PR #2504, opened 2015-02-05
actually read from the buffer returned by rcDataObjRead

### LINKED ISSUE #2503: _ifuseFileCacheRead does not actually read to buf when fileCache->state == NO_FILE_CACHE
For both current master and stable, if `fileCache->state == NO_FILE_CACHE`, _fuseFileCacheRead (in `iFuseLib.FileCache.cpp`) sets `dataObjReadOutBBuf.buf` to be the pointer `buf` specified in its constructor, but when `rcDataObjRead` is called, it eventually calls `read_bytes_buf` in libtcp.cpp, which re-mallocs the buf field of this struct, (line 503).

Upon return from `rcDataObjRead`, `dataObjReadOutBBuf.buf` is never copied back to `buf`.

bpow:
I believe this fixes #2503 (but I'm not sure if there isn't a better way)

### DIFF
--- a/iRODS/clients/fuse/src/iFuseLib.FileCache.cpp
+++ b/iRODS/clients/fuse/src/iFuseLib.FileCache.cpp
@@ -442,8 +442,7 @@ int _ifuseFileCacheRead( fileCache_t *fileCache, char *buf, size_t size, off_t o
 
 
         bzero( &dataObjReadInp, sizeof( dataObjReadInp ) );
-        dataObjReadOutBBuf.buf = buf;
-        dataObjReadOutBBuf.len = size;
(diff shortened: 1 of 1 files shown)


jasoncoposky:
i believe this is fixed in #2538, please confirm.

bpow:
#2538 does appear to fix this. I would have responded earlier, but irodsFuse was broken again in the meantime so I could not verify (see #2554 for that issue and its proposed fix).

trel:
Can you verify now?  We just fixed #2554 with your pull request.

trel:
@bpow awaiting confirmation - will close soon.

bpow:
Sorry about that. This is working for me now in the testing that I have
done, thanks.

PR #2534, opened 2015-02-20
Regression in ability to compile C clients against 4.1.0

keithj:
In running test builds of baton against master I came up against a regression that prevented
compilation. The minimal changes required to get something to compile are in this pull request.
They're not pretty, but they work around the problem.

Added ifdefs to avoid C++ code appearing in headers included by C clients.

Commented one of the two declarations of getAndConnRcatHost. This does
permit a C client to be compiled. However I don't know whether this is
correct.


trel:
thanks!   just cherry-picked into master.

keithj:
This broken again in 5164451

PR #2540, opened 2015-02-24
Fix *_plugin_context::valid() pointer checks

jmarshall:
These `valid()` checks currently always succeed, because `boost::dynamic_pointer_cast<>` and `dynamic_cast<type*>` fail by returning a null pointer rather than by throwing std::bad_cast.

Once that is fixed, `valid().ok()` is still okay because PASSMSG just propagates the `SUCCESS` from the initialisation of `ret`.  This suggested patch uses ERROR instead with what seem like suitable arguments, but I don't know the code and perhaps there's a better approach here.

### DIFF
--- a/iRODS/lib/core/include/irods_auth_plugin_context.hpp
+++ b/iRODS/lib/core/include/irods_auth_plugin_context.hpp
@@ -44,11 +44,9 @@ namespace irods {
 
                 // =-=-=-=-=-=-=
                 // trap case of incorrect type for first class object
-                try {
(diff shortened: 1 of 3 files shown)


jmarshall:
Rebased to current **master**.  It would be great if this PR could be considered, as it will save the next person with a broken shared library build about three hours of frustrating debugging!

trel:
thanks - it's considered!

marked as 4.1.0 (and pinging @tempoz) - and we'll get it in as soon as we confirm the intended behavior.

PR #2554, opened 2015-03-05
Actually use iFuseConn if opened

bpow:
if an iFuseConn can be opened, there is no reason to fail with `SYS_MAX_CONNECT_COUNT_EXCEEDED`. I think the intent is to have this as the `else` for the `if ( listSize( ConnectedConn ) < MAX_NUM_CONN)`.

irodsFuse seems to be completely broken since e36c63ded853a36b0290756b6c0bcf48b741eea8 without this change, despite several subsequent commits. Do people test locally before committing to master?


trel:
Thanks!

Issue #1998, opened 2014-03-21
use openssl MD5 implementation


trel:
remove dependency on local implementation of MD5.

PR #2589, opened 2015-03-30
OS X patch to use CommonCrypto MD5 instead of bundled MD5 implementation

ilarik:
Change in the MD5 code in the native authentication module to solve the OS X dynamic linker runtime conflict of MD5 symbols.

### DIFF
--- a/packaging/build.sh
+++ b/packaging/build.sh
@@ -812,7 +812,7 @@ else
     echo "Detected unixODBC library [$UNIXODBC]"
 fi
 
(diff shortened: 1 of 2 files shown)


jmarshall:
The problem is not unique to libnative.so.

kellerb:
Is removing the bundled implementation of MD5 and using OpenSSL's instead #1998 sufficient to close this issue?

ilarik:
Should be enough for now. Perhaps later on it would be better to migrate to
Apple CommonCrypto MD5 and SHA2 implementations on OS X builds.

PR #2644, opened 2015-04-22
Use printf("%lld") rather than the unusual "%Ld"

jmarshall:
The only warnings produced when compiling the icommands with Clang are

> iRODS/lib/core/src/phybunUtil.cpp:130:40: warning: length modifier 'L' results in undefined behavior or no effect with 'd' conversion specifier [-Wformat]
> `snprintf( tmpStr1, NAME_LEN, "%Ld", rodsArgs->size );`
>
> iRODS/clients/icommands/src/iinit.cpp:403:52: warning: length modifier 'L' results in undefined behavior or no effect with 'd' conversion specifier [-Wformat]
> `"failed to get environment file - %Ld\n",`

Fix these by using `%lld` as is used frequently elsewhere in the codebase.

### DIFF
--- a/iRODS/clients/icommands/src/iinit.cpp
+++ b/iRODS/clients/icommands/src/iinit.cpp
@@ -400,7 +400,7 @@ main( int argc, char **argv ) {
         }
         else {
             printf(
-                "failed to get environment file - %Ld\n",
+                "failed to get environment file - %lld\n",
                 ret.code() );
 
         }
--- a/iRODS/lib/core/src/phybunUtil.cpp
+++ b/iRODS/lib/core/src/phybunUtil.cpp
@@ -127,7 +127,7 @@ initCondForPhybunOpr( rodsArguments_t *rodsArgs,
     }
 
     if ( rodsArgs->sizeFlag == True ) {
-        snprintf( tmpStr1, NAME_LEN, "%Ld", rodsArgs->size );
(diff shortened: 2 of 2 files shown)


kellerb:
Can you confirm if `intmax_t` and `uintmax_t` and their associated format strings `"%ji"` and `"%ju"` are available on the flavor of Mac you are using? If so, I'll make this fix but with those instead.

jmarshall:
OS X compilers have provided a C99 environment for a long time, and `intmax_t`/`%jd` etc work fine.  OTOH last I looked Microsoft's library runtime still does not support `printf("%jd")` — but portability to Windows may or may not be a concern.

kellerb:
Thanks for reporting this.

I pushed the `intmax_t` change.

Feel free to close the issue if this fixes it.

jmarshall:
That does fix it, though recall that for the _iinit.cpp_ one, `ret.code()` in fact **is** a `long long`, so casting seems a bit unnecessary.

BTW when I said these were "the only warnings", I was glossing over the dozens of

> warning: 'foo' has C-linkage specified, but returns user-defined type 'irods::error' which is incompatible with C [-Wreturn-type-c-linkage]

that compiling the plugins produces.

kellerb:
We made a new issue for the c-linkage errors here: #2648

Thanks for the note on the iinit.cpp cast. I tend to prefer explicit `?intmax_t` casts for c formatting calls in case a later refactor misses something. Though if the compiler warns about type mismatches, that might not be necessary anymore.

PR #2666, opened 2015-04-29
Add --verbose option for less quiet builds

jmarshall:
Especially for those not intimately familiar with iRODS source, seeing the actual compilation commands used is important in various debugging scenarios.

Also update the `packaging/build.sh` usage display.  I had been hacking the code that sets `$MAKEJCMD` for ages and was unaware a `--jobs` option had been added.


trel:
This looks good - will rebase and get it in today.

Thanks!

trel:
merged

jmarshall:
Thanks; however see note on 80aa0b6936644a075458971b2619e29c5a9add1c: you can see the explosion by deleting external/irods_external.tar.gz and rerunning `build.sh`.

trel:
should be addressed in 6c1eece0f9

and i've sent the verbosity down to the database plugin as well

Issue #2837, opened 2015-08-14
Memory leak at clearDataObjSqlResult at miscUtil.cpp


iychoi:
Hi, 

Current implementation of fuse uses `rclOpenCollection`/`rclReadCollection`/`rclCloseCollection` functions.
Within these functions, `collHandle_t.dataObjSqlResult.resc_hier.value` field is set to have a pointer to a heap memory (`malloc`'d) and this isn't `free`'d during `clearDataObjSqlResult` function called by `rclCloseCollection`. Also, `dataType` field in the `DataObjSqlResult` structure is not handled.

This caused small memory leak on running iRodsFuse.

kellerb:
Thanks for reporting this. Do you have a commit that fixes this in your repo?

iychoi:
No I didn't fixed it on the same file in my repo since I don't know whether
it is intended for some reasons or missed mistakenly. But I manually
freed it at the fuse side.

kellerb:
We are going to fix this by adding the two missing `free()`'s (`resc_hier.value` and `dataType.value`) to `clearDataObjSqlResult()`

iychoi:
Thanks!

PR #2844, opened 2015-08-19
Performance improvement on iRods Fuse

iychoi:
Handle more corner cases such as malloc failures.
Introduce "preload" which boosts read performance for large files.
Remove and reorganize some locks to provide more concurrency.
Refactor "mysql iCAT driver patch".


trel:
Hi @iychoi, can you please rebase this onto irods:master so:
- your pull request applies cleanly (no conflicts)
- fuse bug fixes are committed first
- new fuse features (preload, etc.) are committed last
- you include no memory leak fixes to rclCloseCollection (has been handled on master via #2837)

Thanks!  We're excited to see these included.

iychoi:
Hi Terrell,

I am making a new pull request based on the latest irods:master. I'll
request again when it is ready so please ignore the request that I made
previously.

Thanks!

PR #2846, opened 2015-08-19
Performance Improvement on iRods Fuse

iychoi:
Introduce "preload" which reads ahead for better file read performance.
Remove unnecessary seek operations at file read and write.
Reorganize locks to allow concurrent file accesses.
Fix potential memory access to a freed memory.


iychoi:
Rebased to fix unidentified authors in commits.

trel:
This passed all the tests - will apply to master ASAP.  Thanks!

PR #2877, opened 2015-09-14
Handle unreachable iCAT or invalid irods environment configuration

iychoi:
This handles unreachable iCAT or invalid irods environment configuration by making a test connection before iRODS FUSE starts FUSE client routine.

Incorrect configuration or unreachable iCAT host will cause iRODS FUSE to terminate immediately with error messages.


trel:
Can you only include the last commit?  It looks like your tracking branch grabbed older things as well.

iychoi:
I trimmed out all previous commits that are already included in irods master repo.

iychoi:
Patch is applied to irods_fuse repo.

https://github.com/irods/irods_client_fuse

PR #3043, opened 2016-02-11
Fixed libodbc detection on Solaris for build

tedgin:
Solaris doesn't use `ldconfig`, so `ld` needs to be used directly.


trel:
@tedgin No more bash scripts.  Is this still relevant in the new CMake/Clang build system?

tedgin:
I doubt it. We don't use Solaris anymore, so feel free to close this PR.

trel:
Oh, the best of answers.   Thanks!

Issue #3092, opened 2016-03-11
list with over 20 elements crashes rule exec


tsmeele:
irods 4.1.8 bug:  rule crashes when it defines a list with more than 20 elements.

Sample code:
```
test{
   *experimentTypes = list("1","2","3","4","5","6","7","8","9","10",
                           "11","12","13","14","15","16","17","18","19","20","21");
}
input *dummy="x"
output ruleExecOut
```
result:
```
libtcp.cpp:240:tcp_read_msg_header :  status [SYS_HEADER_READ_LEN_ERR]  errno [] -- message [read 0 expected 4]
```

Remove one element and the rule execs nicely.

xu-hao:
Microservices support a limited number of arguments. 

A workaround is: you can use the cons microservice to construct lists of arbitrary length. 
cons adds an element and a list into a new list.
For example, cons(1, cons(2, cons(3, list()))). 
I'm working on a patch that provides clearer error messages

PR #3095, opened 2016-03-14
[#3092] fix segfault with more than 20 arguments

xu-hao:


kellerb:
Manually rebased in https://github.com/irods/irods/issues/3092

PR #3240, opened 2016-08-01
Specify paths for each external dependency (XBB-2)

jrandall:
Add ability for user to specify path to each external dependency
by passing in IRODS_EXTERNAL_FULLPATH_${DEPENDENCY_NAME} --
e.g. `cmake ... -DIRODS_EXTERNALS_FULLPATH_AVRO=/usr/local/avro-1.7.7`

Fix missing OPENSSL libs when compiling genOSAuth

Fix missing OPENSSL libs when compiling cll/genq/genu database plugin tests


kellerb:
How did you notice that `genOSAuth` and the database plugin test binaries were missing the OpenSSL libraries?

jrandall:
I'm not sure exactly why it isn't a problem for others. 

My guess would be that is has to do with building in what is probably an as of yet unsupported configuration in which each of my external dependencies are in their own separate path prefix (this is how we manage different versions of software on our systems, bringing them into the environment as needed using Environment Modules). I would guess that most people have a suitable openssl in a system path - in my case there is an openssl in the system path but it is not compatible with the one I am building against. Thus, getting the correct values for `OPENSSL_SSL_LIBRARY` and `OPENSSL_CRYPTO_LIBRARY` is important with this setup, while it may not be necessary if the linker can find a suitable one in a system library path. That's my best guess.

kellerb:
This pull request was automatically closed because the XBB-2 branch was deleted from github, but we are still planning on incorporating this change. You can open a new pull request on the master branch, or I can create a commit for it manually (I can use your email address so you still get credit).

kellerb:
I committed the `IRODS_EXTERNALS_FULLPATH_...` part. I think the SSL linking part should be fixed by the work done in https://github.com/irods/irods/issues/3272, which let us add the "no unlinked symbols at build time" flag to all the shared libraries we produce.

PR #3260, opened 2016-08-18
Aid in debugging plugin-related crash; build tweak

jmarshall:
It seems that the slightest mismatch between how client-side plugins and the client iRODS libraries have been compiled leads to crashes — we haven't fully worked out what's going on here.

In particular, if long-forgotten _rods_environment.json_ settings contain an unexpected `irods_plugins_home` path, the resulting crash can be hard to debug as what's crashed is only vaguely logged (this is what I was complaining about [recently on Twitter](https://twitter.com/jomarnz/status/765342213018619904)).  This adds logging of exactly the plugin used just prior to the crash that potentially occurs when invoking it.

Also: set `CPACK_GENERATOR` appropriately when running on Fedora.


jmarshall:
Ping? :smile:

Could 615979ed8e1ac7ae4f17cbf434919a14aae15cf6 be applied to the **4-1-stable** branch too, please?

kellerb:
Applied the fedora cmake change to master and the plugin loading debug message to master and 4-1-stable

PR #3484, opened 2017-02-21
Fixed issue in irods::configuration_parser::convert_json

jjacquay712:
Booleans in JSON config were always returning true.

http://jansson.readthedocs.io/en/2.7/apiref.html#c.json_boolean

JSON_TRUE and JSON_FALSE will both evaluate to `json_t*` with JSON_TRUE type if passed directly to `json_boolean()` macro.

### DIFF
--- a/lib/core/src/irods_configuration_parser.cpp
+++ b/lib/core/src/irods_configuration_parser.cpp
@@ -127,8 +127,10 @@ namespace irods {
             return boost::any(json_real_value(_val));
 
             case JSON_TRUE:
+            return boost::any(json_true());
+            
             case JSON_FALSE:
-            return boost::any(json_boolean( _val ));
+            return boost::any(json_false());
 
             case JSON_NULL:
             return boost::any(std::string("NULL"));


jacquayj:
Here's more information on how I encountered this bug:

I'm developing a custom rule engine plugin, using this block of code to fetch the JSON config:

```c++

std::string _instance_name = "my_plugin";

const auto& re_plugin_arr = irods::get_server_property< const std::vector< boost::any >& >( 
std::vector< std::string >{ irods::CFG_PLUGIN_CONFIGURATION_KW, irods::PLUGIN_TYPE_RULE_ENGINE } );

for ( const auto& elem : re_plugin_arr ) {
    const auto& plugin_config = boost::any_cast< const std::unordered_map< std::string, boost::any>& >( elem );
    const auto& inst_name = boost::any_cast< const std::string& >( plugin_config.at( irods::CFG_INSTANCE_NAME_KW ) );
    if ( inst_name == _instance_name ) {
        return &plugin_config.at(irods::CFG_PLUGIN_SPECIFIC_CONFIGURATION_KW);
    }
}

```

This is an excerpt of my `/etc/irods/server_config.json`
```json
"rule_engines": [
    {
        "instance_name": "my_plugin-instance",
        "plugin_name": "my_plugin",
        "plugin_specific_configuration": {
            "load_default_rules": false
        }
    }
]
```

When fetching boolean values from the JSON object, in this case `"load_default_rules"`, I was always receiving a `json_t*` (type = JSON_TRUE) regardless of whether the actual JSON in the config file was set to true or false.

I rebuilt `/usr/lib/libirods_common.so.4.2.0` from the 4.2.0 tag, with my changes from this pull request included, and installed the shared object. That resolved the bug for me. Hope this info helps.

Thanks,
John

trel:
Thanks @jacquayj and @jjacquay712!

PR #3494, opened 2017-03-17
using modernize-* checkers from clang-tidy 3.8

trel:
Inspired by today's post:
  https://www.kdab.com/clang-tidy-part-1-modernize-source-code-using-c11c14/

With clang-tidy 3.8:
```
$ clang-tidy --list-checks -checks='*' | grep "modernize"
modernize-loop-convert
modernize-make-unique
modernize-pass-by-value
modernize-redundant-void-arg
modernize-replace-auto-ptr
modernize-shrink-to-fit
modernize-use-auto
modernize-use-default
modernize-use-nullptr
modernize-use-override
```
Eight checkers are included in this pull request.

Two checkers produced no effect:
```
modernize-make-unique
modernize-shrink-to-fit
```


korydraughn:
I think this PR can be closed given that we have a clang-tidy workflow. Also, the codebase has moved since this was created.

trel:
i don't think our current clang-tidy does anything with the modernize checkers.

but agreed - this PR is not helpful as is except for pointers to 'how to' information.

PR #3514, opened 2017-04-12
In usage comment dataObjInp_t is declared as phyPathRegInp then used as dataObjInp.

merimus:
Looks like a minor error in the comment.


trel:
Thanks!

Issue #3593, opened 2017-06-28
reclassify pattern mismatch messages as debug messages


xu-hao:
From:

https://groups.google.com/forum/#!topic/iROD-Chat/pIJ0QeqCMP0

The messages from pattern mismatch in the rule engine should be changed from error message to debug message.

PR #3626, opened 2017-08-16
[#3593] log pattern mismatches as debug messages

xu-hao:


trel:
Thanks @xu-hao.   I've cleanly applied this to 4-2-stable and master, locally.

Any reason you think this shouldn't be good for those branches as well?

xu-hao:
No reason it shouldn't work for 4.2.

Issue #3621, opened 2017-08-04
double free corruption by calling a rule from within a delay block


tedgin:
I'm running iRODS 4.1.10.

I have the following rules.

``` php
_syncReplicas(*CollPath, *DataName) {
  foreach (*cnt in SELECT COUNT(DATA_REPL_NUM)
                   WHERE COLL_NAME = '*CollPath'
                     AND DATA_NAME = '*DataName'
                     AND DATA_REPL_STATUS = '0') {}
}

_scheduleSyncReplicas(*CollPath, *DataName, *DelayTime) {
  delay('<PLUSET>' ++ str(*DelayTime) ++ 's</PLUSET><EF>8h REPEAT UNTIL SUCCESS</EF>') {
    _syncReplicas(*CollPath, *DataName);
    writeLine('serverLog', '_syncReplicas(*CollPath, *DataName)');
  }
}
```

When I trigger the rule `_scheduleSyncReplicas`, a double free error occurs after the `writeLine` statement in the `delay` body. Here's the entry from the reLog file.

```
Aug  4 10:39:37 pid:22449 NOTICE: reServerMain: checking the queue for jobs
Aug  4 10:39:37 pid:16439 NOTICE: writeLine: inString = _syncReplicas(/iplant/home/shared/aegis, clean-irods.txt)
*** glibc detected *** irodsReServer: double free or corruption (!prev): 0x00000000079843f0 ***
```

Here's the corresponding [backtrace](https://github.com/irods/irods/files/1201225/backtrace.txt).

As a work around, the `_syncReplicas` rule can be inlined in the body of the `delay` call. This prevents the error.

xu-hao:
Hi Tony,

Thanks for the bug report. According to Terrell, you are running CentOS 6.
Is that still true?

I tried to reproduce this error on CentOS 6 running iRODS 4.1.10, but
couldn't.

Here is what I did:


On a clean install. Append

_syncReplicas(*CollPath, *DataName) {
  foreach (*cnt in SELECT COUNT(DATA_REPL_NUM)
                   WHERE COLL_NAME = '*CollPath'
                     AND DATA_NAME = '*DataName'
                     AND DATA_REPL_STATUS = '0') {}
}

_scheduleSyncReplicas(*CollPath, *DataName, *DelayTime) {
  delay('<PLUSET>' ++ str(*DelayTime) ++ 's</PLUSET><EF>8h REPEAT
UNTIL SUCCESS</EF>') {
[...]

PR #3627, opened 2017-08-16
[#3621] set no save rei in delayed rules

xu-hao:


trel:
Can you submit a patch for this against 4-2-stable as well?   I didn't see the obvious similar place in the code after the reorganization of 4.2.

xu-hao:
A quick search shows no SAVE_REI in delay execution source code. If there is no SAVE_REI, this patch is no longer necessary.

Issue #3575, opened 2017-06-07
Casting a variable changes its type elsewhere in a rule


TD-er:
I am not entirely sure what triggers the error, but I can display it with this code snippet:
```
stateToInt(*in_State, *out_intValue) {
  # initialize to error value
  *out_intValue = "-1";
  if (*in_State == "Error")     { *out_intValue = "-1";}
  if (*in_State == "0")         { *out_intValue = "0";}
  if (*in_State == "NotSet")    { *out_intValue = "0";}
  if (*in_State == "New")       { *out_intValue = "1";}
  if (*in_State == "Allocated") { *out_intValue = "2";}
  writeLine("serverLog","[stateToInt] *in_State => *out_intValue");
}


testBugIntCast(*Coll, *State) {
  *Newstate_intvalue = "0";
  stateToInt(*State, *Newstate_intvalue);
  if (*Newstate_intvalue == "-1") {
      writeLine("serverLog","Yay");
  } else {
      if (0 < int(*Newstate_intvalue)) {
        writeLine("serverLog","else....");
      }
  }
}
```

The error is:
```
gijs@gijs-X556UB:~/alta_svn/src/irods/ingest$ irule -F testBugIntCast.r 
ERROR: rcExecMyRule error.  status = -1230000 RE_TYPE_ERROR
Level 0: DEBUG: simplify: unsolvable typing constraint INTEGER * STRING < FLEX VAR 724{INTEGER DOUBLE STRING } * FLEX VAR 724{INTEGER DOUBLE STRING }.

line 220, col 6, rule base altaRules
  if (*Newstate_intvalue == "-1") {
      ^
```

And my testBugIntCast.r:
```
testBugIntCastRule {
  *Newstate_intvalue = "0";
  setState(*Coll, "New");
  testBugIntCast(*Coll, "New");
}
INPUT *Coll="/$rodsZoneClient/home/$userNameClient/sub1"
OUTPUT ruleExecOut
```

The problem is that this if-statement only works when '-1' is not a string.
  if (*Newstate_intvalue == -1) {

When the cast using int() is omitted, the if statement can operate on the string "-1"
If you remove the call to stateToInt(), the *Newstate_intvalue is seen as a string.

Is this a bug, or am I missing something here?

trel:
Hi @TD-er, what version of iRODS is this?

cc @xu-hao

TD-er:
iRODS 4.2.0.
[...]

PR #3632, opened 2017-08-22
[#3575] type ascription

xu-hao:
This add a feature to the rule language that allows specifying the type of an expression to avoid the type checker from using the default type that may cause type errors.


trel:
This PR broke federation in CI.  Investigating.

PR #3633, opened 2017-08-22
itrim 4-1-stable

jassigill2000:


tests/pydevtest/resource_suite.py
```diff
@@ -1013,3 +1013,36 @@ def test_itrim_displays_correct_count__ticket_3531(self):
...
         if os.path.exists(filepath):
             os.unlink(filepath)
+
+    def test_itrim_returns_on_negative_status__ticket_3531(self):
+        # local setup
+        filename = "filetotesterror.txt"
+        filepath = lib.create_local_testfile(filename)
+
+        itrimReplResc = "itrimReplResc"
+        self.admin.assert_icommand("iadmin mkresc {itrimReplResc} replication".format(**locals()), 'STDOUT_SINGLELINE', "replication")
```
kellerb: initial comment

tests/pydevtest/resource_suite.py
```diff
@@ -1013,3 +1013,36 @@ def test_itrim_displays_correct_count__ticket_3531(self):
...
+        resc1 = self.testresc
+        resc2 = self.anotherresc
+
+        self.user0.assert_icommand("iput -R {resc1} {filename}".format(**locals()), 'EMPTY')
+
+        # check if file was added
+        self.user0.assert_icommand("ils -L", 'STDOUT_SINGLELINE', filename)
+
+        # add resources to the replNode
```
kellerb: added comment

kellerb: reply to comment

trel:
Please combine the two code commits and then separately, combine the two test commits, and I'll merge them in.

kellerb:
It looks like if [the return value of `rcObjStat()`](https://github.com/irods/irods/pull/3633/commits/3f0ee3fc549feebe8780d41fc3f65fbeb06c2631#diff-a2236a709de79723104aaff13644e685R113) is positive, then it will be one of [these values](https://github.com/irods/irods/blob/4.1.10/iRODS/lib/core/include/rodsType.h#L42-L50). Instead of checking if `objStatus` is positive, may want to check instead if it is equal to `DATA_OBJ_T` and error out if it is not (since we expect the object being trimmed to be a data object). Otherwise everything seems good.

PR #3640, opened 2017-08-25
mem leak

xu-hao:


trel:
please remove the first two commits from this PR, just the mem leak fix

xu-hao:
fixed the problems from previous pr

Issue #3641, opened 2017-08-25
packing instruction memory leak


xu-hao:
- [x] master
- [X] 4-2-stable

----

while fixing #2934:
```
14 bytes in 1 blocks are definitely lost in loss record 373 of 1,139
   at 0x4C2AB80: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
   by 0xA381B49: strdup (strdup.c:42)
   by 0x9B80496: xmlStrToStr (packStruct.cpp:1365)
   by 0x9B81909: unpackXmlString (packStruct.cpp:1913)
   by 0x9B81310: unpackString (packStruct.cpp:1854)
   by 0x9B81045: unpackNonpointerItem (packStruct.cpp:1707)
   by 0x9B8059F: unpackItem (packStruct.cpp:1658)
   by 0x9B7B141: unpackChildStruct (packStruct.cpp:2492)
   by 0x9B7AEEF: unpackStruct (packStruct.cpp:72)
   by 0x6B3F09B: readMsgHeader(boost::shared_ptr<irods::network_object>, msgHeader*, timeval*) (sockComm.cpp:207)
   by 0x6B077A6: irods::read_client_server_negotiation_message(boost::shared_ptr<irods::network_object>, boost::shared_ptr<irods::cs_neg_t>&) (irods_client_negotiation.cpp:455)
   by 0x6712577: irods::client_server_negotiation_for_server(boost::shared_ptr<irods::network_object>, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >&) (irods_server_negotiation.cpp:148)

55 bytes in 4 blocks are definitely lost in loss record 726 of 1,139
   at 0x4C2AB80: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
   by 0xA381B49: strdup (strdup.c:42)
   by 0x9B80496: xmlStrToStr (packStruct.cpp:1365)
   by 0x9B81909: unpackXmlString (packStruct.cpp:1913)
   by 0x9B81310: unpackString (packStruct.cpp:1854)
   by 0x9B81045: unpackNonpointerItem (packStruct.cpp:1707)
   by 0x9B8059F: unpackItem (packStruct.cpp:1658)
   by 0x9B7B141: unpackChildStruct (packStruct.cpp:2492)
   by 0x9B7AEEF: unpackStruct (packStruct.cpp:72)
[...]

PR #3644, opened 2017-08-29
fix mem leak

xu-hao:


trel:
does this address all of #3641 ?

xu-hao:
Yes, it does.

PR #3651, opened 2017-09-05
hash to detect update to rules

xu-hao:


plugins/rule_engines/irods_rule_engine_plugin-irods_rule_language/configuration.cpp
```diff
@@ -304,136 +309,247 @@ int unlinkFuncDescIndex() {
...
         rSplitStr( r2, r1, NAME_LEN, r3, RULE_SET_DEF_LENGTH, ',' );
-        if (getRuleBasePath( r1, fn )==nullptr) {
-            return -1;
+        irbs.push_back(r1);
+        snprintf( r2, sizeof( r2 ), "%s", r3 );
+    }
+    return irbs;
+}
+
+#define HASH_BUF_SZ (1024*1024)
```
trel: new magic number?

xu-hao: source: https://github.com/irods/irods/blob/master/lib/hasher/src/checksum.cpp#L16

Issue #3679, opened 2017-10-12
[20170929-scanbuild] Fix bug type: Dereference of null pointer


jaking92:
master

- [x] lib/core/src/packStruct.cpp, line 47
- [x] lib/core/src/packStruct.cpp, line 1932
- [x] plugins/auth/native/libnative.cpp, line 404*
- [x] plugins/auth/native/libnative.cpp, line 476*
- [x] plugins/auth/native/libnative.cpp, line 513*
- [x] plugins/auth/native/libnative.cpp, line 517*
- [x] plugins/network/ssl/libssl.cpp, line 1062
- [x] plugins/network/ssl/libssl.cpp, line 1064
- [x] plugins/network/ssl/libssl.cpp, line 1075
- [x] plugins/network/ssl/libssl.cpp, line 1079
- [x] plugins/network/ssl/libssl.cpp, line 1090
- [x] plugins/network/ssl/libssl.cpp, line 1093
- [x] plugins/resources/deferred/libdeferred.cpp, line 800*
- [x] plugins/resources/load_balanced/libload_balanced.cpp, line 936*
- [x] plugins/resources/mockarchive/libmockarchive.cpp, line 697*
- [x] plugins/resources/mockarchive/libmockarchive.cpp, line 700*
- [x] plugins/resources/nonblocking/libnonblocking.cpp, line 1439*
- [x] plugins/resources/nonblocking/libnonblocking.cpp, line 1442*
- [x] plugins/resources/nonblocking/libnonblocking.cpp, line 1450*
- [x] plugins/resources/random/librandom.cpp, line 801*
- [x] plugins/resources/unixfilesystem/libunixfilesystem.cpp, line 1532*
- [x] plugins/resources/unixfilesystem/libunixfilesystem.cpp, line 1537*
- [x] plugins/resources/unixfilesystem/libunixfilesystem.cpp, line 1544*
- [x] plugins/rule_engines/irods_rule_engine_plugin-irods_rule_language/arithmetics.cpp, line 1052
- [x] plugins/rule_engines/irods_rule_engine_plugin-irods_rule_language/arithmetics.cpp, line 1134
- [x] plugins/rule_engines/irods_rule_engine_plugin-irods_rule_language/utils.cpp, line 831
- [x] server/api/src/rsCollCreate.cpp, line 163
- [x] server/api/src/rsDataObjRename.cpp, line 118
- [x] server/core/src/dataObjOpr.cpp, line 1784
[...]

PR #3691, opened 2017-10-20
[#3679] Fix dereference of null pointers in packStruct.cpp

jaking92:
extendPackedOutput is being used to populate outPtr in two places. If this function fails, outPtr is explicitly set to null. This change adds checks to ensure that extendPackedOuput did not fail (and, therefore, that outPtr is not null) before dereferencing outPtr.

### DIFF
--- a/lib/core/src/packStruct.cpp
+++ b/lib/core/src/packStruct.cpp
@@ -42,8 +42,12 @@ packStruct( const void *inStruct, bytesBuf_t **packedResult, const char *packIns
 
     if ( irodsProt == XML_PROT ) {
         void *outPtr;
+        status = extendPackedOutput( packedOutput, 1, outPtr ); 
+        if ( SYS_MALLOC_ERR == status ) {
+            return status;
+        }
+
         /* add a NULL termination */
-        extendPackedOutput( packedOutput, 1, outPtr );
         *static_cast<char*>(outPtr) = '\0';
         if ( getRodsLogLevel() >= LOG_DEBUG9 ) {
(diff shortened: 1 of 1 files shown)


jaking92:
Jenkins tests passed for this.

lib/core/src/packStruct.cpp
```diff
@@ -1917,19 +1919,18 @@ unpackXmlString( const void *&inPtr, packedOutput_t &unpackedOutput, int maxStrL
...
     }
-    else {
-        extendPackedOutput( unpackedOutput, maxStrLen, outPtr );
-    }
 
-    if ( myStrlen > 0 ) {
-        strncpy( static_cast<char*>(outPtr), strBuf, myStrlen );
-        outStr = static_cast<char*>(outPtr);
-        outPtr = static_cast<char*>(outPtr) + myStrlen;
+    if ( SYS_MALLOC_ERR != extendPackedOutput( unpackedOutput, extLen, outPtr ) ) {
```
jaking92: Would it be better to return SYS_MALLOC_ERR on failure instead of continuing on? There are several "Argument with 'nonnull' attribute passed null" bugs in this file that are almost all related to this situation

jasoncoposky: I would agree that you should return SYS_MALLOC_ERR if that is the result of extendedPackedOutput().  Can that function return any other possible error?  Why not capture the error in a variable and test and return that?

jaking92: extendedPackedOutput() can only return 0 or SYS_MALLOC_ERR. I will add a variable to capture the error. Thanks!

PR #3698, opened 2017-10-25
[#3679] Fix dereference of null pointers in execMicroService3

jaking92:
When freeing the argument vector, need to make sure the args structures are not null before dereferencing to get the exprType. exprType should be non-null, since it is a pointer in the node structure which is dereferenced.


jaking92:
Jekins test failed:
`[test-cen6-oracle] Failed in branch test-cen6-oracle`

jaking92:
Jenkins tests passed

PR #3699, opened 2017-10-25
[#3679] Add null pointer checks in a few files

jaking92:
All null pointer checks are roughly the same. Grouped together since there were no real logic path changes and so review is a little easier to do.

**In l3Mkdir**:
Fixed situation where dataObjInfo could be null when passed into l3Mkdir. The intention was likely to only call l3Mkdir if specColl was null or colClass == LINKED_COLL, not when dataObjInfo itself is null.

**In updateInEnv:**
updateInEnv takes an Env pointer as argument, and the pointer can be null. Since the entire function depends on env, ensure that it is not null before proceeding with any operations.

**In getDataObjInfoIncSpecColl:**
This change adds a null check before dereferencing dataObjInfo or checking specColl.

Also fixed constants going on the left in change with SHA: eaab3c4549b954d2df701117d8a7b631449998fb

### DIFF
--- a/plugins/network/ssl/libssl.cpp
+++ b/plugins/network/ssl/libssl.cpp
@@ -1056,7 +1056,7 @@ irods::error ssl_send_rods_msg(
                 // =-=-=-=-=-=-=-
                 // send the message buffer
                 int bytes_written = 0;
-                if ( _msg_buf != NULL &&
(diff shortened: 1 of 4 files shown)


jaking92:
Jenkins tests failed:
`[test-ub14-oracle] Failed in branch test-ub14-oracle`

PR #3700, opened 2017-10-25
[#3679] Fix null pointer dereference in rsDataObjRename

jaking92:
This change fixes a situation where the specColl pointer member of srcDataObjInfo can be dereferenced when null because only srcDataObjInfo is null-checked before the dereference.

Slight change in the if-else tree to make it easier to read.


jaking92:
Jenkins tests failed:
`[test-ub12-oracle] Failed in branch test-ub12-oracle`
`[test-ub16-oracle] Failed in branch test-ub16-oracle`

PR #3702, opened 2017-10-26
[#3679] Fix dereference of null pointer in ssl_send_rods_msg (#1)

jaking92:
_msg_buf, _error_buf, and _stream_bbuf are parameters of the function and are not null-checked before dereferencing when printing to log or when attempting to write to socket.

This change adds the needed null checks to accompany the zero-length buffer checks such that the result returned by the function is the same for either case.

The buffer lengths "should" never be non-zero when _msg_buf, _error_buf, and _stream_bbuf are null, so the null pointer dereferences "should" never happen. This adds an extra layer of protection.

(cherry-picked from SHA: eaab3c4549b954d2df701117d8a7b631449998fb)

### DIFF
--- a/plugins/network/ssl/libssl.cpp
+++ b/plugins/network/ssl/libssl.cpp
@@ -1056,7 +1056,8 @@ irods::error ssl_send_rods_msg(
                 // =-=-=-=-=-=-=-
                 // send the message buffer
                 int bytes_written = 0;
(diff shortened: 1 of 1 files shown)


trel:
got it - will hold until after 4.2.2 is out.

jaking92:
This passed Jenkins tests on Oct. 26. I can run it through again, if desired.

Issue #3706, opened 2017-10-29
occasional test failure in Test_Resource_Replication.test_num_repl_policy__ticket_2851


trel:
- [x] master
- [x] 4-2-stable

--
irods.test.test_resource_types.Test_Resource_Replication.test_num_repl_policy__ticket_2851

Sometimes, an assertion error occurs:

```
Traceback (most recent call last):
  File "/var/lib/irods/scripts/irods/test/test_resource_types.py", line 3569, in test_num_repl_policy__ticket_2851
    assert len(hier_ctr) == 3
AssertionError
```

seen on:
- ub16-mysql

Not enough determinism in this assertion:
  https://github.com/irods/irods/blob/130279ec1e4181af0daed1925750af4f1bba2068/scripts/irods/test/test_resource_types.py#L3568-L3569

Need a better test.

jaking92:
Just saw this on cen7-mysql

trel:
and on ub16-psql

trel:
rewrite test to just count the number of replicas for each of `num_repl=1` through `num_repl=3`.

PR #3713, opened 2017-10-31
[#3679] Add null pointer checks in a few files

jaking92:
In l3Mkdir:
Added early return in l3Mkdir if dataObjInfo is null.

In updateInEnv:
updateInEnv takes an Env pointer as argument, and the pointer can be
null. Since the entire function depends on env, ensure that it is
not null before proceeding with any operations.

In getDataObjInfoIncSpecColl:
This change adds a null check before dereferencing dataObjInfo or
checking specColl.

Also fixed constants going on the left in change with SHA:
eaab3c4549b954d2df701117d8a7b631449998fb

### DIFF
--- a/server/api/src/rsCollCreate.cpp
+++ b/server/api/src/rsCollCreate.cpp
@@ -147,6 +147,10 @@ rsCollCreate( rsComm_t *rsComm, collInp_t *collCreateInp ) {
 
 int
 l3Mkdir( rsComm_t *rsComm, dataObjInfo_t *dataObjInfo ) {
+    if ( NULL == dataObjInfo ) {
+        return SYS_NULL_INPUT;
+    }
+
     //int rescTypeInx;
     fileMkdirInp_t fileMkdirInp;
     int status;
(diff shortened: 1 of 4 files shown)


jaking92:
Passed Jenkins tests. This was meant to be an amendment to this PR: https://github.com/irods/irods/pull/3699

server/api/src/rsCollCreate.cpp
```diff
@@ -147,6 +147,10 @@ rsCollCreate( rsComm_t *rsComm, collInp_t *collCreateInp ) {
 
 int
 l3Mkdir( rsComm_t *rsComm, dataObjInfo_t *dataObjInfo ) {
+    if ( NULL == dataObjInfo ) {
+        return -1;
```
trel: is there a better value to return than -1?  @jasoncoposky @tempoz

jasoncoposky: use SYS_NULL_INPUT instead.

jaking92:
Replaced return value. All Jenkins tests passed except for failure matching the one described in #3706 (ub16-psql).

Issue #3701, opened 2017-10-25
[20170929-scanbuild] Fix bug type: Memory leak


jaking92:
**master:**
- [x] lib/core/src/packStruct.cpp, line 1917
- [x] lib/core/src/rcPortalOpr.cpp, line 214
- [x] lib/core/src/rcPortalOpr.cpp, line 910
- [x] plugins/rule_engines/irods_rule_engine_plugin-irods_rule_language/functions.cpp, line 882
- [x] plugins/rule_engines/irods_rule_engine_plugin-irods_rule_language/index.cpp, line 29
- [x] plugins/rule_engines/irods_rule_engine_plugin-irods_rule_language/index.cpp, line 273
- [x] server/core/src/rodsServer.cpp, line 243
- [x] server/core/src/miscServerFunct.cpp, line 1418
- [x] server/core/src/miscServerFunct.cpp, line 1465
- [x] server/core/src/miscServerFunct.cpp, line 1594
- [x] server/re/src/extractAvuMS.cpp, line 223
---
**4-2-stable:**
- [x] lib/core/src/packStruct.cpp, line 1917
- [x] lib/core/src/rcPortalOpr.cpp, line 214
- [x] lib/core/src/rcPortalOpr.cpp, line 910
- [x] plugins/rule_engines/irods_rule_engine_plugin-irods_rule_language/functions.cpp, line 882
- [x] plugins/rule_engines/irods_rule_engine_plugin-irods_rule_language/index.cpp, line 29
- [x] plugins/rule_engines/irods_rule_engine_plugin-irods_rule_language/index.cpp, line 273
- [x] server/core/src/rodsServer.cpp, line 243
- [x] server/core/src/miscServerFunct.cpp, line 1418
- [x] server/core/src/miscServerFunct.cpp, line 1465
- [x] server/core/src/miscServerFunct.cpp, line 1594
- [x] server/re/src/extractAvuMS.cpp, line 223
---
**Bug Group:** Memory Error
**Bug Type:** Memory leak
**Quantity:** 11

PR #3715, opened 2017-11-03
[#3701] Fix missing frees on early returns

jaking92:
There are several places where the program can return early due to some kind of error. Make sure any allocated memory is freed up before returning.


jaking92:
Jenkins tests passed

Issue #3678, opened 2017-10-12
[20170929-scanbuild] Fix bug type: Argument with 'nonnull' attribute passed null


jaking92:
**master:**
- [x] lib/core/src/packStruct.cpp, line 1222
- [x] lib/core/src/packStruct.cpp, line 1225
- [x] lib/core/src/packStruct.cpp, line 1266
- [x] lib/core/src/packStruct.cpp, line 1269
- [x] lib/core/src/packStruct.cpp, line 1376
- [x] lib/core/src/packStruct.cpp, line 1840
- [x] lib/core/src/packStruct.cpp, line 1881
- [x] lib/core/src/packStruct.cpp, line 1884
- [x] lib/core/src/packStruct.cpp, line 1928
- [x] lib/core/src/packStruct.cpp, line 2939
- [x] server/core/src/physPath.cpp, line 1284
---
**4-2-stable:**
- [x] lib/core/src/packStruct.cpp, line 1222
- [x] lib/core/src/packStruct.cpp, line 1225
- [x] lib/core/src/packStruct.cpp, line 1266
- [x] lib/core/src/packStruct.cpp, line 1269
- [x] lib/core/src/packStruct.cpp, line 1376
- [x] lib/core/src/packStruct.cpp, line 1840
- [x] lib/core/src/packStruct.cpp, line 1881
- [x] lib/core/src/packStruct.cpp, line 1884
- [x] lib/core/src/packStruct.cpp, line 1928
- [x] lib/core/src/packStruct.cpp, line 2939
- [x] server/core/src/physPath.cpp, line 1284
---
**Bug Group:** API
**Bug Type:** Argument with 'nonnull' attribute passed null
**Quantity:** 11

Issue #3689, opened 2017-10-19
occasional test failure in Test_Native_Rule_Engine_Plugin.test_network_pep


trel:
irods.test.test_native_rule_engine_plugin.Test_Native_Rule_Engine_Plugin.test_network_pep

Sometimes, an assertion error occurs:

```
Traceback (most recent call last):
  File "/var/lib/irods/scripts/irods/test/test_native_rule_engine_plugin.py", line 45, in test_network_pep
    self.helper_test_pep(rule_texts[self.plugin_name][self.class_name][inspect.currentframe().f_code.co_name], "iput -f --metadata ATTR;VALUE;UNIT "+self.testfile)
  File "/var/lib/irods/scripts/irods/test/test_native_rule_engine_plugin.py", line 42, in helper_test_pep
    assert number_of_strings_to_look_for == count, 'Found {0} instead of {1} occurrences of {2}'.format(count, number_of_strings_to_look_for, s)
AssertionError: Found 2 instead of 1 occurrences of THIS IS AN OUT VARIABLE
```

tempoz:
Just saw this on ub16 oracle

PR #3724, opened 2017-11-10
[#3678] Fix null parameters going into non-null arguments

jaking92:
This fixes a few places where null parameters are being passed into functions which do not accept null parameters for their arguments.


jaking92:
Jenkins tests passed except for failure described in #3689

Issue #3711, opened 2017-10-30
[scanbuild] Fix bug group: Dead store


jaking92:
- [x] server/core/src/rodsAgent.cpp, line 93
- [x] server/core/src/objMetaOpr.cpp, line 130
- [x] server/core/src/initServer.cpp, line 152
- [x] plugins/rule_engines/irods_rule_engine_plugin-irods_rule_language/functions.cpp, line 1255
- [x] plugins/resources/replication/librepl.cpp, line 1481
- [x] plugins/database/src/db_plugin.cpp, line 10733
- [x] lib/core/src/rcMisc.cpp, line 3924
- [x] server/api/src/rsRmColl.cpp, line 379
- [x] server/api/src/rsRmColl.cpp, line 386
- [x] server/api/src/rsRmColl.cpp, line 389
- [x] Other dead assignments (156)
---
**Bug Group:** Dead store
**Bug Type:** Dead increment
**Bug Type:** Dead initialization
These are minor bugs and so will only go into master. Dead store detected in zmq header file has been removed from the list as this has been addressed upstream (i.e. updating zmq should fix this).

jaking92:
https://github.com/irods/externals/issues/11 created to track upgrade of zmq.

PR #3726, opened 2017-11-10
[#3711] Fix dead initializations and increments

jaking92:
Fixed many places where variables were being initialized and assigned a new value before the initial value was read.

Note: Not to go to 4-2-stable.


jaking92:
Passed Jenkins tests

jaking92:
Seems reasonable. I only had in mind to maintain the current functionality (for most cases) but it would probably be better to do the checking.

trel:
Since we know cppcheck is now being run by codacy... I'm going to merge this and trust the tools :)

Issue #3712, opened 2017-10-30
[scanbuild] Fix bug type: Garbage results/returns/assignments


jaking92:
**master:**
- [x] plugins/database/src/db_plugin.cpp, line 644
- [x] plugins/database/src/mid_level_routines.cpp, line 686
- [x] plugins/database/src/mid_level_routines.cpp, line 838
- [x] plugins/database/src/mid_level_routines.cpp, line 884
- [x] plugins/database/src/mid_level_routines.cpp, line 1050
- [x] plugins/database/src/mid_level_routines.cpp, line 1099
- [x] server/core/src/rodsAgent.cpp, line 254
---
**4-2-stable:**
- [x] plugins/database/src/db_plugin.cpp, line 644
- [x] plugins/database/src/mid_level_routines.cpp, line 686
- [x] plugins/database/src/mid_level_routines.cpp, line 838
- [x] plugins/database/src/mid_level_routines.cpp, line 884
- [x] plugins/database/src/mid_level_routines.cpp, line 1050
- [x] plugins/database/src/mid_level_routines.cpp, line 1099
- [x] server/core/src/rodsAgent.cpp, line 254
---
**Bug Group:** Logic error
**Bug Type:** Assigned value is garbage or undefined
**Bug Type:** Garbage return value
**Bug Type:** Result of operation is garbage or undefined
**Quantity:** 7

jaking92:
Paths taken by static analyzer for all bugs found in plugins/database/src/mid_level_routines.cpp are not possible. In all cases, iVal is claimed to have never received a value and then is returned (garbage return). The path taken specifies "Assuming 'status' is >= 0" in each case, where status = cmlGetIntegerValueFromSql(...). cmlGetIntegerValueFromSql cannot return a value greater than 0, so status must be 0. One level down, cmlGetOneRowFromSqlBV returns error codes (<0) or 1. If 1, cmlGetIntegerValueFromSql returns 0 or an error code. If 0, iVal received a value. Since status cannot be more than 0 and the static analyzer assumes status >= 0, then iVal must have received a value. Thoughts?

plugins/database/src/db_plugin.cpp and server/core/src/rodsAgent.cpp are legitimate concerns, and I should have a patch up soon.

PR #3734, opened 2017-11-13
[#3712] Fix garbage assignments/operations/returns

jaking92:
Fixed a couple of places in code where potentially uninitialzed variables were having their values assigned or used in some way.

Passed Jenkins tests except for the failure described in #3689 on ub16-oracle.


server/core/src/rodsAgent.cpp
```diff
@@ -250,14 +250,13 @@ int initRsCommFromServerSocket( rsComm_t *rsComm, int socket ) {
...
-        setLocalAddr( rsComm->sock, &rsComm->localAddr );
-        setRemoteAddr( rsComm->sock, &rsComm->remoteAddr );
+    if ( receiveSocketFromSocket( socket, &newSocket ) > 0 ) {
+        rsComm->sock = newSocket;
+        if ( 0 != rsComm->sock ) {
+            /* remove error messages from xmsLog */
+            setLocalAddr( rsComm->sock, &rsComm->localAddr );
+            setRemoteAddr( rsComm->sock, &rsComm->remoteAddr );
+        }
     }
```
trel: need an else() here?

jaking92: Upon closer inspection, looks like this function (initRsCommFromServerSocket) is not being referenced anywhere. We could fix it up or just remove it. Thoughts, @jasoncoposky?

trel: not used anywhere... torch it.

jaking92:
Jenkins tests passed for new commit.

Issue #3705, opened 2017-10-27
iquest returning DATA_ID when asked for DATA_RESC_HIER


trel:
- [x] master
- [x] 4-2-stable

iRODS 4.2.1

When asking for a DATA_RESC_HIER alongside DATA_NAME, the wrong column DATA_RESC_ID is shown:
```
$ iquest "select DATA_NAME, DATA_RESC_HIER where DATA_RESC_HIER = 'demoResc'"
DATA_NAME = bar
DATA_RESC_ID = 10014
------------------------------------------------------------
DATA_NAME = foo
DATA_RESC_ID = 10014
------------------------------------------------------------
```

But when only asked for DATA_RESC_HIER, it is returned as expected:
```
$ iquest "select DATA_RESC_HIER where DATA_RESC_HIER = 'demoResc'"
DATA_RESC_HIER = demoResc
------------------------------------------------------------

$ iquest "select DATA_RESC_HIER where DATA_NAME = 'foo'"
DATA_RESC_HIER = demoResc
------------------------------------------------------------

$ iquest "select DATA_RESC_HIER where DATA_NAME = 'boom'"
DATA_RESC_HIER = pt1;pt2;u1
------------------------------------------------------------
```

trel:
add to scripts/irods/test/test_iquest.py:
```python
    def test_iquest_data_name_with_data_resc_hier__3705(self):
        filename = 'test_iquest_data_name_with_data_resc_hier__3705'
        lib.make_file(filename, 1)
        self.admin.assert_icommand(['iput', filename])
        self.admin.assert_icommand(['iquest', "select DATA_NAME, DATA_RESC_HIER where DATA_RESC_HIER = '{0}'".format(self.admin.default_resource)], 'STDOUT_SINGLELINE', 'DATA_RESC_HIER')
        os.remove(filename)
```

jaking92:
Additional/potentially related observation:

$ iquest "select DATA_RESC_ID where DATA_RESC_HIER = 'demoResc'"
DATA_RESC_HIER = demoResc
------------------------------------------------------------

PR #3737, opened 2017-11-21
[#3705] Fix iquest not returning DATA_RESC_HIER

jaking92:
Adds some tests for iquest in DATA_RESC_HIER which expose an issue
where the hier name is not being substituted correctly in the output.

If DATA_RESC_HIER is used in a GenQuery, the DATA_RESC_ID for the
hierarchy is used instead in the actual query and then replaced at the end
with DATA_RESC_HIER again. There is a bug where selecting DATA_RESC_HIER
as a column and using the DATA_RESC_HIER as a filter causes the replacement
of the ID at the end to fail. This change fixes this issue by not
allowing the inputs for the filter to override the saved position of
the column in the output.


jaking92:
Codacy does not like the asserts that I put in, apparently, despite the fact that there are asserts in many places in the test code. I can look into alternatives, if desired.

trel:
Codacy is complaining about bytecode - which is fine since we are interested in the assertion itself, not the speed/efficiency of the Python itself.  There may be a way to make codacy happy, but let’s worry about that later.  Looks good to me.

jaking92:
Passed Jenkins tests.

trel:
good stuff - please cherry pick these two to 4-2-stable in a new PR and we'll get it closed.

PR #3738, opened 2017-11-21
[#3679] Fix null pointer dereferences from ASSERT_ERROR/PASS

jaking92:
Fixes the remaining usages of ASSERT_ERROR/ASSERT_PASS which can cause null pointer dereferences in certain code paths.

Jenkins tests passed.

Note: I can try to tackle breaking up native_auth_agent_response (see comment at top of function) but I thought fixing the potential null pointer dereferences (per scanbuild) would be a better first step.


trel:
agreed, first step, make it quiet... then worry about refactoring.

PR #3745, opened 2017-12-01
[#3679] Fix null pointer dereferences in libnative.cpp

jaking92:
Even though it is not a valid code path, the static analyzer has detected some potential null pointer dereferences.

Passed Jenkins tests.

Note: This should be the last of the null pointer dereferences from scanbuild. Must have missed these in the last round.


trel:
fix typo "derefences" in commit message

Issue #3675, opened 2017-10-04
support for cockroachdb


xu-hao:
add a "generic" option for database plugin

Pete-C:
+1

trel:
I think this is complete.  Full test suite passes.

Leaving open until we get cockroachdb into CI.

PR #3764, opened 2017-12-15
[#3675] add generic db type

xu-hao:
add generic db type


server/core/src/irods_generic_object.cpp
```diff
@@ -0,0 +1,111 @@
...
+            std::stringstream msg;
+            msg << "generic_object does not support a [";
+            msg << _interface;
+            msg << "] plugin interface";
+            return ERROR( SYS_INVALID_INPUT_PARAM, msg.str() );
+
+        }
+
+        // =-=-=-=-=-=-=-
+        // ask the database manager for a postgres resource
```
trel: postgres?

xu-hao: thanks for catching that. will update the commit

Issue #3722, opened 2017-11-10
irule output in iRODS 4.2.2


ccacciari:
- [x] master
- [x] 4-2-stable
-----

Hi,

I have just deloyed iRODS 4.2.2 on a centos 7.4.
When I execute the following rule:

```
$ irule -vF EUDATReplication_TestRule2.r
rcExecMyRule: myTestRule {

    *source = "/cinecaDMPZone2/home/claudio/test_replica.txt";
    *destination= "/cinecaDMPZone1/home/claudio#cinecaDMPZone2/test_replica2.txt";
    *recursive = "true";
    *registered = "true";
    *dest_res = "cinecaRes1";
    *status = EUDATReplication(*source, *destination, *dest_res, *registered, *recursive, *response);

    if (*status) {
        writeLine("stdout", "Success!");
    }
    else {
        writeLine("stdout", "Failed: *response");
    }

}

outParamDesc: ruleExecOut
ExecMyRule completed successfully.    Output
```

The rule completes succesfully, but I cannot see any message in standard output.
The EUDATReplication rule is quite complex, but the variable *status is just a boolean, while the variable *response is a string. On the server side I see this error:

```
Nov 10 10:23:16 pid:10648 remote addresses: 127.0.0.1, 130.xxx.xx.xxx, ::1 ERROR: addMsParam: NULL msParamArray or label input
Nov 10 10:23:16 pid:10648 remote addresses: 127.0.0.1, 130.xxx.xx.xxx, ::1 ERROR: addMsParam: NULL msParamArray or label input
```
It worked in 4.2.1.

trel:
That ERROR is coming from this line:
  https://github.com/irods/irods/blob/master/lib/core/src/msParam.cpp#L54

The type checking is a bit tighter now.  Sorry the output is so sparse.

You'll need to look into EUDATReplication to find where a parameter or label is being passed without being set (or it's set to NULL).

ccacciari:
I dig a bit more into my EUDATReplication rule and I discovered that the problem is related to the remote execution microservice.
In version 4.2.2, if I execute the following rule in the local zone 130.186.2.1:
```
$ cat myTestRule.r
myTestRule {

        remote("130.186.2.2", "<ZONE>myRemoteZone</ZONE>") {
            writeLine("serverLog", " -- Remote Zone Execution");
        }

}
INPUT *myvar="hello"
OUTPUT ruleExecOut

$irule -F myTestRule.r
[...]

PR #3780, opened 2018-01-09
[#3722] Fix segfault in remote rule execution

jaking92:
Remote rule execution will fail in versions >4.2.1 due to a segmentation fault in convertHashtableToMsParamArray. To prevent this, msParamArray is being generated for rei before calling the offending function in the event that it is made null.

Tests write to the server log using the remote execution microservice. First test writes to the local zone's log. Second test writes to the remote zone's log (configured in test_federation).

Passed Jenkins tests.


trel:
Looks good - consider 'Hastable' -> 'Hashtable'

PR #3811, opened 2018-02-03
fix --run_s bug caused by adding --skip_until

xu-hao:


scripts/run_tests.py
```diff
@@ -58,49 +59,45 @@ def optparse_callback_federation(option, opt_str, value, parser):
 def run_tests_from_names(names, buffer_test_output, xml_output, skipUntil):
     loader = unittest.TestLoader()
     suites0 = [loader.loadTestsFromName('irods.test.' + name) for name in names] # test files used to be standalone python packages, now that they are in the irods python module, they cannot be loaded directly, but must be loaded with the full module path
```
trel: can we change this to just 'suites'?

scripts/run_tests.py
```diff
@@ -58,49 +59,45 @@ def optparse_callback_federation(option, opt_str, value, parser):
...
+            if len(d["filtered_markers"]) == 0:
+                d["filtered_markers"] = filter(lambda marker: filter_testcase(suite, marker[0]), markers)
+            if len(d["filtered_markers"]) != 0:
+                suitelist.append(suite)
+                d["filtered_markers"] = filter(lambda marker: marker[-1] == "" or not filter_testcase(suite, marker[-1]), d["filtered_markers"])
+        else:
+            for subsuite in suite:
+                filter_testsuite(subsuite)
+
+    filter_testsuite(suites0)
```
trel: just 'suites'?

scripts/run_tests.py
xu-hao: will do

PR #3823, opened 2018-02-21
A candidate fix for host names longer than 64 characters.

HowardLander:
4.2.2 Ubuntu 7

### What did you try to do?
Install iRODS on an VM declared in azure and run the test suite for the newly developed azure driver

When you get a vm on azure the irods hostname code in  lib.py code grabs the fully qualified name for the host.  Unfortunately this can be longer than the existing iRODS 64 character limit. The patch seems to resolve this issue.

### DIFF
--- a/scripts/irods/lib.py
+++ b/scripts/irods/lib.py
@@ -32,10 +32,15 @@ from . import execute
 from . import paths
 
 # get the fully qualified domain name
-#(no, really, getfqdn() is insufficient)
+#(no, really, getfqdn() is insufficient), but some environments (Azure for example) produce
+# hostnames longer than the irods max of 64 characters. If that happens, we revert to the "ordinary"
+# hostname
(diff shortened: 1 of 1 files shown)


HowardLander:
It turns out that there is another, arguably better solution to this problem. Here's the text from the relevant section of the README.md file of the irods azure driver (https://github.com/heliumdatacommons/irods_resource_plugin_azure)

Here's a better work around that seems to fix the problems:

Step one: assign a shorter, more reasonable hostname for the azure client. Follow this link to explain how to do this:

https://docs.microsoft.com/en-us/azure/virtual-machines/linux/portal-create-fqdn

Step two: Once you've got your new hostname, make sure the host uses it with the command

sudo hostnamectl set-hostname new-host-name

trel:
Got it - thanks.

PR #3843, opened 2018-03-09
Add functionality for generic auth object

theferrit32:
Fall through in auth factory to resolve a string value to a library containing an arbitrary subclass of generic_auth_object. 

Utilized by an openid auth plugin here: https://github.com/irods-contrib/irods_auth_plugin_openid


jasoncoposky:
Reviewed and looks good to merge.

Issue #3658, opened 2017-09-12
iscan should trim trailing slash


trel:
```
$ time ireg -R newResc -C /tmp/thesourceondisk/ /tempZone/home/public/destcollection
```

This surprisingly registers `/tmp/thesourceondisk//LICENSE` into the catalog (with double slash).

The double slash makes `iscan` unhappy since the single slash version of the same path reports as not being registered.

The paths need to be canonicalized and/or the trailing slash should be trimmed before registration.

TD-er:
Similar behavior also occurs with iput.

The behavior of iput is very counter intuitive when it comes to trailing slashes.
When adding a trailing slash to destination (collection)path, its behavior changes to including the parent dir of the source.
Something like reported in this issue may happen when including both trailing slash in source path and destination path.

Well, if it isn't a bug, it is at least a very hard to document  feature ;)

andrew-irods:
Summarizing the root factors of this issue:

1. The ireg utility passes to the server a physical folder path which may be terminated by a "/" character, depending on the parameter passed with the "-C" flag.

2. When this (the above) happens, that trailing slash makes it all the way into the catalog maintained by the server as the physical path of a folder being associated with a specific resource.

3. When recreating the physical paths of all files within that specific folder, the physical path of the folder is taken from the catalog and pre-pended (including the "/" if there were one) to the filename (which is kept in the catalog without a path).  This operation always puts a slash "/" between the path of the directory and the file name, which sometimes results in a "/path//file" construct, which is the root cause of this issue. 

To fix the issue, two things are done:

In lib/core/src/rodsPath.cpp (line# 268), a check is made for trailing slashes in the physical directory path, and they are removed if they're there. (This is the code path for dealing with "ireg -C ......").

In server/api/src/rsPhyPathReg.cpp (line# 730), a check is made by the server for trailing slashes in the physical directory path, and they are removed if they're there before a /filename is appended to that string.  This takes care of physical directory paths associated with resources in the catalog pre-dating this fix. 

This addresses all three causes described above.

andrew-irods:
CI Test:  Build #683 - http://172.25.14.125:8080/view/2.%20Personal/job/irods-build-and-test-workflow/683/ 

(This run has a single test failure unrelated to code changes. Past that, a single underscore was added to the test method name in test_log_levels.py).

Issue #3865, opened 2018-03-30
irodsServer: Environment variable spLogSql does not produce output in rodsLog


andrew-irods:
- [ ] master
- [x] 4-2-stable
-------

### iRODS Version, OS and Version
irods 4-2-stable
Ubuntu 16.04 

### What did you try to do?
export spLogSql=1
irodsctl restart

### Expected behavior
Plenty of log lines in rodsLog labeled "LOG_SQL: ", showing various sql operations at runtime

### Observed behavior
Not a single sql log line was created after irodsServer was restarted.

### Steps to reproduce

export spLogSql=1
irodsctl restart

andrew-irods:
Fixed: Immediately at the beginning of main() in rodsServer (around line #174 of server/core/src/rodsServer.cpp), the environment is queried for the existence of "spLogSql", and if it exists, the internal log level for sql is set by calling rodsLogSqlReq(1).

andrew-irods:
CI Test:  Build #683 - http://172.25.14.125:8080/view/2.%20Personal/job/irods-build-and-test-workflow/683/ 

(This run has a single test failure unrelated to code changes. Past that, a single underscore was added to the test method name in test_log_levels.py).

andrew-irods:
| iCommand           | return code | Error Message                             |
|:-------------------|------------:|------------------------------------------:|
| iCommand           | return code | Error Message                             |

andrew-irods:
We need to fill out this functional matrix:

| iCommand           | return code | Error Message | Comment                   |
|:-------------------|:-----------:|:--------------|:--------------------------|
|                    |             |               |                           |
| ils -L ldir     |             |                  | missing pfile in pdir represented by ldir    |
|                    |             |               |                           |
| ils -L lfile     |             |                 | missing pfile equivalent to lfile   |
|                    |             |               |                           |
| iscan -r pdir      |             |               | missing local pfile in pdir     |
|                    |             |               |                           |
| iscan pfile   |             |                    | missing local pfile    |
|                    |             |               |                           |
| iscan -rd ldir |             |                   | missing local pfile in pdir equivalent to ldir   |
|                    |             |               |                           |
| iscan -d lfile    |             |                | missing local pfile equivalent to lfile        |
|                    |             |               |                           |

Where:

ldir is logical irods directory

pdir is physical local directory

lfile is logical irods file

pfile is physical local file

Issue #858, opened 2013-11-26
General Housekeeping - Whitespace, typos, test cleanup


trel:
_Originally Opened: @trel (2012-06-22 16:54:25_)

This is my housekeeping bug.

It's terribly exciting.
## 

From: https://code.renci.org/gf/project/e-irods/tracker/?action=TrackerItemEdit&tracker_item_id=858

PR #3870, opened 2018-04-05
Collection of bug fixes: [#3658] [#858] [#3865]

andrew-irods:
Resolves:

#3658 irods/irods ireg should trim trailing slash

In lib/core/src/rodsPath.cpp (around line# 268), a check is made for trailing slashes in the physical directory path, and they are removed if they're there. (This is the code path for dealing with "ireg -C ......").

In server/api/src/rsPhyPathReg.cpp (line# around 730), a check is made by the server for trailing slashes in the physical directory path, and they are removed if they're there before a /filename is appended to that string. This takes care of physical directory paths associated with resources in the catalog pre-dating this fix.

Test case:  test_ireg_double_slashes__issue_3658() in scripts/irods/test/test_ireg.py

Continuous Integration:  Build 713 - http://172.25.14.125:8080/view/2.%20Personal/job/irods-build-and-test-workflow/713/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 tags
 .dir-locals.el
+.project
+.cproject
+*.swp
 *.pyc
--- a/lib/core/src/rodsPath.cpp
+++ b/lib/core/src/rodsPath.cpp
@@ -265,6 +265,23 @@ addSrcInPath( rodsPathInp_t *rodsPathInp, const char *inPath ) {
         return USER__NULL_INPUT_ERR;
     }
 
+    // Issue #3658 - This section removes trailing slashes from the 
+    // incoming directory path in an invoking icommand (e.g. ireg).
+    //
+    // TODO:  This is a localized fix that addresses any trailing slashes
+    // in inPath. Any future refactoring of the code that deals with paths
+    // would most probably take care of this issue, and the code segment
+    // below would/should then be removed.
+
+    char tmpStr[MAX_NAME_LEN];
+
+    rstrcpy( tmpStr, inPath, MAX_NAME_LEN );
+    size_t len = strlen(tmpStr);
+    for (size_t i = len-1; i > 0 && tmpStr[i] == '/'; i--) {
+        tmpStr[i] = '\0';
+    }
+    const char *modInPath = const_cast<const char *>(tmpStr);   // end of 3658 fix
+
     if ( ( rodsPathInp->numSrc % PTR_ARRAY_MALLOC_LEN ) == 0 ) {
         newNumSrc = rodsPathInp->numSrc + PTR_ARRAY_MALLOC_LEN;
         newSrcPath = ( rodsPath_t * ) malloc( newNumSrc * sizeof( rodsPath_t ) );
@@ -285,7 +302,7 @@ addSrcInPath( rodsPathInp_t *rodsPathInp, const char *inPath ) {
     else {
         newSrcPath = rodsPathInp->srcPath;
     }
-    rstrcpy( newSrcPath[rodsPathInp->numSrc].inPath, inPath, MAX_NAME_LEN );
+    rstrcpy( newSrcPath[rodsPathInp->numSrc].inPath, modInPath, MAX_NAME_LEN );
     rodsPathInp->numSrc++;
 
     return 0;
@@ -478,3 +495,4 @@ clearRodsPath( rodsPath_t *rodsPath ) {
 
     return;
 }
+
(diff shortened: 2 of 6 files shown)


trel:
way too many commits - please squash, and re-submit.

consider a single pull request for each issue.

andrew-irods:
Squashed too many commits as requested.
I'm leaving this as a single pull request since the two issues had more than one commit interspersed with each other.  Will keep issues separate in the future.  Thanks.

.gitignore
```diff
@@ -1,2 +1,5 @@
 tags
 .dir-locals.el
+.project
+.cproject
+*.swp
```
trel: these are unrelated to this issue...

move this edit to a different PR against #858

server/api/src/rsPhyPathReg.cpp
```diff
@@ -727,7 +727,32 @@ dirPathReg( rsComm_t *rsComm, dataObjInp_t *phyPathRegInp, char *filePath,
         fileStatInp_t fileStatInp;
         memset( &fileStatInp, 0, sizeof( fileStatInp ) );
 
-        snprintf( fileStatInp.fileName, MAX_NAME_LEN, "%s/%s", filePath, rodsDirent->d_name );
+        // Issue #3658 - https://github.com/irods/irods/issues/3658
+        //
+        // TODO:  This is a localized fix that addresses any trailing slashes
```
trel: why TODO?

server/api/src/rsPhyPathReg.cpp
```diff
@@ -727,7 +727,32 @@ dirPathReg( rsComm_t *rsComm, dataObjInp_t *phyPathRegInp, char *filePath,
...
+        // in inPath. Any future refactoring of the code that deals with paths
+        // would most probably take care of this issue, and the code segment
+        // below would/should then be removed.
+        //
+        // This is the second part of a two part fix.  This section removes trailing
```
trel: reduce comment to just:
`This section removes trailing slashes from the directory path in the server when the path is already in the catalog.`

server/api/src/rsPhyPathReg.cpp
```diff
@@ -727,7 +727,32 @@ dirPathReg( rsComm_t *rsComm, dataObjInp_t *phyPathRegInp, char *filePath,
...
+        rstrcpy( tmpStr, filePath, MAX_NAME_LEN );
+        size_t len = strlen(tmpStr);
+
+        for (size_t i = len-1; i > 0 && tmpStr[i] == '/'; i--) {
+            tmpStr[i] = '\0';
+        }
+
+        snprintf( fileStatInp.fileName, MAX_NAME_LEN, "%s/%s", tmpStr, rodsDirent->d_name );
+
+        //   For debug purposes (issue# 3658)
```
trel: remove debug code

server/api/src/rsPhyPathReg.cpp
```diff
@@ -727,7 +727,32 @@ dirPathReg( rsComm_t *rsComm, dataObjInp_t *phyPathRegInp, char *filePath,
         fileStatInp_t fileStatInp;
         memset( &fileStatInp, 0, sizeof( fileStatInp ) );
 
-        snprintf( fileStatInp.fileName, MAX_NAME_LEN, "%s/%s", filePath, rodsDirent->d_name );
+        // Issue #3658 - https://github.com/irods/irods/issues/3658
```
trel: this will be visible due to history - no need to declare in the comment.

lib/core/src/rodsPath.cpp
```diff
@@ -265,6 +265,27 @@ addSrcInPath( rodsPathInp_t *rodsPathInp, const char *inPath ) {
         return USER__NULL_INPUT_ERR;
     }
 
+    // Issue #3658 - https://github.com/irods/irods/issues/3658
```
trel: entire comment should just be:
`This section removes trailing slashes from the incoming directory path in an invoking icommand (e.g. ireg).`

andrew-irods: Theses are my opinions:  

The function addSrcInPath() is called from parseCmdLinePath() which is used by a total of nineteen icommands.   Since this removal of a trailing slash from the directory path affects so many utilities, I think it's a good idea to be very verbose here about what's being changed and why. 

I don't think that assuming that the developer looking at this would figure out to go look at the source code control history, or at the git issue detail, is a reasonable assumption. 

In addition, while doing this we (Jason, you, me) discussed the issue of normalizing the treatment of all paths in the future.  We agreed (or I thought we did) that this (removal of trailing slashes in various places in the code) deserves a TODO section that would remind the developers who are doing the normalizing, to clean up (what would then be) unnecessary editing of the path string in various places in the code.  This comment also relates to your comments in rsPhyPathReg.cpp above.

Give me the final verdict, and I'll follow your direction.  Thanks.

trel:
So close - one more pass through rebase to bring up to date.

No need for another Jenkins run.

Issue #3436, opened 2017-01-05
add retry capability to replication resource


kript:
- [x] master
- [x] 4-2-stable
- [x] 4-1-stable

---
Related to #3432 and #3409, we have ended up with ~4k objects with no replica, in part because the iput that put them in failed to replicate (it was targeting an iRES that wasn't accepting connections due to a memory leak, we _think_ (but can't prove!), so wasn't marked as down until too late). 

Tagging @keithj for his input as to the exact failure mode if I've mis-stated it..

How do we best make the replication re-try, and if this is up to end user configuration, how is it suggested to implement this with a composite resource tree that contains a replication leaf - I could see delayed rules being an option, but its not quite the same thing...

keithj:
There were also `iput` failures that left single replicates in cases where network latency caused the connection to one resource server to time out, but not the other.

trel:
I don't think I understand the question.

If the system has a good copy/replica somewhere - and the composable resource tree has a replication resource in it specifying that the policy is to have two (or more) good copies... then the answer is to have `iadmin modresc COMPOSABLE_RESC_TREE_ROOT rebalance` run either in a cronjob or a periodic delayed rule.

This will bring all replicas in that tree into compliance with the policy.

kript:
This issue is around what happens when an iput fails due to replication, and whether we could make the replication re-try if that is what would have caused the failure.

trel:
We should retry the replication.

The context string of the replication resource would grow three new eligible settings:
```
retry_attempts=1
first_retry_delay_in_seconds=1
backoff_multiplier=1
```

Each would default to 1.

This would give a default system a single retry attempt where it currently has 0.

backoff_multiplier is a float that would affect the retry delay after the first attempt.

trel:
The retry loop should write to `LOG_NOTICE` to help admins know that the underlying system is behaving intermittently.

jaking92:
`test_scenario` causes test framework to be angry. Fixing name.

trel:
checkboxes and close!

PR #3871, opened 2018-04-05
[#3436] Add retry capability to replication resource

jaking92:
This change adds a configurable retry mechanism to the replication resource. The replication resources can be configured via 3 new settings in the context string of the resource. The retry applies when the replication from an iput or a replication rebalance fails. The default value for each setting is 1, which means a single retry after 1 second has passed.

Passed 4 test runs on Jenkins. Two of the runs hit #3689 for ub16-oracle. One run hit #3706 for topo-cen6-psql-resc-ssl.


jaking92:
Sorry for all the Codacy noise. That should do it. Please don't merge yet - need to add pound sign to issue number in commit message(s) to complete. Trying to reduce noise in referenced issue. :)

PR #3888, opened 2018-04-19
[#3689] Extend RE sleep time for test_network_pep

jaking92:
Rule engine server occasionally wakes up (every 30s by default) and causes this test to erroneously record two messages to the log where it is expecting only one, causing the test to fail.

This change adjusts rule_engine_server_sleep_time_in_seconds in server_config.json to a longer period of time to avoid waking up while the test is running. The setting is restored after the test completes.

--
Jenkins tests passed before rebase on top of 96107c5cc79d47aead0eb0273a8874124d4056dc. Running again.

### DIFF
--- a/scripts/irods/test/test_native_rule_engine_plugin.py
+++ b/scripts/irods/test/test_native_rule_engine_plugin.py
@@ -6,6 +6,7 @@ else:
     import unittest
 
 import inspect
+import json
 import os
 import time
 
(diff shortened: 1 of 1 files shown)


jaking92:
Jenkins tests passed.

trel:
please again on 4-2-stable.   thanks.

PR #3889, opened 2018-04-20
[#3689] Extend RE sleep time for test_network_pep

jaking92:
Rule engine server occasionally wakes up (every 30s by default) and causes this test to erroneously record two messages to the log where it is expecting only one, causing the test to fail.

This change adjusts rule_engine_server_sleep_time_in_seconds in server_config.json to a longer period of time to avoid waking up while the test is running. The setting is restored after the test completes.

(cherry-picked from SHA: b6391778bd2f8a100b688bc84c3225f805746950)

--
Jenkins tests passed.

### DIFF
--- a/scripts/irods/test/test_native_rule_engine_plugin.py
+++ b/scripts/irods/test/test_native_rule_engine_plugin.py
@@ -6,6 +6,7 @@ else:
     import unittest
 
 import inspect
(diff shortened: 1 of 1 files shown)


trel:
add the # and we're good to go

PR #3898, opened 2018-04-24
[#3706] More determinism in test_repl_num_policy

jaking92:
test_repl_num_policy was trying to assert both the number of replicas being made based on the num_repl context string and the destination resources of those replicas. As the destination resource is
determined by voting (locality of reference, order in the catalog, etc.), the test was technically non-deterministic and could fail on occasion.

This change modifies the test to just count the number of replicas being produced by the replication resource to ensure that the num_repl policy is being enforced.

--
Waiting for one more CI test. Do not submit until # is added to issue number in commit message.

### DIFF
--- a/scripts/irods/test/test_resource_types.py
+++ b/scripts/irods/test/test_resource_types.py
@@ -3557,22 +3557,14 @@ class Test_Resource_Replication(ChunkyDevTest, ResourceSuite, unittest.TestCase)
         filename = "test_num_repl_policy__ticket_2851.txt"
         filepath = lib.create_local_testfile(filename)
 
-        for i in range(0,10):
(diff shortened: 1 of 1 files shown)


trel:
Is it sufficient to simply count the number of replicas reported from `ils -l`, even without marking any resources down?   We just want to see 2, not 3, and don't care where they are?

Artificially limiting the count to 2 by marking 'extra' resources 'down' seems to lean on the voting rather than the 'only make 2 replicas' part of this code.

jaking92:
Fair point. I will write the test to satisfy that condition. Thanks!

jaking92:
Passed CI tests.

Issue #3580, opened 2017-06-14
ipasswd issues (was: stty path hard-coded in clientLogin.cpp)


onmh:
- [x] master
- [x] master icommands
- [x] 4-2-stable
- [x] 4-2-stable icommands
-----

This issue has been restated and expanded to include other issues found in ipasswd during implementation of the bug fix.  Associated issues:  #2797  (ipasswd behavior) and #3910 (code cleanup).

The original issue focused on the "stty" utilty being hardcoded as "/bin/stty" in clientLogin.cpp. 

While digging into this issue, other issues emerged, and the issue has been expanded to deal with them:

1.  Remove all implementations of "/bin/stty" from the C/C++ code.  Instead use termios(3) - tcgetattr() and tcsetattr() to turn echo on and off instead.  **DONE**

2. A possible endless loop exists in the code which asks for new password in ipasswd.  Limit it to three tries, and then abort. **DONE**

3. Part of the endless-loop issue is caused by not distinguishing between EOF, error condition, or empty line entered (using std::getline). **DONE**

4. When entering a new password which is short, the message asks for at least 3 characters, but in fact only accepts passwords that are at least 4 characters long.  Modify the behavior so that the 3-count is implemented correctly.  **DONE**

5. There is only partial implementation of the "-e" flag (echo the password as entered).  The flag is impelented in the original challenge for a password, but completely ignored while entering the new password, and echo is turned off regardless of that flag. This has to be fixed. **DONE**

6. In obf.cpp the fgets() of the password input is not checked for error or EOF. Fix. **DONE**

7. Identify all other places where tcgetattr() and tcgetattr() are used to control echo on/off, and improve to include the above functionality (see full list in a comment below) - **See new issue #3910 submitted including uses of repetitious code that works.**

Lastly, the CI test for ipasswd has to be drastically expanded to cover most basic functionality.  **DONE - see attached spreadsheet in comment below.**

The original statement of the issue is below the line.  It is completely addressed by item (1) above:
--------------------------------------------------------------------

Path to stty is hard coded in clientLogin.cpp:

```
        path p( "/bin/stty" );

            if ( 0 != system( "/bin/stty -echo 2> /dev/null" ) ) {

            if ( 0 != system( "/bin/stty echo 2> /dev/null" ) ) {

```
Some systems (e.g. NixOS) don't have /bin/stty and login throws an error:

`sh: /bin/stty: No such file or directory`

The call could be replaced by:

`/usr/bin/env stty -echo`

[...]

PR #3919, opened 2018-05-03
[#3580] ipasswd issues (was: stty path hard-coded in clientLogin.cpp)

andrew-irods:
CI test successful: http://172.25.14.125:8080/view/2.%20Personal/job/irods-build-and-test-workflow/861/

### DIFF
--- a/lib/core/include/termiosUtil.hpp
+++ b/lib/core/include/termiosUtil.hpp
@@ -1,0 +1,95 @@
+#ifndef _TERMIOS_UTIL_HPP_
+#define _TERMIOS_UTIL_HPP_
+
+#include <unistd.h>
+#include <errno.h>
+#include <termios.h>
+
+#ifdef __cplusplus
+
+namespace irods
+{
+    // class termiosUtil - at this time, this class encapsulates the work
+    // which has to be done in order to turn echo on or off for a specifc
+    // file descriptor.  The call can be expanded to encapsulate other
+    // functionality of termios(3).
+
+    class termiosUtil {
+        public:
+            // Class has to be instantiated with a file descriptor.
+            // The default constructor is invalid.
+            explicit termiosUtil(int fd) :
+                valid_(false),
+                errno_copy_(0),
+                fd_(fd),
+                save_c_lflag_(0)
+            {
+                memset(&tty_, 0, sizeof(tty_));
+                if (isatty(fd_))
+                {
+                    if (tcgetattr( fd_, &tty_ ) == 0)
+                    {
+                        save_c_lflag_ = tty_.c_lflag;
+                        valid_ = true;
+                    }
+                    else
+                    {
+                        errno_copy_ = errno;
+                        valid_ = false;
+                    }
+                }
+                else
+                {
+                    errno_copy_ = errno;
+                    valid_ = false;
+                }
+            };
+
+        public:
+            virtual ~termiosUtil(void) = default;
+            int getError(void) const     { return errno_copy_; };
+            bool getValid(void) const     { return valid_; };
+
+            bool echoOff(void)
+            {
+                if (!valid_)
+                {
+                    return false;
+                }
+                tty_.c_lflag &= ~ECHO;
+                if (tcsetattr( fd_, TCSANOW, &tty_ ) == 0)
+                {
+                    return true;
+                }
+                errno_copy_ = errno;
+                valid_ = false;
+                return false;
+            }
+
+            bool echoOn(void)
+            {
+                if (!valid_)
+                {
+                    return false;
+                }
+                tty_.c_lflag = save_c_lflag_;
+                if (tcsetattr( fd_, TCSANOW, &tty_ ) == 0)
+                {
+                    return true;
+                }
+                errno_copy_ = errno;
+                valid_ = false;
+                return false;
+            }
+
+       private:
+            bool valid_;            // fd_ >= 0, etc
+            int errno_copy_;        // if fd_ is not a tty, or other termios error, this has the error
+            int fd_;                // File descriptor for this tty_ structure
+            struct termios tty_;
+            tcflag_t save_c_lflag_;
+    };
+}; // namespace irods
+
+#endif // __cplusplus
+#endif // _TERMIOS_UTIL_HPP_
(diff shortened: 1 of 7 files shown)


lib/core/src/obf.cpp
```diff
@@ -289,30 +292,34 @@ obfSavePw( int promptOpt, int fileOpt, int printOpt, const char *pwArg ) {
 
     envVal = obfiGetEnvKey();
 
-    if ( strlen( pwArg ) == 0 ) {
-
+    if ( pwArg == NULL || strlen( pwArg ) == 0 ) {
```
trel: more Yoda?

lib/core/src/obf.cpp
```diff
@@ -289,30 +292,34 @@ obfSavePw( int promptOpt, int fileOpt, int printOpt, const char *pwArg ) {
...
+            {
+                printf( "WARNING: Error %d disabling echo mode. Password will be displayed in plaintext.\n", tiosutl.getError() );
             }
         }
 
         printf( "Enter your current iRODS password:" );
-        if ( NULL == fgets( inbuf, MAX_PASSWORD_LEN + 50, stdin ) ) {
-            // end of line reached or no input
-        }
+        const char *fgets_return = fgets( inbuf, MAX_PASSWORD_LEN + 50, stdin );
```
trel: magic number?

lib/core/include/termiosUtil.hpp
```diff
@@ -0,0 +1,99 @@
...
+                else
+                {
+                    errno_copy_ = errno;
+                    valid_ = false;
+                }
+            };
+
+        private:
+            // Invalid constructor.  Ensures instantiation with fd parameter
+            termiosUtil(void) {};
```
korydraughn: Your constructor accepting a file descriptor will stop the compiler from generating a default constructor, so the private default constructor can be removed.

lib/core/include/termiosUtil.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+        public:
+            virtual ~termiosUtil(void) = default;
```
korydraughn: Is it necessary to declare the destructor as virtual?  Looking over this pull request, I haven't spotted a need for this.  Are you expecting this class to be used as a base?

lib/core/include/termiosUtil.hpp
```diff
@@ -0,0 +1,99 @@
...
+                errno_copy_ = errno;
+                valid_ = false;
+                return false;
+            }
+
+       private:
+            bool valid_;            // fd_ >= 0, etc
+            int errno_copy_;        // if fd_ is not a tty, or other termios error, this has the error
+            int fd_;                // File descriptor for this tty_ structure
+            struct termios tty_;
```
korydraughn: Is `struct` needed here?

lib/core/include/termiosUtil.hpp
```diff
@@ -0,0 +1,99 @@
...
+        private:
+            // Invalid constructor.  Ensures instantiation with fd parameter
+            termiosUtil(void) {};
+
+        public:
+            virtual ~termiosUtil(void) = default;
+            int getError(void) const     { return errno_copy_; };
+            bool getValid(void) const     { return valid_; };
+
+            bool echoOff(void)
```
korydraughn: `void` is probably not needed.  An empty set of parenthesis is pretty clear.

lib/core/src/obf.cpp
andrew-irods: Discuss this we did. Change this I will not.

lib/core/include/termiosUtil.hpp
```diff
@@ -0,0 +1,99 @@
+#ifndef _TERMIOS_UTIL_HPP_
+#define _TERMIOS_UTIL_HPP_
+
+#include <unistd.h>
+#include <errno.h>
+#include <termios.h>
+
+#ifdef __cplusplus
```
korydraughn: Why is this needed?

lib/core/src/obf.cpp
andrew-irods: "Not my fault".  I just added the examination of the return value.  But I know what you mean.

lib/core/include/termiosUtil.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+namespace irods
+{
+    // class termiosUtil - at this time, this class encapsulates the work
+    // which has to be done in order to turn echo on or off for a specifc
+    // file desctiptor.  The call can be expanded to encapsulate other
```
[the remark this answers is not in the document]
andrew-irods: Will change.  Thanks.

lib/core/include/termiosUtil.hpp
andrew-irods: Will change.  Thanks.

lib/core/include/termiosUtil.hpp
andrew-irods: Yes yes and yes.  I don't know that it won't be used as a base class 2 years from now.  (My opinion, which means that this stays).

lib/core/include/termiosUtil.hpp
andrew-irods: No it's not.  It's staying (it's a choice, not a rationale).

lib/core/include/termiosUtil.hpp
andrew-irods: Again - this is a choice.  It stays.

lib/core/include/termiosUtil.hpp
andrew-irods: Over abundance of being careful.

lib/core/include/termiosUtil.hpp
```diff
@@ -0,0 +1,95 @@
...
+                {
+                    errno_copy_ = errno;
+                    valid_ = false;
+                }
+            };
+
+        public:
+            virtual ~termiosUtil(void) = default;
+            int getError(void) const     { return errno_copy_; };
+            bool getValid(void) const     { return valid_; };
```
korydraughn: After reviewing the pull request for the icommand [ipasswd], I think `getValid()` should be changed to `valid()` or `isValid()`.

Issue #3928, opened 2018-05-08
Throw error when root directory (/) is used as a vault path


jaking92:
- [x] master
- [x] 4-2-stable
- [x] 4-1-stable
---
## Feature

Using the root directory as a vault path for a resource is a very bad idea. We should issue a warning to the user when this is done (similar to the warning issued when using `localhost` as a hostname for a resource).

There should also be an error when the root directory is used to update/calculate free space on a resource (see discussion in #3841).

jaking92:
We need to skip test_unix_filesystem_free_space_on_root__3928 when the native rule language rule engine plugin is not being used because it breaks the test suite for the python rule engine plugin (and probably others).

trel:
skipped on 4-2-stable and master (no rule engine plugins in 4.1.x) - closing.

PR #3929, opened 2018-05-10
[#3928] Error when root dir used with vault paths (4-1-stable)

jaking92:
When msi_update_unixfilesystem_resource_free_space() updates the free space on a non-existent vault path (or one that boost::filesystem interprets as non-existent), it checks the parent directories of the resource vault path until it finds one that exists, even if it is the root directory ('/' on UNIX systems).

Root directory should not be used as a vault path. An error is now produced if the vault path is set to root directory. Additionally, the root path should not be used in updating the free space on a resource. This change causes msi_update_unixfilesystem_resource_free_space() to log an error and exit if the path_to_stat becomes the root directory because this gives misleading information about the free space on a resource.

---
CI tests passed.

Could not find any portable way to determine that a path is the root directory, so open to suggestions there (see line 569 of plugins/database/src/db_plugin.cpp). Also, potentially could move the change "higher up" (rcGeneralAdmin/rsGeneralAdmin) but db_plugin is where the `localhost` warning is being generated. Not partial to one or the other.

### DIFF
--- a/plugins/database/src/db_plugin.cpp
+++ b/plugins/database/src/db_plugin.cpp
@@ -563,6 +563,17 @@ _resolveHostName(rsComm_t* _rsComm, const char* _hostAddress) {
     return 0;
 }
 
+// Returns success if path is not root; otherwise, generates an error
+irods::error
+verify_non_root_vault_path(irods::plugin_context& _ctx, const std::string& path) {
+    if (0 == path.compare("/")) {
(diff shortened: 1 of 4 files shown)


jaking92:
Running through CI again after some minor changes. Still need # as well.

jaking92:
CI tests passed.

Issue #3932, opened 2018-05-14
Large rulesets processed very slowly at startup


dworkin:
- [x] master
- [x] 4-2-stable

---
### iRODS Version, OS and Version
iRODS master/4.2.2, Centos 7.4

### What did you try to do?

Start iRODS with about 9000 lines of rule code.

### Expected behavior

We expect the server to be available instantaneously.

### Observed behavior (including steps to reproduce, if applicable)

The server is busy for about a minute with 100% CPU usage.  Commands are not handled during this time, but will start another irodsServer process that also uses 100% CPU for a minute.

Investigating the issue showed that iRODS is busy in copyCache, trying to squeeze 249643 items into a hash table of size 100.
```
    Hashtable *objectMap = newHashTable( 100 );

    MK_PTR( RuleSet, coreRuleSet );
```

PR #3933, opened 2018-05-14
Significantly enlarge objectMap hashtable to speed up rule caching.

dworkin:
See https://github.com/irods/irods/issues/3932

### DIFF
--- a/plugins/rule_engines/irods_rule_engine_plugin-irods_rule_language/cache.cpp
+++ b/plugins/rule_engines/irods_rule_engine_plugin-irods_rule_language/cache.cpp
@@ -36,7 +36,7 @@ Cache *copyCache( unsigned char **p, size_t size, Cache *ptr ) {
 
     MK_POINTER( &( ecopy->address ) );
     MK_POINTER( &( ecopy->pointers ) );
-    Hashtable *objectMap = newHashTable( 100 );
+    Hashtable *objectMap = newHashTable( size / 40 );
 
     MK_PTR( RuleSet, coreRuleSet );
     ecopy->coreRuleSetStatus = COMPRESSED;


plugins/rule_engines/irods_rule_engine_plugin-irods_rule_language/cache.cpp
```diff
@@ -36,7 +36,7 @@ Cache *copyCache( unsigned char **p, size_t size, Cache *ptr ) {
 
     MK_POINTER( &( ecopy->address ) );
     MK_POINTER( &( ecopy->pointers ) );
-    Hashtable *objectMap = newHashTable( 100 );
+    Hashtable *objectMap = newHashTable( size / 40 );
```
trel: this is great - thank you.

two questions:

- why divided by 40, specifically?
- what happens if size is <40? (do we need some check/guard before using `size`)

trel: This explanation is great - I definitely appreciate the answer.

Regarding my <40 question, I meant what happens if `size / 40` is less than 1 and therefore goes to zero... (`newHashTable( 0 )`)... Do we need an additional check/guard for that case?

dworkin: When copyCache is called with size < 40, it will start writing memory out of bounds 4 lines earlier:
```
    allocateInBuffer( Cache, ecopy, ptr );
```
since the size of the Cache struct is well over 40. So while I don't think that an *additional* guard is warranted for this case, a *general* one would be useful at the start of the function:
```
    if ( size < DEFAULT_BLOCK_SIZE || size % REGION_ALIGNMENT != 0 ) { /* size should be large enough and divisible by ALIGNMENT */
        return NULL;
    }
```

Issue #3927, opened 2018-05-07
Add PID/hostname info to server log messages


jaking92:
- [x] 4-1-stable only
---
## Feature

Due to concerns raised in #3665 and the lack of ability to reproduce the issue in most environments, it has been determined that adding an enhanced logging feature would be the best way to try to diagnose the problem. The enhanced logging should be implemented in the form of an irods_server.config setting.

jaking92:
We can reopen/adjust information included in the logs as needed.

PR #3935, opened 2018-05-15
[#3927] Add enhanced logging server feature (4-1-stable)

jaking92:
Adds a class called enhanced_logger which holds a map of information which can be accessed by logging utilities. The feature is enabled through a new server_config.json string called `enhanced_logging`. This is simply configured by configuring the specified setting to 0 for disable, and any non-zero value for enable. Defaults to 0.

Also added some additional logging in replication resource and the unixfilesystem resource.

---
CI tests still running. Commit message needs #, too. Also remember when reviewing that 4.1.x is pre-C++11. :)


jaking92:
CI tests passed.

PR #3936, opened 2018-05-15
Include irods_linked_list_iterator.hpp in dev package

theferrit32:
Building S3 plugin ran into an issue with this file missing.  It is included in dataObjOpr.hpp


theferrit32:
closing

Issue #3941, opened 2018-05-17
When delayed rules are run in parallel by at most N forked processes, at most N delayed rules are run per batch


dworkin:
- [x] master
- [x] 4-2-stable
-----

### iRODS Version, OS and Version

iRODS master / 4.2.2

### What did you try to do?

When ``maximum_number_of_concurrent_rule_engine_server_processes`` is 4, at most 4 delayed rules will be processed per batch, and ``rule_engine_server_execution_time_in_seconds`` is ignored.  This seems odd, given that when parallelisation is disabled, and delayed rules are run sequentially, more than 4 delayed rules can be processed per batch. This can work out to single-threaded processing of delayed rules being faster.

### Expected behavior

**When** ``maximum_number_of_concurrent_rule_engine_server_processes`` is 4, ``rule_engine_server_sleep_time_in_seconds`` is 30, ``rule_engine_server_execution_time_in_seconds`` is 120, there are 10 delayed rules in the queue, and each delayed rule takes about a second to process,
**then** I expect all 10 to be processed in the same batch, at most 4 concurrently.

### Observed behavior (including steps to reproduce, if applicable)

Only 4 delayed rules will be processed.

jaking92:
Hi @dworkin. Thanks for all the good work in the pull request!

I ran the following rule:
```
myTestRule {
    for(*i = 0; *i < 10; *i = *i + 1) {
        delay("<PLUSET>10s</PLUSET>") {
            writeLine("serverLog", "delay *i");
        }
    }
}
OUTPUT ruleExecOut
```
...with the following settings in `server_config.json` (same as those described above):
```
"maximum_number_of_concurrent_rule_engine_server_processes": 4, 
"rule_engine_server_execution_time_in_seconds": 120, 
"rule_engine_server_sleep_time_in_seconds": 30, 
```
...and got the following output in the log:
```
Jul 10 13:15:13 pid:37530 NOTICE: writeLine: inString = delay 0
Jul 10 13:15:13 pid:37534 NOTICE: writeLine: inString = delay 2
Jul 10 13:15:13 pid:37535 NOTICE: writeLine: inString = delay 3
Jul 10 13:15:13 pid:37532 NOTICE: writeLine: inString = delay 1
Jul 10 13:15:13 pid:37544 NOTICE: writeLine: inString = delay 5
Jul 10 13:15:13 pid:37542 NOTICE: writeLine: inString = delay 4
[...]

PR #3942, opened 2018-05-17
[#3941] Process more delayed rules per batch

dworkin:
When forking to run delayed rules, process more than
maximum_number_of_concurrent_rule_engine_server_processes per batch.

### DIFF
--- a/server/core/src/reServerLib.cpp
+++ b/server/core/src/reServerLib.cpp
@@ -409,7 +409,15 @@ runQueuedRuleExec( rcComm_t *rcComm, reExec_t *reExec,
     int thrInx;
 
     inx = -1;
-    while ( time( NULL ) <= endTime && ( thrInx = allocReThr( reExec ) ) >= 0 ) {
+    while ( time( NULL ) <= endTime ) {
+	if ( ( thrInx = allocReThr( reExec ) ) < 0 ) {
+	    if ( reExec->doFork == 0 || waitAndFreeReThr( rcComm, reExec ) < 0 ) {
+		break;
+	    }
+	    /* another job can be run */
+	    continue;
+	}
+
         myRuleExecInp = &reExec->reExecProc[thrInx].ruleExecSubmitInp;
         if ( ( inx = getNextQueuedRuleExec( genQueryOut, inx + 1,
                                             myRuleExecInp, reExec, jobType ) ) < 0 ) {


server/core/src/reServerLib.cpp
```diff
@@ -409,7 +409,15 @@ runQueuedRuleExec( rcComm_t *rcComm, reExec_t *reExec,
     int thrInx;
 
     inx = -1;
-    while ( time( NULL ) <= endTime && ( thrInx = allocReThr( reExec ) ) >= 0 ) {
+    while ( time( NULL ) <= endTime ) {
+	if ( ( thrInx = allocReThr( reExec ) ) < 0 ) {
```
dworkin: Up to here there is no real change.  Follow this up with ``break;`` to get the original behaviour.

server/core/src/reServerLib.cpp
```diff
@@ -409,7 +409,15 @@ runQueuedRuleExec( rcComm_t *rcComm, reExec_t *reExec,
...
+	    if ( reExec->doFork == 0 || waitAndFreeReThr( rcComm, reExec ) < 0 ) {
+		break;
+	    }
```
dworkin: At this point there is no reExecProc available to run a delayed rule. Break out of the loop immediately when running single-threaded.

Otherwise, wait for a child process to exit; if there is any kind of error, managing subprocesses has failed and we won't assume they can safely be restarted.

server/core/src/reServerLib.cpp
```diff
@@ -409,7 +409,15 @@ runQueuedRuleExec( rcComm_t *rcComm, reExec_t *reExec,
...
+	    /* another job can be run */
+	    continue;
```
dworkin: At this point, an idle subprocess slot is available. Go back to the beginning to see if there is still time to run another delated rule.

dworkin:
The goal is to keep running at most ``maximum_number_of_concurrent_rule_engine_server_processes`` processes at a time to handle delayed rules, until the time limit is reached or there are no more delayed rules ready to run.  When one process is finished and there are still delayed rules waiting to be processed, start another one to take its place.

The change itself is simple: when no subprocess slots are available, stay in the loop and wait until one becomes available, then try again.  I have commented the changes to further elucidate this.

jaking92:
This change passes CI tests.

trel:
We should get a test in place before pulling this in.

jaking92:
Created a separate pull request for the test and double exec issues (still linked to 3941).

trel:
will cherry pick to 4-2-stable in a separate PR

PR #3946, opened 2018-05-21
Remove double increment for loop counter in rodsMonPerfLog

xtus:
Hi,

The loop counter `index` is double incremented, leading to discard half of the metrics update.

The issue appeared in 4.1.0 with commit c4d070d .

Regards,
Marco


trel:
Excellent find.  Thanks.

trel:
awaiting CI.

PR #3954, opened 2018-05-23
python rule engine failing

jassigill2000:


trel:
passed CI - but awaiting a different fix that will skip native rule engine rules altogether.

Issue #3913, opened 2018-05-02
silent failure of imkdir -p with long path containing white space (blanks)


trel:
- [x] master
- [x] 4-2-stable
-------

### iRODS Version, OS and Version

4.2.2
### What did you try to do?
`imkdir -p` with too long path

### Expected behavior
either no problem (path cannot be too long), or an appropriate error message returned to the user

### Observed behavior (including steps to reproduce, if applicable)

no error message, no return code, and no subcollection creation
```
$ imkdir -p "research-acceptatietestj/gfasgasgfisjafipkaweopfkpwkpvfksawdogkvloafsadlkskdjfdsajfkl asjdfuihbuihuiajujgfawnfuiaheuirfhsaduifhvnasindfuihawsuihseuihruiashdfjkashdjkfhajksdfdsaf/fasdfwaguwioefhiuawshfuiwheafuihaweuifhuiawehfuihuwaeihfuihweauihfuihguiwhauighuiewhguahwseuighuiaehguheuighuawghaeguihwguhuiwehheuahguiwsehghawsghuiashuigheui/sadgasdgadwsfjkasdj fhjkasdhjfjkasdfjjesawedfjjfdijkasdfjsdafjsadjkjksdajkasdjklfsadjkjklifr weierwiorweiorweiorweiokjfdskjajlioeriorweiorwrewisdajkfdsfdasrweoiroerowakljsdlfklaskleiowiorewijasdfkljfdklsaiowropiewkliakl dfas/dafsfasdfjkfdsajikuadsfjkdsafjkdfsajkdfseajerwrrwareafldfadffdsafdrefwierioaiorioarifrfifdfakrw/dsdfsdf/afhsjdifhajksdhfjk   dsajkflksdajfkla  dalksflk jas asd lkjasdj lkjasdl kas  aslkdjfj kljalksdj salkd jkajsld a sdf asd    dsaf sdaf/gfasgasgfisjafipkaweopfkpwkpvfksawdogkvloafsadlkskdjfdsajfklasjdf"
$ echo $?
0
$ ils
/tempZone/home/rods:
$
```

in rodsLog:
```
May  2 08:20:51 pid:20338 remote addresses: 127.0.0.1, 152.54.8.147 ERROR: rstrcpy not enough space in dest, slen:796, maxLen:719, stacktrace:

Dumping stack trace
[...]

PR #3963, opened 2018-06-06
[#3913] silent failure of imkdir -p with long path containing white space

andrew-irods:
[#3913] silent failure of imkdir -p with long path containing white space


andrew-irods:
CI test for the fix (build 1036) was successful.

Issue #3853, opened 2018-03-21
clear out old resc_name and resc_hier entries since 4.2 uses resc_id


trel:
- [x] master
- [x] 4-2-stable

---
From https://groups.google.com/forum/#!topic/irod-chat/j09x4NJM5bM:

> When I run iquest "SELECT DATA_NAME, COLL_NAME WHERE DATA_RESC_NAME = '${resName}'" I get a list of collections and object (theoretically all on the resource specified via ${resName}).

> If I concatenate each data_name and coll_name, I get an absolute data object path.
If I run ils -L against these paths, I find the objects are not systematically on the resource specified in the iquest command.

> I have had a look in our database, in r_data_main. The dataobjects concerned have data in the resc_hier and resc_name fields of this table and this data is not consistent with the data held in the resc_id field of the same table!

We should clear the resc_name and resc_hier fields on upgrade to 4.2 (and/or every new write to r_data_main).  We hesitated before, as leaving the information there would 'be harmless', but we did not consider this use case of 'old queries' still using resc_name and resc_hier and matching stale information.

onmh:
Quick update on the issue: resc_hier is sometimes NULL and sometimes an empty string ''. It seems, from my tests, that it is NULL for data_repl_num = 0 and a empty string in other cases (replicas 1 through 4 confirmed). The behavior is systematic in our test and production environments.

trel:
Thanks - we'll definitely clear these out.

onmh:
Extra update on the issue: I've set the fields to 'EMPTY_RESC_NAME' for resc_name and NULL for resc_hier via psql update on our production environments and haven't noted any adverse effects.

trel:
ah, very good.

PR #3979, opened 2018-06-17
[#3853] Use known strings in obsolete columns

jaking92:
Obsolete columns in R_DATA_MAIN have inconsistent empty values. This change uses well-defined values (EMPTY_*) in a few unused columns for all new data objects registered in iRODS. Replicas made of said data objects will contain the same garbage information. Any data objects or replicas registered before this change will have the out of date information (pre-4.2) or empty values (as of 4.2).

resc_hier and resc_name do not actually store the resc_hier or resc_name anymore, so any code related to repaving these has been removed. Nothing that was removed is in use anymore.

---
[Passed CI](http://172.25.14.125:8080/view/2.%20Personal/job/irods-build-and-test-workflow/1053/) before rebase. Running again.

### DIFF
--- a/plugins/database/src/db_plugin.cpp
+++ b/plugins/database/src/db_plugin.cpp
@@ -919,74 +919,6 @@ static int _delColl( rsComm_t *rsComm, collInfo_t *collInfo ) {
 
 } // _delColl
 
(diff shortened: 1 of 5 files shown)


jaking92:
CI tests passed after rebase.

Issue #4016, opened 2018-07-12
iput displays confusing error message when it detects a file system loop


andrew-irods:
- [x] master
- [x] 4-2-stable
-----
## Bug Report

When the iput utility is used with a symbolic link on the command line, such that the symbolic link creates a file system loop by pointing to itself, it reacts correctly by displaying an error message, and aborts the operation.

However, the error message is one line out of dozens of lines displayed, since the utility puts out its usage message right after the error message.  This is what it looks like:

On stdout: 

~~~
iRODS Version 4.2.4                iput
~~~

On stderr:

~~~
boost::filesystem::status: Too many levels of symbolic links: "dir2/selfsym"
Usage: iput [-abfIkKPQrtTUvV] [-D dataType] [-N numThreads] [-n replNum]
             [-p physicalPath] [-R resource] [-X restartFile] [--link]
             [--lfrestart lfRestartFile] [--retries count] [--wlock]
             [--purgec] [--kv_pass=key-value-string] [--metadata=avu-string]
             [--acl=acl-string]  localSrcFile|localSrcDir ...  destDataObj|destColl
Usage: iput [-abfIkKPQtTUvV] [-D dataType] [-N numThreads] [-n replNum] 
             [-p physicalPath] [-R resource] [-X restartFile] [--link]
             [--lfrestart lfRestartFile] [--retries count] [--wlock]
             [--purgec] [--kv_pass=key-value-string] [--metadata=avu-string]
             [--acl=acl-string]  localSrcFile
 
Store a file into iRODS.  If the destination data-object or collection are
not provided, the current iRODS directory and the input file name are used.
 
The -X option specifies that the restart option is on and the operation
is restartable. The restartFile input specifies a local file that contains
the restart information. If the restartFile does not exist, it will be created
and used for recording subsequent restart information. If it exists and is not
empty, the restart information contained in this file will be used to restart
the operation. Note that the operation is not restarted automatically
when it failed. Another iput -X run must be made to continue from where
. . .
~~~

and this goes on to conclusion of the 116 lines of usage help.

This buries the important information included in the first line shown on stderr:

~~~
boost::filesystem::status: Too many levels of symbolic links: "dir2/selfsym"
~~~

which is all the user needs to know.  Everything else is confusing and unnecessary.

(Note: the irsync utility reacts to the problem much better with a clear error message, and no usage information.)

### iRODS Version, OS and Version
All 4.2-XXX versions, all OS versions (I use ubuntu 16.04)

[...]

PR #4026, opened 2018-07-18
[#4016] When a command line parameter causes a boost::filesystem exce…

andrew-irods:
…ption, do not display the usage message.

### This PR should be merged at the same time as https://github.com/irods/irods_client_icommands/pull/58


andrew-irods:
CI 1172 running currently.

Issue #3906, opened 2018-04-26
Delayed rule can't be scheduled when top-level call has output parameters


dworkin:
- [x] master
- [x] 4-2-stable

---
### iRODS Version, OS and Version
iRODS 4.2.2, Centos 7

### What did you try to do?
```
test() {
        delay ("<PLUSET>1s</PLUSET>") {
                writeLine("serverLog", "delayed rule executed");
        }
        *status = "OK";
}
input null
output *status
```
### Expected behavior

I expect "delayed rule executed" to appear in the log eventually.

### Observed behavior (including steps to reproduce, if applicable)

```
Apr 26 16:59:56 pid:21632 remote addresses: 127.0.0.1 ERROR: resolveStrInItem: Cannot resolve type in inOutStruct
Apr 26 16:59:56 pid:21632 remote addresses: 127.0.0.1 ERROR: packReiAndArg: packStruct error. status = -15000
Apr 26 16:59:56 pid:21632 remote addresses: 127.0.0.1 ERROR: executeRuleAction Failed for delayExec status = -15000 SYS_PACK_INSTRUCT_FORMAT_ERR
Apr 26 16:59:56 pid:21632 NOTICE: executeRuleBody: Microservice or Action delayExec Failed with status -15000
Apr 26 16:59:56 pid:21632 NOTICE: execRuleNodeRes: applyRule Failed: test with status -15000
Apr 26 16:59:56 pid:21632 DEBUG:
Apr 26 16:59:56 pid:21632 NOTICE: execMyRule @external
test() {
        delay ("<PLUSET>1s</PLUSET>") {
                writeLine("serverLog", "delayed rule executed");
        }
}
 Failed with status -15000
```

This will happen even when the usage of delay is several calls deep, as long as the top level call has an output parameter.

alanking:
Need to add python version of the test for PREP.

PR #4031, opened 2018-07-26
[#3906] Pack and unpack output parameters 

dworkin:
When creating a delayed rule, a copy of the REI and top-level arguments is preserved.  Unfortunately, when one of the top level arguments is an output parameter that has not been assigned a value yet, packing would fail with the error SYS_PACK_INSTRUCT_FORMAT_ERR, preventing the delayed rule from being started.

Handle the particular case that prevented successful packing: when packing an uninitialised parameter, the `type` is a NULL pointer, and thus the type of `inOutStruct` cannot be determined. However, `inOutStruct` is also a NULL pointer, so the type is irrelevant for the purpose of packing.  Force the type to `STR_PTR_PI`.

It would be cleaner to introduce a new type for this particular purpose, something like `VOID_PTR_PI`, but such a change would affect much more code.

### DIFF
--- a/lib/core/src/packStruct.cpp
+++ b/lib/core/src/packStruct.cpp
@@ -384,20 +384,36 @@ resolvePackedItem( packItem_t &myPackedItem, const void *&inPtr, packOpr_t packO
 
     /* set up the pointer */
 
+    const void *ptr = inPtr;
     if ( myPackedItem.pointerType > 0 ) {
         if ( packOpr == PACK_OPR ) {
             /* align the address */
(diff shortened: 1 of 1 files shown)


jaking92:
Test written based on @dworkin's example scenario in #3906. [CI tests passed.](http://172.25.14.125:8080/view/2.%20Personal/job/irods-build-and-test-workflow/1224/)

trel:
thanks!  test is written and being merged soon.   then we'll cherry-pick both to master.

PR #4039, opened 2018-08-01
[#3906] Add test for output with delay block (4-2-stable)

jaking92:
Output parameter causing writeLine output in delay block to be short-circuited. Need a test to ensure messages from a delay block are written when an output variable is being used.

---
[CI tests running.](http://172.25.14.125:8080/view/2.%20Personal/job/irods-build-and-test-workflow/1236/)


jaking92:
Passed CI tests (had to re-run test-ub14-psql).

Issue #3981, opened 2018-06-18
Create 4.2.4 upgrade script to update unused/obsolete columns


alanking:
- [x] master
- [x] 4-2-stable

---
## Feature

It was discovered that some columns which are no longer used as of 4.2 had inconsistent empty or NULL values. Any rows created before 4.2 populated these columns with actual data, but it is no longer updated (meaning, it could be outdated). As a result, any specific-queries which select these columns could return out of date or "inconsistently empty" results. As of the resolution of #3853 any new entries will have known strings to represent the empty values.

- Should update all entries of R_DATA_MAIN with the following values:
-- `resc_name`: 'EMPTY_RESC_NAME'
-- `resc_hier`: 'EMPTY_RESC_HIER'
-- `resc_group_name`: 'EMPTY_RESC_GROUP_NAME'
- Should validate the following items before making changes for each row to prevent loss of critical catalog data:
-- `resc_id` column is populated and valid for this row
-- `resc_id` found contains the replica this row represents
- Should be able to interrupt and restart the column update cleanly.
- Should allow user to configure the "batch size" to control how many rows are updated before committing (default to 500)
- Should have the ability to "dry run" the column updates to see how many rows will be affected (anything made before 4.2.4 will be affected).

Running the script will be manual, at the discretion of the iRODS admin in charge of the zone.

PR #4058, opened 2018-08-13
[#3981] Add script for updating obsolete columns in R_DATA_MAIN (master)

alanking:
Adds a script to be run when updating past 4.2.4 which updates the following columns in `R_DATA_MAIN` to contain "known values":
 - resc_name: EMPTY_RESC_NAME
 - resc_hier: EMPTY_RESC_HIER
 - resc_group_name: EMPTY_RESC_GROUP_NAME

Skips any rows with "invalid" resc_id column. A resc_id is considered invalid if it does not exist in R_RESC_MAIN or it maps to a known non-storage resource (i.e. coordinating resource).

Added get_database_type to database_interface for reusability.

---
This script must be run manually.

### DIFF
--- a/scripts/irods/database_interface.py
+++ b/scripts/irods/database_interface.py
@@ -77,9 +77,7 @@ def database_already_in_use_by_irods(irods_config):
             else:
                 return False
 
-def setup_database_config(irods_config):
-    l = logging.getLogger(__name__)
-
+def get_database_type():
     if os.path.exists(os.path.join(paths.plugins_directory(), 'database', 'libpostgres.so')):
         db_type = 'postgres'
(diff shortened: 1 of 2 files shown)


alanking:
Moved connection acquisition block out to "main" function to handle failure to obtain database connection from one place. An exception is raised in `database_connection` if run from the catalog consumer and this is captured and reported. Additional comments added to help messages as well.

trel:
I think the catching of the exception is fine - this is a helper/scrubber script.  If it can self-diagnose *any* problems, that's great.  It should be run by someone who knows what they're doing.  The header text should be explicit about that (and I think it is).

Looks good to me...

Issue #4074, opened 2018-08-23
Enable C++17 Support


korydraughn:
- [x] master
- [ ] 4-2-stable **(not supported in this branch)**

Issue #3132, opened 2016-04-06
syslog export of rodsLog


kript:
Related to #3131 and [this thread](https://groups.google.com/forum/#!topic/irod-chat/xMltK0UDuNo) in irods-chat, its not clear how one redirects the rodsLog to go to syslog instead. Under V3 this was a compile option in the config/config.mk, but there is nothing in the documentation about how to do this in 4.X, and compilation is no longer the recommended way to deploy a production instance.

If we could redirect to syslog, we could also allow another group access to it, control rotation etc, so that would be helpful.

trel:
related to https://github.com/irods/irods/issues/2730 and https://github.com/irods/irods/issues/2112

korydraughn:
I think this can be closed.

Issue #3131, opened 2016-04-06
Group access to rodsLog


kript:
(as per irods-chat post, creating an issue/feature request  so it doesn't get lost)
Hi Folks,

I have a development server that the developers want to see the rodslog for as they write their code. Now that its possible to set the group for the irods processes in `/etc/irods/service_account.config`, I thought I'd change the group to one that the developers are members of, chown the `/var/lib/irods` and `/etc/irods` and off they go.

Unfortunately, the rodslog files are set to be read by the irods user only;

```
me@irods-dev:~$ ls -lah /var/lib/irods/iRODS/server/log/
total 1.9M
drwxr-xr-x 3 irods-user-dev irods 4.0K Mar 30 12:04 .
drwxr-xr-x 8 irods-user-dev irods 4.0K Mar 22 15:52 ..
drwx------ 2 irods-user-dev irods 4.0K Mar 30 12:09 proc
-rw------- 1 irods-user-dev irods 722K Mar 26 00:18 reLog.2016.03.21
-rw------- 1 irods-user-dev irods 993K Mar 30 14:53 reLog.2016.03.26
-rw------- 1 irods-user-dev irods 132K Mar 30 12:04 rodsLog.2016.03.21
-rw------- 1 irods-user-dev irods 5.6K Mar 30 12:09 rodsLog.2016.03.26
```

Any ideas how I change the umask for rodslog files, or a similar solution? Any better ideas for providing access to the logs for the developers?

kript:
Any chance of getting this into 4.1.9 if its not a big change?

trel:
Yes, we'll get it included.

A change to the permissions on creation of the log file should be a minimal tweak.

trel:
Something like... `server/core/src/rodsServer.cpp:240`:

```diff
#else
     LogFd = open( logFile, O_CREAT | O_WRONLY | O_APPEND, 0644 );
+    close(LogFd);
+    // get username and groupname from service_account.config file?
+    chown( logFile, getpwnam(username)->pw_uid, getgrnam(groupname)->gr_gid );
+    LogFd = open( logFile, O_CREAT | O_WRONLY | O_APPEND, 0644 );
#endif
```

trel:
Alternately...

@kript If it suits you, I think you can change the `umask( 077 )` of the server process in the `iRODS/scripts/perl/irodsctl.pl` script.  In 4-1-stable, it can be changed to `umask( 037 )` and the `rodsLog*` should then be `640`.

[...]

Issue #2730, opened 2015-06-03
Move handling of log renaming from rodsLog to external service


kellerb:
Have rodsLog write to a constant filename.

Rotation, etc. can be provided by external services such as logrotate.

This has the side-effect of making it easier to find the active logs from other programs, e.g. when scanning the log from tests.

trel:
related to  #2112

bzizou:
+1
The filename should be constant, but the location of the logfiles should also be customizable.
It would be also great to provide a default logrotate config file.

iychoi:
I felt the same as @bzizou . How about a existing log libraries such as log4c?

bzizou:
I also noticed that I have a lot of log files into /var/lib/irods/iRODS/server/bin ! 32000 files, 500MB  in a week. Anyway to disable this?

jasoncoposky:
can you provide the list of files in server/bin?

bzizou:
[...]
-rw------- 1 irods irods     273 Sep 18 11:08 9931.log
-rw------- 1 irods irods     273 Sep 18 11:08 9933.log
-rw------- 1 irods irods     273 Sep 18 11:08 9935.log
-rw------- 1 irods irods     273 Sep 18 11:08 9937.log
-rw------- 1 irods irods     273 Sep 18 11:08 9939.log
-rw------- 1 irods irods     273 Sep 18 11:08 9941.log
-rw------- 1 irods irods     273 Sep 18 11:08 9943.log
-rw------- 1 irods irods     273 Sep 18 11:08 9945.log
-rw------- 1 irods irods     273 Sep 18 11:08 9947.log
-rw------- 1 irods irods     273 Sep 18 11:08 9949.log
-rw------- 1 irods irods     273 Sep 18 11:08 9951.log
-rw------- 1 irods irods     273 Sep 18 11:08 9953.log
-rw------- 1 irods irods     273 Sep 18 11:08 9956.log
-rw------- 1 irods irods     273 Sep 18 11:08 9958.log
-rw------- 1 irods irods     273 Sep 18 11:08 9960.log
-rw------- 1 irods irods     273 Sep 18 11:08 9962.log
-rw------- 1 irods irods     273 Sep 18 11:08 9964.log
-rw------- 1 irods irods     273 Sep 18 11:09 9966.log
[...]
And each file contains:
DSN info: DSN='postgres',server='127.0.0.1',port='5433',dbase='icat',user='',passwd='xxxxx'
[...]

PR #4075, opened 2018-08-23
[#4074] Introduces new logging API and enables C++17 support.

korydraughn:
Depends on: [irods/externals#15]

Fixes Issues: [#2112][#2730][#3131][#3132]

New logging API built on top of the spdlog logging library.  The new logging API
writes to the syslog only.  Messages are formatted as JSON to allow easy
integration with other tools.  All rodsLog style functions also forward the
message to the new logging API.

The server now returns a trace of what happened while processing a
request whenever the verbose flag is detected.  This allows the client to
diagnose what the server was doing without needing to view any log
files.

Replaces std::random_shuffle with std::shuffle.

Requires custom Avro build that replaces deprecated code (std::auto_ptr)
with supported code (std::unique_ptr).
  - Dependency satisfied (built using irods-externals-avropre190cpp17-0 package).

Removed XMSG python test (fails on this for unknown reason).
  - XMSG is deprecated and will be removed.

Passed CI.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,11 +30,13 @@ endmacro()
 IRODS_MACRO_CHECK_DEPENDENCY_SET_FULLPATH(CLANG clang6.0-0)
 IRODS_MACRO_CHECK_DEPENDENCY_SET_FULLPATH(CPPZMQ cppzmq4.2.3-0)
 IRODS_MACRO_CHECK_DEPENDENCY_SET_FULLPATH_ADD_TO_IRODS_PACKAGE_DEPENDENCIES_LIST(ARCHIVE libarchive3.3.2-1)
-IRODS_MACRO_CHECK_DEPENDENCY_SET_FULLPATH_ADD_TO_IRODS_PACKAGE_DEPENDENCIES_LIST(AVRO avro1.8.2-0)
+IRODS_MACRO_CHECK_DEPENDENCY_SET_FULLPATH_ADD_TO_IRODS_PACKAGE_DEPENDENCIES_LIST(AVRO avropre190cpp17-0)
 IRODS_MACRO_CHECK_DEPENDENCY_SET_FULLPATH_ADD_TO_IRODS_PACKAGE_DEPENDENCIES_LIST(BOOST boost1.67.0-0)
 IRODS_MACRO_CHECK_DEPENDENCY_SET_FULLPATH_ADD_TO_IRODS_PACKAGE_DEPENDENCIES_LIST(CLANG_RUNTIME clang-runtime6.0-0)
 IRODS_MACRO_CHECK_DEPENDENCY_SET_FULLPATH_ADD_TO_IRODS_PACKAGE_DEPENDENCIES_LIST(JANSSON jansson2.7-0)
 IRODS_MACRO_CHECK_DEPENDENCY_SET_FULLPATH_ADD_TO_IRODS_PACKAGE_DEPENDENCIES_LIST(ZMQ zeromq4-14.1.6-0)
+IRODS_MACRO_CHECK_DEPENDENCY_SET_FULLPATH_ADD_TO_IRODS_PACKAGE_DEPENDENCIES_LIST(JSON json3.1.2-0)
+IRODS_MACRO_CHECK_DEPENDENCY_SET_FULLPATH_ADD_TO_IRODS_PACKAGE_DEPENDENCIES_LIST(SPDLOG spdlog0.17.0-0)
 
 string(REPLACE ";" ", " IRODS_PACKAGE_DEPENDENCIES_STRING "${IRODS_PACKAGE_DEPENDENCIES_LIST}")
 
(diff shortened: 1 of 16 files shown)


CMakeLists.txt
```diff
@@ -782,7 +787,7 @@ target_include_directories(
 set_property(TARGET irods_common PROPERTY CXX_STANDARD ${IRODS_CXX_STANDARD})
 set_property(TARGET irods_common PROPERTY VERSION ${IRODS_VERSION})
 set_property(TARGET irods_common PROPERTY SOVERSION ${IRODS_VERSION})
-target_compile_definitions(irods_common PRIVATE ${IRODS_COMPILE_DEFINITIONS})
+target_compile_definitions(irods_common PRIVATE ${IRODS_COMPILE_DEFINITIONS} RODS_SERVER SPDLOG_ENABLE_SYSLOG)
```
trel: do we want RODS_SERVER here?

korydraughn: iRODS cannot be compiled without it mainly because of libRodsAPIs.  The RODS_SERVER macro is used to enable code that forwards messages from rodsLog to the new logging API.  It can be handled in other ways, it's just a matter of which way we prefer.

Is libRodsAPIs still needed?

If libRodsAPIs is no longer supported and we remove it, then I think the RODS_SERVER macro can be removed.

korydraughn: Code Update:
- Resolved compilation issues with libRodsAPIs.
- Removed dependency on RODS_SERVER macro.
- Moved SPDLOG_ENABLE_SYSLOG macro into header file and introduced  IRODS_ENABLE_SYSLOG macro for control.

server/core/include/irods_logger.tpp
```diff
@@ -0,0 +1,217 @@
...
+
+        return json(data).dump();
+    }
+
+    template <typename ForwardIt>
+    void append_to_r_error_stack(ForwardIt _first, ForwardIt _last) const
+    {
+        if (error_ && write_to_error_object_)
+        {
+            std::for_each(_first, _last, [](const auto& _key_value) {
```
tempoz: If we use a [boost::iterator_range](https://www.boost.org/doc/libs/1_36_0/libs/range/doc/utility_class.html#iter_range) here, we can use a [range-based for-loop and use structured binding declarations](https://en.cppreference.com/w/cpp/language/range-for) instead of needing to use 'first' and 'second'. Also classic for loops have preferable syntax to std::for_each.

server/core/include/irods_logger.tpp
```diff
@@ -0,0 +1,217 @@
...
+                std::string msg = _key_value.first;
+                msg += ": ";
+                msg += _key_value.second;
+
+                // Force addRErrorMsg() to skip prefixing the message with 'Level N:'.
+                // When the output is printed on the client-side, it will very close to
+                // the log file.
+                constexpr int disable_level_prefix = STDOUT_STATUS;
+
+                if (const auto ec = addRErrorMsg(error_, disable_level_prefix, msg.c_str()); ec)
```
tempoz: I like this style, we should make sure to do it. Even when we are evaluating the we initialize, explicitly evaluating it afterwards makes it more consistent and clearer to read.

lib/core/include/apiHandler.hpp
```diff
@@ -67,7 +72,46 @@ namespace irods {
...
+                else if constexpr (std::is_same_v<dataObjCopyInp_t*, T>)
+                {
+                    logger_updated = true;
+                    const auto* flag_0 = getValByKey(&_arg->srcDataObjInp.condInput, VERY_VERBOSE_KW);
+                    const auto* flag_1 = getValByKey(&_arg->destDataObjInp.condInput, VERY_VERBOSE_KW);
+                    irods::experimental::log::write_to_error_object(flag_0 && flag_1); 
+                }
+            };
+
+            (update_logger(std::forward<types_t>(_t)), ...);
```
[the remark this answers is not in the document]
korydraughn: The comma is actually required.  Without, it won't compile.

server/core/include/irods_logger.tpp
korydraughn: I totally agree.

lib/core/include/apiHandler.hpp
```diff
@@ -67,7 +72,46 @@ namespace irods {
         api_call_adaptor( std::function<int(rsComm_t*, types_t...)> _fcn ): fcn_(_fcn) {
         }
 
-        irods::error operator()( irods::plugin_context&, rsComm_t* _comm, types_t... _t ) {
+        irods::error operator()( irods::plugin_context&, rsComm_t* _comm, types_t... _t )
+        {
+#ifdef IRODS_ENABLE_SYSLOG
+            bool logger_updated = false;
+
+            auto update_logger = [&logger_updated](auto&& _arg)
```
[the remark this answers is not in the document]
korydraughn: Are you referring to the **update_logger** variable or the **_arg** parameter?

lib/core/src/rodsLog.cpp
```diff
@@ -77,6 +79,52 @@ std::string create_log_error_prefix() {
...
+        {LOG_DEBUG6,      debug},
+        {LOG_NOTICE,      info},
+        {LOG_WARNING,     warn},
+        {LOG_ERROR,       error},
+        {LOG_SYS_WARNING, critical},
+        {LOG_SYS_FATAL,   critical}
+    };
+    // clang-format on
+
+    if (auto iter = msg_handlers.find(_log_level); std::end(msg_handlers) != iter)
```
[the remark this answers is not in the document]
korydraughn: Done.

lib/core/src/rcMisc.cpp
```diff
@@ -4065,7 +4066,10 @@ getRandomArray( int **randomArray, int size ) {
     for ( int i = 0; i < size; i++ ) {
         ( *randomArray )[i] = i + 1;
     }
-    std::random_shuffle( *randomArray, *randomArray + size );
+
+    static std::random_device rng;
+    static std::mt19937 urng{rng()};
```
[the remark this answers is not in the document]
korydraughn: Done.

server/core/include/irods_logger.tpp
korydraughn: Done.

Issue #4007, opened 2018-06-29
Update rule engine plugin rule strings to support the new PREP model for irule and remote commands


jasoncoposky:
reference irods/irods_rule_engine_plugin_python#9

PR #4079, opened 2018-08-24
[#4007] fixed regression in PREP test (for 4-2-stable)

jassigill2000:
fixed some typos that caused regression in PREP testing


jassigill2000:
fixed some typos that caused regression in PREP testing

PR #4081, opened 2018-08-27
[#4007] fixed regression in PREP test (for master)

jassigill2000:
Fixed typos in previous commits that caused regression.


jassigill2000:
Fixed typos in previous commits that caused regression.

Issue #4066, opened 2018-08-20
Add support for data size and modify time keywords in register API


alanking:
- [x] master
- [x] 4-2-stable

---
## Feature

Add support for clients that want to provide data size and modify time directly to the dataObjInfo instead of relying on the utilities in use by the registration API. This is needed to register files quickly and with the correct metadata attributes in the [ingest tool](https://github.com/irods/irods_capability_automated_ingest). Clients which do not use the keywords will have the information provided by default, as shown below...

Data size is retrieved via file stat on the destination resource:
https://github.com/irods/irods/blob/a04846b338e76822ae2c4c28aac73484f60ef33e/server/api/src/rsPhyPathReg.cpp#L524

Modify time is the current time:
https://github.com/irods/irods/blob/a04846b338e76822ae2c4c28aac73484f60ef33e/plugins/database/src/db_plugin.cpp#L2704

alanking:
We need to bypass stat'ing altogether in the presence of the size KW. Re-opening to resolve this issue.

PR #4093, opened 2018-08-31
[#4066] Skip redirect in reg with size KW (master)

alanking:
When the size KW is provided in the call to rsPhyPathReg, there is no need to stat the file. As such, there is no need to redirect to the remote server at all. This also resolves an issue with stat'ing 0-length files even in the presence of the size KW.

(cherry-picked from SHA: 14a868df0fbaf9ca09b0ab13adb41da5d6a237df)

---
[CI tests running.](http://172.25.14.63:8080/view/Personal/job/irods-build-and-test-workflow/1337/) Will update commit message with new SHA when 4-2-stable version is merged.


trel:
4-2-stable merged, please to SHA and #

Issue #4100, opened 2018-09-05
rcGetHostForGet assumes data object resides on "home zone"


alanking:
- [x] master
- [x] 4-2-stable
- [x] 4-1-stable

---
### iRODS Version
4.1.12

### Observed behavior (including steps to reproduce, if applicable)
Running [tears](https://github.com/whitwham/tears) in a federated environment without the `-d` flag will result in failure because of a bug in rcGetHostForGet.

My work machine is the client machine connected to zoneA. I have a remote machine running an icat server which serves as zoneA. zoneA is federated with zoneB. zoneB has an icat server and 2 resource servers. zoneB has the following resource tree:
```
$ ilsresc
demoResc:passthru
└── replicate:replication
    ├── rand1:random
    │   ├── resc1
    │   └── resc1alt
    └── rand2:random
        ├── resc2
        └── resc2alt
```
I built tears using the following steps:
```
$ autoreconf configure.ac
$ automake --add-missing
$ ./configure --with-irods=/path/to/irods/installation/iRODS
$ make
```
Running tears on a file in the remote zone results in an error:
```
$ ./tears -f -r /tempZone/home/rods/secretfile > secretfile
Error: rcGetHostForGet failed with status -1803000:HIERARCHY_ERROR
```
On server log:
```
Jul 31 16:59:32 pid:6200 NOTICE: Agent process 18687 started for puser=rods and cuser=rods from xxx.xxx.xxx.xxx
    Jul 31 16:59:32 pid:18687 ["agent_host":host,"agent_pid":18687,"api_inx":694,"client_ip":xxx.xxx.xxx.xxx,"client_user":rods,"proxy_user":rods] ERROR: [-]    iRODS/server/api/src/rsGetHostForGet.cpp:49:rsGetHostForGet :  status [HIERARCHY_ERROR]  errno [] -- message [rsGetHostForGet :: failed in irods::resolve_resource_hierarchy for [/tempZone/home/rods/secretfile]]
            [-]     iRODS/server/core/src/irods_resource_redirect.cpp:156:resolve_hier_for_open_or_write_without_keyword :  status [HIERARCHY_ERROR]  errno [] -- message [no valid resource found for data object]

    Jul 31 16:59:32 pid:18687 ["agent_host":host,"agent_pid":18687,"api_inx":694,"client_ip":xxx.xxx.xxx.xxx,"client_user":rods,"proxy_user":rods] NOTICE: readAndProcClientMsg: received disconnect msg from client
```
[...]

PR #4111, opened 2018-09-19
[#4100] Add zone checking to rsGetHostForGet/Put (master)

alanking:
rsGetHostForGet and rsGetHostForPut need to make sure they are connected to the correct zone before attempting any operation. If a resource hierarchy is involved, the operation will fail because
the target could be on a remote zone with a resource heirarchy which differs from the connected zone (or worse - one which doesn't differ and data is pulled/put from/to the wrong zone).

(cherry-picked from SHA: eec41ba84bde5687f8fa83acbd377c41fcdfa821)

---
[CI tests running](http://172.25.14.63:8080/view/Personal/job/irods-build-and-test-workflow/1404/)


alanking:
CI tests passed

Issue #4119, opened 2018-09-21
Handle DATA_SIZE_KW in register replica API


alanking:
- [x] master
- [x] 4-2-stable

---
### iRODS Version
4.2.4

### What did you try to do?
Use the [iRODS automated ingest tool](https://github.com/irods/irods_capability_automated_ingest) to register data objects as replicas on a resource server which has no access to the physical files underlying the data objects (no mountpoint).

### Expected behavior
Files are registered using the `DATA_SIZE_KW` just like a regular registration (see #4066).

### Observed behavior (including steps to reproduce, if applicable)
Unhandled exception occurs in `get_file_size_from_filesystem` because the files cannot be accessed.

This stat should not be necessary with the presence of `DATA_SIZE_KW` in the `regReplica_t`. Add handling of this keyword to the `rsRegReplica` API call.

PR #4126, opened 2018-09-24
[#4119] Honor DATA_SIZE_KW in rsRegReplica

alanking:
rsRegReplica was not honoring DATA_SIZE_KW which means it had to do a stat for every registration. This change adds handling for the DATA_SIZE_KW when registering a data object as a replica.

---
[CI tests passed](http://172.25.14.63:8080/view/Personal/job/irods-build-and-test-workflow/1429/)

No tests added because `DATA_SIZE_KW` is not used with `ireg`


trel:
add # to commit message and we're good.

Issue #4089, opened 2018-08-30
Quotas are not being calculated and checked properly


JustinKyleJames:
### iRODS Version, OS and Version

iRODS 4.2.4
Tested on Ubuntu 14.04

### Observed behavior (including steps to reproduce, if applicable)

I see two separate issues:

1. When a quota is set at multiple levels (resource and total) the quota check passes if any quota is not exceeded.  It should fail if any quota is exceeded.

2.  If you set a quota on a coordinating resource, the "iadmin cu" does not update the usage on that coordinating resource.

JustinKyleJames:
There are a couple of issues I noticed in the code.  

1.  In plugins/database/src/general_query.cpp, in generateSpecialQuery() the queries return a list ordered by quota_over descending.  This ordering is correct.  However, in queRescQuota() in server/api/src/rsGetRescQuota.cpp, a link list is being created from bottom to top so the first one on the list is the one with the lowest quota_over.  Then in setRescQuota() it just takes the first one from the list.

2.  The actual queries return information for all resources.  That makes me believe even if the issue above was fixed, it would still not work because we shouldn't care about the quota on a resource that is not being written to.  These queries need to be rewritten.

trel:
Is this still true?   Do we not have tests around this behavior?

korydraughn:
We have https://github.com/irods/irods/blob/ad59ae9c52331ddbc19dac71bc05461b00f46885/scripts/irods/test/test_quotas.py.

korydraughn:
I don't think there are any other tests around quotas.

trel:
Okay, so confirmed, there is nothing in that file that tests when both types of quotas are in play at the same time.

korydraughn:
> 2. If you set a quota on a coordinating resource, the "iadmin cu" does not update the usage on that coordinating resource.

Just one comment to add following discussion about this.

Q. Should physical quotas be allowed on coordinating resources?
A. We feel it should. Whether that's a good use of the feature is up to the admin.

---

The question is motivated by a resource hierarchy containing a replication resource and three unixfilesystem resources. If a quota is attached to the replication resource and a user stores a file there, the replication resource will replicate it to its children.

What should the quota report for the replication resource?
Should the user's file cost 3x its size due to there being three physical replicas?

This situation led us to the answer of ... the quota system is about physical storage and should count all replicas. The quota system is a tool and admins are given freedom over where it is applied.

trel:
[...]

PR #4151, opened 2018-10-10
[4089] Updated code to handle quotas

JustinKyleJames:
[#4089] Update code to handle quotas (passed test)

1.  Moved the quota check to resolve_resource_hierarchy as the full
hierarch needs to be known to properly check for quota issues.

2.  Send the resource to the reoutine that does the quota check.

3.  When checking quotas on resource, filter out all quotas on other
resources.

4.  Create new error code when a put succeeds but replication fails due
to quota exceeded.

5.  Do not perform retry logic when replication fails due to quota
issue.

6.  Provide an error to the user if she attempts to set a quota on a
coordinating resource.

Note:  Quotas on coordinating resources have no effect.  Updated output on iadmin suq indicating this.


JustinKyleJames:
This is an old pull request that has been superseded by other work.  Closing it.

Issue #4099, opened 2018-09-04
file_modified should be called only when data has changed


bh9:
- [x] master
- [x] 4-2-stable
- [x] 4-1-stable
-----

4.1.10
```
bh9@irods-sanger1:~$ iput -K thisiswrong
bh9@irods-sanger1:~$ ils -L thisiswrong
  bh9               0 root;replicate;seq-red;red4;irods-seq-i23-fg          373 2018-09-04.15:30 & thisiswrong
    fdc45f5eb8963c5a53b71a00a998839a    generic    /irods-seq-i23-fg/home/bh9#Sanger1/thisiswrong
  bh9               1 root;replicate;seq-green;green5;irods-seq-sr08-fg          373 2018-09-04.15:30 & thisiswrong
    fdc45f5eb8963c5a53b71a00a998839a    generic    /irods-seq-sr08-fg/home/bh9#Sanger1/thisiswrong
#Edit replica 0 on disk (e.g. to emulate corruption)
bh9@irods-sanger1:~$ ils -L thisiswrong
  bh9               0 root;replicate;seq-red;red4;irods-seq-i23-fg          373 2018-09-04.15:30 & thisiswrong
    fdc45f5eb8963c5a53b71a00a998839a    generic    /irods-seq-i23-fg/home/bh9#Sanger1/thisiswrong
  bh9               1 root;replicate;seq-green;green5;irods-seq-sr08-fg          373 2018-09-04.15:30 & thisiswrong
    fdc45f5eb8963c5a53b71a00a998839a    generic    /irods-seq-sr08-fg/home/bh9#Sanger1/thisiswrong
bh9@irods-sanger1:~$ ichksum -Kn 0 thisiswrong
ERROR: chksumDataObjUtil: rcDataObjChksum error for /seq/home/bh9#Sanger1/thisiswrong status = -314000 USER_CHKSUM_MISMATCH
ERROR: chksumUtil: chksum error for /seq/home/bh9#Sanger1/thisiswrong, status = -314000 status = -314000 USER_CHKSUM_MISMATCH
Total checksum performed = 1, Failed checksum = 1
bh9@irods-sanger1:~$ ichksum -Kn 1 thisiswrong
    thisiswrong                       fdc45f5eb8963c5a53b71a00a998839a
Total checksum performed = 1, Failed checksum = 0
bh9@irods-sanger1:~$ ils -L thisiswrong
[...]

PR #4156, opened 2018-10-21
[#4099] Prevent file_modified on non-write ops (4-1-stable)

alanking:
This change restricts when fileModified is called in rsModDataObjMeta. This should prevent replication (in the case of replication or compound hierarchies) on any non-write/create operation.

The replication resource has been modified such that the child replication list is generated at the time of fileModified instead of in resolve resource hierarchy. This ensures that the replication list is made available to the machine doing the replication as the property map may not be populated from machine to machine.

---
[CI tests passed](http://172.25.14.63:8080/job/test-ub14-oracle/1651/testReport/). Topo tests pending.

### DIFF
--- a/plugins/resources/replication/librepl.cpp
+++ b/plugins/resources/replication/librepl.cpp
@@ -215,6 +215,17 @@ extern "C" {
         return result;
     }
 
+    bool is_self_in_hier(
+        irods::resource_plugin_context& _ctx,
+        const irods::hierarchy_parser &_parser) {
+        std::string name;
+        irods::error ret = _ctx.prop_map().get<std::string>(irods::RESOURCE_NAME, name);
+        if (!ret.ok()) {
+            THROW(ret.code(), ret.result());
+        }
+        return _parser.resc_in_hier(name);
+    }
+
     irods::error get_selected_hierarchy(
         irods::resource_plugin_context& _ctx,
         std::string& _hier_string,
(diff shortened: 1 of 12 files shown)


iRODS/server/config/lockFileDir/.gitignore
```diff
@@ -1,2 +0,0 @@
-*
```
trel: was this removal intended?

plugins/resources/replication/librepl.cpp
```diff
@@ -378,50 +416,177 @@ extern "C" {
...
+        ret = selected_hier_parser.str(current_resc_hier, name);
+        if (!ret.ok()) {
+            return PASSMSG(
+                       (boost::format(
+                        "[%s] - Failed to get hierarchy string for resc name [%s].") %
+                        __FUNCTION__ % name).str(), ret);
+        }
+        irods::hierarchy_parser current_hier_parser;
+        current_hier_parser.set_string(current_resc_hier);
+        //rodsLog(LOG_NOTICE, "[%s] - pared down hier to [%s]", __FUNCTION__, current_resc_hier.c_str());
```
trel: LOG_DEBUG?

iRODS/server/config/lockFileDir/.gitignore
alanking: Nope, must have added it by accident. 4.1's build process (or at least run-in-place) seems to touch this file every time.

plugins/resources/replication/librepl.cpp
```diff
@@ -378,50 +416,177 @@ extern "C" {
...
+        }
+        const std::string host_name(host_name_str);
+
+        // Loop over children and create a list of hierarchies to which objects will be replicated
+        float out_vote;
+        child_list_t repl_vector;
+        irods::error result = SUCCESS();
+
+        irods::resource_child_map::iterator it;
+        for ( it = _ctx.child_map().begin(); result.ok() && it != _ctx.child_map().end(); ++it ) {
```
korydraughn: Is `result.ok()` necessary in the condition?  It doesn't appear to change in the loop at all.

plugins/resources/replication/librepl.cpp
```diff
@@ -215,22 +215,60 @@ extern "C" {
...
+            irods::file_object_ptr file_obj = boost::dynamic_pointer_cast<irods::file_object >( _ctx.fco() );
+            if (file_obj->l1_desc_idx() > 0) {
+                char* hier_str = getValByKey(&L1desc[file_obj->l1_desc_idx()].dataObjInp->condInput, RESC_HIER_STR_KW);
+                if (!hier_str) {
+                    return ERROR(SYS_INTERNAL_NULL_INPUT_ERR,
+                                 (boost::format(
+                                  "[%s] - No hierarchy string found in keywords for file object.") %
+                                  __FUNCTION__).str().c_str());
+                }
+                selected_parser.set_string(std::string(hier_str));
```
korydraughn: Is wrapping `hier_str` in `std::string` required?

plugins/resources/replication/librepl.cpp
```diff
@@ -1369,47 +1446,36 @@ extern "C" {
...
-        // Make sure this is a valid repl operation.
-        if ( !( ret = replValidOperation( _ctx ) ).ok() ) {
-            return PASSMSG(
-                       (boost::format(
-                        "[%s] - Invalid operation on replicating resource.") %
-                        __FUNCTION__).str(), ret );
-        }
         // add ourselves to the hierarchy parser
-        else if ( !( ret = replAddSelfToHierarchy( _ctx, parser ) ).ok() ) {
+        irods::hierarchy_parser parser = *_inout_parser;
```
korydraughn: Would this be better if it were a reference or is it cheap to copy?

plugins/resources/replication/librepl.cpp
alanking: Doesn't look like it. Removing!

plugins/resources/replication/librepl.cpp
alanking: Good point. I believe I based this on the logic found in `replRedirectToChildren` wherein result is updated in the conditional block on 486. result is returned immediately after the loop. We should probably return the `PASSMSG` created there instead of just logging it.

plugins/resources/replication/librepl.cpp
alanking: We can investigate that possibility for the 4.2+ implementation. Here, we are maintaining the existing behavior, and `resolve_resource_hierarchy` is not called frequently anyway. I am unsure of the ramifications of overwriting the value of the passed-in parser before calling `replSelectChild`.

Issue #3883, opened 2018-04-17
Using iput on a compound resource with --purgec and -k option with an empty file generates an error.


JustinKyleJames:
- [x] master
- [x] 4-2-stable
------

## Bug Report

Using iput on a compound resource with --purgec and -k option with an empty file generates an error.

### iRODS Version, OS and Version

This was tested on iRODS 4.2.2 on both Ubuntu 14.04 and CentOS 7.3.

### Test setup

```
iadmin mkresc mycompoundresc compound
iadmin mkresc mycacheresc unixfilesystem `hostname`:/var/lib/irods/mycacheresc
iadmin mkresc myarchiveresc unixfilesystem `hostname`:/var/lib/irods/myarchiveresc
iadmin addchildtoresc mycompoundresc mycacheresc cache
iadmin addchildtoresc mycompoundresc myarchiveresc archive
```

### Test Steps
```
touch emptyfile
iput -R mycompoundresc --purgec -k emptyfile
```

### Expected behavior

No error would be generated.  The behavior would be the same as if this was performed on a non-empty file.

### Observed behavior (including steps to reproduce, if applicable)

After the put the empty file is on the archive resource and is missing from the cache.  This is good.  However, the iput generates the following error:

remote addresses: 127.0.0.1 ERROR: putUtil: put error for /tempZone/home/rods/emptyfile, status = -510002 status = -510002 UNIX_FILE_OPEN_ERR, No such file or directory

This seems to be somewhat related to [#3127] and [#3094].

trel:
this is into a compound resource?

please capture the setup and then usage and error.

PR #4159, opened 2018-10-24
[3883] Updated code for zero length files

JustinKyleJames:
[#3883] Updated the code to treat zero length files the same as non-zero length
files.  The iput with --purgec and -k option now works the same for
both.


server/core/src/objDesc.cpp
```diff
@@ -156,6 +156,8 @@ fillL1desc( int l1descInx, dataObjInp_t *dataObjInp,
     keyValPair_t *condInput;
     char *tmpPtr;
 
+    L1desc[l1descInx].bytesWritten = -1;
```
trel: please comment and explain 'why'

PR #4160, opened 2018-10-24
[#3883] Updated code for zero length files

JustinKyleJames:
[#3883] Updated the code to treat zero length files the same as non-zero length
files.  The iput with --purgec and -k option now works the same for
both.


trel:
tests pass?  if so, add the # and we'll merge it.

Issue #2010, opened 2014-03-26
"iput -r ./" will create a collection named "."


trel:
- [x] master
- [x] 4-2-stable
---------

This is a surprise and is unusable.

Ideally, the "./" will get interpreted and converted to a real basename value.

Trailing slash is a tricky problem.

andrew-irods:
Possibly fixed with #3658 - test to prove.

trel:
The above table of options is no longer deemed desirable.

We are now resolving the `.` and `./` early and `iput -r` will create the target collection if it doesn't already exist.  `irsync` and `icp` now behave very similarly.

Confirmed fixed in 4.2.3.

Issue #1805, opened 2013-11-26
improve code coverage


trel:
_Originally Opened: @trel (2013-11-11 13:24:39_)

refactor to remove lines of red in our code coverage

most commonly, this has been using the new error assertion technique.
## 

From: https://code.renci.org/gf/project/e-irods/tracker/?action=TrackerItemEdit&tracker_item_id=1805

trel:
closing - adding new tests for coverage should be added as part of specific issues.

PR #4184, opened 2018-11-27
[#1805] wrote tests for PREP genquery iterators

d-w-moore:
** DO NOT MERGE YET ** pending resolution of 
   - successful testing of Python RE Genquery Iterators with iRODS master/4.3.0 in CI
-----
Cherry-picked from newdir_for_iput_options_test on 4-2-stable.

Created new tests, in test_prep_genquery_iterator.py, for
the Python RE plugin's upcoming (as of iRODS 4.2.5)
[genquery iterator](https://github.com/irods/irods_rule_engine_plugin_python).

Also ... altered test of recursive iput from unix path "./" (related to
irods/irods [#2010]), for reasons that one or more of the PREP iterator tests
listed below seemed at first to interact badly with, and cause spurious
failures of, the aforementioned iput test.  In the end, the instability with
iput is a real problem but not relevant to the tests described here.

Test of the PREP genquery iterator ***currently in force*** include:

    1 test_multiples_256

      - tests that we can query from 512 total results, yielding exactly
        that many rows. We augment the python rulebase in core.py with a
        function that directly executes the query, then generate and run
        a ".r" rule file in Python to call that function as the admin user.

    2 test_with_modified_rulebase_on_boundary

      - queries 512 objects, from the user0 account, using different
        page sizes; then checks the results (number of iterations and
        remainder , related to the quotient and modulo results of dividing
        512 by the page size) against expected/computed results.
        The test does this via a native RE ".r" file which calls a
        unit test function embedded in the genquery.py module.
        Page sizes are "0" (i.e. use row iterator), "7", "8" and "9" (for
        these last three values, the paged iterator is used with the
        corresponding integer #rows per page).

    3 test_with_modified_rulebase_off_boundary

      - Same as above but query off of a 256*N boundary, this time against
        (512 + 7 * 8 * 9) total results.

Tests of the iterator ***currently being skipped*** include:

    * The next 2 tests are skipped because they were simply rewrites of the
      "on/off boundary" tests described above:

    6 test_PREPonly_modified_rulebase_on_boundary

      - no longer needed (stood in for test#2 during test troubleshooting)

    7 test_PREPonly_modified_rulebase_off_boundary

      - no longer needed (stood in for test#3 during test troubleshooting)

    * The next two nested-iteration tests currently are skipped because
      of uncertain msiCloseGenQuery semantics in which the irods core code
      __possibly_but_not_yet_provably__ frees genquery object(s) invisibly
      to the caller despite a value of 1 returned in continueInx, but the
      free'ing behavior appears to happen exclusively in the case of a query
      ending on a boundary of N*256 results. Consistent with this conjecture
      is the strong indication that the internal free'ing behavior seems
      to correspond with a code of CAT_NO_ROWS_FOUND being returned to the
      caller (based on observations gleaned from monitoring memory usage
      while the "nested" tests ran at the bench).

    8 test_nested_paged_iterators

      - nests paged+row+paged iterators of non-overlapping sets, on three
        hierarchical levels, to produce 300*2*300 results total

    9 test_multiples_256_nested
      - the ambitious version of test # 1 above, wherein  2*256*512 results
        are queried in mixed row/row/paged iterator loops three-deep.

Conflicts:
	scripts/run_tests.py

### DIFF
--- a/scripts/irods/core_file.py
+++ b/scripts/irods/core_file.py
@@ -22,6 +22,16 @@ class CoreFile:
             raise IrodsError('unsupported rule engine for testing: {0}'.format(self.plugin_name))
         self.filepath = os.path.join(irods_config.core_re_directory, filename)
 
+    def prepend_to_imports(self, rule_text):
+
+        if not(isinstance(rule_text,str)) or not rule_text : return
+
+        if self.plugin_name == PYTHON_RULE_ENGINE_PLUGIN_NAME:
+            rule_text = rule_text.lstrip() + ("\n" if rule_text[-1:] != "\n" else "")
+            lib.prepend_string_to_file(rule_text, self.filepath)
+        else:
+            raise IrodsError('core rulebase cannot contain imports unless Python RE plugin installed')
+
     def add_rule(self, rule_text):
         if self.plugin_name == IRODS_RULE_LANGUAGE_RULE_ENGINE_PLUGIN_NAME:
             lib.prepend_string_to_file(rule_text, self.filepath)
--- a/scripts/irods/test/test_iput_options.py
+++ b/scripts/irods/test/test_iput_options.py
@@ -2,6 +2,7 @@ import os
 import re
 import stat
 import sys
+import shutil
 import ustrings
 
 if sys.version_info < (2, 7):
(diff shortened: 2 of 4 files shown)


trel:
This should now be possible to merge - please confirm commit message is true and that all tests run cleanly.

Otherwise, force-push and we'll let the tests run.

trel:
closing in favor of https://github.com/irods/irods/pull/4215

Issue #3729, opened 2017-11-11
remove XMsgServer


trel:
Externalize XMsg code to https://github.com/irods/deprecated_irods_xmsg_server

PR #4193, opened 2018-11-27
[#3729] Removed all code related to Xmsg.

korydraughn:
Created api number validator.
Cleaned up api number indentation and wrapped it in guards for clang-format.


trel:
looks good to me - awaiting codacy

Issue #4204, opened 2018-11-30
Paging results in irods_query results in stacktrace


alanking:
- [x] master
- [x] 4-2-stable

---

### iRODS Version, OS and Version
4.2.5

### What did you try to do?
Page through results using new query iterator. For example, running a query which returns 500 results would require a page fetch since results are only returned 256 (`MAX_SQL_ROWS`) at a time.

Originally discovered in irods/irods_capability_storage_tiering#51

### Expected behavior
All results in all sets are processed without error.

### Observed behavior (including steps to reproduce, if applicable)
When fetching the next page of results, a stacktrace is produced.

This is due to a combination of two things:

1. The `row_idx_` is not being reset between page fetches
2. All statements are being closed in the *_query_impl destructor:

https://github.com/irods/irods/blob/96a4f704351418f199f4e31194d1e28533bcfee0/lib/core/include/irods_query.hpp#L162-L180

PR #4205, opened 2018-12-01
[#4204] Fix paging results and closing statements (4-2-stable)

alanking:
Fixes row_idx_ going out of bounds between page_fetch() calls by
resetting to 0 in the query iterator.

Fixes stacktrace caused by closing all statements in the statement
table in the destructor of the query implementation object. This
has been changed to only close statements generated by that query.

---
[CI tests passed](http://172.25.14.63:8080/view/Personal/job/irods-build-and-test-workflow/1679/) (and [topo-resc](http://172.25.14.63:8080/job/topo-ub16-psql-resc/249/))

### DIFF
--- a/lib/core/include/irods_query.hpp
+++ b/lib/core/include/irods_query.hpp
@@ -107,10 +107,7 @@ namespace irods {
             }
 
             bool page_in_flight(const int row_idx_) {
-                if(row_idx_ < row_cnt()) {
-                    return true;
-                }
-                return false;
+                return (row_idx_ < row_cnt());
             }
 
             bool query_complete() {
@@ -141,6 +138,7 @@ namespace irods {
             }
 
             virtual int fetch_page() = 0;
+            virtual void reset_for_page_boundary() = 0;
             virtual ~query_impl_base() {
             }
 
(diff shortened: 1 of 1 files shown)


trel:
seems good - please add #

lib/core/include/irods_query.hpp
```diff
@@ -106,7 +106,7 @@ namespace irods {
                 return query_string_;
             }
 
-            bool page_in_flight(const int row_idx_) {
+            bool page_in_flight(const int& row_idx_) {
```
korydraughn: I don't think the `&` is really needed.  Creating a copy seems better, but the `const` pretty much achieves the same result.

lib/core/include/irods_query.hpp
```diff
@@ -106,7 +106,7 @@ namespace irods {
...
                 if(row_idx_ < row_cnt()) {
```
korydraughn: You can replace this `if` statement with a single `return`.

lib/core/include/irods_query.hpp
alanking: Got it. Thanks

lib/core/include/irods_query.hpp
alanking: Done

alanking:
[CI tests passed](http://172.25.14.63:8080/job/rule-engine-plugin-storage-tiering-build-test-workflow/237/) for storage tiering (currently the only consumer of this file)

PR #4209, opened 2018-12-05
Rule Engine Plugin Framework now supports continuation

korydraughn:
Please review the python test, including it's name.

Are we satisfied with the error codes used in the Passthrough REP?

Should an issue be created for additions to rodsAgent.cpp?
I added lines to set the log levels for each logger category.

[Passed CI](http://172.25.14.63:8080/view/Personal/job/irods-build-and-test-workflow/1700/)

### DIFF
--- a/plugins/rule_engines/irods_rule_engine_plugin-passthrough/libirods_rule_engine_plugin-passthrough.cpp
+++ b/plugins/rule_engines/irods_rule_engine_plugin-passthrough/libirods_rule_engine_plugin-passthrough.cpp
@@ -1,0 +1,167 @@
+#include "irods_plugin_context.hpp"
+#include "irods_re_plugin.hpp"
+#include "irods_configuration_keywords.hpp"
+#include "irods_get_full_path_for_config_file.hpp"
+#include "irods_logger.hpp"
+#include "irods_state_table.h"
+#include "rodsError.h"
+
+#include <string>
+#include <functional>
+#include <vector>
+#include <regex>
+#include <fstream>
+
+#include "json.hpp"
+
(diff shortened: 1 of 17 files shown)


plugins/rule_engines/irods_rule_engine_plugin-passthrough/libirods_rule_engine_plugin-passthrough.cpp
```diff
@@ -0,0 +1,162 @@
...
+    try {
+        for (const auto& re : config.at("plugin_configuration").at("rule_engines")) {
+            if (_instance_name == re.at("instance_name").get<std::string>()) {
+                for (const auto& e : re.at("plugin_specific_configuration").at("return_codes_for_peps")) {
+                    pep_configs.push_back({std::regex{e.at("regex").get<std::string>()}, e.at("code").get<int>()});
+                }
+
+                return SUCCESS();
+            }
+        }
```
alanking: Would it be helpful to break out the contents of the try block into a separate function and/or leaving an explanatory comment? Seems pretty straightforward, but I had to stare at it for a second to understand what was happening.

korydraughn: It should use the keywords if available.  I will update the code to use `irods::CFG_PLUGIN_SPECIFIC_CONFIGURATION_KW`.

I thought about making new ones, but other plugins do not define their keywords under `irods/irods`.  So, I think I'm going to leave the plugin specific keywords as is unless others feel differently.

PR #4215, opened 2018-12-07
[#1805] wrote tests for PREP genquery iterators

d-w-moore:
Created new tests  (test_prep_genquery_iterator.py) for the Python RE
plugin's upcoming [genquery iterator](https://github.com/irods/irods_rule_engine_plugin_python) capability.

In addition a change was made to a test of recursive iput from "./" in
test_iput_options.py that related to irods/irods [#2010].  This was in order
to stabilize an erratic behavior which (in addition to being problem in
itself) interacted badly in CI with the tests outlined here.

The tests are dependent on the existence of /etc/irods/genquery.py which,
as alluded to, will be present in versions >= 4.2.5  of the python rule plugin.

If the module is not seen at the path /etc/irods/genquery.py (such as if the
python rule plugin is not installed, or outdated), all tests of it will be
skipped.

Two iteration methods are offered by the module, namely result-at-a-time
and paged.

The tests currently-in-force are:

   1. test_multiples_256

      - tests that we can query from 512 total results, yielding exactly
        that many rows. We augment the python rulebase in core.py with a
        function that directly executes the query, then generate and run
        a ".r" rule file in Python to call that function as the admin user.

   2. test_with_modified_rulebase_on_boundary

      - queries 512 objects, from the user0 account, using different
        page sizes; then checks the results (number of iterations and
        remainder , related to the quotient and modulo results of dividing
        512 by the page size) against expected/computed results.
        The test does this via a native RE ".r" file which calls a
        unit test function embedded in the genquery.py module.
        Page sizes are "0" (i.e. use row iterator), "7", "8" and "9" (for
        these last three values, the paged iterator is used with the
        corresponding integer #rows per page).

   3. test_with_modified_rulebase_off_boundary

      - Same as above but query off of a 256*N boundary, this time against
        (512 + 7 * 8 * 9) total results.

The next two nested-iteration tests currently are skipped because
of uncertain msiCloseGenQuery semantics. There are times when iRODS
appears to internally free the objects (during the rsGenQuery call)
that hold the state of the query.  Of particular interest in applications
using the GenQuery is the part of that state that tracks the progress of
pagination (SQL_MAX_ROWS = 256 at a time) through the yielded results.

A caller typically monitors the value of continueInx returned by
msiGetMoreRows() to determine whether or not the call just made to that
microservice should be the last one. Ideally, msiCloseGenquery is also
called after the last call to msiGetMoreRows.

But there is also strong evidence that when msiGetMoreRows returns an error
code of CAT_NO_ROWS_FOUND, the appropriate assumption to be made by the calling
application is that iRODS has already terminated the pagination of results
and freed the memory and objects associated with the query.  This happens
in particular on the last page of query results returned when the total number
of matching results for the query is (N * SQL_MAX_ROWS) with N a positive
integer.

In genquery.py, we currently choose a conservative resource management policy
in this regard: i.e., in the Python code, we do not close the query ourselves.
This is not a problem for a short-running iRODS agent, but may lead to the
consumption of large amounts of memory if there are many queries made within
the lifetime of said iRODS agent. For this reason, we offer a choice:

   * A user can alter the iterator module policy can be altered from its default
   to that of ensuring msiCloseGenQuery is called at the proper times, based on
   the above heuristic.  This is configurable by assigning a boolean True value
   to a flag declared in the module (`genquery.AUTO_FREE_QUERIES`)

   * For the present time, however, `AUTO_FREE_QUERIES` is false by default, and
   thus we disable the especially query-heavy tests mentioned below. However,
   once the microservice API is boxed in and documented a bit better, we fully
   expect to be able to switch them back on.

These are the resource-heavy tests which are now disabled by default:

   4. test_nested_paged_iterators

      - nests paged+row+paged iterators of non-overlapping sets, on three
        hierarchical levels, to produce 300*2*300 results total

   5. test_multiples_256_nested

      - the ambitious version of test # 1 above, wherein  2*256*512 results
        are queried in mixed row/row/paged iterator loops three-deep.

### DIFF
--- a/scripts/irods/core_file.py
+++ b/scripts/irods/core_file.py
@@ -22,6 +22,16 @@ class CoreFile:
             raise IrodsError('unsupported rule engine for testing: {0}'.format(self.plugin_name))
         self.filepath = os.path.join(irods_config.core_re_directory, filename)
 
+    def prepend_to_imports(self, rule_text):
+
+        if not(isinstance(rule_text,str)) or not rule_text : return
+
+        if self.plugin_name == PYTHON_RULE_ENGINE_PLUGIN_NAME:
+            rule_text = rule_text.lstrip() + ("\n" if rule_text[-1:] != "\n" else "")
+            lib.prepend_string_to_file(rule_text, self.filepath)
+        else:
+            raise IrodsError('core rulebase cannot contain imports unless Python RE plugin installed')
+
     def add_rule(self, rule_text):
         if self.plugin_name == IRODS_RULE_LANGUAGE_RULE_ENGINE_PLUGIN_NAME:
             lib.prepend_string_to_file(rule_text, self.filepath)
--- a/scripts/irods/test/test_iput_options.py
+++ b/scripts/irods/test/test_iput_options.py
@@ -2,6 +2,7 @@ import os
 import re
 import stat
 import sys
+import shutil
 import ustrings
 import commands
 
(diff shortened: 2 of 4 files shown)


d-w-moore:
Pending successful run in CI.

PR #4218, opened 2018-12-11
[4197] moved stop operation into the cleanup

tempoz:
This addresses the resource cleanup problem with rule engines, and in particular fixes a problem expressed by the python rule engine plugin where it would not import a new core file for new agents.


tempoz:
When we are happy with the state of this commit, I need to cherry-pick it to 4-2-stable.

Issue #3776, opened 2018-01-04
Double free in main irodsServer during topology test


alanking:
- [x] master
- [x] 4-2-stable

---
When running tests in a resource server in a topology, there is a double free which occurs when attempting a socket creation. The double free is potentially tied to the failure described in #3775

Output from logs:
**icat.example.org** (rodsLog)
```
Jan  4 09:57:26 pid:12071 NOTICE: connectToRhostWithTout: myval error, errno = 111
Jan  4 09:57:26 pid:12071 NOTICE: connectToRhostWithTout: myval error, errno = 111
Jan  4 09:57:26 pid:12071 NOTICE: connectToRhostWithTout: myval error, errno = 111
Jan  4 09:57:26 pid:12071 NOTICE: connectToRhost: connect to host resource2.example.org on port 1247 failed, status = -305111 status = -305111 USER_SOCK_CONNECT_ERR, Connection refused
Jan  4 09:57:26 pid:12071 remote addresses: 172.25.14.106, ::1 ERROR: _rcConnect: connectToRhost error, server on resource2.example.org:1247 is probably down status = -305111 USER_SOCK_CONNECT_ERR, Connection refused
Jan  4 09:57:26 pid:12071 NOTICE: connectToRhostWithTout: myval error, errno = 111
Jan  4 09:57:26 pid:12071 NOTICE: connectToRhostWithTout: myval error, errno = 111
Jan  4 09:57:26 pid:12071 NOTICE: connectToRhostWithTout: myval error, errno = 111
Jan  4 09:57:26 pid:12071 NOTICE: connectToRhost: connect to host resource2.example.org on port 1247 failed, status = -305111 status = -305111 USER_SOCK_CONNECT_ERR, Connection refused
Jan  4 09:57:26 pid:12071 remote addresses: 172.25.14.106, ::1 ERROR: _rcConnect: connectToRhost error, server on resource2.example.org:1247 is probably down status = -305111 USER_SOCK_CONNECT_ERR, Connection refused
Jan  4 09:57:26 pid:12071 remote addresses: 172.25.14.106, ::1 ERROR: syncDataObjPhyPath:rsFileRename from /var/lib/irods/unix2RescVault/home/public/testfile.txt to /var/lib/irods/unix2RescVault/trash/home/otherrods/public/testfile.txt failed,status=-305111
```
**resource1.example.org:** (rodsLog)
```
[...]

PR #4235, opened 2019-01-14
[#3776] Prevent double free in agent spawner (master)

alanking:
The main server agent spawner which handles connection requests can
fail when communicating with the agent factory. On failure, the
connection request is freed by the spawning thread and continues.
Later, procChildren comes to clean up the connection request queue
and also frees the connection request, which results in a double free.

This change puts a new message exchange between the main server and
the agent factory when establishing the connection. After the connection
is made, an acknowledgement is sent from the main server to the agent
factory. If anything other than the expected value (a success message)
is received, the agent factory will not fork an agent process.

The server process now consistently cleans up the open sockets and
associated socket files before returning from execAgent.

---
[CI tests running](http://172.25.14.63:8080/view/Personal/job/irods-build-and-test-workflow/1743/)

### DIFF
--- a/server/core/include/rodsServer.hpp
+++ b/server/core/include/rodsServer.hpp
@@ -69,9 +69,6 @@ spawnAgent( agentProc_t *connReq, agentProc_t **agentProcHead );
 int
 execAgent( int newSock, startupPack_t *startupPack );
 int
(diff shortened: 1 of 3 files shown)


trel:
tests passed

PR #4252, opened 2019-01-29
[4040] add test for ireg with unresolvable owner (4-2-stable)

trel:


scripts/irods/test/test_ireg.py
```diff
@@ -209,3 +209,11 @@ def test_ireg_double_slashes__issue_3658(self):
         self.admin.assert_icommand('iscan {0}'.format(os.path.abspath(dirname)))
         shutil.rmtree(os.path.abspath(dirname), ignore_errors=True)
 
+    def test_ireg_file_with_unresolvable_owner__issue_4040(self):
+        filename = '../test/unresolvable_uid_testfile__issue_4040'
+        fullpath = os.path.abspath(filename)
+        # if file is there (should be for a package installation)
+        # attempt the ireg, should pass now that 4040 is fixed
+        if os.path.isfile(fullpath):
```
alanking: Should the test fail in the event that the file does not exist or otherwise indicate that the test did not run?

scripts/irods/test/test_ireg.py
```diff
@@ -209,3 +209,11 @@ def test_ireg_double_slashes__issue_3658(self):
...
+            self.admin.assert_icommand('ireg {0} {1}'.format(fullpath, self.admin.session_collection+'/'+os.path.basename(filename)))
```
alanking: Minor/optional: Could you use a list here to avoid the use of format/{0}{1}? Something like:
```
['ireg', fullpath, self.admin.session_collection+'/'+os.path.basename(filename))]
```

scripts/irods/test/test_ireg.py
trel: thought about issuing a pass or skipIf, but just not running 'ireg' if the file wasn't there seemed easier.   how would we indicate that it didn't run?

scripts/irods/test/test_ireg.py
trel: sure....  is avoiding format() something to aspire to?

scripts/irods/test/test_ireg.py
```diff
@@ -209,3 +209,13 @@ def test_ireg_double_slashes__issue_3658(self):
         self.admin.assert_icommand('iscan {0}'.format(os.path.abspath(dirname)))
         shutil.rmtree(os.path.abspath(dirname), ignore_errors=True)
 
+    def test_ireg_file_with_unresolvable_owner__issue_4040(self):
+        filename = '/tmp/irods_unresolvable_uid_testfile__issue_4040'
```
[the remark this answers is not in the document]
trel: hard-coded in the postinstall.sh script - need to hard-code it here

Issue #4040, opened 2018-08-01
username/groupname lookup fails


trel:
### iRODS Version, OS and Version

4.2.3, all OSs

### What did you try to do?

Register a file whose owner has been deleted from the system (no entry present in `/etc/passwd`)

### Expected behavior

Expected the file to be registered without incident

### Observed behavior (including steps to reproduce, if applicable)
```
$ ireg /tmp/alicefile /tempZone/home/rods/thealicefile
remote addresses: X.X.X.X ERROR: regUtil: reg error for /tempZone/home/rods/thealicefile, status = -122107 status = -122107 SYS_USER_RETRIEVE_ERR, Transport endpoint is not connected
 ERROR: main: regUtil error.  status = -122107 SYS_USER_RETRIEVE_ERR, Transport endpoint is not connected
```

It has been determined that this lookup is actually unnecessary/unused for registration.  iRODS stores the uid/gid gathered from the stat() command, and any username/groupname resolution can occur later if necessary.

Jollyfant:
@trel, we are getting an identical error and it seems to have to do with owner of the physical file location. Since we are running iRODS in a Docker container and mount an external archive that needs to be registered. This means that the owner of the archive does not exist inside our Docker container. Can you confirm this is probably the same issue?

trel:
That is correct.  We were resolving the uid, and if that resolution fails, an error was returned.

trel:
This fix was included in 4.2.4, needs a test to be closed.

PR #4258, opened 2019-01-30
[#4040] add test for ireg with unresolvable owner (4-2-stable)

trel:


trel:
oops, already upstream.  closing.

Issue #3998, opened 2018-06-28
remote microservice broken in iRODS 4.2.3


cookie33:
- [x] master
- [x] 4-2-stable

---
## Bug Report

When doing a `remote` microservice something is not properly handled/terminated. Because the next simple command fails and gives some errors in the iRODS logfile.
 
### iRODS Version, OS and Version

iRODS 4.2.3, Centos 7.5

### What did you try to do?

Run a test script to do remote execution. The test script `test_remote_6 .r` is
```
test {
   remote("sara-irods4.irods.surfsara.nl","<ZONE>tempZone</ZONE>") { writeLine("serverLog", " -- Remote Execution in Local Zone");  }
   writeLine("stdout", "Success!");
}
OUTPUT ruleExecOut
```

### Expected behavior

screen shows:
```
$ irule -F test_remote_6.r 
Success!
$
```

irods logfile contains:
```
Jun 28 12:11:11 pid:23182 NOTICE: writeLine: inString =  -- Remote Execution in Local Zone
```

### Observed behavior (including steps to reproduce, if applicable)

test run:
```
$ irule -F test_remote_6.r 

$
```
irods logfile contains:
```
Jun 28 12:30:29 pid:25034 NOTICE: writeLine: inString =  -- Remote Execution in Local Zone
Jun 28 12:30:29 pid:25034 remote addresses: 127.0.0.1, 145.100.4.16 ERROR: addMsParam: NULL msParamArray or label input
Jun 28 12:30:29 pid:25034 remote addresses: 127.0.0.1, 145.100.4.16 ERROR: addMsParam: NULL msParamArray or label input
```

We are missing the out of "Succes!" on the terminal where the rule was run.
We are also surprised by the extra ERROR data in the irods logfile

cookie33:
See also #3722. There the `remote` microservice was fixed.

xu-hao:
This bug was introduced in a recent refactoring to allow python rule engine to access `ruleExecOut` from `msParamArray` in `rei`. `remoteExec` needs to be updated to reflect that: 

https://github.com/irods/irods/blob/4-2-stable/plugins/rule_engines/irods_rule_engine_plugin-irods_rule_language/functions.cpp#L1920

trel:
We have a candidate fix - it's under review:
  https://github.com/irods/irods/pull/4011

beheerderdag:
Hi @trel, 
Is this slotted for 4.2.5? 
[...]

PR #4264, opened 2019-02-13
[#3998] Prevent reading of msParam with no type (master)

alanking:
In remote execution blocks, ruleExecOut is not always assigned a type.
This causes a strcmp on a null pointer to char in addMsParamToArray. This
change adds a check in smsi_remoteExec before copying around the ruleExecOut
param to ensure that its type is not null. A check is also done in
addMsParamToArray to ensure type is not null before attempting the strcmp.

---

[CI tests passed](http://172.25.14.63:8080/view/Personal/job/irods-build-and-test-workflow/1778/)


trel:
looks good - please add #s

Issue #3995, opened 2018-06-25
rcMisc.cpp, rsyncUtil.cpp have a global "using boost::filesystem" statement - change to local-scope namespace statements.


andrew-irods:
- [x] master
- [x] 4-2-stable
----

## Bug Report

The source files rcMisc.cpp and rsyncUtil.cpp have a global (available in the entire source file) namespace statement (right at the beginning of each file):

~~~
using namespace boost::filesystem;
~~~

This should be changed to use the namespace statement instead, within the scope of each function and object defined in this source file, which need boost::filesystem objects:

~~~
function / object definition ...
{
    namespace fs = boost::filesystem;
    .   .   .
    if (fs::exists(....) {
    .   .   .
etc.
~~~

### iRODS Version, OS and Version
4-2-stable (relevant to all versions and OS's)

PR #4285, opened 2019-02-19
[#3995] Moved usage of namespace boost::filesystem into functions (master)

korydraughn:
Replaced using namespace directives with namespace alias.


trel:
ready for # if you are

Issue #4301, opened 2019-03-01
boost::any_cast throws exception when renaming a data object and pep_api_data_obj_rename_post is defined


korydraughn:
- [x] master
- [x] 4-2-stable

### iRODS Version, OS and Version
iRODS v4.2.5+

### What did you try to do?
1. Added `pep_api_data_obj_rename_post` to `core.re`
2. Renamed an existing data object (e.g. `imv a.txt b.txt`)

### Expected behavior
No errors should appear in the log file.

### Observed behavior (including steps to reproduce, if applicable)
```
{"log_category":"legacy","log_facility":"local0","log_level":"error","log_message":"[-]\t../server/re/src/irods_re_serialization.cpp:689:irods::error irods::re_serialization::serialize_dataObjCopyInp_ptr(boost::any, irods::re_serialization::serialized_parameter_t &) :  status [INVALID_ANY_CAST]  errno [] -- message [failed to cast dataObjInp ptr]\n\t[-]\
t../server/re/src/irods_re_serialization.cpp:367:irods::error irods::re_serialization::serialize_dataObjInp_ptr(boost::any, irods::re_serialization::serialized_parameter_t &) :  status [INVALID_ANY_CAST]  errno [] -- message [failed to cast dataObjInp ptr]\n\n","request_api_name":"DATA_OBJ_RENAME_AN","request_api_number":627,"request_api_version":"d","req
uest_client_user":"rods","request_host":"152.54.8.75","request_proxy_user":"rods","request_release_version":"rods4.3.0","server_host":"kdd-ws","server_pid":14202,"server_timestamp":"2019-03-01T19:04:36.649830","server_type":"agent"}
 {"log_category":"legacy","log_facility":"local0","log_level":"error","log_message":"[-]\t../server/re/src/irods_re_serialization.cpp:703:irods::error irods::re_serialization::serialize_dataObjCopyInp_ptr(boost::any, irods::re_serialization::serialized_parameter_t &) :  status [INVALID_ANY_CAST]  errno [] -- message [failed to cast dataObjInp ptr]\n\t[-]\
[...]

PR #4304, opened 2019-03-04
[#4301] Fixed boost::any_cast and added test (master)

korydraughn:


trel:
looks good to me, please add #

Issue #4319, opened 2019-03-20
irepl to replication resource with all local leaf resources throws errors and creates orphan files


alanking:
- [x] master
- [x] 4-2-stable
---
### iRODS Version, OS and Version
4.1.12, 4.2.5

### What did you try to do?
```
$ ilsresc                  
replResc:replication
├── ufs1:unixfilesystem # all 3 resources reside on catalog consumer
├── ufs2:unixfilesystem
└── ufs3:unixfilesystem
rootResc:passthru
└── demoResc:unixfilesystem # resides on catalog provider

$ fallocate -l 3M /tmp/3M.test
$ iput -KR rootResc /tmp/3M.test
$ irepl -R replResc 3M.test
```

### Expected behavior
```
$ ils -L 3M.test
  rods              0 rootResc;demoResc      3145728 2019-03-20.11:01 & 3M.test
    sha2:u9Bc9gl6ybH4nqKdJULBt7Z+5GhIOTiV9ankP6H2IeU=    generic    /var/lib/irods/Vault/home/rods/3M.test
  rods              1 replResc;ufs1      3145728 2019-03-20.11:01 & 3M.test
    sha2:u9Bc9gl6ybH4nqKdJULBt7Z+5GhIOTiV9ankP6H2IeU=    generic    /tmp/irods/ufs1vault/home/rods/3M.test
  rods              2 replResc;ufs2      3145728 2019-03-20.11:01 & 3M.test
    sha2:u9Bc9gl6ybH4nqKdJULBt7Z+5GhIOTiV9ankP6H2IeU=    generic    /tmp/irods/ufs2vault/home/rods/3M.test
  rods              3 replResc;ufs3      3145728 2019-03-20.11:01 & 3M.test
    sha2:u9Bc9gl6ybH4nqKdJULBt7Z+5GhIOTiV9ankP6H2IeU=    generic    /tmp/irods/ufs3vault/home/rods/3M.test
```
No errors

### Observed behavior (including steps to reproduce, if applicable)
When running commands above on the catalog provider, all replicas appear correctly on disk and in the catalog. However, errors appear in the log and 0-length orphan files appear on disk:
```
$ ls -l /tmp/irods/ufs1vault/home/rods/3M.test
-rw------- 1 irods irods 3145728 Mar 20 11:01 /tmp/irods/ufs1vault/home/rods/3M.test
$ ls -l /tmp/irods/ufs2vault/home/rods/3M.test
[...]

PR #4320, opened 2019-03-21
[#4319] Convert source l3descInx for remote copies (4-2-stable)

alanking:
When performing a remote data copy, the destination l3descInx in the
dataOprInp structure passed to rcDataCopy is converted to match a valid
value on the remote machine. This also needs to be done for the source
l3descInx or else the remote copy may fail if it is not valid on the
other machine. This has been demonstrated on locally executed replications
being driven by a remote call.

### DIFF
--- a/scripts/irods/test/test_resource_types.py
+++ b/scripts/irods/test/test_resource_types.py
@@ -4433,6 +4433,32 @@ OUTPUT ruleExecOut
         self.assertEqual(num_out_of_date, 0)
         os.unlink(filename)
 
+    def test_irepl_to_consumer_repl_hier_from_provider__4319(self):
+        filename = 'test_irepl_to_consumer_repl_hier_from_provider__4319'
+        lib.make_file(filename, 1 * 1024 * 1024 + 1)
+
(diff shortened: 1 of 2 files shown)


scripts/irods/test/test_resource_types.py
```diff
@@ -4433,6 +4433,29 @@ def count_good_and_bad(ils_l_output):
         self.assertEqual(num_out_of_date, 0)
         os.unlink(filename)
 
+    def test_replicate_to_consumer_repl_hier_from_provider__4319(self):
+        filename = 'test_replicate_to_consumer_repl_hier_from_provider__4319'
+        lib.make_file(filename, 3 * 1024 * 1024)
```
trel: 1 * 1024 * 1024 + 1 (as minimal test?)

alanking: Will do. CI did not fail as expected without the fix commit, so I am looking into this test.

alanking:
Test reliably demonstrates the issue, but the bug seems not to manifest itself on fresh machines (possibly due to memory layout).

PR #4327, opened 2019-04-04
[4326] Update so that the plugins must set the physical path (4-2-stable)

JustinKyleJames:
[4326] Do not set the physical path in rsFileRename.cpp.  Set it in libunixfilesystem.cpp and libunivmss.cpp.


trel:
looks good - add some #

Issue #4356, opened 2019-04-19
Log says "Warning, pending SQL at cllDisconnect" a lot


alanking:
- [x] master
---
After the removal of audit trails (#4337) a message began appearing in the logs which is a lie. It should only appear if there is indeed "pending SQL at cllDisconnect".

PR #4357, opened 2019-04-19
[#4356] Make "pending SQL at cllDisconnect" true (master)

alanking:
A warning message is appearing in the logs stating that there is pending
SQL at cllDisconnect when in fact the number of pending statements is 0.
This change ensures that there are pending statements before attempting
to process or log them.

---
[CI tests running](http://172.25.14.63:8080/view/Personal/job/irods-build-and-test-workflow/1944/)

For reference, the comparison used to be:
https://github.com/irods/irods/blob/e521a02daa8af971fddc9beb62ad02dd57242632/plugins/database/src/low_level_odbc.cpp#L284

### DIFF
--- a/plugins/database/src/low_level_odbc.cpp
+++ b/plugins/database/src/low_level_odbc.cpp
@@ -272,43 +272,45 @@ cllCheckPending( const char *sql, int option, int dbType ) {
         return 0;
     }
 
-    /* but ignore a single pending "begin" which can be normal */
-    if ( pendingIx == 1 ) {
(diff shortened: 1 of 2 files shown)


alanking:
[CI tests passed](http://172.25.14.63:8080/view/Personal/job/irods-build-and-test-workflow/1950/) with addition of 30 second sleep in setUp of `test_icommands_file_operations`, except for ub14 because the Internet moved again. Created #4359 to address the sleep's necessity.

Issue #4344, opened 2019-04-08
Use new logging utilities in delay server


alanking:

alanking:
Delay server now logs to the main log file using the new logging utilities (master only).

trel:
reopening for last removal of delay log file

PR #4366, opened 2019-04-30
[#4344] Use syslog utility in delay server (master)

alanking:
As of this change, irodsReServer will no longer write to the reLog.

Adjusts tests to check inside server log for delay server messages.

---
[CI tests passed](http://172.25.14.63:8080/view/Personal/job/irods-build-and-test-workflow/)

Not sure why the irods::experimental::log::init() call is necessary (not needed for agent spawner process, which is what I based this change upon), but it seems that the irodsReServer process becomes a zombie if not. Am I holding it wrong, @korydraughn?

### DIFF
--- a/scripts/irods/configuration.py
+++ b/scripts/irods/configuration.py
@@ -388,10 +388,6 @@ class IrodsConfig(object):
         return paths.server_log_path()
 
     @property
-    def re_log_path(self):
-        return paths.re_log_path()
-
-    @property
     def server_bin_directory(self):
         return paths.server_bin_directory()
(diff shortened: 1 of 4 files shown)


korydraughn:
Your code is correct.  After the fork, the address space is replaced when `execv` is called.

See the line below:
https://github.com/irods/irods/blob/99ccfd2a9a29968b1e24ea2f4899c10a858d2205/server/core/src/rodsServer.cpp#L1216)

trel:
seems good - please #

Issue #4363, opened 2019-04-26
Can create resource hierarchy cycles


trel:
- [x] master
- [x] 4-2-stable
-------

### iRODS Version, OS and Version

4.2.5

### What did you try to do?

Add a child resource to self as parent (mistakenly)

### Expected behavior

See an error message saying
```
Cannot add child [<resc_name>] to parent [<resc_name>]
```

### Observed behavior (including steps to reproduce, if applicable)

Child was added successfully.

Resource now invisible to `ilsresc`

alanking:
Apparently, you can also add a child as a parent to its parent or grandparent (and so on) and it is hidden from ilsresc:
```
$ ilsresc
test_passthru0:passthru
└── test_passthru1:passthru
    └── test_ufs:unixfilesystem
$ iadmin addchildtoresc test_ufs test_passthru0
```
In other words, we should probably rename this issue to address general hierarchical cycles.

trel:
agreed.

alanking:
This should be addressed with the following logic:
1. Child-to-be must be a root (aka no parent)
2. Parent-to-be must not have child-to-be as its root

Pseudocode:
```
if !child_to_be.parent.empty:
    return CHILD_HAS_PARENT
if parent_to_be.root == child_to_be:
    return HIERARCHY_ERROR
addchildtoresc()
```

Needs two tests based on this simple case:
```
A
└── B
```
1. Add A as a child of A
2. Add A as a child of B

PR #4376, opened 2019-05-06
[#4363] Return error when adding child resource creates a hierarchy cycle (master)

korydraughn:
Docker - Passed
[CI](http://172.25.14.63:8080/view/Personal/job/irods-build-and-test-workflow/2104/) - Passed (with false positive)

### DIFF
--- a/lib/core/include/rodsErrorTable.h
+++ b/lib/core/include/rodsErrorTable.h
@@ -474,6 +474,7 @@ NEW_ERROR(CAT_INVALID_CHILD,                           -857000)
 NEW_ERROR(CAT_INVALID_OBJ_COUNT,                       -858000) // hcj
 NEW_ERROR(CAT_INVALID_RESOURCE_NAME,                   -859000) // JMC
 NEW_ERROR(CAT_STATEMENT_TABLE_FULL,                    -860000) // JMC
+NEW_ERROR(CAT_RESOURCE_NAME_LENGTH_EXCEEDED,           -861000)
 /** @} */
 
 /* 880,000 to 889,000  Deprecated  */
--- a/server/api/src/rsGeneralAdmin.cpp
+++ b/server/api/src/rsGeneralAdmin.cpp
@@ -17,6 +17,8 @@
 // stl includes
 #include <iostream>
 #include <string>
+#include <tuple>
+#include <optional>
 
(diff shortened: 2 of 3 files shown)


server/api/src/rsGeneralAdmin.cpp
```diff
@@ -233,8 +236,28 @@ _addChildToResource(
...
+        if (std::strncmp(_generalAdminInp->arg2, _generalAdminInp->arg3, parent_resc_length) == 0) {
+            std::string msg = "Cannot add child resource [";
+            msg += _generalAdminInp->arg3;
+            msg += "] to parent resource [";
+            msg += _generalAdminInp->arg2;
+            msg += "].";
+
+            // TODO Investigate how to detect the verbose flag.
+            // Add this message to the rError stack should happen automatically.
+            // The apiHanlder does not know about the verbose flag for iadmin yet.
```
trel: apiHanlder -> apiHandler

lib/core/include/rodsErrorTable.h
```diff
@@ -784,6 +784,7 @@ NEW_ERROR(INVALID_KVP_STRING,                   -1826000)
 NEW_ERROR(PLUGIN_ERROR_MISSING_SHARED_OBJECT,   -1827000)
 NEW_ERROR(RULE_ENGINE_ERROR,                    -1828000)
 NEW_ERROR(REBALANCE_ALREADY_ACTIVE_ON_RESOURCE, -1829000)
+NEW_ERROR(INVALID_RESOURCE_HIERARCHY,           -1830000)
```
alanking: FWIW, `HIERARCHY_ERROR` is an existing error which may be equally effective. Otherwise, I would argue this should have a more specific name to inform the user as to what happened. Although, I suppose the client error stack will contain this information...

trel: i think i agree, HIERARCHY_ERROR should already communicate enough, along with the specific error message being sent back.

korydraughn:
Are we happy with the error messages and the new error code?

trel:
Does the SYS_RESC_DOES_NOT_EXIST error specify which one was 'bad'?

korydraughn:
I'll try and restore the error codes.

korydraughn:
Added #

PR #4377, opened 2019-05-06
[#4363] Return error when adding child resource creates a hierarchy cycle (4-2-stable)

korydraughn:
Docker - Passed
[CI](http://172.25.14.63:8080/view/Personal/job/irods-build-and-test-workflow/2105/) - Passed


lib/core/include/rodsErrorTable.h
```diff
@@ -770,6 +770,7 @@ NEW_ERROR(INVALID_KVP_STRING,                   -1826000)
 NEW_ERROR(PLUGIN_ERROR_MISSING_SHARED_OBJECT,   -1827000)
 NEW_ERROR(RULE_ENGINE_ERROR,                    -1828000)
 NEW_ERROR(REBALANCE_ALREADY_ACTIVE_ON_RESOURCE, -1829000)
+NEW_ERROR(INVALID_RESOURCE_HIERARCHY,           -1830000)
```
trel: could be convinced this needs a new error type... or that we could use `HIERARCHY_ERROR` as is...

korydraughn:
Added #

Issue #4369, opened 2019-05-03
Create irods::query_processor class


alanking:
- [x] master
- [x] 4-2-stable

## Feature

Encapsulate the `irods::thread_pool` into an interface which takes a query string and a `std::function` representing a unit of work. The `std::function` will by contract take a `irods::query::value_type` which represents the results of the query.

This object holds the selected columns in order as described by the query string. In the below example, these would be COLL_NAME and DATA_NAME.

The object could return a `std::future<std::vector<std::tuple<int, std::string>>>` which would represent the error codes/messages of any failed jobs.

Use cases:
- Delay server
- Rebalance
- Recursive operations (remove, replication, rename, ...)

Proposed usage for recursive unlink:
```c++
irods::thread_pool thread_pool{std::thread::hardware_concurrency() / 2};
std::string query_string{
    "select COLL_NAME, DATA_NAME where COLL_NAME like '/tempZone/home/rods/delete_me%'"};
auto job{[](const auto& _query_result){
        try {
            irods::filesystem::remove(
                irods::filesystem::path(_query_result[0]) / _query_result[1]);
        } catch(const irods::exception& e) {
            // log message for server, stderr for client
        }
    }};
irods::query_processor job_processor{query_string, job}; // new!
auto errors = job_processor.execute(thread_pool);

for(const auto& error : errors.get()) {
    // process each returned error
}
```

PR #4388, opened 2019-05-14
[#4369] Added new query_processor library (master)

korydraughn:
Also adds a new error code for situations where the error is unknown.

[Passed CI](http://172.25.14.63:8080/view/Personal/job/irods-build-and-test-workflow/1998/)

### DIFF
--- a/cmake/development_library.cmake
+++ b/cmake/development_library.cmake
@@ -87,6 +87,7 @@ set(
   ${CMAKE_SOURCE_DIR}/lib/core/include/irods_tcp_object.hpp
   ${CMAKE_SOURCE_DIR}/lib/core/include/irods_threads.hpp
   ${CMAKE_SOURCE_DIR}/lib/core/include/irods_virtual_path.hpp
+  ${CMAKE_SOURCE_DIR}/lib/core/include/query_processor.hpp
   ${CMAKE_SOURCE_DIR}/lib/core/include/lsUtil.h
   ${CMAKE_SOURCE_DIR}/lib/core/include/mcollUtil.h
   ${CMAKE_SOURCE_DIR}/lib/core/include/microservice.hpp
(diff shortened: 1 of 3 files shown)


trel:
seems good to me, let's # and get it merged - would be nice to have in 4.2.6, and won't hurt anything since we aren't wiring it up for use, yet.

cmake/development_library.cmake
```diff
@@ -87,6 +87,7 @@ set(
   ${CMAKE_SOURCE_DIR}/lib/core/include/irods_tcp_object.hpp
   ${CMAKE_SOURCE_DIR}/lib/core/include/irods_threads.hpp
   ${CMAKE_SOURCE_DIR}/lib/core/include/irods_virtual_path.hpp
+  ${CMAKE_SOURCE_DIR}/lib/core/include/query_processor.hpp
```
alanking: Nit: I haven't audited this list but alphabetization would be nice here.

lib/core/include/rodsErrorTable.h
```diff
@@ -233,6 +233,7 @@ NEW_ERROR(SYS_SOCK_WRITE_ERR,                          -161000)
 NEW_ERROR(SYS_SOCK_CONNECT_ERR,                        -162000)
 NEW_ERROR(SYS_OPERATION_IN_PROGRESS,                   -163000)
 NEW_ERROR(SYS_REPLICA_DOES_NOT_EXIST,                  -164000)
+NEW_ERROR(SYS_UNKNOWN_ERROR,                           -165000)
```
alanking: `SYS_INTERNAL_ERR` may communicate the same amount of information if we want to avoid adding an new error.

korydraughn: This new error code was introduced during a group code review for cases where devs choose to use -1 to indicate an error occurred, but aren't sure what happened.

korydraughn:
Added #.

Issue #4391, opened 2019-05-15
ifsck has an integer overflow


trel:
- [x] master
- [x] 4-2-stable

## Bug Report

From https://groups.google.com/forum/#!topic/irod-chat/9XCUxSIHcOE

### iRODS Version, OS and Version

4.2+

### What did you try to do?

ifsck

### Expected behavior

good files not singled out with "CORRUPTION"

### Observed behavior (including steps to reproduce, if applicable)

```
CORRUPTION: local file [/asresc/home/adam/projects/RIS/data/fastq/wgs/429TG/HYKKCCCXY__AS_429TG_S7_L007_R1_001.fastq.gz] size [46089239651] not consistent with iRODS object [/resarchivezone/home/adam/projects/RIS/data/fastq/wgs/429TG/HYKKCCCXY__AS_429TG_S7_L007_R1_001.fastq.gz] size [-1155400605].
CORRUPTION: local file [/asresc/home/adam/projects/RIS/data/alignment/wgs/421TG/421TG.realigned-recalibrated.bam] size [179052998583] not consistent with iRODS object [/resarchivezone/home/adam/projects/RIS/data/alignment/wgs/421TG/421TG.realigned-recalibrated.bam] size [-1335627849].
CORRUPTION: local file [/asresc/home/adam/projects/RIS/data/alignment/wgs/44/RIS44.realigned-recalibrated.bam] size [186157451760] not consistent with iRODS object [/resarchivezone/home/adam/projects/RIS/data/alignment/wgs/44/RIS44.realigned-recalibrated.bam] size [1473858032].
```

Note...
`46089239651 + 1155400605 = 11 * 2^32`
`179052998583 + 1335627849 = 42 * 2^32`
`186157451760 + 1473858032 = 43 * 2^32`

trel:
Threshold appears to be 2G.

```
$ truncate -s2G 2gfile
$ truncate -s2147483647 justunder2g
$ imkdir ifscktest
$ iput 2gfile ifscktest
$ iput justunder2g ifscktest
$ ils -L ifscktest
/tempZone/home/rods/ifscktest:
  rods              0 demoResc   2147483648 2019-05-14.21:52 & 2gfile
        generic    /var/lib/irods/Vault/home/rods/ifscktest/2gfile
  rods              0 demoResc   2147483647 2019-05-14.21:52 & justunder2g
        generic    /var/lib/irods/Vault/home/rods/ifscktest/justunder2g
[...]

Issue #1887, opened 2014-01-12
Test suite is not idempotent


sjdrodge:
First run on a clean install turned up two failing tests, second run turned up three failing tests.

List of failed tests (first one shows up on the second run only):

18 - iadmin mkresc testresource2 "unix file system" "turing:/tmp/testresc2"  ====> error code = 1024
241 - iadmin rmresc testresource2  ====> error code = 1024
247 - iadmin lr  ====> error code = 0

alanking:
The test suite is quite stable at this point. Should any specific issues arise which have not already been addressed, please open new issues for those.

PR #4397, opened 2019-05-17
[#4391] Fixed integer overflow for ifsck (master)

korydraughn:
[Passed CI](http://172.25.14.63:8080/view/Personal/job/irods-build-and-test-workflow/2013/) (centos 7 - journald suppressed expected output) and Docker.

### DIFF
--- a/scripts/irods/test/test_ifsck.py
+++ b/scripts/irods/test/test_ifsck.py
@@ -1,0 +1,45 @@
+import os
+import sys
+
+if sys.version_info < (2, 7):
+    import unittest2 as unittest
+else:
+    import unittest
+
+from . import session
+from .. import test
+from .. import lib
+
+class Test_Ifsck(session.make_sessions_mixin([('otherrods', 'rods')], []), unittest.TestCase):
+
+    def setUp(self):
+        super(Test_Ifsck, self).setUp()
+        self.admin = self.admin_sessions[0]
+
+    def tearDown(self):
+        super(Test_Ifsck, self).tearDown()
+
+    @unittest.skipIf(test.settings.RUN_IN_TOPOLOGY, "Skip for Topology Testing")
+    def test_ifsck_has_integer_overflow__issue_4391(self):
+        # Create a file that is exactly 2G in size.
+        file_2g = '2g_issue_4391.txt'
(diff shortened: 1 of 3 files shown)


scripts/irods/test/test_ifsck.py
```diff
@@ -0,0 +1,45 @@
...
+
+    @unittest.skipIf(test.settings.RUN_IN_TOPOLOGY, "Skip for Topology Testing")
+    def test_ifsck_has_integer_overflow__issue_4391(self):
+        # Create a file that is exactly 2G in size.
+        file_2g = '2g_issue_4391.txt'
+        lib.make_file(file_2g, 2147483648, 'arbitrary')
+
+        # Create another file that is one byte smaller than 2G in size.
+        file_just_under_2g = 'just_under_2g_issue_4391.txt'
+        lib.make_file(file_just_under_2g, 2147483647, 'arbitrary')
```
alanking: I am just now noticing that there are several places in our tests where we are calling `lib.make_file` and then failing to unlink/remove the file at the end of the test. So, it would be nice to have a couple of `os.unlink` calls at the end of this test for these two files (especially given the size).  Not so big of a deal for CI/docker but I thought I'd point it out just in case. #1887 may be the place for addressing these types of concerns...

trel: Agreed - for completeness... but, note that if we never leave the 'session' collection, then they all get cleaned up with the session 'irm -rf' cleanup.

alanking: Sorry, I should have clarified that I meant the file on disk that we are `iput`-ing - which is separate from the file created in the vault for the data object represented in iRODS - is not being cleaned up. Another example is `test_rebalance_different_sized_replicas__3486`. I have a local file called "test_rebalance_different_sized_replicas__3486" leftover in the directory from which the test was run. I end up having to clean up the file myself later. Like I said, not a big deal, and actually doesn't affect the result of running the same tests over and over in these two cases.

trel: Okay - yes, I thought those all were/are being cleaned up.   We should make the tests leave no trace.

PR #4399, opened 2019-05-18
[4369] add query limit to query_processor (4-2-stable)

jasoncoposky:


trel:
noob question... is this all it needs?   where is it handled?

jasoncoposky:
yes, it is passed through to the query iterator which [enforces the limit](https://github.com/irods/irods/blob/master/lib/core/include/irods_query.hpp#L84).

trel:
easy enough

trel:
i note that limit is `uint32_t`, and this one is `uintmax_t`

jasoncoposky:
fair.

lib/core/include/query_processor.hpp
```diff
@@ -69,6 +70,7 @@ namespace irods
     private:
         std::string query_;
         job job_;
+        uintmax_t limit_;
```
trel: still uintmax_t

PR #4405, opened 2019-05-21
[4369] add query type to query_processor (master)

jasoncoposky:


lib/core/include/query_processor.hpp
```diff
@@ -23,12 +23,14 @@ namespace irods
         using errors     = std::vector<error>;
         using result_row = typename query<ConnectionType>::value_type;
         using job        = std::function<void (const result_row&)>;
+        using query_type = typename query<ConnectionType>::query_type;
```
korydraughn: We should probably add `_type` to the rest of the type aliases since they are in the same class.

korydraughn: I just realized that query_type is already in line with the other aliases. Its fine as it is.

trel:
ready to add the # ?

PR #4436, opened 2019-07-05
Bundle of dstream fixes (master)

korydraughn:


trel:
we need tests for these?

korydraughn:
I plan on writing unit tests for these. These changes allow dstream objects to support move semantics. We don't any way in 4-2-stable to test these changes though.

korydraughn:
Added #

Issue #4153, opened 2018-10-16
irmdir -p removes home directory (if empty)


alanking:
- [x] master
- [x] 4-2-stable
------

### iRODS Version, OS and Version
4.2.4

### What did you try to do?
As user `rods`:
```
$ ils -l
/tempZone/home/rods:
  C- /tempZone/home/rods/testdir
$ irmdir -p testdir
Failed to remove [/tempZone/home]: Collection is not empty
$ ils -l
remote addresses: x.x.x.x ERROR: lsUtil: srcPath /tempZone/home/rods does not exist or user lacks access permission
$ ils -l /tempZone/home
/tempZone/home:
  C- /tempZone/home/public
```

### Expected behavior
Should we delete all the way up to (and including) the user's home directory? Or is this the expected behavior?

korydraughn:
I think the behavior is wrong. Assuming this is based on `rmdir`, it should delete only the elements in the path given. `rmdir` provides an example that demonstrates what using `-p` is like. It appears to treat relative paths differently than absolute paths.
```bash
'rmdir -p a/b/c' is similar to 'rmdir a/b/c a/b a'
```
From stackexchange: https://unix.stackexchange.com/questions/22104/what-exactly-does-rmdir-p-do

So for the example provided in the issue, `testdir` should be the only thing deleted. The parent collections should not have been touched.

PR #4445, opened 2019-07-10
[#4153] irmdir now properly handles "-p" (4-2-stable)

korydraughn:
[Passed CI with false positive](http://172.25.14.63:8080/view/Personal/job/irods-build-and-test-workflow/2110/)


trel:
looks like a false negative ( a red dot due to a vSphere hiccup )

Issue #4060, opened 2018-08-13
collection with backslash in name causes problems for ipwd


tedgin:
## Bug Report

- [x] master
- [x] 4-2-stable
- [x] icommands master
- [x] icommands 4-2-stable

### iRODS Version, OS and Version
I'm using iRODS server version 4.1.10 on CentOS 6.

### What did you try to do?
I tried to display the current working collection for a collection with a backslash in its name.  It rendered the path with the `\f` being interpreted as a control character.

```bash
$ icd 'test\folder'
$ ipwd
/iplant/home/tedgin/test
                        older
```

### Expected behavior
I expected to see the absolute path to the collection proper collection name without the `\f` being interpreted as a line feed.

trel:
also, `ils`:

```
$ imkdir 'test\folder'
$ icd 'test\folder'
$ ipwd
/tempZone/home/rods/test
                        older
$ ils
remote addresses: 152.54.8.147 ERROR: lsUtil: srcPath /tempZone/home/rods/test
                                                                              older does not exist or user lacks access permission
ERROR(4) $ icd
$ irm -rf 'test\folder'
$
```

PR #4447, opened 2019-07-12
[#4060] Paths now escape '\f' character (master)

korydraughn:
Passed in new CI (core tests only)


trel:
did this succeed?

trel:
how are these looking?

korydraughn:
Added #.

PR #4448, opened 2019-07-12
[#4060] Paths now escape '\f' character (4-2-stable)

korydraughn:
Passed in new CI (core tests and federation only)


korydraughn:
Added #.

Issue #4438, opened 2019-07-06
Statement table entries exhausted by sequential python genquery calls


d-w-moore:
- [x] master
- [x] 4-2-stable
-----

### iRODS Version, OS and Version
irods 4.2.6 on Ubuntu 16.04 , python RE plugin loaded
`irule -a` produces:
```
Level 0: Available rule engine plugin instances:
	irods_rule_engine_plugin-irods_rule_language-instance
	irods_rule_engine_plugin-python-instance
```

### What did you try to do?
Use the python genquery row_iterator to produce >S query objects sequentially (S being 50, the size of the statement table) within the same agent process.
Here, we reproduce the same problem reported in [this post](https://groups.google.com/d/msg/iROD-Chat/u-fds230M4U/YTlCLqtmCAAJ)

### Expected behavior
Each query should auto-destruct as soon as it goes out of scope, allowing an indefinite number of queries to be done within the same process.

### Observed behavior (including steps to reproduce, if applicable)
Place the following text at head of `/etc/irods/core.py`:
```
import genquery
def query_51(rule_args, callback, rei):
    for y in range(51):
        for x in genquery.row_iterator( 'COLL_NAME', "COLL_NAME like '{}'".format("/tempZone/home/rods"),
                                        genquery.AS_LIST, callback):
            pass
```
Then do the following at the `irods` command prompt:
```
irule query_51 null null
```

This produces:
```
remote addresses: 192.168.122.200 ERROR: rcExecMyRule error.  status = -860000 CAT_STATEMENT_TABLE_FULL
Level 0: DEBUG: 
```
and looking at the logs, we see the telltale **too many concurrent statements** :
```
Jul  6 06:39:11 pid:6008 remote addresses: 192.168.122.200, ::1 ERROR: cllExecSqlWithResult: too many concurrent statements
Jul  6 06:39:11 pid:6008 NOTICE: chlGenQuery cmlGetFirstRowFromSql failure -860000
Jul  6 06:39:11 pid:6008 NOTICE: _rsGenQuery: genQuery status = -860000
Jul  6 06:39:11 pid:6008 NOTICE: rsGenQuery: _rsGenQuery failed, status = -860000
Jul  6 06:39:11 pid:6008 remote addresses: 192.168.122.200, ::1 ERROR: caught python exception: Traceback (most recent call last):
  File "/etc/irods/core.py", line 31, in query_51
    genquery.AS_LIST, callback):
[...]

PR #4452, opened 2019-07-16
[#4438] test fix for PREP genquery iter 'too many concurrent stmts'

d-w-moore:


d-w-moore:
The 'unused variable' objections by Codacy are all attributable to use of the Python idiom: "...{var}...".format(**locals()) and can be safely ignored

Issue #4459, opened 2019-07-25
upgrading on oracle from 4.1 to 4.2 populates resc_id incorrectly


trel:
- [x] master
- [x] 4-2-stable
------

#  Bug Report

### iRODS Version, OS and Version

4.2.x

### What did you try to do?

Upgrade oracle provider from 4.1 to 4.2.6

### Expected behavior

Data is visible after upgrade, same as before upgrade.

### Observed behavior (including steps to reproduce, if applicable)

Upgrade 'completes'... server stands up...

```
$ ils
foo
$ ils -L
$
```

The `resc_id` field in the `r_resc_main` table has the correct entries... *except* they're missing the last digit.  Therefore, `ils -L` cannot match the correct resource to display, and returns an empty (or incorrect) list.

```
Connected to:
Oracle Database 11g Enterprise Edition Release 11.2.0.1.0 - 64bit Production
With the Partitioning, OLAP, Data Mining and Real Application Testing options

SQL> select RESC_NAME, RESC_ID from r_resc_main;

RESC_NAME
--------------------------------------------------------------------------------
   RESC_ID
----------
bundleResc
      9100

demoResc
      9101

SQL> select DATA_NAME, DATA_ID, RESC_ID from r_data_main;

DATA_NAME
--------------------------------------------------------------------------------
   DATA_ID    RESC_ID
---------- ----------
foo
     10513	 910
```

aka, the `resc_id` entry for `foo` should be '9101' but the database only has '910'.

trel:
local test run...  appears to be an implicit cast bug...
```
irods@icat:~/scripts$ python -c 'import irods.trel.dbtest; irods.trel.dbtest.trel_db_testing()'
------- show db right now
('foo', Decimal('10513'), Decimal('9101'))
------- set to 1234 state
------- show 1234 state
('foo', Decimal('10513'), Decimal('1234'))
------- walk the resources, update r_data_main -- as current 41->42 code
------- show the new stuff
('foo', Decimal('10513'), Decimal('910'))
------- walk the resources, update r_data_main -- with int()
------- show the new stuff
('foo', Decimal('10513'), Decimal('9101'))
```

PR #4460, opened 2019-07-25
[4459] fix implicit cast bug when talking to oracle (4-2-stable)

trel:
When pypyodbc talks to oracle, the implicit cast from
string->int->numbers(38) was dropping the last character of the string.

It is not yet clear if this is a pypyodbc bug or Oracle ODBC bug.

--

Needs to go through CI to confirm mysql/postgres not affected by this new int().


trel:
forgot the # for #4459

PR #4465, opened 2019-07-31
[#4446] Removed icommands as an installation dependency (master)

korydraughn:


test/post_install_test/main.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+        if (!out) {
+            std::cerr << "Error: cannot open data object for writing [" << data_obj_path << "]\n";
+            return 1;
+        }
+
+        out << test_string;
+    }
+
+    // Check that the new data object contains the correct content.
```
trel: should also check the timestamp is fresh?   in case there was already one there from an earlier test run?

korydraughn: Added timestamp to name.

korydraughn:
Are the permissions correct for this binary?  In particular, SETUID?
Are we okay with the naming of things?

PR #4469, opened 2019-08-05
Issue 4387 recursive collection tickets (4-2-stable)

jasoncoposky:


plugins/database/src/general_query.cpp
```diff
@@ -2206,6 +2206,7 @@ checkCondInputAccess( genQueryInp_t genQueryInp, int statementNum,
         else {
             zoneName = genQueryInp.condInput.value[zoneIx];
         }
+rodsLog(LOG_NOTICE, "XXXX %s:%d", __FILE__, __LINE__);
```
trel: remove debug

jasoncoposky: Previous there was no recursive behavior.  The original query worked as intended, the only ticket applied to a collection of data objects was a ticket of the immediate parent.  This commit provides the desired recursive behavior by modifying the query and walking the collection path.

trel:
> Previous there was no recursive behavior. The original query worked as intended, the only ticket applied to a collection of data objects was a ticket of the immediate parent. This commit provides the desired recursive behavior by modifying the query and walking the collection path.

 i like it.    if passing tests, please to add #.

Issue #4495, opened 2019-08-30
ibun fails to unzip a 3.2GB zip file


hyi:
- [x] master
- [x] 4-2-stable
---
This issue can be reproduced with this 3.2GB zip file: http://people.renci.org/~hongyi/irods/OcheyedanRiver.zip using this command ```ibun -xDzip```
It can be reproduced in iRODS version 4.2.6. @trel is able to reproduce the issue in iRODS 4.2.6. When running this ibun command to unzip this file, ibun immediately returns without doing anything. No relevant messages are recorded in the log.

alanking:
The approach we have for fixing the extract operation uses libarchive to iterate over the structured file and register its contents in the catalog rather than pulling the entire file into memory and reading it in that way. See [libarchive documentation](https://github.com/libarchive/libarchive/wiki/Examples) for examples. This includes a hard-coded value of 1MB for the "block size" in the structfile resource plugin. Here I will provide some justification for our decision using benchmark test results. Notes about the test:
1. Used a RAM disk for the unixfilesystem resource vault
2. Figures are approximate
3. `unzip` took ~44 seconds to extract the contents of the file, for comparison

block size (KB) | 4 | 8 | 16 | 32 | 64 | 128 | 256 | 512 | 1024
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
ibun -x time (sec) | ~129 | ~79 | ~54 | ~43 | ~36 | ~34 | ~32 | ~31 | ~30

![image](https://user-images.githubusercontent.com/8021100/64458012-33096f80-d0c2-11e9-909c-becd4eb6e262.png)

The data show a significant initial performance gain as block size increases from 4KB with the effect tapering off as the value increases. This is because the number of iterations needed to read the file decreases as the block size increases.

PaulVanSchayck:
Would this also help for issue #4118?

trel:
This issue was due to crossing an arbitrary 4GB threshold defined by using an integer and not iterating over blocks of the zip file in question.  The files in #4118 are much bigger, so ... it's probably not exactly the same problem.  We'll investigate there as well.

alanking:
I was able to demonstrate that the linked code changes also fix the problem described in #4118.

PaulVanSchayck:
Thanks Alan. The issues sounded so similar that I thought there must be overlap.

alanking:
Test needs to copy a different directory as CI runs out of `/var/lib/irods/scripts`

PR #4498, opened 2019-09-06
[#4495] Chunk libarchive reads for extractions (4-2-stable)

alanking:
The extract operation in structfile resource plugin reads in entire files into a memory buffer without regard for size. If the file in question overflows an int when its size in bytes is calculated, libarchive will fail to read the file and extraction will do nothing. This change iterates over the file using a block size and the buffer pointer in use today. The block_size has been hard-coded to 1MB as performance gains above this are negligible, and block sizes below this only become increasingly worse as the value decreases.

The make_file utility in lib.py grew an option to pass a block size as well. This allows for much larger files than was possible before as dd will not always be able to generate a file large enough to meet the requested size in the current implementation. Now the count and bs parameters are used to generate files properly with an append of any leftover bytes in the event of an uneven division.

### DIFF
--- a/plugins/resources/structfile/libstructfile.cpp
+++ b/plugins/resources/structfile/libstructfile.cpp
@@ -205,10 +205,10 @@ int irods_file_open_for_write(
 ssize_t irods_file_read(
     struct archive* _arch,
     void*           _data,
-    const void**    _buff ) {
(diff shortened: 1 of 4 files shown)


trel:
seems good to me - # please

Issue #4118, opened 2018-09-21
msiTarFileExtract not working on .tar files over 11GB in size


matthewsaum:
- [x] master
- [x] 4-2-stable
---
## Bug Report
Running the "msiTarFileExtract" microservice is not re-creating the files from within the tarball larger than about 11GB. Haven't nailed down exact size to bytes.
### iRODS Version, OS and Version
iRODS 4.1.10 and iRODS 4.2.4
CentOS 7.5
### What did you try to do?
Run msiCreateTarfile on a collection, then attempt to run msiTarFileExtract on the newly created tarball.

### Expected behavior
The new empty collection would be re-populated with the contents of the tarball

### Observed behavior (including steps to reproduce, if applicable)
The tarball did NOT extract the files. The rule finishes nearly instantly.

The example tar.r rule

> myTestRule {
>    msiTarFileCreate(*File,*Coll,*Resc,*Flag);
>    writeLine("stdout","Created tar file *File for collection *Coll on resource *Resc");
>  }
> INPUT *File="/tempZone/home/rods/test/testcoll.tar", *Coll="/tempZone/home/rods/sub1", *Resc="demoResc", *Flag=""
> OUTPUT ruleExecOut

The example untar.r rule
> myTestRule {
>    msiTarFileExtract(*File,*Coll,*Resc,*Status);
>    writeLine("stdout","Extract files from a tar file *File into collection *Coll on resource *Resc");
>  }
> INPUT *File="/tempZone/home/rods/test/testcoll.tar", *Coll="/tempZone/home/rods/ruletest/fromtar", *Resc="demoResc"
> OUTPUT ruleExecOut

Here, creating the collection structure and putting the files in. Directory sub1 contains 11 separate 1GB text files:
> imkdir test
> iput -R demoResc -rb sub1
> irule -F tar.r

At this point, I have a successful tar file created. It looks good, even if I download it and manually untar it outside of iRODS.

However, if I am to start the untar rule, it will "succeed" without reporting failure, despite having unpacakged nothing:

> imkdir fromtar
> irule -F untar.r
> ils fromtar

I do not see any files. This has happened with a test of 15 1GB files, as well as with a single 13GB file

alanking:
This issue is fixed with the resolution of #4495. We will add a test for this issue (i.e. tar file) and close it.

PR #4499, opened 2019-09-09
[#4118] Add test for extracting large tar file (master)

alanking:
ibun -x of sufficiently large tar files needs a test to match the
test for extraction of sufficiently large zip files.

Also, fixed bug with "leftover" size for creating files in lib.make_file.


trel:
please to #

Issue #4492, opened 2019-08-28
irods::filesystem does not return all permissions for an object


korydraughn:
- [x] master
- [x] 4-2-stable

### iRODS Version, OS and Version
iRODS v4.2.6

### What did you try to do?
I tried to get the list of users and their permissions on a shared object.
```c++
namespace fs = irods::experimental::filesystem;
const auto s = fs::client::status(conn, "/path/to/shared_object");
const auto p = s.permissions();
```
`p` holds permissions for one user and the user is not known.

### Expected behavior
`p` should be a list of users and their permission on the object.

PR #4504, opened 2019-09-13
[#4492] Fixed irods::filesystem permissions handling (master)

korydraughn:


lib/filesystem/src/filesystem.cpp
```diff
@@ -395,7 +415,11 @@ namespace irods::experimental::filesystem::NAMESPACE_IMPL
         }
 
         create_collection(_comm, _p);
-        permissions(_comm, _p, s.permissions());
+
+        // TODO This needs to done using a batch operation to decrease the chance of failure.
```
trel: does this need to be addressed now?

trel:
if this is good - mark it with some #

korydraughn:
Added #

Issue #4521, opened 2019-09-23
add microservice to modify full AVU (including units)


d-w-moore:
- [x] master
- [x] 4-2-stable
----

## Feature

Set/Modify an AVU's 3rd field, Units, from rule code. 

### Description

Though iRODS clients are able to tag objects in the ICAT with "full-spec" AVU metadata records (meaning they can set all three fields, AttrName , Value, and Units), rule code is only able to query and read the Units field, not modify it. That is because the KeyValuePair container structures that function as a  source of input for the AVU manipulation microservices ( for example msiSetKeyValuePairsToObj and msiAssociateKeyValuePairsToObj)  give leverage for setting Attribute and Value only.

This means, among other things, that if an AVU of ["myAttr","myVal", "myUnit"] is set on a data object `afile.txt`, one cannot -- at least, not *directly* from rule code --  alter the Value field without erasing the Units field:
```   
$ touch afile.txt ; iput afile.txt; imeta set -d afile.txt myAttr myVal myUnit
$ imeta ls -d afile.txt 
AVUs defined for dataObj afile.txt:
attribute: myAttr
value: myVal
units: myUnit
$ irule "*x.myAttr='myVal_2'; msiSetKeyValuePairsToObj(*x,'$(ipwd)/afile.txt','-d')" null null
$ imeta ls -d afile.txt
AVUs defined for dataObj afile.txt:
attribute: myAttr
value: myVal_2
units: 
```
nor is it feasible to set AVU's that have Units != "" or to associate multiple AVU's to an object that might differ only in the Units field.

At the very least , a microservice could be added in with core iRODS microservices, one that thinly wraps the `rsModAVUMetadata` call and allows "add", "set", "rm", and "rmw" operations. (Possibly more? open to suggestions).

This would be more general, and only slightly more onerous to use, than:  [#4185 msiAddAVU]

trel:
If this is done, please checkbox and close.

PR #4523, opened 2019-09-26
[#4521] Add msvc for setting/modifying Units in AVU

d-w-moore:


server/re/src/extractAvuMS.cpp
```diff
@@ -661,3 +664,114 @@ msiRemoveKeyValuePairsFromObj( msParam_t *metadataParam, msParam_t* objParam,
...
+ *
+ * \DolVarDependence none
+ * \DolVarModified none
+ * \iCatAttrDependence none
+ * \iCatAttrModified AVU triples add/set/rm/rm-with-wildcard(rmw)
+ * \sideeffect none
+ *
+ * \return integer
+ * \retval 0 on success
+ * \retval USER_PARAM_TYP_ERROR wheninput  param dont match the type
```
trel: when input param doesn't match the type

Issue #4527, opened 2019-09-27
test_rulebase.py fails on call to lib.make_file(filename, '64M')


korydraughn:
- [x] master
- [x] 4-2-stable

## Bug Report
This test fails because the call to `lib.make_file(...)` is missing `contents='arbitrary`' .
```bash
======================================================================
ERROR [4.018s]: test_acDataDeletePolicy (irods.test.test_rulebase.Test_Resource_Session_Vars__3024)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/var/lib/irods/scripts/irods/test/test_rulebase.py", line 377, in setUp
    lib.make_file(self.large_file, '64M')
  File "/var/lib/irods/scripts/irods/lib.py", line 302, in make_file
    count = f_size / block_size_in_bytes
TypeError: unsupported operand type(s) for /: 'str' and 'int'
```

https://github.com/irods/irods/blob/73c2bacbf9e0539c5b3dccbcc40331213233459a/scripts/irods/lib.py#L193

PR #4528, opened 2019-09-27
[#4527] Added missing argument to function call (master)

korydraughn:


scripts/irods/test/test_rulebase.py
```diff
@@ -374,7 +374,7 @@ def setUp(self):
 
         # make large test file
         self.large_file = '/tmp/largefile'
-        lib.make_file(self.large_file, '64M')
+        lib.make_file(self.large_file, '64M', 'arbitrary')
```
alanking: This was due to a change I made in `lib.make_file` to properly use `dd` for generating large files. I changed the usage of the `f_size` parameter to be an int used in a calculation rather than passing the contents directly to `dd`. The solution presented here side-steps the problem by taking the 'arbitrary' codepath, which passes the argument directly to `truncate` as a string. So this is fine as a way of getting the tests to pass.

I do have a couple of suggestions for a more complete solution. We could modify `lib.make_file` to catch the `TypeError` being raised and revert to the old behavior (i.e. pass the argument directly to `dd` as a string). It could look something like this (see [existing implementation](https://github.com/irods/irods/blob/73c2bacbf9e0539c5b3dccbcc40331213233459a/scripts/irods/lib.py#L193-L209) for reference):
```

def make_file(f_name, f_size, contents='zero', block_size_in_bytes=1000):
    assert contents in ['arbitrary', 'random', 'zero']
    if contents == 'arbitrary' or f_size == 0:
        execute_command(['truncate', '-s', str(f_size), f_name])
        return

    source = {'zero': '/dev/zero',
              'random': '/dev/urandom'}[contents]

    try:
        count = f_size / block_size_in_bytes
        if count > 0:
            execute_command(['dd', 'if='+source, 'of='+f_name, 'count='+str(count), 'bs='+str(block_size_in_bytes)])
            leftover_size = f_size % block_size_in_bytes
            if leftover_size > 0:
                execute_command(['dd', 'if='+source, 'of='+f_name, 'count=1', 'bs='+str(leftover_size), 'oflag=append', 'conv=notrunc'])
            return
    except TypeError:
        pass
    execute_command(['dd', 'if='+source, 'of='+f_name, 'count=1', 'bs='+str(f_size)])
```

Alternatively, we could change the parameter name from `f_size` to `f_size_in_bytes` to make it more clear what we are expecting and just ensure all existing calls conform to this new requirement. This option may be clearer and will certainly be simpler to implement.

Sorry about this oversight...

trel:
definitely, add #

korydraughn:
Added #

scripts/irods/test/test_rulebase.py
trel: I vote for option B.   clear is good.  less magic.

korydraughn: With option B (f_size_in_bytes), we remove the `contents` parameter and fix all tests using this function? Everything goes uses `dd` at that point, right?

alanking: No, I think we'd want to keep the ability to control the contents of the file. This would just be imposing a type requirement on the user of `lib.make_file()`.

I wonder if that's an anti-pattern in Python...

alanking: Actually, the easiest thing to do would probably be to change the default of `contents` to 'arbitrary'. That way, we wouldn't have to touch all the tests.

korydraughn: I'll make a new issue to clean this up.

PR #4539, opened 2019-09-30
[4537] Updated the message on the irods::error constructor if a new m…

JustinKyleJames:
…essage is passed in.

### DIFF
--- a/lib/core/src/irods_error.cpp
+++ b/lib/core/src/irods_error.cpp
@@ -136,6 +136,11 @@ const char *error::status_token_     = " status [";
         std::string  _fcn,
         const error& _rhs ) :
         error(_rhs) {
+ 
+        if (!_msg.empty()) {
+            message_ = _msg;
+        }
+
         if (exception_) {
             exception_->add_message(_msg + ": " + build_result_string( _file, _line, _fcn ));
             return;


lib/core/src/irods_error.cpp
```diff
@@ -136,6 +136,11 @@ const char *error::status_token_     = " status [";
         std::string  _fcn,
         const error& _rhs ) :
         error(_rhs) {
+ 
+        if (_msg != "") {
```
korydraughn: Is this statement checking for emptiness? If so, the **empty** member function should be used. 
```
if (!_msg.empty()) {
    ...
}
```

lib/core/src/irods_error.cpp
```diff
@@ -136,6 +136,11 @@ const char *error::status_token_     = " status [";
...
+            message_ = _msg;
```
korydraughn: You can possibly use `std::move` here given that `_msg` arrives as a copy.

JustinKyleJames: Could not use std::move since _msg is being used below this point.

trel:
looks good - please #.

JustinKyleJames:
\# added

Issue #4551, opened 2019-10-10
irods::filesystem::path::parent_path() is not compliant with the C++17 standard


korydraughn:
- [x] master
- [x] 4-2-stable

### iRODS Version, OS and Version
iRODS v4.2.6

### What did you try to do?
Write a unit test that verifies that the iRODS filesystem implementation produces the same results as the C++17 standard implementation.
```c++
namespace fs = irods::experimental::filesystem;
REQUIRE(fs::path{"/"}.parent_path() == "/");
```

### Expected behavior
The expression within the `REQUIRE` macro should evaluate to **true**.

### Observed behavior (including steps to reproduce, if applicable)
The expression fails because the current implementation returns an empty string (just like  boost::filesystem).

PR #4552, opened 2019-10-10
[#4551] irods::filesystem::path::parent_path() is compliant with the standard (master)

korydraughn:
Passed core tests in CI.


trel:
passes?  add the #

korydraughn:
Added #

Issue #4566, opened 2019-10-29
Connection pool should allow ownership of connections to be transferred


korydraughn:
- [x] master
- [x] 4-2-stable

## Feature
This feature would be very similar to `std::unique_ptr`.
```c++
auto smart_ptr = std::make_unique<int>(100);
auto* p = smart_ptr.release(); // The smart pointer no longer manages the memory.
```
Once ownership is transferred, the connection pool will create a new connection (in the slot previously released) on the next request.

PR #4567, opened 2019-10-29
[#4566] Connections can now be released from a connection pool (master)

korydraughn:
Core tests passed in CI.


korydraughn:
Added #

Issue #4570, opened 2019-10-30
irods::filesystem::status throws exception during "ils" federation testing


korydraughn:
- [x] master
- [x] 4-2-stable

## Bug Report
Federation tests that invoke `ils` on a collection in a remote zone fail because`irods::filesystem::status` does not set the **zone hint** on the specific query. 
https://github.com/irods/irods/blob/9ae8a022f7a8cc815c3c5b71ca4a7771effd8508/lib/filesystem/src/filesystem.cpp#L148-L154

PR #4571, opened 2019-10-30
[#4570] Zone hint is now set in irods::experimental::filesystem::status (master)

korydraughn:
Passed Core Tests.


korydraughn:
Added #

Issue #4574, opened 2019-11-04
Implement builder class to simplify construction of irods::query objects


korydraughn:
- [x] master
- [x] 4-2-stable

Issue #4573, opened 2019-11-01
irods::query needs to allow setting a zone hint


korydraughn:
- [x] master
- [x] 4-2-stable

## Bug Report
The zone hint is required to allow iRODS query to work with federated zones.

## Solution
Add a new parameter for specifying the zone hint.

Issue #4493, opened 2019-08-28
irods::query does not allow binding arguments to specific queries that require them


korydraughn:
- [x] master
- [x] 4-2-stable
-------
## Bug Report
The current interface does not give the developer a way to pass arguments to the query.
```c++
using query_type = irods::query<rcComm_t>::query_type;
const std::string specific_query_alias = "ShowCollAcls";
irods::query q{conn, specific_query_alias, query_type::SPECIFIC};
```

### Possible Solutions
- Pass a vector or tuple of values to the constructor.
- Provide a member function for binding values to specific parameters.

PR #4576, opened 2019-11-05
[#4493][#4573][#4574] irods::query now accepts a zone hint (4-2-stable)

korydraughn:
Passed Core Tests.


trel:
i like this very much.    please add the #.

PR #4581, opened 2019-11-08
Slow rebalance with iRODS 4.2.4 and PostgreSQL 11.1

jthiltges:
The following query has been running for over 15 days on a system with ~8M records and two replicas.

```select distinct data_id from R_DATA_MAIN where data_id in (select data_id from R_DATA_MAIN where resc_id in (10016)) and data_id not in (select data_id from R_DATA_MAIN where resc_id in (10014)) and modify_ts <= '01571939628' limit 500```

Rewriting the query to use `EXCEPT` for the set math seems much faster, returning results in about 70 seconds. The change is implemented in this PR.

This PR is submitted as draft, as I have not tested iRODS with this change. I'd appreciate a review from someone who understands the spirit of the query.

### DIFF
--- a/plugins/database/src/db_plugin.cpp
+++ b/plugins/database/src/db_plugin.cpp
@@ -13587,8 +13587,23 @@ irods::error db_get_repl_list_for_leaf_bundles_op(
 
 #ifdef ORA_ICAT
     const std::string query = (boost::format("select data_id from (select distinct data_id from R_DATA_MAIN where data_id in (select data_id from R_DATA_MAIN where resc_id in (%s)) and data_id not in (select data_id from R_DATA_MAIN where resc_id in (%s)) and modify_ts <= '%s') where rownum <= %d") % not_child_array % child_array % _invocation_timestamp->c_str() % _count).str();
+#elif MY_ICAT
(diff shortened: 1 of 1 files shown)


trel:
This is great - will look a bit closer in the next few days.

Also note... MySQL doesn't implement `EXCEPT` - but can use a `LEFT JOIN` to simulate the `MINUS`/`EXCEPT` notation.

http://www.mysqltutorial.org/mysql-minus/

jthiltges:
Reopening to retry travis-ci build

alanking:
I have run core tests on this change and all tests passed as expected.

anderbubble:
@trel what is standing in the way of this getting merged? Can it be in the next release?

trel:
Yes, we can get it included if it passes all three database suites in CI.

Please squash these three commits down to one and update the commit message to begin with our bracketed notation [#xxxx] to link with issue 4731.

jthiltges:
Commits are squashed and rebased against master. Thank you for taking a look at it.

trel:
running through CI again, will remove draft status, and get it merged - thanks!

Issue #4430, opened 2019-06-19
Refactor delay server as an irods::query_processor


alanking:
- [x] 4-2-stable
- [x] master
---
The delay server is basically an implementation of the new `irods::query_processor`, so it could be refactored to use this for executing delayed rules. The result would be that the delay server would not need to manage its own thread/connection pools and the code would be reduced to the query for fetching delayed rules to be executed and the instructions for executing them.

PR #4596, opened 2019-11-26
[#4430] Implement delay server as query processor (master)

alanking:


server/core/src/irodsReServer.cpp
```diff
@@ -337,63 +373,37 @@ int main() {
...
-                    }
-                    logger::delay_server::trace((boost::format("Enqueueing rule [%s]") % rule_id).str());
-                    queue.enqueue_rule(rule_id);
-                    irods::thread_pool::post(thread_pool, [conn_pool, &queue, result] {
-                        execute_rule(conn_pool, queue, result, re_server_terminated);
-                    });
+            auto future = delay_queue_processor.execute(thread_pool, static_cast<rcComm_t&>(query_conn));
+            auto errors = future.get();
+            if(errors.size() > 0) {
+                for(const auto [code, msg] : errors) {
```
korydraughn: You can add a `&` to avoid copying (e.g. `for (const auto& [code, msg] : errors)`).

alanking: Got it, thanks!

alanking:
These changes should also address #4616 as the worker connection pool is now instantiated outside the main loop.

lib/core/include/connection_pool.hpp
```diff
@@ -57,6 +57,8 @@ namespace irods
 
         connection_proxy get_connection();
 
+        static std::shared_ptr<connection_pool> make_one();
```
korydraughn: This should probably be moved outside of the class as it doesn't depend on anything within it. This would allow the name to be more descriptive as well as accept an argument for specifying the size of the pool.
```c++
std::shared_ptr<connection_pool> make_connection_pool(int size = 1);
```

alanking: I like it

alanking:
Tests are passing as expected.

trel:
please to #

alanking:
Added #

Issue #4617, opened 2019-12-03
Query iterator throws on a specific query resulting in CAT_NO_ROWS_FOUND


korydraughn:
- [x] master
- [x] 4-2-stable
---

### iRODS Version, OS and Version
4.2.6

### What did you try to do?
Tried to build a specific query that returns nothing.

### Expected behavior
We expected it to quietly carry on.

### Observed behavior (including steps to reproduce, if applicable)
A stacktrace was produced.

PR #4618, opened 2019-12-03
[#4617] Query iterator no longer throws on CAT_NO_ROWS_FOUND (master)

korydraughn:


trel:
please to # when ready

Issue #4636, opened 2020-01-08
Crash while listing files with ils after upgrade to 4.2.7 version of the iRODS packages


batkiewicz:
- [x] master
- [x] 4-2-stable
-----

### iRODS Version, OS and Version
iRODS version 4.2.7
OS: Ubuntu 16.04 (Problem observed also on CentOS 7)

### What did you try to do?
I am using irods packages to access KEK iRODS.
Everything is working fine for version 4.2.6.
After upgrade to 4.2.7 I am able to initialize the session,
but command ils causes the crash.
The error disappeared after downgrading irods-runtime and irods-icommands to version 4.2.6.

### Observed behavior (including steps to reproduce, if applicable)

After creating iRODS session (iinit command, everthing is fine at this stage),
I write in termianl ils command and get the following error:
```
terminating with uncaught exception of type irods::exception: iRODS Exception:
    file: /irods/lib/core/include/irods_query.hpp
    function: irods::query<rcComm_t>::spec_query_impl::spec_query_impl(connection_type *, int, int, const std::string &, const std::string &, const std::vector<std::string> *) [connection_type = rcComm_t]
    line: 287
    code: -853000 (CAT_UNKNOWN_SPECIFIC_QUERY)
    message:
        query fill failed for [ShowCollAcls]
stack trace:
--------------
Dumping stack trace
<0>	Offset: 0x65 	Address: 0x7fe3a25589e5	irods::exception::exception(long, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, unsigned int, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&)
<1>	Offset: 0x43 	Address: 0x7fe3a2558b63	irods::exception::exception(long, boost::basic_format<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, unsigned int, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&)
[...]

PR #4638, opened 2020-01-11
[#4636] GenQuery fallback for retrieving collection permissions (master)

korydraughn:


lib/filesystem/src/filesystem.cpp
```diff
@@ -160,17 +160,31 @@ namespace irods::experimental::filesystem::NAMESPACE_IMPL
...
+                    std::vector<std::string> args{_p.string()};
+
+                    qb.type(irods::experimental::query_type::specific)
+                      .bind_arguments(args);
+
+                    for (const auto& row : qb.build(_comm, "ShowCollAcls")) {
+                        _s.prms.push_back({row[0], to_permission_enum(row[2])});
+                    }
+                }
+                catch (...) {
```
trel: can we / do we want to avoid the (...)?  a more specific exception to catch?

korydraughn: I thought about that and ultimately decided it wasn't needed. The current fsys impl does not log anything. In nearly every case, it throws an exception for the user to handle. With that, I didn't see any benefit to specifying a type if it was not going to be used.

trel: cool.   maybe add a comment to save that thinking/why.

# away.

korydraughn: Sure. After reviewing this a bit more, I don't understand why ils succeeds with this change. I understand the fsys fix, but I would expect ils to still fail given that iquest fails to link tables when executing the same genquery string used by the following function:  https://github.com/irods/irods/blob/0a31d045e43396e198b7a127419dc30d60df3151/lib/core/src/lsUtil.cpp#L666

korydraughn: It turns out that `queryCollAcl` uses columns that are not mapped. Because of that, `fillGenQueryInpFromStrCond` cannot find the required columns when attempting to generate the correct SQL.

Created issues and I also have fixes for this included in this PR as of now.

trel: very nice - remarkable that those columns weren't mapped.    they now appear in `iquest attrs` and hadn't before?

trel: confirming they didn't in 4.2.7...

```
$ imiscsvrinfo | grep relVersion
relVersion=rods4.2.7
$ iquest attrs | grep COLL_USER_NAME
$ echo $?
1
$ iquest attrs | grep COLL_ZONE_NAME
$ echo $?
1
```

korydraughn: Yes, they do appear in `iquest attrs`.

korydraughn:
Added #

Issue #4649, opened 2020-01-16
Server leaks memory via (python-)rule genqueries and built-in microservices


cjsmeele:
- [x] master
- [x] 4-2-stable
------

## Bug Report

Memory allocated via genquery and certain microservices is not freed, when called through the Python rule engine. Equivalent operations from the iRODS rule language appear either not to leak or to leak less memory.

This makes it difficult to have long-running Python rules, or to use a single iRODS session for multiple (Python) rule calls.

### iRODS Version, OS and Version

iRODS 4.2.7, CentOS 7

### What did you try to do?

Call a long-running Python rule containing various queries or MSI calls, or alternatively call multiple Python rules using a single iRODS connection (via N `rcExecMyRule` API calls).

### Expected behavior

`irodsServer` memory consumption remains near-constant across rule calls. Completing e.g. a python-rule-initiated query frees related resources (either immediately or at some memory usage threshold).

### Observed behavior (including steps to reproduce, if applicable)

`irodsServer` memory consumption increases after MSI calls and does not decrease. Completion of the rcExecMyRule API call does not free allocated server resources either: repeatedly calling (Python) rules using the same iRODS session eventually results in an OOM-kill for the server process handling those requests.

### Reproducing

For example, the following two rules are functionally identical: They both execute 1000 genqueries. However, by checking irodsServer memory consumption (`watch -n.1 'ps -o pid,start,rss,args --sort start_time,-rss -C irodsServer'`) the Python variant can be seen to cause memory usage to increase steadily over time (~3MB/s). In contrast, the memory usage of the iRODS rule language variant appears to remain constant (and lower).

```python
import time
import genquery
def dummy(rule_args, callback, rei):
    for i in range(1000):
        # Some query that has 0 results.
        for x in genquery.row_iterator('COLL_NAME', "COLL_NAME = 'unicorn'", genquery.AS_LIST, callback):
            pass
        time.sleep(0.01)
```

```
dummyR() {
    for (*i = 0; *i < 1000; *i = *i + 1) {
        foreach (*row in SELECT COLL_NAME WHERE COLL_NAME = 'unicorn') { nop; }
        msiSleep('0', '10000');
    }
}
```

Calling the same Python rule multiple times sequentially (via multiple `rcExecMyRule` API calls) multiplies memory consumption. Only closing the connection releases the allocated resources.

Another example:
A data object is repeatedly opened and closed. The close is expected to free the resources related to the handle.

```python
[...]

PR #4654, opened 2020-01-22
[#4649] Use fillMsParam in convertToMsParam (4-2-stable)

alanking:
...instead of replMsParam. Fixes memory leak in conversions
made in rule engine plugins.

---

CI tests to be run.


alanking:
This change will be part of a larger effort to fix the remaining leaks. Closing this PR.

PR #4709, opened 2020-02-06
[4708] use capital table name to be database agnostic

jassigill2000:


korydraughn:
Looks good to me.

trel:
agreed - good catch.

Issue #4712, opened 2020-02-10
lifetime_manager unit-test segfaults


korydraughn:
- [x] master
- [x] 4-2-stable
---
## Bug Report
The segfault occurs because the `test_class::str_` member variable is not initialized to a `nullptr` during the construction of the following array.
https://github.com/irods/irods/blob/e563c57c6813463aa731fecd04e368da5a23b138/unit_tests/src/test_irods_lifetime_manager.cpp#L130

PR #4714, opened 2020-02-10
[#4712] Added user-defined default constructor to fix segfault in lifetime_manager unit-test (4-2-stable)

korydraughn:


korydraughn:
Added #

Issue #4506, opened 2019-09-13
Add iunreg iCommand


alanking:
- [x] master
- [x] 4-2-stable
---
## Feature

Related to #3821 and #3451 (commit https://github.com/irods/irods/commit/344426eca96285984bbd01ff06f7e5d73fc02f09).

With the deprecation of `irm -n`, `irm -U -n` is no longer recommended. We need a way to trim a specific replica from the catalog without touching physical storage (aka unregister).

Expected example usage with `-n` (`REPL_NUM_KW`):
```
$ ils -l foo
$ ils -l
/tempZone/home/rods:
  rods              0 demoResc        6 2019-09-12.14:03 & foo
  rods              1 ufs1            6 2019-09-12.14:03 & foo
  rods              2 ufs2            6 2019-09-12.14:03 & foo
$ itrim -U -n1 foo
$ ils -l
/tempZone/home/rods:
  rods              0 demoResc        6 2019-09-12.14:03 & foo
  rods              2 ufs2            6 2019-09-12.14:03 & foo
```

Expected example usage with `-S` (`RESC_NAME_KW`) where all replicas under the named source resource are eligible for trimming:
```
$ ils -l foo
$ ils -l
/tempZone/home/rods:
  rods              0 demoResc        6 2019-09-12.14:03 & foo
  rods              1 repl;ufs1       6 2019-09-12.14:03 & foo
  rods              2 repl;ufs2       6 2019-09-12.14:03 & foo
$ itrim -U -N1 -S repl foo
$ ils -l
/tempZone/home/rods:
  rods              0 demoResc        6 2019-09-12.14:03 & foo
```

Expected example usage with `-S` (`RESC_NAME_KW`) where some - but not all - replicas under the named source resource are eligible for trimming:
```
$ ils -l foo
$ ils -l
/tempZone/home/rods:
  rods              0 demoResc        6 2019-09-12.14:03 & foo
  rods              1 repl;ufs1       6 2019-09-12.14:03 & foo
  rods              2 repl;ufs2       6 2019-09-12.14:03 & foo
$ itrim -U -S repl foo
$ ils -l
/tempZone/home/rods:
  rods              0 demoResc        6 2019-09-12.14:03 & foo
  rods              2 repl;ufs2       6 2019-09-12.14:03 & foo
```
Even when the target resource is marked down, the eligible replica(s) should be unregistered without complaint.

tsmeele:
Ideally this command option to unregister is authorized similar to ireg:
```
[...]

PR #4730, opened 2020-02-14
[#4506] Add support for iunreg + fix tests checking logs (4-2-stable)

alanking:
CI tests are running

Requires irods/irods_client_icommands#114


alanking:
#s added. Alleviates some of the test failures, but log flushing continues to be an ongoing issue.

Issue #4410, opened 2019-05-25
Unauthorized iput over existing data object across the network creates orphan file


alanking:
- [x] master
- [x] 4-2-stable
---
### iRODS Version, OS and Version
4.2.5

### Observed behavior (including steps to reproduce, if applicable)
Some notes (possibly unrelated):
 - `resource1.example.org` is a catalog consumer
 - `icat.example.org` is a catalog provider
 - `demoResc` resides on `icat.example.org`
 - This issue was originally discovered when looking into `test_iticket` failing for files which trigger parallel transfer
```
irods@resource1:~$ ienv | grep irods_version && ilsresc -l demoResc | grep location
irods_version - 4.2.5
location: icat.example.org
irods@resource1:~$ iput -R demoResc VERSION.json foo
irods@resource1:~$ ils -L foo
rods              0 demoResc          224 2019-05-25.14:59 & foo
generic    /var/lib/irods/Vault/home/rods/foo
root@icat:/# ls -al /var/lib/irods/Vault/home/rods/foo
-rw------- 1 irods irods 224 May 25 14:59 /var/lib/irods/Vault/home/rods/foo
# switch to regular rodsuser and iput to a place you have no permissions using ticket
root@resource1:/# ienv | grep user
irods_user_name - mob
root@resource1:/# iput -R demoResc -f 1G-file.bin /tempZone/home/rods/foo
remote addresses: 127.0.0.1 ERROR: putUtil: put error for /tempZone/home/rods/foo, status = -24000 status = -24000 SYS_INTERNAL_NULL_INPUT_ERR
# After an unhelpful error message... the catalog seems fine
irods@resource1:~$ ils -L foo
rods              0 demoResc          224 2019-05-25.14:59 & foo
generic    /var/lib/irods/Vault/home/rods/foo
# ...but the data is gone
root@icat:/# ls -al /var/lib/irods/Vault/home/rods/foo
ls: cannot access '/var/lib/irods/Vault/home/rods/foo': No such file or directory
```
Cannot reproduce when client is local to the machine on which the resource resides.

alanking:
Contrary to the last line in the original issue comment, the target resource can be on the server to which the client is connected and hit this issue *provided the server is a catalog consumer*. Client connected to the catalog provider did not demonstrate this behavior.

alanking:
Likely will be addressed as part of #4343 work.

alanking:
[...]

PR #4747, opened 2020-02-26
[#4410] Attempt registration before moving data (4-2-stable)

alanking:
rsDataObjPut and rsDataObjCreate now register data objects before
touching the disk. This prevents situations in which existing data
objects being overwritten by unauthorized users would have their
associated files unlinked from the data object (orphan).

Adds a test to put to a resource with a vault path for which iRODS
has no write access. Modifies ticket put/get tests to trigger parallel
transfer.

---

CI tests passed. More tests to come later as this may fix additional issues, so hold off on merging.


alanking:
Tests passed, #'s added.

Issue #4740, opened 2020-02-21
Add ability to modify any R_DATA_MAIN column with 'iadmin modrepl'


alanking:
- [x] master
- [x] 4-2-stable
---
## Feature

`isysmeta` should be able to modify any column for a given row of R_DATA_MAIN. As such, the command would require the following to be provided:

- subcommand (mod, ls, etc.)
- path to data object (or data id?)
- replica number or hierarchy/leaf resource
For mod:
- column name (e.g. DATA_STATUS)
- the value to inject

The motivation for this feature is the need for the ability for administrators to modify replica status in the event that intermediate replicas (#4343) cannot mark themselves as good or stale on close. #4331 should handle this in the future, but a direct modification of the catalog could be a helpful tool regardless.

trel:
I'd suggest that maybe there should be no default behavior (aka replica 0).

Force the admin to know/mean what they're doing.

alanking:
Agreed. I've adjusted the description to remove the suggestion of a default value for replica number and separated the requirement of column name and value for the mod subcommand only. Thanks!

alanking:
`isysmeta` uses `rcModDataObjMeta` which does not require admin privileges. As such we are tweaking this slightly to be a new `iadmin` subcommand called `modrepl`. Usage:
```
iadmin modrepl logical_path [ --replica_number, -n | --resource, -R ] ATTR_NAME VALUE
```
The replica number or resource hierarchy will be required in order to specify a row to modify. ATTR_NAME will be the GenQuery attribute to modify (e.g. `DATA_REPL_STATUS` -- maps to column `data_is_dirty`) and VALUE will be the value to insert in that column. The `ADMIN_KW` will be used to enforce permissions

trel:
Related work will involved updating 'iadmin ls' to present the current information / GenQuery attributes for a particular logical_path / replica.

Together, these will provide tools for both inspection and modification of replica information to an administrator.

PR #4749, opened 2020-03-02
[#4740] Add support for iadmin modrepl (master)

alanking:
Some cleanup in the mod_data_obj_meta database op.

---

CI tests running

### DIFF
--- a/plugins/database/src/db_plugin.cpp
+++ b/plugins/database/src/db_plugin.cpp
@@ -43,6 +43,7 @@
 // stl includes
 #include <sstream>
 #include <string>
+#include <string_view>
 #include <iostream>
 #include <vector>
 #include <boost/regex.hpp>
@@ -2081,7 +2082,7 @@ irods::error db_mod_data_obj_meta_op(
 //        _ctx.prop_map().get< icatSessionStruct >( ICSS_PROP, icss );
 
 
-    int i = 0, j = 0, status = 0, upCols = 0;
+    int status = 0, upCols = 0;
     rodsLong_t iVal = 0; // JMC cppcheck - uninit var
     int status2 = 0;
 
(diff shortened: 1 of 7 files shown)


plugins/database/src/db_plugin.cpp
```diff
@@ -2141,23 +2163,26 @@ irods::error db_mod_data_obj_meta_op(
...
-            if( std::string( "resc_name") == colNames[i]) {
+    std::size_t i = 0, j = 0;
+    for (i = 0, j = 0; i < regParamNames.size(); i++) {
+        theVal = getValByKey(_reg_param, regParamNames[i].data());
+        if (theVal) {
+            if(std::string( "resc_name") == colNames[i]) {
                 continue;
             }
-            else if( std::string( "resc_hier") == colNames[i]) {
+            else if(std::string( "resc_hier") == colNames[i]) {
```
korydraughn: the `std::string` wrapping `"resc_hier"` is probably not needed given `colNames[i]` is a `std::string_view`.

alanking: Good catch - will fix

alanking:
CI tests passed.

server/core/src/irodsReServer.cpp
```diff
@@ -244,15 +244,7 @@ namespace {
...
-                rodsEnv env{};
-                _getRodsEnv(env);
-                auto tmp_pool = std::make_shared<irods::connection_pool>(
-                    1,
-                    env.rodsHost,
-                    env.rodsPort,
-                    env.rodsUserName,
-                    env.rodsZone,
-                    env.irodsConnectionPoolRefreshTime);
+                auto tmp_pool = irods::make_connection_pool();
```
trel: should this be a separate commit/issue?

trel:
looks good otherwise, # when ready.

server/core/src/irodsReServer.cpp
alanking: Yup, probably. Removed for later fixing.

Issue #4735, opened 2020-02-18
Reinstate connection pooling in the delay server


alanking:
## Feature

Due to the issue described in #4616, the delay server will not be using a proper connection pool for the time being. A single connection will be created for each result in the queue up to the configured number of concurrent RE processes.

The attempted pooling implementation leads to long-held connections not updating appropriately when changes to configured rule bases are made (see #2585). Add some implementation-defined mechanism for the connection pool that detects a need for refreshing the connection at the time it is used.

PR #4754, opened 2020-03-03
[#4735] Use free function to create single-conn pool (master)

alanking:
Passed CI tests as part of #4749 and removed due to being a separate issue.


trel:
looks good - please #

alanking:
#'d

Issue #4650, opened 2020-01-16
Add library that enables management of users and groups


korydraughn:
- [x] master
- [x] 4-2-stable
---
## Feature
This library should enable the following:
- Creation/Removal of users and groups
- Modifying user information
  - Password
  - User Type
- Adding/Removing users to/from groups
- Querying user/group information
  - Is some user a member of some group?
  - Does the user or group exist?

PR #4776, opened 2020-03-08
[#4650] User Administration Library (4-2-stable)

korydraughn:
- Added new specific query for listing all groups a user is a member of.
- Unit-testing of remote operations is disabled by default. To enable,
  define the macro "IRODS_TEST_REMOTE_USER_ADMINISTRATION_OPERATIONS".

### DIFF
--- a/lib/core/include/user_administration.hpp
+++ b/lib/core/include/user_administration.hpp
@@ -1,0 +1,113 @@
+#ifndef IRODS_USER_ADMINISTRATION_HPP
+#define IRODS_USER_ADMINISTRATION_HPP
+
+#undef NAMESPACE_IMPL
+#undef rxComm
+
+#ifdef IRODS_USER_ADMINISTRATION_ENABLE_SERVER_SIDE_API
+    #define NAMESPACE_IMPL      server
+    #define rxComm              rsComm_t
+#else 
+    #define NAMESPACE_IMPL      client
+    #define rxComm              rcComm_t
+#endif // IRODS_USER_ADMINISTRATION_ENABLE_SERVER_SIDE_API
+
+#include "rcConnect.h"
+#include "user.hpp"
+#include "group.hpp"
+
+#include <vector>
+#include <string>
+#include <optional>
+#include <ostream>
+#include <system_error>
+#include <stdexcept>
+
+namespace irods::experimental::administration
+{
+    inline namespace v1
+    {
+        // Types
+
+        enum class user_type
+        {
+            rodsuser,
+            groupadmin,
+            rodsadmin
+        }; // user_type
+
+        enum class zone_type
+        {
+            local,
+            remote
+        }; // zone_type
+
+        // Exceptions
+
+        class user_management_error
+            : std::runtime_error
+        {
+        public:
+            using std::runtime_error::runtime_error;
+        };
+    } // namespace v1
+} // irods::experimental::administration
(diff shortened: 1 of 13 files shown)


lib/core/include/user_administration.hpp
```diff
@@ -0,0 +1,113 @@
...
+namespace irods::experimental::administration::NAMESPACE_IMPL
+{
+    inline namespace v1
+    {
+        // User Management
+
+        auto make_user(rxComm& conn,
+                       const user& user,
+                       user_type user_type = user_type::rodsuser,
+                       zone_type zone_type = zone_type::local) -> std::error_code;
```
korydraughn: iRODS has `iadmin mkuser` and `msiCreateUser`. C++ tends to use `make_*` for convenience functions (e.g. `std::make_unique`, `std::make_shared`) and iRODS has started to adopt this scheme as well (e.g. `irods::make_connection_pool`).

I feel like the **make** prefix is a little confusing here. This function **adds** a user to iRODS. When I read `make_user`, my first thought would be I'm instantiating a user object, not adding a new user to iRODS. With that, it seems clearer to me to change **make** to **add**.

This fits in line with Linux's `adduser` and `deluser` commands too. Also, add and remove are a more common pair for operations such as these.

korydraughn: I feel `iadmin mkuser` should probably change to use `adduser` as well given this thinking.

lib/core/include/user_administration.hpp
```diff
@@ -0,0 +1,113 @@
...
+
+        auto set_user_type(rxComm& conn, const user& user, user_type new_user_type) -> std::error_code;
+
+        auto add_user_auth(rxComm& conn, const user& user, std::string_view auth) -> std::error_code;
+
+        auto remove_user_auth(rxComm& conn, const user& user, std::string_view auth) -> std::error_code;
+
+        // Group Management
+
+        auto make_group(rxComm& conn, const group& group) -> std::error_code;
```
korydraughn: See comment about `make_user`.

lib/core/include/user_administration.hpp
trel: We're not going to change the admin-facing interface (or the `msiCreateUser` microservice name) until we create the one true `irods` command-line tool, but we could add an alias subcommand `iadmin adduser` -> `iadmin mkuser`...

Internal function calls should be renamed to be consistent, yes.

korydraughn: Do we want to change the function referenced to `add_user` or keep it how it is?

The aliases for iadmin and msiCreateUser can wait since they aren't urgent.

trel: i don't have a strong feeling either way... but `add_user` to me feels like adding a user to a group...  will there be a `make_user` function that does something else if we rename this to `add_user`?

korydraughn: I can't say there won't ever be a make_user function. make_* functions are normally there to ease construction of something. I don't see that happening with the user objects given they are very tiny objects with very simple constructors. There is also a function for adding users to groups, `add_user_to_group` so there shouldn't be any confusion about what is happening.

trel: go for it.

korydraughn:
unit-tests pass.  Added #

PR #4798, opened 2020-03-12
timing tests separated into their own suite (master)

alanking:


alanking:
Added notes about cherry-pick origins. Codacy can be ignored as we are using subprocess module in other places and this is simply lifting existing tests into a different file.

trel:
these seem okay to add # ?

Issue #3548, opened 2017-05-16
deprecate num_repl keyword for replication resource


trel:
#2851 introduced `num_repl=N` for randomly selecting N children to receive replicas under a replication resource.

We need to confirm/fix the behavior when `iadmin modresc rebalance` is run against a replication resource with `num_repl` defined.

trel:
This keyword breaks the policy encoded in the replication resource.  Marking as deprecated - will remove in 4.3.0.

PR #4801, opened 2020-03-13
[#3548] Deprecate num_repl keyword (4-2-stable)

alanking:
CI tests running

### DIFF
--- a/plugins/resources/replication/librepl.cpp
+++ b/plugins/resources/replication/librepl.cpp
@@ -1631,13 +1631,21 @@ class repl_resource : public irods::resource {
                 }
 
                 if ( kvp_map.find( NUM_REPL_KW ) != kvp_map.end() ) {
+                    irods::log(LOG_WARNING,
+                               (boost::format(
+                               "[%s] - resource [%s] is using deprecated context string [%s]") %
+                               __FUNCTION__ %
+                               _inst_name %
+                               NUM_REPL_KW).str());
                     try {
                         int num_repl = boost::lexical_cast< int >( kvp_map[ NUM_REPL_KW ] );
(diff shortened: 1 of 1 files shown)


plugins/resources/replication/librepl.cpp
```diff
@@ -1641,6 +1641,9 @@ class repl_resource : public irods::resource {
                         }
                         else {
                             properties_.set< size_t >( NUM_REPL_KW, static_cast< size_t >( num_repl ) );
+                            irods::log(LOG_NOTICE,
```
trel: LOG_WARNING ?    are our other deprecation messages LOG_NOTICE?

alanking: unixfilesystem resource is the only one which logs deprecation messages and it is using LOG_NOTICE:

https://github.com/irods/irods/blob/d91ec25e619e159c5b9416c7fed2408d31dbf833/plugins/resources/unixfilesystem/libunixfilesystem.cpp#L464-L467

I was also wondering if the message should be logged just here (standing up the resource plugins), or also at the time that the policy is applied to the child list (each time a create is performed in this resource).

alanking: It may be worth the format call to specify which resource this is, too.

trel: definitely worth the time to format a helpful message.   and if production systems are set to WARNING, we still want them to hear about this.

at time of use is fine rather than creation - it will already be in their logs as soon as they use it.  and we don't parse things for meaning at creation/edit of the context string.

trel:
looks pretty good to me.

alanking:
CI tests pass

PR #4837, opened 2020-04-02
New library and bug fixes (master)

korydraughn:


scripts/irods/test/test_rule_engine_plugin_framework.py
```diff
@@ -282,6 +282,71 @@ def test_repf_supports_skipping_operations_and_nrep_returns_positive_error_codes
...
+                lib.make_file(filename, 1, 'arbitrary')
+
+                # Capture the current size of the log file. This will be used as the starting
+                # point for searching the log file for a particular string.
+                log_offset = lib.get_file_size_by_path(paths.server_log_path())
+
+                self.admin.assert_icommand_fail(['iput', filename])
+
+                msg = 'RULE_ENGINE_SKIP_OPERATION (5001000) incorrectly returned from PEP [pep_api_data_obj_put_finally]'
+                lib.delayAssert(lambda: lib.log_message_occurrences_greater_than_count(msg=msg, count=0, start_index=log_offset))
```
korydraughn, line 348: This could be refactored into a loop to avoid duplicate code.

server/api/src/rsDataObjUnlink.cpp
```diff
@@ -34,12 +34,58 @@
...
+        const fs::path parent = vault_path;
+
+        auto p_iter = std::begin(parent);
+        auto p_last = std::end(parent);
+        auto c_iter = std::begin(child);
+        auto c_last = std::end(child);
+
+        for (; p_iter != p_last && c_iter != c_last && *p_iter == *c_iter; ++p_iter, ++c_iter);
+
+        return {0, p_iter == p_last};
```
korydraughn, line 85: This code exists in the filesystem implementation and should me externalized to a separate function (i.e. is_prefix_of).

server/core/include/scoped_client_identity.hpp
```diff
@@ -0,0 +1,56 @@
...
+    {
+    public:
+        /// Constructs a scoped_client_identity object and sets the username of
+        /// \b comm to \b new_username.
+        ///
+        /// \param[in] comm         The server communication object.
+        /// \param[in] new_username The name of the iRODS user to become.
+        scoped_client_identity(rsComm_t& comm, std::string_view new_username)
+            : conn_{comm}
+            , username_{conn_.clientUser.userName}
```
korydraughn, line 31: If an exception is thrown from the constructor, then we lose cycles by doing this step. However, the cost of doing this is likely very small and the implementation is correct as is.

server/api/src/rsDataObjUnlink.cpp
```diff
@@ -34,12 +34,58 @@
...
+    }
```
alanking, line 86: Just making a note that server/core/include/physPath.hpp define `isInVault` which ostensibly performs the same task. This seems like a nice replacement. Perhaps a future improvement.

scripts/irods/test/test_rule_engine_plugin_framework.py
```diff
@@ -282,6 +282,71 @@ def test_repf_supports_skipping_operations_and_nrep_returns_positive_error_codes
...
+                lib.make_file(filename, 1, 'arbitrary')
+
+                # Capture the current size of the log file. This will be used as the starting
+                # point for searching the log file for a particular string.
+                log_offset = lib.get_file_size_by_path(paths.server_log_path())
+
+                self.admin.assert_icommand(['iput', filename])
+
+                msg = 'RULE_ENGINE_SKIP_OPERATION (5001000) incorrectly returned from PEP [pep_api_data_obj_put_post]'
+                lib.delayAssert(lambda: lib.log_message_occurrences_greater_than_count(msg=msg, count=0, start_index=log_offset))
```
alanking, line 310: Should we skip this test in topology testing? We are reading a server log here.

korydraughn: That sounds correct to me. Will make the change.

server/api/src/rsDataObjUnlink.cpp
korydraughn: Lets make a new issue for this one.

server/api/src/rsDataObjUnlink.cpp
korydraughn: The fsys implementation is specialized for fsys paths, therefore we will let fsys keep it's own implementation and move this version out of rsDataObjUnlink.cpp and into rodsPath.h so that others can use it.

scripts/irods/test/test_rule_engine_plugin_framework.py
korydraughn: done.

korydraughn:
CI core tests passed on Centos 7.
Did not test Ubuntu 16,
Added #

server/api/src/rsDataObjUnlink.cpp
korydraughn: I reviewed the version you mentioned and I agree the previous version should be replaced.

The `isInVault` implementation actually has a substring bug.

trel:
good to merge?

korydraughn:
Yes

PR #4838, opened 2020-04-02
New library and bug fixes (4-2-stable)

korydraughn:


korydraughn:
CI core tests passed on Centos 7.
Did not test Ubuntu 16,
Added #

server/api/src/rsDataObjUnlink.cpp
```diff
@@ -198,32 +248,30 @@ rsDataObjUnlink( rsComm_t *rsComm, dataObjInp_t *dataObjUnlinkInp ) {
...
+    int status = chkPreProcDeleteRule( rsComm, dataObjUnlinkInp, *dataObjInfoHead );
     if ( status < 0 ) {
         return status;
     }
 
-
     myDataObjInfoHead = *dataObjInfoHead;
     if ( strstr( myDataObjInfoHead->dataType, BUNDLE_STR ) != NULL ) { // JMC - backport 4658
-        int numSubfiles; // JMC - backport 4552
+        // TODO Shouldn't this be "clientUser" instead of "proxyUser"?
```
trel, line 263: @jasoncoposky @alanking I'm 0-for-1000 on getting this right.   Thoughts?

server/core/src/client_api_whitelist.cpp
```diff
@@ -225,13 +225,12 @@ namespace irods
...
         try {
-            rodsLog(LOG_DEBUG, "Added API number [%d] to the client API whitelist.", api_number);
             api_numbers_.push_back(api_number);
+            rodsLog(LOG_DEBUG, "Added API number [%d] to the client API whitelist.", api_number);
         }
         catch (const std::exception& e) {
-            const auto ec = SYS_INTERNAL_ERR;
-            const auto* msg = "Could not add API number to whitelist [error_code => %d, exception => %s]";
-            rodsLog(LOG_ERROR, msg, ec, e.what());
+            rodsLog(LOG_ERROR, "Could not add API number to whitelist [error_code => %d, exception => %s]",
```
trel, line 232: want to log the number attempted?    "Could not add API number [%d] to whitelist..."

server/api/src/rsDataObjUnlink.cpp
korydraughn: This could be correct. I grep'd the codebase and found several other places that do a similar check of the proxy user. In my experience, I've always modified or checked the `clientUser` info depending on what needed to be done. This just had me wondering when should the proxy user be inspected? Always? Should it be checked whenever the client user's info is checked?  It is not clear to me. Is the proxy user just for making the connection to other servers?

server/core/src/client_api_whitelist.cpp
korydraughn: Adding ...

korydraughn: done.

trel:
good to merge?

korydraughn:
Yes

PR #4842, opened 2020-04-04
[4612] no longer attempt to cleanup in preremove.sh (4-2-stable)

trel:
The script was attempting to cleanup the iRODS catalog of any
iRODS resources resident on this to-be-removed catalog consumer server.

Any cleanup should be performed by a rodsadmin at a later time.

Removing this code in service of the 'Principle of Least Surprise'.


trel:
@bh9 @mcv21 @kript This won't help with the move from 4.1 because the `preremove.sh` for 4.1.x is already on the machine that's getting upgraded.  However, we avoid any future conversations about this kind of thing since the code is gone :)

PR #4845, opened 2020-04-04
[4307] add helper script for 41 to 42 upgrade

trel:
provides additional confidence in light of any confusion
from the packaged preremove.sh script included in 4.1.x.


trel:
@bh9 @mcv21 @kript comments welcome.

this will be accompanied by a note in the upgrade docs pointing to this script and the best practice around using the local server's hostname in the service account's `irods_environment.json`.

Issue #4856, opened 2020-04-13
Bump spdlog dependency version


alanking:
spdlog was bumped in externals here: https://github.com/irods/externals/commit/c21f9db0f571e6093efc01fe0d18a97a1aad716c

Needs to be bumped in CMakeLists and the build hook.

korydraughn:
See https://github.com/irods/irods/issues/4810

alanking:
Closing as duplicate. Thanks!

PR #4857, opened 2020-04-13
[#4856] Bump spdlog dependency (master)

alanking:
Required a couple of header file changes and syslog_sink has become a
templated class for single-threaded or multi-threaded contexts. Assuming
use of single-threaded for now.

---

Builds in CI, but has not been tested.


server/core/src/irods_logger.cpp
```diff
@@ -126,7 +129,7 @@ void log::init(bool _write_to_stdout, bool _enable_test_mode) noexcept
     }
     else {
         static const char* id = "";
-        sinks.push_back(std::make_shared<spdlog::sinks::syslog_sink>(id, LOG_PID, LOG_LOCAL0));
+        sinks.push_back(std::make_shared<spdlog::sinks::syslog_sink_st>(id, LOG_PID, LOG_LOCAL0, false));
```
alanking, line 132: @korydraughn - Has the `spdlog::sinks::syslog_sink` been used in a multi-threaded context this whole time, or single-threaded? I assumed single-threaded because we are using `spdlog::details::null_mutex` for the other two sinks, but let me know if I misread this.

korydraughn: i actually have these same changes along with others to make this work (due to logical quotas).  there are changes that need to be made to force spdlog to use our packaged fmt library as well.  currently, spdlog will choose to use it's internal version of fmt.

alanking: Got it, thanks. I'll close this PR, but I'll keep the PR just in case.

Issue #4732, opened 2020-02-16
CAT_STATEMENT_TABLE_FULL during ichksum


anderbubble:
- [x] master
- [x] 4-2-stable
------

### iRODS Version, OS and Version

```
$ rpm -q irods-server
irods-server-4.2.7-1.x86_64

$ cat /etc/redhat-release 
CentOS Linux release 7.7.1908 (Core)
```

### What did you try to do?

```
$ sudo -u irods ichksum -MaK --verify -vr "/civilfritz.net/home/janderson/games"
```

### Expected behavior

Successful ichksum of all files under `/civilfritz.net/home/janderson/games`, like

```
    BRAIN.NFO    sha2:V+5dNdmpltyFekNXLA0JyLi8LIoKMVYtIz5nUDTIFqg=
   BRAIN.NFO                  0.039 sec
```

### Observed behavior (including steps to reproduce, if applicable)

```
remote addresses: 128.138.138.142 ERROR: chksumDataObjUtil: rcDataObjChksum error for /civilfritz.net/home/janderson/games/castle/drbrain/BRAINSG.000 status = -860000 CAT_STATEMENT_TABLE_FULL
remote addresses: 128.138.138.142 ERROR: chksumCollUtil:chksumDataObjU failed for /civilfritz.net/home/janderson/games/castle/drbrain/BRAINSG.000.stat = -860000 status = -860000 CAT_STATEMENT_TABLE_FULL
```

d-w-moore:
@anderbubble  -  So far I haven't been able to reproduce this on a vanilla Centos 7.7 setup. Can you share any additional system details? Is this a single-node system? Have any rules been configured which would trigger on opening or reading data objects?

anderbubble:
@d-w-moore it's a single-node system, yes. I had an early version of the logical quotas stuff enabled at one point, but I think it's all torn out now (and I definitely removed all the relevant imeta).

I'm happy to share my config; just let me know what you want to see.

d-w-moore:
@anderbubble  Still working through the simpler possibilities at the moment. Perhaps you could send an 'ilsresc -l' output and a rough idea of the object hierarchy itself -- a listing of the output of "ils -lr /civilfritz.net/home/janderson/games" if it's not confidential, or too large.

anderbubble:
```
[janderson@fox1 ~]$ ilsresc
rootResc:passthru
└── replResc:replication
    ├── fox1Resc:unixfilesystem
[...]

PR #4869, opened 2020-04-20
[#4732] (master) use query iterator for proper cleanup

d-w-moore:
Will soon be ready for merge - pending successful CI testing and review.

### DIFF
--- a/server/api/src/rsDataObjChksum.cpp
+++ b/server/api/src/rsDataObjChksum.cpp
@@ -12,11 +12,18 @@
 #include "rsFileStat.hpp"
 #include "boost/lexical_cast.hpp"
 #include "rsGenQuery.hpp"
+#include "irods_logger.hpp"
+#define RODS_SERVER
+#include "irods_query.hpp"
+#undef RODS_SERVER
 
 // =-=-=-=-=-=-=-
 #include "irods_resource_backport.hpp"
 #include "irods_resource_redirect.hpp"
 
+#include <optional>
+
+using logger = irods::experimental::log;
 
 namespace {
     // assumes zone redirection has already occurred, and that the function is being called on a server in the zone the object belongs to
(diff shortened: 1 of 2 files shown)


server/api/src/rsDataObjChksum.cpp
```diff
@@ -34,34 +38,22 @@ namespace {
...
-        genQueryInp.maxRows = 1;
-        genQueryOut_t *genQueryOut = nullptr;
-        const int status_rsGenQuery = rsGenQuery(rsComm, &genQueryInp, &genQueryOut);
-        clearGenQueryInp(&genQueryInp);
-        if (status_rsGenQuery < 0) {
-            if (status_rsGenQuery == CAT_NO_ROWS_FOUND) {
-                freeGenQueryOut(&genQueryOut);
+        std::optional<rodsLong_t> size_in_database {};
+        std::string select_and_condition {
+            (boost::format("select DATA_SIZE where DATA_PATH = '%s'") % dataObjInfo->filePath).str()
```
korydraughn, line 43: There is nothing wrong with this code, but use the new `fmt` library instead of `boost::format`. The `fmt` library is significantly faster. With that, this line could be this:
```c++
const auto gql = fmt::format("select DATA_SIZE where DATA_PATH = {}", dataObjInfo->filePath);
```
This library uses Python-like format specifiers, and it automatically returns a `std::string` when called so you can use `auto` to decrease typing and keep the code tighter.

server/api/src/rsDataObjChksum.cpp
```diff
@@ -34,34 +38,22 @@ namespace {
...
+        };
+        for (auto&& row : irods::query{rsComm, select_and_condition}) {
+            try {
+                size_in_database = boost::lexical_cast<rodsLong_t>(row[0]);
+            } catch (const boost::bad_lexical_cast&) {
+                rodsLog(LOG_ERROR, "_rsFileChksum: lexical_cast of [%s] for [%s] [%s] [%s] failed", row[0].c_str(), dataObjInfo->filePath, dataObjInfo->rescHier, dataObjInfo->objPath);
```
korydraughn, line 49: On the `master` branch, use the new logger instead of `rodsLog`. This line could be:
```c++
using log = irods::experimental::log; // Just after #includes.
                                      // Could also be another name if "log" is already used.

// Here is how you'd use the new logger.
// You can use a different category if you like. "api" is not required to be used here, but just used
// as an example. See "irods_logger.hpp" for the full list of categories.
log::api::error("_rsFileChksum: lexical_cast of [{}] for [{}] [{}] [{}] failed",
                row[0],
                dataObjInfo->filePath,
                dataObjInfo->rescHier,
                dataObjInfo->objPath);
```

trel:
seems good to me - @korydraughn ?

codacy things appear harmless

d-w-moore:
No don't merge yet though, please. Just had a failed test run and Alan spotted why . My newly fmt-formatted query doesn't include the single-quotes around the { }

d-w-moore:
Will integrate the first codacy (unused variable ) suggestion, for the final testing here. 

Wrt the bare except -- I don't think it is bad usage in this case, as I'm just trying to guard against the clean-up failing "for any reason".

korydraughn:
Looks ok to me.

trel:
yep, looks good - please squash with the # and we'll merge

d-w-moore:
completely tested and ready for merge (master)

PR #4870, opened 2020-04-20
[4866] resolve hostnames with hosts_config.json (master)

jasoncoposky:


lib/core/src/rcMisc.cpp
```diff
@@ -4447,12 +4450,58 @@ static std::string stringify_addrinfo_hints(const struct addrinfo *_hints) {
...
+                    HOST_CONFIG_FILE);
+                return "INVALID_HOST_NAME";
+            }
+
+            hosts_config = json::parse(std::ifstream{cfg_file});
+        }
+
+        for(const auto entry : hosts_config.at("host_entries")) {
+            const std::string address0 = entry.at("addresses").at(0).at("address");
+            const std::string address1 = entry.at("addresses").at(1).at("address");
```
trel, line 4478: needs to loop through all aliases... not just checking the first?

PR #4872, opened 2020-04-21
with_durability, fmt, and non-vault replicas (4-2-stable)

korydraughn:


korydraughn:
Tests passed on centos 7.
Not tested on ubuntu 16.

korydraughn:
Updated and added #

PR #4879, opened 2020-04-23
[4848] Update resource suite to clean up registered file

JustinKyleJames:
Files registered outside of vault are no longer deleted so they need to be cleaned up in test.


trel:
looks good - please #

PR #4889, opened 2020-04-28
[4860] Updated test cases for deprecated message

JustinKyleJames:
about specifying minimum number of replicas to keep


trel:
please # and we'll see if they get a green dot...

trel:
ah, deprecation warning won't appear in master... closing.

Issue #4898, opened 2020-05-01
Add atomic bulk metadata operations to filesystem library


korydraughn:
- [x] master
- [x] 4-2-stable

## Feature
The filesystem library supports adding, setting, and removing a single metadata AVU on a data object or collection. With the use of the `atomic_apply_metadata_operations` API plugin, the filesystem library can support applying multiple AVUs on a single data object or collection.

This enhancement should add multiple function overloads to the filesystem library without breaking existing usage.

PR #4900, opened 2020-05-01
[#4898] Filesystem library now supports atomic bulk metadata operations (4-2-stable)

korydraughn:
Passed unit-tests.

### DIFF
--- a/lib/filesystem/include/filesystem/filesystem.tpp
+++ b/lib/filesystem/include/filesystem/filesystem.tpp
@@ -1,0 +1,74 @@
+namespace
+{
+    template <typename Iterator>
+    auto do_metadata_op(rxComm& _comm,
+                        const path& _path,
+                        Iterator _first,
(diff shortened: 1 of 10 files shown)


lib/filesystem/include/filesystem/filesystem.tpp
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        return true;
+    }
+} // anonymous namespace
+
+template <typename Iterator>
+auto add_metadata(rxComm& _comm, const path& _path, Iterator _first, Iterator _last) -> bool
+{
+    return do_metadata_op(_comm, _path, _first, _last, "set");
```
korydraughn, line 57: "set" should be "add"

trel:
i think we're good, yes?

korydraughn:
I changed the return types for all of the filesystem metadata functions.  That change is in the second commit.

Added #

Issue #2915, opened 2015-09-30
change all mention of SID to zone_key


trel:
- [ ] Update any error messages.
- [ ] Update all variable names.

Examples include:
LocalZoneSID
RemoteZoneSID
SIDKey
REMOTE_SERVER_SID
REMOTE_ZONE_SID_KW
zoneSID
LOCAL_ZONE_SID_KW
CS_NEG_SID_KW
REMOTE_SERVER_SID_NOT_DEFINED
localSID
SID_KEY_KW
requireSIDs

trel:
Maybe we cannot get rid of these until 5.0 (because 4.x may still federate with 3.x and need to know these keywords/error codes).

PR #4910, opened 2020-05-05
[#2915] Server now uses the correct keyword when fetching the local SID (4-2-stable)

korydraughn:


korydraughn:
federation tests pass on ub16/cen7 in CI.

trel:
very good - please #

korydraughn:
Added #

PR #4915, opened 2020-05-08
add timing tests to the python suite (4-2-stable)

jassigill2000:


trel:
seems good to me - does it work?

 if kory has another commit, we'll get this in too.

jassigill2000:
yes it works.

trel:
okay, thanks - holding for now.

trel:
think about which issue to point this at

trel:
does this close those issues?   or should we leave them in 4.2 Backlog?

jassigill2000:
we may want to pull them back where they belong once we move to Kubernetes. So I guess we leave them in the backlog.

trel:
agreed

jassigill2000:
added the # sign

Issue #4916, opened 2020-05-09
Atomic metadata API plugin does not properly support Oracle and MySQL


korydraughn:
- [x] master
- [x] 4-2-stable
---
##  Bug Report
The API plugin uses macros to determine what SQL syntax should be used. However, these macros are never defined because the plugin is not compiled for different databases. Instead, the plugin must inspect the server configuration to make decisions about what SQL syntax should be used.

PR #4918, opened 2020-05-11
[#4916] Atomic API plugin detects database flavor and uses the correct SQL syntax (master)

korydraughn:
Disables all tests that call the database plugin and nanodbc within the same process.


korydraughn:
Added #

PR #4919, opened 2020-05-11
[#4916] Atomic API plugin detects database flavor and uses the correct SQL syntax (4-2-stable)

korydraughn:
Disables all tests that call the database plugin and nanodbc within the same process.

### DIFF
--- a/plugins/api/src/atomic_apply_metadata_operations.cpp
+++ b/plugins/api/src/atomic_apply_metadata_operations.cpp
@@ -58,20 +58,42 @@ namespace
     //
 
     auto call_atomic_apply_metadata_operations(irods::api_entry*, rsComm_t*, bytesBuf_t*, bytesBuf_t**) -> int;
+
     auto is_input_valid(const bytesBuf_t*) -> std::tuple<bool, std::string>;
+
     auto to_bytes_buffer(const std::string& _s) -> bytesBuf_t*;
+
     auto get_file_descriptor(const bytesBuf_t& _buf) -> int;
+
     auto make_error_object(const json& _op, int _op_index, const std::string& _error_msg) -> json;
+
     auto get_object_id(rsComm_t& _comm, const std::string& _entity_name, const std::string& _entity_type) -> int;
-    auto new_database_connection() -> nanodbc::connection;
-    auto user_has_permission_to_modify_metadata(rsComm_t& _comm, nanodbc::connection& _db_conn, int _object_id, const std::string& _entity_type) -> bool;
+
+    auto new_database_connection() -> std::tuple<std::string, nanodbc::connection>;
+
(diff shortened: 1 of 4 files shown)


unit_tests/src/filesystem/test_filesystem.cpp
```diff
@@ -547,8 +547,17 @@ TEST_CASE("filesystem")
             REQUIRE_THROWS(fs::client::remove_metadata(conn, "invalid_path", metadata), "cannot apply metadata operations: unknown object type");
         }
 
+#ifdef IRODS_ENABLE_ALL_UNIT_TESTS
         SECTION("atomic operations")
         {
+            // IMPORTANT
+            // ~~~~~~~~~
+            // This test will fail against databases that have the transaction isolation
+            // level set to REPEATABLE-READ (e.g. MySQL). This is because the database
```
trel, line 556: 'REPEATABLE-READ or higher (MySQL by default)'

korydraughn: Updated.

trel:
looks good - are we happy with it?

plugins/api/src/atomic_apply_metadata_operations.cpp
```diff
@@ -188,12 +210,24 @@ namespace
         }
 
         try {
-            const auto& db_plugin_config = config.at(irods::CFG_PLUGIN_CONFIGURATION_KW).at(irods::PLUGIN_TYPE_DATABASE);
+            auto db_plugin_config = std::move(config.at(irods::CFG_PLUGIN_CONFIGURATION_KW).at(irods::PLUGIN_TYPE_DATABASE));
```
korydraughn, line 213: instead of capturing the json object, this should just capture the property name as a string

plugins/api/src/atomic_apply_metadata_operations.cpp
```diff
@@ -271,17 +305,25 @@ namespace
...
         nanodbc::statement stmt{_db_conn};
 
-#ifdef ORA_ICAT
-        prepare(stmt, "insert into R_META_MAIN (meta_id, meta_attr_name, meta_attr_value, meta_attr_unit, create_ts, modify_ts) "
-                      "values (select R_OBJECTID.nextval from DUAL, ?, ?, ?, ?, ?)");
-#else
-        prepare(stmt, "insert into R_META_MAIN (meta_id, meta_attr_name, meta_attr_value, meta_attr_unit, create_ts, modify_ts) "
-                      "values (nextval('R_OBJECTID'), ?, ?, ?, ?, ?)");
-#endif // ORA_ICAT
+        if (_db_plugin_config.contains("oracle")) {
```
korydraughn, line 312: at the very least, we are paying the prices of a lookup and string comparison multiple times.

korydraughn:
Added #

PR #4920, opened 2020-05-11
Removed Timing tests in their own suite and added it to the python suite (master)

jassigill2000:


korydraughn:
Shouldn't **Removed** be **Moved** in the first commit message?

jassigill2000:
Well 4-2-stable has the same commit message. I removed the tests from the original and moved it to the new place.

trel:
will merge.

PR #4929, opened 2020-05-14
[4862] test now manages core.py correctly (4-2-stable)

d-w-moore:


trel:
looks good to me - let's # it

PR #4939, opened 2020-05-14
Skip tests for prep (4-2-stable)

d-w-moore:


trel:
see codacy opinion

trel:
rest looks good

d-w-moore:
That's a good catch by codacy.

d-w-moore:
Ok, these all behave. I verified with bench-tests  that the modules load and the tests are skipped, without throwing an error, when run individually.

Issue #4943, opened 2020-05-18
Race conditions in logging code lead to `terminating with uncaught exception of type boost::filesystem::filesystem_error: boost::filesystem::directory_iterator::operator++: Not a directory: "/proc/self/fd"`


mcv21:
- [x] master
- [x] 4-2-stable
----
### iRODS Version, OS and Version
iRODS 4.2.7, Ubuntu 18.04

### What did you try to do?

We're sometimes seeing overnight crashes in one of the two iRODS server processes (which confused systemd, which fails to stop the other one properly before trying to start iRODS up again, but that's another story), at almost exactly the time iRODS is rotating its logs

### Observed behavior (including steps to reproduce, if applicable)

The rodsServerLog contains:
```
terminating with uncaught exception of type boost::filesystem::filesystem_error: boost::filesystem::directory_iterator::operator++: Not a directory: "/proc/self/fd"
```

Which was unusual enough to cause me to go digging. The error comes from `get_open_socket_file_descriptors` https://github.com/irods/irods/blob/87e9b6c31ee01b1df3118ccef3cc22b480927785/lib/core/src/irods_socket_information.cpp#L37-L55

(which is in turn called by `create_log_error_prefix`, called by `rodsLog`). 

Now in chkLogfileName, after opening the new log file, it closes fds 0,1,2 before calling `dup2` on them: https://github.com/irods/irods/blob/87e9b6c31ee01b1df3118ccef3cc22b480927785/server/core/src/rsLog.cpp#L123-L129

There is a race here, as alluded to in the `dup2` manual:

> The steps of closing and reusing the file  descriptor  newfd  are  per‐
       formed  atomically.   This  is  important,  because trying to implement
       equivalent functionality using close(2) and dup() would be  subject  to
       race  conditions,  whereby newfd might be reused between the two steps.

In particular, the observed behaviour would be consistent with one thread calling `chkLogfileName` while another calls `rodsLog` - such that the `rodsLog` thread opens `/proc/self/fd` between the `close` and `dup2` calls, gets fd 0 (WLOG) back, and then the `chkLogfileName` calls `dup2` and replaces fd0 with the fd pointing to the new log file. Then `getdents(2)` is called on this fd (by the boost directory iterator), and returns `ENOTDIR`, resulting in the error above.

[...]

PR #4944, opened 2020-05-18
Don't close() fds before calling dup2() on them

mcv21:
It is unnecessary to close() a fd before calling dup2() on it - dup2
will close the fd if necessary. Importantly, dup2 will do so
atomically, whereas the code here is racy - some other thread could
open() something inbetween the close() and dup2(), in which case that
other thread's FD will be overwritten.

This is a partial fix for #4943
Signed-off-by: Matthew Vernon <mv3@sanger.ac.uk>

### DIFF
--- a/server/core/src/rsLog.cpp
+++ b/server/core/src/rsLog.cpp
@@ -120,9 +120,6 @@ chkLogfileName( const char *logDir, const char *logFileName ) {
 
     CurLogfileName = logFile;
 
-    close( 0 );
-    close( 1 );
-    close( 2 );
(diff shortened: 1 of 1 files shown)


mcv21:
[the CI seems to have timed out in build...]

trel:
Yes, Travis has a 50min build limit, and we're exceeding that now on master...  we have some plans to optimize that, but all things in priority order...   4-2-stable is still succeeding under 50min. 

We're looking into reproducing this log rotation issue - thanks for the PR.

PR #4953, opened 2020-05-29
Include <cstring> if using std:mem functions

mp15:
It is a little naughty to use std::memcpy and similar without directly including #include <cstring> and this behaviour is not guaranteed to work on all platforms or with future C++ library versions.

### DIFF
--- a/lib/core/include/dstream.hpp
+++ b/lib/core/include/dstream.hpp
@@ -11,6 +11,7 @@
 #include <stdexcept>
 #include <algorithm>
 #include <utility>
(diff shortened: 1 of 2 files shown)


korydraughn:
Great find!

I've created an issue for this: #4954
Please update your PR to include this issue number.

mp15:
Done

trel:
okay, that's pretty quick :)

we'll get it through CI and merged ... and then cherry-picked to 4-2-stable.

PR #4957, opened 2020-06-01
Replace snprintf with strncat in get_canonical_name

cinek810:
This operation is performed very often and strncat should generally
perform better.

### DIFF
--- a/lib/core/src/rcMisc.cpp
+++ b/lib/core/src/rcMisc.cpp
@@ -4497,7 +4497,7 @@ int get_canonical_name(const char *_hostname, char* _buf, size_t _len) {
     if (ret_getaddrinfo_with_retry ) {
         return ret_getaddrinfo_with_retry;
     }
-    snprintf(_buf, _len, "%s", p_addrinfo->ai_canonname);
+    strlcat(_buf, p_addrinfo->ai_canonname, _len);
     freeaddrinfo(p_addrinfo);
     return 0;
 }


cinek810:
I'm submitting this as a question - our team analysis suggests that timer returned by error message `get_canonical_name of %s is taking %d seconds. This could severely affect the interactivity of your iRODS system.` may be even two times more than what we get measuring DNS responses (under very high load) than what we get from appropriate `time dig ...`. 
We don't see any retry/error messages in log, so maybe.. it's related with snprintf?

trel:
`strlcat` is faster because there's no parsing of the format string?

Can you tell us more about the DNS measurement issue you're seeing?  That seems important and "two times more" certainly seems a problem we should investigate.

cinek810:
Initially, I thought about  strncat, but it failed one of CI checks :) maybe:
```
memcpy(_buf, p_addrinfo->ai_canonname, _len);
_buf[_len-1] = '\0';
```
 We've noticed multiple log messages indicating that DNS query takes more than 5s. DNS resolver for a specific domain runs at the same server as irods catalog (I'll tell more about that on the group meeting). During this time we did an experiment with dig command line tool querying for the same DNS name (more than 1k times) and the maximal query time was ~2s. I know it's not very accurate, but.. looking at the code I found only this.

alanking:
We were going through some older issues and found this.

While we appreciate the contribution and acknowledge that this may introduce a performance boost, we feel that any improvement in the event of a DNS resolution failure will be marginal (i.e. not on the order of seconds). Considering this and in light of other efforts going on within the Consortium, we are going to close this without merging.

If you feel that we should reconsider, please open a new pull request and we can have another look in a separate review. Thanks again for your contribution!

PR #4962, opened 2020-06-08
Remove some server log reading in tests (master)

alanking:
In order to increase coverage in distributed testing and prevent timing issues with respect to flushing to logs, this change uses `msiAssociateKeyValuePairsToObj` instead of `writeLine` in several rules in the delay queue tests and the NREP tests. Also adds a new microservice to get the PID of the agent executing it.

Tests are being run now.

### DIFF
--- a/plugins/microservices/msi_get_agent_pid/libmsi_get_agent_pid.cpp
+++ b/plugins/microservices/msi_get_agent_pid/libmsi_get_agent_pid.cpp
@@ -1,0 +1,62 @@
+#include "irods_ms_plugin.hpp"
+#include "irods_re_structs.hpp"
+#include "msParam.h"
+#include "rodsErrorTable.h"
+#include "irods_error.hpp"
+#include "irods_logger.hpp"
+
+#include <exception>
+#include <functional>
+#include <string>
+
+namespace
+{
+    using log = irods::experimental::log;
+
+    auto msi_impl(msParam_t* _out_pid, ruleExecInfo_t* rei) -> int
+    {
+        try {
+            const auto pid_str = std::to_string(getpid());
+            fillStrInMsParam(_out_pid, pid_str.c_str());
+            return 0;
+        }
+        catch (const std::exception& e) {
+            log::microservice::error(e.what());
+            return SYS_INTERNAL_ERR;
+        }
+        catch (...) {
+            log::microservice::error("An unknown error occurred while processing the request.");
+            return SYS_UNKNOWN_ERROR;
+        }
+    } // msi_impl
+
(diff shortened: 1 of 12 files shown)


alanking:
Passed CI tests

trel:
please review/re-run tests and # if ready

alanking:
CI tests running

plugins/microservices/msi_get_agent_pid/libmsi_get_agent_pid.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include <string>
+
+namespace
+{
+    using log = irods::experimental::log;
+
+    auto msi_impl(msParam_t* out_pid, ruleExecInfo_t* rei) -> int
+    {
+        try {
+            const char* pid = std::to_string(getpid()).c_str();
```
korydraughn, line 19: this is effectively invoking undefined behavior because the temporary produced by `std::to_string` goes away.  you can capture the `std::string` and invoke `.c_str()` on the next call.

plugins/microservices/msi_get_agent_pid/libmsi_get_agent_pid.cpp
```diff
@@ -0,0 +1,47 @@
...
+        return msi;
+    } // make_msi
+} // anonymous namespace
+
+extern "C"
+auto plugin_factory() -> irods::ms_table_entry*
+{
+    return make_msi<msParam_t*>("msi_get_agent_pid", msi_impl);
+}
+
```
korydraughn, line 47: I'm pretty sure you were going to document this microservice, but just so you don't run into the same issues as me, see the following cpp files for examples of how to make doxygen include your microservice documentation.
https://github.com/irods/irods/blob/8ed1608b21682edcfd77a973b5f4cb5b3447e54d/plugins/microservices/msi_atomic_apply_acl_operations/libmsi_atomic_apply_acl_operations.cpp#L79
https://github.com/irods/irods/blob/8ed1608b21682edcfd77a973b5f4cb5b3447e54d/plugins/microservices/msi_atomic_apply_metadata_operations/libmsi_atomic_apply_metadata_operations.cpp#L85

alanking:
Thanks for the review comments. They have been addressed.

All CI tests passed except for one, which was amended to do the right thing.

trel:
Nice.  Please squash when ready.   We'll # after one more look.

alanking:
Fixing the Codacy complaint and running CI one last time...

alanking:
Octothorpe'd and green'd

Issue #4978, opened 2020-06-18
Fix test_federation for older iRODS versions


alanking:
- [x] master
- [x] 4-2-stable
---
- Skip `test_remote_writeLine_remotezone_3722` and `test_remote_writeLine_localzone_3722` for versions older than 4.2.3
- Change the remote vault path if the remote zone is pre-4.x

PR #4979, opened 2020-06-18
[#4978] Fix for federation tests in 3.x (master)

alanking:
Skips two tests that were fixed in 4.2.3 for any version older than
4.2.3. Modifies the remote vault path if the remote version is less
than 4.0.0 as /var/lib/irods does not exist in those versions.

---

Tested successfully against 4-2-stable, need to run against master.


alanking:
Federation tests passed (note: with intermediate replica changes!)

PR #4980, opened 2020-06-18
[#4978] Fix for federation tests in 3.x (4-2-stable)

alanking:
Skips two tests that were fixed in 4.2.3 for any version older than
4.2.3. Modifies the remote vault path if the remote version is less
than 4.0.0 as /var/lib/irods does not exist in those versions.

---

Tests pass


trel:
please to # if happy.

alanking:
#'d

PR #4982, opened 2020-06-22
Add const struct support in key_value_proxy and change interface a little (master)

alanking:
Adds template class for key_value_proxy to support const keyValPair_t
as well as enforcing the fact that the proxy only manages keyValPair_t's.

Several missing pieces were added to the iterator and handle classes to
flesh out the interface a little more.

`get()` was changed to return a pointer instead of a reference to the
underlying object in order to maintain const-ness of underlying struct.
Without this, a const keyValPair_t could be given to a proxy object and
returned as non-const.

### DIFF
--- a/lib/core/include/key_value_proxy.hpp
+++ b/lib/core/include/key_value_proxy.hpp
@@ -5,16 +5,17 @@
 #include "rcMisc.h"
 
 #include "lifetime_manager.hpp"
+
 #include <algorithm>
 #include <limits>
 #include <string>
 #include <string_view>
 
-// TODO: take a pass over std::string allocations
-
 namespace irods::experimental {
-    /// \class key_value_proxy
-    ///
+    /// \brief Tag which indicates that a missing key will be inserted with an empty value.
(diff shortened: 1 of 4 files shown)


alanking:
Tied to irods/irods_client_icommands#135

lib/core/include/key_value_proxy.hpp
```diff
@@ -535,8 +568,23 @@ namespace irods::experimental {
...
+    {
+        return {std::move(key_value_proxy{kvp}), lifetime_manager{kvp}};
+    }
+
+    template<typename kvp_type>
+    static auto make_key_value_proxy(const key_value_proxy<kvp_type>& p) -> lifetime_manager<kvp_type>
+    {
+        return {p.get()};
+    }
+#endif
```
alanking, line 586: delet this

Also, add...
```
template<typename kvp_type>
auto make_key_value_proxy(kvp_type& kvp) -> key_value_proxy<kvp_type>;
```

Issue #4972, opened 2020-06-17
key_value_proxy does not support const keyValPair_t's


korydraughn:
- [x] master
- [x] 4-2-stable
---
### iRODS Version, OS and Version
iRODS v4.2.8

### What did you try to do?
I tried to construct a `key_value_proxy` using a `const keyValPair_t`.

### Expected behavior
It would compile and enforce only read access to the underlying object.

### Observed behavior (including steps to reproduce, if applicable)
Compiler errors.

### Solution
- Change the implementation to a template and adjust the implementation based on the constness.
- Implement a new class that is identical to `key_value_proxy`, but accepts `const keyValPair_t`.

alanking:
I cannot checkbox this, but I believe this is now done in both major branches.

PR #4991, opened 2020-07-01
[#4972] Support const struct in key_value_proxy (4-2-stable)

alanking:
Adds template class for key_value_proxy to support const keyValPair_t
as well as enforcing the fact that the proxy only manages keyValPair_t's.

Several missing pieces were added to the iterator and handle classes to
flesh out the interface a little more.

get() was changed to return a pointer instead of a reference to the
underlying object in order to maintain const-ness of underlying struct.
Without this, a const keyValPair_t could be given to a proxy object and
returned as non-const.

### DIFF
--- a/lib/core/include/key_value_proxy.hpp
+++ b/lib/core/include/key_value_proxy.hpp
@@ -5,14 +5,16 @@
 #include "objInfo.h"
 #include "rcMisc.h"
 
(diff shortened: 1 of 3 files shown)


alanking:
Tied to irods/irods_client_icommands#136

PR #4995, opened 2020-07-03
Increase max password size

emedernach:
In order to use openid tokens for authentication we needed to increase the maximum password length.

modified: lib/api/include/authenticate.h
modified: plugins/database/src/db_plugin.cpp

### DIFF
--- a/lib/api/include/authenticate.h
+++ b/lib/api/include/authenticate.h
@@ -6,7 +6,8 @@
 #define AUTH_SUBOP_REQ_AUTH "request challenge"
 #define AUTH_SUBOP_RESP "challenge response"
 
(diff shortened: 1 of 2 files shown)


trel:
@jasoncoposky @d-w-moore Does this affect how we handle the passwords in PRC or in the database or new flexible auth plugin framework?   Any other surprises you see by bumping this value?

trel:
This PR will break federation and not fit within RESPONSE_LEN for the rest of the conversation between client and server.

We're working with the [Authentication Working Group](https://github.com/trel/irods_working_group_authentication) now to develop a new framework to allow the client to determine how to authenticate and how to handle tokens and the flow itself.  OpenID Connect is a primary target use case for that working group.

PR #4998, opened 2020-07-06
[3461] change default server SSL policy to CS_NEG_REFUSE (master)

trel:
needs a run through CI...

probably missed a couple places in the tests that need to be tweaked for the new default.

### DIFF
--- a/packaging/core.re.template
+++ b/packaging/core.re.template
@@ -9,7 +9,7 @@ printHello { print_hello; }
 #
 # These are sys admin rules for creating and deleting users and renaming
 # the local zone.
-acPreConnect(*OUT) { *OUT="CS_NEG_DONT_CARE"; }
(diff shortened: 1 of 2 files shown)


trel:
need to change the python core.py.template to remain consistent as well

trel:
#'d here and https://github.com/irods/irods_rule_engine_plugin_python/pull/117

oh, also need to do the default cpp...

korydraughn (approved):

alanking:
I have observed some failures in the tests when I cherry-picked this change into a local branch, so I will need to examine this more carefully...

alanking:
Tried again and the tests passed. I think we may have an intermittently failing test which we should watch for (`test_iadmin.Test_Iadmin.test_host_access_control`). It is not related to this change as far as I can tell.

alanking (approved):

PR #4999, opened 2020-07-07
Dependency on OpenSSL 1.1 for debian

stefan-wolfsheimer:
Debian does not support OpenSSL 1.0.0.
We have successfully build irods and icommands with this patch

maybe it would be useful for future releases.


trel:
at this time...

- debian9 has libssl1.0.2 and libssl1.1 available
- debian10 has only libssl1.1 available

trel:
I'm pretty sure this is now moot, we've got deb11 building with the latest/included libssl - @SwooshyCueb can you confirm?

SwooshyCueb:
I haven't tried building for debian and I don't really have a workflow for doing so at the moment. Assigning myself so I can remember to look at this later

alanking:
I've added support for building iRODS (main only) for Debian 11 in the development environment. I could give it a try if you don't feel like getting all that machinery set up.

PR #5009, opened 2020-07-10
[#5011] Add catalog utility libraries (master)

alanking:
Adds a new catalog library for general catalog transaction execution.

Adds a new catalog utilities library for adding more specific operations.

---

Currently leaning on tests for atomic_apply_metadata_operations API plugin, but probably needs further testing (hence, draft for now).

### DIFF
--- a/server/core/src/catalog.cpp
+++ b/server/core/src/catalog.cpp
@@ -1,0 +1,95 @@
+#include <fstream>
+#include <functional>
+#include <stdexcept>
+
+#include "irods_configuration_keywords.hpp"
+#include "irods_get_full_path_for_config_file.hpp"
+#include "irods_logger.hpp"
+#include "irods_stacktrace.hpp"
+#include "objDesc.hpp"
+#include "rodsConnect.h"
+
+#include "catalog.hpp"
+
+#include "json.hpp"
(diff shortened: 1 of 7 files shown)


alanking:
For some reason, the PR is not updating with new commits I've pushed to this branch, but if you check my branch you can see the new changes. I'm going to mark this as ready for review but if it doesn't eventually update the PR, I guess I will open a new PR. Tests pass in CI with the new commit.

server/core/src/catalog.cpp
```diff
@@ -0,0 +1,93 @@
...
+#include "objDesc.hpp"
+#include "rodsConnect.h"
+
+#include "catalog.hpp"
+
+#include "json.hpp"
+#include "fmt/format.h"
+#include "nanodbc/nanodbc.h"
+
+namespace irods::experimental::catalog {
```
trel, line 18: might consider some ::trace calls in here if someone is debugging...

alanking: I will add the DSN string when the connection attempt is made. Otherwise, I think the error messages sent back from nanodbc and our own exceptions should be sufficient without leaking sensitive information in the logs.

PR #5013, opened 2020-07-15
Parallel Transfer Engine and Atomic ACLs (master)

korydraughn:
This is still a work in progress, but is close to being ready for review.


lib/core/include/dstream.hpp
```diff
@@ -163,6 +230,26 @@ namespace irods::experimental::io
...
+        int replica_number() const
+        {
+            return transport_->replica_number();
+        }
+
+        std::string replica_token() const
+        {
+            return transport_->replica_token();
+        }
+
```
korydraughn, line 252: Review whether these functions are named properly and needed at all. Use better return types.

lib/core/include/dstream.hpp
```diff
@@ -419,11 +506,50 @@ namespace irods::experimental::io
...
+                      const filesystem::path& _p,
+                      const leaf_resource_name& _leaf_resource_name,
+                      std::ios_base::openmode _mode = default_openmode<GeneralStream>)
+            : basic_dstream{}
+        {
+            open(_transport, _p, _leaf_resource_name, _mode);
+        }
+
+        basic_dstream(transport<char_type>& _transport,
+                      std::string_view replica_token,
```
korydraughn, line 526: Replace string_view with a replica_token type.

lib/core/include/dstream.hpp
```diff
@@ -516,6 +697,26 @@ namespace irods::experimental::io
...
+
+        int replica_number() const
+        {
+            return buf_.replica_number();
+        }
+
+        std::string replica_token() const
+        {
+            return buf_.replica_token();
+        }
```
korydraughn, line 718: See previous comment about these.

lib/core/include/parallel_transfer_engine.hpp
```diff
@@ -0,0 +1,612 @@
...
+
+            return header;
+        }
+
+        auto restart_file_directory_path() -> boost::filesystem::path
+        {
+            // FIXME Permission error on this path.
+            //return boost::filesystem::temp_directory_path() / "irods/parallel_transfer_engine/restart_file";
+            return boost::filesystem::temp_directory_path() / "irods_parallel_transfer_engine/restart_file";
+        }
```
korydraughn, line 300: The client should probably be able to specify a different location.

lib/core/include/stream_factory_utility.hpp
```diff
@@ -0,0 +1,33 @@
...
+            // is requesting a stream that may need information only available
+            // from the first stream (e.g. a replica number or resource hierarchy).
+            if (_base) {
+                return _factory(_base->replica_token(), replica_number{_base->replica_number()}, _mode);
+            }
+        }
+
+        //return _factory("child_b", _mode);      // ERROR: DIRECT_CHILD_ACCESS
+        //return _factory("root;child_b", _mode); // ERROR: SYS_RESOURCE_DOES_NOT_EXIST
+        //return _factory("other", _mode);
```
korydraughn, line 28: Remove these once everything else is handled and the tests pass.

lib/core/include/transport/default_transport.hpp
```diff
@@ -79,50 +89,131 @@ namespace irods::experimental::io::NAMESPACE_IMPL
...
+                  const root_resource_name& _root_resource_name,
                   std::ios_base::openmode _mode) override
         {
             if (is_open()) {
                 return false;
             }
 
-            return open_impl(_p, _mode, [&_resource_name](auto& _input) {
-                addKeyVal(&_input.condInput, RESC_NAME_KW, _resource_name.c_str());
+            // TODO This leaves the decision making to the server (i.e. the policy).
```
korydraughn, line 128: TODO

lib/core/include/transport/default_transport.hpp
```diff
@@ -79,50 +89,131 @@ namespace irods::experimental::io::NAMESPACE_IMPL
...
+                  const leaf_resource_name& _leaf_resource_name,
+                  std::ios_base::openmode _mode) override
         {
-            openedDataObjInp_t input{};
-            input.l1descInx = fd_;
+            if (is_open()) {
+                return false;
+            }
+
+            // TODO This is when the client knows exactly where the replica should reside.
```
korydraughn, line 142: remove TODO

lib/core/include/transport/default_transport.hpp
```diff
@@ -79,50 +89,131 @@ namespace irods::experimental::io::NAMESPACE_IMPL
...
+            return open_impl(_p, _mode, [&_leaf_resource_name](auto& _input) {
+                addKeyVal(&_input.condInput, LEAF_RESOURCE_NAME_KW, _leaf_resource_name.value.c_str());
+            });
+        }
+
+        bool open(std::string_view replica_token,
```
korydraughn, line 148: use concrete type for replica_token instead of string_view

lib/core/include/transport/default_transport.hpp
```diff
@@ -267,16 +373,25 @@ namespace irods::experimental::io::NAMESPACE_IMPL
 
             _func(input);
 
-            // TODO Modularize the block of code below.
-
-            const auto fd = rxDataObjOpen(comm_, &input);
+            // FIXME The memory referenced by this pointer will need to be freed
+            // by the shared library.
+            char* json_output{}; 
+            const auto fd = rx_replica_open(comm_, &input, &json_output);
```
korydraughn, line 379: Need to think about this one.

lib/core/include/transport/transport.hpp
```diff
@@ -5,9 +5,31 @@
 
 #include <ios>
 #include <string>
+#include <vector>
```
korydraughn, line 8: is this needed?

lib/core/include/transport/transport.hpp
```diff
@@ -5,9 +5,31 @@
...
+
+    struct leaf_resource_name
+    {
+        std::string value;
+    };
+
+    struct replica_number
+    {
+        int value;
+    };
```
korydraughn, line 25: add a replica_token type.

server/api/src/rsDataObjClose.cpp
```diff
@@ -81,16 +90,107 @@ namespace{
...
+                ix::replica_access_table::instance().erase_pid(token.data(), getpid());
+            }
+            else {
+                ix::log::api::warn("Missing replica access token. Replica access token leaked!");
+            }
+        }
+    }
+
+    return ec;
+#endif
```
korydraughn, line 187: remove this once everything is working and passing tests.

server/api/src/rsDataObjOpen.cpp
```diff
@@ -58,15 +63,86 @@
...
 
 using logger = irods::experimental::log;
 
 namespace {
 
+enum class update_operation
+{
+    create,
+    update
+};
```
korydraughn, line 93: this feels too generic. maybe use better names? `update_operation` tells us nothing about what it relates to.

server/api/src/rsDataObjOpen.cpp
```diff
@@ -526,6 +602,39 @@ int change_replica_status_to_intermediate(
     return status;
 } // change_replica_status_to_intermediate
 
+int change_replica_status(rsComm_t& rsComm, dataObjInp_t& dataObjInp, int new_replica_status)
```
korydraughn, line 605: this could possibly replace the other functions that do a similar thing.

server/api/src/rsDataObjOpen.cpp
```diff
@@ -577,21 +690,51 @@ int _rsDataObjOpen(
                     unlock_data_obj();
                 }
             }
-            return l1descInx;
-        }
+            
+            // TODO Shouldn't the l1descInx be compared to three instead of zero
+            // given that the first three file descriptors are special?
+            if (l1descInx >= 0) {
```
korydraughn, line 696: See comment.

server/api/src/rsDataObjOpen.cpp
```diff
@@ -710,13 +907,32 @@ int rsDataObjOpen(
         return l1descInx;
     }
 
+    using irods::experimental::key_value_proxy;
+
+    key_value_proxy kvp{dataObjInp->condInput};
+
+    // If the client specified a leaf resource, then resolve the hierarchy and store it in the keyValPair_t.
+    // This instructs the iRODS server to create the replica at the specified resource if it does not exist.
+    if (kvp.contains(LEAF_RESOURCE_NAME_KW)) {
```
korydraughn, line 916: need a unit-test for this.

server/core/include/objDesc.hpp
```diff
@@ -58,6 +57,8 @@ typedef struct l1desc {
                                      * on close */
     rodsServerHost_t *remoteZoneHost;
     char in_pdmo[MAX_NAME_LEN];
+
+    std::string replica_token;
```
korydraughn, line 61: this variable needs to be cleared either on init/deinit of the L1 descriptor.

server/core/include/replica_access_table.hpp
```diff
@@ -0,0 +1,109 @@
+#ifndef IRODS_REPLICA_ACCESS_TABLE_HPP
```
korydraughn, line 1: this file needs to be documented.

server/core/include/replica_access_table.hpp
```diff
@@ -0,0 +1,109 @@
...
+        ///
+        auto contains(replica_token_view_type token,
+                      data_id_type data_id,
+                      replica_number_type replica_number) -> bool;
+
+        ///
+        auto erase_pid(replica_token_view_type token, pid_t pid) -> std::optional<restorable_entry>;
+
+        ///
+        auto restore(const restorable_entry& entry) -> void;
```
korydraughn, line 102: consider using different names and function signatures.

server/core/include/server_utilities.hpp
```diff
@@ -0,0 +1,28 @@
...
+    ///
+    /// \user Server
+    ///
+    /// \param[in] comm  A reference to the communication object.
+    /// \param[in] input A reference to the \p dataObjInp_t containing the \p keyValPair_t.
+    ///
+    /// \return A boolean value indicating whether the FORCE_FLAG_KW is required or not.
+    /// \retval true  If the keyword is required.
+    /// \retval false If the keyword is not required.
+    auto is_force_flag_required(rsComm_t& comm, /*const*/ dataObjInp_t& input) -> bool;
```
korydraughn, line 24: review whether `const` can be used.

server/core/src/replica_access_table.cpp
```diff
@@ -0,0 +1,245 @@
...
+#include <boost/uuid/uuid.hpp>
+#include <boost/uuid/uuid_generators.hpp>
+#include <boost/uuid/uuid_io.hpp>
+
+#include <memory>
+
+// TODO Shared Memory Resource via PMR Allocators!
+// Requires:
+// - An implementation of a std::pmr::memory_resource holding the shared memory.
+// - Previously allocated shared memory pool.
```
korydraughn, line 20: investigate later, but remove when everything is working and passing tests.

server/core/src/replica_access_table.cpp
```diff
@@ -0,0 +1,245 @@
...
+        const char* g_mutex_name = "irods_replica_access_table_mutex";
+        const char* g_segment_name = "irods_replica_access_table";
+        const std::size_t g_segment_size = 1'000'000; // TODO This could be a knob.
+
+        bool g_initialized = false;
+        pid_t g_owner_pid;
+        std::unique_ptr<bi::managed_shared_memory> g_segment;
+        std::unique_ptr<void_allocator_type> g_allocator;
+        std::unique_ptr<bi::named_mutex> g_mutex;
+        map_type* g_fd_info_map;
```
korydraughn, line 70: maybe document these.

server/core/src/replica_access_table.cpp
```diff
@@ -0,0 +1,245 @@
...
+        using replica_token_view_type = replica_access_table::replica_token_view_type;
+        using data_id_type            = replica_access_table::data_id_type;
+        using replica_number_type     = replica_access_table::replica_number_type;
+        using restorable_entry_type   = replica_access_table::restorable_entry;
+
+        using segment_manager_type  = bi::managed_shared_memory::segment_manager;
+        using void_allocator_type   = bi::allocator<void, segment_manager_type>;
+        // clang-format on
+
+        struct file_descriptor_info
```
korydraughn, line 39: document this

server/core/src/rodsServer.cpp
```diff
@@ -261,6 +263,10 @@ main( int argc, char **argv )
 
     init_logger(write_to_stdout, enable_test_mode);
 
+    irods::experimental::replica_access_table::init();
+    irods::at_scope_exit deinit_fd_table{[] { irods::experimental::replica_access_table::deinit(); }};
+    // TODO Consider looking at "serverExit" to deinit the replica access table.
```
korydraughn, line 268: See comment

server/core/src/server_utilities.cpp
```diff
@@ -0,0 +1,22 @@
...
+#include "dataObjInpOut.h"
+#include "key_value_proxy.hpp"
+
+#define IRODS_FILESYSTEM_ENABLE_SERVER_SIDE_API
+#include "filesystem.hpp"
+
+namespace irods
+{
+    // FIXME The key_value_proxy class does not support const keyValPair_t's.
+    auto is_force_flag_required(rsComm_t& comm, /*const*/ dataObjInp_t& input) -> bool
```
korydraughn, line 13: the `const` keyword can be enabled now.

server/core/src/voting.cpp
```diff
@@ -47,7 +51,24 @@ namespace {
 
         const auto& r = repl->get();
         if (INTERMEDIATE_REPLICA == r.replica_status()) {
-            return vote::zero;
+            // Because the replica is in an intermediate state, we must check if the client
+            // provided a stream token. The stream token represents a piece of information that
```
korydraughn, line 55: replace stream token with replica token.

unit_tests/src/test_dstream.cpp
```diff
@@ -141,5 +141,104 @@ TEST_CASE("dstream", "[iostreams]")
...
+
+    SECTION("bad leaf resoure")
+    {
+
+    }
+
+    SECTION("phymv replica")
+    {
+
+    }
```
korydraughn, line 242: implement these if applicable.

unit_tests/src/test_parallel_transfer_engine.cpp
```diff
@@ -0,0 +1,239 @@
+#include "catch.hpp"
```
korydraughn, line 1: remove hard-coded values and update tests.

unit_tests/src/test_replica_open_and_close.cpp
```diff
@@ -0,0 +1,57 @@
...
+    std::strcpy(input.objPath, path.c_str());
+
+    char* json_output = nullptr;
+
+    const auto fd = rc_replica_open(static_cast<RcComm*>(conn), &input, &json_output);
+    REQUIRE(fd >= 3);
+
+    const auto close_input = json{
+        {"update_catalog", true},
+        {"compute_checksum", false},
```
korydraughn, line 50: test needs to verify checksum support.

korydraughn:
The parallel transfer engine library may be easier to follow if implemented using inheritance instead of templates.

However, the template situation could be resolved when the compiler used supports Concepts.

korydraughn:
Tests passed for ubuntu 16.

server/core/src/rodsServer.cpp
korydraughn: this is likely ok because the init call performs cleanup before setting up the world.

lib/core/include/dstream.hpp
```diff
@@ -163,6 +230,26 @@ namespace irods::experimental::io
...
+        replica_number replica_number() const
+        {
+            return transport_->replica_number();
+        }
+
+        replica_token replica_token() const
+        {
+            return transport_->replica_token();
+        }
+
```
korydraughn, line 252: these should probably be returned by reference.  how expensive is a copy here, though?

server/api/src/rsDataObjOpen.cpp
```diff
@@ -526,6 +602,39 @@ int change_replica_status_to_intermediate(
...
+    }
+
+    dataObjInfo_t info{};
+    rstrcpy(info.objPath, dataObjInp.objPath, MAX_NAME_LEN);
+
+    keyValPair_t kvp{};
+    irods::experimental::key_value_proxy dst{kvp};
+    irods::experimental::key_value_proxy src{dataObjInp.condInput};
+
+    dst[RESC_HIER_STR_KW] = src[RESC_HIER_STR_KW];
```
korydraughn, line 621: the resource hierarchy is supposed to be set in the dataObjInfo_t, not the keyValPair_t.

server/api/src/rsDataObjOpen.cpp
korydraughn: this will likely be an istream test.

server/api/src/rsDataObjOpen.cpp
korydraughn: this will likely be an istream test.

server/core/include/replica_access_table.hpp
korydraughn: done.

korydraughn:
Passed CI tests on Centos 7

server/api/src/rsDataObjOpen.cpp
korydraughn: add a comment, so this is fine.

server/api/src/rsDataObjOpen.cpp
korydraughn: This is likely something that will need to be addressed throughout the code base. Also, this is not necessarily wrong.

unit_tests/src/test_replica_open_and_close.cpp
korydraughn: done

plugins/api/src/replica_close.cpp
```diff
@@ -0,0 +1,378 @@
...
+            const auto l1desc_index = json_input.at("file_descriptor").get<int>();
+
+            log::api::trace("Checking file descriptor ...");
+
+            if (l1desc_index < 0 || l1desc_index >= NUM_L1_DESC) {
+                log::api::error("Invalid file descriptor [error_code={}, fd={}].", BAD_INPUT_DESC_INDEX, l1desc_index);
+                return BAD_INPUT_DESC_INDEX;
+            }
+
+            const auto& l1desc = L1desc[l1desc_index];
```
korydraughn, line 194: should this plugin be using the L1 descriptor table at all given the L1 descriptor information was provided by the client?

lib/core/include/dstream.hpp
jasoncoposky: allocating a std::string could potentially be expensive.

plugins/api/src/replica_close.cpp
jasoncoposky: there is also the dataObjInfo from the catalog?

plugins/api/src/replica_close.cpp
```diff
@@ -0,0 +1,378 @@
...
+                    const auto compute_checksum = (!json_input.contains("compute_checksum") ||
+                                                    json_input.at("compute_checksum").get<bool>());
+
+                    if (compute_checksum) {
+                        log::api::trace("Computing checksum ...");
+                        const int replica_number = fd_info.at("data_object_info").at("replica_number").get<int>();
+                        constexpr const auto calculation = fs::verification_calculation::always;
+                        fs::server::data_object_checksum(*_comm, logical_path, replica_number, calculation);
+                    }
+                }
```
korydraughn, line 293: this code feels too loose. it can be invoked with any JSON data following the correct structure.

lib/core/include/dstream.hpp
korydraughn: True. I think there are a lot of things to consider here.

The values returned are constructed on demand making it difficult to impossible to return a reference. There is also the case of how often will these functions be invoked. The results will likely be short strings more often than not so they should hit the SBO and avoid heap allocation. These may also hit RVO too.

With that, i feel that the copies here aren't that bad (they may be optimized away).

This is under the experimental namespace too, so we have that going for us :)

server/api/src/rsDataObjOpen.cpp
korydraughn: This keyword allows any client to bypass voting and specify where a replica should be created, if it does not exist. Normally, clients are only allowed to specify a root resource and the server handles the rest.

Should this keyword be accessible only on the server-side?

server/api/src/rsDataObjOpen.cpp
korydraughn: This can be revisited in another PR.

lib/core/include/transport/default_transport.hpp
korydraughn: iRODS needs to provide its own heap memory deallocation function (i.e. `irods_free`). The client can't assume that memory allocated by the API is safe to be freed using `std::free`.

korydraughn: Or the iRODS API must clearly state how allocated memory should be deallocated. This could be as simple as stating that all API functions are required to use `malloc` and `free` for memory allocation. This information should be documented in the header file of each API function.

jasoncoposky: We had a head start with the life_time_manager library, is this still a place where we want to handle this?

korydraughn: C clients can't use that library and that library doesn't exactly fix the issue of mixed memory allocation functions.  This is mostly about guaranteeing that users of the C API free memory correctly. The lifetime manager would use this C function underneath guaranteeing safe C++ code.

server/api/src/rsDataObjOpen.cpp
korydraughn: replace "resolve" with "discover" in comment.

It has been decided that "LEAF_RESOURCE_NAME_KW" should be a server-side thing only.

plugins/api/src/replica_close.cpp
```diff
@@ -0,0 +1,377 @@
...
+                    return ec;
+                }
+
+                freeL1desc(l1desc_index);
+
+                return 0;
+            }
+#endif
+
+            if (l1desc.inuseFlag == FD_INUSE) {
```
korydraughn, line 240: swap this if-stmt and the one on line 265

korydraughn: also make sure the l1desc info matches the JSON data.

plugins/api/src/replica_open.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+        log::api::trace("Constructing JSON ...");
+
+        const auto json_input = nlohmann::json{{"fd", fd}}.dump();
+        char* json_output{};
+
+        log::api::trace("JSON = {}", json_input);
+        log::api::trace("Retrieving file descriptor information ...");
+
+        if (const auto ec = rs_get_file_descriptor_info(_comm, json_input.data(), &json_output); ec != 0) {
```
korydraughn, line 80: this function needs to support redirect because the actual info is in the remote computer.

server/api/src/rsDataObjOpen.cpp
korydraughn: this cannot be made a server-side only keyword due to the iRODS system not being able to distinguish clients and servers. this keyword does not introduce a new situation either. it only provides a convenient way to generate the resource hierarchy.

marking as resolved.

lib/core/include/transport/default_transport.hpp
korydraughn: this needs further investigation. leaving resolution of this issue out of this PR.

plugins/api/src/replica_close.cpp
korydraughn: the JSON data is not needed anymore given the swapping of the operations.

plugins/api/src/replica_close.cpp
korydraughn: no longer an issue due to swapping of operations.

plugins/api/src/replica_close.cpp
```diff
@@ -0,0 +1,342 @@
...
+    auto free_l1_descriptor(int _l1desc_index) -> int
+    {
+        // FIXME If this fails, what should we do?
+        // Should failing be treated as an error or be ignored?
+        if (const auto ec = freeL1desc(_l1desc_index); ec != 0) {
+            log::api::error("Failed to release L1 descriptor [error_code={}].", ec);
+            return ec;
+        }
+
+        return 0;
```
korydraughn, line 173: See comment in highlighted code.

korydraughn: This function is called just before returning.

server/core/src/rodsAgent.cpp
```diff
@@ -271,7 +274,15 @@ runIrodsAgentFactory( sockaddr_un agent_addr ) {
...
             }
+
             rmProcLog( reaped_pid );
+
+            // TODO This should be scheduled on a thread pool to avoid blocking.
+            // Or maybe the replica access table allows marking PIDs for removal and
+            // invoking normal operations cause the table to clean-up before carrying out
+            // the operation.
+            ix::log::agent_factory::trace("Scheduling removal of agent PID [{}] from the replica access table ...", reaped_pid);
+            ix::replica_access_table::instance().erase_pid(reaped_pid);
```
korydraughn, line 285: this needs to be scheduled on a thread pool.

unit_tests/src/test_dstream.cpp
```diff
@@ -141,5 +137,190 @@ TEST_CASE("dstream", "[iostreams]")
...
+
+            // Create the second replica.
+            io::client::native_transport tp{conn};
+            io::odstream out{tp, path, io::root_resource_name{"ut_ufs_resc_2"}};
+            REQUIRE(out);
+            out << "Hello, iRODS!";
+            out.close();
+
+            // TODO Remove this
+            //std::system(("ils -l " + path.string()).data());
```
korydraughn, line 262: remove this

unit_tests/src/test_dstream.cpp
```diff
@@ -141,5 +137,190 @@ TEST_CASE("dstream", "[iostreams]")
...
+    {
+        io::client::native_transport tp{conn};
+        io::odstream out{tp, sandbox / "should_not_exist", io::replica_number{1}};
+        REQUIRE_FALSE(out);
+    }
+
+    SECTION("phymv replica")
+    {
+
+    }
```
korydraughn, line 299: remove this

plugins/api/src/replica_open.cpp
korydraughn: done.

server/core/src/rodsAgent.cpp
korydraughn: done.  however, there is the situation of multiple clients connecting to the same server and writing to it. this will lead to lock contention. a different approach may be needed to handle that (e.g. marking PIDs for removal and letting the thread pool erase the PIDs in bulk).

korydraughn: until this becomes a problem, we'll let this solution ride.

PR #5034, opened 2020-08-04
[5007] remove filesystem unit test

jasoncoposky:


unit_tests/src/filesystem/test_filesystem.cpp
```diff
@@ -130,6 +130,17 @@ TEST_CASE("filesystem")
...
 
+    SECTION("create and remove collections with extended options")
+    {
+        const fs::path col1 = sandbox / "col1";
+        REQUIRE(fs::client::create_collection(conn, col1));
+        REQUIRE(fs::client::create_collection(conn, sandbox / "col2", col1));
+        REQUIRE(fs::client::create_collections(conn, sandbox / "col2/col3/col4/col5"));
+        REQUIRE(fs::client::remove(conn, col1, {true, false, false, true, false}));
+        REQUIRE(fs::client::remove_all(conn, sandbox / "col2/col3/col4", {true, false, false, true, false}));
+        REQUIRE(fs::client::remove_all(conn, sandbox / "col2", {true, false, false, true, false}));
```
korydraughn, line 141: test seems fine. it would likely be better if these booleans were given names or pulled out and set in the actual struct.
```c++
fs::extended_remove_options opts;
opts.recursive = true;
opts.no_trash = false;
// etc.
```
Maybe this gets refactored in the future to use an enum class supporting bitwise operations.
```c++
using rm_opt = extended_remove_options;
fs::client::remove(conn, col, rm_opt::recursive | rm_opt::no_trash | rm_opt::progress);
```

Issue #4969, opened 2020-06-17
Cannot open multiple streams to the same replica


korydraughn:
- [x] master
- [x] 4-2-stable

## Bug Report
When a data object is in an intermediate state, the voting mechanism returns a vote of zero causing subsequent calls to `rxDataObjOpen` to fail.
https://github.com/irods/irods/blob/baef7769a0a8ee6561f562f2d4b0c5fc6987657d/server/core/src/voting.cpp#L49-L51

korydraughn:
Handled by #5038 (master)

Issue #4970, opened 2020-06-17
Cannot open an existing replica using the replica number


korydraughn:
- [x] master
- [x] 4-2-stable

## Bug Report
Attempting to open a specific replica using the replica number triggers the `CREATE_OPERATION` code path which does not make sense. If a replica number is provided on a call to `rxDataObjOpen`, this implies that the replica exists and the server should handle it as a `WRITE_OPERATION`.

korydraughn:
Handled by #5038 (master)

PR #5038, opened 2020-08-06
Parallel Transfer Engine, Atomic ACLs, Replica Tokens! (master)

korydraughn:
Passes all core and federation tests.

Replaces #5013 

Requires irods/irods_client_icommands#138


korydraughn:
Added #

korydraughn:
Resolves #4969 and #4970

PR #5042, opened 2020-08-10
[5007] add replication api

jasoncoposky:


plugins/experimental/parallel_filesystem_operation.hpp
```diff
@@ -0,0 +1,170 @@
...
+
+                if(psc.contains("api") && psc.at("api").contains(instance_name_)) {
+                    auto in = psc.at("api").at(instance_name_);
+                    if(in.contains("thread_count")) {
+                        return in.at("thread_count").get<int>();
+                    }
+                }
+            }
+
+            return 4;
```
trel, line 49: comment about why 4 is a default here?

scripts/irods/test/test_experimental_parallel_replication.py
```diff
@@ -0,0 +1,153 @@
...
+from . import session
+
+class TestParallelReplication(session.make_sessions_mixin([], []), unittest.TestCase):
+
+    def setUp(self):
+        super(TestParallelReplication, self).setUp()
+        with session.make_session_for_existing_admin() as admin:
+             admin.assert_icommand(['iadmin', "mkresc", "testResc", 'unixfilesystem', lib.get_hostname() + ":" + "/tmp/irods/testResc"], 'STDOUT_SINGLELINE', 'unixfilesystem')
+             admin.assert_icommand(['iadmin', "mkresc", "anotherResc", 'unixfilesystem', lib.get_hostname() + ":" + "/tmp/irods/anotherResc"], 'STDOUT_SINGLELINE', 'unixfilesystem')
+             admin.assert_icommand(['iadmin', "mkuser", "alice", 'rodsuser'])
```
trel, line 18: i think these are already created by the test harness... surprised you didn't hit duplicate entries...

plugins/experimental/parallel_filesystem_operation.hpp
jasoncoposky: It is the default for the delayed execution server and a polite number at around 10% the core/thread count of a typical server.  I should be using a constant here but was not sure if we wanted those two to be the same or name another thing locally.

scripts/irods/test/test_experimental_parallel_replication.py
jasoncoposky: I am not inheriting from the resource base class so I had to make them myself.

scripts/irods/test/test_experimental_parallel_replication.py
```diff
@@ -0,0 +1,153 @@
...
+                admin.assert_icommand(['ils', '-l', file_name], 'STDOUT_SINGLELINE', file_name)
+                admin.assert_icommand(['irepl', '-R', 'anotherResc', file_name])
+                admin.assert_icommand(['ils', '-l', file_name], 'STDOUT_SINGLELINE', file_name)
+                admin.assert_icommand(['iput', '-f', '-n', '0', file_name])
+                admin.assert_icommand(['ils', '-l', file_name], 'STDOUT_SINGLELINE', file_name)
+
+                endpoint = 'replicate'
+                options  = """{{"logical_path" : "{lp}",
+                                "update_all_replicas" : true,
+                                "source_resource" : "{src}"}}""".format(lp = logical_path, src = "demoResc", dst = "testResc")
```
trel, line 141: codacy says dst not used, then please #

trel:
i see 3 commits - please squash to 1

Issue #5043, opened 2020-08-12
Given an error name, need a function to look up the integer code


d-w-moore:
- [x] master
- [x] 4-2-stable
------

## Feature

In rule code, given an error name, need a function or microservice that returns the corresponding error code, ie `error( name ) -> int`. This could help the code to be more self-documenting.

For  Instance, `fail (-66000)` could be replaced by `fail(error( "SYS_NOT_SUPPORTED"))`
 
Or, in the following:
```
  *x = errorcode( msi_something( ... ) )
  if (*x == -808000) { ...  # not clear what we are testing for
```
We could instead say : `if (*x == error("CAT_NO_ROWS_FOUND"))` which would be much clearer.

d-w-moore:
Also there is duplication in the test suite, in lines such as [this one](https://github.com/irods/irods/blob/a2b6aed1617f075269c1b3a53431eaa6ba7cbc41/scripts/irods/test/test_rule_engine_plugin_framework.py#L49), but that's not difficult to fix with a module like this --
```
__all__ = []

import re
import os

def include_finder ( base_file_name ):
    return os.path.join( 
        os.path.dirname(__file__),
        '../../lib/core/include',
        base_file_name)

def add_constants( g_dict, pattern, file_name):
    with open( include_finder( file_name ), 'r') as include_file:
        for line in include_file:
            m = pattern.match( line )
            if m:
                g_dict [ m.group(1) ] = int(m.group(2))
                __all__ . append( m.group(1) )
                
add_constants( globals(),
               re.compile(r'^\s*NEW_ERROR\s*\(\s*(\w+)\s*,\s*((?:[\-\+]?)[\d]+)\)'),
               'rodsErrorTable.h' )

add_constants( globals(),
               re.compile(r'^\s*NEW_STATE\s*\(\s*(\w+)\s*,\s*((?:[\-\+]?)[\d]+)\)'),
               'irods_state_table.h' )
```

d-w-moore:
Ok that won't do it. Of course not. The include file isn't there at run time.

PR #5044, opened 2020-08-12
[#5043] error() msvc - look up error code from name (4-2-stable)

d-w-moore:
Adds `error( "ERROR_NAME" ) -> [negative integer]`  to the system microservice table.

Tests not yet written.

### DIFF
--- a/plugins/rule_engines/irods_rule_engine_plugin-irods_rule_language/functions.cpp
+++ b/plugins/rule_engines/irods_rule_engine_plugin-irods_rule_language/functions.cpp
@@ -1,5 +1,8 @@
 /* For copyright information please refer to files in the COPYRIGHT directory
  */
+
+#define MAKE_IRODS_ERROR_MAP
+#include "rodsErrorTable.h"
 #include "index.hpp"
 #include "functions.hpp"
 #include "arithmetics.hpp"
@@ -2048,6 +2051,19 @@ Res *smsi_substr( Node** paramsr, int, Node* node, ruleExecInfo_t*, int, Env*, r
     return retres;
 }
 
+Res *smsi_error( Node** paramsr, int, Node* node, ruleExecInfo_t*, int, Env*, rError_t* errmsg, Region* r ) {
+	Res * errName = paramsr[0];
+	const auto& name_map = irods_error_map_construction::irods_error_name_map;
+	int i {};
+	try { i = name_map.at( errName->text ); } catch(...) { }
+	if ( i == int{} )
+	{
+            generateAndAddErrMsg( "invalid error name", node, RE_RUNTIME_ERROR, errmsg );
+	    return newErrorRes(r, RE_RUNTIME_ERROR);
+	}
+	return newIntRes(r, -abs(i));
+}
+
 Res *smsi_split( Node** paramsr, int, Node*, ruleExecInfo_t*, int, Env*, rError_t*, Region* r ) {
     Res *strres = ( Res * )paramsr[0];
     Res *delimres = ( Res * )paramsr[1];
@@ -2672,6 +2688,7 @@ void getSystemFunctions( Hashtable *ft, Region* r ) {
     insertIntoHashTable( ft, "strlen", newFunctionFD( "string->integer", smsi_strlen, r ) );
     insertIntoHashTable( ft, "substr", newFunctionFD( "string * integer * integer->string", smsi_substr, r ) );
     insertIntoHashTable( ft, "split", newFunctionFD( "string * string -> list string", smsi_split, r ) );
+    insertIntoHashTable( ft, "error", newFunctionFD( "string -> integer", smsi_error, r ) );
     insertIntoHashTable( ft, "execCmdArg", newFunctionFD( "f string->string", smsi_execCmdArg, r ) );
     insertIntoHashTable( ft, "query", newFunctionFD( "expression ? -> `GenQueryInp_PI` * `GenQueryOut_PI`", smsi_query, r ) );
     insertIntoHashTable( ft, "unspeced", newFunctionFD( "-> ?", smsi_undefined, r ) );


plugins/rule_engines/irods_rule_engine_plugin-irods_rule_language/functions.cpp
```diff
@@ -2048,6 +2051,19 @@ Res *smsi_substr( Node** paramsr, int, Node* node, ruleExecInfo_t*, int, Env*, r
...
 }
 
+Res *smsi_error( Node** paramsr, int, Node* node, ruleExecInfo_t*, int, Env*, rError_t* errmsg, Region* r ) {
+	Res * errName = paramsr[0];
+	const auto& name_map = irods_error_map_construction::irods_error_name_map;
+	int i {};
+	try { i = name_map.at( errName->text ); } catch(...) { }
+	if ( i == int{} )
+	{
+            generateAndAddErrMsg( "invalid error name", node, RE_RUNTIME_ERROR, errmsg );
```
trel, line 2061: whitespace

jasoncoposky:
> seems good to me - only comment would be to consider the name 'errorname', to be a bit more parallel with existing 'errorcode'.
> 
> and we should update the irods_docs to have this information and an example.

This is returning an error code given an error name, and since errorcode() was already taken we went with error().

trel:
> since errorcode() was already taken

gotcha.

plugins/rule_engines/irods_rule_engine_plugin-irods_rule_language/functions.cpp
```diff
@@ -2048,6 +2051,19 @@ Res *smsi_substr( Node** paramsr, int, Node* node, ruleExecInfo_t*, int, Env*, r
     return retres;
 }
 
+Res *smsi_error( Node** paramsr, int, Node* node, ruleExecInfo_t*, int, Env*, rError_t* errmsg, Region* r ) {
+	Res * errName = paramsr[0];
+	const auto& name_map = irods_error_map_construction::irods_error_name_map;
+	int i {};
+	try { i = name_map.at( errName->text ); } catch(...) { }
+	if ( i == int{} )
```
korydraughn, line 2059: what does using `int{}` give us over `0`?  this feels like it can confuse people down the road.

plugins/rule_engines/irods_rule_engine_plugin-irods_rule_language/functions.cpp
```diff
@@ -1,5 +1,8 @@
 /* For copyright information please refer to files in the COPYRIGHT directory
  */
+
+#define MAKE_IRODS_ERROR_MAP
+#include "rodsErrorTable.h"
```
korydraughn, line 5: i'd move this under the other includes so that they aren't affected by the macro.  you could also undef it following the include.  this is more of a recommendation than anything.

plugins/rule_engines/irods_rule_engine_plugin-irods_rule_language/functions.cpp
```diff
@@ -2048,6 +2051,19 @@ Res *smsi_substr( Node** paramsr, int, Node* node, ruleExecInfo_t*, int, Env*, r
...
+	Res * errName = paramsr[0];
+	const auto& name_map = irods_error_map_construction::irods_error_name_map;
+	int i {};
+	try { i = name_map.at( errName->text ); } catch(...) { }
+	if ( i == int{} )
+	{
+            generateAndAddErrMsg( "invalid error name", node, RE_RUNTIME_ERROR, errmsg );
+	    return newErrorRes(r, RE_RUNTIME_ERROR);
+	}
+	return newIntRes(r, -abs(i));
```
korydraughn, line 2064: should negation be used here?  the integer value is already negative, right?

korydraughn:
should this consider the state codes too?

d-w-moore:
Right, and they would be positive which - if we include them - corroborates the removal of the -abs( ... ) part.

korydraughn:
irods_state_table.h

d-w-moore:
So my Q's right now, 1 ) Does the "error" name stay , when /  if we're also returning the non-error states ?
... and 2) do I build the corresponding name map for states at initialization time , which would parallel the [irods_error_name_map](https://github.com/irods/irods/blob/baef7769a0a8ee6561f562f2d4b0c5fc6987657d/lib/core/include/rodsErrorTable.h#L56) ? that way it would be available for the msvc call and not need rebuilding each time.

plugins/rule_engines/irods_rule_engine_plugin-irods_rule_language/functions.cpp
d-w-moore: I tried moving it under the other includes, but the namespaces I needed were then no longer visible.  I could still do it but it would require making a new compilation unit just for the new microservice.

plugins/rule_engines/irods_rule_engine_plugin-irods_rule_language/functions.cpp
d-w-moore: Yes, agreed. changing as suggested.

korydraughn:
i think the answer depends on what the requirements are for this msi.
- Do we need to expose the state table codes?
- Would it be better to provide a different msi for the state table codes?

d-w-moore:
ok thanks @korydraughn . went ahead and made a new commit, which I'll submit shortly. it works both ways, we'll see how it settles out in the approval department : )
```
$ irule -r irods_rule_engine_plugin-irods_rule_language-instance '*x=error("RULE_ENGINE_CONTINUE"); writeLine("stdout",(*x))' null ruleExecOut
5000000
$ irule -r irods_rule_engine_plugin-irods_rule_language-instance '*x=state("RULE_ENGINE_CONTINUE"); writeLine("stdout",(*x))' null ruleExecOut
5000000
```

plugins/rule_engines/irods_rule_engine_plugin-irods_rule_language/functions.cpp
korydraughn: that is likely due to the other headers including rodsErrorTable.h.  carry on :)

plugins/rule_engines/irods_rule_engine_plugin-irods_rule_language/functions.cpp
```diff
@@ -2048,6 +2055,29 @@ Res *smsi_substr( Node** paramsr, int, Node* node, ruleExecInfo_t*, int, Env*, r
...
+{
+        Res * errName = paramsr[0];
+        int i {0};
+        for (auto & the_map : {
+                                  irods_error_map_construction::irods_error_name_map,
+                                  irods_state_map_construction::irods_state_name_map
+                              })
+        {
+            try { i = the_map.at( errName->text ); } catch(...) { }
+            if (i != 0) break;
```
trel, line 2068: two lines please

trel:
if happy, please squash, #, and remove draft status.

d-w-moore:
Yes will do. Writing tests!

d-w-moore:
Closed in favor of #5066

PR #5047, opened 2020-08-13
5007  fixes for build (master)

jasoncoposky:


trel:
i still see reference to recursive remove in here... is that correct?   if so, # it.

Issue #5057, opened 2020-08-18
Add ability for irods::hierarchy_parser to add parent


alanking:
- [x] master
- [x] 4-2-stable
---
## Feature

Extend interface to allow adding a parent resource to an existing hierarchy, optionally specifying a node in the hierarchy by name where the parent will be inserted (default to adding to beginning of hierarchy).

Example usage:
```
irods::hierarchy_parser p{"a;b;c"};
p.add_parent("z");
REQUIRE("a" == p.next("z"));
p.add_parent("c2", "c");
REQUIRE("c" == p.next("c2"));
```

PR #5059, opened 2020-08-18
[#5057] Add add_parent to hierarchy_parser (4-2-stable)

alanking:
Unit tests pass

### DIFF
--- a/lib/core/src/irods_hierarchy_parser.cpp
+++ b/lib/core/src/irods_hierarchy_parser.cpp
@@ -6,6 +6,8 @@
 #include <iostream>
 #include <sstream>
 
+#include "fmt/format.h"
+
 namespace irods {
 
     static const std::string DELIM = ";";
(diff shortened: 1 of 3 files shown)


trel:
seems good - any other way to break it?   add parent can happen anywhere in the tree... so 'nonexistent' and 'bad input' seems to be the only things...    if good, please add #

lib/core/src/irods_hierarchy_parser.cpp
```diff
@@ -87,10 +87,29 @@ namespace irods {
...
+        const std::string& _parent,
+        const std::string& _child) {
+        if (_parent == hierarchy_parser::delimiter()) {
+            THROW(SYS_INVALID_INPUT_PARAM, "invalid resource name");
+        }
+        auto it = resc_list_.begin();
+        if (!_child.empty()) {
+            it = std::find(resc_list_.begin(), resc_list_.end(), _child);
+            if (it == resc_list_.end()) {
+                THROW(CHILD_NOT_FOUND, (boost::format(
```
alanking, line 108: Need to change this to use `fmt::format`... been sitting on this commit for quite some time!

alanking:
I added a test for the beginning, end, middle, non-existent, and with invalid input (i.e. `;`) so I think that about covers it.

trel:
good good.  # when ready

alanking:
Done. Replaced `boost::format` usage with `fmt::format` as well

PR #5077, opened 2020-08-26
Issue 5030 (4-2-stable)

JustinKyleJames:


server/api/src/rsDataObjRepl.cpp
```diff
@@ -424,6 +425,11 @@ int dataObjCopy(
     int srcRemoteFlag{};
     int source_l1descInx = L1desc[_destination_l1descInx].srcL1descInx;
     int srcL3descInx = L1desc[source_l1descInx].l3descInx;
+
+    // copy data size from src catalog to destination
+    L1desc[_destination_l1descInx].dataSize = L1desc[source_l1descInx].dataObjInfo->dataSize;
+    //L1desc[source_l1descInx].dataSize = L1desc[source_l1descInx].dataObjInfo->dataSize;
```
trel, line 431: leftover?

JustinKyleJames: Left it in there in case someone decides it is needed in the fruture.  Can delete

trel:
do we need a new issue for the first commit?

server/api/src/rsDataObjRepl.cpp
trel: delete it if it's not needed today.   leave a helpful comment if there's something to say.

JustinKyleJames: Probably just left over.

trel:
we need any tests for this?   handled by existing tests elsewhere?

JustinKyleJames:
For 5078, the new S3 plugin won't work without it so it would be tested with the plugin itself.  Not sure how to test it in the core itself.

For 4030, the resource plugin has to set that.  If it is tested it would be in the plugin.

trel:
Got it - thanks.

trel:
If we're happy with this, please # it.

JustinKyleJames:
> If we're happy with this, please # it.

Done

PR #5094, opened 2020-09-02
DNS bug fix and macros (master)

korydraughn:
Passed CI.

This change may require updates to any code that used `irods::query`.


trel:
adaptor -> adapter in the commit message  ... then # em

korydraughn:
Verified that the experimental plugin adapter tests pass and added #

PR #5095, opened 2020-09-02
DNS bug fix and macros (4-2-stable)

korydraughn:
This change may require updates to any code that used `irods::query`.


korydraughn:
Passed CI and added #

PR #5098, opened 2020-09-03
Metadata API

jasoncoposky:


lib/core/include/metadata.hpp
```diff
@@ -0,0 +1,232 @@
...
+            };
+
+            return rxModAVUMetadata(&comm, &inp);
+
+        } // mod_avu_meta
+
+    } // namespace
+
+
+    auto to_type(const std::string& _t)
```
korydraughn, line 91: this would be clearer and safer if this were renamed to `to_entity_type`.

lib/core/include/metadata.hpp
```diff
@@ -0,0 +1,232 @@
...
+
+    } // to_type
+
+
+    auto to_token(entity_type _t)
+    {
+
+        if(type_to_token.find(_t) == type_to_token.end()) {
+            return tokens::none;
+        }
```
korydraughn, line 108: this is unnecessary given that the incoming enumeration guarantees that a valid type exists.

lib/core/include/metadata.hpp
```diff
@@ -0,0 +1,232 @@
...
+        if(token_to_type.find(_t) == token_to_type.end()) {
+            return entity_type::none;
+        }
+
+        return token_to_type.at(_t);
+
+    } // to_type
+
+
+    auto to_token(entity_type _t)
```
korydraughn, line 103: this is safe due to the parameter type, but it might be clearer and fit inline with the changes if it were renamed to `to_entity_token`

lib/core/include/metadata.hpp
```diff
@@ -0,0 +1,232 @@
...
+        return eq;
+    }
+
+    namespace tokens {
+        const std::string none{};
+        const std::string collection{"-C"};
+        const std::string object{"-d"};
+        const std::string user{"-u"};
+        const std::string resource{"-R"};
+    }
```
korydraughn, line 46: is this an implementation detail?  if so, this can be moved into the anonymous namespace below.

lib/core/include/metadata.hpp
```diff
@@ -0,0 +1,232 @@
...
+
+    namespace {
+
+        using bf = boost::format;
+
+        using entity_type = irods::experimental::entity::type;
```
korydraughn, line 52: this can be removed if `type` is renamed

lib/core/include/metadata.hpp
```diff
@@ -0,0 +1,232 @@
...
+        };
+
+        const std::unordered_map<std::string, entity_type> token_to_type {
+            {tokens::collection, entity_type::collection},
+            {tokens::object,     entity_type::object},
+            {tokens::user,       entity_type::user},
+            {tokens::resource,   entity_type::resource}
+        };
+
+        auto mod_avu_meta(rxComm_t& comm, const avu& _md, const avu& _dmd, entity_type _et, const std::string& _tgt, const std::string& _op)
```
korydraughn, line 68: missing leading underscore on `comm`

lib/core/include/metadata.hpp
```diff
@@ -0,0 +1,232 @@
...
+        const std::string none{};
+        const std::string collection{"-C"};
+        const std::string object{"-d"};
+        const std::string user{"-u"};
+        const std::string resource{"-R"};
+    }
+
+    namespace {
+
+        using bf = boost::format;
```
korydraughn, line 50: prefer `fmt` library.  it is faster and is also part of the upcoming standard

lib/core/include/metadata.hpp
```diff
@@ -0,0 +1,223 @@
...
+    using avus = std::vector<avu>;
+
+
+    auto get(rxComm_t& _comm, entity_type _et, const std::string& _tgt)
+    {
+        namespace fs = irods::experimental::filesystem;
+
+        if(entity_type::none == _et) {
+            THROW(SYS_INVALID_INPUT_PARAM, "entity type of none is invalid");
+        }
```
korydraughn, line 169: the removal of `entity_type::none` means this can be removed

unit_tests/src/test_metadata.cpp
```diff
@@ -0,0 +1,285 @@
...
+    const int cp_refresh_time = 600;
+
+    irods::connection_pool conn_pool{cp_size,
+                                     env.rodsHost,
+                                     env.rodsPort,
+                                     env.rodsUserName,
+                                     env.rodsZone,
+                                     cp_refresh_time};
+
+    auto conn = conn_pool.get_connection();
```
korydraughn, line 38: this can be replaced with `irods::experimental::client_connection`, but this is also fine as is

lib/core/include/metadata.hpp
```diff
@@ -0,0 +1,223 @@
...
+        bool eq = _lhs.attribute == _rhs.attribute &&
+                  _lhs.value     == _rhs.value &&
+                  _lhs.units     == _rhs.units;
+        return eq;
+    }
+
+    namespace tokens {
+        const std::string none{};
+        const std::string collection{"-C"};
+        const std::string object{"-d"};
```
korydraughn, line 43: maybe this becomes `data_object`

lib/core/include/entity.hpp
```diff
@@ -0,0 +1,17 @@
+#ifndef ENTITY_HPP
+#define ENTITY_HPP
+
+namespace irods::experimental::entity {
+
+    enum class type {
+        none,
+        collection,
+        object,
```
korydraughn, line 9: maybe this becomes `data_object`

lib/core/include/metadata.hpp
```diff
@@ -0,0 +1,223 @@
...
+
+        return type_to_token.at(_t);
+
+    } // to_token
+
+
+    void set(rxComm_t& _comm, const avu& _md, entity_type _et, const std::string& _tgt)
+    {
+        if(auto ec = mod_avu_meta(_comm, _md, {}, _et, _tgt, "set"); ec < 0) {
+            THROW(ec, bf{"metadata set failed for [%s]"} % _tgt);
```
korydraughn, line 118: fmt library

lib/core/include/metadata.hpp
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+    } // set
+
+
+    void add(rxComm_t& _comm, const avu& _md, entity_type _et, const std::string& _tgt)
+    {
+
+        if(auto ec = mod_avu_meta(_comm, _md, {}, _et, _tgt, "add"); ec < 0) {
+            THROW(ec, bf{"metadata add failed for [%s]"} % _tgt);
```
korydraughn, line 128: fmt library

lib/core/include/metadata.hpp
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+    } // add
+
+
+    void remove(rxComm_t& _comm, const avu& _md, entity_type _et, const std::string& _tgt)
+    {
+
+        if(auto ec = mod_avu_meta(_comm, _md, {}, _et, _tgt, "rm"); ec < 0) {
+            THROW(ec, bf{"metadata remove failed for [%s]"} % _tgt);
```
korydraughn, line 138: fmt library

lib/core/include/metadata.hpp
```diff
@@ -0,0 +1,223 @@
...
+    {
+
+        avu tmp{
+              std::string{"n:"} + _dmd.attribute
+            , std::string{"v:"} + _dmd.value
+            , std::string{"u:"} + _dmd.units
+        };
+
+        if(auto ec = mod_avu_meta(_comm, _md, tmp, _et, _tgt, "mod"); ec < 0) {
+            THROW(ec, bf{"metadata modify failed for [%s]"} % _tgt);
```
korydraughn, line 154: fmt library

lib/core/include/metadata.hpp
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+    } // modify
+
+
+    using avus = std::vector<avu>;
```
korydraughn, line 160: is this a public type?

if not, it should be moved into the function that uses it.

lib/core/include/metadata.hpp
```diff
@@ -0,0 +1,223 @@
...
+            where_fmt = fmt::format(where, coll, obj);
+        }
+        else {
+            where_fmt = fmt::format(where, _tgt);
+        }
+
+        auto type_str  = type_to_string.at(_et);
+        auto query_str = fmt::format(temp, type_str, type_str, type_str, where_fmt);
+
+        auto res = avus{};
```
korydraughn, line 204: nothing else uses this type alias.  this can be `std::vector<avu> res;`

lib/core/include/metadata.hpp
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+        return type_to_token.at(_t);
+
+    } // to_token
+
+
+    void set(rxComm_t& _comm, const avu& _md, entity_type _et, const std::string& _tgt)
+    {
+        if(auto ec = mod_avu_meta(_comm, _md, {}, _et, _tgt, "set"); ec < 0) {
```
korydraughn, line 117: `ec` can be marked `const`

korydraughn:
i'm not sure if function names like `set`, `get`, `remove` (etc.) provide enough information to the reader.  the short names may require readers to inspect the area around these calls to figure out what they are doing.

given that this is in experimental, we don't have to change these names now.  as we gain experience with these, we will figure out if they work well from a maintainability standpoint or not.

however, i feel that the functions that use `auto` as their return type need to specify the trailing return type so that users don't have to read the body to figure it out.  this is more towards functions that are more than a couple lines long.

lib/core/include/metadata.hpp
```diff
@@ -0,0 +1,223 @@
...
+
+    auto operator==(const avu& _lhs, const avu& _rhs) {
+        bool eq = _lhs.attribute == _rhs.attribute &&
+                  _lhs.value     == _rhs.value &&
+                  _lhs.units     == _rhs.units;
+        return eq;
+    }
+
+    namespace tokens {
+        const std::string none{};
```
korydraughn, line 41: `none` is likely not needed

jasoncoposky: this mirrors the entity type enumeration in order to model the negative space.

korydraughn: true, but looking at the impl, it appears that this value will never be encountered or used.  when might someone see this value?

lib/core/include/metadata.hpp
```diff
@@ -134,9 +133,7 @@ namespace irods::experimental::metadata {
     } // set
 
 
-    /// \returns xyz
-    /// \retval xyz
-    /// \brief xyz
+    /// \brief Add a metadata triple to a given entity type without the possibly of an overwrite
```
trel, line 136: possibility

trel:
with typo fix - i think we're ready for the squashing

trel:
looks good - please to # it

PR #5106, opened 2020-09-08
Replica library and new proxy objects (master)

alanking:
CI tests are still running, but expected to pass.

Two things I'd like to bring attention to:

1. The TODO in lib/core/include/replica.hpp about making the query into an API plugin. I have another implementation with this separated out in a different branch but decided to exclude it in this PR so that other work can move forward.
2. The use of `RODS_SERVER` macro in the replica library to determine includes and the presence of `IRODS_FILESYSTEM_ENABLE_SERVER_SIDE_API`. I didn't know of any other way to go about this other than to introduce another "enable server-side API" #define for the replica library. Please advise.


lib/core/include/data_object_proxy.hpp
```diff
@@ -0,0 +1,212 @@
...
+        /// \brief Set the collection id for the data object (i.e. all replicas)
+        /// \param[in] _cid - New value for collection id
+        /// \returns void
+        /// \since 4.2.9
+        template<
+            typename P = doi_type,
+            typename = std::enable_if_t<!std::is_const_v<P>>>
+        auto collection_id(const rodsLong_t _cid) -> void 
+        {
+            // TODO: should this set the logical path too?
```
trel, line 113: i think it must?    what does it mean not to?

lib/core/include/data_object_proxy.hpp
```diff
@@ -0,0 +1,212 @@
...
+        /// \brief Set the logical path for the data object (i.e. all replicas)
+        /// \param[in] _lp - New value for logical path
+        /// \returns void
+        /// \since 4.2.9
+        template<
+            typename P = doi_type,
+            typename = std::enable_if_t<!std::is_const_v<P>>>
+        auto logical_path(std::string_view _lp) -> void
+        {
+            // TODO: should this set the collId too?
```
trel, line 128: i think it must?    what does it mean not to?

lib/core/include/data_object_proxy.hpp
```diff
@@ -0,0 +1,212 @@
...
+
+        /// \brief Generates replica proxy objects from doi_type's linked list and stores them in a list
+        /// \since 4.2.9
+        static auto fill_replica_list(doi_type& _doi) -> replica_list
+        {
+            replica_list r;
+            for (doi_type* d = &_doi; d; d = d->next) {
+                r.push_back(replica::replica_proxy{*d});
+            }
+            // TODO: do we allow empty replica list?
```
trel, line 195: no?    what's the use case for that?    can anyone hold it wrong?

alanking: Hmmm... you're right. I guess it doesn't make any sense - it has to be at least 1 replica long as the very existence of a `dataObjInfo_t` suggests a replica. I will remove this

lib/core/include/data_object_proxy.hpp
alanking: Well, the setting of the logical path via change in collection_id is necessarily a hit to the database in order to resolve the ID to a logical path so that we can set it (unless we have a resource_manager-esque entity for collections that I do not know about). Are we willing to take that hit, or is it up to the caller (or a separate member function) to do this if desired?

lib/core/include/data_object_proxy.hpp
alanking: See comment in collection_id above

lib/core/include/data_object_proxy.hpp
trel: is there any upside to allowing someone to get the database 'out of sync' in this way?   i can only think of how this can be bad.

lib/filesystem/src/filesystem.cpp
```diff
@@ -559,25 +546,6 @@ namespace irods::experimental::filesystem::NAMESPACE_IMPL
...
-            throw filesystem_error{"path does not exist", _p, make_error_code(OBJ_PATH_DOES_NOT_EXIST)};
-        }
-
-        if (s.type == DATA_OBJ_T) {
-            return static_cast<std::uintmax_t>(s.size);
-        }
-
-        return 0;
-    }
-
```
korydraughn, line 580: this function's implementation will change in the future, so this can stay here for now.

lib/core/include/data_object_proxy.hpp
alanking: If this does not set the logical path (or if setting the logical path does not set the collection id), one or the other is out of sync within this data model. We plan to use this structure to generate r_data_main updates which will be applied in the finalize API. As collection name is not stored in the catalog, I think this is okay. The only time this would be happening would be in a rename operation anyway, so the physical path would need to be updated for each replica as well.

After some more thinking, I feel this is the responsibility of a higher-level call, whether in the caller or in a convenience function. Thoughts?

lib/filesystem/include/filesystem/path.hpp
```diff
@@ -383,6 +384,8 @@ namespace irods::experimental::filesystem
     inline auto swap(path& _lhs, path& _rhs) noexcept -> void { _lhs.swap(_rhs); }
 
     auto hash_value(const path& _p) noexcept -> std::size_t;
+
+    auto get_zone_name(const path& _p) -> std::optional<std::string>;
```
korydraughn, line 388: maybe this gets renamed to just `zone_name` to fit in with the other functions around it.

lib/filesystem/src/filesystem.cpp
```diff
@@ -592,15 +560,11 @@ namespace irods::experimental::filesystem::NAMESPACE_IMPL
     {
         const auto s = status(_comm, _p);
 
-        if (is_data_object(s)) {
-            return data_object_size(_comm, _p) == 0;
-        }
-
```
korydraughn, line 598: with the changes coming for `data_object_size`, this code can be restored.

lib/filesystem/src/filesystem.cpp
```diff
@@ -592,15 +560,11 @@ namespace irods::experimental::filesystem::NAMESPACE_IMPL
...
         if (is_collection(s)) {
             return is_collection_empty(_comm, _p);
         }
 
-        throw filesystem_error{"cannot check emptiness: unknown object type", _p, make_error_code(CAT_NOT_A_DATAOBJ_AND_NOT_A_COLLECTION)};
+        throw filesystem_error{"cannot check emptiness: object type unsupported", _p, make_error_code(CAT_NOT_A_DATAOBJ_AND_NOT_A_COLLECTION)};
```
korydraughn, line 567: there are other places in this file where **unknown object type** is used.  either the other references need to be updated to use the same phrase or this should be restored to fall in line with the others.

lib/filesystem/src/filesystem.cpp
```diff
@@ -627,7 +591,7 @@ namespace irods::experimental::filesystem::NAMESPACE_IMPL
     {
         const auto s = stat(_comm, _p);
 
-        if (s.error < 0 || s.type == UNKNOWN_OBJ_T) {
+        if (s.error < 0 || UNKNOWN_OBJ_T == s.type || DATA_OBJ_T == s.type) {
```
korydraughn, line 594: this can be restored. this function will be fixed in a later PR.

lib/filesystem/src/path.cpp
```diff
@@ -438,5 +438,19 @@ namespace irods::experimental::filesystem
         const auto pstr = _p.string();
         return boost::hash_range(std::begin(pstr), std::end(pstr));
     }
+
+    auto get_zone_name(const path& _p) -> std::optional<std::string>
```
korydraughn, line 442: consider renaming to `zone_name` to fit with rest of API.

unit_tests/src/filesystem/test_filesystem.cpp
```diff
@@ -157,21 +157,6 @@ TEST_CASE("filesystem")
...
-            default_transport tp{conn};
-            odstream{tp, p} << "hello world!";
-        }
-
-        REQUIRE(fs::client::exists(conn, p));
-        REQUIRE(fs::client::data_object_size(conn, p) == 12);
-        REQUIRE(fs::client::data_object_checksum(conn, p, fs::replica_number::all).size() == 1);
-        REQUIRE(fs::client::remove(conn, p, fs::remove_options::no_trash));
-    }
-
```
korydraughn, line 174: this can be restored (minus the checksum call).

lib/core/include/replica.hpp
```diff
@@ -0,0 +1,312 @@
...
+
+#include <chrono>
+#include <iomanip>
+#include <string_view>
+
+namespace irods::experimental::replica
+{
+    namespace {
+        namespace fs = irods::experimental::filesystem;
+    } // anonymous namespace
```
korydraughn, line 30: it is probably best if anonymous namespaces are not used in headers.  i believe for every translation unit that includes this header, this namespace will result in a new symbol.

prefer using the fully qualified names for headers.  there is no good way around this as far as i know.

lib/core/include/replica.hpp
```diff
@@ -0,0 +1,312 @@
...
+    ///
+    /// \since 4.2.9
+    template<typename rxComm>
+    auto replica_size(rxComm& _comm, const fs::path& _p, const replica_number_type _rn) -> std::uintmax_t
+    {
+        detail::throw_if_replica_input_is_invalid(_comm, _p, _rn);
+
+        const auto result = get_replica_info(_comm, _p, _rn);
+
+        return static_cast<std::uintmax_t>(std::stoull(result[6])); // 6 == index of DATA_SIZE column from SELECT statement
```
korydraughn, line 168: the `6` could use a name here.  see my previous comment.

lib/filesystem/src/path.cpp
```diff
@@ -438,5 +438,19 @@ namespace irods::experimental::filesystem
...
+
+        auto iter = std::begin(_p);
+
+        if (constexpr difference_type hops = 2; std::distance(iter, std::end(_p)) >= hops) {
+            return *++iter;
+        }
+        
+        return std::nullopt;
+    }
+
```
korydraughn, line 454: remove whitespace

unit_tests/src/test_replica.cpp
```diff
@@ -0,0 +1,87 @@
...
+namespace fs = irods::experimental::filesystem;
+namespace io = irods::experimental::io;
+namespace replica = irods::experimental::replica;
+
+TEST_CASE("replica", "[replica]")
+{
+    load_client_api_plugins();
+
+    auto conn_pool = irods::make_connection_pool();
+    auto conn = static_cast<rcComm_t>(conn_pool->get_connection());
```
korydraughn, line 24: this results in a copy (i think).  im not entirely sure what this means for the connection proxy and pool.

change to `rcComm_t&`

lib/filesystem/include/filesystem/collection_entry.hpp
```diff
@@ -28,7 +28,6 @@ namespace irods::experimental::filesystem
         auto is_data_object() const noexcept -> bool              { return filesystem::NAMESPACE_IMPL::is_data_object(status_); }
         auto is_collection() const noexcept -> bool               { return filesystem::NAMESPACE_IMPL::is_collection(status_); }
         auto is_other() const noexcept -> bool                    { return filesystem::NAMESPACE_IMPL::is_other(status_); }
-        auto data_object_size() const noexcept -> std::uintmax_t  { return data_size_; }
```
korydraughn, line 31: this can be restored.  will fix in a later PR.

korydraughn (commented):
In regards to the comments about restoring things, I'm not sure if it is worth actually doing that (we have git :) ). The plan is to make those functions fetch the latest replica's information. That way, the filesystem lib can still provide information as it relates to the logical layer.

lib/core/include/data_object_proxy.hpp
korydraughn: I agree.  This sounds like the responsibility of another API.

Is it the proxy's responsibility to keep things in sync or is it something that simply provides a better interface to the C type?

lib/core/include/replica.hpp
```diff
@@ -0,0 +1,312 @@
...
+        } // get_replica_info
+
+        template<typename rxComm>
+        auto replica_mtime(
+            rxComm& _comm,
+            const fs::path& _p,
+            const replica_number_type _rn) -> std::string
+        {
+            const auto result = get_replica_info(_comm, _p, _rn);
+            return result[18]; // 18 == index of MODIFY_TIME column from SELECT statement
```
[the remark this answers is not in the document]
trel: yes please

lib/core/include/data_object_proxy.hpp
```diff
@@ -0,0 +1,212 @@
...
+
+        /// \returns size_type
+        /// \retval Number of replicas for this data object
+        /// \since 4.2.9
+        auto replica_count() const -> size_type { return r_.size(); }
+
+        /// \returns const replica_list&
+        /// \retval Reference to the list of replica_proxy objects
+        /// \since 4.2.9
+        auto replicas() const -> const replica_list& { return r_; }
```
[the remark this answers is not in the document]
alanking: The `doi_pointer_type` will never be `nullptr`, so the replica list will never be empty. The members of the underlying struct are not guaranteed to be populated if a zero-initialized struct was used to construct the proxy object. I'll perform an audit of these things and mark `noexcept` where appropriate.

lib/core/include/data_object_proxy.hpp
```diff
@@ -0,0 +1,212 @@
...
+            return r;
+        } // fill_replica_list
+    }; // data_object_proxy
+
+    /// \brief Wraps an existing doi_type with a proxy object.
+    /// \param[in] _doi - Pre-existing doi_type which will be wrapped by the returned proxy.
+    /// \return data_object_proxy
+    /// \since 4.2.9
+    template<typename doi_type>
+    static auto make_data_object_proxy(doi_type& _doi) -> data_object_proxy<doi_type>
```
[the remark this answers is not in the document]
alanking: I will look into this. Currently, this compiles, but inlining it is probably the right call regardless.

lib/core/include/replica.hpp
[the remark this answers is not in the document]
alanking: I can look into this. In the other branch I mentioned, I am constructing a map of the result set using the genquery column names as keys so that they can be referred to directly, regardless of ordering. Thinking on it further, I think having the indexes laid out in a SPOT may just be better/easier to manage overall anyway.

lib/core/include/replica_proxy.hpp
```diff
@@ -0,0 +1,482 @@
...
+        /// \retval specColl pointer for the dataObjInfo_t node
+        /// \since 4.2.9
+        auto special_collection_info() const -> const specColl_t*
+        {
+            return doi_->specColl;
+        }
+
+        /// \returns std::string_view
+        /// \since 4.2.9
+        auto sub_path()         const -> std::string_view { return doi_->subPath; }
```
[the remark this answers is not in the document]
alanking: I had a comment at the beginning of this section that I removed but I have no idea what a lot of these members are for (including expiry_ts). So I just named them what they are in r_data_main so as to not provide incorrect definitions.

I am happy to name them something else or provide documentation, but I am not aware of what anything is from sub_path to resource.

lib/filesystem/include/filesystem/filesystem.hpp
```diff
@@ -90,8 +70,6 @@ namespace irods::experimental::filesystem
 
         auto equivalent(rxComm& _comm, const path& _p1, const path& _p2) -> bool;
 
-        auto data_object_size(rxComm& _comm, const path& _p) -> std::uintmax_t;
```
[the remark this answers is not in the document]
alanking: I was not aware that this was the plan. Can you point me to an issue number so I can leave some TODOs or somesuch? Or am I expected to go ahead and implement this? Either way is fine, just making sure.

lib/filesystem/src/filesystem.cpp
alanking: In this case, this code will be restored along with the data_object_size family of functionality, so I will just revert these pieces of the change.

lib/filesystem/include/filesystem/filesystem.hpp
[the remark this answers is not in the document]
korydraughn: There is no issue at this time. It was something I was planning on looking into. I think you should continue with your current impl. Ignore my comments about **restoring** removed code. That can be revisited after your PR is merged.

lib/core/include/data_object_proxy.hpp
alanking: I think if the need ever arises, a new member function can be introduced to keep everything aligned. For now, it won't be concerned with keeping things in sync to keep things simple and the caller will need to make sure that the information is correct if it is going to be updating the catalog with information from these structs.

trel: seems fair.

lib/core/include/replica_proxy.hpp
```diff
@@ -0,0 +1,482 @@
...
+        } // set_string_property
+
+    }; // class replica_proxy
+
+    /// \brief Factory method for making a replica proxy object for a dataObjInfo_t
+    /// \param[in] _doi - Pre-existing doi_type which will be wrapped by the returned proxy.
+    /// \return replica_proxy
+    /// \since 4.2.9
+    template<typename doi_type>
+    static auto make_replica_proxy(doi_type& _doi) -> replica_proxy<doi_type>
```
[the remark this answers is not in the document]
alanking: It does. This is how make_key_value_proxy was implemented as well.

lib/core/include/data_object_proxy.hpp
alanking: After offline discussion, we have determined both member functions should be replaced by a single `rename` function which updates both the collection id and the logical path as provided by the caller so that the ORM cannot be out of sync.

unit_tests/src/test_replica.cpp
```diff
@@ -21,7 +21,7 @@ TEST_CASE("replica", "[replica]")
     load_client_api_plugins();
 
     auto conn_pool = irods::make_connection_pool();
-    auto conn = static_cast<rcComm_t>(conn_pool->get_connection());
+    auto conn = static_cast<rcComm_t&>(conn_pool->get_connection());
```
korydraughn, line 24: This is actually a bug and needs to be adjusted.  With the current impl, the connection is being returned back to the pool, but it is still being used throughout the test.  This is fine because there is only one connection being used and it is synchronous, but its usage is incorrect.

Here is an alternative.
```c++
auto conn = conn_pool->get_connection();
rcComm_t& conn_ref = conn;
```

alanking:
CI tests pass except for some known unrelated failures

trel:
ready to see squashed version - then can # after confirmation.

trel:
please # when ready.

alanking:
#'d

Issue #5107, opened 2020-09-08
istream does not allow writing to existing replicas via the replica number


korydraughn:
- [x] master
- [x] 4-2-stable
---
## Bug Report
`istream` should allow replica numbers to be used when writing to existing replicas.  `istream` will silently write to replica zero if a replica number is provided.

PR #5109, opened 2020-09-09
[#5107] Added test to verify istream supports replica numbers when writing (4-2-stable)

korydraughn:


korydraughn:
test added.

added #

PR #5130, opened 2020-09-13
Collection mtimes, library improvements, and more documentation (master) 

korydraughn:


lib/filesystem/src/filesystem.cpp
```diff
@@ -546,6 +536,58 @@ namespace irods::experimental::filesystem::NAMESPACE_IMPL
...
+        if (query.size() == 0) {
+            throw filesystem_error{"no good replica found", _p, make_error_code(SYS_NO_GOOD_REPLICA)};
+        }
+
+        // This implementation assumes that any good replica will always satisfy
+        // the requirement within the loop, therefore the first iteration always causes
+        // the size to be captured. The size object should be empty if and only if there
+        // are no good replicas.
+        std::uint64_t latest_mtime = 0;
+        std::uintmax_t size = 0;
```
korydraughn, line 577: this probably needs to be wrapped in a `std::optional` because the query could result in no good replicas.

korydraughn: instead of using `std::optional`, the query object should be checked to see if any results exist and act appropriately.

lib/filesystem/src/filesystem.cpp
```diff
@@ -848,6 +902,52 @@ namespace irods::experimental::filesystem::NAMESPACE_IMPL
...
+        }
+
+        std::uintmax_t latest_mtime = 0;
+        std::string checksum;
+
+        // This implementation assumes that any good replica will always satisfy
+        // the requirement within the loop, therefore the first iteration always causes
+        // the checksum to be captured. The checksum object should be empty if and only if there
+        // are no good replicas.
+        for (const auto& row : qb.build(_comm, gql)) {
```
korydraughn, line 939: see comment about `data_object_size`

server/api/src/rsDataObjRename.cpp
```diff
@@ -258,142 +264,174 @@ namespace {
...
+        const auto src_parent_path = fs::path{dataObjRenameInp->srcDataObjInp.objPath}.parent_path();
+        const auto dst_parent_path = fs::path{dataObjRenameInp->destDataObjInp.objPath}.parent_path();
+
+        using std::chrono::system_clock;
+        using std::chrono::time_point_cast;
+
+        const auto mtime = time_point_cast<fs::object_time_type::duration>(system_clock::now());
+
+        try {
+            fs::server::last_write_time(*rsComm, src_parent_path, mtime);
```
alanking, line 426: Very minor thought: Would it be worth (perhaps in the future) having a convenience/helper method to set the last_write_time to now?

Also, the commit message should probably read: "Parent collection's mtime now updated when contents are modified." or something to that effect.

lib/filesystem/src/filesystem.cpp
```diff
@@ -859,6 +858,52 @@ namespace irods::experimental::filesystem::NAMESPACE_IMPL
...
+        }
+
+        detail::throw_if_path_length_exceeds_limit(_p);
+
+        if (!is_data_object(_comm, _p)) {
+            throw filesystem_error{"path does not point to a data object", _p, make_error_code(SYS_INVALID_INPUT_PARAM)};
+        }
+
+        // TODO Determine whether "DATA_REPL_NUM" is needed here.
+        const auto gql = fmt::format("select DATA_CHECKSUM, DATA_REPL_NUM, DATA_MODIFY_TIME "
```
alanking, line 918: Is DATA_REPL_NUM used anywhere? Looks like you could get away with just selecting DATA_CHECKSUM and DATA_MODIFY_TIME

unit_tests/src/filesystem/test_filesystem.cpp
```diff
@@ -180,6 +264,81 @@ TEST_CASE("filesystem")
...
+
+        const fs::path p = sandbox / "data_object";
+
+        {
+            irods::experimental::client_connection conn;
+            default_transport tp{conn};
+            odstream{tp, p} << "hello world!";
+        }
+
+        const auto first_replica_mtime = fs::client::last_write_time(conn, p);
```
alanking, line 279: Would it be worth adding a test for the appropriate filesystem error for attempting to modify the last_write_time of a data object? I removed the test when the function was moved to the replica library, but maybe good for completeness.

alanking (commented):
Looks good overall. A few minor comments and some cringing wrt future rebasing pain ;)

unit_tests/src/filesystem/test_filesystem.cpp
korydraughn: sure.  will add that.

unit_tests/src/filesystem/test_filesystem.cpp
```diff
@@ -180,6 +264,81 @@ TEST_CASE("filesystem")
...
+        std::this_thread::sleep_for(2s);
+
+        namespace adm = irods::experimental::administration;
+
+        // Get hostname so that we can create a new resource.
+        char host_name[64] {};
```
trel, line 285: same with 64 here.

unit_tests/src/filesystem/test_filesystem.cpp
```diff
@@ -180,6 +264,81 @@ TEST_CASE("filesystem")
...
+        const fs::path p = sandbox / "data_object";
+
+        {
+            irods::experimental::client_connection conn;
+            default_transport tp{conn};
+            odstream{tp, p} << "hello world!";
+        }
+
+        const auto first_replica_mtime = fs::client::last_write_time(conn, p);
+        std::this_thread::sleep_for(2s);
```
trel, line 280: why 2s?

unit_tests/src/filesystem/test_filesystem.cpp
```diff
@@ -180,6 +264,81 @@ TEST_CASE("filesystem")
...
+
+        {
+            // Create a new resource.
+            irods::experimental::client_connection conn;
+            REQUIRE(adm::client::add_resource(conn, ufs_info).value() == 0);
+        }
+
+        {
+            // Replicate the existing data object and verify that the time
+            // returned comes from the most up-to-date good replica.
```
trel, line 308: most 'recent'.... they're all up-to-date?

lib/filesystem/src/filesystem.cpp
korydraughn: i don't think it is used.  i'm hoping i'm able to remove that column once this PR is complete.

trel (commented):
i saw a few places with very similar code around updating a collection mtime... could possibly be pulled out into a cleaner function call?

server/api/src/rsDataObjRename.cpp
korydraughn: that idea has crossed my mind and i'm not sure.  maybe a different function that emphasizes what it is doing?  i don't think `last_write_time` would work given that we can't default the last argument.

i'll update the commit message to be clearer.

unit_tests/src/filesystem/test_filesystem.cpp
```diff
@@ -157,7 +166,80 @@ TEST_CASE("filesystem")
...
+        }
+
+        REQUIRE(fs::client::exists(conn, p));
+        REQUIRE(fs::client::data_object_size(conn, p) == 12);
+        REQUIRE(fs::client::data_object_checksum(conn, p) == "");
+
+        namespace adm = irods::experimental::administration;
+
+        // Get hostname so that we can create a new resource.
+        char host_name[64] {};
```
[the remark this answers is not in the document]
korydraughn: No, i just chose a size i felt covers most host name lengths.

unit_tests/src/filesystem/test_filesystem.cpp
korydraughn: it is using std::chrono's user-defined literals to denote 2 seconds. the "s" makes this equivalent to `std::chrono::seconds{2}`. does that answer the question?

unit_tests/src/filesystem/test_filesystem.cpp
korydraughn: yes, but i think the mtimes are different so the mtime of the second replica will be returned.

lib/filesystem/src/filesystem.cpp
```diff
@@ -848,6 +902,52 @@ namespace irods::experimental::filesystem::NAMESPACE_IMPL
...
+            throw filesystem_error{"empty path", make_error_code(SYS_INVALID_INPUT_PARAM)};
+        }
+
+        detail::throw_if_path_length_exceeds_limit(_p);
+
+        if (!is_data_object(_comm, _p)) {
+            throw filesystem_error{"path does not point to a data object", _p, make_error_code(SYS_INVALID_INPUT_PARAM)};
+        }
+
+        // TODO Determine whether "DATA_REPL_NUM" is needed here.
```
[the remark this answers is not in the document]
korydraughn: yes

lib/filesystem/src/filesystem.cpp
```diff
@@ -546,6 +536,58 @@ namespace irods::experimental::filesystem::NAMESPACE_IMPL
...
```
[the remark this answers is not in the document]
korydraughn: yes

unit_tests/src/filesystem/test_filesystem.cpp
korydraughn: added test.

korydraughn:
> i saw a few places with very similar code around updating a collection mtime... could possibly be pulled out into a cleaner function call?

@trel Can you point out the code you were referring to?

server/api/src/rsDataObjRename.cpp
korydraughn: After coming back and reading this again, I do think convenience functions would be helpful. `std::chrono` is not the easiest library to use when just starting out.

Where would these functions live?

I can imagine having functions that do the following:
- One that returns the current time in various formats.
- One that converts a string (or other type) to an iRODS time type and vice versa.

This should probably have its own issue.

Thoughts on any of that?

trel:
I see six instances of '// Update the parent collection's mtime.' in this diff - seemed plausible there is a way to do it more concisely.  Definitely not a blocker.

server/api/src/rsDataObjRename.cpp
trel: a generic irods::time or irods::chrono thinger might be useful - but willing to wait for that.  can be a 'later' thing, no problem.

korydraughn:
Okay. I'll review those and see if I can generalize that code more. It should be possible.

korydraughn:
If the tests pass, then I may let the current PR be merged and leave the generalization of that mtime code to a later PR.

trel:
definitely PR is good as is - perhaps a new issue for consolidation of time stuffs.

korydraughn:
tests passed.

Added #

PR #5147, opened 2020-09-17
New API Plugin and Microservice for updating mtimes (master)

korydraughn:
Bundled with irods/irods_client_icommands#152


plugins/api/src/touch.cpp
```diff
@@ -0,0 +1,512 @@
...
+ {
+     // Must be an absolute path.
+     // Cannot be empty.
+     "logical_path": string, 
+
+     // TODO It would be great if we can get rid of this.
+     // Doing so may require a slight change in the API.
+     //
+     // Cannot be empty.
+     "default_resource": string,
```
trel, line 54: yes, agreed, no default determination magic should be happening in this API call.

plugins/api/src/touch.cpp
```diff
@@ -0,0 +1,544 @@
...
+                THROW(OBJ_PATH_DOES_NOT_EXIST, "Cannot update modification time (object does not exist).");
+            }
+
+            if (opts.contains(prop_replica_number)) {
+                THROW(USER_INCOMPATIBLE_PARAMS, "Replica numbers cannot be used when creating new data objects.");
+            }
+
+            // Create the data object at the leaf resource if the user provided one.
+            // This could be explicitly provided by the client (i.e. itouch -R <resc>) or
+            // implicitly if a default resource is defined in the client's irods_environment.json file.
```
trel, line 371: comment should now talk about the server's default, not the client's.

plugins/api/src/touch.cpp
```diff
@@ -0,0 +1,541 @@
...
+        // Get ready to create a new data object.
+        io::server::default_transport tp{_comm};
+
+        if (_json_input.contains(prop_options)) {
+            const auto& opts = _json_input.at(prop_options.data());
+
+            // Return immediately if the object does not exist and the user specified to not
+            // create any new data objects.
+            if (opts.contains(prop_no_create) && opts.at(prop_no_create.data()).get<bool>()) {
+                // TODO Use better error code.
```
trel, line 358: `USER_INCOMPATIBLE_PARAMS`?

korydraughn: maybe this shouldn't be considered an error at all. `touch --no-create does_not_exist.txt` returns `0` which seems like the right thing to do.

trel: ah, very well.

`0` it is - no work was requested, and no work was done.

trel:
let's # it

korydraughn:
Added #

PR #5148, opened 2020-09-17
New API Plugin and Microservice for updating mtimes (4-2-stable)

korydraughn:
Bundled with irods/irods_client_icommands#153

This PR requires https://github.com/irods/irods/pull/5158 to be merged.


scripts/irods/test/test_itouch.py
```diff
@@ -0,0 +1,216 @@
...
+            # original replica.
+            sleep(2) 
+
+            # Replicate the original replica and show that their mtimes are different.
+            self.admin.assert_icommand(['irepl', '-R', resc_0, data_object])
+            mtime = self.get_replica_mtime(data_object, 0)
+            self.assertNotEqual(mtime, self.get_replica_mtime(data_object, 1))
+
+            # Make the new replica have the same mtime as the original.
+            self.admin.assert_icommand(['itouch', '-n1', '-s', str(mtime), data_object])
```
trel, line 100: not wrong, but let's use `'-n', '1',`

scripts/irods/test/test_itouch.py
```diff
@@ -0,0 +1,216 @@
...
+
+            # Clean up.
+            self.admin.assert_icommand(['irm', '-f', data_object])
+        except Exception as e:
+            print(e)
+        finally:
+            self.admin.assert_icommand(['iadmin', 'rmresc', resc_0])
+
+    @unittest.skipIf(test.settings.RUN_IN_TOPOLOGY, "Skip for Topology Testing")
+    def test_updates_mtime_of_replica_using_leaf_resource_option(self):
```
trel, line 111: consider putting `resc_0` under a `passthru` to truly test the `leaf`yiness

scripts/irods/test/test_itouch.py
```diff
@@ -0,0 +1,216 @@
...
+            self.assertEqual(self.get_replica_mtime(data_object_1, 0), mtime)
+        except Exception as e:
+            print(e)
+
+    @unittest.skipIf(test.settings.RUN_IN_TOPOLOGY, "Skip for Topology Testing")
+    def test_seconds_since_epoch_is_honored(self):
+        data_object = 'foo'
+
+        # 584582400 seconds since epoch => 1988-07-11 00:00:00.0 UTC
+        #                               => 1988-07-10 20:00:00.0 Local Time
```
trel, line 169: this test will only pass in the eastern time zone during daylight saving time?

scripts/irods/test/test_itouch.py
```diff
@@ -0,0 +1,216 @@
...
+            sleep(2)
+            self.admin.assert_icommand(['itouch', self.admin.session_collection])
+            self.assertNotEqual(old_mtime, self.get_collection_mtime(self.admin.session_collection))
+        except Exception as e:
+            print(e)
+
+    @unittest.skipIf(test.settings.RUN_IN_TOPOLOGY, "Skip for Topology Testing")
+    def test_mutually_exclusive_options_are_handled(self):
+        error_msg = 'Error: -n and -R cannot be used together.'
+        self.admin.assert_icommand(['itouch', '-n0', '-R', 'demoResc', 'foo'], 'STDERR', [error_msg])
```
trel, line 186: consider splitting -n0

scripts/irods/test/test_itouch.py
```diff
@@ -0,0 +1,216 @@
...
+
+        error_msg = 'Error: -r and -s cannot be used together.'
+        self.admin.assert_icommand(['itouch', '-r', 'source', '-s', '1', 'foo'], 'STDERR', [error_msg])
+
+    @unittest.skipIf(test.settings.RUN_IN_TOPOLOGY, "Skip for Topology Testing")
+    def test_replica_number_cannot_be_used_to_create_new_replicas(self):
+        error_msg = 'Error: Replica numbers cannot be used when creating new data objects.'
+        self.admin.assert_icommand(['itouch', '-n3', 'foo'], 'STDERR', [error_msg])
```
trel, line 194: consider splitting -n3

scripts/irods/test/test_itouch.py
```diff
@@ -0,0 +1,216 @@
...
+        data_object = os.path.basename(path)
+        mtime, ec, rc = self.admin.run_icommand('iquest %s "select DATA_MODIFY_TIME where COLL_NAME = \'{0}\' and DATA_NAME = \'{1}\' and DATA_REPL_NUM = \'{2}\'"'
+            .format(collection, data_object, replica_number))
+        return int(mtime)
+
+    def get_collection_mtime(self, path):
+        mtime, ec, rc = self.admin.run_icommand('iquest %s "select COLL_MODIFY_TIME where COLL_NAME = \'{0}\'"'.format(path))
+        return int(mtime)
+
+    def create_resource(self, resource_name):
```
trel, line 209: this is fine... but we may already have a helper for this - not sure...

trel:
please add a test for each:
 - touching a non-existent replica by resource name
 - touching a non-existent replica by replica number

korydraughn:
i think the second list item may be covered by `test_replica_number_cannot_be_used_to_create_new_replicas` unless you are talking about the case where the data object exists, but the target replica does not.

can you explain these tests in a little more detail?

trel:
yes, both are for data objects that already exist and have replicas... but the requested replica (either by number or by resource name) does not exist.

korydraughn:
the replica number test makes sense, replica numbers cannot create new replicas.

but what about the leaf resource option.  i want to say that istream allows users to create new replicas using that option.  do we want to keep the same behavior?  seems so to me.

trel:
hmm... so if you touch a new replica by resource name for an existing data object... it would be empty (0-bytes) and marked good (&), which suggests it is the new correct replica because the other replicas are now marked stale...

that seems... not what we want.

I think touching a replica that DOES NOT exist for a data object that DOES exist... should be an ERROR

scripts/irods/test/test_itouch.py
korydraughn: i'm going to leave it for now.  the tests are passing and i don't want to hold this up at this time.

trel:
looks good

trel:
#5158 is merged - please rebase when ready

korydraughn:
Added #

PR #5154, opened 2020-09-18
Catalog/Replica/Proxy library enhancements, Data Object Finalize API, tilling for logical locking (master)

alanking:
CI tests running

PR #5161, opened 2020-09-23
Fix rsPhyPathReg for create + lifetime_manager tweaks + suppress some non-error log messages (master)

alanking:
CI tests are running - previously failing tests passed at the bench.

Some notes here...

1. rsPhyPathReg must now return the dataObjInfo_t of the replica that was just registered in the catalog. As such, some of the free's have been removed. This is not leaking these structs, but rather transfers ownership to the global L1 descriptor table, which is cleaned up on agent exit. Otherwise, the lifetime_manager provided by the replica_proxy factory method cleans up the dataObjInfo_t.

2. The shift to returning pointers in the get() and release() methods of the lifetime_manager is consistent with the proxy objects' get() methods. This was needed in order to prevent freeing the dataObjInfo_t generated in rsPhyPathReg.

### DIFF
--- a/server/api/src/rsPhyPathReg.cpp
+++ b/server/api/src/rsPhyPathReg.cpp
@@ -1,8 +1,3 @@
-/*** Copyright (c), The Regents of the University of California            ***
- *** For more information please refer to files in the COPYRIGHT directory ***/
-/* This is script-generated code (for the most part).  */
-/* See phyPathReg.h for a description of this API call.*/
-
 #include "fileStat.h"
 #include "key_value_proxy.hpp"
 #include "phyPathReg.h"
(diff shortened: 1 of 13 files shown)


server/api/src/rsPhyPathReg.cpp
```diff
@@ -482,7 +508,7 @@ namespace {
             fileStatInp_t fileStatInp;
             memset( &fileStatInp, 0, sizeof( fileStatInp ) );
 
-            // Issue #3658 - This section removes trailing slashes from the 
+            // Issue #3658 - This section removes trailing slashes from the
             // directory path in the server when the path is already in the catalog.
             //
             // TODO:  This is a localized fix that addresses any trailing slashes
```
trel, line 514: did we find a better/existing way to do this work?

trel:
looks good to me - please # when ready

server/api/src/rsPhyPathReg.cpp
alanking: Yes, a suggestion was made here: https://github.com/irods/irods/pull/5154#discussion_r491771767

I can fix this if desired, just not clear how to test other than running the suite again

trel: Nope, no worries - just looked familiar.   Carry on.

trel: Please make a new issue so we don't lose the suggestion

alanking:
Tests passed

alanking:
Actually, there was one failure, which had to do with the KEY_VALUE_PASSTHROUGH_KW test in Test_Resource_Unixfilesystem. Hold, please...

trel:
and codacy has a single thing.

alanking:
Fixed failing test and addressed Codacy issue.

Issue #5170, opened 2020-10-02
zonereport does not include resc_id information


trel:
- [x] master
- [x] 4-2-stable
-------

### iRODS Version, OS and Version
4.2.8

### What did you try to do?

izonereport - and then construct the resource hierarchies from the included information

### Expected behavior

izonereport should provide `resc_id` information for each named/included resource

### Observed behavior (including steps to reproduce, if applicable)

`parent_id` is provided, but no `resc_id` - so it's impossible to know how to build the tree(s)

SwooshyCueb:
Just to make sure I'm on the same page, `parent_id` is currently provided as `parent_resource`, yes?

trel:
that is correct - in 4.1, all resources were communicated by string name, and in 4.2, we moved to integers for a variety of performance and consistency reasons.

we now provide the id, rather than the names when populating the zonereport.

SwooshyCueb:
#5173 is merged, master can be checked off

trel:
got it

PR #5173, opened 2020-10-06
[#5170] Add id to resource json serialization (master)

SwooshyCueb:
This is meant to address #5170. I'll make another pull request for the `4-2-stable` branch once this one has passed review.

If a test case is required, I may need some direction. I've tried adding one to `test_resource_administration.cpp`, but the resources added with `adm::client` do not show up in the output from `rcServerReport`.

### DIFF
--- a/scripts/irods/test/test_izonereport.py
+++ b/scripts/irods/test/test_izonereport.py
@@ -1,0 +1,79 @@
+from __future__ import print_function
+import sys
+if sys.version_info >= (2, 7):
+    import unittest
+else:
+    import unittest2 as unittest
+import os
+import json
+from .. import lib
+from . import session
+from ..test.command import assert_command
+from ..configuration import IrodsConfig
+
+class Test_Izonereport(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        cls.admin = session.mkuser_and_return_session('rodsadmin', 'otherrods', 'rods', lib.get_hostname())
+        cls.admin.assert_icommand(['iadmin', 'mkresc', 'rand_resc', 'random'], 'STDOUT_SINGLELINE', 'rand_resc')
+        cls.admin.assert_icommand(['iadmin', 'mkresc', 'repl_resc', 'replication'], 'STDOUT_SINGLELINE', 'repl_resc')
+        cls.admin.assert_icommand(['iadmin', 'mkresc', 'comp_resc', 'compound'], 'STDOUT_SINGLELINE', 'comp_resc')
+        cls.admin.assert_icommand(['iadmin', 'modresc', 'comp_resc', 'context', 'comp_resc_context'])
+        cls.admin.assert_icommand(['iadmin', 'modresc', 'repl_resc', 'context', 'repl_resc_context'])
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.admin.assert_icommand(['iadmin', 'rmresc', 'rand_resc'])
+        cls.admin.assert_icommand(['iadmin', 'rmresc', 'repl_resc'])
+        cls.admin.assert_icommand(['iadmin', 'rmresc', 'comp_resc'])
+        with session.make_session_for_existing_admin() as admin_session:
+            cls.admin.__exit__()
+            admin_session.assert_icommand(['iadmin', 'rmuser', cls.admin.username])
+
+    def test_izonereport_key_sanitization(self):
+        self.admin.assert_icommand("izonereport | grep key | grep -v XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX",
+                                   'STDOUT_SINGLELINE', '"irods_encryption_key_size": 32,', use_unsafe_shell=True)
+
+    def test_izonereport_with_coordinating_resources__ticket_3303(self):
+        _, stdout, _ = self.admin.assert_icommand('izonereport', 'STDOUT_SINGLELINE', 'comp_resc')
+
+        expected_names = [
+            'rand_resc',
+            'repl_resc',
+            'comp_resc'
+        ]
+
+        icat_server_object = json.loads(stdout)['zones'][0]['icat_server']
+        self.assertIn('coordinating_resources', icat_server_object.keys())
+        coord_array = icat_server_object['coordinating_resources']
+        coord_names = map(lambda r : r['name'], coord_array)
+
+        for n in expected_names:
+            self.assertIn(n, coord_names)
+
+    @unittest.skip('FIXME: Remove this line once we figure out why the test fails in ci')
+    def test_izonereport_and_validate(self):
+        jsonschema_installed = True
+        if lib.get_os_distribution() == 'Ubuntu' and lib.get_os_distribution_version_major() == '12':
+            jsonschema_installed = False
+
+        validate_json_path = os.path.join(IrodsConfig().scripts_directory, 'validate_json.py')
+        zone_report = os.path.join(self.admin.local_session_dir, 'out.txt')
+        # bad URL
+        self.admin.assert_icommand("izonereport > %s" % (zone_report), use_unsafe_shell=True)
+        if jsonschema_installed:
+            assert_command('python %s %s https://irods.org/badurl' % (validate_json_path, zone_report), 'STDERR_MULTILINE',
+                               ['WARNING: Validation Failed'], desired_rc=2)
+        else:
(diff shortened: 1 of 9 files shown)


trel:
code looks good to me - yes, let's discuss some tests...

and we're avoiding using the # autolinker until things are nice and tight - because it puts a notification on the issue itself, which can become noisy if done more than once/twice.

SwooshyCueb:
I've added a test to the python test suite that checks the output of `izonereport`. The test assumes that there is at least one zone actually present in the output, but I think that's an okay assumption to make.

scripts/irods/test/test_izonereport.py
```diff
@@ -0,0 +1,32 @@
...
+             admin.assert_icommand(['iadmin', "rmresc", "testResc"])
+             admin.assert_icommand(['iadmin', "rmresc", "anotherResc"])
+
+    def test_resource_json_has_id(self):
+        with session.make_session_for_existing_admin() as admin:
+            _, out, _ = admin.assert_icommand(['izonereport'], 'STDOUT')
+            zone_report = json.loads(out)
+            self.assertGreaterEqual(len(out_j["zones"][0]["icat_server"]["resources"]), 1)
+            for resource in out_j["zones"][0]["icat_server"]["resources"]:
+                self.assertTrue('id' in resource)
```
alanking, line 32: Would it be worth verifying that the IDs are correct? I suppose that is more testing the resource property getter/setter than `izonereport`. Just a thought.

Also, minor: Are we going to be pulling `test_izonereport_with_coordinating_resources__ticket_3303` and `test_izonereport_key_sanitization` into this new suite? I recall talking about that, but maybe I am mistaken.

server/core/src/irods_report_plugins_in_json.cpp
```diff
@@ -97,6 +97,13 @@ namespace irods {
             return PASS(ret);
         }
 
+        rodsLong_t id;
+        ret = _resc->get_property< rodsLong_t >( irods::RESOURCE_ID, id );
+        if ( !ret.ok() ) {
+            return PASS(ret);
+        }
+        std::string id_str = std::to_string( id );
```
alanking, line 105: Could the `std::to_string` call occur on 156 at the time of insertion into the `_entry` map?

alanking (approved):
Looks good overall. A couple of minor comments. Also, the commit message for the test should read "izonereport"

scripts/irods/test/test_izonereport.py
trel: yes, we did.    let's get the izonereport tests into a single well-named spot.

SwooshyCueb: Will do.

In regards to testing the validity of the ids, the c++ test I had written earlier covers this. I recall someone mentioning that the reason I was running into trouble was because I needed to use a new connection to call `rcServerReport`, so I can fix that up and commit it.

server/core/src/irods_report_plugins_in_json.cpp
SwooshyCueb: Sure thing

scripts/irods/test/test_izonereport.py
alanking: Right, like you say, the issue from the C++ tests was that you were re-using the same connection in the administration library calls (which is certainly the intuitive thing to do). Unfortunately, the resource manager is not updated in the connected agent which is why the new resources were not appearing. In the python suite, we are using iCommands which results in a new connection for each call, so it should work if a query or `ilsresc` with some parsing is used.

Hope that helps!

unit_tests/src/test_resource_administration.cpp
```diff
@@ -77,6 +80,33 @@ TEST_CASE("resource administration")
...
+        auto conn_pool = irods::make_connection_pool();
+        auto conn = conn_pool->get_connection();
+
+        auto [ec, info] = adm::client::resource_info(conn, ufs_info.resource_name);
+        REQUIRE(ec.value() == 0);
+        REQUIRE(info->type() == ufs_info.resource_type);
+
+        bytesBuf_t *jbuf;
+        REQUIRE(rcServerReport((rcComm_t*)conn, &jbuf) == 0);
+        std::string report_str((char*)jbuf->buf, jbuf->len);
```
korydraughn, line 96: try to avoid using C style casts as they can introduce bugs.

prefer one of the C++ casting operators.  in this particular case, `static_cast`.

unit_tests/src/test_resource_administration.cpp
```diff
@@ -77,6 +80,33 @@ TEST_CASE("resource administration")
...
+
+        auto report = nlohmann::json::parse(report_str);
```
korydraughn, line 98: this can be marked as `const` since the code only reads from it.

unit_tests/src/test_resource_administration.cpp
```diff
@@ -77,6 +80,33 @@ TEST_CASE("resource administration")
...
+        REQUIRE(report["resources"].size() >= 2);
+        std::string id_str;
+        for (auto &resource : report["resources"]) {
```
korydraughn, line 101: `resource` can be marked `const` as well.  this can help the compiler produce more optimized code.

unit_tests/src/test_resource_administration.cpp
```diff
@@ -77,6 +80,33 @@ TEST_CASE("resource administration")
...
+            std::string name = resource["name"].get<std::string>();
```
korydraughn, line 102: `name` can be marked `const`.

unit_tests/src/test_resource_administration.cpp
```diff
@@ -77,6 +80,33 @@ TEST_CASE("resource administration")
...
+    {
+        // we need a fresh connection becuase the cache is probably out of date
+        auto conn_pool = irods::make_connection_pool();
+        auto conn = conn_pool->get_connection();
+
+        auto [ec, info] = adm::client::resource_info(conn, ufs_info.resource_name);
+        REQUIRE(ec.value() == 0);
+        REQUIRE(info->type() == ufs_info.resource_type);
+
+        bytesBuf_t *jbuf;
```
korydraughn, line 94: prefer initializing your variables (unless you can make certain guarantees).  right now, this could point to anything in memory and cause the program to crash.

you can always use curly braces to zero out primitive types and C structures.  for example:
```c++
bytesBuf_t* bbuf{};  // Equivalent to: bytesBuf_t* bbuf = nullptr;
```

korydraughn (commented):
Looks good to me.

However, the test in the resource administration file should probably be moved to its own file.  The tests in the resource admin file verify that the C++ resource administration library produces the correct results.

The resource id test should have a dedicated file for verifying the behavior of `rcZoneReport`.

unit_tests/src/test_zone_report.cpp
```diff
@@ -0,0 +1,62 @@
...
+    irods::at_scope_exit remove_resources{[&conn, &ufs_info] {
+        adm::client::remove_resource(conn, ufs_info.resource_name);
+    }};
+
+    REQUIRE(adm::client::add_resource(conn, ufs_info).value() == 0);
+
+    // see issue #5170
+    SECTION("verify resource id from json serialization")
+    {
+        // we need a fresh connection becuase the cache is probably out of date
```
trel, line 39: becuase -> because

trel:
please mark comments as resolved as you handle them - this is looking good.

SwooshyCueb:
I realized that the resources created in the setup function for the python test class don't actually show up in the `resources` array in the server report.
I've addressed this by adding some resources that do, and checking the ones that show up in the `coordinating_resources` array as well.

Additionally, I've also changed the test class to use the *class* setup/teardown functions and use a single session object for everything. This improves performance of the test class a bit, though this particular test class doesn't benefit as much from this change as others would, as it's only got four test methods, one of which is always skipped.

Is there anything else that needs to be addressed before I push this through jenkins?

trel:
i think there may be some duplication of resources listed in a multi-server zone - but that could be a separate issue.

consider an [858] on the second commit - otherwise, i think we're good.

SwooshyCueb:
All tests passed in jenkins except the one that doesn't, so I've added the #s

trel:
just caught a typo in the last two commit messages - 'resouce'->'resource'.

tweak and i'll merge.  thanks.

SwooshyCueb:
Whoops. Taken care of.

trel:
checkbox and cherry-pick for 4-2-stable por favor.

PR #5175, opened 2020-10-07
No more REI files! (4-2-stable)

korydraughn:
These code changes do not come with automated tests, but they have been manually tested.

CI Test Results: pending

See [RFC](https://github.com/irods/irods_rfcs/blob/master/0006_static_peps_and_delay_rules.md) for details regarding the REI migration plan.


scripts/setup_irods.py
```diff
@@ -263,7 +265,20 @@ def setup_service_account(irods_config, irods_user, irods_group):
...
     irods_config.clear_cache()
 
-def setup_server_config(irods_config):
+def setup_run_delay_server_configuration_option(irods_config, run_delay_server=None):
+    l = logging.getLogger(__name__)
+    l.debug('Setting up advanced_settings.run_delay_server configuration option ...')
+
+    # Determine if the rule execution (delay) server should be launched on startup.
+    if run_delay_server != None:
+        irods_config.server_config['advanced_settings']['run_delay_server'] = run_delay_server.lower() in ['true', 'y', 'yes', '1']
```
trel, line 274: consider 'True' also...

trel, line 274: parentheses needed/helpful around rhs of this assignment?   same as two lines down?

trel:
this is great!

three things
- haven't added the new database column yet - still reusing `reiFilePath`

- talked about numbering the processed files in the logging, to give more context/scope to human if something goes sideways during the migration

- consider... can this be re-run safely?  what is the status/state of the files/database if only some get migrated cleanly?   a second running will currently... overwrite an earlier attempt's logfile?  consider a timestamp in the log filename to avoid this.

saw a few other TODOs still in there as well.

scripts/setup_irods.py
korydraughn: done

korydraughn:
Added numbering to the migration log.

The migration app is safe to re-run.  The log file's name has been updated to include the current time (in seconds).

scripts/irods/controller.py
```diff
@@ -32,6 +32,35 @@ def check_config(self):
...
+        if not os.path.isdir(rei_file_dir) or len(os.listdir(rei_file_dir)) == 0:
+            return
+
+        db_config = self.config.server_config['plugin_configuration']['database'].values()[0]
+        db_name = 'iRODS Catalog'
+        db_user = db_config['db_username']
+        db_password = db_config['db_password']
+
+        l.info('Migrating REI files to catalog ...')
+        cmd = [paths.migrate_rei_files_into_catalog_executable(), rei_file_dir, db_name, db_user, db_password]
```
trel, line 49: Does this leak the password into the process status (i.e another user could see the password via `ps`)?

scripts/irods/database_upgrade.py
```diff
@@ -88,6 +88,12 @@ def run_update(irods_config, cursor):
                         "'1580297960');")
         database_connect.execute_sql_statement(cursor, sql)
 
+    elif new_schema_version == 8:
+        if irods_config.catalog_database_type == 'postgres':
+            database_connect.execute_sql_statement(cursor, "alter table R_RULE_EXEC add column exe_context text;")
```
trel, line 93: we need some discourse/comments here ...

seems like TEXT is good for mysql/mariadb too?   and clob perhaps instead of blob?

scripts/setup_irods.py
```diff
@@ -263,6 +265,20 @@ def setup_service_account(irods_config, irods_user, irods_group):
...
+
+    # Determine if the rule execution (delay) server should be launched on startup.
+    if run_delay_server != None:
+        irods_config.server_config['advanced_settings']['run_delay_server'] = (run_delay_server.lower() in ['true', 'y', 'yes', '1'])
+    else:
+        role = irods_config.server_config['catalog_service_role'].lower()
+        irods_config.server_config['advanced_settings']['run_delay_server'] = (role == 'provider')
+
+    l.debug('Updating configuration file ...')
+    irods.lib.update_json_file_from_dict(irods.paths.server_config_path(), irods_config.server_config)
```
trel, line 280: add an l.debug line to print the decision made?

server/api/include/rsExecRuleExpression.hpp
```diff
@@ -3,7 +3,6 @@
 
 #include "exec_rule_expression.h"
 
-int rsExecRuleExpression(rsComm_t*,exec_rule_expression_t*);
+int rsExecRuleExpression(rsComm_t* comm,exec_rule_expression_t* input);
```
trel, line 6: space after comma

trel:
please consider sharing a successful example migration output in the commit message, so people can know what to expect.  

may also get that into the release notes and/or docs.

scripts/irods/controller.py
korydraughn: Maybe.  How do we want to approach this?

It is implemented this way for flexibility and simplicity.

Do we lose anything by having the migration app read the server config and pull that db info?
Can we guarantee that the REI file directory is always the same path?

trel: all fine questions.   will need to think through non-package installs perhaps.   but i think reading the database password from the config *should* be safe... can discuss.

server/api/include/rsExecRuleExpression.hpp
korydraughn: don't know how i missed that.  it has been updated.

scripts/irods/database_upgrade.py
korydraughn: what would the comments be about?  the data types selected?

based on what i've read about MySQL, TEXT has a max size of 2^16.  LONGTEXT is 4GB.  Which one do we want?

trel:
looks good - the commit message needs to be updated with the latest example output (new migrated directory, etc.)

test/migrate_rei_files_into_catalog/main.cpp
```diff
@@ -0,0 +1,431 @@
...
+            }
+        }
+
+        // Move the REI files out of the way.
+        for (auto&& p : processed_files) {
+            fs::rename(p, processed_rei_dir / p.filename());
+        }
+
+        fmt::print("Total REI files processed: {}/{}\n", processed_files.size(), total_files);
+        fmt::print("Migrated REI files location: {}.\n", processed_rei_dir.c_str());
```
trel, line 164: no period at end

test/migrate_rei_files_into_catalog/main.cpp
```diff
@@ -0,0 +1,431 @@
...
+        fmt::print("REI Migration log file location: {}.\n", log_path.c_str());
```
trel, line 165: no period at end

test/migrate_rei_files_into_catalog/main.cpp
korydraughn: done

test/migrate_rei_files_into_catalog/main.cpp
korydraughn: done

korydraughn:
We may need to discuss the migration tool and how it should handle REI files that were generated by static PEPs.

trel:
Yes, perhaps it inspects them, moves them aside and reports loudly about its findings.  But if it's only 1 of 35 in the queue...

Not sure what we can *DO* with them if they've already upgraded.

And no real way to do a 'pre-flight' check.

trel:
Pre-flight check is interesting... perhaps someone still running 4.2.8 could point something we build/provide at their `packedRei` directory to 'see' what might be in there... but there's still nothing actionable to do except for drain them all before upgrade, right?

korydraughn:
we could inspect the REI for things that we don't support just before Jsonifying it.  that would at least allow us to scream loudly.

korydraughn:
that sounds like a good solution.

korydraughn:
yes, they would have to drain those.

korydraughn:
will do

korydraughn:
Adjusted implementation based on community feedback (i.e. downtime concerns).

korydraughn:
All (expected) tests passed in CI.

All that is left is to review and squash this.

scripts/irods/test/test_delay_queue.py
```diff
@@ -478,6 +478,99 @@ def test_exception_in_delay_server(self):
...
+                        {
+                            delay('<INST_NAME>irods_rule_engine_plugin-irods_rule_language-instance</INST_NAME>') {
+                                *kvp.'delay_rule_fired' = 'YES';
+                                msiSetKeyValuePairsToObj(*kvp, *object_path, "-d");
+                            }
+                        }
+
+                        acPostProcForPut
+                        {
+                            schedule_delay_rule($userNameClient, $objPath);
```
trel, line 555: two parameters going in - but line 545 only takes the object path?

korydraughn: Will fix that asap.

korydraughn: Done

server/core/include/catalog_utilities.hpp
```diff
@@ -2,17 +2,21 @@
 #define IRODS_CATALOG_UTILITIES_HPP
 
 #include "irods_exception.hpp"
-#include "rcConnect.h"
-#include "rodsConnect.h"
+//#include "rcConnect.h"
+//#include "rodsConnect.h"
```
korydraughn, line 6: Remove these

server/api/src/rsRuleExecSubmit.cpp
```diff
@@ -1,189 +1,138 @@
...
 
-    status = write( reiFd, ruleExecSubmitInp->packedReiAndArgBBuf->buf,
-                    ruleExecSubmitInp->packedReiAndArgBBuf->len );
+    namespace ic = irods::experimental::catalog;
 
-    close( reiFd );
+    // TODO Re-evaluate the following code.
+    // The original implementation called getAndConnReHost(). The new code
+    // is likely wrong because it is querying about the provider.
+#if 0
```
korydraughn, line 98: Investigate / discuss

server/core/src/server_utilities.cpp
```diff
@@ -16,5 +18,26 @@ namespace irods
         return fs::server::exists(_comm, _input.objPath) &&
                !ix::key_value_proxy{_input.condInput}.contains(FORCE_FLAG_KW);
     }
+
+    auto contains_static_pep_session_variables(const std::string_view _rule_text) -> bool
+    {
+        static const std::regex session_vars_pattern{
```
korydraughn, line 24: This does not handle session variables in comments.

lib/core/src/irods_configuration_keywords.cpp
```diff
@@ -0,0 +1,149 @@
...
+    // plugin types
+    const std::string PLUGIN_TYPE_API( "api" );
+    const std::string PLUGIN_TYPE_RULE_ENGINE( "rule_engines" );
+    const std::string PLUGIN_TYPE_AUTHENTICATION( "auth" );
+    const std::string PLUGIN_TYPE_NETWORK( "network" );
+    const std::string PLUGIN_TYPE_DATABASE( "database" );
+    const std::string PLUGIN_TYPE_RESOURCE( "resources" );
+    const std::string PLUGIN_TYPE_MICROSERVICE( "microservices" );
+
+
```
korydraughn, line 138: Remove extra line

plugins/database/src/db_plugin.cpp
```diff
@@ -1,5 +1,6 @@
 // =-=-=-=-=-=-=-
 // irods includes
+#include "low_level_odbc.hpp"
```
korydraughn, line 3: This is probably not needed.

server/api/src/rsExecRuleExpression.cpp
```diff
@@ -1,77 +1,59 @@
...
+    if (rei->doi) {
+        if (rei->doi->next) {
+            free(rei->doi->next);
+            rei->doi->next = nullptr;
         }
     }
 
     std::string instance_name;
-    if(strlen(rei->pluginInstanceName) > 0) {
+    if (std::strlen(rei->pluginInstanceName) > 0) {
```
korydraughn, line 43: What happens if the string is not null terminated? This needs to use a safer function to check for empty.

server/core/src/server_utilities.cpp
korydraughn: Capture this as an issue once this is merged.

trel:
looks good - are we waiting on anything else before adding the #s?

korydraughn:
I'm running it through CI one more time to make sure everything works as expected.

korydraughn:
Passed CI and added #

Issue #5111, opened 2020-09-10
imeta does not correctly parse `-lC`


kript:
- [x] master
- [ ] 4-2-stable
----

### iRODS Version, OS and Version

iRODS 4.2.7
Ubuntu 18.04

### What did you try to do?

long list metadata on a collection with `imeta ls -lC /collection` (provides timestamp IIRC)

### Expected behavior

similar behaviour to `imeta ls -ld /collectionpath/object`

### Observed behavior (including steps to reproduce, if applicable)

```
$ imeta ls -C /Sanger1/home/jc18/tears_testing
AVUs defined for collection /Sanger1/home/jc18/tears_testing:
None
$ imeta ls -lC /Sanger1/home/jc18/tears_testing

Error: No object type descriptor (-d/C/R/u) specified
$ imeta ls -Cl /Sanger1/home/jc18/tears_testing

Error: No object type descriptor (-d/C/R/u) specified
```

trel:
Reading the help text, it seems... this is only intended(?) for data-objects?
```
 ls  -[l]d|C|R|u Name [AttName] (List existing AVUs for item Name)
```

Source says... the current behavior is the intent... but I agree that it's a bug and should show the time set regardless of what the AVU is attached to.

search for `longMode` in https://github.com/irods/irods_client_icommands/blob/4-2-stable/src/imeta.cpp

SwooshyCueb:
This has been addressed in the master branch, so that can be checked off.

trel:
Done.

SwooshyCueb:
4-2-stable can be checked off as well, now

PR #5176, opened 2020-10-07
[#5111] add test to verify existence of mtime in imeta ls -l output (master)

SwooshyCueb:
This is a companion pull request to irods/irods_client_icommands#155.

I've added a test case for the long mode output from `imeta ls` on the supported object types.


trel:
looks good to me - verify it passes for you, add the # to the commit message, and force push to update this PR.

SwooshyCueb:
# added

Issue #2279, opened 2014-08-15
core.re cache not updating after rapid sequential changes to core.re


kellerb:
- [x] main
- [X] 4-2-stable
- [x] main - remove all sleep()s
- [x] 4-3-stable - remove all sleep()s
------

Use hash (in addition to / instead of ) file timestamp to determine if core.re has been modified. Expected to resolve issues seen in #2276.

Once fixed, revert devtest workaround 2942df8ff405ab9df8d3401edb9418c396c2be0a.

trel:
if i get a vote, i'd say let's leave 4.2 as is.   the sleeps are only in the tests, right?

korydraughn:
I agree. Let's leave the sleeps in 4-2-stable and only target 4-3-stable for this.

korydraughn:
@MartinFlores751 Don't forget to cherry-pick #6765 to 4-3-stable.

trel:
and then this becomes a 4.2.12, yes?

korydraughn:
Because of the original work?

korydraughn:
Changed milestone to 4.2.12 because this issue is linked to commits that predate 4.2.

alanking:
So... we do need to remove the sleeps from 4-2-stable before closing?

trel:
> I agree. Let's leave the sleeps in 4-2-stable and only target 4-3-stable for this.


nopes.

korydraughn:
The removal of sleeps are for 4.3 only.

And given this comment, https://github.com/irods/irods/issues/2279#issuecomment-1363391140, there is no more work to do here.

Closing.

PR #5182, opened 2020-10-08
Remove some server log reading in tests (4-2-stable)

alanking:
CI tests are running


alanking:
Final commit fixes 1 test which failed - cherry-pick issue.

trel:
I notice that https://github.com/irods/irods/issues/2279 is also a thing that could be revisited soon - It appears we can remove sleep()s when ready.   Otherwise, if this is passing, please squash em.   Then we'll # em.

alanking:
Octothorpe'd and green'd

PR #5191, opened 2020-10-14
Use proxy objects in rsDataObjClose + New replica/data_object features (master)

alanking:
Making boundaries for logical locking more clear by refactoring rsDataObjClose a bit. Some other library features needed for logical locking as well.

CI tests passed except for one, which was fixed and passed at the bench.


trel:
two codacy thingers - otherwise, squash and # when happy

alanking:
The Codacy issues have been addressed with the use of static_cast, so I'm not sure why it is still complaining.

trel:
you fixed two, it found two others - different line numbers it seems

trel:
green codacy, please to #

alanking:
#'d

Issue #4949, opened 2020-05-22
Remove use of server log for testing where possible


alanking:
- [x] master
- [x] 4-2-stable
---

Reading the server log in the tests is not always necessary. This can be replaced, for the most part, through the use of metadata on data objects or resources.

PR #5196, opened 2020-10-15
[#4949] Link libboost_system in msi_get_agent_pid (master)

alanking:
Does not link for debug builds, apparently. Made to match other microservice plugins.


alanking:
#'d

Issue #5081, opened 2020-08-27
imeta now exits when it encounters an empty line on STDIN, rather than ^D


kjsanger:
- [x] master
- [x] 4-2-stable
------

iRODS 4.2.7, Ubuntu bionic.

What I did:

Pipe a series of imeta commands to STDIN of imeta for the form:

```
add -d /testZone/home/irods/test/data/f1.txt attr1 value1 units1
add -d /testZone/home/irods/test/data/f2.txt attr1 value1 units1
add -d /testZone/home/irods/test/data/f3.txt attr1 value1 units1

add -d /testZone/home/irods/test/data/a/f4.txt attr1 value1 units1
add -d /testZone/home/irods/test/data/a/f5.txt attr1 value1 units1
add -d /testZone/home/irods/test/data/a/f6.txt attr1 value1 units1

add -d /testZone/home/irods/test/data/a/x/f7.txt attr1 value1 units1
add -d /testZone/home/irods/test/data/a/x/f8.txt attr1 value1 units1
add -d /testZone/home/irods/test/data/a/x/f9.txt attr1 value1 units1
```

What I expected to happen:

9 commands to be executed. This is the behaviour in iRODS 3.x, 4.0 and 4.1

What happened:

3 commands were executed and imeta exited when it  encountered the blank line

This also happens in REPL mode. This is iRODS 4.1 where returns are accepted
as input and don't end the session, The session is ended by ^D:

```
$ imeta
imeta>
imeta>
imeta>
imeta>
```

In 4.2, any return prematurely ends the session.

kjsanger:
Also exits from REPL mode as soon

PR #5204, opened 2020-10-21
[#5081] Add test class for interactive imeta (master)

SwooshyCueb:
Companion pull request for irods/irods_client_icommands#163

A few tests for interactive `imeta`

Addresses #5081

### DIFF
--- a/scripts/core_tests_list.json
+++ b/scripts/core_tests_list.json
@@ -15,6 +15,7 @@
     "test_ils",
     "test_imeta_error_handling",
     "test_imeta_help",
+    "test_imeta_interactive",
     "test_imeta_set",
     "test_imv",
     "test_ipasswd",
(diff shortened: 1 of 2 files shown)


alanking (approved):
Looks good to me.

Without deeply evaluating, it looks like this could possibly benefit from a utility in our testing suite called `delayAssert` which repeatedly tries an assertion up to a certain number or if it passes. But this is fine how it is, just wanted to make mention of that.

Also, per commit message, should we create an issue for REPL testing?

SwooshyCueb:
> Also, per commit message, should we create an issue for REPL testing?

Yeah, I think that might be a good idea

trel:
Agreed, please create an issue for that - and tweak the commit message to leave out the suggestion.

If tests are passing, please # it.

SwooshyCueb:
All tests passed except the one that doesn't. #s added

PR #5243, opened 2020-11-12
Miscellaneous machinery in run-up to logical locking (4-2-stable)

alanking:
CI tests running


alanking:
CI tests passed except for `test_irodsctl`, which passes at the bench.

Added another commit to remove the moved test class from the parallel core tests run by CI.

Issue #5264, opened 2020-11-21
delay server should report itself cleanly to ips via spOption


trel:
- [x] master
- [x] 4-2-stable
-----

## Feature

The `irodsReServer`, soon to be known as the delay server, should appear in `ips` more cleanly by setting `spOption` when waking up and connecting.  It currently appears as the default 'UNKNOWN'.

PR #5270, opened 2020-11-24
[#5264] Delay server name now shows up in ips output (master)

korydraughn:


korydraughn:
Hold on merging this. This code needs to be adjusted so that other tools have access to it (e.g. istream and itouch)

korydraughn:
After further thought, the code is fine. There's no need to create a utility function for this given it just sets an environment variable.

korydraughn:
Added #

Issue #5193, opened 2020-10-14
close_file after creating a data object in PRC results in intermediate replica


alanking:
- [x] master
- [x] 4-2-stable
---
### iRODS Version, OS and Version
4-2-stable

### What did you try to do?
Run the following script provided by @d-w-moore:
```
import os
import ssl
from irods.session import iRODSSession
try:
    env_file = os.environ['IRODS_ENVIRONMENT_FILE']
except KeyError:
    env_file = os.path.expanduser('~/.irods/irods_environment.json')

ssl_context = ssl.create_default_context(purpose=ssl.Purpose.SERVER_AUTH, cafile=None, capath=None, cadata=None)
ssl_settings = {'ssl_context': ssl_context}

with iRODSSession(irods_env_file=env_file, **ssl_settings) as session:
    dobj_path = '/tempZone/home/rods/VERSION.json'

    try:
        obj = session.data_objects.create(dobj_path)
    except:
        print('failed on create of data object')
```

### Expected behavior
0-length replica marked as good

### Observed behavior (including steps to reproduce, if applicable)
0-length replica marked as intermediate

rsDataObjClose is not stamping the catalog with the good replica status for whatever reason, possibly related to having written 0 bytes. Please investigate.

trel:
yes, works as expected on 4.2.8, as released.

PR #5283, opened 2020-12-04
[#5193] Mark newly created replicas as good with no bytes written (master)

alanking:
CI tests running.

A note about the TODO: This is technically incorrect as we need to check for the expected size of the replica before we determine that it is good. This PR will restore the historical behavior of assuming new replicas with no bytes written are good. Future changes will require more vetting with respect to the expected size of the replica.


alanking:
#'d

PR #5284, opened 2020-12-04
[#5193] Mark newly created replicas as good with no bytes written (4-2-stable)

alanking:
CI tests running.

A note about the TODO: This is technically incorrect as we need to check for the expected size of the replica before we determine that it is good. This PR will restore the historical behavior of assuming new replicas with no bytes written are good. Future changes will require more vetting with respect to the expected size of the replica.

### DIFF
--- a/server/api/src/rsDataObjClose.cpp
+++ b/server/api/src/rsDataObjClose.cpp
@@ -558,20 +558,25 @@ namespace
             throw;
         }
 
-        if (PUT_OPR != l1desc.oprType || !l1desc.chksumFlag) {
-            // return success as the rest is put-specific...
-            return;
-        }
-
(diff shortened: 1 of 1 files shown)


alanking:
Another note - if something goes wrong with the checksum process for PUTs here, the replica will be stuck in the intermediate state instead of moving to the stale state. This behavior will be overwritten in a future commit, but adding this here in case it comes up later.

trel:
looks fine to me - let's # it and get dan happy.

d-w-moore:
@alanking , great, thanks! in my repo now and building

d-w-moore:
thanks @trel yes, merge at leisure and I'll test based on what I've imported from alan's 4-2-stable branch . ... I'm sure the SHA's will match up when I check'em later....

alanking:
Had 1 test fail in CI but passes at the bench.

Issue #5252, opened 2020-11-17
ichksum --verify doesn't verify size of unchecksummed replicas


bh9:
### iRODS Version, OS and Version
4.2.7, Ubuntu 18.04 (using 16.04 packages)

### What did you try to do?
While trying to establish if we could use ichksum to safely checksum all our unchecksummed files, I wanted to check if it would pick up unchecksummed files with the wrong size and a different checksum to the other replica (see #5251), so I created a replicated unchecksummed object, checksummed one replica, and truncated the other replica:
```
bh9@deskpro110321:~$ ils -L ichksumtests/ichksum-case-8
  irods-ires        0 root;replicate;green;green8;irods-seq-sr02-ddn-rd10-9-10-11         1523 2020-11-17.11:58 & ichksum-case-8
    9e0d157af0519e85e2978b2a1d84ef9f    generic    /irods-seq-sr02-ddn-rd10-9-10-11/home/irods-ires/ichksumtests/ichksum-case-8
  irods-ires        1 root;replicate;red;red2;irods_r2_dev_sdb         1523 2020-11-17.11:58 & ichksum-case-8
        generic    /irods_r2_dev_sdb/home/irods-ires/ichksumtests/ichksum-case-8
```
where `/irods_r2_dev_sdb/home/irods-ires/ichksumtests/ichksum-case-8` is not the right length:
```
irods-r2-dev@irods-r2-dev:~$ wc -c  /irods_r2_dev_sdb/home/irods-ires/ichksumtests/ichksum-case-8
605 /irods_r2_dev_sdb/home/irods-ires/ichksumtests/ichksum-case-8
```
```
irods-g2-dev@irods-g2-dev:~$ wc -c /irods-seq-sr02-ddn-rd10-9-10-11/home/irods-ires/ichksumtests/ichksum-case-8
1523 /irods-seq-sr02-ddn-rd10-9-10-11/home/irods-ires/ichksumtests/ichksum-case-8
```

### Expected behavior
`--verify` checks that the length is not the same, and the checksum is not added to the catalog, a USER_FILE_SIZE_MISMATCH error is returned, and ichksum returns -3

### Observed behavior (including steps to reproduce, if applicable)
While the replica has no checksum, `--verify` does not check the size, but once there is a checksum, `--verify` does return the USER_FILE_SIZE_MISMATCH and rc 3
```
bh9@deskpro110321:~$ ichksum -aKMv --verify ichksumtests/ichksum-case-8
[...]

Issue #5251, opened 2020-11-17
ichksum -aK on mismatched replicas


bh9:
### iRODS Version, OS and Version
4.2.7, Ubuntu 18.04 (with 16.04 packages)

### What did you try to do?
We're attempting to fix a number of unchecksummed files and identify if any have corrupted since they were uploaded. Reading the ichksum man page, I saw 
```
       -a     checksum  all  replicas.  ils  -L should be used to list the values of all replicas If used together
              with -K, and the replicas have different checksums, an error message is displayed, but  no  checksum
              is updated in the catalog.

       -K     verify  the  checksum value in iCAT. If the checksum value does not exist, compute and register one.
              If the catalog checksum exists and the file checksum differs, returns USER_CHKSUM_MISMATCH and  does
              not update in the catalog.
```
I produced a test for this, by creating a replicated object without checksums, then modified one of the replicas on disk to simulate a corruption:
```
bh9@deskpro110321:~$ ils -L ichksumtests/ichksum-case-10
  irods-ires        0 root;replicate;green;green3;irods-seq-sr02-ddn-rd10-21-22-23         1523 2020-11-17.11:59 & ichksum-case-10
        generic    /irods-seq-sr02-ddn-rd10-21-22-23/home/irods-ires/ichksumtests/ichksum-case-10
  irods-ires        1 root;replicate;red;red2;irods_r2_dev_sdb         1523 2020-11-17.11:59 & ichksum-case-10
        generic    /irods_r2_dev_sdb/home/irods-ires/ichksumtests/ichksum-case-10
```
Where `/irods_r2_dev_sdb/home/irods-ires/ichksumtests/ichksum-case-10` has a single character change:
```
irods-r2-dev@irods-r2-dev:~$ head -n1 /irods_r2_dev_sdb/home/irods-ires/ichksumtests/ichksum-case-10
localhoot provider-build
```
```
irods-g2-dev@irods-g2-dev:~$ head -n1 /irods-seq-sr02-ddn-rd10-21-22-23/home/irods-ires/ichksumtests/ichksum-case-10
localhost provider-build
```

### Expected behavior
The man page entry for `-a` says that using `-aK` should report if the checksums differ, and not update their checksums in the catalog, so I would expect an error message, and a non-zero return code

### Observed behavior (including steps to reproduce, if applicable)
`ichksum` returns 0, and updates the checksums in the catalog, producing a pair of replicas with mismatched checksums:
```
[...]

PR #5286, opened 2020-12-04
[#5251][#5252] Fixed the checksum API (master)

korydraughn:
Needs to be run through CI.

However, all tests in `test_ichksum.py` pass.

### DIFF
--- a/lib/api/include/dataObjChksum.h
+++ b/lib/api/include/dataObjChksum.h
@@ -1,29 +1,117 @@
 #ifndef DATA_OBJ_CHKSUM_H__
 #define DATA_OBJ_CHKSUM_H__
 
-#include "rcConnect.h"
-#include "dataObjInpOut.h"
-#include "objInfo.h"
+/// \file
+
+struct RcComm;
+struct DataObjInp;
 
 #ifdef __cplusplus
-extern "C"
+extern "C" {
 #endif
-int rcDataObjChksum( rcComm_t *conn, dataObjInp_t *dataObjChksumInp, char **outChksum );
 
-/* rcDataObjChksum - Chksum a iRODS data object.
- * Input -
- *   rcComm_t *conn - The client connection handle.
- *   dataObjInp_t *dataObjInp - generic dataObj input. Relevant items are:
- *	objPath - the path of the data object.
- *	openFlags - should be set to O_RDONLY.
- *   condInput - FORCE_CHKSUM_KW - force checksum even if one exist.
- *		 CHKSUM_ALL_KW - checksum all replica.
- *		 VERIFY_CHKSUM_KW - verify the checksum value in icat.
- * 		 REPL_NUM_KW - the "value" gives the replica number of the
- *		   copy to checksum.
- * OutPut -
- *   char **outChksum - the chksum string
- *   return value - The status of the operation.
- */
+/// Computes and stores a checksum for a data object.
+///
+/// The checksum algorithm used by this function is determined by the client and server's configuration,
+/// but defaults to SHA256.
+///
+/// This API function supports two modes of operation:
+/// - \p Lookup/Update (default): Deals with computation and updates of checksum information.
+/// - \p Verification:            Deals with verifying the checksum information. This mode does not perform catalog updates.
+///
+/// Operations that target multiple replicas will only affect replicas that are marked good. This means intermediate,
+/// locked, and stale replicas will be ignored.
+///
+/// Operations that target a specific replica are allowed to operate on stale replicas.
+///
+/// \param[in]  comm             The communication object.
+/// \param[in]  dataObjChksumInp \parblock
+/// The bundle of input arguments that dictate what happens.
+///
+/// The relevant member variables of this structure are:
+/// - \p objPath:   The absolute path to a data object.
+/// - \p condInput: A list of keywords and values that instruct the server on how to carry out the operation.
+/// 
+/// Keywords supported by \p condInput:
+/// - \p FORCE_CHKSUM_KW:
+///     - Instructs the server to compute and update the checksum.
+///     - Must be set to an empty string.
+/// - \p REPL_NUM_KW:
+///     - Identifies a specific replica by the replica number.
+///     - Accepts the replica number as a string.
+///     - Incompatible with \p RESC_NAME_KW.
+///     - Incompatible with \p CHKSUM_ALL_KW.
+/// - \p RESC_NAME_KW:
+///     - Identifies a specific replica by the leaf resource name.
+///     - Accepts the resource name as a string.
+///     - Incompatible with \p REPL_NUM_KW.
+///     - Incompatible with \p CHKSUM_ALL_KW.
+/// - \p CHKSUM_ALL_KW:
+///     - Instructs the server to operate on all replicas.
+///     - Must be set to an empty string.
+///     - In \p lookup/update mode, reports if the replicas do not share identical checksums if no errors occur.
+///     - Incompatible with \p RESC_NAME_KW.
+///     - Incompatible with \p REPL_NUM_KW.
+/// - \p ADMIN_KW:
+///     - Instructs the server to execute the operation as an administrator.
+///     - Must be set to an empty string.
+/// - \p VERIFY_CHKSUM_KW:
+///     - Instructs the server to verify the checksum information (enables verification mode).
+///     - Must be set to an empty string.
+///     - Operates on all replicas unless a specific replica is targeted via \p REPL_NAME_KW or \p RESC_NAME_KW.
+///     - No checksum is returned to the client when in verification mode.
+///     - The following operations are performed:
+///         1. Reports replicas with mismatched size information (physical vs catalog).
+///         2. Reports replicas that are missing checksums to the client.
+///         3. Reports replicas with mismatched checksums (computed vs catalog).
+///         4. Reports if the replicas do not share identical checksums.
+///     - If \p NO_COMPUTE_KW is set, step 3 will not be performed.
+///     - If a specific replica is targeted, step 4 will not be performed.
+///     - Verification results are reported to the client via the RcComm::rError object.
+/// - \p NO_COMPUTE_KW:
+///     - Instructs the server to not compute a checksum in verification mode potentially leading to a performance boost.
+///     - Must be set to an empty string.
+///     - A modifier for \p VERIFY_CHKSUM_KW.
+/// \endparblock
+/// \param[out] outChksum        Holds the returned checksum if available.
+///
+/// \returns An integer.
+/// \retval 0        On success.
+/// \retval Non-zero On failure.
+///
+/// \b Example
+/// \code{.cpp}
+/// DataObjInp input;
+/// memset(&input, 0, sizeof(DataObjInp));
+///
+/// // Set the absolute path to the data object.
+/// rstrcpy(input.objPath, "/tempZone/home/rods/data_object", MAX_NAME_LEN);
+///
+/// // Verify the checksum information for replica number 2.
+/// addKeyVal(&input.condInput, VERIFY_CHKSUM_KW, "");
+/// addKeyVal(&input.condInput, REPL_NUM_KW, "2");
+///
+/// RcComm* comm = // Our iRODS connection.
+/// char* computed_checksum = NULL;
+/// const int ec = rcDataObjChksum(comm, &input, &computed_checksum);
+///
+/// if (ec < 0) {
+///     // Handle error.
+/// }
+///
+/// //
+/// // Do something with "computed_checksum".
+/// //
+///
+/// free(computed_checksum);
+/// \endcode
+int rcDataObjChksum(struct RcComm* comm,
+                    struct DataObjInp* dataObjChksumInp,
+                    char** outChksum);
 
+#ifdef __cplusplus
+} // extern "C"
 #endif
+
+#endif // DATA_OBJ_CHKSUM_H__
+
--- a/lib/core/include/rodsKeyWdDef.h
+++ b/lib/core/include/rodsKeyWdDef.h
@@ -104,6 +104,7 @@
 #define REG_REPL_KW                                 "regRepl"  /* register replica */
 #define AGE_KW                                      "age"  /* age of the file for itrim */
 #define DRYRUN_KW                                   "dryrun"  /* do a dry run */
+#define NO_COMPUTE_KW                               "no_compute"  /* do not compute anything (similar to dryrun) */
 #define ACL_COLLECTION_KW                           "aclCollection"  /* the collection from which
 * the ACL should be used */
 #define NO_CHK_COPY_LEN_KW                          "noChkCopyLen"  /* Don't check the len
--- a/scripts/irods/test/test_chunkydevtest.py
+++ b/scripts/irods/test/test_chunkydevtest.py
@@ -62,7 +62,7 @@ class ChunkyDevTest(ResourceBase):
         # put and list basic file information
         self.admin.assert_icommand("ils -AL", 'STDOUT_SINGLELINE', "home")  # debug
         self.admin.assert_icommand("iput -K --wlock " + test_file + " " + irodshome + "/icmdtest/foo1")
-        self.admin.assert_icommand("ichksum -f " + irodshome + "/icmdtest/foo1", 'STDOUT_SINGLELINE', "performed = 1")
+        self.admin.assert_icommand("ichksum -f " + irodshome + "/icmdtest/foo1", 'STDOUT_SINGLELINE', "    sha2:")
         self.admin.assert_icommand("iput -kf " + test_file + " " + irodshome + "/icmdtest/foo1")
         self.admin.assert_icommand("ils " + irodshome + "/icmdtest/foo1", 'STDOUT_SINGLELINE', "foo1")
         self.admin.assert_icommand("ils -l " + irodshome + "/icmdtest/foo1", 'STDOUT_SINGLELINE', ["foo1", myssize])
@@ -642,7 +642,7 @@ class ChunkyDevTest(ResourceBase):
                                    irodshome + "/icmdtest1/foo1 " + irodshome + "/icmdtest1/foo2")
         self.admin.assert_icommand("imv " + irodshome + "/icmdtest1/foo2 " + irodshome + "/icmdtest1/foo4")
         self.admin.assert_icommand("imv " + irodshome + "/icmdtest1/foo4 " + irodshome + "/icmdtest1/foo2")
-        self.admin.assert_icommand("ichksum -K " + irodshome + "/icmdtest1/foo2", 'STDOUT_SINGLELINE', "foo2")
+        self.admin.assert_icommand("ichksum -K " + irodshome + "/icmdtest1/foo2")
         self.admin.assert_icommand("iget -f -K " + irodshome + "/icmdtest1/foo2 " + dir_w)
         os.unlink(dir_w + "/foo2")
         self.admin.assert_icommand("irsync " + test_file + " i:" + irodshome + "/icmdtest1/foo1")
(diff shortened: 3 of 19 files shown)


server/api/src/rsDataObjChksum.cpp
```diff
@@ -330,19 +505,22 @@ verifyDatObjChksum( rsComm_t *rsComm, dataObjInfo_t *dataObjInfo,
                  dataObjInfo->objPath, status );
         return status;
     }
-    if ( *outChksumStr == NULL ) {
-        rodsLog( LOG_ERROR, "verifyDatObjChksum: outChkSumStr is null." );
+
+    if ( *outChksumStr == nullptr ) {
+        rodsLog( LOG_ERROR, "verifyDatObjChksum: outChkSumStr is nullptr." );
```
korydraughn, line 510: revert nullptr to null.

unit_tests/src/test_rc_data_obj.cpp
```diff
@@ -356,3 +361,86 @@ TEST_CASE("open,read,write,close")
...
+        io::odstream{tp, data_object} << "some important data.";
+    }
+
+    DataObjInp input{};
+    std::strcpy(input.objPath, data_object.c_str());
+    ix::key_value_proxy kvp{input.condInput};
+
+    char* ignore_checksum{};
+
+    SECTION("replica number flag and resource name flag")
```
korydraughn, line 402: the descriptions of these sections need to be better.

server/api/src/rsDataObjChksum.cpp
```diff
@@ -315,13 +491,12 @@ dataObjChksumAndRegInfo( rsComm_t *rsComm, dataObjInfo_t *dataObjInfo,
     return status;
 }
 
-int
-verifyDatObjChksum( rsComm_t *rsComm, dataObjInfo_t *dataObjInfo,
-                    char **outChksumStr ) {
-    int status;
-
+int verifyDatObjChksum(rsComm_t* rsComm,
```
trel, line 494: Consider marking `verifyDatObjChksum` as deprecated, and have it and everywhere else call a new, correctly spelled `verifyDataObjChksum` with an `a`.

lib/core/include/rodsKeyWdDef.h
```diff
@@ -104,6 +104,7 @@
 #define REG_REPL_KW                                 "regRepl"  /* register replica */
 #define AGE_KW                                      "age"  /* age of the file for itrim */
 #define DRYRUN_KW                                   "dryrun"  /* do a dry run */
+#define NO_COMPUTE_KW                               "no_compute"  /* do not compute anything (similar to dryrun) */
```
trel, line 107: I notice this is the first(?) with an underscore... I think we want `noCompute` instead for consistency?

korydraughn: It isn't the first, but I can change it if you like. I'm pretty sure there are other instances in that file.

server/api/src/rsDataObjChksum.cpp
```diff
@@ -1,97 +1,400 @@
...
+        return irods::RESOURCE_CLASS_BUNDLE == resc_class;
+    } // is_member_of_bundle_resource
+
+    std::vector<DataObjInfo*> get_good_replicas(DataObjInfo* _replicas)
+    {
+        std::vector<DataObjInfo*> good_replicas;
+        good_replicas.reserve(std::distance(begin(_replicas), end(_replicas)));
+
+        for (auto&& r : _replicas) {
+            if (!is_stale_replica(r) && !is_member_of_bundle_resource(r.rescId)) {
```
trel, line 111: careful here, opposite of stale is not necessarily good.    we now have intermediate...

server/api/src/rsDataObjChksum.cpp
```diff
@@ -1,97 +1,400 @@
...
+                good_replicas.push_back(&r);
+            }
+        }
+
+        good_replicas.shrink_to_fit();
+
+        return good_replicas;
+    } // get_good_replicas
+
+    void report_number_of_stale_replicas_skipped(const DataObjInfo* _replicas, rError_t& _error_stack)
```
trel, line 121: might be more correct to be 'not good replicas'... b/c intermediate and locked, etc...

server/api/src/rsDataObjChksum.cpp
```diff
@@ -1,97 +1,400 @@
...
+        if (report_replicas_without_checksums(good_replicas, _comm.rError)) {
+            // Return immediately because this function requires that all replicas
+            // have a checksum after this point.
+            return 0;
+        }
+
+        //
+        // At this point, all replicas are guaranteed to have a checksum.
+        //
+
```
trel, line 254: consider possibility of other replicas being created in the meantime?    or are we within a giant transaction?

lib/core/include/rodsKeyWdDef.h
trel: oooh, i see all the new ones have underscores.    resolving.   carry on.

server/api/src/rsDataObjChksum.cpp
korydraughn: I agree. How should this believe when intermediate and lock states are encountered?

korydraughn: *believe -> behave

trel: we should ignore anything other than 'good'.    right?

server/api/src/rsDataObjChksum.cpp
korydraughn: this function uses only the list of replicas retrieved by `getDataObjInfoIncSpecColl`, so new replicas are not a concern at this point.

and no, this code is not within transaction.

server/api/src/rsDataObjChksum.cpp
korydraughn: i think you're right. with intermediates and logical locking, if someone opens a data object for write and another person tries to do checksum related operations, what do we expect to happen? my guess is that the checksum call would fail because the server said the data object is locked and being operated on. however, if that is not the case, then it may be possible for the checksum call to compute and update checksums for SOME of the replicas and not the ones being written to.

server/api/src/rsDataObjChksum.cpp
```diff
@@ -1,97 +1,437 @@
...
+        return good_replicas;
+    } // get_good_replicas
+
+    void report_number_of_replicas_skipped(const DataObjInfo* _replicas, rError_t& _error_stack)
+    {
+        std::size_t count = 0;
+
+        for (const auto& r : _replicas) {
+            if (!is_good_replica(r)) {
```
korydraughn, line 129: the bundle resource needs to be considered here. just like in `get_good_replicas`.

scripts/irods/test/test_ichksum.py
```diff
@@ -168,3 +168,112 @@ def test_chksum_catalog_verify_not_exhausting_statements__issue_4732(self):
...
+        self.admin.assert_icommand(['ichksum', '-K', data_object], 'STDOUT', ['ERROR: Physical size does not match size in catalog for replica [0].'])
+
+        # Show that --verify is an alias for -K.
+        self.admin.assert_icommand(['ichksum', '--verify', data_object], 'STDOUT', ['ERROR: Physical size does not match size in catalog for replica [0].'])
+
+    def test_ichksum_reports_when_replicas_do_not_share_identical_checksums__issue_5252(self):
+        data_object = 'foo'
+        self.admin.assert_icommand(['istream', 'write', data_object], input='some data')
+        self.admin.assert_icommand(['irepl', '-R', self.testresc, data_object])
+        self.admin.assert_icommand(['istream', 'write', '-n0', '-o0', '--no-trunc', data_object], input='XXXX')
```
korydraughn, line 207: these tests need to be revisited. if multiple replicas exist and istream is used to append/overwrite data in one of the replicas, should the sibling replicas be marked stale. if yes, then some of these tests are incorrect because they assume istream will not cause replicas to be marked stale (which is temporarily the case right now).

trel: Yes, sibling replicas should be marked stale by iRODS today, and the same when @alanking lands his intermediate/logical_locking code.   The tests should be changed/updated.

trel:
Looking very good...

In your commit message (and perhaps the docs, when that is done)...

> Operations that target multiple replicas always ignore stale replicas.
> Operations that target a specific replica are allowed to operate on stale replicas.

This should probably mention intermediate and locked replicas to be thorough.

korydraughn:
Will do.

Please look at the changes to the older tests closely. I want to make sure everything is correct and that the point of the tests are still maintained.

korydraughn:
Passed CI.

scripts/irods/test/test_chunkydevtest.py
```diff
@@ -698,9 +698,9 @@ def test_large_files_from_devtest(self):
             os.unlink(lrsfile)
         if os.path.isfile(rsfile):
             os.unlink(rsfile)
-        self.admin.assert_icommand("iput -vbPKr --retries 10 --wlock -X " + rsfile + " --lfrestart " +
+        self.admin.assert_icommand("iput -vbPkKr --retries 10 --wlock -X " + rsfile + " --lfrestart " +
```
korydraughn, line 701: adding `-k` fixed this assertion. does this change the intention of the test in any way?

trel: It does not - the assertion here is about using the restart file, so I think you're good to go.

korydraughn: So, what is the significance of the ichksum calls/options? Are they needed at all for testing restarts?

trel: I would think not.  I would speculate these are here for coverage more than targeted testing.

I'd be okay if you removed them here.

korydraughn: ok. i'll look into removing the checksum options on iput as well as the ichksum calls.

korydraughn:
All tests passed CI.

trel (commented):
i think it's really close... is there anything else on your list?

scripts/irods/test/test_ichksum.py
```diff
@@ -75,6 +75,7 @@ def test_ichksum_force(self):
         new_chksum = lib.file_digest(full_path, 'sha256', encoding='base64')
         self.admin.assert_icommand(['ichksum', '-f', filename], 'STDOUT_SINGLELINE', new_chksum)
 
+    @unittest.skip('Skipped due to change in ichksum behavior and new verification tests')
```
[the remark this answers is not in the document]
korydraughn: Sure. I didn't know if we wanted to remove these tests or not. Figured skipping was good since deletion felt harsh.

scripts/irods/test/test_ichksum.py
```diff
@@ -168,3 +168,183 @@ def test_chksum_catalog_verify_not_exhausting_statements__issue_4732(self):
...
+        # Show that using --silent produces no output when all replicas are in a good state.
+        self.admin.assert_icommand(['ichksum', '-r', '-a', '--silent', self.admin.session_collection])
+
+    def test_ichksum_ignores_replicas_that_are_not_marked_good_or_stale__issue_5252(self):
+        data_object = os.path.join(self.admin.session_collection, 'foo')
+        self.admin.assert_icommand(['istream', 'write', data_object], input='some data')
+
+        # Set the replica's status to an unknown value, in this test, 3.
+        # If the replica's status was set to intermediate (2), ichksum would return HIERARCHY_ERROR.
+        self.admin.assert_icommand(['iadmin', 'modrepl', 'logical_path', data_object, 'replica_number', '0', 'DATA_REPL_STATUS', '3'])
```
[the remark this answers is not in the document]
korydraughn: Agreed. I set it to 3 to reflect the current state of iRODS. I figured someone would update it following Alan's future changes. That would bring attention to whether the other replica statuses should be handled differently in regards to ichksum.

Ultimately I didn't want to risk introducing issues due to attempts at future proofing the test.

scripts/irods/test/test_icommands_file_operations.py
```diff
@@ -882,6 +882,7 @@ def test_igetwild_with_semicolon_in_filename(self):
             os.unlink(badpath)
         os.unlink(localpath)
 
+    @unittest.skip('Replaced by new test. See test_ichksum.py and look for test names with a suffix of "__issue_5252".')
```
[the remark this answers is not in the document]
korydraughn: Sure

scripts/irods/test/test_icommands_file_operations.py
```diff
@@ -916,6 +917,7 @@ def test_irm_colloprstat__3572(self):
                 start_index=initial_size_of_server_log))
         os.unlink(filename)
 
+    @unittest.skip('Replaced by new test. See test_ichksum.py - test_ichksum_reports_when_replicas_physical_size_does_not_match_size_in_catalog__issue_5252.')
```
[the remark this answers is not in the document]
korydraughn: Sure

korydraughn:
I think the only thing that remains for this PR is documentation of the checksum api functions.

The help text of ichksum can be updated after this PR is merged.

scripts/irods/test/test_ichksum.py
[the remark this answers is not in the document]
trel: agreed.

scripts/irods/test/test_ichksum.py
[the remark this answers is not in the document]
trel: well, either deletion or a stub/pass that just refers to the new thing.   that way, there's a searchable issue number for the old thing too.

trel:
good good.   awaiting docs.

scripts/core_tests_list.json
```diff
@@ -1,71 +1,3 @@
...
-    "test_resource_types.Test_Resource_ReplicationWithinReplication",
-    "test_resource_types.Test_Resource_Unixfilesystem",
-    "test_resource_types.Test_Resource_WeightedPassthru",
-    "test_rule_engine_plugin_framework",
-    "test_rule_engine_plugin_passthrough",
-    "test_rulebase",
-    "test_ssl",
-    "test_stacktrace",
-    "test_symlink_operations"
+    "test_all_rules"
```
trel, line 2: is this correct?   removing all the others?

alanking: I'm familiar with this trick :)

irods/irods_testing_jenkins#107

server/api/src/rsDataObjChksum.cpp
```diff
@@ -1,116 +1,494 @@
...
+    if ( remoteFlag == REMOTE_HOST ) {
+        return rcDataObjChksum( rodsServerHost->conn, dataObjChksumInp, outChksum );
     }
 
     // If the client specified a leaf resource, then discover the hierarchy and
     // store it in the keyValPair_t. This instructs the iRODS server to create
     // the replica at the specified resource if it does not exist.
-    auto kvp = irods::experimental::make_key_value_proxy(dataObjChksumInp->condInput);
+    auto kvp = ix::make_key_value_proxy(dataObjChksumInp->condInput);
     if (kvp.contains(LEAF_RESOURCE_NAME_KW)) {
```
korydraughn, line 473: @alanking should this keyword actually be a part of this API?

scripts/irods/test/test_ichksum.py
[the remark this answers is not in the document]
korydraughn: removed test

scripts/irods/test/test_icommands_file_operations.py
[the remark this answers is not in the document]
korydraughn: removed test

scripts/irods/test/test_icommands_file_operations.py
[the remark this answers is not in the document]
korydraughn: removed test

lib/api/include/dataObjChksum.h
```diff
@@ -1,29 +1,116 @@
...
+/// - \p ADMIN_KW:
+///     - Instructs the server to execute the operation as an administrator.
+///     - Must be set to an empty string.
+/// - \p VERIFY_CHKSUM_KW:
+///     - Instructs the server to verify the checksum information (enables verification mode).
+///     - Must be set to an empty string.
+///     - Operates on all replicas unless a specific replica is targeted via \p REPL_NAME_KW or \p RESC_NAME_KW.
+///     - No checksum is returned to the client when in verification mode.
+///     - The following operations are performed:
+///         1. Reports replicas with mismatch size information (physical vs catalog).
```
trel, line 63: mismatched

lib/api/include/dataObjChksum.h
```diff
@@ -1,29 +1,116 @@
...
+///         2. Reports replicas that are missing checksums to the client.
+///         3. Reports replicas with mismatch checksums (computed vs catalog).
```
trel, line 65: mismatched

server/api/include/rsDataObjChksum.hpp
```diff
@@ -1,13 +1,127 @@
...
+/// - \p ADMIN_KW:
+///     - Instructs the server to execute the operation as an administrator.
+///     - Must be set to an empty string.
+/// - \p VERIFY_CHKSUM_KW:
+///     - Instructs the server to verify the checksum information (enables verification mode).
+///     - Must be set to an empty string.
+///     - Operates on all replicas unless a specific replica is targeted via \p REPL_NAME_KW or \p RESC_NAME_KW.
+///     - No checksum is returned to the client when in verification mode.
+///     - The following operations are performed:
+///         1. Reports replicas with mismatch size information (physical vs catalog).
```
trel, line 60: mismatched

server/api/include/rsDataObjChksum.hpp
```diff
@@ -1,13 +1,127 @@
...
+///         2. Reports replicas that are missing checksums to the client.
+///         3. Reports replicas with mismatch checksums (computed vs catalog).
```
trel, line 62: mismatched

lib/api/include/dataObjChksum.h
```diff
@@ -1,29 +1,116 @@
...
- *		 VERIFY_CHKSUM_KW - verify the checksum value in icat.
- * 		 REPL_NUM_KW - the "value" gives the replica number of the
- *		   copy to checksum.
- * OutPut -
- *   char **outChksum - the chksum string
- *   return value - The status of the operation.
- */
+/// Computes and stores the checksum for a data object.
+///
+/// The default checksum algorithm is SHA256.
```
trel, line 15: hmmm...  could be a tad more descriptive/correct.

a link to the github repo (which has a branch in the URL)...
 - https://github.com/irods/irods_docs/blob/4-2-stable/docs/system_overview/configuration.md#checksum-configuration

or

a link to the rendered docs site (which has a version number in the URL)...
 - https://docs.irods.org/4.2.8/system_overview/configuration/#checksum-configuration

or

"The checksum algorithm used is determined by considering both the sender and receiver's hash schemes and the receiver's policy."

trel: all good points/concerns.    how about...

"The default checksum algorithm for iRODS is SHA256, but the actual algorithm used by this function is determined at runtime by the client and server's configuration."

korydraughn: that is much better imo. i think the statement needs to be swapped though. how do you feel about this:

**The checksum algorithm used by this function is determined by the client and server's configuration, but defaults to SHA256.**

lastly, does this information belong in this function's documentation?

trel: yours is better, definitely.

I think it's... helpful to be here.  How about...

note the 'a' in the first sentence.
```
Computes and stores a checksum for a data object. 

The checksum algorithm used by this function is determined by the client and server's configuration, but defaults to SHA256.
```

korydraughn: maybe "and" should replace "but" in my swap example.

korydraughn: i'll use your first suggestion. the more i read it, i like it.

korydraughn: updated based on your comment.

thanks sir!

trel: i think 'but' is fine.

korydraughn:
passed CI and added #

PR #5295, opened 2020-12-10
[4602] skip touching existing database tables on setup (4-2-stable)

trel:


trel:
ping for @bh9

bh9:
Sorry, I assumed that ping was for awareness rather than review, LGTM

Issue #5300, opened 2020-12-10
Stopping iRODS on the host will kill iRODS servers running in a container


korydraughn:
## Bug Report
Given the following setup:
- One iRODS server running directly on the host machine.
- Another iRODS server running inside of a docker container (on the same host machine).

If you attempt to stop the server running directly on the host machine via the `irodsctl` script, the containerized iRODS server will be killed as well.

Stopping the server running directly on the host should not kill servers running inside of a container.

PR #5304, opened 2020-12-11
[#5300] irodsctl no longer kills containerized iRODS processes (4-2-stable)

korydraughn:
The changes have shown to be reliable in the simplest case. However, we need to consider the stop implementation in controller.py.

When stop is executed and the server takes longer than expected to shutdown, controller.py will begin killing processes. This is potentially bad because these changes rely on the root PID out living all descendants. This is just theory at the moment, but it needs to be investigated before merging this PR.

### DIFF
--- a/scripts/irods/controller.py
+++ b/scripts/irods/controller.py
@@ -96,7 +96,7 @@ class IrodsController(object):
                 env=self.config.execution_environment)
 
             root_pid = lib.get_server_root_pid_with_retry()
-            if root_pid == -1:
+            if root_pid == 0:
                 raise IrodsError('Could not retrieve server root PID.')
 
             try_count = 1
@@ -217,13 +217,10 @@ class IrodsController(object):
                 self.config.xmsg_server_executable]
 
         d = {}
-        root_pid = lib.get_server_root_pid()
-
-        if root_pid != -1:
-            for b in binaries:
-                pids = lib.get_pids_executing_binary_file(b, root_pid)
-                if pids:
-                    d[b] = pids
+        for b in binaries:
+            pids = lib.get_pids_executing_binary_file(b, lib.get_server_root_pid())
+            if pids:
+                d[b] = pids
 
         return d
 
--- a/scripts/irods/controller.py
+++ b/scripts/irods/controller.py
@@ -81,8 +81,8 @@ class IrodsController(object):
                     s.bind(('127.0.0.1', irods_port))
                 except socket.error:
                     six.reraise(IrodsError,
-                            IrodsError('Could not bind port {0}.'.format(irods_port)),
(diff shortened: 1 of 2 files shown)


scripts/irods/controller.py
```diff
@@ -213,11 +215,16 @@ def get_binary_to_pids_dict(self, binaries=None):
...
                 self.config.xmsg_server_executable]
+
         d = {}
-        for b in binaries:
-            pids = lib.get_pids_executing_binary_file(b)
-            if pids:
-                d[b] = pids
+        root_pid = lib.get_server_root_pid()
+
+        if pid != -1:
```
trel, line 222: root_pid?

scripts/irods/controller.py
```diff
@@ -213,11 +215,16 @@ def get_binary_to_pids_dict(self, binaries=None):
                 self.config.server_executable,
                 self.config.rule_engine_executable,
                 self.config.xmsg_server_executable]
+
         d = {}
-        for b in binaries:
-            pids = lib.get_pids_executing_binary_file(b)
-            if pids:
-                d[b] = pids
+        root_pid = lib.get_server_root_pid()
```
korydraughn, line 220: this may need to call the `with_retry` version to future proof the code. not sure yet.

korydraughn:
In regards to the stop operation, it may be worth looking into changing how iRODS shuts down. irodsctl will kill processes labeled `irodsServer` first to stop spawning of new processes. This could be avoided by having the server shutdown sockets and the agent factory.

korydraughn:
> In regards to the stop operation, it may be worth looking into changing how iRODS shuts down. irodsctl will kill processes labeled irodsServer first to stop spawning of new processes. This could be avoided by having the server shutdown sockets and the agent factory. The server or script would then need to kill descendants of the root process before killing the actual root server pid (i.e. grandpa).

This is unnecessary. If the root server process dies before its children, the children will be adopted by init (process 1). Because of this adoption, all the controller needs to do is include a check for a ppid of 1. If the immediate parent of the target process is 1, then we know we're looking at the root server process or the root server process has died and the target process has been adopted by init, making it safe to kill.

kript:
The behaviour of an iRODS server used to be that shutting down the server left agent processes running to allow ongoing operations to finish. Is that still part of the design spec, and if so, will it be affected by this work?

trel:
It will not.   This is just about making the selection of which processes to target a bit more discriminating... not changing the logic of how to deal with them.

korydraughn:
This PR needs to consider if the container was launched with an init process. Consider centos systemd enabled images.

trel:
closing b/c 5300 is now master only... yes?

korydraughn:
Correct

Issue #5317, opened 2020-12-29
Does not build against libc++ 11.0 (-Wdeprecated-declarations)


SwooshyCueb:
- [x] master
- [x] 4-2-stable
----
In C++17, `template<> struct std::allocator<void>` and `std::allocator::allocate( size_t n, const void * hint )` were deprecated. libc++ started enforcing this deprecation in 11.0.

Unfortunately, the boost 1.67 headers are littered with references to these, resulting in build failures due to multiple instances of the following errors:
- `error: 'allocator<void>' is deprecated [-Werror,-Wdeprecated-declarations]`
- `error: 'allocate' is deprecated [-Werror,-Wdeprecated-declarations]`

trel:
So... this goes away with a bump to our boost version?

We can/will do that on master for 4.3.0, I presume.

SwooshyCueb:
I'm not sure. I've done a little more research. Let me give you some more background.

In  C++17, `template<> struct std::allocator<void>` was deprecated in favor of `template< class T >struct std::allocator`. In other words, it wasn't the *template initialization* `std::allocator<void>` that was deprecated, but the *explicit template specialization* of `std::allocator<void>`. In C++20, it's removed.  
In C++17, `pointer std::allocator::allocate( size_type n, const void * hint = 0 )` was *replaced* with `T* std::allocate( std::size_t n )` and `T* std::allocator::allocate( std::size_t n, const void * hint)`, the latter of which is deprecated. In effect, the second optional argument of the original function is deprecated. In C++20, the two-argument function is removed completely.

To my understanding, the `std::allocator<void>` template specialization deprecation *should not affect existing code*. To back that up, when, in 2018, MSVC started throwing warnings about `std::allocator<void>` in boost, the boost team decided that this was improper behavior on MSVC's part (see chriskohlhoff/asio#290), and [Microsoft agrees with them but, get this, *decided it wasn't worth fixing*](https://developercommunity.visualstudio.com/solutions/502232/view.html). Therefore, I have good reason to believe that Clang's warnings about `std::allocator<void>` are *also* improper behavior. The official guidance on dealing with this warning in MSVC, specifically, is to define `_SILENCE_CXX17_ALLOCATOR_VOID_DEPRECATION_WARNING`. But that doesn't do anything for us, as libc++ doesn't look for this define and doesn't seem to have anything equivalent.

[...]

Issue #5250, opened 2020-11-17
Delegate more CMake work to irods-dev package


SwooshyCueb:
- [ ] master
- [ ] 4-2-stable
----
There's a lot of [common functionality](https://github.com/irods/irods_capability_publishing/pull/18#issuecomment-726911773) that could be moved to the cmake files provided by `irods-dev`

Additionally, we could benefit from changing our import targets to include the proper [include directories](https://github.com/irods/irods_rule_engine_plugin_update_collection_mtime/issues/21) and such.

trel:
definitely.    consolidate all the things.

SwooshyCueb:
The vast majority of the remaining work for this is likely not going to make it into 4-2-stable, so I'm closing it in favor of #5669, #5712, and #5717

PR #5318, opened 2021-01-01
[#5250,#5317] Introduce FindLibCXX cmake module and conditionally define _LIBCPP_DISABLE_DEPRECATION_WARNINGS (master)

SwooshyCueb:
In my work on identifying the best possible solution/workaround for #5317, I came to the conclusion that due to the impact of the scope of the available workarounds, it would be best to only apply them conditionally, only if the `libc++` version being used is 11+.
As part of the changes for #5250, I was already planning on writing a cmake find module for libc++ which would also identify the version of the library, so some (most) of the commits in this repository are actually in service of #5250.
Also, it seemed appropriate to make the compatibility flags for Clang 10 conditional as well, so I did that.


CMakeLists.txt
```diff
@@ -97,6 +99,37 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND CMAKE_CXX_COMPILER_VERSION VERSIO
...
+  # existing code, libc++ 11 causes warnings for references to
+  # std::allocator<void>. Maybe a future release of libc++ will fix this.
+  # Should we ever switch to C++20 (or newer), std::allocator<void> problem
+  # will go away, as the explicit template specialization is removed in C++20.
+  #
+  # In the meantime, we have a couple of options for dealing with this.
+  # 1. Define the preprocessor variable _LIBCPP_DISABLE_DEPRECATION_WARNINGS.
+  #    This will silence all deprecation warnings from libc++. References to
+  #    deprecated declarations from other sources will still cause a warning
+  #    and hault the build. This solution/workaround could potentially silence
```
trel, line 119: hault -> halt

CMakeLists.txt
```diff
@@ -97,6 +99,37 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND CMAKE_CXX_COMPILER_VERSION VERSIO
...
+  #    valid warnings that would otherwise come from new code that *we*
+  #    introduce.
+  # 2. Pass the compiler option -Wno-error=deprecated-declarations.
+  #    This will prevent references to deprecated declarations from halting
+  #    the build. Warnings will not be silenced, but references to deprecated
+  #    declarations, even those outside of libc++, will not hault the build.
```
trel, line 125: hault -> halt

CMakeLists.txt
```diff
@@ -97,6 +99,37 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND CMAKE_CXX_COMPILER_VERSION VERSIO
...
+  # Both of these options have their downsides, but given that option 2 is
+  # much greater in scope, affecting more than just libc++, option 1 will bez
```
trel, line 127: bez -> be

trel (commented):
typo - "compaitbility flags" in fourth commit message

CMakeLists.txt
SwooshyCueb: I have just now realized that I have probably been spelling this word wrong my entire life.

trel: amazing.

jasoncoposky (approved):
I've already seen Terrell's comments.

trel:
shall we squash before #?   fine either way.

korydraughn (approved):
Everything looks fine.

@SwooshyCueb If I build my own version of libc++, what options would I have to set to use it?

@trel @jasoncoposky Why do we need a separate clang runtime dep when clang already comes with all the necessary libs?

SwooshyCueb:
> @&zwnj;SwooshyCueb If I build my own version of libc++, what options would I have to set to use it?

At present, the easiest way is to set `IRODS_EXTERNALS_FULLPATH_CLANG_RUNTIME` to the installation prefix of your libc++. You can do this at configure time by passing `-DIRODS_EXTERNALS_FULLPATH_CLANG_RUNTIME=[prefix]` to cmake.
The `FindLibCXX` module I've added adds the ability to instead specify `LIBCXX_LIBRARY`, `LIBCXX_INCLUDE_DIR`, and (optionally) `LIBCXX_LIBCXXABI_LIBRARY`, but at present this only works for the main irods repository.
 
> @&zwnj;trel @&zwnj;jasoncoposky Why do we need a separate clang runtime dep when clang already comes with all the necessary libs?

`IRODS_EXTERNALS_FULLPATH_CLANG_RUNTIME` is a bit of a misnomer, as is the package name `irods-externals-clang-runtime`. What we are calling the "clang runtime" is actually libc++ and libc++abi, which, depending on how it was built, may not be required for clang at all. Furthermore, GCC and ICC both support using libc++instead of libstdc++, just as clang does.

SwooshyCueb:
> shall we squash before #? fine either way.

I'm gonna vote no on that. If everything else is fine, I'll go ahead and \#.

trel:
seems good - # it up

SwooshyCueb:
\#'d

Issue #4761, opened 2020-03-03
ilocate should support case-insensitive search, be documented


anderbubble:
- [x] master
- [x] 4-2-stable
-----

iquest supports case-insensitive search, but it appears that the ilocate command does not.

```
$ ilocate -h
Usage: ilocate [options] searchPattern [searchPattern] [searchPattern] ...

Search through the local Zone for data-object(s) with fullpaths with components
which match the provided search pattern.  Use '%' as a 
wildcard (emulates the sql 'like' operand).  If you need to express
% as a literal, you must encapsulate the search pattern with quotes.

Options are :
-0  Separate output via NULL characters instead of newlines (like find -print0)
-t  Also show objects in trash

iRODS Version 4.2.7                ilocate
```

Also, I _have_ an ilocate command, but it does not appear to be documented at https://docs.irods.org/master/icommands/user/

```
$ rpm -qf $(which ilocate)
irods-icommands-4.2.7-1.x86_64
```

trel:
we did not make it default to case-insensitive search:
 - it would be a change in behavior
 - it feels not as unix-y (technical term)

anderbubble:
Great! Thank you for all the hard work! As long as there's an option for it, I'm satisfied. 👍

PR #5331, opened 2021-01-08
[#4761] Added test for ilocate case-insensitive search (master)

korydraughn:
Passed at bench.

CI not required.

Depends on https://github.com/irods/irods_client_icommands/pull/184


korydraughn:
Added #

PR #5332, opened 2021-01-08
[#4761] Added test for ilocate case-insensitive search (4-2-stable)

korydraughn:
Passed at bench.

CI not required.

Depends on https://github.com/irods/irods_client_icommands/pull/185


korydraughn:
Added #

Issue #5224, opened 2020-11-03
leftover 'temporary' rulefile.re.[pid] causing stacktraces once PIDs wrap


trel:
- [x] master
- [x] 4-2-stable
-----

### iRODS Version, OS and Version
4.2.7

### What did you try to do?
with an additional non-core.re rulebase configured (in this example... `training.re`)...

interact with iRODS, creating an Agent, which happens to re-use an old PID

### Expected behavior

no surprises - things just run

### Observed behavior (including steps to reproduce, if applicable)

`SYS_INTERNAL_ERR` sent to the client...

rodsLog has...
```
Nov  3 02:48:23 pid:24762 NOTICE: proc_results_for_rebalance: creating new replica for data id [39104] from [replResc;child1] on [replResc;child2]
terminating with uncaught exception of type boost::filesystem::filesystem_error: boost::filesystem::copy_file: File exists: "/etc/irods/training.re", "/etc/irods/training.re.18125"
Caught signal [6]. Dumping stacktrace and exiting
```

It seems the iRODS server should remove all leftover *.re.[pid] files on startup (based on its currently configured `re_rulebase_set`...   nobody else should be needing them at that time.

stsnel:
@trel I think this issue is also described in https://github.com/irods/irods/issues/4080 . Maybe these issues can be merged?

trel:
Indeed!  I thought we had already fixed this, but couldn't find it when I went looking...

The partial fix is mentioned in that issue:
    https://github.com/tempoz/irods/commits/cache_fix

We'll leave these both open for now as they have a slightly different focus - but they can both be fixed at the same time.

kript:
any chance of a fix in 4.2.9? We get this too.

trel:
yes, quite possible.

trel:
this is handled with a regex removal for now.

a more complete solution (tempfiles only, no actual pidfiles) can occur later with #4080.

Issue #4080, opened 2018-08-25
use memory-only tmpfiles to track irodsAgent PIDs


jasoncoposky:
- [ ] master
- [ ] 4-2-stable
-----

The native rule engine plugin makes a copy of the rule base, which can occasionally collide with a previous copy should one exist.  If the NREP is configured the irodsctl script should clean up any copies of the rule bases in /etc/irods on server stop.

tempoz:
Simply deleting the files when the server is started or stopped is not sufficient to ensure collisions are avoided, as pids can and do wrap during a single execution of the server. We should make a dedicated directory for cached rulebases and use tmpfile to create the actual cache files to avoid collisions.

trel:
Consider:
`/etc/irods/cached_rulebases/core.re.<tmpfilevalue>`

And clear them out on server startup (not stop).

tsmeele:
Collisions indeed occur after Linux process numbers have wrapped:
```
Apr  6 05:59:45 pid:8674 NOTICE: writeLine: inString = Agent process started for puser=rods and cuser=rods from 127.0.0.1

terminating with uncaught exception of type boost::filesystem::filesystem_error: boost::filesystem::copy_file: File exists: "/etc/irods/rules-dgk.re", "/etc/irods/rules-dgk.re.8683"
Caught signal [6]. Dumping stacktrace and exiting
```
`

trel:
partial solution (mentioned above) from two years ago:
  https://github.com/tempoz/irods/commits/cache_fix

trel:
partial solution with #5224 is landed for 4.2.9 - regex removal of any leftover pidfiles on server startup.

we'll pursue a more complete solution with only tempfiles (no actual pidfiles at all) to close this issue.

kript:
As discussed elsewhere, it's not unusual for us to have services remaining running for months, even years, so being unique is a good thing longer term.

korydraughn:
This issue was resolved by #5210.

Rulebase files are stored in memory now.

PR #5333, opened 2021-01-08
[#5224] Server removes leftover rulebase files on startup (master)

korydraughn:
Passed at bench.

### DIFF
--- a/scripts/irods/test/test_misc.py
+++ b/scripts/irods/test/test_misc.py
@@ -10,7 +10,9 @@ else:
 from . import session
 from .. import test
 from .. import lib
+from .. import paths
 from ..configuration import IrodsConfig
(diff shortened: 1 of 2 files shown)


trel:
regex looks good.   this will match and remove `core.re.*` files, yes?   so also a partial fix for #4080?

korydraughn:
yep. it removed the left over core.re.* files and fake.re.* files from my config dir

trel:
cool - please reference it in the commit message alongside 5224.

then we'll # and merge

korydraughn:
i added the issue number to the commit msg. is that what you meant?

trel:
yeps - looks good.   # it.

korydraughn:
Added #

PR #5342, opened 2021-01-11
Fixed memory leaks and ichksum -r output (master)

korydraughn:
This change will likely break code using the `query_processor` due to a change in the interface. Fixing the broken plugins, etc. should be an easy fix.


trel:
looks good to me - were these found via valgrind?

korydraughn:
Yes

trel:
please # when ready

korydraughn:
Added #

PR #5343, opened 2021-01-11
Fixed memory leaks and ichksum -r output (4-2-stable)

korydraughn:
This change will likely break code using the `query_processor` due to a change in the interface. Fixing the broken plugins, etc. should be an easy fix.


korydraughn:
Added #

PR #5344, opened 2021-01-13
Fixed memory leaks and realloc bug in Pack Struct (master)

korydraughn:


trel:
oh my.

korydraughn:
all tests passed except the irodsctl tests.

trel:
thoughts on the irodsctl tests not passing?    is that known?

korydraughn:
I'm not surprised. I've been seeing validation issues on startup quite a bit. I know Jason brought it up a while back as well. I think it has to do with cmake/cpack and how paths are generated. With that said, I don't think it should stop anything from being merged.

trel:
okay - so perhaps some build artifacts.

it's green - let's # em.

korydraughn:
added #

PR #5345, opened 2021-01-13
Fixed memory leaks and realloc bug in Pack Struct (4-2-stable)

korydraughn:


lib/core/src/packStruct.cpp
```diff
@@ -2704,18 +2691,15 @@ unpackPointerItem( packItem_t &myPackedItem, const void *&inPtr,
     }
     case PACK_INT_TYPE:
         if ( myDim == 0 ) {
-            outPtr = addPointerToPackedOut( unpackedOutput,
-                                            numElement * elementSz, NULL );
-            status = unpackIntToOutPtr( inPtr, outPtr, numElement,
-                                        myPackedItem.name, irodsProt );
+            outPtr = addPointerToPackedOut( unpackedOutput, numElement * elementSz, NULL );
+            unpackIntToOutPtr( inPtr, outPtr, numElement, myPackedItem.name, irodsProt );
```
korydraughn, line 2695: This line now mirrors the master branch. Is it a bug in master that the result is not captured in `status`? I'm not sure which implementation is the correct one, but capturing the result in `status` does feel more correct.

lib/core/src/packStruct.cpp
```diff
@@ -2724,18 +2708,15 @@ unpackPointerItem( packItem_t &myPackedItem, const void *&inPtr,
         break;
     case PACK_INT16_TYPE:
         if ( myDim == 0 ) {
-            outPtr = addPointerToPackedOut( unpackedOutput,
-                                            numElement * elementSz, NULL );
-            status = unpackInt16ToOutPtr( inPtr, outPtr, numElement,
-                                          myPackedItem.name, irodsProt );
+            outPtr = addPointerToPackedOut( unpackedOutput, numElement * elementSz, NULL );
+            unpackInt16ToOutPtr( inPtr, outPtr, numElement, myPackedItem.name, irodsProt );
```
korydraughn, line 2712: This line now mirrors the master branch. Is it a bug in master that the result is not captured in `status`? I'm not sure which implementation is the correct one, but capturing the result in `status` does feel more correct.

lib/core/src/packStruct.cpp
```diff
@@ -2744,18 +2725,15 @@ unpackPointerItem( packItem_t &myPackedItem, const void *&inPtr,
         break;
     case PACK_DOUBLE_TYPE:
         if ( myDim == 0 ) {
-            outPtr = addPointerToPackedOut( unpackedOutput,
-                                            numElement * elementSz, NULL );
-            status = unpackDoubleToOutPtr( inPtr, outPtr, numElement,
-                                           myPackedItem.name, irodsProt );
+            outPtr = addPointerToPackedOut( unpackedOutput, numElement * elementSz, NULL );
+            unpackDoubleToOutPtr( inPtr, outPtr, numElement, myPackedItem.name, irodsProt );
```
korydraughn, line 2729: This line now mirrors the master branch. Is it a bug in master that the result is not captured in `status`? I'm not sure which implementation is the correct one, but capturing the result in `status` does feel more correct.

korydraughn:
Tested in CI. Only the irodsctl tests failed (schema validation, shmem creation/cleanup).

lib/core/src/packStruct.cpp
trel: so the comment says don't check status... because it could be negative... that seems... weird/incorrect?

korydraughn: Yes, that is weird. I didn't see that before.

So 4-2-stable should be updated to mirror master? Both PRs passed CI even with this difference.

korydraughn: It looks like the comment is saying to literally not check the captured value because it will be checked later. If that is the case, maybe the return value should be captured like before.

korydraughn: I think I see what is happening. The assignment to `status` is wasted cpu cycles because it is never read after this point. Removing it does not affect the outcome of this function at all.

korydraughn: Notice there are no loops which allow status to be checked later. On top of that, each case statement that cares about the result checks `status` immediately after it is updated.

trel:
green - let's # em.

lib/core/src/packStruct.cpp
trel: agreed.   carry on.

korydraughn:
added #

Issue #5018, opened 2020-07-20
A msiDataObjCreate followed by msiDataObjClose leaves object in intermediate state


JustinKyleJames:
- [x] master
- [x] 4-2-stable
-------

## Bug Report

When running the following rule, the file (f1) was left in an intermediate state.

```
test_open_write_close {{
    msiDataObjCreate("/tempZone/home/rods/f1", "destRescName=demoResc", *fd)
    msiDataObjClose(*fd, *status)

}}
INPUT null
OUTPUT ruleExecOut
```

### iRODS Version, OS and Version

iRODS on 4-2-stable branch
Ubuntu 16.04

### What did you try to do?

Ran the rule listed above.

### Expected behavior

A zero length object (f1) is created and at the end has a good (&) state.

### Observed behavior (including steps to reproduce, if applicable)

The object has an intermediate (?) state.

trel:
This was 4-2-stable?   No intermediates in 4.2.8 ... right?

alanking:
I can't add checkboxes, but this is fixed now. Here is the related test:

https://github.com/irods/irods/blob/d2bf2063c2992da52c7ee234bb381fbb4de5de80/unit_tests/src/test_rc_data_obj.cpp#L28-L51

Should I add one specifically for the microservice before we close?

trel:
Yes, please add a microservice test as well - thanks.

alanking:
Cannot checkbox, but tests are added

trel:
done - thanks.

PR #5346, opened 2021-01-14
[#5018] add msiDataObjCreate/Close test (master)

alanking:
Adds a test which creates a data object and closes it without writing to
it using the microservices in a rule. The resultant replica should be
marked as good. The python version of this rule was implemented but not
tested.


trel:
if you're happy, # it up.

alanking:
#'d

PR #5355, opened 2021-01-19
[#5354] Handle missing unversioned libc++ symlink in FindLibCXX.cmake (master)

### LINKED ISSUE #5354: Compiled binaries do not include the path to libc++ shared libraries in the rpath
- [x] master
- [x] 4-2-stable
---
## Bug Report
This needs to be set to `${IRODS_EXTERNALS_FULLPATH_CLANG_RUNTIME}/lib` given that the test hook does not install the clang compiler and the server doesn't depend on the compiler.
https://github.com/irods/irods/blob/ee116cbc7cf90f6b0960e596b0b7ebf97877393b/CMakeLists.txt#L80

The following line needs to include `${IRODS_EXTERNALS_FULLPATH_CLANG_RUNTIME}/lib/libc++.so` and `${IRODS_EXTERNALS_FULLPATH_CLANG_RUNTIME}/lib/libc++abi.so`.
https://github.com/irods/irods/blob/ee116cbc7cf90f6b0960e596b0b7ebf97877393b/CMakeLists.txt#L73

SwooshyCueb:
@korydraughn discovered that our irods-externals package for libc++ lacks unversioned symlinks for libc++ and libc++abi, which was causing FindLibCXX to miss them and choose the libraries that lived in our clang irods-externals package.

This PR fixes the issue by having `find_library` additionally search for the verioned library symlink when looking in the irods-externals libc++ prefix.

Fixes #5354 for master.
Touches #5250.

Currently pushing through jenkins

### DIFF
--- a/cmake/Modules/FindLibCXX.cmake
+++ b/cmake/Modules/FindLibCXX.cmake
@@ -36,6 +36,17 @@ The following cache variables may also be set:
 
 get_filename_component(__find_libcxx_dir "${CMAKE_CURRENT_LIST_FILE}" PATH)
 
+# We need to set some variables to manipulate find_library behavior
+set(PRE_CMAKE_FIND_LIBRARY_PREFIXES "${CMAKE_FIND_LIBRARY_PREFIXES}")
+set(PRE_CMAKE_FIND_LIBRARY_SUFFIXES "${CMAKE_FIND_LIBRARY_SUFFIXES}")
+# These libraries always start with lib, even on platforms where libraries
+# usually do not.
+set(CMAKE_FIND_LIBRARY_PREFIXES "lib")
+# our externals package for libc++ might not have unversioned so symlinks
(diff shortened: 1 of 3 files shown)


cmake/Modules/FindLibCXX.cmake
```diff
@@ -36,6 +36,17 @@ The following cache variables may also be set:
 
 get_filename_component(__find_libcxx_dir "${CMAKE_CURRENT_LIST_FILE}" PATH)
 
+# We need to set some variables to manipualte find_library behavior
```
trel, line 39: "manipualte" -> "manipulate"

trel (commented):
5317 mentioned in description, but not in commit messages

korydraughn (approved):
Looks fine to me.

trel:
good good - # it when happy

SwooshyCueb:
Had several tests fail in jenkins, but I can't reproduce any of the failures in my docker containers.
\#'d

PR #5357, opened 2021-01-20
[#5356,#5361,#5363] script output fixes (master)

### LINKED ISSUE #5356: python: execute.execute_command can throw AttributeError "'file' object has no attribute 'splitlines'" when foreground=True
In trying to diagnose a problem in my docker container, I have run into a bug in `execute.py`.

Running `irodsctl` with the `--stdout` argument throws an exception:
```
  File "/var/lib/irods/scripts/irods_control.py", line 127, in <module>
    sys.exit(main())
  File "/var/lib/irods/scripts/irods_control.py", line 116, in main
    operations_dict[operation]()
  File "/var/lib/irods/scripts/irods_control.py", line 70, in <lambda>
    operations_dict['start'] = lambda: irods_controller.start(write_to_stdout=options.write_to_stdout, test_mode=options.test_mode)
  File "/var/lib/irods/scripts/irods/controller.py", line 131, in start
    env=self.config.execution_environment)
  File "/var/lib/irods/scripts/irods/execute.py", line 128, in execute_command
    check_command_return(args, out, err, returncode, **kwargs)
  File "/var/lib/irods/scripts/irods/execute.py", line 122, in check_command_return
    indent(out),
  File "/var/lib/irods/scripts/irods/execute.py", line 22, in indent
    for lines in text])
AttributeError: 'file' object has no attribute 'splitlines'
```

The `--stdout` argument causes `foreground=True` to be passed to `execute_command`:
https://github.com/irods/irods/blob/ee116cbc7cf90f6b0960e596b0b7ebf97877393b/scripts/irods/controller.py#L119-L131
`execute_command` passes it down to `execute_command_permissive`:
https://github.com/irods/irods/blob/ee116cbc7cf90f6b0960e596b0b7ebf97877393b/scripts/irods/execute.py#L126-L127
Normally, `execute_command_permissive` returns two string objects and a return code, but when `foreground` is `True`, it returns two *file objects* and a return code:
https://github.com/irods/irods/blob/ee116cbc7cf90f6b0960e596b0b7ebf97877393b/scripts/irods/execute.py#L95-L105
Those two file objects are then passed into `check_command_return`:
https://github.com/irods/irods/blob/ee116cbc7cf90f6b0960e596b0b7ebf97877393b/scripts/irods/execute.py#L127-L128
`check_command_return` is not set up to handle file objects, so if the process returned a nonzero error code, things go sideways.
https://github.com/irods/irods/blob/ee116cbc7cf90f6b0960e596b0b7ebf97877393b/scripts/irods/execute.py#L114-L124

Additionally, the foreground loop never reads from `p.stderr`. That's probably not great.
https://github.com/irods/irods/blob/ee116cbc7cf90f6b0960e596b0b7ebf97877393b/scripts/irods/execute.py#L97-L101

This only affects master. Stable does not have the relevant functionality.

### LINKED ISSUE #5361: setup_irods.py: log level and days per log arguments not converted to string
- [ ] master
- [ ] 4-2-stable
----
Both `setup_irods.py` and `irods_control.py` accept arguments as defined in `irods/start_options.py`.

However, `irods_control.py` stringifies its log level and days per log arguments:
https://github.com/irods/irods/blob/ee116cbc7cf90f6b0960e596b0b7ebf97877393b/scripts/irods_control.py#L102-L107
and `setup_irods.py` does not:
https://github.com/irods/irods/blob/ee116cbc7cf90f6b0960e596b0b7ebf97877393b/scripts/setup_irods.py#L428-L433

This causes python to get angry when these arguments are passed to `setup_irods.py` because of non-string values in the subprocess environment dict.

SwooshyCueb:
Fixes #5356
Fixes #5361 for master
Addresses the main concern of #5363 for master

### DIFF
--- a/scripts/irods/controller.py
+++ b/scripts/irods/controller.py
@@ -126,7 +126,8 @@ class IrodsController(object):
                     cmd.append('-t')
 
                 lib.execute_command(cmd,
-                                    foreground=True,
+                                    stdout=sys.stdout,
+                                    stderr=sys.stdout,
                                     cwd=self.config.server_bin_directory,
                                     env=self.config.execution_environment)
             else:
--- a/scripts/irods/execute.py
+++ b/scripts/irods/execute.py
@@ -16,10 +16,8 @@ def indent(*text, **kwargs):
     if 'indentation' in kwargs:
         indentation = kwargs['indentation']
     else:
-        indentation='  '
-    return '\n'.join([
-        ''.join([indentation, '\n{0}'.format(indentation).join(lines.splitlines())])
-            for lines in text])
+        indentation = '  '
+    return '\n'.join([''.join([indentation, '\n{0}'.format(indentation).join(lines.splitlines())]) for lines in text])
 
 def safe_shlex_split_for_2_6(args):
     if not isinstance(args, str) and isinstance(args, six.text_type):
@@ -28,7 +26,7 @@ def safe_shlex_split_for_2_6(args):
 
 def communicate_and_log(p, args, input=None):
     l = logging.getLogger(__name__)
-    out, err = [t.decode('utf_8') for t in p.communicate(input=(input.encode('ascii') if input is not None else None))]
+    out, err = [(None if t is None else t.decode('utf_8')) for t in p.communicate(input=(None if input is None else input.encode('ascii')))]
     message = ['Command %s returned with code %s.' % (args, p.returncode)]
     if input:
         message.append('stdin:\n%s' % indent(input))
@@ -49,10 +47,10 @@ def execute_command_nonblocking(args, stdout=subprocess.PIPE, stderr=subprocess.
     if 'env' in kwargs:
         kwargs_without_env = copy.copy(kwargs)
         kwargs_without_env['env'] = 'HIDDEN'
-    else :
+    else:
         kwargs_without_env = kwargs
     l.debug('Calling %s with options:\n%s', args, pprint.pformat(kwargs_without_env))
-    try :
+    try:
         return subprocess.Popen(args, **kwargs)
     except OSError as e:
         six.reraise(IrodsError,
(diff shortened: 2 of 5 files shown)


scripts/irods/controller.py
```diff
@@ -126,7 +126,8 @@ def start(self, write_to_stdout=False, test_mode=False):
                     cmd.append('-t')
 
                 lib.execute_command(cmd,
-                                    foreground=True,
+                                    stdout=sys.stdout,
```
trel, line 129: what was the foreground used for?    are we losing that functionality?

scripts/irods/execute.py
```diff
@@ -80,51 +78,42 @@ def execute_command_timeout(args, timeout=10, **kwargs):
...
-def check_command_return(args, out, err, returncode, **kwargs):
+def check_command_return(args, out, err, returncode, input=None, **kwargs):
     if returncode is not None and returncode != 0:
         if 'env' in kwargs:
             kwargs_without_env = copy.copy(kwargs)
             kwargs_without_env['env'] = 'HIDDEN'
-        else :
+        else:
             kwargs_without_env = kwargs
+        deets = [
```
trel, line 100: details?

scripts/irods/start_options.py
```diff
@@ -1,10 +1,10 @@
 def add_options(parser):
     parser.add_option('-q', '--quiet',
-                      dest='verbose', action='store_false',
+                      dest='verbose', action='store_const', const=0,
```
trel, line 3: i'm not familiar with store_const ... so does `-q` still do what it did?   or do we now just set a numeric value for `-v`?

scripts/irods/controller.py
SwooshyCueb: The `foreground` argument would be passed down to `execute_command_permissive`. If set to True, it would run a loop to manually pipe the subprocess's stdout to the script's stdout.

I've removed this in favor of just passing the script's stdout directly to `subprocess.Popen`, which does the same thing without risk of blocking because of stuff like stderr filling up.

scripts/irods/start_options.py
SwooshyCueb: The behavior of `-q` is effectively the same.

I've changed the `verbose` variable from a boolean to an integer.
You can pass `-v` multiple times, and each time increments `verbose`, allowing for additional levels of verbosity.
`-q` just explicitly sets `verbose` to 0.

scripts/irods/execute.py
SwooshyCueb: We say "opts" for "options" so I figure "deets" for "details" is fair play

scripts/irods/controller.py
korydraughn: Does the output from the Popen'd process still come out as soon as it is available? I believe the foreground option was implemented that way because the output from the irodsServer binary wasn't appearing immediately.

SwooshyCueb: I think so, but I haven't tested that specifically.
My understanding is that the `communicate` call in `communicate_and_log` blocks until the subprocess terminates, so I can force a flush after just to make sure.

korydraughn: We need to verify that the output from grandpa and any forked processes is printed to the console as soon as it is available. The `--stdout` option will cause the `irodsctl` script to never return. The loop/poll/flush code allowed messages to be seen as the server was running.

SwooshyCueb: I'll do some more testing with this.

scripts/irods/start_options.py
trel: ah, nice.   and if they're both passed, which wins?

scripts/irods/execute.py
trel: i'll allow it - don't tell anyone.

scripts/irods/controller.py
SwooshyCueb: I've done some more testing just to double check, ant it looks like the server's stdout is going directly to the console.

SwooshyCueb:
Any objections to \#ing? @trel @korydraughn

trel:
either need to add issue number(s) to the first two commits... or squash... so lets get that done first, then # after one last look.

korydraughn:
Looks good to me.

SwooshyCueb:
Opened 5361 and 5363.
I've covered more in 5363 than is addressed by this PR. I feel there's some decisions to be made about `--quiet` but it's not as important as the rest of the changes in this PR.

scripts/irods/start_options.py
SwooshyCueb: I've preserved the previous behavior of `-q` cancelling out any `-v` that comes before in the argument list.

trel:
cool - let's # em up.

these are pointed at master for now.    let's discuss whether any/all of these get cherry-picked for 4-2-stable.

SwooshyCueb:
\#'d

I'm in favor of cherry-picking the first two commits into stable, the second at the very least.

trel:
agreed - please cherry-pick the top two - thanks

PR #5364, opened 2021-01-22
[#5354] Handle missing unversioned libc++ symlink in FindLibCXX.cmake (4-2-stable)

### LINKED ISSUE #5354: Compiled binaries do not include the path to libc++ shared libraries in the rpath

SwooshyCueb:
Cherry-pick of #5355 into stable

Fixes #5354 for stable.
Touches #5250.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,7 +76,19 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS TRUE)
 include(${CMAKE_SOURCE_DIR}/cmake/InstallDirs.cmake)
 
 set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
-set(CMAKE_INSTALL_RPATH "${LIBCXX_LIBRARY_DIR}")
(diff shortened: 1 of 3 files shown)


trel:
looks good - let's # it.

SwooshyCueb:
\#'d

Issue #5362, opened 2021-01-22
trim/rm leaves registered 'in-vault' replica on disk


trel:
- [x] master
- [x] 4-2-stable
-----

### iRODS Version, OS and Version

4.2.8

### What did you try to do?

we know that 'non-vault' replicas are left on disk (as per #4848 in 4.2.8)

tried to show that an 'in-vault' registered replica would be removed from disk on trim.

### Expected behavior

the replica should have been removed from disk.

### Observed behavior (including steps to reproduce, if applicable)

it was still there.

```
$ # create a couple files in nested directories, simulating a bgfs mountpoint
$ mkdir /tmp/thebgfsmount -p
$ mkdir -p /tmp/thebgfsmount/dir1/dir11/
$ echo "file111" > /tmp/thebgfsmount/dir1/dir11/file111
$ mkdir -p /tmp/thebgfsmount/dir2/dir22/
$ echo "file222" > /tmp/thebgfsmount/dir2/dir22/file222

$ # create bgfsResc
$ iadmin mkresc bgfsResc unixfilesystem $(hostname):/tmp/thebgfsmount/
Creating resource:
Name:           "bgfsResc"
Type:           "unixfilesystem"
Host:           "xxx.xxx.xxx.xxx"
Path:           "/tmp/thebgfsmount/"
Context:        ""

$ # recursively register some existing files into bgfsResc
$ ireg -C -R bgfsResc /tmp/thebgfsmount/ /tempZone/home/rods/stuffinhere/bgfs

$ # show the registered data objects, their resource, and their physical paths
$ # note there has been no data movement
$ ils -rL stuffinhere
/tempZone/home/rods/stuffinhere:
  C- /tempZone/home/rods/stuffinhere/bgfs
/tempZone/home/rods/stuffinhere/bgfs:
  C- /tempZone/home/rods/stuffinhere/bgfs/dir1
/tempZone/home/rods/stuffinhere/bgfs/dir1:
  C- /tempZone/home/rods/stuffinhere/bgfs/dir1/dir11
/tempZone/home/rods/stuffinhere/bgfs/dir1/dir11:
  rods              0 bgfsResc            8 2021-01-22.13:47 & file111
        generic    /tmp/thebgfsmount/dir1/dir11/file111
  C- /tempZone/home/rods/stuffinhere/bgfs/dir2
/tempZone/home/rods/stuffinhere/bgfs/dir2:
[...]

PR #5375, opened 2021-01-24
[#5362] Added test that shows in-vault registered replicas are removed on trim (master)

korydraughn:


trel:
looks good to me - and removing code!    please # it.

bonus points if we can figure out when it was fixed

korydraughn:
added #.

I'll add it to my notes to track down when it was fixed.

PR #5376, opened 2021-01-24
[#5362] Added test that shows in-vault registered replicas are removed on trim (4-2-stable)

korydraughn:


trel:
looks good - please #

korydraughn:
added #

PR #5393, opened 2021-02-03
[#5250] cmake: clean up risidual explicit libc++ references, unset cached LIBCXX version variables if __config changes (master)

SwooshyCueb:
Loads of the catch2 unit tests had explicit linkages to libc++abi. Typically harmless, but no longer needed.
The `irods_post_install_test` project had explicit libc++ includes, which overrode `LIBCXX_INCLUDE_DIR`.

~~If `IRODS_EXTERNALS_FULLPATH_CLANG_RUNTIME` or `IRODS_EXTERNALS_FULLPATH_CLANG` changes, unset cached `LIBCXX_INCLUDE_DIR`, `LIBCXX_LIBRARY`, and `LIBCXX_LIBCXXABI_LIBRARY` so they can be derived with the new paths.~~ Turns out, doing this will also cause CMake to ignore values passed in on the command line. The CMake documentation leads one to believe that cache variables can be unset while preserving regular variables with the same names, but this is not the case.
If the MD5 hash of `${LIBCXX_INCLUDE_DIR}/__config` changes, unset cached `HAVE_LIBCXX_VERSION_INT`, `LIBCXX_VERSION`, and `LIBCXX_VERSION_IN`, so that they can be derived with the new file.

in service of #5250


trel:
very nice -  logic seems good to me.    

please   # when ready

SwooshyCueb:
\#'d

SwooshyCueb:
actually, hang on, don't merge this just yet, I think I broke something

trel:
will await word.

SwooshyCueb:
It was just one of my build scripts acting up, false alarm

PR #5395, opened 2021-02-03
[5392] added additional hash algorithms

JustinKyleJames:
see [irods/irods_client_globus_connector#9]

sha512
sha1
adler32

### DIFF
--- a/lib/hasher/include/ADLER32Strategy.hpp
+++ b/lib/hasher/include/ADLER32Strategy.hpp
@@ -1,17 +1,17 @@
-#ifndef _SHA256_STRATEGY_HPP_
-#define _SHA256_STRATEGY_HPP_
+#ifndef _ADLER32_STRATEGY_HPP_
+#define _ADLER32_STRATEGY_HPP_
 
 #include "HashStrategy.hpp"
 
 namespace irods {
-    extern const std::string SHA256_NAME;
-    class SHA256Strategy : public HashStrategy {
+    extern const std::string ADLER32_NAME;
+    class ADLER32Strategy : public HashStrategy {
         public:
-            SHA256Strategy() {};
-            virtual ~SHA256Strategy() {};
+            ADLER32Strategy() {};
+            virtual ~ADLER32Strategy() {};
 
-            virtual std::string name() const override {
-                return SHA256_NAME;
+            std::string name() const override {
+                return ADLER32_NAME;
             }
             error init( boost::any& context ) const override;
             error update( const std::string& data, boost::any& context ) const override;
@@ -21,4 +21,4 @@ namespace irods {
     };
 } // namespace irods
 
-#endif // _SHA256_STRATEGY_HPP_
+#endif // _ADLER32_STRATEGY_HPP_
--- a/lib/hasher/include/SHA512Strategy.hpp
+++ b/lib/hasher/include/SHA512Strategy.hpp
@@ -1,17 +1,17 @@
-#ifndef _SHA256_STRATEGY_HPP_
-#define _SHA256_STRATEGY_HPP_
+#ifndef _SHA512_STRATEGY_HPP_
+#define _SHA512_STRATEGY_HPP_
 
 #include "HashStrategy.hpp"
 
 namespace irods {
-    extern const std::string SHA256_NAME;
-    class SHA256Strategy : public HashStrategy {
+    extern const std::string SHA512_NAME;
+    class SHA512Strategy : public HashStrategy {
         public:
-            SHA256Strategy() {};
-            virtual ~SHA256Strategy() {};
+            SHA512Strategy() {};
+            virtual ~SHA512Strategy() {};
 
-            virtual std::string name() const override {
-                return SHA256_NAME;
+            std::string name() const override {
+                return SHA512_NAME;
             }
             error init( boost::any& context ) const override;
             error update( const std::string& data, boost::any& context ) const override;
@@ -21,4 +21,4 @@ namespace irods {
     };
 } // namespace irods
 
-#endif // _SHA256_STRATEGY_HPP_
+#endif // _SHA512_STRATEGY_HPP_
--- a/lib/hasher/include/checksum.hpp
+++ b/lib/hasher/include/checksum.hpp
@@ -5,9 +5,9 @@
 #ifndef CHECKSUM_HPP_
 #define CHECKSUM_HPP_
 
-#include <stdio.h>
-#include <time.h>
-#include <string.h>
+#include <cstdio>
+#include <ctime>
+#include <cstring>
 #include "objInfo.h"
 
 #ifdef __cplusplus
@@ -15,6 +15,9 @@ extern "C" {
 #endif
 
 #define SHA256_CHKSUM_PREFIX "sha2:"
+#define SHA512_CHKSUM_PREFIX "sha512:"
+#define ADLER32_CHKSUM_PREFIX "adler32:"
+#define SHA1_CHKSUM_PREFIX "sha1:"
 int verifyChksumLocFile( char *fileName, const char *myChksum, char *chksumStr );
 
 int
--- a/lib/hasher/src/ADLER32Strategy.cpp
+++ b/lib/hasher/src/ADLER32Strategy.cpp
@@ -1,0 +1,79 @@
+#include "ADLER32Strategy.hpp"
+#include "checksum.hpp"
+
+#include <sstream>
+#include <iostream>
+#include <iomanip>
+#include <cstring>
+#include <boost/algorithm/string/predicate.hpp>
+
+#include "base64.h"
+
+namespace irods {
+
+    const std::string ADLER32_NAME( "adler32" );
+
+    struct adler32_parts {
+        uint32_t a;
+        uint32_t b;
+    };
+
+    adler32_parts adler32_init() {
+        return adler32_parts{1, 0};
+    }
+
+    static adler32_parts adler32_update(const adler32_parts& parts, const unsigned char *data, size_t len) {
+
+        const uint32_t MOD_ADLER = 65521;
+
+        uint32_t a = parts.a, b = parts.b;
+
+        // Process each byte of the data in order
+        for (size_t index = 0; index < len; ++index)
+        {
+            a = (a + data[index]) % MOD_ADLER;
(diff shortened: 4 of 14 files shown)


lib/hasher/src/ADLER32Strategy.cpp
```diff
@@ -0,0 +1,79 @@
+#include "ADLER32Strategy.hpp"
+#include "checksum.hpp"
+
+#include <string>
+#include <sstream>
+#include <iostream>
+#include <iomanip>
+#include <string.h>
+#include <boost/algorithm/string/predicate.hpp>
+//#include <zlib.h>
```
trel, line 10: remove if unneeded

lib/hasher/include/checksum.hpp
```diff
@@ -15,6 +15,9 @@ extern "C" {
 #endif
 
 #define SHA256_CHKSUM_PREFIX "sha2:"
+#define SHA512_CHKSUM_PREFIX "sha-512:"
+#define ADLER32_CHKSUM_PREFIX "adler-32:"
```
trel, line 19: remove the hyphens?   is there a reason we need them?

server/api/src/rsFileChksum.cpp
```diff
@@ -165,7 +165,7 @@ int fileChksum(
 
     rodsLog(
         LOG_DEBUG,
-        "fileChksum :: final_scheme [%s]  chkstr_scheme [%s]  svr_hash_policy [%s]  hash_policy [%s]",
+        "fileChksum :: final_scheme [%s]  chkstr_scheme [%s]  hash_policy [%s]",
```
trel, line 168: weird that this was just... wrong

lib/hasher/src/ADLER32Strategy.cpp
```diff
@@ -0,0 +1,79 @@
+#include "ADLER32Strategy.hpp"
+#include "checksum.hpp"
+
+#include <string>
+#include <sstream>
+#include <iostream>
+#include <iomanip>
+#include <string.h>
```
korydraughn, line 8: given that this is a cpp file, this should be `<cstring>`.

lib/hasher/include/SHA512Strategy.hpp
```diff
@@ -0,0 +1,26 @@
+#ifndef _SHA512_STRATEGY_HPP_
+#define _SHA512_STRATEGY_HPP_
+
+#include "HashStrategy.hpp"
+#include <string>
+#include <openssl/sha.h>
+
+namespace irods {
+    const std::string SHA512_NAME( "sha512" );
```
korydraughn, line 9: This will result in multiple unique symbols appearing in the shared libs. This line should be replaced with a declaration:
```c++
extern const std::string SHA512_NAME;
```
The definition should appear in the cpp file like so:
```c++
const std::string SHA512_NAME( "sha512" )
```
These changes need to be applied for all of the hashing hpp and cpp files.

lib/hasher/include/SHA512Strategy.hpp
```diff
@@ -0,0 +1,26 @@
...
+            SHA512Strategy() {};
+            virtual ~SHA512Strategy() {};
+
+            virtual std::string name() const {
+                return SHA512_NAME;
+            }
+            virtual error init( boost::any& context ) const;
+            virtual error update( const std::string& data, boost::any& context ) const;
+            virtual error digest( std::string& messageDigest, boost::any& context ) const;
+            virtual bool isChecksum( const std::string& ) const;
```
korydraughn, line 21: If all of these functions override a virtual function from the base class, you can remove the `virtual` keyword and append the `override` keyword to each one. For example:
```c++
error init(boost::any& context) const override;
```

lib/hasher/src/ADLER32Strategy.cpp
```diff
@@ -0,0 +1,79 @@
...
+        uint32_t a = parts.a, b = parts.b;
+
+		// Process each byte of the data in order
+		for (size_t index = 0; index < len; ++index)
+		{
+			a = (a + data[index]) % MOD_ADLER;
+			b = (b + a) % MOD_ADLER;
+		}
+
+        return adler32_parts{a, b};
```
korydraughn, line 38: indentation

lib/hasher/src/SHA1Strategy.cpp
```diff
@@ -0,0 +1,49 @@
+#include "SHA1Strategy.hpp"
+#include "checksum.hpp"
+
+#include <string>
+#include <sstream>
+#include <iostream>
+#include <iomanip>
+#include <string.h>
```
korydraughn, line 8: <cstring>

lib/hasher/src/SHA512Strategy.cpp
```diff
@@ -0,0 +1,49 @@
+#include "SHA512Strategy.hpp"
+#include "checksum.hpp"
+
+#include <string>
+#include <sstream>
+#include <iostream>
+#include <iomanip>
+#include <string.h>
```
korydraughn, line 8: <cstring>

lib/hasher/include/ADLER32Strategy.hpp
```diff
@@ -0,0 +1,26 @@
+#ifndef _ADLER32_STRATEGY_HPP_
+#define _ADLER32_STRATEGY_HPP_
+
+#include "HashStrategy.hpp"
+#include <string>
+#include <openssl/sha.h>
```
korydraughn, line 6: If `ADLER32Strategy.hpp` doesn't depend on any `<openssl/sha.h>` symbols, then this `#include` can be moved into the cpp file.

lib/hasher/src/ADLER32Strategy.cpp
```diff
@@ -0,0 +1,79 @@
+#include "ADLER32Strategy.hpp"
+#include "checksum.hpp"
+
+#include <string>
```
korydraughn, line 4: This header can be removed because it is already included by `ADLER32Strategy.hpp`.

This should only apply to hpp/cpp pairs.

lib/hasher/src/SHA512Strategy.cpp
korydraughn: `<cstring>`

server/api/src/rsFileChksum.cpp
JustinKyleJames: Yes, I ran into this a while back when I enabled debug logging.

lib/hasher/include/checksum.hpp
JustinKyleJames: Not necessarily.  The hyphen is the normal terminology for them and I used it to distinguish it from sha2 which is also known as sha-256 when it references a specific algorithm rather than the family of sha-2 algorithms.

However, it is known as "sha-2" and we use "sha2" so it makes sense to keep this consistent.  I will remove the hashes.

lib/hasher/src/ADLER32Strategy.cpp
JustinKyleJames: Done

lib/hasher/include/checksum.hpp
JustinKyleJames: Done

lib/hasher/src/ADLER32Strategy.cpp
JustinKyleJames: Done for all algorithms.

lib/hasher/include/SHA512Strategy.hpp
JustinKyleJames: Done for all hashing schemes.

lib/hasher/include/SHA512Strategy.hpp
JustinKyleJames: Done for all hashing schemes.

lib/hasher/src/ADLER32Strategy.cpp
JustinKyleJames: I had accidentally copied in some tab chars which looked fine with my tabstop setting.  Replaced with four spaces.

lib/hasher/src/SHA512Strategy.cpp
JustinKyleJames: done

lib/hasher/src/SHA1Strategy.cpp
JustinKyleJames: done

lib/hasher/include/ADLER32Strategy.hpp
JustinKyleJames: Done for all hash strategies.

trel:
okay, push the new commits and we'll look over it again and resolve the conversations

lib/hasher/src/ADLER32Strategy.cpp
JustinKyleJames: done

lib/hasher/include/ADLER32Strategy.hpp
```diff
@@ -0,0 +1,24 @@
...
+            std::string name() const override {
+                return ADLER32_NAME;
+            }
+            error init( boost::any& context ) const override;
+            error update( const std::string& data, boost::any& context ) const override;
+            error digest( std::string& messageDigest, boost::any& context ) const override;
+            bool isChecksum( const std::string& ) const override;
+
+    };
+}; // namespace irods
```
korydraughn, line 22: namespaces do not require a semicolon

korydraughn: this applies to the other files too.

JustinKyleJames: Dude, you are killing me.  These were just copied.  ;-)

JustinKyleJames: Done

trel:
Looks good - when this is working as expected, please squash down to a single commit without the # and we'll check it one more time.

JustinKyleJames:
> Looks good - when this is working as expected, please squash down to a single commit without the # and we'll check it one more time.

Done.

trel:
please # it

JustinKyleJames:
> please # it

done

PR #5403, opened 2021-02-09
DNS caching and Parallel Transfer + hosts_config.json (4-2-stable)

korydraughn:


lib/core/src/hostname_cache.cpp
```diff
@@ -0,0 +1,193 @@
...
+        return inserted;
+    } // hnc_insert_or_assign
+
+    auto hnc_lookup(const std::string_view _key) -> std::optional<std::string>
+    {
+        bi::scoped_lock lk{*g_mutex};
+
+        if (auto iter = g_map->find(key_type{_key.data(), *g_allocator}); iter != g_map->end()) {
+            if (clock_type::now().time_since_epoch().count() < iter->second.expiration) {
+                iter->second.expiration = clock_type::now().time_since_epoch().count() + iter->second.expires_after;
```
korydraughn, line 145: needs duration_cast to seconds.

lib/core/src/hostname_cache.cpp
```diff
@@ -0,0 +1,193 @@
...
+    {
+        bi::scoped_lock lk{*g_mutex};
+        g_map->erase(key_type{_key.data(), *g_allocator});
+    } // hnc_erase
+
+    auto hnc_erase_expired_entries() -> void
+    {
+        bi::scoped_lock lk{*g_mutex};
+
+        const auto now = clock_type::now().time_since_epoch().count();
```
korydraughn, line 163: duration_cast to seconds (see dns_cache.cpp).

lib/core/src/hostname_cache.cpp
```diff
@@ -0,0 +1,193 @@
...
+        catch (...) {}
+    } // hnc_deinit
+
+    auto hnc_insert_or_assign(const std::string_view _key,
+                              const std::string_view _alias,
+                              std::chrono::seconds _expires_after) -> bool
+    {
+        bi::scoped_lock lk{*g_mutex};
+        key_type key{_key.data(), *g_allocator};
+        const auto expiration = (clock_type::now() + _expires_after).time_since_epoch().count();
```
korydraughn, line 133: duration_cast to seconds.

lib/core/src/rcMisc.cpp
```diff
@@ -4594,12 +4607,42 @@ int get_canonical_name(const char *_hostname, char* _buf, size_t _len) {
...
+        rodsLog(LOG_DEBUG, "%s :: Caching addrinfo for [hostname=%s].", __func__, _hostname);
+
+        //try {
+            const auto age = irods::get_server_property<std::chrono::seconds>(irods::CFG_EVICTION_AGE_IN_SECONDS_KW);
+            net::dnsc_insert_or_assign(target, *p_addrinfo, age);
+        //}
+        //catch (...) {
+            // TODO Should the default be specified in the catch statement.
+            //net::dnsc_insert_or_assign(target, *p_addrinfo, std::chrono::seconds{30});
+        //}
```
korydraughn, line 4623: the outcome of these lines depend on how we want to set the eviction property. see rodsServer.cpp.

lib/core/src/rcMisc.cpp
```diff
@@ -4608,12 +4651,27 @@ int load_in_addr_from_hostname(const char* _hostname, struct in_addr* _out) {
...
```
korydraughn, line 4667: the outcome of these lines depend on how we want to set the eviction property. see rodsServer.cpp.

lib/core/src/sockComm.cpp
```diff
@@ -393,9 +385,23 @@ sockOpenForInConn( rsComm_t *rsComm, int *portNum, char **addr, int proto ) {
...
+        }
+        else {
+            *addr = static_cast<char*>(malloc(sizeof(char) * LONG_NAME_LEN));
+            gethostname(*addr, LONG_NAME_LEN);
+
+            const auto alias = resolve_hostname_from_hosts_config(*addr);
+            net::hnc_insert_or_assign("localhost", alias, std::chrono::seconds{20});
+
+            rstrcpy(*addr, alias.data(), LONG_NAME_LEN);
+        }
```
korydraughn, line 404: do we see any need for the hostname cache here?

`resolve_hostname_from_hosts_config` will read the config file once for every agent. imo, there isn't much gained from using the hostname cache. maybe, this is a little faster given it is just a lookup once cached.

trel: on the server, yes?    anything to reduce network traffic is a speedup.

on the client... no.

korydraughn: Well, that's the thing. I don't think this will affect network traffic. gethostname() is likely just reading a local file.

korydraughn: On a different note, this code returns the first entry from the hosts_config.json.

lib/core/src/sockComm.cpp
```diff
@@ -393,9 +388,44 @@ sockOpenForInConn( rsComm_t *rsComm, int *portNum, char **addr, int proto ) {
...
+
+                if (alias) {
+                    std::strncpy(*addr, alias->data(), alias->size());
+                }
+
+                const auto age = irods::get_server_property<std::chrono::seconds>(
+                    irods::configuration_parser::key_path_t{irods::CFG_HOSTNAME_CACHE_KW,
+                                                            irods::CFG_EVICTION_AGE_IN_SECONDS_KW});
+
+                const auto inserted = net::hnc_insert_or_assign("localhost", *addr, age);
```
korydraughn, line 420: `*addr` contains the hostname obtained via `gethostname()`. The code presented will map `"localhost"` to the value of `gethostname()` if `resolve_hostname()` does not return an alias. This is so that the cache miss message is not logged on every invocation of this function.

Are we okay with this?

trel: This is running and being cached per server... so, yes, I think this is okay.

korydraughn: cool

korydraughn:
This implementation does not support the multiple readers / single writer scenario. This needs to be investigated and modified to support that.

server/core/include/replica_access_table.hpp
```diff
@@ -21,142 +21,149 @@ namespace irods::experimental
...
+    /// This function should only be called on startup of the server.
+    ///
+    /// \param[in] _shm_name The name of the shared memory to create.
+    /// \param[in] _shm_size The size of the shared memory to allocate in bytes.
+    ///
+    /// \since 4.2.9
+    auto rat_init(const std::string_view _shm_name = "irods_replica_access_table",
+                  std::size_t _shm_size = 1'000'000) -> void;
+
+    /// Cleans up any memory previously allocated by calling init().
```
korydraughn, line 74: init() -> rat_init()

server/core/src/rodsAgent.cpp
```diff
@@ -33,18 +25,49 @@
...
+        }
+
+        try {
+            rodsLog(LOG_DEBUG, "Client relVersion (raw): %s", _comm.cliVersion.relVersion);
+            std::istringstream in{_comm.cliVersion.relVersion};
+
+            irods::version version;
+            in.ignore(4) >> version.major;
+            in.ignore(1) >> version.minor;
+            in.ignore(1) >> version.patch;
```
korydraughn, line 60: consider what happens if the version string is corrupt.

korydraughn:
TODO: Make the delay server capture the server version so that it can determine how the packing instructions should be handled. This is likely not necessary because the delay server uses the binary protocol.

lib/core/src/client_properties.cpp
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    auto version::operator!=(const version& _rhs) const noexcept -> bool
+    {
+        return !(*this == _rhs);
+    }
+
+    auto version::operator>(const version& _rhs) const noexcept -> bool
+    {
+        const auto cmp = [](std::uint8_t _lhs, std::uint8_t _rhs, auto _func) noexcept -> bool
```
korydraughn, line 22: replace std::uint8_t with std::uint16_t

lib/core/src/client_properties.cpp
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    auto version::operator>=(const version& _rhs) const noexcept -> bool
+    {
+        return *this == _rhs || *this > _rhs;
+    }
+
+    auto version::operator<(const version& _rhs) const noexcept -> bool
+    {
+        const auto cmp = [](std::uint8_t _lhs, std::uint8_t _rhs, auto _func) noexcept -> bool
```
korydraughn, line 51: replace std::uint8_t with std::uint16_t

server/core/src/server_utilities.cpp
```diff
@@ -39,5 +51,60 @@ namespace irods
...
+
+        return std::chrono::seconds{3600};
+    }
+
+    auto parse_and_store_hosts_configuration_file_as_json() noexcept -> void
+    {
+        try {
+            std::string path;
+
+            if (const auto error = get_full_path_for_config_file("hosts_config.json", path); !error.ok()) {
```
korydraughn, line 94: replace "hosts_config.json" with macro (e.g. HOST_CONFIG_FILE)?

lib/core/src/procApiRequest.cpp
```diff
@@ -321,20 +298,26 @@ readAndProcApiReply( rcComm_t *conn, int apiInx, void **outStruct,
...
-        status = unpackStruct( errorBBuf->buf, ( void ** )( static_cast< void * >( &conn->rError ) ),
-                               "RError_PI", RodsPackTable, conn->irodsProt );
+        status = unpack_struct(errorBBuf->buf,
+                               (void**) (static_cast<void*>(&conn->rError)),
+                               "RError_PI", RodsPackTable, conn->irodsProt,
+                               conn->svrVersion->relVersion);
         if ( status < 0 ) {
             rodsLogError( LOG_ERROR, status,
-                          "readAndProcApiReply:unpackStruct error. status = %d",
+                          "readAndProcApiReply:unpack_sStruct error. status = %d",
```
korydraughn, line 320: fix this

lib/core/src/procApiRequest.cpp
```diff
@@ -363,27 +346,26 @@ procApiReply( rcComm_t *conn, int apiInx, void **outStruct,
...
-            status = unpackStruct( outStructBBuf->buf, ( void ** ) outStruct,
-                                   ( char* )RcApiTable[apiInx]->outPackInstruct, RodsPackTable,
-                                   conn->irodsProt );
+            status = unpack_struct(outStructBBuf->buf, (void**) outStruct,
+                                   (char*) RcApiTable[apiInx]->outPackInstruct, RodsPackTable,
+                                   conn->irodsProt, conn->svrVersion->relVersion);
             if ( status < 0 ) {
                 rodsLogError( LOG_ERROR, status,
-                              "readAndProcApiReply:unpackStruct error. status = %d",
+                              "readAndProcApiReply:unpack_sStruct error. status = %d",
```
korydraughn, line 361: fix these

korydraughn:
All tests passed CI for this PR: https://github.com/korydraughn/irods/commit/e232462fc36c093509c1fc6160456880a9000d00

lib/core/src/packStruct.cpp
```diff
@@ -1,494 +1,708 @@
...
+
+    auto to_version(const std::string& _peer_version) -> std::optional<irods::version>
+    {
+        if (!_peer_version.empty()) {
+            // Expects _peer_version to be something like "rods4.3.0".
+            static const std::regex pattern{R"_(\s*rods(\d+)\.(\d+)\.(\d+)\s*)_"};
+
+            if (std::smatch m; std::regex_match(_peer_version, m, pattern)) {
+                // The first match is the whole string. If there is a match, then we
+                // expect a match size of four.
```
alanking, line 240: Because 3 digits (major, minor, patch) plus first match?

lib/core/src/rcMisc.cpp
```diff
@@ -4729,3 +4757,43 @@ myRead( int sock, void *buf, int len,
...
+        case hostname_resolution_scheme::match_longest:
+            std::string longest;
+
+            for (auto&& address : iter->at("addresses")) {
+                if (auto tmp = address.at("address").get<std::string>(); tmp.size() > longest.size()) {
+                    longest = std::move(tmp);
+                }
+            }
+
+            return longest;
```
alanking, line 4796: Should we handle the default case? I realize it's an enum on which we are switching here, just makes me itch.

Also, my C brain did a double-take when I saw the declaration without curly braces :)

server/core/include/server_utilities.hpp
```diff
@@ -35,6 +36,30 @@ namespace irods
...
+    auto get_dns_cache_eviction_age() noexcept -> std::chrono::seconds;
+
+    /// Returns the hostname cache eviction age from server_config.json.
+    ///
+    /// If the configuration file cannot be read or parsed, returns a default of 30 seconds.
+    ///
+    /// \return std::chrono::seconds
+    ///
+    /// \since 4.2.9
+    auto get_hostname_cache_eviction_age() noexcept -> std::chrono::seconds;
```
alanking, line 56: Is there a reason `get_dns_cache_eviction_age` and `get_hostname_cache_eviction_age` are not part of their respective header files?

unit_tests/src/test_misc.cpp
```diff
@@ -0,0 +1,21 @@
...
+    REQUIRE(v4290 >= v4290);
+
+    constexpr irods::version v4300{4, 3};
+
+    REQUIRE(v4290 != v4300);
+    REQUIRE(v4290 <  v4300);
+    REQUIRE(v4290 <= v4300);
+    REQUIRE(v4300 >  v4290);
+    REQUIRE(v4300 >= v4290);
+}
```
alanking, line 20: Maybe this could be `test_version.cpp`? `test_misc` is unclear to me (is it intended for `rcMisc.h`?)

alanking (commented):
Amazing work as always. Most comments are minor/clarification

lib/core/include/packStruct.h
```diff
@@ -108,190 +99,39 @@ typedef struct PackedOutput {
...
-int
-unpackStruct( const void *inPackStr, void **outStruct, const char *packInstName,
-              const packInstruct_t *myPackTable, irodsProt_t irodsProt );
-int
-parsePackInstruct( const char *packInstruct, packItem_t &packItemHead );
-int
-copyStrFromPiBuf( const char *&inBuf, char *outBuf, int dependentFlag );
-int
-packTypeLookup( const char *typeName );
+__attribute__((deprecated("Use pack_struct")))
```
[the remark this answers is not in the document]
korydraughn: I believe it is because of differences between C and C++. I think using the nice form is a C++ specific thing and can't appear in extern "C" blocks.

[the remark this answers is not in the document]
alanking: Oh, I didn't even see that. Thanks!

lib/core/include/version.hpp
```diff
@@ -0,0 +1,90 @@
...
+    /// A simple struct capable of representing a version number.
+    ///
+    /// \since 4.2.9
+    struct version
+    {
+        // clang-format off
+        std::uint16_t major    = 0;
+        std::uint16_t minor    = 0;
+        std::uint16_t patch    = 0;
+        std::uint16_t revision = 0;
```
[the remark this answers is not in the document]
korydraughn: I always used the trailing underscore to signal to the reader that the member variable is private to the class/struct. They don't have a trailing underscore because these variables are part of the public interface. Willing to discuss more.

lib/core/src/packStruct.cpp
korydraughn: yes, but with 3 numbers.

lib/core/include/dns_cache.hpp
```diff
@@ -0,0 +1,97 @@
...
+    /// \since 4.2.9
+    auto dnsc_init(const std::string_view _shm_name = "irods_dns_cache",
+                   std::size_t _shm_size = 5'000'000) -> void;
+
+    /// Cleans up any resources created via dnsc_init().
+    ///
+    /// This function must be called from the same process that called dnsc_init().
+    ///
+    /// \since 4.2.9
+    auto dnsc_deinit() noexcept -> void;
```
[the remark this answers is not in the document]
korydraughn: consider using true namespaces instead of C style prefixes.

lib/core/src/packStruct.cpp
korydraughn: use a better comment

lib/core/src/rcMisc.cpp
korydraughn: add a comment about the `default` keyword not being included.

unit_tests/src/test_misc.cpp
korydraughn: rename to test_version.cpp

lib/core/src/rcMisc.cpp
korydraughn: On second thought, handling the default case is the better option. If someone casts a random integer to a `hostname_resolution_scheme`, who knows what would happen. At least, the default case would handle that safely.

lib/core/include/dns_cache.hpp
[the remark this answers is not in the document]
korydraughn: I'm leaning towards the following for namespaces:
- dns_cache::
- hostname_cache::
- replica_access_table::
- replica_access::

Thoughts?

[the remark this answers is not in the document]
korydraughn: In particular, `replica_access::` vs `replica_access_table::`.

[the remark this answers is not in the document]
korydraughn: Going with replica_access_table::.

lib/core/src/packStruct.cpp
korydraughn: is the new comment clearer?

trel: it is

korydraughn:
Do we need server config options for controlling the much shared memory needs to be allocated for the DNS cache and Hostname cache?

Right now, the DNS cache is initialized to 5000000 bytes. The Hostname cache is initialized with 1000000 bytes.

trel:
if it's easy enough to add, yes, that sounds good.

'shared_memory_size_in_bytes' for each?

korydraughn:
Will do.

korydraughn:
added #

PR #5407, opened 2021-02-12
[#5317,#5250] cmake: Introduce UseLibCXX module (master)

SwooshyCueb:
Took a break from working on 4973 to work on a couple of #5250-related PRs.

This PR introduces a `UseLibCXX` CMake module that handles invoking the `FindLibCXX` module and setting up the CMake project for libc++ (toolchain variables, directory properites, `std::allocator<void>` workaround, `RPATH` stuff, etc).
Additionally, with this PR, we are now setting `CMAKE_CXX_STANDARD` once (before `project` is called), rather than individually setting the `CXX_STANDARD` property from `IRODS_CXX_STANDARD` for each target.

Some additional relevant touch-ups are included, such as replacing `LINK_LANGUAGE` generator expressions with `COMPILE_LANGUAGE` where appropriate.

Touches #5317 (makes it easier to fix for other projects)
In service of #5250

I don't expect this to change functionality in any way, but I'm pushing it through jenkins just in case.


SwooshyCueb:
Pushed through jenkins. One test failure, but it's unrelated.

trel:
high quality minus.

please # when ready.

SwooshyCueb:
\#'d

PR #5433, opened 2021-02-23
DNS caching and Parallel Transfer + hosts_config.json (master)

korydraughn:
- [x] Make sure all functions in rcMisc.h are still there
- [x] Passed CI


korydraughn:
added #

Issue #5437, opened 2021-02-24
Convenience functions for API plugins should not trigger policy if invoked on the server-side


korydraughn:
- [x] master
- [x] 4-2-stable
---
## Bug Report
Invoking functions like `rs_replica_close()` on the server-side will trigger policy because it invokes the plugin via `server_api_call`. The server-side convenience functions need to invoke the plugins directly.

The following plugins need to be adjusted:
- rs_atomic_apply_acl_operations
- rs_atomic_apply_metadata_operations
- rs_data_object_finalize
- rs_data_object_modify_info
- rs_get_file_descriptor_info
- rs_replica_close
- rs_replica_open
- rs_touch

PR #5440, opened 2021-02-25
[#5437] Server-side API plugin convenience functions do not trigger policy (4-2-stable)

korydraughn:


korydraughn:
passed CI.

added #

PR #5449, opened 2021-03-03
delete s3 shmem files on startup/shutdown

JustinKyleJames:
Clean up shared memory on startup and shutdown of server.


trel:
looks good now - please # it. 

the only thing remaining is to make sure the shmem files being created match this regex.

JustinKyleJames:
> looks good now - please # it.
> 
> the only thing remaining is to make sure the shmem files being created match this regex.

Yes, I was doing that in parallel.

PR #5450, opened 2021-03-03
delete s3 shmem files on startup/shutdown

JustinKyleJames:
Clean up shared memory on startup and shutdown of server.

### DIFF
--- a/scripts/irods/controller.py
+++ b/scripts/irods/controller.py
@@ -72,6 +72,8 @@ class IrodsController(object):
         if self.get_binary_to_pids_dict():
             raise IrodsError('iRODS already running')
 
+        delete_s3_shmem()
+
         self.config.clear_cache()
         if not os.path.exists(self.config.server_executable):
             raise IrodsError(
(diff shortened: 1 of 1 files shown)


scripts/irods/controller.py
```diff
@@ -298,3 +304,12 @@ def delete_cache_files_by_name(*filepaths):
...
         except (IOError, OSError):
             l.warning(lib.indent('Error deleting cache file: %s'), path)
+
+def delete_s3_shmem():
+    # delete s3 shared memory if any exist 
+    s3_plugin_shmem = glob.glob(os.path.join(
+        paths.root_directory(),
+        'dev',
+        'shm',
+        '*s3-shm*'))
```
trel, line 314: we should consider namespacing these shmem files a bit more...

`*irods_s3-shm*` ?

JustinKyleJames: Good idea.  I will do that.

JustinKyleJames: Done

trel: You used `*irods-s3-shm*` rather than an underscore between `irods` and `s3`, like the others we use for the re_cache.

An underscore will keep them sorted/consistent.

JustinKyleJames:
Pound added.

PR #5453, opened 2021-03-03
msiRenameLocalZoneCollection and msiRenameCollection (4-2-stable)

korydraughn:


scripts/irods/test/test_iadmin.py
```diff
@@ -1629,6 +1629,77 @@ def check_and_remove_rebalance_visibility_metadata(attempts):
         with session.make_session_for_existing_admin() as admin_session:
             admin_session.assert_icommand(['ils', '-l', '-r', self.test_dir], 'STDOUT_SINGLELINE', self.test_dir)
 
+class Test_Iadmin_modzone(session.make_sessions_mixin([('otherrods', 'rods')], [('alice', 'apass')]), unittest.TestCase):
```
korydraughn, line 1632: These tests are actually testing `msiRenameLocalZoneCollection` and should probably be moved to test_all_rules.py.

doxygen/microservices.cpp
```diff
@@ -103,6 +103,7 @@
   - #msiDeleteCollByAdmin  - Deletes a collection by administrator
   - #msiRenameLocalZone  - Renames the local zone by updating various tables
```
trel, line 104: does this need to be updated/removed?

korydraughn: See `acRenameLocalZone` in the core.re.template.

Administrators will need to manually update their core.re files for this to work.

trel: Okay - but new installs (where nearly all zone renaming happens...) would have the new core.re...

Do we need to update the implementation in the default_cpp rule engine too?

korydraughn: I believe that change is in the PR

trel: "I see" said the blind man.

trel:
Convinced - please # when ready.

korydraughn:
added #

PR #5466, opened 2021-03-04
[5431] add JSON external for msi_get_file_descriptor_info (4-2-stable)

trel:


trel:
note to self - this PR is serving as a live link in a GitHub support request about how GitHub Actions sometimes is unable to reach https://unstable.irods.org

Can re-run and merge when that support request moves forward.

trel:
we're seeing fewer failures now, so will re-run.

oh... can't re-run... maybe there's a time window for that... interesting.

merging.

Issue #5419, opened 2021-02-18
Uncaught exception when trying to read from a nonexistent data object.


ilarik:
- [x] master
- [x] 4-2-stable
----------

## Bug Report

### iRODS Version, OS and Version

iRODS 4-2-stable (as of commit ID `40ad875c0f09f8e719326e3ad17f418c0545fc2a`) and CentOS Linux release 7.8.2003 (Core).

### What did you try to do?

If one (mistakenly) tries to stream from a data object, which doesn't exist, with `istream`, the server logs uncaught exceptions. The command line tool itself fails gracefully outputting an error message and returns a non-zero status.

### Expected behavior

The iRODS server to catch the exception and handle the error message in a more sysadmin friendly way.

### Observed behavior (including steps to reproduce, if applicable)

When running `istream read VERSION.json1` (the data object does not exist):


```
-bash-4.2$ istream read VERSION.json1
Error: Cannot open data object.
-bash-4.2$ echo $?
1
-bash-4.2$ 
```
the server log `rodsLog` is more verbose:
```
Feb 18 11:32:16 pid:22496 remote addresses: 127.0.0.1, ::1 ERROR: resolve_resource_hierarchy :: failed in file_object_factory                                                                                                                                                 
Feb 18 11:32:16 pid:22496 remote addresses: 127.0.0.1, ::1 ERROR: iRODS Exception:                                                                                                                                                                                            
    file: /irods_source/server/core/src/irods_resource_redirect.cpp                                                                                                                                                                                                           
    function: irods::resolve_hierarchy_result_type irods::resolve_resource_hierarchy(rsComm_t *, const std::string &, dataObjInp_t &, irods::file_object_factory_result &)                                                                                                    
    line: 366                                                                                                                                                                                                                                                                 
    code: -808000 (CAT_NO_ROWS_FOUND)                                                                                                                                                                                                                                         
    message:                                                                                                                                                                                                                                                                  
        resolve_resource_hierarchy :: failed in file_object_factory                                                                                                                                                                                                           
stack trace:                                                                                                                                                                                                                                                                  
--------------                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                              
Dumping stack trace                                                                                                                                                                                                                                                           
[...]

PR #5468, opened 2021-03-05
[#5419] Opening non-existent data object for read does not produce a stacktrace (master)

korydraughn:


korydraughn:
passed CI.

added #

Issue #5401, opened 2021-02-09
ichksum calculating against full file on disk, rather than only 'size' bytes stored in catalog


trel:
- [x] master
- [x] 4-2-stable
----------

### iRODS Version, OS and Version

4-2-stable

### What did you try to do?

interrogate a stale replica to determine whether it is still 'okay'.

### Expected behavior

From https://github.com/irods/irods/issues/5400#issuecomment-776132503...

The size on disk is not actually zero bytes.

But the correct checksum of the full file on disk is correct in the catalog.

```
$ ils -L
/testZone/home/irods:
  irods             0 demoResc            0 2021-02-09.18:04 X eels.txt
    6a57b28b63eede75341f2e06b428cf7a    generic    /var/lib/irods/Vault/home/irods/eels.txt
```

Running `ichksum -f` or `ichksum -K` should calculate and act based on zero bytes of data.

This would be consistent with how `iget` only retrieves the size of bytes in the catalog (which is behaving as expected):

```
$ iget -K eels.txt
remote addresses: 127.0.0.1 ERROR: rcDataObjGet: checksum mismatch error for foo, status = -314000 status = -314000 USER_CHKSUM_MISMATCH
remote addresses: 127.0.0.1 ERROR: getUtil: get error for foo status = -314000 USER_CHKSUM_MISMATCH
```

### Observed behavior (including steps to reproduce, if applicable)

`ichksum -f` is calculating on the full contents of the file on disk and saving the 'incorrect' value into the database.

Issue #3603, opened 2017-07-17
remove iphybun


trel:
marked deprecated in 4.2.0 (https://github.com/irods/irods/issues/3012)

PaulVanSchayck:
Hi Terrell, is the deprecation of iphybun only planned for the iCommand or also planned for the `msiPhyBundleColl`. Or is even the whole bundleResource feature being removed?

trel:
We haven't decided specifically - but input is welcome.

Are you depending on the `phyBundleColl` feature-set?

Candidates include:
- `rsPhyBundleColl`
- `rcPhyBundleColl`
- `remotePhyBundleColl`
- `msiPhyBundleColl`
- `iphybun`

PaulVanSchayck:
Hi Terrell. Thanks for your answer. We're actually considering using the bundle collections for a project that involves moving data to tape, while preserving objects and metadata. 

This is a project together with Matthew Saum (@Msaum87) from SURFsara. He is drafting a reply for this on irods-chat.

PaulVanSchayck:
Matthew started [this conversation](https://groups.google.com/forum/#!topic/irod-chat/CjDTUQ-BSFM) on irods-chat.

trel:
@msaum87 @PaulVanSchayck what is the status on your efforts with this?   Are you using this interface in a way that is helpful/successful?

We're leaning into the release of 4.3.0 and don't want to remove something that is serving a use case for you (or anyone else) without providing good justification or an alternate codepath/functionality.

PaulVanSchayck:
Nope.. we're not using it. We considered, but found another (better) solution not involving tar balls. Using tarballs made us very unhappy.

trel:
Ah, very good.   Excellent news :)

Can you share what you ended up doing?  To help anyone else who finds this discussion later and needs to solve a similar challenge.

---

We're interested in removing this functionality because it's a fixed, inflexible implementation and doesn't preserve the metadata or keep objects and subcollections searchable.

proposal:
 - remove these functions/iCommands
 - move to a more abstract model.... based on annotation and policy invocation
 - this would separate any operations (the what) from their invocation (the when)

new universe:
 - this work would be related to / part of the work on refactoring special collections (using libarchive methods in-memory, rather than with files in a (fake) resource, aka `bundleResc`)
 - new policy for most common use cases could be shipped with the iRODS Server

PaulVanSchayck:
[...]

PR #5475, opened 2021-03-08
[#5401] Checksum API calculates hash using size in catalog (4-2-stable)

korydraughn:


korydraughn:
requires https://github.com/irods/irods_client_icommands/pull/195

trel:
seems good when you're ready to #

server/api/src/rsFileChksum.cpp
```diff
@@ -185,7 +188,37 @@ int fileChksum(rsComm_t* rsComm,
         return error.code();
     }
 
-    auto bytes_remaining = irods::experimental::replica::replica_size(*rsComm, objPath, leaf_resc);
+    std::size_t bytes_remaining;
+
+    namespace ir  = irods::experimental::replica;
+
+    // If we're dealing with an intermediate replica, the data size may not be in the catalog just yet.
```
trel, line 195: we shouldn't be able to calculate/read an intermediate replica at all?

korydraughn: I'll have to check if that is enforced.

This block handles the case around `iput -K`. The checksum api is called before the data size is recorded in the catalog. To handle that, we record the size in the replica state table and the checksum api retrieves the size from the table if the replica is in an intermediate state.

trel: right - okay - if it's **our** intermediate replica, then that seems okay.

korydraughn: Due to changes in the implementation, this conversation no longer applies.

scripts/irods/test/test_resource_types.py
```diff
@@ -77,17 +77,21 @@ def tearDown(self):
 
     @unittest.skipIf(test.settings.RUN_IN_TOPOLOGY, "local filesystem check")
     def test_ichksum_no_file_modified_under_replication__4099(self):
+        #
+        # TODO IS THIS TEST NEEDED ANYMORE? WHAT DOES THIS TEST MEAN?
+        #
+
```
korydraughn, line 83: Do we need this test?

This test does not appear to be useful because the checksum API detects the situation and returns an error. Notice the changes to the test. The reason why this test worked before was because the checksum API read the entire file on disk rather than the data size in the catalog.

scripts/irods/test/test_resource_types.py
```diff
@@ -3678,17 +3688,21 @@ def tearDown(self):
...
```
korydraughn, line 3694: Do we need this test?

This test does not appear to be useful because the checksum API detects the situation and returns an error. Notice the changes to the test. The reason why this test worked before was because the checksum API read the entire file on disk rather than the data size in the catalog.

server/api/include/rsFileChksum.hpp
```diff
@@ -1,13 +1,37 @@
...
+               char* orig_chksum,
+               char *chksumStr);
+
+int file_checksum(RsComm* _comm,
+                  const char* _logical_path,
+                  const char* _filename,
+                  const char* _resource_hierarchy,
+                  const char* _original_checksum,
+                  rodsLong_t _data_size,
+                  char* _calculated_checksum);
```
korydraughn, line 34: Thoughts on the deprecation, name, etc.?

server/core/include/physPath.hpp
```diff
@@ -1,109 +1,117 @@
...
+struct DataObjInfo;
+struct DataObjInp;
+struct FileRenameInp;
+struct KeyValPair;
+struct RsComm;
+struct StructFileExtAndRegInp;
+struct StructFileOprInp;
+struct VaultPathPolicy;
+
+extern "C"
```
korydraughn, line 22: Does this need to be marked as `extern "C"`? It has a C++ header extension.

lib/api/include/fileChksum.h
```diff
@@ -1,25 +1,51 @@
...
     char fileName[MAX_NAME_LEN];
     char rescHier[MAX_NAME_LEN];
     char objPath[MAX_NAME_LEN];
-    int flag;   // not used for now
-    char in_pdmo[MAX_NAME_LEN]; // Flag indicating if we are being executed from a pdmo
-    char orig_chksum[NAME_LEN]; // original incoming checksum
+    int flag;                   // Not used for now.
+    char in_pdmo[MAX_NAME_LEN]; // Flag indicating if we are being executed from a pdmo.
+    char orig_chksum[NAME_LEN]; // Original incoming checksum.
+    rodsLong_t dataSize;
```
trel, line 20: dataSize is added here, but not in the #define a few lines below

scripts/irods/test/test_ichksum.py
```diff
@@ -68,11 +72,18 @@ def test_ichksum_force(self):
...
         lib.cat(full_path, 'XXXXX')
         new_chksum = lib.file_digest(full_path, 'sha256', encoding='base64')
-        self.admin.assert_icommand(['ichksum', '-f', filename], 'STDOUT_SINGLELINE', new_chksum)
+
+        # Show that the new checksum does not match the original.
+        self.assertNotEqual(file_chksum, new_chksum)
+
+        # The checksum will remain the same because ichksum reads, at most, the number of bytes
+        # recorded in the catalog.
+        self.admin.assert_icommand(['ichksum', '-f', filename], 'STDOUT_SINGLELINE', file_chksum)
```
korydraughn, line 86: Does this test still make sense? As in does the behavior match the test's name?

scripts/irods/test/test_resource_types.py
trel: Makes sense to me.

@alanking wrote those a couple years ago - thoughts?   Anything we're missing?

scripts/irods/test/test_resource_types.py
trel: same as above - this case should no longer need a test.

server/api/src/rsDataObjClose.cpp
```diff
@@ -771,6 +769,8 @@ int rsDataObjClose(rsComm_t* rsComm, openedDataObjInp_t* dataObjCloseInp)
                                                 l1desc.dataObjInfo->objPath, l1desc.dataObjInfo->rescHier));
         }
     }
+
+    // TODO Is this comment needed?
```
trel, line 773: clean up on aisle 774

server/core/src/physPath.cpp
```diff
@@ -428,63 +422,38 @@ getSizeInVault( rsComm_t *rsComm, dataObjInfo_t *dataObjInfo ) {
...
 
-        }
-        break;
-    default:
-        rodsLog( LOG_NOTICE,
-                 "_dataObjChksum: rescCat type %d is not recognized", category );
-        status = SYS_INVALID_RESC_TYPE;
-        break;
+    if (ec == DIRECT_ARCHIVE_ACCESS) {
+        const auto msg = fmt::format(R"_(Data object: "{}" is located an an archive resource. Ignoring its checksum.)_",
```
trel, line 451: "an an archive" -> "in an archive"

server/core/src/physPath.cpp
```diff
@@ -428,63 +422,38 @@ getSizeInVault( rsComm_t *rsComm, dataObjInfo_t *dataObjInfo ) {
...
+                                     dataObjInfo->filePath); // TODO Should this be the logical path instead?
```
trel, line 452: let's consider showing all three pieces of information...
 - resource name (perhaps full hierarchy, at this point it would be helpful to the admin)
 - logical path
 - physical path / OID

server/api/src/rsPhyBundleColl.cpp
```diff
@@ -493,8 +486,13 @@ bundleAndRegSubFiles( rsComm_t *rsComm, int l1descInx, char *phyBunDir,
...
         // JMC - backport 4528
         if ( chksumFlag != 0 ) {
-            status = fileChksum( rsComm, regReplicaInp.destDataObjInfo->filePath,
-                                 subPhyPath, regReplicaInp.destDataObjInfo->rescHier, 0, tmpBunReplCache->chksumStr );
+            status = file_checksum(rsComm,
+                                   regReplicaInp.destDataObjInfo->filePath,
+                                   subPhyPath,
+                                   regReplicaInp.destDataObjInfo->rescHier,
+                                   nullptr,
+                                   regReplicaInp.destDataObjInfo->dataSize, // TODO Should this come from the source data object info?
```
trel, line 494: where is it coming from now?   isn't `destDataObjInfo->dataSize` populated with the size from the source already?   or is it coming from a physical stat?  seems too early for that...

server/api/include/rsFileChksum.hpp
trel: looks good to me - was going to say make sure you get all of the server calls... but there were only two and they're handled in this PR.

scripts/irods/test/test_ichksum.py
trel: right, we now trust the catalog, so this is a moot test.

it would behave the same, with and without the `-f` in this case, correct?

lib/api/include/fileChksum.h
korydraughn: updated packing instruction

scripts/irods/test/test_resource_types.py
alanking: The test has to do with triggering `fileModified` after performing a checksum which was not valid at the time of writing the test (see #4099).

The test does the following:
1. Start with 2 identical replicas under a replication resource.
2. Corrupt the data of one of the replicas.
3. Perform a checksum update on the corrupted replica. In the past, this would forcibly update the checksum in the catalog. It seems that this is not what would happen anymore due to file size mismatch (more on this below).
4. Ensure that the other replica was not overwritten by replication triggered by `fileModified`. In other words, make sure a replication did not occur due to the checksum calculation as it is not a valid trigger for `fileModified`.

If the checksum isn't being updated due to file size on disk not matching what is recorded in the catalog, then perhaps the test should corrupt the file in such a way as to match the file size on disk with what is recorded in the catalog so that the checksum calculation goes through.

Would that restore the validity of the test? Happy to discuss whether this is a valid test, but that's the idea behind it.

korydraughn: I'll adjust how much data is written on the overwrite. That should restore its validity.

server/api/src/rsPhyBundleColl.cpp
korydraughn: I'm not sure where this comes from. The CI tests that hit this code appear to behave so this may be fine. I'll spend a few minutes looking over this.

scripts/irods/test/test_ichksum.py
korydraughn: Correct. This PR adds two tests that demonstrate this behavior.

Will remove this test.

scripts/irods/test/test_resource_types.py
korydraughn: Overwriting the file with enough data allows the test to pass. I think it is good now.

scripts/irods/test/test_ichksum.py
korydraughn: Test removed.

server/api/src/rsDataObjClose.cpp
korydraughn: removed

server/core/src/physPath.cpp
korydraughn: done.

server/core/src/physPath.cpp
korydraughn: done.

scripts/irods/test/test_resource_types.py
korydraughn: See comments above.

server/api/src/rsPhyBundleColl.cpp
korydraughn: Reading the code, the data size member of the src/dst dataObjInfo appears to be zero at the point of the `file_checksum` invocation. I don't see the data size member being updated anywhere. The `regReplicaInp` is constructed and initialized within this function.

Not only that, but the phybun APIs and commands appear to be marked as deprecated (see https://github.com/irods/irods/issues/3603).

The tests that invoke `iphybun` are skipped or commented out. [This test](https://github.com/irods/irods/blob/be48c4ab2880b1188ddcf12ec2f44f8d178b434d/scripts/irods/test/test_all_rules.py#L456) is the only one that invokes a phybun operation. However, no checksum flags are passed during this test. The rule file can be found [here](https://github.com/irods/irods_client_icommands/blob/4-2-stable/test/rules/rulemsiPhyBundleColl.r).

Q. Is this API still supported?
Q. What do we want to do about this API?

korydraughn: After discussing this topic with the team, it has been decided that the behavior of the phybun APIs is not something that this PR should be concerned about.

server/core/include/physPath.hpp
korydraughn: removed.

korydraughn:
passed CI.

added #

Issue #4331, opened 2019-04-05
new api plugin - finalize data transfer


trel:
- [x] master
- [x] 4-2-stable
---------

## Feature

A new server api plugin to check integrity of completed transfer, possibly checksum, update catalog, report success.

takes json manifest of
 - data object name
 - data transfer map of parts on disk
 - flags (checksum, force, etc.)
 - atomic metadata
 - atomic acls
 - storage resource
 - physical path

kript:
Naively, this looks like it would be helpful for #4314 and #4170?

trel:
As currently being designed, yes - a new multipart transfer mechanism (https://github.com/irods/irods/issues/4336) would fix a lot of the issues with parallel transfer.

michael-conway:
Is this intended to be a follow up call after oper complete on a parallel
transfer? Should it be baked in to client apis? As an  option? I was not
sure the context for this plugin.

trel:
This is a replacement, not a thing to call at the end of existing parallel transfer.

trel:
@alanking is this complete?   seems it should at least be pulled into the current milestone?

alanking:
I would say this is not quite complete, but it is definitely part of my current efforts. I'll pull it in.

alanking:
I think this will be complete after this upcoming write-lock implementation PR.

I do have some concerns around the API's usage from the client side and as it relates to the iRODS permissions model. These are probably pretty easily resolved, but let's only close it after it has been confirmed that these check out.

alanking:
#5594 (and imminent 4-2-stable equivalent) will close this issue. It addresses the permissions concern.

trel:
@alanking i think this is done?    anything else?

alanking:
Yes, the remaining concerns that I had for this have been addressed.

PR #5476, opened 2021-03-08
[#4331] Update data_object_finalize/RST interfaces (4-2-stable)

alanking:
The data_object_finalize and replica_state_table interfaces are very
closely related and so the JSON passed between the two has been modified
to bring a sense of consistency and clarity to their usage.

The RST interfaces underwent the following changes:
    - RST is now a std::map<std::uint64_t, nlohmann::json> instead of
      simply a JSON map for key-type flexibility
    - Entries are keyed by data_id instead of logical path
    - update now takes only a list of r_data_main column-string pairs
    - publish_to_catalog:
        - now takes a replica number and a JSON object which, if
          present, indicates that file_modified should be triggered
        - JSON object is added as a field to the replica indicated
          by the replica number input

The data_object_finalize API underwent the following changes:
    - data_id field was removed as it served no purpose - this should be
      retrieved from the replica information
    - fileModified is now triggered in finalize by the presence of
      trigger_file_modified boolean
    - The file_modified entry will remain on the affected replica as
      a JSON object

---
CI tests running.

### DIFF
--- a/lib/api/include/data_object_finalize.h
+++ b/lib/api/include/data_object_finalize.h
@@ -16,8 +16,6 @@ extern "C" {
 /// desired modifications is required in the form of a JSON structure. Each replica can also
 /// have "file_modified" key which holds a set of key-value pairs for the file_modified plugin operation.
 ///
-/// The data_id is used to identify the data object being finalized.
-///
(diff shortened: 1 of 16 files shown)


trel:
green - please to # it when happy.

alanking:
CI tests passed

Issue #5478, opened 2021-03-09
Put in replication node on 4-2-stable is single-threaded for the replication


JustinKyleJames:
- [x] master
- [x] 4-2-stable
------

A put in a replication node on 4-2-stable is single threaded.  The first replica is written with N threads but after that the read/writes are done in a single thread.

Added logging to show this.  The thread numbers are in ```[[ ]]```

Last few lines of the first PUT
-----

```
Mar  9 19:48:17 pid:2841 NOTICE: /github/irods/plugins/resources/unixfilesystem/libunixfilesystem.cpp:742 (unix_file_write) [[139941476546304]] write 4194304 bytes
Mar  9 19:48:17 pid:2841 NOTICE: /github/irods/plugins/resources/unixfilesystem/libunixfilesystem.cpp:742 (unix_file_write) [[139942193764096]] write 2649088 bytes
Mar  9 19:48:17 pid:2841 NOTICE: /github/irods/plugins/resources/unixfilesystem/libunixfilesystem.cpp:742 (unix_file_write) [[139941476546304]] write 4194304 bytes
Mar  9 19:48:17 pid:2841 NOTICE: /github/irods/plugins/resources/unixfilesystem/libunixfilesystem.cpp:742 (unix_file_write) [[139941476546304]] write 2649088 bytes
Mar  9 19:48:17 pid:2841 NOTICE: /github/irods/plugins/resources/unixfilesystem/libunixfilesystem.cpp:742 (unix_file_write) [[139941216503552]] write 2649088 bytes
Mar  9 19:48:17 pid:2841 NOTICE: /github/irods/plugins/resources/unixfilesystem/libunixfilesystem.cpp:700 (unix_file_read) [[139942534428608]] read 4194304 bytes
Mar  9 19:48:17 pid:2841 NOTICE: /github/irods/plugins/resources/unixfilesystem/libunixfilesystem.cpp:742 (unix_file_write) [[139942534428608]] write 4194304 bytes
```
First few lines of the READ/WRITE to the Replica
----

```
Mar  9 19:48:17 pid:2841 NOTICE: /github/irods/plugins/resources/unixfilesystem/libunixfilesystem.cpp:700 (unix_file_read) [[139942534428608]] read 4194304 bytes
Mar  9 19:48:17 pid:2841 NOTICE: /github/irods/plugins/resources/unixfilesystem/libunixfilesystem.cpp:742 (unix_file_write) [[139942534428608]] write 4194304 bytes
[...]

PR #5480, opened 2021-03-10
[#5478] Set numThreads in rsDataObjCopy/Repl/Phymv (master)

alanking:
CI tests passed

### DIFF
--- a/scripts/irods/test/test_icp.py
+++ b/scripts/irods/test/test_icp.py
@@ -21,7 +21,6 @@ class Test_Icp(session.make_sessions_mixin([('otherrods', 'rods')], [('alice', '
     class_name = 'Test_Icp'
     def setUp(self):
         super(Test_Icp, self).setUp()
-        self.admin = self.admin_sessions[0]
         self.alice = self.user_sessions[0]
 
     def tearDown(self):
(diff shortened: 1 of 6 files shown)


scripts/irods/test/test_icp.py
```diff
@@ -51,3 +50,26 @@ def test_icp_closes_file_descriptors__4862(self):
...
+
+                self.assertGreaterEqual(int(cp_thread_count), threads)
+
+            finally:
+                self.alice.assert_icommand(['irm', '-f', dest_path])
+                self.alice.assert_icommand(['irm', '-f', another_dest_path])
+
+        do_test(0, 0)
+        do_test(31457280, 1) # 30MiB
+        do_test(52428800, 1) # 50MiB
```
alanking, line 75: Any suggestions on what numbers to check for here? I remember that there is a place where we calculate this kind of thing but I do not now recall where that is. Also, 0 is a little special in that I might need a `self.assertEqual` rather than `self.assertGreaterEqual`, but if we can determine the exact number of threads expected in each case then I can always use `self.assertEqual`.

scripts/irods/test/test_irepl.py
```diff
@@ -13,13 +13,16 @@
 from .. import test
 from ..configuration import IrodsConfig
 
-class Test_IRepl(session.make_sessions_mixin([('otherrods', 'rods')], [('alice', 'apass')]), unittest.TestCase):
+class Test_IRepl_Special(session.make_sessions_mixin([('otherrods', 'rods')], [('alice', 'apass')]), unittest.TestCase):
```
trel, line 16: better/different name than 'special'?

scripts/irods/test/test_icp.py
trel, line 75: 33MiB is above the threshold, could save a tad amount of time

trel: and `msiSetNumThreads` has the real logic

scripts/irods/test/test_irepl.py
trel: me too - but not sure how much of that is Python convention vs unittest module requirements vs Zoey defaults persisting.

scripts/irods/test/test_icp.py
alanking: Yep, here's the magic formula...

https://github.com/irods/irods/blob/9eb6c23df45cdedff8ee9c3af71a65d304635037/server/re/src/reSysDataObjOpr.cpp#L828

alanking: Okay, we can just do 33MiB and 0 for now

scripts/irods/test/test_irepl.py
alanking: Changed the name

alanking:
I think this is good to go unless waiting on more eyeballs

trel:
looks good to me - please squash as desired.    we can # after one more round.

trel:
please revisit the commit message for the first commit.

it's now asserting exactly the correct number of threads, and only testing two sizes, not three.

alanking:
Good call.

trel:
green is good.   # em.

alanking:
#'d

Issue #5506, opened 2021-03-25
"ils" needs option that instructs it to not print the contents of a collection


korydraughn:
- [x] master
- [x] 4-2-stable
---
## Feature
`ils` always prints the contents of a collection regardless of the options. `ils` needs to provide an option that works just like the `-d` option of `ls`. This would avoid listing the contents of the collection. See example below.

### Example
```bash
$ ils
/tempZone/home/rods:
  C -/tempZone/home/rods/1000_files.d
  C -/tempZone/home/rods/10000_files.d
  C -/tempZone/home/rods/100000_files.d
$
$ # The "-d" option instructs ils to not print the contents of the collection.
$ ils -Ad 100000_files.d
/tempZone/home/rods/100000_files.d:
        ACL - rods#tempZone:own   alice#tempZone:read object
        Inheritance - Disabled
```

PR #5510, opened 2021-03-25
[#5506] Implemented new -d option of ils (4-2-stable)

korydraughn:


lib/core/src/lsUtil.cpp
```diff
@@ -407,6 +406,12 @@ lsCollUtil( rcComm_t *conn, rodsPath_t *srcPath, rodsEnv *myRodsEnv,
         printCollInheritance( conn, srcColl );
     }
 
+    // Return immediately if the user passed the "-d" option.
+    // This option allows "ils" to mirror the behavior of "ls -d <directory>".
+    if (True == rodsArgs->dataObjects) {
```
trel, line 411: dataObjects?    rather than 'directory listing' or something?

scripts/irods/test/test_ils.py
```diff
@@ -57,3 +57,20 @@ def test_genquery_fallback_for_collection_permissions__issue_4636(self):
                           " R_OBJT_ACCESS.access_type_id = R_TOKN_MAIN.token_id")
         self.admin.assert_icommand(['iadmin', 'asq', specific_query, 'ShowCollAcls'])
 
+    def test_ils_lists_collections_without_listing_their_contents__issue_5506(self):
```
trel, line 60: need a test to show that `-A` still works?

lib/core/src/lsUtil.cpp
korydraughn: That is the variable name that is assigned to the -d option. I can't change it. No one will see that though.

scripts/irods/test/test_ils.py
korydraughn: Should they be incompatible? Right now, they are ignored if the arg is a collection. Same for the -r option.

trel: i've come around....

for collection
 - `-d` and `-l` are fine - `-d` takes precendence  - ignore `-l`
 - `-d` and `-L` are fine - `-d` takes precendence  - ignore `-L`

for data object
 - `-r` is ignored?   matches behavior of `ls -R` (recursive)
 - `-d` should be... ignored as well... same as (`ls -d foo.txt`)

korydraughn: Great! I believe the code modifications support all of your thoughts right now. The -d option is only checked on the collection listing code path.

I'll add tests so that this behavior is captured.

korydraughn: done.

korydraughn:
passed CI

added #

PR #5511, opened 2021-03-29
[4055] Restore effect of time units s,m,h,d,y for delayExec PLUSET tag - 4-2-stable

d-w-moore:
Test passes at bench, now running in CI

### DIFF
--- a/lib/core/src/rcMisc.cpp
+++ b/lib/core/src/rcMisc.cpp
@@ -1566,17 +1566,44 @@ getLocalTimeStr( struct tm *mytm, char *timeStr ) {
 }
 
 
-/* Get the current time + offset , in the  form: 2006-10-25-10.52.43 */
-/*  offset is a string of the same form  */
-/*                                               0123456789012345678 */
-void
-getOffsetTimeStr( char *timeStr, const char *offSet ) {
-    time_t myTime;
 
-    myTime = time( NULL );
-    myTime += atoi( offSet );
-
-    snprintf( timeStr, TIME_LEN, "%d", ( uint ) myTime );
+/* Get the current time + offset , and store to timeStr (understood max length TIME_LEN) in epoch seconds.
+    OffSet is a string of the form nnnn<Unit> with Unit in {'','s','m','h','d','y'}.
+    An empty Unit field implies seconds. */
+int
+getOffsetTimeStr( char *timeStr, const char *offSet )
+{
+    long seconds_multiplier = 1L;
+    static const std::map<char,long> multiplier {
+        {'\0',1}, {'s',1}, {'m',60}, {'h',3600}, {'d',24*3600}, {'y',24*3600*365}
+    };
+    time_t myTime = time( NULL );
+    if (offSet == nullptr) { return SYS_INTERNAL_NULL_INPUT_ERR; }
+    const char *lTrimOffSet = offSet;
+    while(isspace(*lTrimOffSet)) {++lTrimOffSet;}        // left-trim any whitespace
+    int nonNumberLoc = strspn(lTrimOffSet,"0123456789.");
+    char u = lTrimOffSet[nonNumberLoc];                  // u is value of first nondigit character (at or before end '\0')
+    const char* error_type = (nonNumberLoc == 0 ? "Empty or invalid number field" : "");
+    if (0 == *error_type) {  // There's a time specification, so determine the time unit
+        try {
+            seconds_multiplier = multiplier.at(u);
+        }
+        catch (const std::out_of_range&) {
+            // An unmapped, non-whitespace character as the unit field is an error.
+            if (!(isspace(u))) {
+               error_type = "Bad units (need s, h, m, d, or y)";
+            }
+            // deliberate fall through with seconds_multiplier = 1
+        }
+    }
+    if (NULL != strchr(lTrimOffSet,'.') && seconds_multiplier > 1) { error_type = "Fractions disallowed unless units are seconds"; }
+    if (*error_type) {
+        rodsLog(LOG_ERROR, "Incorrect delay interval '%s'.  Reason:  %s.", lTrimOffSet, error_type);
+        return INPUT_ARG_NOT_WELL_FORMED_ERR;
+    }
+    myTime += atol( lTrimOffSet ) * std::max(1L,seconds_multiplier);     // max function prevents zeroing or negation
+    snprintf( timeStr, TIME_LEN, "%ld", ( long ) myTime );
+    return 0;
 }
 
 /* Update the input time string to be offset minutes ahead of the
--- a/scripts/irods/test/test_rule_engine_plugin_framework.py
+++ b/scripts/irods/test/test_rule_engine_plugin_framework.py
@@ -14,6 +14,7 @@ import time  # remove once file hash fix is commited #2279
 import subprocess
 import mmap
 import shutil
+import re
 
 from .. import lib
 from .. import paths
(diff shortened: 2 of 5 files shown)


trel:
i still see some `mytm->tm_year + 1900` and direct character mapping going on in
https://github.com/irods/irods/blob/9f9d29bca3e9fce9a7f403443abf6436483eb3ba/lib/core/src/rcMisc.cpp#L1592-L1640

let's make this is clear as possible for the next person

also, did this ever work?     last touched 10 years ago?
  https://github.com/irods/irods-legacy/blame/master/iRODS/lib/core/src/rcMisc.c#L1842

d-w-moore:
wrt the confusing/suspect/ancient code noted above:
- I believe `tm_year +1900` is a long standing idiom as POSIX has always stored the year as offset relative to year 1900
- Though inherently distrustful of timestamps that don't include a timezone, the time translations seem to be in working order
- Though the removal of ancient, timezone-agnostic, unix centric C-isms is a noble goal (std::chrono anyone?) I'd need considerably more study to feel good about such changes in this context. Maybe we study more broadly how timestamps are used in iRODS and file an issue relating to saner time keeping? : )

d-w-moore:
I'd totally be for overhauling this comment (among others) which, given the function it precedes and presumably describes, is inaccurate.
https://github.com/irods/irods/blob/9f9d29bca3e9fce9a7f403443abf6436483eb3ba/lib/core/src/rcMisc.cpp#L1570

d-w-moore:
I am pretty sure times are consistently calculated and stored in iRODS as time_t, which is a long integer.  The use of localtime is probably noncritical as a result, but admittedly archaic!

scripts/irods/test/test_rule_engine_plugin_framework.py
```diff
@@ -475,6 +476,38 @@ def setUp(self):
...
+                regex_time = '^\s*RULE_EXEC_TIME\s*=\s*([0-9]+)'
+                regex_id = '^\s*RULE_EXEC_ID\s*=\s*([0-9]+)'
+                dummy_rc, out, dummy_err = self.admin.assert_icommand(
+                    ['iquest', """SELECT RULE_EXEC_ID, RULE_EXEC_NAME, RULE_EXEC_TIME where RULE_EXEC_NAME like '%%4055_%s.%%'"""%units],
+                    'STDOUT_MULTILINE', regex_time, use_regex=True)
+                exec_tm =  int(re.search(regex_time,out,flags=re.MULTILINE).groups(1)[0])
+                rule_id =  re.search(regex_id,out,flags=re.MULTILINE).groups(1)[0]
+                expected_delay = 10 * multiplier
+                self.assertTrue((tm + 0.6*expected_delay) < exec_tm < (tm + 1.4*expected_delay)) # is exec time accurate?
+                self.admin.assert_icommand(['iqdel', rule_id])
```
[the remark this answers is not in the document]
d-w-moore: Very true that comments needed.
I am guessing that bad inputs (units not in [y,m,d,h,s]) produce something like `INPUT_ARG_NOT_WELL_FORMED_ERR`

d-w-moore:
Tests now failing due to the use of fields such as 0.1s being used in PLUSET ...
So, to allow people flexibility of belief and not break their code, I'm going to make a change allowing decimals in the number field prior to the units character.
But  probably, people should be made aware the fractional part is always lost via atoi ( )

lib/core/src/rcMisc.cpp
```diff
@@ -1566,17 +1566,37 @@ getLocalTimeStr( struct tm *mytm, char *timeStr ) {
...
+    int nonDigitLoc = strspn(lTrimOffSet,"0123456789");
+    char u = lTrimOffSet[nonDigitLoc];                          // u is value of first nondigit character
+                                                                // (Guaranteed not beyond the terminating null)
+    try {
+        seconds_multiplier = multiplier.at(u);
+    }
+    catch (std::out_of_range & e) {
+        // An unmapped, non-whitespace character as the unit field is an error.
+        if (!(isspace(u))) {
+            rodsLog(LOG_ERROR, "Incorrect delay interval <%s>; needed <nnnn[U]> with optional U in {s,m,h,d,y}", lTrimOffSet);
```
alanking, line 1592: The error message leads me to believe that I can only use a value up to and not exceeding 9,999, and cannot have whitespace between the value and the optional unit. Is that right?

Also, this function is an excellent candidate for unit tests, but I'll consider that a bonus ;D

Also, the caught exception can be const and actually doesn't need a name as it is not used in this block. Pretty minor.

lib/core/src/rcMisc.cpp
```diff
@@ -1566,17 +1566,37 @@ getLocalTimeStr( struct tm *mytm, char *timeStr ) {
...
+int
+getOffsetTimeStr( char *timeStr, const char *offSet )
+{
+    long seconds_multiplier = 1L;
+    static const std::map<char,long> multiplier {
+        {'\0',1}, {'s',1}, {'m',60}, {'h',3600}, {'d',24*3600}, {'y',24*3600*365}
+    };
+    time_t myTime = time( NULL );
+    const char *lTrimOffSet = offSet;
+    while(isspace(*lTrimOffSet)) {++lTrimOffSet;}               // left-trim any whitespace
```
alanking, line 1582: Would it be worth checking for `nullptr` on the two inputs, or is that overkill in this situation?

alanking (commented):
On the point about supporting/tolerating decimals, I am of the opinion that we should just enforce integers, but if you or others feel differently I guess that's fine. Just needs lots of documentation, as you say.

lib/core/src/rcMisc.cpp
d-w-moore: The documentation says 'nnnn' but the code actually allows more digits.  The only thing that I've done is allowed units to be used again, no new limitations, so for example 86400s = 1d and all that. I've tested beyond the year 2038 ("20y")and that works too.  I am pretty happy with it and it should pass the test suite i'm running now, and unless someone wants to educate me in the catch2 and unit tests department, probably will leave it as is...
I cannot enforce integers as @trel wanted because  test_delay_queue was too finicky; if you have any spare time you might look at  `test_delay_queue_with_long_job`. It was brittle for some reason, didn't like being an integer, so I had to leave the PLUSET field at 0.1s

lib/core/src/rcMisc.cpp
d-w-moore: i had nullptr before but the man page says use NULL.  I know they're the same, but for C style code I generally write in C idioms ; )

alanking: What I meant was that we dereference `lTrimOffset` (which is equivalent to `offSet` at first) without making sure it's a valid pointer and I didn't know if we wanted to check our inputs.

lib/core/src/rcMisc.cpp
alanking: Okay, just wanted to make sure.

As for `test_delay_queue`, what you said seems fine for now. I think the tests should probably be updated but is probably outside of the scope of this issue.

d-w-moore:
This seems ready for merge. If all agree, I'll squash and # here and on master

trel:
Yes, please mark the conversations above resolved if they are resolved.... and then squash down to the exact commits.

We'll # after one last pass...

scripts/irods/test/test_rule_engine_plugin_framework.py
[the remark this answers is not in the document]
d-w-moore: Made test for bad/empty inputs and comments where necessary.

lib/core/src/rcMisc.cpp
d-w-moore: @alanking Ok, let me know what kind of issue if any we should make for updating the funky fractions (`0.1s`) in ` test_delay_queue_with_long_job` and friends.  I read through and wasn't quite clear about the intent of the tests.

alanking: Filed #5526 for this. Thanks

lib/core/src/rcMisc.cpp
d-w-moore: ah good point

d-w-moore:
Squashed and ready for merge,  here and also on master.

PR #5512, opened 2021-03-29
[4055] Restore effect of time units s,m,h,d,y for delayExec PLUSET tag - master

d-w-moore:


trel:
seems like you're ready to squash to intended commits.   then we can # after more eyes.

d-w-moore:
Yes, now testing this on master .  Greenlight on  4055.42s where it greenlighted all tests that were also passing on bare 4-2-stable (yes, the outlier is  test_irodsctl ), so looks like we are almost there.

Issue #5494, opened 2021-03-17
Remove usage of PHYOPEN_BY_SIZE_KW in the server


alanking:
- [x] master
- [x] 4-2-stable
---
`PHYOPEN_BY_SIZE_KW` is used to signal to `rsDataObjOpen` that a physical open of a replica should only take place in the event that the size of the file exceeds the configured `maximum_size_for_single_buffer_in_megabytes`. This has historically only been used in `rsDataObjPut` and `rsDataObjGet` so that they can use tools like `l3FileGetSingleBuf` and `l3FilePutSingleBuf`. Both `rsDataObjPut` and `rsDataObjGet` could be implemented using open/write/close and open/read/close, respectively, and always physically open the file. This will also make the transition to using the new parallel transfer engine internally a little easier to make.

alanking:
For posterity, this change will have no effect on older servers talking to newer servers and vice versa (i.e. federation).

Here is the relevant code in `rsDataObjGet`:
https://github.com/irods/irods/blob/9eb6c23df45cdedff8ee9c3af71a65d304635037/server/api/src/rsDataObjGet.cpp#L148-L149

Here is the relevant code in `rsDataObjOpen`:
https://github.com/irods/irods/blob/9eb6c23df45cdedff8ee9c3af71a65d304635037/server/api/src/rsDataObjOpen.cpp#L368-L389

The `PHYOPEN_BY_SIZE_KW` is added to the `condInput` for the `dataObjInp` given to `rsDataObjOpen`. This is considered in `rsDataObjOpen` and the file is not physically opened if the size of the replica in question is below the configured `maximum_size_for_single_buffer_in_megabytes`. All of this logic is to be removed as a result of resolving this issue.

In `rsDataObjGet`, if the data object resides in a remote zone, a client call is made as shown here:
https://github.com/irods/irods/blob/9eb6c23df45cdedff8ee9c3af71a65d304635037/server/api/src/rsDataObjGet.cpp#L98-L128

The redirection occurs before the `PHYOPEN_BY_SIZE_KW` is considered and because `rsDataObjGet` is the one which adds the keyword, the server in the remote zone will use the keyword if it is a version of iRODS which cares about it (i.e. before the resolution of this issue). The open-by-size logic will be honored on the remote server in this case.

[...]

PR #5514, opened 2021-03-30
[#5494] Remove use of PHYOPEN_BY_SIZE_KW in server (master)

alanking:
Implements single buffer gets (that is, gets of replicas whose size is
under the configured maximum_size_for_single_buffer_in_megabytes value)
as an open/read/close rather than using l3GetDataSingleBuf and
determining whether to perform a physical file open via the
PHYOPEN_BY_SIZE_KW. The open is now performed as normal and the size
information included in the L1 descriptor table is used to determine if
the get will be single buffer or parallel.

---

CI tests passed. I have outlined potential impacts (or lack thereof) here: https://github.com/irods/irods/issues/5494#issuecomment-810417599


trel:
so close to being some minus :)      please to # when ready.

alanking:
#'d. The Codacy issues can be ignored as it does not understand `at_scope_exit`.

PR #5529, opened 2021-04-08
[#5494] Remove use of PHYOPEN_BY_SIZE_KW in server (4-2-stable)

alanking:
Implements single buffer gets (that is, gets of replicas whose size is
under the configured maximum_size_for_single_buffer_in_megabytes value)
as an open/read/close rather than using l3GetDataSingleBuf and
determining whether to perform a physical file open via the
PHYOPEN_BY_SIZE_KW. The open is now performed as normal and the size
information included in the L1 descriptor table is used to determine if
the get will be single buffer or parallel.

---
cherry-picked from ef80a36583bfb6f77856d47163f4662b8c620a39 

CI tests running


alanking:
CI tests passed.

trel:
then let's # it.

alanking:
#'d

Issue #2759, opened 2015-06-22
add rule priority to delayed rules


xu-hao:
- [x] master
- [x] 4-2-stable
----------

This feature defines priority order of execution after time-based selection, it does not create an interrupt-based system for rules to preempt currently/already executing rules.

- [x] `iqmod` can manipulate priority
- [x] `delay` "hints" should support setting `PRI` (varchar in db, but use as integer, larger is higher priority)
- [x] set default priority at `5` for all rules (allows new rules to be defined ahead or behind default)
- [x] `irodsReServer` should sort selected/ready rules by priority for execution
- [x] update documentation (plugins/pluggable_rule_engine/#delay-execution) to include `<PRI>`

- valid priority values will be integers within 1-9, inclusive
- invalid priority values will be set/interpreted as default value of `5`

trel:
@hao, can you write more about what this issue is meant to cover?

And what would be affected?

xu-hao:
Added more details. The only thing that was not clear to me was "upgrade support default rule". (This issue was originally based on a request from Jean-Yves.)

trel:
Okay, this seems to be two separate issues.   Maybe we can break them up...

1) One for the priority system (w/ iCommand support and tests), and
2) the other for naming the delay queue (w/ iCommand support and tests).

I'm not sure what "upgrade support default rule" should mean either, @nief?

nief:
I don't know what it means too :-).
I looked back at the email exchanges with you and Hao and I did not find anything related to this mysterious item: I think you can safely forget about it.

trel:
Okay, I've split this issue into two...

#2759 add rule priority to delayed rules

#3796 add rule name to delayed rules

xu-hao:
thanks

trel:
checkboxes checked.   closing!

PR #5530, opened 2021-04-08
[#2759] Delay server now honors a rule's priority (4-2-stable)

korydraughn:
Valid priority levels are defined by the following range: [1 - 9]

The delay server gathers and schedules rules in descending order of
priority level.

`<PRI>` tag is no longer supported by the delay(). Use `<PRIORITY>` instead.
Attempting to use `<PRI>` will result in an error.

Reoccurring rules that have an empty priority level will be updated to have a
priority level of 5 following successful execution. This also applies to
rules that have an invalid priority level.

### DIFF
--- a/plugins/database/src/db_plugin.cpp
+++ b/plugins/database/src/db_plugin.cpp
@@ -39,9 +39,11 @@
 #include "rods.h"
 #include "rcMisc.h"
 #include "miscServerFunct.hpp"
+#include "rodsErrorTable.h"
 
 // =-=-=-=-=-=-=-
 // stl includes
+#include <cstring>
 #include <sstream>
 #include <string>
 #include <string_view>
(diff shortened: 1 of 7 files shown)


korydraughn:
`iqmod` can set the priority to invalid values. The API that allows this needs to be updated to enforce valid priority values.

korydraughn:
How do we want to handle the `<PRI>` tag?
- Do we return an error when it is provided?
- Do we accept it and log a message about it being deprecated and up for removal in 4.3?
- Do we do something else?

trel:
Since there are zero known usages for it today, I think we're okay treating `PRI` as an error and providing a good error message.

trel:
nope - seems good.

korydraughn:
Added tests.

The tests don't check the log file for messages because their appearance is unreliable. However, I have verified at the bench that appropriate messages appear in the log file.

Running this PR through CI.

korydraughn:
passed CI

added #

PR #5536, opened 2021-04-13
[5520] use BinBytesBuf_PI for new API's (4-2-stable)

d-w-moore:
Use `BinBytesBuf_PI` (instead of `BytesBuf_PI`) as packing instructions when contained text fields are likely to contain XML special characters (ie `<`, `>`, `&`) or other arbitrary content possibly unsuitable for use with iRODS `XML_PROT` (JSON api requests and/or responses for example).

### DIFF
--- a/unit_tests/src/test_json_apis_from_client.cpp
+++ b/unit_tests/src/test_json_apis_from_client.cpp
@@ -1,0 +1,124 @@
+#include "catch.hpp"
+
+#include "atomic_apply_metadata_operations.h"
+#include "connection_pool.hpp"
+#include "dstream.hpp"
+#include "filesystem.hpp"
+#include "irods_at_scope_exit.hpp"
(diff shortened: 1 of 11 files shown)


korydraughn (commented):
See my comments on #5537

unit_tests/src/test_json_apis_from_client.cpp
```diff
@@ -0,0 +1,129 @@
...
+    auto conn_pool = irods::make_connection_pool(2);
+
+    rodsEnv env;
+    _getRodsEnv(env);
+
+    const auto sandbox = fs::path{env.rodsHome} / "unit_testing_sandbox";
+
+/*
+    auto conn = conn_pool->get_connection();
+*/
```
trel, line 79: if you want to leave this here, add some context... otherwise, just remove it.

PR #5537, opened 2021-04-13
[5520] use BinBytesBuf_PI for new API's (master)

d-w-moore:
Use `BinBytesBuf_PI` (instead of `BytesBuf_PI`) as packing instructions when contained text fields are likely to contain XML special characters (ie `<`, `>`, `&`) or other arbitrary content possibly unsuitable for use with iRODS `XML_PROT` (JSON api requests and/or responses for example).

### DIFF
--- a/plugins/api/src/atomic_apply_acl_operations.cpp
+++ b/plugins/api/src/atomic_apply_acl_operations.cpp
@@ -537,8 +537,8 @@ auto plugin_factory(const std::string& _instance_name,
                         RODS_API_VERSION,                           // API version
                         NO_USER_AUTH,                               // Client auth
                         NO_USER_AUTH,                               // Proxy auth
-                        "BytesBuf_PI", 0,                           // In PI / bs flag
-                        "BytesBuf_PI", 0,                           // Out PI / bs flag
+                        "BinBytesBuf_PI", 0,                        // In PI / bs flag
+                        "BinBytesBuf_PI", 0,                        // Out PI / bs flag
                         op,                                         // Operation
                         "api_atomic_apply_acl_operations",          // Operation name
                         nullptr,                                    // Null clear function
@@ -547,10 +547,10 @@ auto plugin_factory(const std::string& _instance_name,
 
     auto* api = new irods::api_entry{def};
 
-    api->in_pack_key = "BytesBuf_PI";
+    api->in_pack_key = "BinBytesBuf_PI";
     api->in_pack_value = BytesBuf_PI;
 
-    api->out_pack_key = "BytesBuf_PI";
+    api->out_pack_key = "BinBytesBuf_PI";
     api->out_pack_value = BytesBuf_PI;
 
     return api;
--- a/plugins/api/src/atomic_apply_metadata_operations.cpp
+++ b/plugins/api/src/atomic_apply_metadata_operations.cpp
@@ -523,8 +523,8 @@ auto plugin_factory(const std::string& _instance_name,
                         RODS_API_VERSION,                           // API version
                         NO_USER_AUTH,                               // Client auth
                         NO_USER_AUTH,                               // Proxy auth
-                        "BytesBuf_PI", 0,                           // In PI / bs flag
-                        "BytesBuf_PI", 0,                           // Out PI / bs flag
+                        "BinBytesBuf_PI", 0,                        // In PI / bs flag
+                        "BinBytesBuf_PI", 0,                        // Out PI / bs flag
                         op,                                         // Operation
                         "api_atomic_apply_metadata_operations",     // Operation name
                         nullptr,                                    // Null clear function
@@ -533,10 +533,10 @@ auto plugin_factory(const std::string& _instance_name,
 
     auto* api = new irods::api_entry{def};
 
-    api->in_pack_key = "BytesBuf_PI";
+    api->in_pack_key = "BinBytesBuf_PI";
     api->in_pack_value = BytesBuf_PI;
 
-    api->out_pack_key = "BytesBuf_PI";
+    api->out_pack_key = "BinBytesBuf_PI";
     api->out_pack_value = BytesBuf_PI;
 
     return api;
--- a/plugins/api/src/data_object_finalize.cpp
+++ b/plugins/api/src/data_object_finalize.cpp
@@ -410,8 +410,8 @@ auto plugin_factory(const std::string& _instance_name,
                         RODS_API_VERSION,                           // API version
                         NO_USER_AUTH,                               // Client auth
                         NO_USER_AUTH,                               // Proxy auth
-                        "BytesBuf_PI", 0,                           // In PI / bs flag
-                        "BytesBuf_PI", 0,                           // Out PI / bs flag
+                        "BinBytesBuf_PI", 0,                        // In PI / bs flag
+                        "BinBytesBuf_PI", 0,                        // Out PI / bs flag
                         op,                                         // Operation
                         "data_object_finalize",                     // Operation name
                         nullptr,                                    // Null clear function
@@ -420,10 +420,10 @@ auto plugin_factory(const std::string& _instance_name,
 
     auto* api = new irods::api_entry{def};
 
-    api->in_pack_key = "BytesBuf_PI";
+    api->in_pack_key = "BinBytesBuf_PI";
     api->in_pack_value = BytesBuf_PI;
 
-    api->out_pack_key = "BytesBuf_PI";
+    api->out_pack_key = "BinBytesBuf_PI";
     api->out_pack_value = BytesBuf_PI;
 
     return api;
--- a/plugins/api/src/experimental_api_plugin_adaptor.cpp
+++ b/plugins/api/src/experimental_api_plugin_adaptor.cpp
@@ -298,15 +298,15 @@ extern "C" {
 
         // =-=-=-=-=-=-=-
         // create a api def object
-        irods::apidef_t def{ADAPTER_APN,        // api number
-                            RODS_API_VERSION,   // api version
-                            NO_USER_AUTH,       // client auth
-                            NO_USER_AUTH,       // proxy auth
-                            "BytesBuf_PI", 0,   // in PI / bs flag
-                            "BytesBuf_PI", 0,   // out PI / bs flag
-                            std::function<int(rsComm_t*, bytesBuf_t*, bytesBuf_t**)>(adapter), // operation
+        irods::apidef_t def{ADAPTER_APN,                // api number
+                            RODS_API_VERSION,           // api version
+                            NO_USER_AUTH,               // client auth
+                            NO_USER_AUTH,               // proxy auth
+                            "BinBytesBuf_PI", 0,        // in PI / bs flag
+                            "BinBytesBuf_PI", 0,        // out PI / bs flag
+                            std::function<int(rsComm_t*, bytesBuf_t*, bytesBuf_t**)>(adapter),  // operation
                             "experimental_api_adaptor", // operation name
-                            nullptr,            // clear fcn
+                            nullptr,                    // clear fcn
                             (funcPtr)CALL_ADAPTOR};
         return new irods::api_entry(def);
     } // plugin_factory
--- a/plugins/api/src/get_file_descriptor_info.cpp
+++ b/plugins/api/src/get_file_descriptor_info.cpp
@@ -364,8 +364,8 @@ auto plugin_factory(const std::string& _instance_name,
                         RODS_API_VERSION,                // API version
                         NO_USER_AUTH,                    // Client auth
                         NO_USER_AUTH,                    // Proxy auth
-                        "BytesBuf_PI", 0,                // In PI / bs flag
-                        "BytesBuf_PI", 0,                // Out PI / bs flag
+                        "BinBytesBuf_PI", 0,             // In PI / bs flag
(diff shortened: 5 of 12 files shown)


trel:
end of comment in first commit is missing its closing parenthesis

trel:
this PR has an extra commit from 4-2-stable... for ADAPTOR_APN?

plugins/api/src/atomic_apply_metadata_operations.cpp
```diff
@@ -523,8 +523,8 @@ auto plugin_factory(const std::string& _instance_name,
                         RODS_API_VERSION,                           // API version
                         NO_USER_AUTH,                               // Client auth
                         NO_USER_AUTH,                               // Proxy auth
-                        "BytesBuf_PI", 0,                           // In PI / bs flag
-                        "BytesBuf_PI", 0,                           // Out PI / bs flag
+                        "BinBytesBuf_PI", 0,                           // In PI / bs flag
```
trel, line 526: line up the comments

plugins/api/src/atomic_apply_acl_operations.cpp
```diff
@@ -537,8 +537,8 @@ auto plugin_factory(const std::string& _instance_name,
                         RODS_API_VERSION,                           // API version
                         NO_USER_AUTH,                               // Client auth
                         NO_USER_AUTH,                               // Proxy auth
-                        "BytesBuf_PI", 0,                           // In PI / bs flag
-                        "BytesBuf_PI", 0,                           // Out PI / bs flag
+                        "BinBytesBuf_PI", 0,                           // In PI / bs flag
+                        "BinBytesBuf_PI", 0,                           // Out PI / bs flag
```
trel, line 541: line up the comments

plugins/api/src/data_object_finalize.cpp
```diff
@@ -410,8 +410,8 @@ auto plugin_factory(const std::string& _instance_name,
                         RODS_API_VERSION,                           // API version
                         NO_USER_AUTH,                               // Client auth
                         NO_USER_AUTH,                               // Proxy auth
-                        "BytesBuf_PI", 0,                           // In PI / bs flag
-                        "BytesBuf_PI", 0,                           // Out PI / bs flag
+                        "BinBytesBuf_PI", 0,                           // In PI / bs flag
+                        "BinBytesBuf_PI", 0,                           // Out PI / bs flag
```
trel, line 414: line up the comments

plugins/api/src/get_file_descriptor_info.cpp
```diff
@@ -364,8 +364,8 @@ auto plugin_factory(const std::string& _instance_name,
                         RODS_API_VERSION,                // API version
                         NO_USER_AUTH,                    // Client auth
                         NO_USER_AUTH,                    // Proxy auth
-                        "BytesBuf_PI", 0,                // In PI / bs flag
-                        "BytesBuf_PI", 0,                // Out PI / bs flag
+                        "BinBytesBuf_PI", 0,                // In PI / bs flag
+                        "BinBytesBuf_PI", 0,                // Out PI / bs flag
```
trel, line 368: line up the comments

plugins/api/src/replica_close.cpp
```diff
@@ -531,7 +531,7 @@ auto plugin_factory(const std::string& _instance_name,
                         RODS_API_VERSION,                 // API version
                         NO_USER_AUTH,                     // Client auth
                         NO_USER_AUTH,                     // Proxy auth
-                        "BytesBuf_PI", 0,                 // In PI / bs flag
+                        "BinBytesBuf_PI", 0,                 // In PI / bs flag
```
trel, line 534: line up the comments

plugins/api/src/replica_open.cpp
```diff
@@ -138,7 +138,7 @@ auto plugin_factory(const std::string& _instance_name,
                         NO_USER_AUTH,                    // Client auth
                         NO_USER_AUTH,                    // Proxy auth
                         "DataObjInp_PI", 0,              // In PI / bs flag
-                        "BytesBuf_PI", 0,                // Out PI / bs flag
+                        "BinBytesBuf_PI", 0,                // Out PI / bs flag
```
trel, line 141: line up the comments

plugins/api/src/touch.cpp
```diff
@@ -505,7 +505,7 @@ auto plugin_factory(const std::string& _instance_name,
                         RODS_API_VERSION,       // API version
                         NO_USER_AUTH,           // Client auth
                         NO_USER_AUTH,           // Proxy auth
-                        "BytesBuf_PI", 0,       // In PI / bs flag
+                        "BinBytesBuf_PI", 0,       // In PI / bs flag
```
trel, line 508: line up the comments

plugins/api/src/experimental_api_plugin_adaptor.cpp
```diff
@@ -302,8 +302,8 @@ extern "C" {
                             RODS_API_VERSION,   // api version
                             NO_USER_AUTH,       // client auth
                             NO_USER_AUTH,       // proxy auth
-                            "BytesBuf_PI", 0,   // in PI / bs flag
-                            "BytesBuf_PI", 0,   // out PI / bs flag
+                            "BinBytesBuf_PI", 0,   // in PI / bs flag
+                            "BinBytesBuf_PI", 0,   // out PI / bs flag
```
trel, line 306: line up the comments

plugins/api/src/atomic_apply_metadata_operations.cpp
d-w-moore: now done on both branches, correcting comment alignment

unit_tests/src/test_json_apis_from_client.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+    auto conn_pool = irods::make_connection_pool(2);
+
+    rodsEnv env;
+    _getRodsEnv(env);
+
+    const auto sandbox = fs::path{env.rodsHome} / "unit_testing_sandbox";
+    auto conn = conn_pool->get_connection();
+
+    REQUIRE(fs::client::create_collection(conn, sandbox));
```
korydraughn, line 75: This collection is never deleted. Use `at_scope_exit` to delete it. See other tests for examples.

unit_tests/src/test_json_apis_from_client.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+TEST_CASE("test atomic metadata with BinBytesBuf", "[xml]")
+{
+    char* save_protocol;
+
+    change_environment_variable("irodsProt",save_protocol,"1");
+
+    load_client_api_plugins();
+
+    auto conn_pool = irods::make_connection_pool(2);
```
korydraughn, line 67: This creates two connections, but only one is used throughout the test.

unit_tests/src/test_json_apis_from_client.cpp
```diff
@@ -0,0 +1,110 @@
...
+                {"value", "john <&doe&>"},
+                {"units", ""}
+            }
+        })}
+    }.dump();
+
+    char* json_error_string{};
+    irods::at_scope_exit free_memory{[&json_error_string] { std::free(json_error_string); }};
+
+    auto* conn_ptr = &static_cast<rcComm_t&>(conn);
```
korydraughn, line 98: `conn` can be casted directly to a pointer.

unit_tests/src/test_json_apis_from_client.cpp
```diff
@@ -0,0 +1,110 @@
...
+            charptr_tmp = nullptr;
+        }
+    }
+}
+
+TEST_CASE("test atomic metadata with BinBytesBuf", "[xml]")
+{
+    char* save_protocol;
+
+    change_environment_variable("irodsProt",save_protocol,"1");
```
korydraughn, line 63: This will cause a memory leak if an exception is thrown. You can use `std::string` to avoid this or `at_scope_exit` to free the allocated memory.

d-w-moore: I believe `save_protocol` ends up pointing at the application's environment string space, so I'm not sure how I could free it without potentially corrupting the heap. But I am open to better ways of changing environment variables if they exist !

unit_tests/src/test_json_apis_from_client.cpp
d-w-moore: Should I leave this at a max 2 conns in the event the at_scope_exit brings up a new connection while deleting the sandbox, but before the other connection has been deleted?

unit_tests/src/test_json_apis_from_client.cpp
d-w-moore: yep- agreed

unit_tests/src/test_json_apis_from_client.cpp
korydraughn: you call `strdup` on the environment variable so it points to memory you are responsible for.

unit_tests/src/test_json_apis_from_client.cpp
korydraughn: The code doesn't do anything asynchronously so one connection should be safe. You can reuse the connection handle obtained from the first call to `conn_pool->get_connection()`. You could replace the connection pool with `irods::experimental::client_connection` if you wanted. A connection pool isn't necessary for the test, but isn't wrong either.

unit_tests/src/test_json_apis_from_client.cpp
d-w-moore: you are right. my (brain) memory seems to have failed me.... thanks!

unit_tests/src/test_json_apis_from_client.cpp
```diff
@@ -0,0 +1,110 @@
...
+    irods::at_scope_exit free_memory{[&json_error_string] { std::free(json_error_string); }};
+
+    auto* conn_ptr = &static_cast<rcComm_t&>(conn);
+    REQUIRE(rc_atomic_apply_metadata_operations(conn_ptr, json_input.c_str(), &json_error_string) == 0);
+    REQUIRE(json_error_string == "{}"s);
+
+    restore_environment_variable("irodsProt",save_protocol);
+
+    const auto gql = fmt::format("select COUNT(COLL_ID) where COLL_NAME = '{}' and META_COLL_ATTR_VALUE like '{}'",
+                                 sandbox.string().c_str(), "%" "<&" "%" "&>" "%");
```
korydraughn, line 105: Why chop the third argument to `fmt::format` like this? Just curious.

d-w-moore: I thought it had better clarity, but it's a subjective thing of course.

unit_tests/src/test_json_apis_from_client.cpp
d-w-moore: changed the pool to be size 1 and it hung ... not sure why .  (New issue possibly? ) But this unit test runs flawlessly and cleans up after itself as long as I keep the pool at size 2.

unit_tests/src/test_json_apis_from_client.cpp
d-w-moore: left strings in concatenated form but parenthesized to make clear it's a single argument

d-w-moore:
Seems ready for the squash and # ...
There is the matter of the too-large irods connection pool (size 2 when 1 would do -- unresolved conversation above) to avoid causing a hang, but not sure that intersects with this issue.

trel:
did you try with `irods::experimental::client_connection`?     otherwise, yes, please squash to single commit and we'll eyeball again before #.

d-w-moore:
Using the `client_connection` worked at bench.  Doing one final round of unit test in CI, cherry picked to master.

unit_tests/src/test_json_apis_from_client.cpp
```diff
@@ -0,0 +1,129 @@
...
+    irods::at_scope_exit free_memory{[&json_error_string] { std::free(json_error_string); }};
+
+    auto* conn_ptr = static_cast<rcComm_t*>(conn);
+    REQUIRE(rc_atomic_apply_metadata_operations(conn_ptr, json_input.c_str(), &json_error_string) == 0);
+    REQUIRE(json_error_string == "{}"s);
+
+    restore_environment_variable("irodsProt",save_protocol);
+
+    const auto gql = fmt::format("select COUNT(COLL_ID) where COLL_NAME = '{}' and META_COLL_ATTR_VALUE like '{}'",
+                                 sandbox.string().c_str(),
```
korydraughn, line 122: while this is not wrong, it does result in a temporary. the temporary is avoided by using `sandbox.c_str()` instead.

d-w-moore:
Passing tests in CI, squashed here and on 4-2-stable. Awaiting final eyes.

trel:
so i think the client connection discussion above can be resolved, and... the one temporary kory mentioned is the only thing left?

unit_tests/src/test_json_apis_from_client.cpp
d-w-moore: Corrected this and ran again in CI with a green.

d-w-moore:
> so i think the client connection discussion above can be resolved, and... the one temporary kory mentioned is the only thing left?

Yes, squashing and #'ing now

trel:
okay, so you went back to a connection_pool(2).      will merge, but we should probably understand that a bit more moving forward.

Issue #5360, opened 2021-01-21
iphymv (et al?) should exit with more specific values


mcv21:
- [x] master server
- [x] master icommands
- [x] 4-2-stable server
- [x] 4-2-stable icommands
--------

## Feature

We would like to use some icommands in an automated fashion; doing so usefully involves knowing what the cause of a failure was. The easiest way to do this is to inspect the return value (i.e. what was passed to `exit();` mod 255).

Currently iphymv only uses a few non-zero values - `1` for usage error or environment error, `2` for no input or connection error, 7 for login error, and 3 for all other errors. It would be nice to use the space more systematically and to differentiate more failure modes (rather than us as users having to parse whatever appears on stderr).

Less abstractly, we're using `iphymv` to migrate data off old hardware onto new hardware, and there are (at least) 3 failures it would be nice to differentiate between since we want to handle them differently:

1. No space left on destination (stop this migration, move to next target)
2. Network failures of various sorts (try exponential back-off a few times then give up)
3. checksum error or file not found on filesystem (give up on this object, move on to the next)

Having to do this with ad-hoc parsing of stderr from iphymv is a pain, and doing it based on `$?` would be much nicer :)

alanking:
I am finally swinging around to look at this.

What specific error code(s) correspond(s) with each item in your list? And what return code(s) would you expect upon receipt of said error code(s)?

I would guess at least some of the following may apply, though I imagine there are more...

1. `USER_FILE_TOO_LARGE` (maybe `UNIX_FILE_GET_FS_FREESPACE_ERR`?)
2. `SYS_GET_HOSTNAME_ERR`, `SYS_SVR_TO_SVR_CONNECT_FAILED`, `SYS_INVALID_SERVER_HOST`, etc.?
3. `USER_CHKSUM_MISMATCH`, `UNIX_FILE_OPEN_ERR` (with UNIX code of 2)

It's difficult for us to classify some iRODS error codes as some of them are so generic (e.g. "user errors", "system errors", etc.) that it's difficult to define a range of errors to catch. So, if you have some specific errors you are sniffing or encounter a lot, that could be helpful in developing this enhancement. There are some that are grouped nicely, however, for example:

https://github.com/irods/irods/blob/d6c0eb90d5425242ddcf4a95a08cd58d7904a6d2/lib/core/include/rodsErrorTable.h#L313-L341

[...]

PR #5555, opened 2021-04-21
[#5360] Add iphymv exit code tests (master)

alanking:
Still need to run the mungefs tests, but here is the start. Companion PR is irods/irods_client_icommands#201

### DIFF
--- a/plugins/database/src/db_plugin.cpp
+++ b/plugins/database/src/db_plugin.cpp
@@ -2892,6 +2892,8 @@ irods::error db_reg_replica_op(
 //        int IX_RESC_GROUP_NAME = 7; /* index into theColls */
     const int IX_RESC_ID = 10;
     const int IX_DATA_PATH = 11;    /* index into theColls */
+    const int IX_REPLICA_STATUS = 14;
+    const int IX_DATA_STATUS = 15;
 
     const int IX_DATA_MODE = 19;
     const int IX_CREATE_TS = 21;
@@ -2991,6 +2993,20 @@ irods::error db_reg_replica_op(
     cVal[IX_RESC_ID]       = (char*)resc_id_str.c_str();
     cVal[IX_DATA_PATH]       = _dst_data_obj_info->filePath;
     cVal[IX_DATA_MODE]       = _dst_data_obj_info->dataMode;
+
+    // The caller has requested that the replica be registered as intermediate. This
+    // means that the replica will be written or changed at a future time. Otherwise,
+    // the replica will take the replica status of the source replica. The const must
+    // be cast away due to cVal being a char*[] and not a const char*[].
+    char* intermediate_replica_status = const_cast<char*>(std::to_string(INTERMEDIATE_REPLICA).data());
+    if (getValByKey(_cond_input, REGISTER_AS_INTERMEDIATE_KW)) {
+        cVal[IX_REPLICA_STATUS] = intermediate_replica_status;
+    }
+
+    // data_status tracks replica status for logical locking - this is a new replica,
+    // so make sure the data_status column is empty at registration time.
+    std::snprintf(_dst_data_obj_info->statusString, NAME_LEN, "");
+    cVal[IX_DATA_STATUS] = _dst_data_obj_info->statusString;
 
     getNowStr( myTime );
     cVal[IX_MODIFY_TS] = myTime;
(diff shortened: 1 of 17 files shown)


trel:
very nice.    i see all seven error codes.

alanking:
I added more assertions to the `iphymv` tests as well as some checksum mismatch tests for `icp`, `iput`, and `irepl` to test behavior in those API endpoints as well. This uncovered an issue in `icp` where error codes in finalize were not being propagated back to the user (although the replica was correctly staled). I need to do another CI run but my Jenkins is down, so might take a while to get there.

plugins/database/src/db_plugin.cpp
```diff
@@ -2992,6 +2994,19 @@ irods::error db_reg_replica_op(
     cVal[IX_DATA_PATH]       = _dst_data_obj_info->filePath;
     cVal[IX_DATA_MODE]       = _dst_data_obj_info->dataMode;
 
+    // The caller has requested that the replica be registered as intermediate. This
+    // means that the replica will be written or changed at a future time. Otherwise,
+    // the replica will take the replica status of the source replica.
+    char* intermediate_replica_status = "2";
```
trel, line 3000: this isn't defined somewhere else already as '2'?

server/api/src/rsDataObjUnlink.cpp
```diff
@@ -547,125 +564,95 @@ int dataObjUnlinkS(rsComm_t* rsComm,
...
+        const auto cond_input = irods::experimental::make_key_value_proxy(dataObjUnlinkInp->condInput);
+        info.in_pdmo(cond_input.contains(IN_PDMO_KW) ? cond_input.at(IN_PDMO_KW).value() : "");
+
+        if (const auto ec = l3Unlink(rsComm, info.get()); ec < 0) {
+            irods::log(LOG_NOTICE, fmt::format(
+                "[{}:{}] - failed to physically unlink replica "
+                "[error_code=[{}], path=[{}], hierarchy=[{}]",
+                __FUNCTION__, __LINE__, ec, info.logical_path(), info.hierarchy()));
+
+            // TODO: why do we allow NOENT and EACCES to unregister?
```
trel, line 626: b/c the physical file already has a problem?   NOENT means it's not even there... and EACCES means the service account can't read it anyway...

server/api/src/rsDataObjUnlink.cpp
```diff
@@ -547,125 +564,95 @@ int dataObjUnlinkS(rsComm_t* rsComm,
...
-                    }
-                    break;
-                }
-                return status;
-            }
-            else {
-                status = 0;
-            }
-        }
+        // TODO: Should the replica be marked stale if the unregister itself fails?
```
trel, line 649: if the data on disk and the entry in the catalog are known to be good... we should leave it good.   for *anything* other case, we can't vouch for it, so yes, stale it.

alanking: I think I will revisit this on the pass to prevent unlinks on locked data objects.

server/api/src/rsDataObjUnlink.cpp
alanking: I think I will revisit this on the pass to prevent unlinks on locked data objects.

plugins/database/src/db_plugin.cpp
alanking: What I normally do is a `std::to_string` on the `INTERMEDIATE_REPLICA` enumerated value, but that won't play well here as `cVal` is an array of pointer to non-const `char`s. I could cast away the const or perform a `std::snprintf` to a pre-existing char array. Casting away the const seems pretty safe given it's coming from a temporary anyway.

alanking:
Force push was from an accidental rebase, but no conflicts were found. Addressed comment in db_plugin. Changes before latest commit passed core tests in CI, awaiting unit tests and timing tests.

alanking:
Everything passed except for a couple of unit test assertions which I found are likely due to timing issues in updating the catalog on agent teardown. I'm going to address those in the test commit.

alanking:
All tests passed in CI. Squashed down commits to RST work, tests, and phymv updates.

trel:
in your 5360 commit message ... "propogating" -> "propagating"

everything else looks great.

alanking:
Fixed commit message and #'d

trel:
awaiting green

PR #5558, opened 2021-04-22
Little fixes (4-2-stable)

korydraughn:
Passed CI.

### DIFF
--- a/lib/core/include/irods_server_properties.hpp
+++ b/lib/core/include/irods_server_properties.hpp
@@ -9,6 +9,7 @@
 #include "irods_exception.hpp"
 
 #include <string>
+#include <string_view>
 
 namespace irods
 {
@@ -59,6 +60,10 @@ namespace irods
         /// @brief capture server_config.json
         void capture_json(const std::string& _filename);
 
+        bool contains(const std::string_view _key) const {
+            return config_props_.has_entry(_key);
+        }
+
         template< typename T >
         T& get_property( const std::string& _key ) {
             return config_props_.get< T >( _key );
@@ -95,6 +100,10 @@ namespace irods
         /// @brief properties lookup table
         configuration_parser config_props_;
     }; // class server_properties
+
+    inline bool server_property_exists(const std::string_view _prop) {
+        return irods::server_properties::instance().contains(_prop);
+    }
 
     template< typename T >
     T& get_server_property( const std::string& _prop ) {
--- a/server/core/src/client_api_whitelist.cpp
+++ b/server/core/src/client_api_whitelist.cpp
@@ -12,15 +12,7 @@ namespace
 {
     auto is_client_to_agent_connection() -> bool
     {
-        try {
-            irods::get_server_property<std::string>(irods::AGENT_CONN_KW);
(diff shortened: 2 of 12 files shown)


korydraughn:
@jasoncoposky  @alanking  Are there any translation issues (throwing -> non-throwing) around the `AGENT_CONN_KW`?

server/core/src/client_api_whitelist.cpp
```diff
@@ -12,6 +12,7 @@ namespace
 {
     auto is_client_to_agent_connection() -> bool
     {
+#if 0
```
trel, line 15: torch if happy with the #else

server/core/src/dataObjOpr.cpp
```diff
@@ -830,6 +830,7 @@ chkOrphanFile(
     addKeyVal( &genQueryInp.condInput, DISABLE_STRICT_ACL_KW, "disable" );
 
     std::string svr_sid;
+#if 0
```
trel, line 833: same

lib/core/include/irods_server_properties.hpp
```diff
@@ -96,6 +101,10 @@ namespace irods
         configuration_parser config_props_;
     }; // class server_properties
 
+    inline bool server_property_exists(const std::string_view _prop) {
```
alanking, line 104: Thoughts on the name `server_property_contains`?

server/core/src/dataObjOpr.cpp
```diff
@@ -848,6 +849,22 @@ chkOrphanFile(
             return e.code();
         }
     }
+#else
```
alanking, line 852: I'm assuming the #if 0 block will be deleted in a future commit

alanking (approved):
The new code looks like it would behave equivalently and is much nicer. :)

server/core/src/dataObjOpr.cpp
korydraughn: the plan was to let the team see the old code beside the new code and verify that they are equivalent.

lib/core/include/irods_server_properties.hpp
alanking: I was just thinking about our other key-value things lining up with maps terminology. I am fine with this name too

korydraughn: i think `exists` works better given the preceding characters. if we use `contains`, i feel `server_property_` needs to be tweaked, else it sounds like `server_property_contains` is asking `does the server property contain _prop`. hopefully that made sense.

alanking: Totally

server/core/src/client_api_whitelist.cpp
korydraughn: done

server/core/src/dataObjOpr.cpp
korydraughn: done

trel:
i'm good for #s.

korydraughn:
added #

PR #5559, opened 2021-04-23
Little fixes (master)

korydraughn:
Waiting for CI results

### DIFF
--- a/lib/core/include/irods_server_properties.hpp
+++ b/lib/core/include/irods_server_properties.hpp
@@ -9,6 +9,7 @@
 #include "irods_exception.hpp"
 
 #include <string>
+#include <string_view>
 
 namespace irods
 {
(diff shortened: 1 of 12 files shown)


lib/core/include/irods_server_properties.hpp
```diff
@@ -56,6 +57,10 @@ namespace irods
         /// @brief capture server_config.json
         void capture_json(const std::string& _filename);
 
+        bool exists(const std::string_view _key) const {
```
korydraughn, line 60: @alanking @trel I can see this being renamed to `contains`. What do you think?

trel: if we're doing full name comparison...  i think `exists` is fine/better.

korydraughn: I changed it to `contains` before you responded. Figured it made sense because the `key_value_proxy` uses `contains` instead of `exists`.

korydraughn:
Passed CI

Issue #5496, opened 2021-03-22
A force-put (e.g. iput -f) should erase the existing checksum


korydraughn:
- [x] master
- [x] 4-2-stable
---
## Feature
Erasing the checksum is necessary so that the catalog stays consistent. Right now, a force-put overwrites the replica and recalculates the checksum if one exists.

**This is a change in behavior!**

trel:
Note, this is `iput -f` without the `-k` or `-K` options.

Either of those options would calculate and store the new checksum in the catalog.

korydraughn:
Should writing to an existing replica cause the checksum to be erased? For example, using `istream` to overwrite part of a replica.

IMO, yes because the checksum will be incorrect if the data actually changed.

trel:
Yes, we don't want incorrect information in the catalog.

And we shouldn't assume the user wants to calculate a checksum on an arbitrarily large file.

korydraughn:
This also affects bulk put.

PR #5564, opened 2021-04-26
[#5496] Checksums are erased on overwrite of replica (4-2-stable)

korydraughn:
Passed CI before trace log message removal. Running through CI again.

Do we want to keep any of the trace messages for 4-2-stable? master? Click on commit ba949ed to see the removed messages.

### DIFF
--- a/server/api/src/rsDataObjClose.cpp
+++ b/server/api/src/rsDataObjClose.cpp
@@ -130,7 +130,7 @@ namespace
 
         auto destination_replica = ir::replica_proxy_t{*l1desc.dataObjInfo};
 
-        // This section only exists for the compound resource as replication handles its own finalizing logic
+        // This section only exists for the compound resource as replication handles its own finalizing logic.
         if (oprType == REPLICATE_DEST) {
             const int srcL1descInx = l1desc.srcL1descInx;
             if (srcL1descInx < 3) {
@@ -176,7 +176,7 @@ namespace
             }
         }
 
-        if (!l1desc.chksumFlag) {
+        if (0 == l1desc.chksumFlag) {
             if (destination_replica.checksum().empty()) {
                 return {};
             }
@@ -188,6 +188,7 @@ namespace
                 return irods::verify_checksum(_comm, *destination_replica.get(), l1desc.chksum);
             }
 
+            // Specifically for compound resources.
             if (REPLICATE_DEST == oprType) {
                 if (!destination_replica.checksum().empty()) {
                     destination_replica.cond_input()[ORIG_CHKSUM_KW] = destination_replica.checksum();
@@ -212,7 +213,7 @@ namespace
                     }
                 }
 
-                return {checksum_string};
+                return checksum_string;
             }
 
             // TODO come back to this case...
@@ -249,7 +250,7 @@ namespace
                     }
                 }
 
-                return {checksum_string};
+                return checksum_string;
             }
 
             return {};
@@ -417,7 +418,8 @@ namespace
         }
 
         if (const bool verify_size = !getValByKey(&l1desc.dataObjInp->condInput, NO_CHK_COPY_LEN_KW);
-            verify_size && l1desc.dataSize > 0 && size_in_vault != l1desc.dataSize) {
+            verify_size && l1desc.dataSize > 0 && size_in_vault != l1desc.dataSize)
+        {
             // The size in the vault does not match the expected size in the catalog.
             // Update the size in the RST and unlock the data object and throw.
             r.size(size_in_vault);
@@ -449,16 +451,34 @@ namespace
         auto r = ir::make_replica_proxy(*l1desc.dataObjInfo);
 
         try {
-            const auto size_should_be_verified    = !getValByKey(&l1desc.dataObjInp->condInput, NO_CHK_COPY_LEN_KW);
-            const auto replica_opened_for_write   = OPEN_FOR_WRITE_TYPE == l1desc.openType || CREATE_TYPE == l1desc.openType;
-            const auto replica_has_a_checksum     = !r.checksum().empty();
-            const auto checksum_should_be_updated = replica_opened_for_write && replica_has_a_checksum;
+            if (PUT_OPR == l1desc.oprType) {
+                if (REG_CHKSUM == l1desc.chksumFlag || VERIFY_CHKSUM == l1desc.chksumFlag) {
+                    // Update the replica proxy's checksum value so that if an exception is thrown,
+                    // the checksum value will still be stored in the catalog.
+                    r.checksum(irods::register_new_checksum(_comm, *r.get(), l1desc.chksum));
+                    rst::update(r.data_id(), r);
 
-            if (checksum_should_be_updated) {
-                l1desc.chksumFlag = REG_CHKSUM;
+                    if (VERIFY_CHKSUM == l1desc.chksumFlag) {
+                        rodsLog(LOG_DEBUG,
+                                "[%s:%d] Verifying checksum [calculated_checksum=%s, checksum_from_client=%s, path=%s, replica_number=%d] ...",
+                                __func__, __LINE__, r.checksum().data(), l1desc.chksum, r.logical_path().data(), r.replica_number());
+
+                        if (r.checksum() != l1desc.chksum) {
+                            THROW(USER_CHKSUM_MISMATCH, fmt::format("[{}:{}] Mismatch checksum for {}.inp={}, compute {}",
+                                  __FUNCTION__, __LINE__, r.logical_path(), l1desc.chksum, r.checksum()));
+                        }
+                    }
+
+                    // Copy the new checksum value into the conditional input so that file modified is triggered.
+                    cond_input[CHKSUM_KW] = r.checksum();
+                }
+
+                return;
             }
 
-            if (checksum_should_be_updated || size_should_be_verified) {
+            const auto size_should_be_verified = !getValByKey(&l1desc.dataObjInp->condInput, NO_CHK_COPY_LEN_KW);
+
+            if (size_should_be_verified) {
                 if (const std::string checksum = perform_checksum_operation_for_finalize(_comm, _fd); !checksum.empty()) {
                     cond_input[CHKSUM_KW] = checksum;
                 }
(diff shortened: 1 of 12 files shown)


server/api/src/rsDataObjPut.cpp
```diff
@@ -123,6 +115,9 @@ namespace
         reg_param[REPL_STATUS_KW] = std::to_string(STALE_REPLICA);
         reg_param[STALE_ALL_INTERMEDIATE_REPLICAS_KW] = "";
         if (cond_input.contains(ADMIN_KW)) {
+            // TODO Shouldn't this always be set?
+            // The user's perms would need to be elevated as well.
+            // rsModDataObjMeta should never fail due to perms here.
             reg_param[ADMIN_KW] = "";
         }
```
korydraughn, line 122: Shouldn't `ADMIN_KW` always be set here? The existence check seems unnecessary.

unit_tests/src/test_rc_data_obj.cpp
```diff
@@ -548,3 +664,13 @@ auto create_empty_replica(const fs::path& _path) -> void
     REQUIRE(GOOD_REPLICA == replica::replica_status(comm, _path, 0));
 } // create_empty_replica
 
+auto create_local_file_40mb(const std::string_view _path, char _ch) -> void
```
korydraughn, line 667: This could be made a little more generic and moved into the unit test utility header.

unit_tests/src/test_replica.cpp
```diff
@@ -121,8 +122,15 @@ TEST_CASE("replica", "[replica]")
...
-        REQUIRE(replica::replica_checksum(comm, target_object, second_replica) == expected_checksum);
+        // show that the checksum has been erased
+        {
+            const auto gql = fmt::format("select DATA_CHECKSUM where COLL_NAME = '{}' and DATA_NAME = '{}' and DATA_REPL_NUM = '{}'",
+                                         target_object.parent_path().c_str(),
+                                         target_object.object_name().c_str(),
+                                         second_replica);
+
+            REQUIRE(irods::query{&comm, gql}.front()[0].empty());
+        }
```
korydraughn, line 133: Look into replacing this with the replica library.

unit_tests/src/test_replica.cpp
```diff
@@ -179,8 +187,15 @@ TEST_CASE("replica", "[replica]")
...
-        REQUIRE(replica::replica_checksum(comm, target_object, resc_name) == expected_checksum);
+        // show that the checksum has been erased
+        {
+            const auto gql = fmt::format("select DATA_CHECKSUM where COLL_NAME = '{}' and DATA_NAME = '{}' and DATA_RESC_NAME = '{}'",
+                                         target_object.parent_path().c_str(),
+                                         target_object.object_name().c_str(),
+                                         resc_name);
+
+            REQUIRE(irods::query{&comm, gql}.front()[0].empty());
+        }
```
korydraughn, line 198: Look into replacing this with the replica library.

unit_tests/src/test_rc_data_obj.cpp
```diff
@@ -69,6 +70,59 @@ TEST_CASE("open,read,write,close")
...
+        SECTION("#5496")
+        {
+            std::string_view path_str = target_object.c_str();
+
+            irods::experimental::client_connection conn;
+            RcComm& comm = static_cast<RcComm&>(conn);
+
+            dataObjInp_t open_inp{};
+            std::snprintf(open_inp.objPath, sizeof(open_inp.objPath), "%s", path_str.data());
+            open_inp.openFlags = O_CREAT | O_TRUNC | O_WRONLY;
```
korydraughn, line 82: This is unnecessary given that the server overwrites these.

server/api/src/rsDataObjPut.cpp
trel: it's just propagating it if it exists, right now, right?   so the question is what bad can happen if we just always set it?

korydraughn:
passed CI again

trel:
> this could be kept as a LOG_DEBUG

this was at https://github.com/irods/irods/commit/ba949ed51fca47c738ee7ed50ac172e1463a653f#diff-5201f58857336c4ef35965351ba3331f50a9fa1ac20c34fe48c0ca341a3c9bd4L465

server/api/src/rsDataObjPut.cpp
korydraughn: this code block's job is to update the db with the correct info. imo, the update should never fail because of the user's lack of access level.

i guess the real question is whether a non-admin user can successfully execute this code path without error.

alanking: Could it fail due to permissions because the permissions model for opening is different from that of modifying the system metadata (i.e. `rsModDataObjMeta`)? Certainly not a normal (or even fully defensible) use case, but... possible?

korydraughn: I hit this error case before. Looking back now, I can see it was because the `ADMIN_KW` was set while I was a `rodsuser`. 
https://github.com/irods/irods/blob/e4fd2a6a59dbc85c376e30e0481d8f3339da2893/plugins/database/src/db_plugin.cpp#L2206-L2211

I think this is okay.

server/core/src/initServer.cpp
```diff
@@ -133,15 +95,12 @@ namespace
             r.mtime(SET_TIME_TO_NOW_KW);
         }
 
-        if (cond_input.contains(CHKSUM_KW)) {
-            r.checksum(cond_input.at(CHKSUM_KW).value());
-        }
-
+        r.checksum(""); // TODO Should the checksum be cleared?
```
korydraughn, line 98: @trel Should the checksum field be erased if the client does not call close? Could do this based on whether bytes were written.

trel: If bytes were written, we definitely clear...

if no bytes were written... I think it would be okay to leave it as it was... could that be misleading in any case?

trel: counter question... any benefit to leaving it as is...?

trel:
> > this could be kept as a LOG_DEBUG
> 
> this was at [ba949ed#diff-5201f58857336c4ef35965351ba3331f50a9fa1ac20c34fe48c0ca341a3c9bd4L465](https://github.com/irods/irods/commit/ba949ed51fca47c738ee7ed50ac172e1463a653f#diff-5201f58857336c4ef35965351ba3331f50a9fa1ac20c34fe48c0ca341a3c9bd4L465)

not sure you saw this - i think adding a debug here makes sense.

korydraughn:
Passed CI.

server/core/src/initServer.cpp
korydraughn: Erasing the checksum if bytes were written seems safe. Any operation that truncates a replica on open will automatically erase the checksum too, which is good.

As for the counter question, I don't know if there are benefits to leaving the checksum as is. I'm leaning towards it being more confusing.

trel: i think so too.   in an error case, i think we should torch it - nobody will have a false positive that way.  and worst case, they have to recalculate on a known good replica.

korydraughn:
> > > this could be kept as a LOG_DEBUG
> > 
> > 
> > this was at [ba949ed#diff-5201f58857336c4ef35965351ba3331f50a9fa1ac20c34fe48c0ca341a3c9bd4L465](https://github.com/irods/irods/commit/ba949ed51fca47c738ee7ed50ac172e1463a653f#diff-5201f58857336c4ef35965351ba3331f50a9fa1ac20c34fe48c0ca341a3c9bd4L465)
> 
> not sure you saw this - i think adding a debug here makes sense.

I did see your comment. I can add the trace message back it. I imagine we want to include info that identifies the replica as well?

Do we want to start introducing trace messages throughout the code? A lot of the messages (previously in this PR) around closing a replica were extremely helpful for understanding what was happening.

server/api/src/rsDataObjClose.cpp
```diff
@@ -449,16 +450,42 @@ namespace
...
+            //const auto checksum_should_be_updated = replica_opened_for_write && replica_has_a_checksum;
+            // clang-format on
+
+            // TODO This appears to be the auto-update case. Lets disable it and see what CI
+            // complains about.
+            //if (checksum_should_be_updated) {
+                //l1desc.chksumFlag = REG_CHKSUM;
+            //}
+
+            //if (checksum_should_be_updated || size_should_be_verified) {
```
korydraughn, line 487: This code is safe to remove given that this PR passed CI.

trel:
> I imagine we want to include info that identifies the replica as well?

Yes, please.

The other lines didn't seem as useful to someone after this logic and functionality is developed and working.   But I'm certainly not against having more LOG_DEBUG in places if that feels helpful.   Up to you.

server/api/src/rsDataObjClose.cpp
trel: agreed

server/api/src/rsDataObjClose.cpp
```diff
@@ -449,16 +450,42 @@ namespace
...
+                }
 
-            if (checksum_should_be_updated) {
-                l1desc.chksumFlag = REG_CHKSUM;
+                return;
             }
 
-            if (checksum_should_be_updated || size_should_be_verified) {
+            // clang-format off
+            const auto size_should_be_verified    = !getValByKey(&l1desc.dataObjInp->condInput, NO_CHK_COPY_LEN_KW);
```
korydraughn, line 475: This can probably be removed as well. The `NO_CHK_COPY_LEN_KW` is not used anymore.

korydraughn: This line also happens to be true almost 100% of the time.

alanking: I think `NO_CHK_COPY_LEN_KW` is used in the compound resource when opening a replica on the archive resource: https://github.com/irods/irods/blob/5f4a213c79711fab65375ad3da9b89aa5184377c/plugins/resources/compound/libcompound.cpp#L1612-L1615

korydraughn: Cool. I'll leave this as is.

server/api/src/rsDataObjClose.cpp
korydraughn: done

server/core/src/initServer.cpp
korydraughn: What do you mean by error case?

The latest changes and tests show that the agent can erase the checksum based on whether bytes were written or not. Is that still what we want or did we want to always erase the checksum if the client doesn't call close?

trel: sorry, not error case... just unknown reason for 'lack of close'.

i think we agreed if the client doesn't call close on a write operation, they don't get a checksum generated, so... we should clear the checksum, regardless.  it can't be misleading that way.

korydraughn: Ok. I'll update the code to always erase the checksum.

trel:
looks good.   more CI?  or ready to # if green?

korydraughn:
A replication/compound resource test failed. I need to verify if the failure was real.

korydraughn:
I ran the test at the bench and it passed.

Are there any last changes I need to make before adding the #?

korydraughn:
Just an fyi, I rebased on top of 42s and ran the test, test_resource_types.Test_Resource_ReplicationToTwoCompoundResourcesWithPreferArchive.test_iphymv_admin_mode, at the bench.

korydraughn:
I'll rerun it though CI one more time just to be safe.

trel:
yes please.   thanks.

korydraughn:
passed CI

added #

Issue #3091, opened 2016-03-09
Cancelling an 'iput -f' breaks the integrity of a replica whilst preserving the previous checksum.


triitaojcsc:
- [x] master
- [x] 4-2-stable
-------

Cancelling an `iput -f` breaks the integrity of a replica whilst preserving the previous checksum.
# Environment

```
[user@localhost ~]$ cat /etc/redhat-release
CentOS release 6.6 (Final)

[user@localhost ~]$ rpm -qa | grep irods
irods-database-plugin-postgres93-1.6-0.x86_64
irods-icat-4.1.5-0.x86_64
```

Note: Also reproducible on 4.1.8.

```
[user@localhost ~]$ ilsresc --tree
testrepl:replication
├── res3
└── res4
```
# Changes in core.re

```
acPostProcForPut {
    msiSysChksumDataObj;
}
acPostProcForRepl {
    msiSysChksumDataObj;
}
```
# Starting situation

User has uploaded a file '100M', which has two replicas, two identical checksums and two identical file sizes.

```
[user@localhost ~]$ ils -L 100M
  user          0 testrepl;res4    104857600 2016-03-09.11:20 & 100M
    sha2:IEkqTQ2E+L6xdn9mFiKfhdRMKCe2S9v7Jg7hL6EQng4=    generic    /var/lib/irods/iRODS/Vault-res4/home/user/100M
  user          1 testrepl;res3    104857600 2016-03-09.11:20 & 100M
    sha2:IEkqTQ2E+L6xdn9mFiKfhdRMKCe2S9v7Jg7hL6EQng4=    generic    /var/lib/irods/iRODS/Vault-res3/home/user/100M
```
# Action that creates the problem

User decides to overwrite the '100M' file with a 2147483648 byte (2GB) file, but cancels the transfer.

```
[user@localhost ~]$ iput -f 2GB 100M
^C
```

(Cancelling can be done with CTRL-C, but note: for testing purposes perhaps timeouts should also be simulated? Pretty sure this also occurs when the connection is abrubtly closed for whatever reasons.)
# The problem

Now the user sees two replicas, two identical checksums and two differing file sizes. We can see that the 2147483648 byte file did not transfer completely, because the user cancelled the transfer.

```
[user@localhost ~]$ ils -L 100M
  user          0 testrepl;res4   2020194304 2016-03-09.11:20 & 100M
[...]

PR #5595, opened 2021-05-04
[#3091] Cancel large iput -f test (master)

alanking:
Adds a test similar to test_cancel_large_iput but for a forced overwrite
of a data object. The test especially ensures that the checksum is
updated and is not equal to the old checksum which would be incorrect
given the partial overwrite of the data.

---

As this only adds a test, I only ran the one test and it passes. The old test which was renamed passes as well.

### DIFF
--- a/scripts/irods/test/test_icommands_file_operations.py
+++ b/scripts/irods/test/test_icommands_file_operations.py
@@ -660,7 +660,7 @@ class Test_ICommands_File_Operations(resource_suite.ResourceBase, unittest.TestC
                         msg="Files missing from vault:\n" + str(local_files - files_in_vault) + "\n\n" +
                             "Extra files in vault:\n" + str(files_in_vault - local_files))
 
-    def test_cancel_large_iput(self):
+    def test_cancel_large_iput_for_create(self):
         base_name = 'test_cancel_large_put'
         local_dir = os.path.join(self.testing_tmp_dir, base_name)
         file_size = pow(2, 30)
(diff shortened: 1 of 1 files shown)


scripts/irods/test/test_icommands_file_operations.py
```diff
@@ -683,6 +683,70 @@ def test_cancel_large_iput(self):
...
+            if not os.path.exists(large_local_file):
+                lib.make_file(large_local_file, file_size)
+
+            iput_cmd = 'iput -f -K {0} {1}'.format(large_local_file, logical_path)
+            file_vault_full_path = os.path.join(self.user0.get_vault_session_path(), filename)
+            self.user0.interrupt_icommand(iput_cmd, file_vault_full_path, 2048)
+            self.user0.assert_icommand('ils -L', 'STDOUT_SINGLELINE', filename)
+
+            # multiple threads could still be writing on the server side, so we need to wait for
+            # the size in the vault to converge - then were done.
```
trel, line 720: we're

scripts/irods/test/test_icommands_file_operations.py
```diff
@@ -683,6 +683,70 @@ def test_cancel_large_iput(self):
...
+            old_size = None
+            new_size = os.path.getsize(file_vault_full_path)
+            while old_size != new_size:
+                time.sleep(1)
+                old_size = new_size
+                new_size = os.path.getsize(file_vault_full_path)
+
+            # wait for select() call to timeout, set to "SELECT_TIMEOUT_FOR_CONN", which is 60 seconds
```
trel, line 728: 'set slightly above'

trel (commented):
i like it a lot.    please tweak comments and #.

scripts/irods/test/test_icommands_file_operations.py
alanking: Done

scripts/irods/test/test_icommands_file_operations.py
alanking: Done

alanking:
Fixed comments, squashed, #'d

Issue #5592, opened 2021-05-04
Corrupted file replication from coordinating replication resource to unixfilesystem resource will not fail and only lead to error in rodsLog


PaulVanSchayck:
- [x] master
- [x] 4-2-stable
---------

### iRODS Version, OS and Version
iRODS: 4.2.8
OS: Ubuntu 18.04 (in Docker)

We first discovered this on our production environment (4.2.6), but now also tested in a minimal Docker environment (4.2.8), with no catalog consumers and a default core.re.

### What did you try to do?
I've got a coordinating replication resource and demoResc

```
$ ilsresc
demoResc:unixfilesystem
replRescUM01:replication
├── UM-hnas-4k:unixfilesystem
└── UM-hnas-4k-repl:unixfilesyste
```

I put a file on `replRescUM01` and checksum it:

```
$ echo "bar" > /tmp/foo
$ iput -R replRescUM01 /tmp/foo
$ ichksum foo
    foo    sha2:fYZelZskZpGMmGOvypQtD7idfJrAyZuvw3SVBN7ZdzA=
$ ils -L foo
  rods              0 replRescUM01;UM-hnas-4k-repl            4 2021-05-04.13:46 & foo
    sha2:fYZelZskZpGMmGOvypQtD7idfJrAyZuvw3SVBN7ZdzA=    generic    /mnt/repl/home/rods/foo
  rods              1 replRescUM01;UM-hnas-4k            4 2021-05-04.13:46 & foo
        generic    /mnt/resc/home/rods/foo
```

I then simulate a corrupted data transfer by modifying the `foo` file on the physical file path of `UM-hnas-4k`. Only modifying one character
```
$ echo "far" > /mnt/repl/home/rods/foo
```
I then replicate to demoResc:
```
$ irepl -R demoResc foo
$ ils -L foo
  rods              0 replRescUM01;UM-hnas-4k-repl            4 2021-05-04.13:46 & foo
    sha2:fYZelZskZpGMmGOvypQtD7idfJrAyZuvw3SVBN7ZdzA=    generic    /mnt/repl/home/rods/foo
  rods              1 replRescUM01;UM-hnas-4k            4 2021-05-04.13:46 & foo
        generic    /mnt/resc/home/rods/foo
  rods              2 demoResc            4 2021-05-04.13:48 & foo
        generic    /var/lib/irods/vault/home/rods/foo
$ cat /var/lib/irods/vault/home/rods/foo
bar
```
[...]

PR #5601, opened 2021-05-05
[#5592] Added test for irepl (4-2-stable)

korydraughn:
Do we want to erase the checksum if replication fails?

Right now, if a replication fails, the replica is marked as stale, but keeps the bad checksum.

### DIFF
--- a/scripts/irods/test/test_irepl.py
+++ b/scripts/irods/test/test_irepl.py
@@ -1,6 +1,7 @@
 from __future__ import print_function
 import os
 import sys
+
 if sys.version_info < (2, 7):
     import unittest2 as unittest
 else:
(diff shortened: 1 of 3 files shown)


scripts/irods/test/test_irepl.py
```diff
@@ -11,9 +12,140 @@
...
+            self.admin.run_icommand(['irm', '-f', data_object])
+
+            for resc_name in ufs_resources:
+                self.admin.run_icommand(['iadmin', 'rmchildfromresc', repl_resc, resc_name])
+                self.admin.run_icommand(['iadmin', 'rmresc', resc_name])
+
+            self.admin.run_icommand(['iadmin', 'rmresc', repl_resc])
+
+    def test_irepl_reports_checksum_mismatch_error_when_bytes_within_the_data_size_are_overwritten__issue_5592(self):
+        # Passing "True" causes the test to trigger a checksum mismatch by overwritting a character
```
trel, line 138: overwritting -> overwriting

scripts/irods/test/test_irepl.py
```diff
@@ -11,9 +12,140 @@
...
+        # within a replica. The overwrite will not cause a mismatch in the size. Only the data within
+        # the replica will change.
+        self.do_test_for_issue_5592(trigger_checksum_mismatch_error=True)
+
+    def test_irepl_does_not_return_a_checksum_mismatch_error_after_appending_data_to_a_replica_out_of_band__issue_5592(self):
+        # Passing "False" causes the test to append a character to a replica. This will not trigger
+        # a checksum mismatch error because the API honor the data size in the catalog when calculating
```
trel, line 145: honor -> honors

trel (commented):
Yes, let's erase known bad checksums.   I don't see any good to them being in there and being known-wrong.

trel:
so now the test is the way we want it - but the code still leaves the checksum there?  so we'll await a code change too?

korydraughn:
I've updated the server to store the incorrect checksum so that users/admins can see that there is a difference. This is how the checksums are handled by the other APIs.

scripts/irods/test/test_irepl.py
```diff
@@ -11,9 +12,140 @@
...
+                self.admin.assert_icommand(['irepl', '-R', 'demoResc', data_object], 'STDERR', ['status = -314000 USER_CHKSUM_MISMATCH'])
+
+                # Show that the replicas under the replication resource are still marked as "good"
+                # and that the newest replica is marked as "stale" due to the error.
+                self.assertEqual(get_replica_status(data_object, '0'), '1')
+                self.assertEqual(get_replica_status(data_object, '1'), '1')
+                self.assertEqual(get_replica_status(data_object, '2'), '1')
+                self.assertEqual(get_replica_status(data_object, '3'), '0')
+
+                # Show that the newest replica does has a different checksum.
```
trel, line 102: does has -> does have
or
remove does

trel:
and then # it up.

scripts/irods/test/test_irepl.py
```diff
@@ -11,9 +12,140 @@
...
+            self.admin.assert_icommand(['istream', 'write', '-R', repl_resc, data_object], input='bar')
+            self.admin.assert_icommand(['ils', '-l', data_object], 'STDOUT', [
+                ' {};{}            3 '.format(repl_resc, ufs_resources[0]),
+                ' {};{}            3 '.format(repl_resc, ufs_resources[1]),
+                ' {};{}            3 '.format(repl_resc, ufs_resources[2])
+            ])
+
+            # Checksum the first replica.
+            # TODO Should ALL of the replicas be checksum'd?
+            self.admin.assert_icommand(['ichksum', '-n', '0', data_object], 'STDOUT', ['sha2:'])
```
korydraughn, line 79: @trel Any thoughts on calculating checksums for one or all replicas?

Also, I'm going to remove the `-n0` on the `ichksum` call because that will bring the test closer to what was described in the issue. The downside is that we are relying on the server to vote and pick the replica with the checksum as the replication source. What if the voting picks a replica without a checksum.

Any thoughts on that?

trel: We want to test what happens when there is a checksum mismatch on the replication... so i think the `-a` makes sense here.   I don't know that testing the case of replication without a checksum tells us anything.

korydraughn: Cool. Will add the -a

Issue #5220, opened 2020-10-30
iRODS buildsystem does not make proper use of CMAKE_INSTALL_PREFIX or GNUInstallDirs


SwooshyCueb:
For the this repository and all plugins and C++ clients:
- [x] master
- [x] 4-2-stable
----

This is most evident when using `make install` with `DESTDIR` set, when `cmake` was run with `CMAKE_INSTALL_PREFIX` set to `/usr/local`. You wind up with directories like `$DESTDIR/usr/local/usr/lib`

trel:
This will need to be balanced with our desire/requirement to provide a non-package-install option for HPC use cases.

SwooshyCueb:
*My* desire for a proper non-package install is actually part of my motivation for self-assigning this.

trel:
I've merged all I could find for master.

SwooshyCueb:
You've got them all.

I wouldn't rush to mark of master as complete just yet, as I've only submitted PRs for repos that were either
- updated in the past year or so, targeted iRODS 4.3.0 on the master branch, AND buildable in the development environment docker containers without too much fuss;
OR
- mentioned in that list you gave me

trel:
fair enough, i retract the checkmark and leave it to you.

SwooshyCueb:
The next round of PRs is ready for review. All tests are passing in Jenkins.

After poking, prodding, and careful consideration, I have decided that the globus connector's buildsystem differs enough from our own to exclude it from this round of PRs.

trel:
got it - have noted on each.     and made sure codacy is on for all those branches (had missed 4-5).

agreed on the globus connector - it's early days there.

trel:
all merged.     checkbox baton is yours.

PR #5606, opened 2021-05-05
[#5220,#5250] Fix bad assumptions about default paths; cmake touch-ups for icommands userspace packaging (4-2-stable)

SwooshyCueb:
Backport of #5573 into stable.

In service of #5220 and #5250.


SwooshyCueb:
\#'d

@trel

PR #5612, opened 2021-05-07
Quick fixes (4-2-stable)

korydraughn:


server/re/src/keyValPairMS.cpp
```diff
@@ -1,35 +1,24 @@
...
-//#include "reFuncDefs.hpp"
 #include "icatHighLevelRoutines.hpp"
+#include "msParam.h"
 #include "rcMisc.h"
-
 #include "irods_re_structs.hpp"
+#include "rodsErrorTable.h"
 
 /**
- * \fn msiGetValByKey(msParam_t* inKVPair, msParam_t* inKey, msParam_t* outVal, ruleExecInfo_t *rei)
```
trel, line 16: i think doxygen still wants this line?

trel (commented):
just the \fn, and then # if tests pass

server/re/src/keyValPairMS.cpp
korydraughn: it doesn't need those lines when the documentation is next to the function. assuming the function name and parameters match.

trel: oh?   fine if true.

korydraughn:
Please review `rsDataObjRead` and `rsDataObjLseek`'s handling of the file read position.

server/re/src/keyValPairMS.cpp
korydraughn: i'll verify if this is needed before the end of this weekend.

server/api/src/rsDataObjLseek.cpp
```diff
@@ -10,112 +7,102 @@
...
-        memset( &subStructFileLseekInp, 0, sizeof( subStructFileLseekInp ) );
+    // If the replica was opened in read-only mode, then don't allow the client
+    // to seek past the data size in the catalog. This is necessary because the
+    // size of the replica in storage could be larger than the data size in the
+    // catalog.
+    //
+    // For all other modes, let the seek operation do what it normally does.
+    const auto offset = (O_RDONLY == (l1desc.dataObjInp->openFlags & O_ACCMODE))
+        ? std::min(dataObjLseekInp->offset, dataObjInfo->dataSize)
+        : dataObjLseekInp->offset;
```
trel, line 65: yes, this looks good

server/api/src/rsDataObjRead.cpp
```diff
@@ -55,40 +61,64 @@ applyRuleForPostProcForRead( rsComm_t *rsComm, bytesBuf_t *dataObjReadOutBBuf, c
...
+        // If the file read position is greater than or equal to the data size,
+        // then return immediately.
+        if (output->offset >= l1desc.dataObjInfo->dataSize) {
+            return 0;
+        }
+
+        // At this point, we know the file read position is less than the replica's
+        // recorded data size. We now have to adjust the requested number of bytes to
+        // read so that the client does not read past the recorded data size.
+        dataObjReadInp->len = std::min<int>(dataObjReadInp->len, l1desc.dataObjInfo->dataSize - output->offset);
```
trel, line 112: the remainder math looks good

server/api/src/rsDataObjRead.cpp
```diff
@@ -55,40 +61,64 @@ applyRuleForPostProcForRead( rsComm_t *rsComm, bytesBuf_t *dataObjReadOutBBuf, c
...
+
+        if (const auto ec = rsDataObjLseek(rsComm, &input, &output); ec < 0) {
+            rodsLog(LOG_ERROR, "%s: Could not retrieve the current file read position.", __func__, ec);
+            return ec;
         }
+
+        // If the file read position is greater than or equal to the data size,
+        // then return immediately.
+        if (output->offset >= l1desc.dataObjInfo->dataSize) {
+            return 0;
```
trel, line 106: is this correct?    not an error?    it's someone else's job to interpret this?

korydraughn: This is the system telling the client that no bytes were read because they have reached the end of the file.

server/api/src/rsDataObjLseek.cpp
korydraughn: I may need to define the integer type for std::min here.

server/api/src/rsDataObjRead.cpp
trel: or it's just saying it literally read zero bytes.

server/api/src/rsDataObjRead.cpp
```diff
@@ -55,40 +61,64 @@ applyRuleForPostProcForRead( rsComm_t *rsComm, bytesBuf_t *dataObjReadOutBBuf, c
...
-        int i;
-        bytesRead = l3Read( rsComm, l1descInx, dataObjReadInp->len,
-                            dataObjReadOutBBuf );
-        i = applyRuleForPostProcForRead( rsComm, dataObjReadOutBBuf,
-                                         L1desc[l1descInx].dataObjInfo->objPath );
-        if ( i < 0 ) {
-            return i;
+
+    // Do not let the client read past the data size in the catalog if the replica
+    // was opened in read-only mode.
```
alanking, line 90: Might be useful to have the explanatory "This is necessary because..." comment here as well

alanking (commented):
Approach seems good to me

server/api/src/rsDataObjRead.cpp
```diff
@@ -55,40 +61,64 @@ applyRuleForPostProcForRead( rsComm_t *rsComm, bytesBuf_t *dataObjReadOutBBuf, c
...
-        rodsLog( LOG_NOTICE,
-                 "rsDataObjRead: l1descInx %d out of range",
-                 l1descInx );
+rsDataObjRead(rsComm_t* rsComm,
+              openedDataObjInp_t* dataObjReadInp,
+              bytesBuf_t* dataObjReadOutBBuf)
+{
+    const int l1descInx = dataObjReadInp->l1descInx;
+
+    if (l1descInx < 2 || l1descInx >= NUM_L1_DESC) {
```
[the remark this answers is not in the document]
korydraughn: Yes. Will change.

server/api/src/rsDataObjRead.cpp
korydraughn: I'm not sure what else to say here. Is there something you feel needs to be added that could help us understand this six months from now?

alanking: I was just noticing that the text is very similar to that found in the lseek code above, so it might be useful here too:

    // If the replica was opened in read-only mode, then don't allow the client
    // to seek past the data size in the catalog. This is necessary because the
    // size of the replica in storage could be larger than the data size in the
    // catalog.

korydraughn: Gotcha. Will add more comments like these.

server/api/src/rsDataObjRead.cpp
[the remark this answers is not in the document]
korydraughn: done

server/api/src/rsDataObjLseek.cpp
```diff
@@ -10,112 +7,102 @@
...
+
+    // If the replica was opened in read-only mode, then don't allow the client
+    // to seek past the data size in the catalog. This is necessary because the
+    // size of the replica in storage could be larger than the data size in the
+    // catalog.
+    //
+    // For all other modes, let the seek operation do what it normally does.
+    const auto offset = (O_RDONLY == (l1desc.dataObjInp->openFlags & O_ACCMODE))
+        ? std::min(dataObjLseekInp->offset, dataObjInfo->dataSize)
+        : dataObjLseekInp->offset;
```
korydraughn, line 76: Moved the offset adjustment below the special collections code. CI tests involving special collections were failing.

It is probably best to ignore special collections for this anyway. Thoughts?

server/api/src/rsDataObjRead.cpp
```diff
@@ -55,40 +61,70 @@ applyRuleForPostProcForRead( rsComm_t *rsComm, bytesBuf_t *dataObjReadOutBBuf, c
...
+
+    auto* dataObjInfo = l1desc.dataObjInfo;
+
+    // If the replica was opened in read-only mode, then don't allow the client
+    // to read past the data size in the catalog. This is necessary because the
+    // size of the replica in storage could be larger than the data size in the
+    // catalog.
+    //
+    // For all other modes, let the read operation do what it normally does.
+    if (!dataObjInfo->specColl && O_RDONLY == (l1desc.dataObjInp->openFlags & O_ACCMODE)) {
```
korydraughn, line 97: Ignore this check if the special collections pointer is valid. This change was due to failing CI tests.

Does this seem reasonable?

korydraughn: Also, is using `getStructFileType(dataObjInfo->specColl) >= 0` better than just checking the pointer for special collections?

What is the difference?

korydraughn:
Passed CI.

Waiting on response to comments.

server/api/src/rsDataObjLseek.cpp
trel: you're only adding a new safety check - certainly not changing the current behavior for them until we swing around again and tackle them all head on seems fine.  @alanking ?

alanking: I have found ignoring special collections and/or pushing them to the fringe cases is a good policy, so seems fine to me.

server/api/src/rsDataObjRead.cpp
alanking: The struct file check is kind of a subset of special collections. It's a type of special collection which represents a mounted struct file (e.g. tar). I've struggled with the difference myself because it seems that sometimes the struct files do things like regular data objects but at other times they don't, so I'm not 100% sure when it is appropriate to use one over the other.

TL;DR: if you just want to avoid special collections here, checking the `specColl` pointer should be sufficient

korydraughn: Ok. Marking as resolved.

server/api/src/rsDataObjLseek.cpp
korydraughn: Ok. Marking as resolved.

trel:
got em - already merged those two.

korydraughn:
passed CI

added #

PR #5620, opened 2021-05-10
Fixed nanodbc dependency in build hook (4-2-stable)

korydraughn:
This is required so that iRODS can be built in CI


korydraughn:
added #

Issue #5605, opened 2021-05-05
irods_rc_data_obj test intermittently fails due to a race condition


SwooshyCueb:
Observed in master, not tested in stable yet.

While testing 5573 in Jenkins, I encountered test failures in `irods_rc_data_obj` that did not occur at the bench.

```
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
  <testsuite name="irods_rc_data_obj" errors="0" failures="4" tests="108" hostname="tbd" time="48.7649" timestamp="2021-05-05T13:18:41Z">
    <testcase classname="irods_rc_data_obj.global" name="open,read,write,close" time="4.791027"/>
    <testcase classname="irods_rc_data_obj.global" name="open,read,write,close/create,no close" time="2.587055">
      <failure message="0 == 2" type="CHECK">
at /irods_git_repo/unit_tests/src/test_rc_data_obj.cpp:97
      </failure>
    </testcase>
    <testcase classname="irods_rc_data_obj.global" name="open,read,write,close/open for read,close" time="2.757982"/>
    <testcase classname="irods_rc_data_obj.global" name="open,read,write,close/open for write,close" time="3.585973"/>
    <testcase classname="irods_rc_data_obj.global" name="open,read,write,close/open for read,no close" time="3.066706"/>
    <testcase classname="irods_rc_data_obj.global" name="open,read,write,close/open for write,no close" time="3.087368">
      <failure message="0 == 2" type="REQUIRE">
at /irods_git_repo/unit_tests/src/test_rc_data_obj.cpp:207
      </failure>
    </testcase>
    <testcase classname="irods_rc_data_obj.global" name="open,read,write,close/create,no close,open for write,close" time="3.29922"/>
    <testcase classname="irods_rc_data_obj.global" name="open,read,write,close/create,write,close" time="2.543977"/>
    <testcase classname="irods_rc_data_obj.global" name="open,read,write,close/create,write,no close" time="2.792813">
      <failure message="9 == 0" type="CHECK">
at /irods_git_repo/unit_tests/src/test_rc_data_obj.cpp:349
      </failure>
      <failure message="0 == 2" type="CHECK">
at /irods_git_repo/unit_tests/src/test_rc_data_obj.cpp:350
      </failure>
    </testcase>
    <testcase classname="irods_rc_data_obj.global" name="rxDataObjChksum" time="2.355791"/>
[...]

PR #5624, opened 2021-05-11
[#5605] Wait for agent close before asserting (master)

alanking:
Adds a unit test utility to get the PID of the connected agent. This
allows us to wait for the remote agent to go away after suddenly
disconnecting before making any assertions about what should and should
not be in the catalog.

---

As only a unit test was touched for this change, I only ran the affected unit tests and they pass.


alanking:
#'d. Passed CI several times

Issue #5426, opened 2021-02-19
ibun atop existing archive file into replication resource leaves zero-length replicas


trel:
- [x] master
- [x] 4-2-stable
----------

### iRODS Version, OS and Version

4.2.8

### What did you try to do?

ibun a 'large' collection into a replication resource where the target archive file already exists

### Expected behavior

two new correct zip (or other archive type) files in the right place

### Observed behavior (including steps to reproduce, if applicable)

zero-length truncation of the replicated file(s)

```
$ ils -rl thedir ; echo "------"; for SUFFIX in tar zip gzip bzip2 ; do for X in 1 2 ; do echo "--- ${SUFFIX} --- ${X} ---"; ibun -cD${SUFFIX} -f /tempZone/home/rods/thedir.${SUFFIX} /tempZone/home/rods/thedir ; ils -l thedir.${SUFFIX} ; done ; done
/tempZone/home/rods/thedir:
  rods              0 hydro;replresc;pt2;ufs2   2147483648 2021-02-19.17:46 & 2gfile
  rods              1 hydro;replresc;pt1;ufs1   2147483648 2021-02-19.17:47 & 2gfile
  rods              0 hydro;replresc;pt2;ufs2     36700160 2021-02-19.17:11 & 35mbfile
  rods              1 hydro;replresc;pt1;ufs1     36700160 2021-02-19.17:11 & 35mbfile
  rods              0 hydro;replresc;pt1;ufs1          224 2021-02-18.21:16 & foo
  rods              1 hydro;replresc;pt2;ufs2          224 2021-02-19.18:02 & foo
  C- /tempZone/home/rods/thedir/subcoll  
/tempZone/home/rods/thedir/subcoll:
  rods              0 hydro;replresc;pt1;ufs1          166 2021-02-18.21:16 & bar
  rods              1 hydro;replresc;pt2;ufs2          166 2021-02-19.18:02 & bar
------
--- tar --- 1 ---
  rods              0 hydro;replresc;pt2;ufs2   2184192000 2021-02-19.18:10 & thedir.tar
  rods              1 hydro;replresc;pt1;ufs1   2184192000 2021-02-19.18:11 & thedir.tar
--- tar --- 2 ---
  rods              0 hydro;replresc;pt2;ufs2            0 2021-02-19.18:12 & thedir.tar
  rods              1 hydro;replresc;pt1;ufs1   2184192000 2021-02-19.18:13 & thedir.tar
--- zip --- 1 ---
[...]

PR #5629, opened 2021-05-11
[#5426] Fix force flag usage in rsStructFileBundle (master)

alanking:
New test passes, will run through CI, too

### DIFF
--- a/scripts/irods/test/test_ibun.py
+++ b/scripts/irods/test/test_ibun.py
@@ -11,10 +11,9 @@ import tempfile
 
 from . import resource_suite
 from .. import lib
+from .. import test
 
-@unittest.skip('Generation of large file causes I/O thrashing... skip for now')
 class Test_Ibun(resource_suite.ResourceBase, unittest.TestCase):
-
     def setUp(self):
         super(Test_Ibun, self).setUp()
         self.known_file_name = 'known_file'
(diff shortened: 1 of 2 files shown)


scripts/irods/test/test_ibun.py
```diff
@@ -100,3 +101,80 @@ def test_ibun_extraction_of_big_tar_file__issue_4118(self):
...
+                self.admin.run_icommand(['irm', '-r', '-f', collection_path])
+                self.admin.run_icommand(['irm', '-f', archive_path])
+                lib.remove_child_resource(replication, resource_0, self.admin)
+                lib.remove_child_resource(replication, resource_1, self.admin)
+                lib.remove_resource(replication, self.admin)
+                lib.remove_resource(resource_0, self.admin)
+                lib.remove_resource(resource_1, self.admin)
+
+        do_test_ibun_atop_existing_archive_file_in_replication_hierarchy(self, 1039425, 1044480)
+        do_test_ibun_atop_existing_archive_file_in_replication_hierarchy(self, 1039426, 1054720)
```
trel, line 180: *sometimes* this one byte size delta is not enough... so might want to make it 5 bytes or so between 'under' and 'over' scenarios.

alanking: Okay. Should I bump back the smaller size as well to ensure that the archive file stays under 1 MiB? I'm not quite sure what the threshold is. Does it depend on the libarchive version and/or underlying machinery in the OS?

From https://github.com/irods/irods/issues/5426#issuecomment-782907963:

> The bug is triggered as soon as the archive file is larger than 1 mebibyte (1024*1024 = 1048576 bytes).

alanking: I bumped by 1KB each in either direction and test passes for me

PR #5637, opened 2021-05-13
[#5220] Fix makeinstall-time schema updating (master)

SwooshyCueb:
This fixes the disappearing `/var` issue for master.

Related to #5220


SwooshyCueb:
\#'d

Issue #4720, opened 2020-02-11
Test failure (master only): test_all_rules.Test_AllRules.test_rulemsiServerMonPerf_r


alanking:
- [ ] master
---
```
 --- IrodsSession: icommand executed by [rods] [irule -vF /var/lib/irods/clients/icommands/test/rules/rulemsiServerMonPerf.r] --- 
Assert Command: irule -vF /var/lib/irods/clients/icommands/test/rules/rulemsiServerMonPerf.r
Expecting STDOUT_SINGLELINE: ['completed successfully']
  stdout:
    | rcExecMyRule: acServerMonPerf {
    | #This microservice invokes a command in server/bin/cmd
    | #    irodsServerMonPerf     - a perl script to get monitoring information
    |       msiServerMonPerf("default","default");
    |       msiServerMonPerf("verbose","default");
    | }   
    |     
    | outParamDesc: ruleExecOut
  stderr:
    | remote addresses: 172.17.0.10 ERROR: rcExecMyRule error.  status = -154000 SYS_INTERNAL_ERR
Unexpected output on stderr

FAILED TESTING ASSERTION
```

<details>
<summary>
A stacktrace in the log... (click to expand)
</summary>

```json
{
  "log_category": "agent",
  "log_facility": "local0",
  "log_level": "info",
  "log_message": "I'm an agent!",
  "server_host": "icat.example.org",
  "server_pid": 2200,
  "server_timestamp": "2020-02-11T14:48:45.645930",
  "server_type": "agent"
}
{
  "log_category": "legacy",
  "log_facility": "local0",
  "log_level": "info",
  "log_message": "execCmd:../../var/lib/irods/msiExecCmd_bin/irodsServerMonPerf argv: -t 10 -fs /tmp/irods/AnotherResc,/tmp/irods/testallrulesResc,/tmp/irods/TestResc,/var/lib/irods/Vault",
  "request_api_name": "EXEC_MY_RULE_AN",
  "request_api_number": 625,
  "request_api_version": "d",
  "request_client_user": "rods",
  "request_host": "172.17.0.10",
  "request_proxy_user": "rods",
  "request_release_version": "rods4.3.0",
  "server_host": "icat.example.org",
  "server_pid": 2203,
  "server_timestamp": "2020-02-11T14:48:45.126680",
  "server_type": "agent"
}
{
  "log_category": "legacy",
  "log_facility": "local0",
  "log_level": "error",
[...]

PR #5642, opened 2021-05-14
[#4331] Add ticket support to data_object_finalize (master)

alanking:
CI tests are running

### DIFF
--- a/server/core/src/logical_locking.cpp
+++ b/server/core/src/logical_locking.cpp
@@ -3,6 +3,7 @@
 #include "irods_log.hpp"
 #include "replica_access_table.hpp"
 #include "replica_state_table.hpp"
+#include "scoped_privileged_client.hpp"
 
 namespace
 {
@@ -19,7 +20,7 @@ namespace
         {ill::lock_type::write, WRITE_LOCKED}
     };
 
-    auto get_original_replica_status_impl(const json& _json_replica) -> repl_status_t
+    auto get_original_replica_status_impl(const json& _json_replica) -> int
     {
         const auto data_status_str = _json_replica.at("after").at("data_status").get<std::string>();
 
@@ -317,7 +318,7 @@ namespace irods::logical_locking
 {
     auto get_original_replica_status(
         const std::uint64_t _data_id,
-        const int           _replica_number) -> repl_status_t
+        const int           _replica_number) -> int
     {
         return get_original_replica_status_impl(rst::at(_data_id, _replica_number));
     } // get_original_replica_status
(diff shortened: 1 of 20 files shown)


alanking:
Codacy issue is invalid. `ec` is set to 0 before the lambda provided to `irods::at_scope_exit` captures it by reference. `ec` is later set to be whatever `rs_data_object_finalize` returns, which may not be 0.

server/core/src/logical_locking.cpp
```diff
@@ -367,7 +368,20 @@ namespace irods::logical_locking
             return ec;
         }
 
-        if (const int ec = rst::publish::to_catalog(_comm, _ctx); ec < 0) {
+        // TODO: why have the retry? What are the implications of elevating the privileges the first time?
```
trel, line 371: yes, what case does retrying the database failure exercise... a transient network error?  a locked database during the first attempt?

trel:
otherwise, looks good - please # when satisfied.

server/core/src/logical_locking.cpp
alanking: Let's see... there's only one case I care about when it comes to unlocking with elevated privileges and that's the case where the database update failed for some non-database-related reason - be it permissions, tickets, or some other restriction imposed by iRODS. In this case, I just want to make sure that the object is unlocked.

Perhaps it would be better in the case of failure to just return the error to the caller and let them decide what to do. This is server-side only anyway, so the caller may decide to stale the offending replica or something like that (i.e. call `stale_target_replica_and_publish`).

I'm leaning toward removing the retry.

trel: seems reasonable.  we can't know the intent of the caller.

alanking:
CI tests passed. I removed the retry. Going to investigate one more thing, maybe run through CI again for sanity. Then cherry-pick to 4-2-stable and do it again.

server/core/src/logical_locking.cpp
alanking: This is more of a convenience function than anything, and is also internal-only so I will assume the user knows what they're doing with it and what they should do in failure. Removed the retry, thanks.

alanking:
CI tests passed with retry removal, except for 2. One is captured in #4720 (intermittent) and the other is `test_imeta_interactive` again, which seems to fail intermittently as well. Both of these pass at the bench.

alanking:
Added test for write byte count updating when the size of the data object does not change. The test passes locally

trel:
test looks good, please squash and # em.

alanking:
#'d

Issue #5017, opened 2020-07-17
Use of microservices in dynamic PEPs breaks large file transfers across federation


alanking:
- [x] master
- [x] 4-2-stable
---

### iRODS Version, OS and Version
4.2.8

### What did you try to do?
Put a large file across federation while connected server is configured with a rulebase which includes `pep_api_data_obj_put_post` and calls any microservice.

### Expected behavior
Things just work.

### Observed behavior (including steps to reproduce, if applicable)
Include the following stanza in a configured rulebase on the catalog provider for the home zone:
```
pep_api_data_obj_put_post(*INSTANCE_NAME, *COMM, *DATAOBJINP, *BUFFER, *PORTAL_OPR_OUT) {
    msiObjStat("/remoteZoneHint/path/to/whatever", *out);
}
```
Put a large file across federation to the remote zone, connected and authenticated in the  home zone:
```
$ ienv | grep zone
irods_zone_name - otherZone
$ iput big /tempZone/home/rods#otherZone/foo1
remote addresses: 172.22.0.3 ERROR: connectToRhostPortal: connectTo Rhost icat.tempZone.example.org port 20080 error, status = -305111
remote addresses: 172.22.0.3 ERROR: putUtil: put error for /tempZone/home/rods#otherZone/foo1, status = -305111 status = -305111 USER_SOCK_CONNECT_ERR, Connection refused
```

If the microservice is called from a different dynamic or static PEP, things work as expected. This has been demonstrated on 4.2.7 as well.

Please investigate - this is likely a parsing/memory layout issue in the NREP.

alanking:
After some investigation, we have a better handle on the problem now: The dynamic PEP in the rulebase for the home zone catalog service provider fires before returning to the client, and then executes the microservice.

If this happens to be a data object or catalog operation in the remote zone, the same agent will reach out to the remote catalog service provider using a new connection on the zone port. The agent process standing up the parallel transfer portal in the remote zone CSP times out when `select`ing on the socket (because the portal information has not been returned to the client yet), fails, and services the request for the post-PEP. The home zone CSP then returns to the client with the portal information for the parallel transfer, the client arrives to connect to the portal, and the remote server refuses the connection because by this point the portal has gone away.

Still investigating a fix which can allow the home zone CSP in this scenario to continue issuing requests to the remote zone CSP so that the parallel transfer can complete in this scenario

[...]

PR #5645, opened 2021-05-17
Various fixes for federation (master)

alanking:
Fixes several issues concerning federation (mostly parallel transfer related) which have occurred over time in master and 4-2-stable:
  - `icp` does not need to resolve the resource hierarchy for the destination replica as this can be done in `rsDataObjOpen`. This logic was here mainly to uphold the `OVERWRITE_WITHOUT_FORCE_FLAG` rules for `icp`, but this is being handled earlier in the function and the deleted logic was only checking for existing replicas anyway.
  - The `dataSize` parameter and `bytesWritten` are very touchy in `rsDataObjCopy` so I've put back a couple of these to more closely resemble how they were being set in the past. This mainly involves using the size of the data object for the source data object as the "expected" dataSize for the destination data object as well as the input dataSize for the destination data object and the bytesWritten for the `transStat`.
  - Prevent updating collection mtime in open and put when the path is not in the local zone because of the issue described in #5017 and also because it is unnecessary - the remote zone will take care of updating the collection mtime if it cares about that because it is part of the logic for these API endpoints.
  - You'll find the new test specifically for the linked issue at the bottom of `test_federation.py`. It uses an existing rule for dynamic PEPs which is exactly what I needed for the federation test, so that's why it is referring to the other issue number in the name of the rule text.

Federation test suite passed at the bench, but running through CI as well as the usual suite. CI tests are running.

### DIFF
--- a/scripts/irods/test/test_federation.py
+++ b/scripts/irods/test/test_federation.py
@@ -6,18 +6,20 @@ if sys.version_info >= (2, 7):
 else:
     import unittest2 as unittest
 import os
-import tempfile
-import time
 import shutil
 import socket
+import tempfile
+import time
 import ustrings
 
+from . import session
+from . import settings
+from .. import lib
 from .. import paths
 from .. import test
-from . import settings
-from . import session
-from .. import lib
 from ..configuration import IrodsConfig
+from ..core_file import temporary_core_file
+from .rule_texts_for_tests import rule_texts
 
 SessionsMixin = session.make_sessions_mixin(
     test.settings.FEDERATION.RODSADMIN_NAME_PASSWORD_LIST, test.settings.FEDERATION.RODSUSER_NAME_PASSWORD_LIST)
@@ -510,7 +512,6 @@ class Test_ICommands(SessionsMixin, unittest.TestCase):
         filename = 'icp_test_file'
         filesize = self.config['test_file_size']
         filepath = os.path.join(self.local_test_dir_path, filename)
-        lib.make_file(filepath, filesize)
 
         # test specific parameters
         parameters = self.config.copy()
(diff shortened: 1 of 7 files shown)


alanking:
Also, I didn't prevent the collection mtime updates in copy or rename because I figured these would be handled by `rsDataObjOpen`. I can add the checks for copy and rename if anybody feels strongly about it.

scripts/irods/test/test_federation.py
```diff
@@ -1442,3 +1471,90 @@ def test_icp_tree_remotezone_in_home(self):
...
+                        lambda: lib.metadata_attr_with_value_exists(self.admin, attr, pep),
+                        interval=1,
+                        maxrep=10
+                    )
+
+                self.assertFalse(lib.metadata_attr_with_value_exists(self.admin, attr, 'pep-obj-put-except'))
+
+                reset_resource(self)
+
+                # put to same logical path, targeting a non-existent replica (resulting in error)
```
alanking, line 1544: Change to: put to same logical path without force flag, resulting in an error and (hopefully) triggering the except pep

trel:
i think cp is definitely fine b/c open... and a rename can't... rename across a zone boundary (that's called a copy)... so yes, i think we're okay for now.

scripts/irods/test/test_federation.py
```diff
@@ -1442,3 +1471,87 @@ def test_icp_tree_remotezone_in_home(self):
...
+        logical_path = os.path.join(remote_home_collection, filename)
+        metadata_resource = 'test_peps_for_parallel_mode_transfers_resource'
+        file_size = 40 * 1024 * 1024 # 40MB
+        attr = 'test_peps_for_parallel_mode_transfers__4404_put'
+
+        try:
+            if not os.path.exists(local_file):
+                lib.make_file(local_file, file_size)
+
+            # PEPs will fire locally on connected server, so metadata will applied to local resource
```
trel, line 1510: be applied

trel (commented):
very nice.

scripts/irods/test/test_federation.py
alanking: Done

alanking:
CI tests passed except for federation, but this was a failure in the test machinery to set up the remote zone. I'll have to debug Jenkins some other time - the suite passed at the bench in a docker-compose'd federation between two zones. Added #

trel:
awaiting green, will merge.

trel:
(!)

PR #5650, opened 2021-05-18
Implemented new library: scoped_permission (4-2-stable)

korydraughn:
Allows operations to temporarily grant permissions on a collection or data
object for the duration of a scoped block.

### DIFF
--- a/lib/filesystem/include/filesystem/filesystem.hpp
+++ b/lib/filesystem/include/filesystem/filesystem.hpp
@@ -44,6 +44,9 @@ namespace irods::experimental::filesystem
         std::string units;
     };
 
+    /// A tag type used to instruct an operation to run in administrator mode.
+    inline struct admin_tag {} admin;
+
     namespace NAMESPACE_IMPL
     {
         // Operational functions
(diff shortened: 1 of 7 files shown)


korydraughn:
This could be renamed to `scoped_access` and it could take a third argument that specifies what to set the permissions to.

trel:
does changing it for a data object also need to change it for the parent collection?

trel:
scoped_access seems more flexible, but there is more error checking to be done... but we could easily compare against known values... right?

korydraughn:
I don't think permissions need to be adjusted on the parent collection because `rodsadmin`s can see inside any collection. The always elevates privileges to `rodsadmin` status before adjusting the permissions so I think collections are handled.

As for the more flexible `scoped_access` option, the fsys library only accepts enumerations when setting permissions, so it is hard to misuse the library. Does that answer your question about error checking?

korydraughn:
Now that I think about it, the parent collection may need to be adjusted as well because the `rodsadmin` status is not maintained for the duration of the scoped block.

trel:
not convinced `OWN` is a good default...

no default would be better?  force caller to understand what they want?

korydraughn:
Sounds good. Removing default

korydraughn:
Maybe, parent collections of data objects shouldn't be touched. Instead, the dev is responsible for using this class on the parent collection as well, for example:
```c++
RsComm comm = // Communication object representing iRODS user "alice".
fs::path data_object = "/tempZone/home/rods/foo.txt";

{
    scoped_permission sp1{comm, data_object.parent_path(), fs::perms::read);
    scoped_permission sp2{comm, data_object, fs::perms::write);

    // Now add/remove metadata on the data object owned by "rods".
}
```

server/core/include/scoped_permission.hpp
```diff
@@ -0,0 +1,91 @@
...
+                    entity.prms != fs::perms::inherit &&
+                    entity.prms != fs::perms::noinherit)
+                {
+                    old_perm_ = entity.prms;
+                    break;
+                }
+            }
+
+            // Set the client's requested permissions on the collection or data object.
+            fs::server::permissions(comm_, path_, comm_.clientUser.userName, _perm);
```
korydraughn, line 58: this line will throw because the fsys library doesn't support the `ADMIN_KW`.

the fix for this is to either use the C API directly or provide a way to set the `ADMIN_KW` through the fsys library.

korydraughn:
Are there any other thoughts on this PR @trel @alanking @SwooshyCueb ?

alanking:
More explanation as to the motivation for these changes would be good, but I don't see any obvious issues

korydraughn:
This library captures a scenario encountered while implementing the hard links REP. I figured there are likely other places where this facility may be useful.

It also includes changes to the filesystem library in regards to collection inheritance.

lib/filesystem/include/filesystem/filesystem.hpp
```diff
@@ -142,6 +145,47 @@ namespace irods::experimental::filesystem
...
+        ///                          reference a collection.
+        ///
+        /// \param[in] _admin A tag which instructs the function to operate in administrator mode. The client
+        ///                   must be an administrator to use this function.
+        /// \param[in] _comm  The communication object.
+        /// \param[in] _p     The path to a collection.
+        /// \param[in] _value A boolean indicating whether inheritance should be enabled or not.
+        ///
+        /// \since 4.2.11
+        auto enable_inheritance(admin_tag _admin, rxComm& _comm, const path& _p, bool _value) -> void;
```
trel, line 187: was it intentional to put the admin flag first?    rather than last?   then the two signatures (admin and non-admin) would be more similar?

first is definitely more 'visible'.

trel:
and... the current implementation does *not* touch the parent collection, right?

lib/filesystem/include/filesystem/filesystem.hpp
korydraughn: Yes, it was intentional. The admin flag being first makes it so that no one can miss it. That isn't the only reason, but it is a big one.

korydraughn:
It does not. It only touches the path you give it.

korydraughn:
Devs are still free to implement more convenient things on top of this if they, but we don't make any assumptions about what they are trying to do. They still need to understand iRODS permissions and things.

lib/filesystem/include/filesystem/object_status.hpp
```diff
@@ -48,17 +49,20 @@ namespace irods::experimental::filesystem
 
         auto type() const noexcept -> object_type { return type_; }
         auto permissions() const noexcept -> const std::vector<entity_permission>& { return perms_; }
+        auto is_inheritance_enabled() const noexcept -> bool { return inheritance_; }
```
korydraughn, line 52: @trel @alanking Right now, the fsys library treats the inheritance flag as a boolean. How do we feel about that?

Would it be better to store the value from the catalog and leave the interpretation of it on the developer? We could provide utility functions for translation as well.

This is just me thinking about alternatives to what is presented.

lib/filesystem/include/filesystem/filesystem.hpp
trel: sold.

lib/filesystem/include/filesystem/object_status.hpp
trel: well, inheritance *is* a boolean, as far as I know... a collection has it turned on or not.

korydraughn: Cool. I think the implementation is as it should be then.

korydraughn:
added #

PR #5655, opened 2021-05-19
[5654] change default maximum threads to "default" (4-2-stable)

trel:
shown to be enough to saturate a connection

defined in server_config.json as advanced_settings.default_number_of_transfer_threads


trel:
CI running

trel:
passed unit, core, timing, federation...  working on topology provider and consumer

PR #5665, opened 2021-05-21
[_XXXX] add DB plugin op for data_object_finalize (master)

alanking:
This adds a database plugin operation for use in the
data_object_finalize API plugin. This is meant to act as interchangeable
logic with the nanodbc-based database interactions which exists in the
API plugin today. This change is needed because of some possible bugs in
the implementation and/or in the nanodbc library itself. For more
details on the issue, see: <insert URL here for nanodbc issue 267>

---

Have not tested for master yet

### DIFF
--- a/plugins/database/src/db_plugin.cpp
+++ b/plugins/database/src/db_plugin.cpp
@@ -51,6 +51,10 @@
 #include <vector>
 #include <boost/regex.hpp>
 #include <boost/lexical_cast.hpp>
+#include <chrono>
+
+#include "json.hpp"
+#include "fmt/format.h"
 
 #include "irods_lexical_cast.hpp"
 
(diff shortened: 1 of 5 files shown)


trel:
consider we just use issue 4331 for this work?   no new issue?

plugins/database/src/db_plugin.cpp
```diff
@@ -14762,6 +14766,181 @@ auto db_update_ticket_write_byte_count_op(
...
+            cllBindVars[cllBindVarCount++] = data_expiry_ts.data();
+            cllBindVars[cllBindVarCount++] = data_map_id.data();
+            cllBindVars[cllBindVarCount++] = data_mode.data();
+            cllBindVars[cllBindVarCount++] = r_comment.data();
+            cllBindVars[cllBindVarCount++] = create_ts.data();
+            cllBindVars[cllBindVarCount++] = modify_ts.data();
+            cllBindVars[cllBindVarCount++] = resc_id.data();
+
+            cllBindVars[cllBindVarCount++] = where_data_id.data();
+            cllBindVars[cllBindVarCount++] = where_resc_id.data();
```
korydraughn, line 14908: Do we bind strings for all database operations?

I'm curious to know if many/all of the existing iRODS DB functions do this. If I had to guess, we probably bind strings for every DB op due to the use of `cllBindVars`.

If all supported databases accept strings for every column, it may be worth changing the nanodbc-based API plugins to bind everything as a string as well. We'd likely pay the cost of conversion (i.e. string -> db column type) by the DB, but we'd get the benefit of not having to deal with DB differences. Plus, the code would be a little simpler.

alanking: Yes, `cllBindVars` is an array of `const char*`. I'm pretty sure everything that calls `cmlExecuteNoAnswer` uses `cllBindVars`, but I based this implementation mainly off of `db_mod_data_obj_meta_op`.

alanking: If it works for nanodbc then that seems fine to me. Wouldn't be any worse than the current implementation from my view.

alanking:
This change may not be necessary in light of #5666 so stand by...

trel:
please mark as draft - removes the accidental merge button...

alanking:
Done

alanking:
Superseded by #7033. Closing.

PR #5666, opened 2021-05-24
Oracle support and Large IDs (4-2-stable)

korydraughn:
- Verified that the atomic ACLs/Metadata API plugins work against Oracle.
- Verified that the data object finalize API plugin (using nanodbc) works against Oracle.

This needs to be run through CI.


trel:
great - do we need a couple issue numbers for this work?   what does the question mark on the isolation level commit represent?

korydraughn:
That question mark will be removed in the final PR.

trel:
issue numbers looks good - please # when ready.

korydraughn:
added #'s

PR #5680, opened 2021-06-01
MySQL ODBC config / plugin_base and triggering PEPs (4-2-stable)

korydraughn:


korydraughn:
passed CI
- cen7/postgres
- ub16/postgres
- ub18/postgres

added #

Issue #5682, opened 2021-06-02
ibun fails to finalize archive file due to missing replica information after redirect


alanking:
- [x] master
- [x] 4-2-stable
---
`rsStructFileBundle` redirects to the server which owns the storage resource before calling `rsDataObjCreate` for the archive file which it is supposed to create. This results in a redirect to the catalog provider when `rsPhyPathReg` is called as part of the creation process. The information provided by the database operation is not passed back to the caller as a result of a limitation in the interface, so we must implement a way for `rsDataObjOpen` to get this information if it is not on the catalog provider.

PR #5686, opened 2021-06-02
[#5682] Query for missing replica info in create (4-2-stable)

alanking:
When an operation requires that the connection be redirected to the
catalog consumer which owns the storage resource, the replica
information from the database operation to register the new replica in
the catalog is lost because the rsPhyPathReg interface has no way of
returning the information.

This change allows the logic for creating a new replica to detect the
situation in which the rsPhyPathReg call had to redirect to the catalog
provider and queries the database in order to retrieve the information
which was generated by the registration database operation.

---

CI tests for topology are currently running. Testing branch based on branch in #5680

### DIFF
--- a/server/api/src/rsDataObjOpen.cpp
+++ b/server/api/src/rsDataObjOpen.cpp
@@ -21,6 +21,7 @@
 #include "objDesc.hpp"
 #include "objInfo.h"
 #include "objMetaOpr.hpp"
+#include "miscServerFunct.hpp"
 #include "physPath.hpp"
 #include "rcGlobalExtern.h"
 #include "rcMisc.h"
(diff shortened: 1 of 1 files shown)


server/api/src/rsDataObjOpen.cpp
```diff
@@ -391,6 +392,42 @@ namespace
...
+            // Get the service role for the server to determine whether the catalog-generated information for
+            // the newly created replica is available in the L1 descriptor.
+            std::string service_role;
+            if (const auto ret = get_catalog_service_role(service_role); !ret.ok()) {
+                irods::log(PASS(ret));
+            }
+
+            // Need to get information from the catalog which was populated in the database operation if
+            // this server is not a catalog provider. In that instance, the dataObjInfo in the L1 descriptor
+            // could not have the information.
```
trel, line 404: just asking...  is it 'will not have' or 'cannot have' ?

trel:
looks good, please # when happy.

server/api/src/rsDataObjOpen.cpp
alanking: I think it's "will not have"

alanking:
#'d. CI tests are running but topology tests passed

Issue #5691, opened 2021-06-04
Forced overwrite with iput to default resource which differs from server configuration leaves locked replicas


alanking:
- [x] master
- [x] 4-2-stable
---------
### iRODS Version, OS and Version
4-2-stable

### What did you try to do?
As a user without a configured default resource, `iput -f` over a file which exists on some resource which is neither the default resource configured in the server nor that in the configured rulebase(s) (e.g. `/etc/irods/core.re`).

### Expected behavior
In 4.2.8, this would select an existing replica and overwrite it. We would like for this to be more surgical in the future, but this behavior is acceptable for now.

### Observed behavior (including steps to reproduce, if applicable)
Creates a new replica which is successfully finalized., but then leaves a locked replica

PR #5692, opened 2021-06-04
[#5691] Relax usage of default resource for put (4-2-stable)

alanking:
Matches 4.2.8 behavior. Topology tests still running, but passed in CI for timing, unit, core, and federation tests.


trel:
please # when happy... can make the last one 858 if you want

alanking:
#'d

Issue #5693, opened 2021-06-05
msiRenameLocalZoneCollection does not update the catalog


korydraughn:
- [x] master
- [x] 4-2-stable
---
### iRODS Version, OS and Version
tip of 4-2-stable

### What did you try to do?
Rename the local zone collection using `msiRenameLocalZoneCollection`.

### Expected behavior
I expected all paths in r_coll_main to have the new zone collection prefix.

### Observed behavior (including steps to reproduce, if applicable)
The database table was never updated.

The log file contains messages like the following.
```bash
Jun  4 23:58:51 pid:19415 NOTICE: Warning, pending SQL at cllDisconnect, count: 4
Jun  4 23:58:51 pid:19415 NOTICE: Warning, pending SQL: begin ...
Jun  4 23:58:51 pid:19415 NOTICE: Warning, pending SQL: update R_COLL_MAIN set coll_n ...
Jun  4 23:58:51 pid:19415 NOTICE: Warning, pending SQL: update R_COLL_MAIN set parent ...
Jun  4 23:58:51 pid:19415 NOTICE: Warning, pending SQL: update R_COLL_MAIN set coll_n ...
```
This implies that the database updates are not being committed.

PR #5694, opened 2021-06-05
[#5693] msiRenameLocalZoneCollection commits changes to database (4-2-stable)

korydraughn:


trel:
by removing this test - now we don't appear to have a test that changes the zone name?

i don't see a substitute/replacement `iadmin modzone` test anywhere else in the suite.

korydraughn:
I am planning on adding one. This currently passes CI on ub16/postgres. I'm waiting for MySQL now.

I have manually verified that the database is updated correctly. I will add a test and make sure it passes today.

trel:
gotcha - thanks.    this is the last one(!).

korydraughn:
passed CI

added #

Issue #5352, opened 2021-01-19
rcDataObjRead should verify offset is not past end of file (size in the catalog)


kript:
- [x] master
- [x] 4-2-stable
-----------

### iRODS Version, OS and Version

4.2.7
Ubuntu 18.04

### What did you try to do?

Read a file using samtools/htslib which under the hood uses `rcDataObjLseek`.
The file is marked zero length, but actually contains data on disk, due to past issues.

### Expected behavior

a zero length file is provided as that is the size held in the catalog.

### Observed behavior (including steps to reproduce, if applicable)

The underlying file is read and passed through to the samtools program.
This actually worked out for us, but I thought I'd raise it in case its an unintended consequence...

kript:
Not that I expect you to look at samtools/htslib, but in case; https://github.com/samtools/htslib

trel:
We think the use case that suggests it needs to be possible (and not verify catalog size) is if a data_object is opened with `O_RDRW`...

A `0`-length data_object could be opened, and then `rcDataObjLseek` `50` bytes, and then written (which seems fine, regardless)... or read (which seems... possibly problematic as you mention), but that's what POSIX allows.  It's up to the user to hold things correctly.

If this is a security argument, then we'd like to hear more use cases / scenarios... but for now, I think this is working as intended.

trel:
@kript @korydraughn  any further thoughts?    i don't know that there's anything actionable here at this time.

korydraughn:
I agree with your previous statement. I don't think there is anything actionable here. Also, changing how seek works could negatively affect the parallel transfer engine.

kript:
As with other tools, though, if the catalog says the file size is zero, 0 bytes are what should be passed back?
If the user attempts to seek past the 'end' of the file on a filesystem they would get an error, and I would presume the same here?

trel:
But... that's not true?

```
The lseek() function allows the file offset to be set beyond the end of the file (but this does not change the size of the file). If data is later written at this point, subsequent reads of the data in the gap (a "hole") return null bytes (aq\0aq) until data is actually written into the gap.
```

kjsanger:
I think @trel is correct and I would expect to be able to seek beyond the end of the file. But I wouldn't expect to be able to read data from there.

https://linux.die.net/man/2/read: 

> If the current file offset is at or past the end of file, no bytes are read, and read() returns zero.

trel:
Ah, very good.

[...]

PR #5708, opened 2021-06-20
limit chunking to integer size (4-2-stable)

d-w-moore:
in rsDataObjRead(), dataObjReadInp->len sometimes became negative
because `min<int>` casted  (dataSize - offset) without checking
that the rodsLong_t result would fit into the int (=int32) type, thus causing
no bytes ever to be read by a client for data object sizes greater
than 2 GB.
See https://github.com/irods/python-irodsclient/issues/280

### DIFF
--- a/server/api/src/rsDataObjRead.cpp
+++ b/server/api/src/rsDataObjRead.cpp
@@ -15,6 +15,7 @@
 #include "rsDataObjLseek.hpp"
 
 #include <algorithm>
+#include <limits>
 
 int
 applyRuleForPostProcForRead( rsComm_t *rsComm, bytesBuf_t *dataObjReadOutBBuf, char *objPath ) {
@@ -117,7 +118,15 @@ rsDataObjRead(rsComm_t* rsComm,
         // At this point, we know the file read position is less than the replica's
         // recorded data size. We now have to adjust the requested number of bytes to
         // read so that the client does not read past the recorded data size.
-        dataObjReadInp->len = std::min<int>(dataObjReadInp->len, dataObjInfo->dataSize - output->offset);
+
+        // Perform:
+        //     Len = min(Len, min(DataSize-Offset,INT_MAX))
+        // in a way that doesn't lop off the upper bits to fit a rodsLong_t into the smaller 'int' type
+        // needed by a data object read operation.
+
+        const auto buffer_size = std::min<rodsLong_t>(dataObjReadInp->len, dataObjInfo->dataSize - output->offset);
+        using limits_type = std::numeric_limits<decltype(dataObjReadInp->len)>;
+        dataObjReadInp->len = (buffer_size > static_cast<rodsLong_t>(limits_type::max())) ? limits_type::max() : buffer_size;
     }
 
     const auto bytes_read = l3Read(rsComm, l1descInx, dataObjReadInp->len, dataObjReadOutBBuf);
--- a/scripts/irods/test/rule_texts_for_tests.py
+++ b/scripts/irods/test/rule_texts_for_tests.py
@@ -31,6 +31,12 @@ acAclPolicy {
 #===== Test_AllRules =====
 
 rule_texts['irods_rule_engine_plugin-irods_rule_language']['Test_AllRules'] = {}
+rule_texts['irods_rule_engine_plugin-irods_rule_language']['Test_AllRules']['test_data_obj_read_for_2100MB_file__5709'] = '''
+        msiDataObjOpen("objPath={0}", *fd);
+        *len = {1};
+        msiDataObjRead(*fd, *len, *buf);   # test of read from  data object >= 2GB in size, irods/irods#5709
+        writeLine("stdout", *buf);
+'''
 rule_texts['irods_rule_engine_plugin-irods_rule_language']['Test_AllRules']['test_msiTarFileExtract_big_file__issue_4118'] = '''
 myTestRule {{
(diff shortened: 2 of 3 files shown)


kript:
Interested in the scope of this - will this just affect python-irodsclient clients, or other API such as C/C++ (I'm thinking especially of baton here), and/or does it affect icommands? 

Given the fix appears to be in iRODs itself, I am concerned that, if I read this correctly, one cannot download files larger than 2GB from 4.2.9, is that a correct interpretation?

d-w-moore:
I too am interested, and not quite sure the scope of it. Theoretically this affects any and all programs operating through the rsDataObjRead to fetch their data from an iRODS server, whether the request was mediated from a client or not. As it happens, Python uses data "READ" for its get( ) method, as opposed to going through the data "GET" api.

d-w-moore:
As one might expect, this does not just apply to the python client. 
Note the comments to see what number-of-bytes-read-out values are produced in the serverLog ...
```                                             
read_test() {                                                                          
       msiDataObjOpen("objPath=/tempZone/home/rods/junk.put", *fd);   # a data object of 2100 MB (2.1G)
*len = 2048;    
        msiDataObjRead(*fd, *len, *buf); 
        writeLine("serverLog", "--- size of data block read  ---   ");                              
        writeLine("serverLog", *buf);  
                         # -> "0" on 4.2.9 for a data object of size [ > 2G, < 4G ] but "2048" for other objects >= 2048 bytes    
                         # -> "2048" on 4.2.8 for any known size of data object       >= 2048 bytes                            
        msiDataObjClose(*fd, *close_out);
}
```

trel:
Does this suggest `istream` can also trigger this limitation?

Seems like `iget` is working as expected in this case.

d-w-moore:
It's possible istream uses the read API. will check....

d-w-moore:
Yes, istream also has problems
```
$ dd if=/dev/urandom count=2140k bs=1k  of=2.1Gthing.put
$  iput -f 2.1Gthing.put 
$  istream read  /tempZone/home/rods/2.1Gthing.put >/tmp/stuff
 Error: Failed to read requested number of bytes.
ERROR(1) 10:26AM 
```

d-w-moore:
iget is also affected for a 2-4GB file if "-N1" is used.  I'd assume the same is true if we apply N=3 threads to a larger transfer.  The "buggy" part of the DataObjRead( ) was that if bit#31, ie. the sign bit of (data_size - offset) is set, we apply a min of that 
 suddenly negative ceiling to the "desired" number of bytes to be transferred.

d-w-moore:
Interestingly rsDataObjRead( ) is called for an iget of an 2500MB object using  "-N1" (effectively single buffer get) and also for an iget of a 4500MB object with "-N2". Both cases fail in 4.2.9, but are rectified with the fix we've got in #5708 .

Strangely,the GET api avoids calling rsDataObjRead for all of the larger transfers and #threads (e.g. "-N3") I've tried (ergo irrelevant(?)), and istream appears fine with the fix, as well.

trel:
This is related / due to the fix put in for https://github.com/irods/irods/issues/5352

server/api/src/rsDataObjRead.cpp
```diff
@@ -114,10 +116,27 @@ rsDataObjRead(rsComm_t* rsComm,
             return 0;
         }
 
+        // Store into 'dst' (an integer of bitwidth B) the value of 'src' (bitwidth >= B))
+        //   or the maximum value that will fit into dst, whichever is less.
+
+        auto limit_to_dsttype = [](auto & dst, const auto& src) noexcept {
+            using srcI = std::remove_reference_t< decltype(src) >;
+            using dstI = std::remove_reference_t< decltype(dst) >;
+            static_assert( std::is_signed_v<srcI> == std::is_signed_v<dstI>
```
korydraughn, line 125: Why do we need a static_assert here? It the data types ever result in a failure, we'd have to reevaluate this.

korydraughn (commented):
This solution seems a bit complex. The following achieves the same result.
```c++
const auto t = std::min<rodsLong_t>(dataObjReadInp->len, dataObjInfo->dataSize - output->offset);
using limits_type = std::numeric_limits<decltype(dataObjReadInp->len)>;
dataObjReadInp->len = (t > static_cast<rodsLong_t>(limits_type::max())) ? limits_type::max() : t;
```

server/api/src/rsDataObjRead.cpp
korydraughn: I was just curious.

Consider using a better name than `t`. Perhaps, `new_buffer_size`, `corrected_length`, etc.

d-w-moore: Agreed, using `new_buffer_size` - and,this way is less confusing.

d-w-moore:
Awaiting tests

server/api/src/rsDataObjRead.cpp
```diff
@@ -15,6 +15,8 @@
 #include "rsDataObjLseek.hpp"
 
 #include <algorithm>
+#include <type_traits>
```
korydraughn, line 18: Is this header needed?

d-w-moore: will remove if not needed

d-w-moore:
Ready - force pushed, #'d , took care of extra include and codacy concerns

Issue #5689, opened 2021-06-03
data_object_finalize should only redirect to catalog service provider for catalog operation


alanking:
- [x] master
- [x] 4-2-stable
---
The `data_object_finalize` API plugin redirects to the catalog service provider (CSP) in the connected zone before beginning its operations. However, it remains connected to the CSP for the duration of the operation, including the call to the `fileModified` interface. The redirection should only last long enough to allow the database operation to complete and then return to the agent which called it so that `fileModified` will execute somewhere other than the CSP. This could lead to heavy loads on the CSP which has performance implications.

PR #5714, opened 2021-06-23
[#5689] Limit redirection in data_object_finalize (4-2-stable)

alanking:
In cases where the connected server is not the catalog service
provider, the call to the remote data_object_finalize should not attempt
to perform the file_modified operation, should it be requested in the
input. After the catalog has been updated in the CSP, the RPC should
return and the control should continue on the connected server to
execute the file_modified operation.

This change now explicitly turns off trigger_file_modified in the input
to the redirected call to data_object_finalize in order to only update
the catalog on the CSP (again, if it is not the connected server).

---

CI tests running. Will run tests in topology overnight.

### DIFF
--- a/plugins/api/src/data_object_finalize.cpp
+++ b/plugins/api/src/data_object_finalize.cpp
@@ -362,94 +362,13 @@ namespace
         }
     } // invoke_file_modified
 
-    auto rs_data_object_finalize(
-        RsComm* _comm,
-        BytesBuf* _input,
+    auto finalize(
+        RsComm& _comm,
+        json& _replicas,
+        const std::string_view _bytes_written,
+        const bool _admin_operation,
         BytesBuf** _output) -> int
     {
-        // Connect to the catalog provider before proceeding as this API
-        // deals almost exclusively in database transactions.
-        try {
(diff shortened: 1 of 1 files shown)


plugins/api/src/data_object_finalize.cpp
```diff
@@ -362,40 +362,113 @@ namespace
...
+            *_output = to_bytes_buffer(make_error_object(msg).dump());
 
-                const auto ec = rc_data_object_finalize(host_info.conn, json_input.data(), &json_output);
-                *_output = to_bytes_buffer(json_output);
+            return SYS_CONFIG_FILE_ERR;
+        }
 
-                return ec;
+        try {
+            // This section perform permissions checks and update ticket information
```
trel, line 392: 'perform' -> 'will perform'

trel (commented):
seems good - awaiting topo test results...

plugins/api/src/data_object_finalize.cpp
alanking: Done

alanking:
Fixed typo. Topology testing demonstrated no regressions.

trel:
yeps, please # it

alanking:
#'d

Issue #5723, opened 2021-06-29
Putting and getting a zero length file returns a non-zero file.


JustinKyleJames:
- [x] master
- [x] 4-2-stable
---------

## Bug Report

Putting and getting a zero length file returns a non-zero file.

### iRODS Version, OS and Version

iRODS 4.2.9 tested on Ubuntu 16.

### What did you try to do?

Get a zero length file from iRODS.

### Expected behavior

The file should be zero length.

### Observed behavior (including steps to reproduce, if applicable)

The file was non-zero length.  In my case 32 MiB.

```
irods@justin:~$ touch zerofile
irods@justin:~$ ls -l zerofile
-rw-rw-r-- 1 irods irods 0 Jun 29 10:10 zerofile
irods@justin:~$ iput zerofile
irods@justin:~$ iget zerofile -f
irods@justin:~$ ls -l zerofile
-rw-rw-r-- 1 irods irods 33554432 Jun 29 10:10 zerofile
```

PR #5725, opened 2021-06-29
[#5723] rsDataObjGet sets output buffer size to zero for zero length data object (master)

korydraughn:


server/api/src/rsDataObjGet.cpp
```diff
@@ -82,6 +82,16 @@ namespace
                 __FUNCTION__, __LINE__, bytes_read,
                 replica.logical_path(), replica.hierarchy()));
         }
+        else if (0 == bytes_read) {
+            // Because clients may expect a buffer to always be returned, we set the
+            // length of the returned buffer to zero to indicate that it doesn't contain
+            // any valid data. The real solution is to not allocate anything and the clients
+            // use the bytes read as the indicator for how to move forward.
```
alanking, line 89: Are we going to file an issue to implement the "real solution"?

korydraughn: I should probably reword that.

Allocating a buffer isn't that big of a deal. It's not wrong, just inefficient. If we made this change, we may break clients (which have a bug if they are affected by this PR). Is the real solution a 4.3 or 5.0 thing?

SwooshyCueb: Sounds like maybe a 5.0 thing.

trel: I think we're okay for this PR to be 42s - oh I see... this is about the wording of a 'real' solution in the comment...

korydraughn: i think @SwooshyCueb is right.

this PR is good though

korydraughn:
passed CI

added #

PR #5728, opened 2021-07-02
Delay server and Agent server watchers

epsilon-phase:
This provides a partial solution to issue 4977 and provides a framework which provides for tasks that should be executed at a given interval to be built.

### DIFF
--- a/scripts/irods/test/test_misc.py
+++ b/scripts/irods/test/test_misc.py
@@ -196,3 +196,44 @@ class Test_Misc(session.make_sessions_mixin([('otherrods', 'rods')], []), unitte
             for fn in [pid_file_1, pid_file_2, pid_file_3]:
                 self.assertFalse(os.path.exists(fn))
 
+    @unittest.skipIf(test.settings.RUN_IN_TOPOLOGY, "skip for topology testing")
+    def test_server_respawns_processes__issue_4977(self):
+        import psutil
+        import time
+        from collections import defaultdict
+        main_server = None
+        sescoln = self.admin.session_collection
+        # Find the main server process
+        for proc in psutil.process_iter():
+            # It won't have any parents.
+            if proc.name() == 'irodsServer' \
+               and proc.parent().name() != 'irodsServer':
+                main_server = proc
+                # Kill the agent factory and the rule server
+                for child in proc.children(recursive=True):
+                    child.kill()
+                break
+        #It should run every 5 seconds but it's probably best to wait a lot longer just in case
+        time.sleep(15)
+        object_name = sescoln+"/hello"
+        self.assertTrue(len(main_server.children()) >= 2)
+        # If the API is up after this it seems it should be entirely working.
+        self.admin.assert_icommand(['itouch', object_name])
+        seen_process_names = defaultdict(int)
+        for i in main_server.children():
+            seen_process_names[i.name()] += 1
+        # There should only be a single delay server instance
+        self.assertTrue(seen_process_names['irodsReServer'] == 1)
+        self.assertTrue(seen_process_names['irodsServer'] == 1)
+        attribute = "a1"
+        # Test the delay server
+        self.admin.assert_icommand(['irule', '-r',
+                                    "irods_rule_engine_plugin-irods_rule_language-instance",
+                                    'delay("<INST_NAME>irods_rule_engine_plugin-irods_rule_language-instance</INST_NAME>") {{ msiModAVUMetadata("-d", "{0}", "add", "{1}", "v1", "u2") }}'.format(object_name, attribute),
+                                    'null',
+                                    'ruleExecOut'])
+
+        # Wait for the delay server to process the rule
+        time.sleep(45)
+        self.admin.assert_icommand('imeta ls -d %s' % (object_name), 'STDOUT_SINGLELINE', ['attribute: ' + attribute])
+        self.admin.assert_icommand(['irm', "-f", 'hello'])
--- a/server/core/include/process_manager.hpp
+++ b/server/core/include/process_manager.hpp
@@ -1,0 +1,193 @@
+#ifndef IRODS_PROCESS_MANAGER_HPP
+#define IRODS_PROCESS_MANAGER_HPP
+
+#include <unistd.h>
+#include <functional>
+#include <chrono>
+#include <any>
+#include <thread>
+#include <boost/lexical_cast.hpp>
+#include <memory>
+
+#include "irods_logger.hpp"
+
+/// \file
+/// \parblock
+/// The main use case of this facility is to provide a means of running tasks intermittently.
+/// that is, in an interval.
+/// \endparblock
+/// TODO add since information
+
+namespace irods::experimental::cron{
+
+    /// \brief a cron task, that is, one run intermittently.
+    ///
+    /// TODO add some /since information once it's known what release it's going to be in.
+    class cron_task{
+        using func = std::function<void(std::any&)>;
+        /// \brief the main function in the task
+        std::function<void(std::any&)> execute_;
+        /// \brief The function to be run before the main function of the task
+        /// run only once when #run_pre_once_only is true
+        std::function<void(std::any&)> pre_;
+        /// \brief The function to be run after the main function of the task
+        /// run only once when #run_post_once_only is true
+        std::function<void(std::any&)> post_;
+        /// \brief The length of time between runs, in seconds
+        std::chrono::seconds interval_ = std::chrono::seconds(0);
+        /// \brief The timestamp of the last time this task was run
+        std::chrono::time_point<std::chrono::system_clock> last_run_ =
+            std::chrono::time_point<std::chrono::system_clock>(std::chrono::seconds(0));
+        /// \brief Only run the pre function once.
+        bool run_pre_once_only_ = true;
+        /// \brief Run post only once.
+        bool run_post_once_only_  = true;
+        /// \brief User defined data to be handled by the task.
+        std::any data_;
+
+        /// \brief construct a cron_task
+        /// \param[in] e The function that will be called as the main execution phase
+        /// \param[in] pre The function to call before the main execution. May only be executed once if run_pre_once is set to true
+        /// \param[in] post The function to call after the main execution. May only be executed once if run_post_once is set to true
+        /// \param[in] interval The number of seconds between execution
+        /// \param[in] run_pre_once determines if the pre function is only called at the first invocation of the task's main function
+        /// \param[in] run_post_once determines if the post function is only called at the first invocation of the task's main function
+        /// \param[in] data user supplied data that may be manipulated by the task's function
+        cron_task(func e, func pre, func post, std::chrono::seconds interval, bool run_pre_once, bool run_post_once,
+                  std::any data):
+            execute_(e),
+            pre_(pre),
+            post_(post),
+            interval_(interval),
+            last_run_(std::chrono::seconds(0)),
+            run_pre_once_only_(run_pre_once),
+            run_post_once_only_(run_post_once),
+            data_(data)
+        {
+        }
+    public:
+
+        /// \brief Returns true if the task is ready to run
+        /// \retval true if sufficient time has elapsed
+        bool ready() const noexcept {
+            auto now = std::chrono::system_clock::now();
+            return (now - last_run_) > interval_;
+        }
+        /**
+         * \brief Returns true if the task has run before, as determined by the last run timestamp
+         *
+         * \retval true if the task has run before
+         */
+        bool has_run()const noexcept{
+            return last_run_.time_since_epoch().count() != 0;
+        }
+        /**
+         * \brief Run the task
+         */
+        void run(){
+            if( !ready() ){
+                return;
+            }
+            if(pre_ && ((run_pre_once_only_ && !has_run()) || !run_pre_once_only_)){
+                pre_(data_);
+            }
+            execute_(data_);
+            if(post_ && ((run_post_once_only_ && !has_run() ) || !run_post_once_only_) ){
+                post_(data_);
+            }
+            last_run_ = std::chrono::system_clock::now();
+        }
+        friend struct cron_builder;
+    }; // class cron_task
+
+    /// \brief Builder for a CronTask.
+    struct cron_builder {
+        std::function<void(std::any&)> to_execute_;
+        std::function<void(std::any&)> pre_function_;
+        std::function<void(std::any&)> post_function_;
+        bool run_pre_once_ = true;
+        bool run_post_once_ = true;
+        std::chrono::seconds interval_ = std::chrono::seconds(0);
+        std::any data_;
+
+        /// \brief set the pre action to be run before the main function
+        ///
+        /// \param[in] f The function to run before the main event, whatever that might be
+        cron_builder& pre(std::function<void(std::any&)> f){ pre_function_ = f; return *this; }
+
+        /// \brief set the post function, to be executed before the execution function
+        ///
+        /// \param[in] f The function to be run after the primary function of the task. By default only run once
+        cron_builder& post(std::function<void(std::any&)> f){ post_function_ = f; return *this; }
+
+        /// \brief Set the execution function to be run.
+        ///
+        /// \param[in] The main function to be executed
+        cron_builder& to_execute(std::function<void(std::any&)> f) noexcept { to_execute_ = f; return *this; }
+
+        /// \brief Set the interval in seconds between runs of this task. A value of Zero(the default), results
+        /// in the task being run every time the cron manager loops.
+        ///
+        /// \param[in] n The number of seconds between runs of the task
+        cron_builder& interval(int n) noexcept { interval_ = std::chrono::seconds(n); return *this; }
+
+        /// \brief Set the user defined data to be made available to the task in question
+        ///
+        /// \param[in] a The user defined data to be made available.
+        cron_builder& data(std::any& a) noexcept { data_ = a; return *this; }
+
+        /// \brief Sets whether or not post function should run more than once. Defaults to true.
+        ///
+        /// \param[in] a a bool indicating whether the post function should only run once.
+        cron_builder& run_post_op_once(bool a) noexcept { run_post_once_ = a; return *this; }
+
+        /// \brief Sets whether or not the pre function should be run every time the
+        /// main function is run. Defaults to true.
+        ///
+        /// \param[in] a a bool indicating whether or not the pre function should only run once.
+        cron_builder& run_pre_op_once(bool a) noexcept { run_pre_once_ = a; return *this; }
+
+        /// \brief Convert the cron_builder into a cron_task instance.
+        ///
+        /// \retval A cron_task object, ready to be inserted into the cron system.
+        cron_task build() const noexcept{
+            return cron_task(to_execute_,
+                             pre_function_,
+                             post_function_,
+                             interval_,
+                             run_pre_once_,
+                             run_post_once_,
+                             data_);
+        }
+    }; // struct cron_builder
+
+    /// \brief Management struct for the facility. A singleton.
+    struct cron{
+        /// \brief Loop across all the tasks in the cron structure
+        void run(){
+            for(auto& task: tasks) {
+                task.run();
+            }
+
+        }
+        /// \brief Get the single instance of the cron facility.
+        static cron* get() noexcept {
+            static cron singleton;
+            return &singleton;
+        }
+        /// \brief Add a task to the cron facility.
+        void add_task(cron_task&& ct){
+            tasks.emplace_back(std::move(ct));
+        }
+    private:
+        std::vector<cron_task> tasks;
+        cron(){}
+        cron(cron&) = delete;
+        cron(const cron&) = delete;
+        cron(cron&&) = delete;
+        cron& operator=(const cron&)=delete;
+
+    }; // struct cron
+} // namespace irods::experimental::cron
+
+#endif // IRODS_PROCESS_MANAGER_HPP
--- a/server/core/src/rodsServer.cpp
+++ b/server/core/src/rodsServer.cpp
@@ -27,6 +27,7 @@
 #include "hostname_cache.hpp"
 #include "dns_cache.hpp"
 #include "server_utilities.hpp"
+#include "process_manager.hpp"
 
 #include <pthread.h>
 #include <sys/socket.h>
@@ -48,6 +49,7 @@
 #include <fstream>
 #include <regex>
 #include <algorithm>
+#include <optional>
 #include <iterator>
 
 // clang-format off
@@ -330,7 +332,7 @@ int main(int argc, char** argv)
     if (!write_to_stdout) {
         daemonize();
     }
-
+    
     init_logger(write_to_stdout, enable_test_mode);
 
     ix::log::server::info("Initializing server ...");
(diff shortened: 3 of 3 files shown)


server/core/src/rodsServer.cpp
```diff
@@ -421,14 +422,49 @@ int main(int argc, char** argv)
...
+                    int saved_errno = errno;
+                    if ( ( time( 0 ) - sock_connect_start_time ) > 5 ) {
+                        rodsLog(LOG_ERROR, "Error connecting to agent factory socket, errno = [%d]: %s",
+                                saved_errno, strerror( saved_errno ) );
+                        exit(SYS_SOCK_CONNECT_ERR);
+                    }
+
+                }
+                agent_spawning_pid=new_pid;
+            }
```
korydraughn, line 463: maybe wrap in a function to avoid duplicate code.

server/core/src/rodsServer.cpp
```diff
@@ -1306,7 +1343,39 @@ int initServerMain(
...
+                    argv.push_back(nullptr);
+
+                    // Launch the delay server!
+                    execv(argv[0], &argv[0]);
+                    exit(1);
+                }
+                else {
+                    irods::set_server_property<int>(irods::RE_PID_KW, pid);
+                }
+            }
```
korydraughn, line 1375: wrap in a function to avoid code duplication

server/core/include/process_manager.hpp
```diff
@@ -0,0 +1,86 @@
...
+                setup(data);
+            execute(data);
+            if((run_post_once_only&&!has_run())||!run_post_once_only)
+                post(data);
+            last_run=std::chrono::system_clock::now();
+        }
+        std::any data;
+    };
+    /// Builder for a CronTask.
+    struct cron_builder {
```
korydraughn, line 45: consider special member functions (constructors, destructors, etc.).

consider copy/move semantics.

server/core/include/process_manager.hpp
```diff
@@ -0,0 +1,86 @@
...
+        cron_builder& post(std::function<void(std::any&)> f){_post_function=f;return *this;}
+        cron_builder& to_execute(std::function<void(std::any&)> f){_to_execute=f;return *this;}
+        cron_builder& interval(int n){_interval = n; return *this;}
+        virtual cron_task finalize(){
+            return {_to_execute,_setup_function,_post_function,std::chrono::seconds(_interval),
+                std::chrono::time_point<std::chrono::system_clock>(std::chrono::seconds(0)),_run_setup_once,_run_post_once,_data};
+        }
+    };
+
+    struct cron{
```
korydraughn, line 63: consider special member functions (constructors, destructors, etc.).

consider copy/move semantics.

scripts/irods/test/test_misc.py
```diff
@@ -196,3 +196,24 @@ def test_server_removes_leftover_temporary_rulebase_pid_files_on_startup__issue_
             for fn in [pid_file_1, pid_file_2, pid_file_3]:
                 self.assertFalse(os.path.exists(fn))
 
+    @unittest.skipIf(test.settings.RUN_IN_TOPOLOGY, "skip for topology testing")
+    def test_server_respawns_processes__issue_4977(self):
```
korydraughn, line 200: handle the delay server.

make sure it doesn't stand up more than one delay server.

korydraughn (commented):
please add doxygen documentation.

some examples:
- replica_access_table.hpp
- replica_state_table.hpp
- client_connection.hpp

server/core/src/rodsServer.cpp
epsilon-phase: :+1:

scripts/irods/test/test_misc.py
epsilon-phase: :+1:

trel:
green - good stuff.

i see one remaining curly braces comment.

scripts/irods/test/test_misc.py
```diff
@@ -196,3 +196,37 @@ def test_server_removes_leftover_temporary_rulebase_pid_files_on_startup__issue_
...
+        for i in main_server.children():
+            seen_process_names[i.name()] += 1
+        # There should only be a single delay server instance
+        self.assertTrue(seen_process_names['irodsReServer'] == 1)
+        # Test the delay server
+        self.admin.assert_icommand(['irule', '-r',
+                                    "irods_rule_engine_plugin-irods_rule_language-instance",
+                                    "'delay(\"\") { writeLine(\"serverLog\", \"DID IT!\"); }'",
+                                    'null',
+                                    'ruleExecOut'])
```
trel, line 232: rather than writing to the log... we could add an AVU to something, sleep 31s, and then confirm that the AVU is in the system.   no log checking.

epsilon-phase: Alright, I'm gonna need some guidence on how to make that work.

server/core/src/rodsServer.cpp
```diff
@@ -394,41 +395,46 @@ int main(int argc, char** argv)
...
+    ix::cron::cron_builder agent_watcher;
+    auto start_agent_server = [&](std::any& _) {
+        int status;
+        int w = waitpid(agent_spawning_pid, &status, WNOHANG);
+        if( w != 0 ) {
+            ix::log::server::info("Restarting agent spawner");
+            auto new_pid = fork();
+            if( new_pid == 0 ) {
+                ProcessType = AGENT_PT;
+                runIrodsAgentFactory(local_addr);
```
korydraughn, line 407: The original code returned the result of this function call. We need to investigate how to maintain the original behavior here.

epsilon-phase: where did it expose it?

Didn't it just end up getting gobbled up by the fork without it ever being recorded?

epsilon-phase: though, if it's about the process needing to exit with a specifc code, I guess that `exit(runIrodsAgentFactory(local_addr));` would probably suffice

scripts/irods/test/test_misc.py
trel: `msiModAVUMetadata` is the thing in the rule itself to set an AVU... and then an assert in python with `imeta ls -d` to confirm it made it in after the delay server fires the rule.

https://docs.irods.org/4.2.9/doxygen/extractAvuMS_8cpp.html#a547b0640dacd3b8d58979688f6608b17

something like....
```
self.admin.assert_icommand('imeta ls -d %s' % (object_name), 'STDOUT_SINGLELINE', ['attribute: ' + attribute])
```

server/core/src/rodsServer.cpp
korydraughn: `exit` may work, but I'm nervous that it may not do the full job. Some code in `main` expects to run during the call to destructors and `exit` doesn't guarantee that.

see https://en.cppreference.com/w/cpp/utility/program/exit

epsilon-phase: Does the stack get unwound when the process aborts or is killed any other way?

scripts/irods/test/test_misc.py
```diff
@@ -196,3 +196,43 @@ def test_server_removes_leftover_temporary_rulebase_pid_files_on_startup__issue_
...
+        seen_process_names = defaultdict(int)
+        for i in main_server.children():
+            seen_process_names[i.name()] += 1
+        # There should only be a single delay server instance
+        self.assertTrue(seen_process_names['irodsReServer'] == 1)
+        attribute = "a1"
+        # Test the delay server
+        self.admin.assert_icommand(['irule', '-r',
+                                    "irods_rule_engine_plugin-irods_rule_language-instance",
+                                    "'delay(\"\") { msiModAVUMetadata(\"-d\",\"%s\",\"add\",\"a1\",\"v1\",\"u2\"); writeLine(\"serverLog\",\"Did it, %s\") }'"%(object_name,object_name),
```
korydraughn, line 231: The first thing i see is that your call to `delay()` is missing the hints argument. The `delay()` function needs a target to run the body of the delay rule against. Try replacing line 231 with the following:
```bash
'delay("<INST_NAME>irods_rule_engine_plugin-irods_rule_language-instance</INST_NAME>") { msiModAVUMetadata("-d", "{0}", "add", "a1", "v1", "u2"); writeLine("serverLog", "XXX FROM TEST XXX"); }'.format(object_name),
```
Another thing to keep in mind is that the delay server runs every 30 seconds by default. Normally, the tests adjusts the config file so that the delay server wakes up more often (like every second). For your test, I'd recommend going into `/etc/irods/server_config.json` and adjusting the delay server options under `advanced_settings`. The option you need to adjust is called `rule_engine_server_sleep_time_in_seconds`. You'll need to restart the server after updating the config file.

Here is the link explaining each option: https://docs.irods.org/4.2.9/system_overview/configuration/

server/core/src/rodsServer.cpp
epsilon-phase: Okay, so I can cause the destructors to fire off if I throw, for example, the return code from the forked process.

trel:
please make any remaining changes in new commits until it's time to squash/merge - we can't see a diff with the force pushes.

epsilon-phase:
Alright, will do

-------- Original Message --------

server/core/src/rodsServer.cpp
```diff
@@ -1349,66 +1316,48 @@ int initServerMain(
...
-          }
-        }).interval(5);
-
-        ix::cron::Cron::get()->tasks.push_back(delay_server.finalize());
-        const int pid = RODS_FORK();
-        
-        if (pid == 0) {
-            close(svrComm->sock);
+            if(delay_pid == -2 || waitpid(delay_pid,nullptr,WNOHANG)!=0){
+                ix::log::server::info("Restarting rule server");
```
trel, line 1333: "rule server" -> "delay server"

server/core/src/rodsServer.cpp
```diff
@@ -395,82 +395,49 @@ int main(int argc, char** argv)
...
+            auto new_pid = fork();
+            if( new_pid == 0 ) {
+                ProcessType = AGENT_PT;
+
+                // Throwing the return code causes it to unwind the stack completely
+                // and execute all the destructors.
+                throw runIrodsAgentFactory(local_addr);
+            } else {
+                close(agent_conn_socket);
+                ix::log::server::info("Restarting agent spawner");
```
trel, line 413: "agent spawner" -> "agent factory"

server/core/src/rodsServer.cpp
```diff
@@ -395,82 +395,49 @@ int main(int argc, char** argv)
...
-        ssize_t status = connect( agent_conn_socket, (const struct sockaddr*) &local_addr, len );
-        if ( status >= 0 ) {
-            break;
-        }
+    ix::cron::cron_builder agent_watcher;
+    auto start_agent_server = [&](std::any& _) {
+        int status;
+        int w = waitpid(agent_spawning_pid, &status, WNOHANG);
+        if( w != 0 ) {
+            ix::log::server::info("Restarting agent spawner");
```
trel, line 403: "Restarting agent spawner" -> "Starting agent factory"

server/core/src/rodsServer.cpp
```diff
@@ -438,7 +438,11 @@ int main(int argc, char** argv)
     start_agent_server(dummy);
     agent_watcher.to_execute(start_agent_server).interval(5);
     ix::cron::cron::get()->add_task(agent_watcher.finalize());
-    return serverMain(enable_test_mode, write_to_stdout);
+    try{
+        return serverMain(enable_test_mode, write_to_stdout);
+    }catch(int i){
```
trel, line 443: just style: newline before catch

scripts/irods/test/test_misc.py
```diff
@@ -196,3 +196,43 @@ def test_server_removes_leftover_temporary_rulebase_pid_files_on_startup__issue_
...
+        seen_process_names = defaultdict(int)
+        for i in main_server.children():
+            seen_process_names[i.name()] += 1
+        # There should only be a single delay server instance
+        self.assertTrue(seen_process_names['irodsReServer'] == 1)
+        attribute = "a1"
+        # Test the delay server
+        self.admin.assert_icommand(['irule', '-r',
+                                    "irods_rule_engine_plugin-irods_rule_language-instance",
+                                    'delay("<INST_NAME>irods_rule_engine_plugin-irods_rule_language-instance</INST_NAME>") {{ msiModAVUMetadata("-d", "{0}", "add", "a1", "v1", "u2"); writeLine("serverLog", "XXX FROM TEST XXX"); }}'.format(object_name),
```
korydraughn, line 231: remove call to writeLine

scripts/irods/test/test_misc.py
```diff
@@ -196,3 +196,43 @@ def test_server_removes_leftover_temporary_rulebase_pid_files_on_startup__issue_
...
+                                    'null',
+                                    'ruleExecOut'])
+
+        # Wait for the delay server to process the rule
+        time.sleep(45)
+        self.admin.assert_icommand('imeta ls -d %s' % (object_name), 'STDOUT_SINGLELINE', ['attribute: ' + attribute])
+        self.admin.assert_icommand(['irm','hello'])
```
korydraughn, line 238: prefer `irm -f ...` to avoid the trash collection

server/core/include/process_manager.hpp
```diff
@@ -0,0 +1,171 @@
...
+        /// \brief the main function in the task
+        std::function<void(std::any&)> execute;
+        /// \brief The function to be run before the main function of the task
+        /// run only once when #run_pre_once_only is true
+        std::function<void(std::any&)> pre;
+        /// \brief The function to be run after the main function of the task
+        /// run only once when #run_post_once_only is true
+        std::function<void(std::any&)> post;
+        /// \brief The length of time between runs, in seconds
+        std::chrono::seconds interval;
```
korydraughn, line 40: confirm initialization

server/core/include/process_manager.hpp
```diff
@@ -0,0 +1,171 @@
...
+    /// TODO add some /since information once it's known what release it's going to be in.
+    struct cron_task{
+        /// \brief the main function in the task
+        std::function<void(std::any&)> execute;
+        /// \brief The function to be run before the main function of the task
+        /// run only once when #run_pre_once_only is true
+        std::function<void(std::any&)> pre;
+        /// \brief The function to be run after the main function of the task
+        /// run only once when #run_post_once_only is true
+        std::function<void(std::any&)> post;
```
korydraughn, line 38: confirm initialization

server/core/include/process_manager.hpp
```diff
@@ -0,0 +1,171 @@
...
+
+        /// \brief set the pre action to be run before the main function
+        ///
+        /// \param[in] f The function to run before the main event, whatever that might be
+        cron_builder& pre(std::function<void(std::any&)> f){pre_function_ = f; return *this;}
+
+        /// \brief set the post function, to be executed before the execution function
+        ///
+        /// \param[in] f The function to be run after the primary function of the task. By default only run once
+        cron_builder& post(std::function<void(std::any&)> f){post_function_ = f;return *this;}
```
korydraughn, line 100: space after semicolon

server/core/include/process_manager.hpp
```diff
@@ -0,0 +1,171 @@
...
+
+        /// \brief Set the execution function to be run.
+        ///
+        /// \param[in] The main function to be executed
+
+        cron_builder& to_execute(std::function<void(std::any&)> f){to_execute_ = f;return *this;}
```
korydraughn, line 106: space after semicolon

server/core/include/process_manager.hpp
```diff
@@ -0,0 +1,171 @@
...
+        /// \brief Set the interval in seconds between runs of this task. A value of Zero(the default), results
+        /// in the task being run every time the cron manager loops.
+        ///
+        /// \param[in] n The number of seconds between runs of the task
+        cron_builder& interval(int n){interval_ = n; return *this;}
+
+        /// \brief Set the user defined data to be made available to the task in question
+        ///
+        /// \param[in] a The user defined data to be made available.
+        cron_builder& data(std::any& a){data_ = a;return *this;}
```
korydraughn, line 117: space after semicolon

server/core/include/process_manager.hpp
```diff
@@ -0,0 +1,171 @@
...
+        ///
+        /// \param[in] The main function to be executed
+
+        cron_builder& to_execute(std::function<void(std::any&)> f){to_execute_ = f;return *this;}
+
+        /// \brief Set the interval in seconds between runs of this task. A value of Zero(the default), results
+        /// in the task being run every time the cron manager loops.
+        ///
+        /// \param[in] n The number of seconds between runs of the task
+        cron_builder& interval(int n){interval_ = n; return *this;}
```
korydraughn, line 112: maybe use `std::chrono::seconds` as the parameter type to express intent.

this is mainly for the reader of your class.

server/core/include/process_manager.hpp
```diff
@@ -0,0 +1,171 @@
...
+                    run_pre_once_,
+                    run_post_once_,
+                    data_};
+        }
+    }; // struct cron_builder
+
+    /// \brief Management struct for the facility. A singleton.
+    struct cron{
+        /// \brief Loop across all the tasks in the cron structure
+        void run(){
```
korydraughn, line 148: consider marking as `const`

server/core/src/rodsServer.cpp
```diff
@@ -394,42 +395,55 @@ int main(int argc, char** argv)
...
+    ix::cron::cron::get()->add_task(agent_watcher.finalize());
+    try{
+        return serverMain(enable_test_mode, write_to_stdout);
 
-        int saved_errno = errno;
-        if ( ( time( 0 ) - sock_connect_start_time ) > 5 ) {
-            rodsLog(LOG_ERROR, "Error connecting to agent factory socket, errno = [%d]: %s", saved_errno, strerror( saved_errno ) );
-            return SYS_SOCK_CONNECT_ERR;
-        }
+    }catch(int i){
```
korydraughn, line 444: use a more descriptive for `i`

server/core/src/rodsServer.cpp
```diff
@@ -1306,34 +1321,48 @@ int initServerMain(
 
     if (reServerHost && LOCAL_HOST == reServerHost->localFlag) {
         rodsLog(LOG_NOTICE, "Forking Rule Execution Server (irodsReServer) ...");
+        ix::cron::cron_builder delay_server;
+        auto start_delay_server = [&](std::any& _){
+            
+            pid_t delay_pid = -1;
```
korydraughn, line 1327: use `std::optional` instead of `-2`

server/core/src/rodsServer.cpp
```diff
@@ -394,42 +395,55 @@ int main(int argc, char** argv)
     snprintf(local_addr.sun_path, sizeof(local_addr.sun_path), "%s", agent_factory_socket_file);
 
     ix::log::server::info("Forking agent factory ...");
+    ix::cron::cron_builder agent_watcher;
+    auto start_agent_server = [&](std::any& _) {
+        int status;
+        int w = waitpid(agent_spawning_pid, &status, WNOHANG);
+        if( w != 0 ) {
+            ix::log::server::info("Restarting agent factory");
```
korydraughn, line 403: swap with line 413

server/core/src/rodsServer.cpp
epsilon-phase: So that doesn't work because there are plenty of `catch (...)` are in the way, it results in far too many agent factories being spawned. I think

trel (commented):
everything looks good - please squash to the 'right number' of commits - maybe just one?

commit title should begin with [4977], we'll # it with a final force-push once letter-perfect.

commit description should tell a story about the what and the why.

server/core/include/process_manager.hpp
```diff
@@ -0,0 +1,185 @@
...
+        /// \brief Run post only once.
+        bool run_post_once_only  = true;
+        /// \brief User defined data to be handled by the task.
+        std::any data;
+        cron_task(func e, func pre, func post, std::chrono::seconds i, bool run_pre_once, bool run_post_once,
+                  std::any data):
+            execute(e),
+            pre(pre),
+            post(post),
+            interval(i),
```
korydraughn, line 56: `last_run` needs to be added to this list.

server/core/include/process_manager.hpp
```diff
@@ -0,0 +1,185 @@
...
+            return (now - last_run) > interval;
+        }
+        /**
+         * \brief Returns true if the task has run before, as determined by the last run timestamp
+         *
+         * \retval true if the task has run before
+         */
+        bool has_run()const noexcept{
+            return last_run.time_since_epoch().count() == 0;
+        }
```
korydraughn, line 77: is this logic correct?

following a successful run, this will return false.

server/core/include/process_manager.hpp
```diff
@@ -0,0 +1,185 @@
...
+        /// \brief Add a task to the cron facility.
+        void add_task(cron_task&& ct){
+            tasks.emplace_back(std::move(ct));
+        }
+    private:
+        std::vector<cron_task> tasks;
+        cron(){}
+        cron(cron&) = delete;
+        cron(cron&&) = delete;
```
korydraughn, line 181: missing:
- `cron& operator=(const cron&) = delete;`
- `cron& operator=(cron&&) = delete;`

you can disable all copy and move support by marking the copy constructor and copy assignment operator as deleted. for example:
```c++
cron(const cron&) = delete;
cron& operator=(const cron&) = delete;
```

server/core/src/rodsServer.cpp
```diff
@@ -393,42 +395,56 @@ int main(int argc, char** argv)
...
-    
-    // Main iRODS server (parent)
-    if (agent_spawning_pid < 0) {
-        rodsLog( LOG_ERROR, "fork() failed when attempting to create agent factory process" );
-        return SYS_FORK_ERROR;
-    }
-
-    ix::log::server::info("Agent factory PID = [{}]", agent_spawning_pid);
+    ix::cron::cron_builder agent_watcher;
+    auto start_agent_server = [&](std::any& _) {
```
korydraughn, line 399: this can use some `const`.

`const auto start_agent_server = ...`

server/core/src/rodsServer.cpp
```diff
@@ -1301,39 +1318,57 @@ int initServerMain(
...
-
-    if (reServerHost && LOCAL_HOST == reServerHost->localFlag) {
-        rodsLog(LOG_NOTICE, "Forking Rule Execution Server (irodsReServer) ...");
 
-        const int pid = RODS_FORK();
-        
-        if (pid == 0) {
-            close(svrComm->sock);
+    ix::cron::cron_builder delay_server;
+    auto start_delay_server = [&](std::any& _){
```
korydraughn, line 1323: this can use some `const`.

`const auto start_delay_server = ...`

korydraughn (commented):
it's getting better and better. we're close.

server/core/src/rodsServer.cpp
```diff
@@ -393,42 +395,56 @@ int main(int argc, char** argv)
...
+        int w = waitpid(agent_spawning_pid, &status, WNOHANG);
+        if( w != 0 ) {
+            ix::log::server::info("Starting agent factory");
+            auto new_pid = fork();
+            if( new_pid == 0 ) {
+                ProcessType = AGENT_PT;
+
+                // Throwing the return code causes it to unwind the stack completely
+                // and execute all the destructors.
+                exit(runIrodsAgentFactory(local_addr));
```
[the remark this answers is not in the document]
epsilon-phase: Yeah, as it turns out, it ends up being caught by a `catch(...)` so there's not much better I can do, from what I can tell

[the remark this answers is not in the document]
korydraughn: ok. we can investigate this later on. leave a comment explaining the situation for the next person.

server/core/include/process_manager.hpp
epsilon-phase: Yeah, that is definitely a mistake

server/core/include/process_manager.hpp
```diff
@@ -0,0 +1,185 @@
...
+/// The main use case of this facility is to provide a means of running tasks intermittently.
+/// that is, in an interval.
+/// \endparblock
+/// TODO add since information
+
+namespace irods::experimental::cron{
+    namespace detail{
+        /// \brief Placeholder function for cron_task
+        void noop(std::any&) noexcept {}
+    } // namespace detail
```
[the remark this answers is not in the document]
epsilon-phase: Alright, that's certainly easy enough

Issue #5733, opened 2021-07-07
Repeated "There is no transaction in progress" warning in database logs


alanking:
### iRODS Version, OS and Version
4.2.9 (possibly earlier versions)

### Observed behavior (including steps to reproduce, if applicable)
Running a docker-compose setup with a separate container to host the catalog based on the official postgres docker image, I see the following warning message coming from the output of `docker logs` 2-5 times per second while running tests:
```
icat_1                              | 2021-07-06 21:26:31.080 UTC [7825] WARNING:  there is no transaction in progress
```
It is unclear at this point what is causing these messages. It doesn't seem to be a big problem (everything is working) but it should be investigated.

alanking:
I have narrowed down the cause of this issue as being the DB operation created for use in `data_object_finalize`, `db_check_permission_to_modify_data_object_op`. It is using `cmlCheckDataObjId`, which will update ticket information should there be a ticket involved in modifying a data object. The DB operation calls "rollback" on failure and "commit" on success to publish any changes to the database with respect to tickets that may have occurred as a result of the call to `cmlCheckDataObjId`. If no such changes occurred, a commit is attempted anyway, which results in this warning because no transaction is in progress.

Here is the output from the posgresql log:
```
icat_1                              | 2021-07-07 17:04:56.511 UTC [36] LOG:  duration: 0.034 ms  execute <unnamed>: select object_id from R_OBJT_ACCESS OA, R_DATA_MAIN DM, R_USER_GROUP UG, R_USER_MAIN UM, R_TOKN_MAIN TM where OA.object_id=$1 and UM.user_name=$2 and UM.zone_name=$3 and UM.us
er_type_name!='rodsgroup' and UM.user_id = UG.user_id and OA.object_id = DM.data_id and UG.group_user_id = OA.user_id and OA.access_type_id >= TM.token_id and  TM.token_namespace ='access_type' and TM.token_name = $4 limit 1
icat_1                              | 2021-07-07 17:04:56.511 UTC [36] DETAIL:  parameters: $1 = '50154', $2 = 'rods', $3 = 'tempZone', $4 = 'modify metadata'  
[...]

PR #5735, opened 2021-07-08
[#5733] Only commit for tickets in check_permission_to_modify_data_object (master)

alanking:
This is not the complete solution for #5733, but does eliminate the bulk of the warning messages in the database logs. This will still result in a warning message when checking the permissions for an object being finalized via ticket access when no update has occurred for the ticket, but this is a much more niche case than every single call to finalize.

Tests running now


trel:
looks good to me - please # once green.

alanking:
Tests passed. #'d

PR #5738, opened 2021-07-08
[#5733] Only commit for tickets in check_permission_to_modify_data_object (4-2-stable)

alanking:
Tests running now


alanking:
Core, federation, and topology-from-icat tests passed. #'d

PR #5743, opened 2021-07-13
[#5300] Server now creates PID file (4-2-stable)

korydraughn:
This PR attempts to keep the irodsctl script from killing iRODS processes inside of a container by introducing a PID file.

With the PID file, the irodsctl script can confidently determine if a process is a descendant or not.


korydraughn:
After discussion, this enhancement has been marked as 4.3.0 only

PR #5751, opened 2021-07-15
Implicit Remote

epsilon-phase:


server/core/src/irodsReServer.cpp
```diff
@@ -77,6 +78,30 @@ namespace {
         }
     }
 
+    void load_available_executors(){
+        auto exec = irods::get_server_property<std::unordered_map<std::string, int> >(irods::HOSTS_AVAILABLE_RULE_EXECUTORS);
+        for(const auto &[hostname, port] : exec){
```
trel, line 83: i think we know the port... they're in the same zone as us... so they have to be using the same port...

hmm, pretty sure about that... @korydraughn @alanking 

if so, don't need a `std::pair`... just the hostnames.

server/core/src/irodsReServer.cpp
```diff
@@ -77,6 +78,30 @@ namespace {
...
+        rodsEnv env;
+        _getRodsEnv(env);
+        available_hosts.push_back(std::make_pair(env.rodsHost,
+                                                 env.rodsPort));
+    }
+    ix::client_connection get_next_connection(){
+        static std::atomic<int> host_index = 0;
+        if(available_hosts.empty())
+            load_available_executors();
+        int cur = (host_index++) % available_hosts.size();
```
trel, line 95: is this random enough?

server/core/src/irodsReServer.cpp
korydraughn: i'm not really sure, but that does sound correct.

we might only care about the port when we're dealing with federation.

trel: yes, and federation will NOT matter here, the delay server is in charge of rules in its Zone only.   if it's a rule that should be run in a different Zone, it will be stored in that other Zone's catalog.

server/core/src/irodsReServer.cpp
korydraughn: do we want/need randomness?

round robin would guarantee that the delay server distributes work evenly across the executors. it would be more deterministic too.

server/core/src/irodsReServer.cpp
alanking: To be honest, I've never actually tried running a remote rule on a remote zone. Could be an interesting experiment. :)

Also, zone's must all have the same port (https://docs.irods.org/4.2.9/system_overview/configuration/): 

> zone_port (required) (default 1247) - The main port used by the Zone for communication. This must be the same across all iRODS servers in a Zone.

korydraughn: nice find.

server/core/src/irodsReServer.cpp
trel: yes?    why do we want determinism?  what happens when someone takes one out of the list?   hmm... more thoughts...

korydraughn: it being deterministic isn't that important, but it can help with testing. i don't see randomness helping much here.

the delay server doesn't see changes to the config file right now. if it did respond to changes to the list of executors, i'd expect the delay server to pick the next executor and attempt to connect and execute the rule.

lib/core/src/irods_configuration_keywords.cpp
```diff
@@ -166,5 +166,6 @@ namespace irods
 
     // misc. keywords
     const std::string HOSTS_CONFIG_JSON_OBJECT_KW("hosts_config_json_object");
+    const std::string HOSTS_AVAILABLE_RULE_EXECUTORS("available_executors");
```
trel, line 169: since things are mostly working now - the naming has come back up...

consider...
- `delay_rule_executors`
- `delayed_rule_executors`
- `delay_rule_executor_hostnames`
- `delayed_rule_executor_hostnames`
- `hostnames_to_execute_delay_rules`
- `hostnames_to_execute_delayed_rules`
- `hostnames_for_delay_execution`
- `hostnames_for_delayed_execution`
- `hostnames_for_execution_of_delay_rules`
- `hostnames_for_execution_of_delayed_rules`

epsilon-phase:
This is ready for review.

scripts/irods/test/test_delay_queue.py
```diff
@@ -602,6 +602,35 @@ def test_delay_rule_does_not_accept_invalid_priority_levels__issue_2759(self):
         self.admin.assert_icommand(['irule', '-r', rep_name, delay_rule.format(rep_name, '10'), 'null', 'ruleExecOut'], 'STDERR', expected_output)
         self.admin.assert_icommand(['iquest', 'select RULE_EXEC_ID'], 'STDOUT', ['CAT_NO_ROWS_FOUND']) # Show that no rules exist.
 
+    @unittest.skipIf(not test.settings.RUN_IN_TOPOLOGY, 'Not running a topology test')
```
korydraughn, line 605: i believe there is also a `skipUnless`. that would allow us to remove the `not` and it might make this a little clearer.

up to you.

server/core/src/irodsReServer.cpp
```diff
@@ -77,6 +78,52 @@ namespace {
...
+        if(available_hosts.empty()){
+            available_hosts.push_back(env.rodsHost);
+        }
+        logger::delay_server::info("Executors populated");
+    }
+
+    /// Get the next connection to use for the next rule
+    ///
+    /// \param[in] user The name of the user to proxy into
+    ix::client_connection get_next_connection(const std::optional<std::string>& _user){
```
korydraughn, line 101: this function name may need to be tweaked. it is returning a new connection every time it is invoked.

imo, having **next** in the name makes it a bit confusing. perhaps `get_connection` or `new_connection`?

lib/core/src/client_connection.cpp
```diff
@@ -24,6 +24,34 @@ namespace irods::experimental
         connect_and_login(_host, _port, _username, _zone);
     }
 
+    client_connection::client_connection(const std::string& _host,
+                                         const int _port,
+                                         const std::string& _proxy_user,
+                                         const std::string& _proxy_zone,
+                                         const std::string& _username,
+                                         const std::string& _zone
```
korydraughn, line 32: `std::string&` -> `std::string_view`

korydraughn (changes requested):
almost there!

server/core/src/irodsReServer.cpp
```diff
@@ -77,6 +78,52 @@ namespace {
         }
     }
 
+    /// Load the available executors from the server's configuration
+    void load_available_executors(){
+        logger::delay_server::info("Picking out available executors");
```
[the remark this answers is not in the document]
trel: this feels like more of a trace - if there is a value printed... i feel that's a bit more helpful for debug.

as for phrasing...  "Loading list of delay rule executors." seems good to me.

server/core/src/irodsReServer.cpp
```diff
@@ -77,6 +78,52 @@ namespace {
...
+            for(const auto &i : hosts){
+                available_hosts.push_back(boost::any_cast<std::string>(i));
+            }
+        }
+        rodsEnv env;
+        _getRodsEnv(env);
+        if(available_hosts.empty()){
+            available_hosts.push_back(env.rodsHost);
+        }
+        logger::delay_server::info("Executors populated");
```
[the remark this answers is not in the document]
trel: "Loading list of delay rule executors - Complete."  ?

server/core/src/irodsReServer.cpp
```diff
@@ -77,6 +78,52 @@ namespace {
...
+    ///
+    /// \param[in] user The name of the user to proxy into
+    ix::client_connection get_next_connection(const std::optional<std::string>& _user){
+        static std::atomic<int> host_index = 0;
+        rodsEnv env{};
+        _getRodsEnv(env);
+        ix::client_connection conn(ix::defer_connection);
+        if(_user.has_value()){
+            int cur = (host_index++) % available_hosts.size();
+            logger::delay_server::info("Sending rule to {}, Using proxy user {} for implicit remote (real user {})", available_hosts[cur], _user.value(), env.rodsUserName);
```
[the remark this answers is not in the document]
trel: this is debug

server/core/src/irodsReServer.cpp
```diff
@@ -77,6 +78,52 @@ namespace {
...
+            // TODO This must proxy the user properly, so that it executes with the right permissions.
+            // Otherwise it doesn't work at all.
+            conn = ix::client_connection(available_hosts[cur],
+                                         env.rodsPort,
+                                         _user.value(),
+                                         env.rodsZone,
+                                         env.rodsUserName,
+                                         env.rodsZone);
+        } else {
+            logger::delay_server::info("Connecting to self");
```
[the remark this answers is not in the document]
trel: i vote trace on this one

server/core/src/irodsReServer.cpp
trel: either seems fine.   or combined... `get_new_connection`?

lib/core/src/irods_configuration_keywords.cpp
trel: delay_rule_executors appears to have been selected.

epsilon-phase:
I think this is ready for a final review pass before rebasing

server/core/src/irodsReServer.cpp
```diff
@@ -77,6 +77,35 @@ namespace {
...
+                                         _user.value(),
+                                         env.rodsZone,
+                                         env.rodsUserName,
+                                         env.rodsZone);
+        } else {
+            logger::delay_server::trace("Connecting to self");
+            return ix::client_connection(env.rodsHost,
+                                         env.rodsPort,
+                                         env.rodsUserName,
+                                         env.rodsZone);
```
korydraughn, line 105: this is equivalent to invoking the default constructor

server/core/src/irodsReServer.cpp
```diff
@@ -77,6 +77,35 @@ namespace {
...
     }
 
+    /// Get the next connection to use for the next rule
+    ///
+    /// \param[in] user The name of the user to proxy into
+    ix::client_connection get_new_connection(const std::optional<std::string>& _user){
+        static std::atomic<int> host_index = 0;
+        static auto &delay_rule_executors = irods::get_advanced_setting<std::vector<boost::any>&>(irods::DELAY_RULE_EXECUTORS_KW);
+        rodsEnv env{};
+        _getRodsEnv(env);
```
korydraughn, line 87: if you use the default constructor for L102, you can move L86,87 into the **true** part of the if-statement

server/core/src/irodsReServer.cpp
```diff
@@ -77,6 +77,35 @@ namespace {
...
+        if(_user.has_value() && !delay_rule_executors.empty() ){
+            int cur = (host_index++) % delay_rule_executors.size();
+            auto &host = boost::any_cast<std::string&>(delay_rule_executors[cur]);
```
korydraughn, line 90: `cur` and `host` can be marked as `const`

lib/core/include/client_connection.hpp
```diff
@@ -98,6 +117,17 @@ namespace irods::experimental
...
                      const std::string_view _zone) -> void;
 
+        /// Connects to the iRODS server with the specified proxy user and zone
+        /// authenticates the user using the information and credentials
+        /// found in the user's irods_environment.json file.
+        ///
+        /// \throws irods::exception If an error occurred.
+        ///
+        /// \param[in] _proxy_user The name of the user to proxy
+        /// \param[in] _proxy_zone The name of the zone to proxy
```
korydraughn, line 127: add `\since 4.3.0` line just like the other functions

lib/core/include/client_connection.hpp
```diff
@@ -52,6 +52,25 @@ namespace irods::experimental
...
+        /// found in the user's irods_environment.json file
+        ///
+        /// \throws irods::exception If an error occured.
+        ///
+        /// \param[in] _host     The host name of the iRODS server.
+        /// \param[in] _port     The port to connect to.
+        /// \param[in] _proxy_user The name of the user to proxy
+        /// \param[in] _proxy_zone The name of the zone to proxy
+        /// \param[in] _username The user to connect as.
+        /// \param[in] _zone     The zone managed by the iRODS server.
```
korydraughn, line 66: add `\since 4.3.0` line just like the other functions

please vertically align the parameter descriptions like the other functions

server/core/src/irodsReServer.cpp
```diff
@@ -541,7 +569,6 @@ int main(int argc, char **argv)
     signal(SIGHUP, signal_exit_handler);
     signal(SIGTERM, signal_exit_handler);
     signal(SIGUSR1, signal_exit_handler);
-
     const auto sleep_time = [] {
```
korydraughn, line 572: lets restore the line above L572 since the code above it is not related to the code that follows it

server/core/src/irodsReServer.cpp
```diff
@@ -453,17 +482,16 @@ namespace {
...
+        
+        ix::client_connection conn = get_new_connection(std::nullopt);
         try {
             rule_exec_submit_inp = fill_rule_exec_submit_inp(conn, rule_id);
         }
         catch (const irods::exception& e) {
             irods::log(e);
             return;
         }
-
```
korydraughn, line 466: lets restore this line since the code above it is not related to the code that follows it

korydraughn (changes requested):
this is looking very good.

korydraughn (approved):

trel (commented):
Looks good - consider the outdated(?) TODO.

Then, we're ready for a # linking to 4429.

epsilon-phase:
Alright, the TODO was indeed outdated, and the issue is linked :)

trel:
awaiting green, and will merge.   good stuff.

Issue #5760, opened 2021-08-03
When iget fails because the open operation returns an error, objects go into stale state.


JustinKyleJames:
- [x] master
- [x] 4-2-stable
--------

## Bug Report

When iget fails because the open operation returns an error, the object goes into stale state.

### iRODS Version, OS and Version

iRODS 4.2.10.  Tested on Ubuntu 16.

### What did you try to do?

Running iget when the open operation returns an error..

In my particular case, I am detecting that the object is in Glacier so the open plugin operation returns an error indicating that it is in Glacier and is being restored.

### Expected behavior

The error will be returned to the client but the object should stay in good state.

### Observed behavior (including steps to reproduce, if applicable)

The object goes to stale state when the open operation returns an error.  From what I have been told, objects should not go into stale state on read only operations.

```
$ ils -l test1
  rods              0 fujifilmresc            6 2021-08-03.14:34 & test1

$ iget test1 - 
remote addresses: 172.17.0.2 ERROR: getUtil: get error for - status = -721000 S3_FILE_IN_GLACIER_ERR
Level 0: [-]    /github/irods_resource_plugin_s3/s3/s3_operations.cpp:304:irods::error irods_s3::restore_s3_object(irods::plugin_context &) :  status [S3_FILE_IN_GLACIER_ERR]  errno [] -- message [Object is in glacier and is currently bei
ng restored.  Try again later.]

$ ils -l test1                                                                                                                                                     
  rods              0 fujifilmresc            6 2021-08-03.14:34 X test1
```

alanking:
I think we have 2 separate issues here:

1. What should we do with a replica that is in the S3 glacier scenario?

I think that this particular behavior does raise a new case for the status of a replica. This was suggested initially to be a replica status stored in the catalog in the `data_is_dirty` column (a la logical locking). After some more discussion, we think it should be sufficient to detect this error in the server as a special case and handle it appropriately.

What the S3 plugin is doing now - returning `S3_FILE_IN_GLACIER_ERR` and adding an appropriate message to the `rError` stack - is what we want to keep doing. The difference for the new approach would be that the name of the error being returned should be made more generic so that it can be reused for equivalent situations in other storage backends (e.g. `REPLICA_IS_BEING_STAGED` or something like that). The server can then detect this error and choose not to mark the replica as stale.

My follow-up question would be what should we do when a user attempts to overwrite the replica? I imagine that when the system goes to open the replica it fails to do so and the replica is then left in whatever state in which it was found.

[...]

PR #5764, opened 2021-08-06
[#5760] Prevent staling replicas on open for read + test_delay_queue resilience (4-2-stable)

alanking:
CI tests passed, federation tests passed, confirmed that `test_delay_queue` is passing both connected to the catalog consumer and provider.

### DIFF
--- a/scripts/irods/test/test_delay_queue.py
+++ b/scripts/irods/test/test_delay_queue.py
@@ -107,14 +107,24 @@ class Test_Delay_Queue(session.make_sessions_mixin([('otherrods', 'rods')], [('a
 
     @unittest.skipIf(plugin_name == 'irods_rule_engine_plugin-python', 'Delete this line on resolution of #4094')
     def test_delay_queue_with_long_job(self):
+        """Test to ensure that a long-running delayed job does not block other jobs from executing.
+
+        Schedules a long-running job, a block of short-running jobs to be run as soon as possible, and a block
+        of short-running jobs to be run at a later time (but before the long-running job completes).
+
+        The delay server is limited to 2 concurrent threads of execution. One of the threads will be occupied
+        with the long-running job while the other will execute the short-running jobs. This demonstrates the
+        delay server's ability to execute multiple jobs per query as well as to pick up new jobs even as a
+        long-running job continues to execute.
+        """
         irodsctl = IrodsController()
         server_config_filename = paths.server_config_path()
 
         delay_job_batch_size = 5
         re_server_sleep_time = 3
-        sooner_delay = 0.1
+        sooner_delay = 0
         later_delay = re_server_sleep_time * 5
-        long_job_run_time = re_server_sleep_time * 10
+        long_job_run_time = later_delay * 10
 
         rule_text_key = 'test_delay_queue_with_long_job'
         parameters = {}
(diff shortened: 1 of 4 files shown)


scripts/irods/test/test_delay_queue.py
```diff
@@ -107,14 +107,25 @@ def test_failed_delay_job(self):
 
     @unittest.skipIf(plugin_name == 'irods_rule_engine_plugin-python', 'Delete this line on resolution of #4094')
     def test_delay_queue_with_long_job(self):
+        """
+        Test to ensure that long-running delayed job does not block other jobs from executing.
```
trel, line 111: that 'a' long-running...

scripts/irods/test/test_delay_queue.py
```diff
@@ -412,13 +455,15 @@ def test_sigpipe_in_delay_server(self):
...
+                lib.delayAssert(self.no_delayed_rules)
 
         finally:
             os.remove(rule_file)
-            self.admin.assert_icommand(['iqdel', '-a'])
+            self.admin.run_icommand(['iqdel', '-a'])
             irodsctl.restart()
 
     @unittest.skipIf(plugin_name == 'irods_rule_engine_plugin-python', 'Delete this line on resolution of #4094')
+    @unittest.skipIf(test.settings.TOPOLOGY_FROM_RESOURCE_SERVER, 'odbc.ini file does not exist on CSC')
```
trel, line 466: we don't use CSC in documentation anywhere... so consider more words here

alanking:
pep257 is attempting to enforce D212 and D213 simultaneously. These are mutually exclusive options. We should either configure this tool to pick one, disable the two, or continue to ignore Codacy. Go here for more information: https://stackoverflow.com/questions/45990301/pep257-d212-and-d213-conflicts

server/api/src/rsDataObjOpen.cpp
```diff
@@ -1019,7 +1021,7 @@ namespace
             }};
 
             if (const int unlock_ec = ill::unlock_and_publish(*rsComm, {replica, admin_op, 0}, ill::restore_status); unlock_ec < 0) {
-                irods::log(LOG_ERROR, fmt::format("[{}:{}] - failed to unlock data object [error_code={}]", __FUNCTION__, __LINE__, unlock_ec));
+                irods::log(LOG_NOTICE, fmt::format("[{}:{}] - failed to unlock data object [error_code={}]", __FUNCTION__, __LINE__, unlock_ec));
```
trel, line 1024: what's the reasoning?   just more visibility?

scripts/irods/test/test_delay_queue.py
alanking: Done

scripts/irods/test/test_delay_queue.py
alanking: Done

server/api/src/rsDataObjOpen.cpp
alanking: After seeing it pop up as a result of a more important overarching error, I didn't feel that it was at the level of ERROR, but perhaps WARNING would be more appropriate. I am willing to be convinced.

trel: so I guess the question is... does it require human eyeballs, that's ERROR...  If it could cause a problem, then WARNING seems right.   If it's merely informative... NOTICE seems good.

alanking: Perhaps ERROR is the best category, then. I'll put it back

trel:
all looks good - please squash cleanly - then we'll # em.

alanking:
Squashed

trel:
yeps, please to #

alanking:
#'d

Issue #5669, opened 2021-05-25
Improve how the buildsystem locates dependencies


SwooshyCueb:
At present, iRODS' buildsystem relies heavily on our irods-externals packages, and getting iRODS to build against equivalent headers/libraries provided by the system can be a bit of a chore.

This was split out from #5082. #5082 was changed to cover the creation of userspace packages for icommands, but I don't want the issue of building without irods-externals to wind up lost. More information about this issue can be found there.

SwooshyCueb:
In order to pass down dependencies from `OBJECT` "library" targets, CMake 3.12 or later is required.

SwooshyCueb:
Note to self: consider using FindPkgConfig for libpam

SwooshyCueb:
Note to self: use find module for ODBC

SwooshyCueb:
I've split out several issues so that this one can now be closed for 4.3.0

Issue #4083, opened 2018-08-28
GCC compatibility


joergsteinkamp:
## Feature: GCC compatibility

Sorry for bothering you again, I know GCC is not on your priority list. However, if the changes made here do not affect the normal builds, I would be happy if you could apply them.

Here is my modified repository: https://github.com/joergsteinkamp/irods/tree/gcc

### Why
1. To reduce the external dependencies from the perspective of a debian system. Debian provides all build tools (autoconf, cmake) and a lot of the external libraries (boost, jansson, jason, libarchive, libs3, ZeroMQ).
2. Since we use Ceph as backend, we can also make use of the debian provided librados2. If iRODS and boost are compiled with clang and linked to the debian provided librados2, which is linked to debian's libboost, a runtime error was caused, when saving large files in the Ceph resource.

### What
Two kind of error occured during compilation:

1. When C-structs are initialize with curly brackets, the order must be equal to the order in the struct definition. Affected files:
   * lib/core/src/packStruct.cpp
   * plugins/resources/structfile/libstructfile.cpp

2. GCC is more strict than LLVM with respect to const and non-const function parameter definition and their respective call. Affected files.
   * plugins/resources/compound/libcompound.cpp
   * plugins/resources/deferred/libdeferred.cpp
   * plugins/resources/load_balanced/libload_balanced.cpp
   * plugins/resources/nonblocking/libnonblocking.cpp
   * plugins/resources/passthru/libpassthru.cpp
   * plugins/resources/random/librandom.cpp
   * plugins/resources/replication/librepl.cpp
   * plugins/resources/roundrobin/libroundrobin.cpp
   * plugins/resources/structfile/libstructfile.cpp
   * plugins/resources/unixfilesystem/libunixfilesystem.cpp
   * server/core/src/irods_resource_plugin_impostor.cpp

From my understanding on how lax clang handles those "errors" my changes should make no difference for clang compiled version even if you remove the preprocessor statements.

### iRODS Version, OS and Version

iRODS 4.3.0
Debian 9 (stretch)

PR #5769, opened 2021-08-11
[#4083] First round of changes for GCC compatibility

SwooshyCueb:
This PR is in service of #4083.

This PR contains a couple of commits in service of #5669:
1. Logging typo fixed in `CheckCPPMacroDefinition.cmake`
2. `IRODS_EXTERNALS_PACKAGE_ROOT`,  `IRODS_MACRO_CHECK_DEPENDENCY_SET_FULLPATH`, and `IRODS_MACRO_CHECK_DEPENDENCY_SET_FULLPATH_ADD_TO_IRODS_PACKAGE_DEPENDENCIES_LIST` definitions moved to new CMake module `IrodsExternals.cmake`.

This PR introduces initial support for building with GCC and `libstdc++`.
**Caveat:** When building against the current set of irods-externals packages, `libc++` must still be used, as `libstdc++` and `libc++` are not ABI compatible.

Some new CMake options have been introduced:
| Variable                     | Type   | Default                                                              | Description |
| :--------------------------- | :----- | :------------------------------------------------------------------- | :---------- |
| `IRODS_BUILD_WITH_CLANG`     | `BOOL` | `TRUE`                                                               | If `TRUE`, CMake will attempt to set `CMAKE_C_COMPILER` and `CMAKE_CXX_COMPILER` to Clang
| `IRODS_BUILD_WITH_WERROR`    | `BOOL` | `TRUE` if the C++ compiler is detected as a Clang, `FALSE` otherwise | If `TRUE`, CMake will pass `-Werror` to the compiler
| `IRODS_BUILD_AGAINST_LIBCXX` | `BOOL` | `TRUE` if the C++ compiler is detected as a Clang, `FALSE` otherwise | If `TRUE`, CMake will attempt to build against `libc++`

Both `IRODS_BUILD_WITH_CLANG` and `IRODS_BUILD_AGAINST_LIBCXX` are in modules included in the `irods-dev` package and are propagated foward in the installed CMake configuration file.  
Don't get too attached to `IRODS_BUILD_WITH_CLANG`, as we will eventually move away from having the buildsystem set `CMAKE_<LANG>_COMPILER` for us.

Building with GCC against `libc++` is supported, but with a few caveats:
- If using irods-externals for `libc++`, the clang package is still required, as it includes the headers.
- Since CMake cannot be made to use the `gcc` frontend for linking without changing the link language variable for the entire binary, `g++` must be explicitly told not to link to `libstdc++`. The least impactful flag to do this is `-nodefaultlibs`, which still overreaches significantly.
    - `libc` and friends are excluded from linking. This is worked around by using `find_library` to locate `libc` and `libm`. The results are added to `CMAKE_CXX_STANDARD_LIBRARIES`. If one or both are not found, `-c` and `-m`, respectively, are added instead. `find_library` provides CMake variables that can be set if alternate/static libraries should be used instead.
    - `libgcc` is excluded from linking. This is worked around in a similar fashion as `libc` and friends. However, this causes flags like `-static-libgcc` to become non-ops, so the developer must pay close attention to CMake configuration to ensure that the proper `libgcc` is used.
    - `-pthread` becomes a non-op during the linking stage. This is worked around by explicitly setting `THREADS_PREFER_PTHREAD_FLAG` to false, as the `-pthread` flag has not been observed to have any impact at compile-time for iRODS.

Other buildsystem changes:
- `-fpermissive` is passed if the C++ compiler is detected as GCC.
- The `Wparentheses` warning is disabled completely if the C++ compiler is detected as GCC.
- The `Wunused-variable` and `Wunused-but-set-variable` warnings are set to not generate errors in `Werror` mode if the C++ compiler is detected as GCC.
- The method for working around the `libc++` 11+ bug where references to `std::allocator<void>` causes deprecation warnings with C++17 has been changed from disabling all libc++ deprecation warnings to setting deprecation warnings to not generate errors.

Due to changes in `UseLibCXX.cmake`, this change is potentially backward incompatible, so companion pull requests may be required in some other repos.

One notable code change is the relocation of `dstream.hpp`'s `mandatory_openmode`. Due to a [known bug in GCC](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=90031), explicitly specialized template variables inside class scopes generate parsing errors at compile time. This is solved by moving `mandatory_openmode` and its explicit specializations outside the `basic_dstream` class, to the same scope as `default_openmode`.


trel:
making some progress.      cc/ @joergsteinkamp

lib/core/include/dstream.hpp
```diff
@@ -435,6 +435,15 @@ namespace irods::experimental::io
     template <typename T>
     inline static constexpr std::ios_base::openmode default_openmode<std::basic_iostream<T>> = std::ios_base::in | std::ios_base::out;
 
+    template <typename>
```
alanking, line 438: I remember us discussing this particular change... did we determine that this is more correct, or should we leave an TODO comment which explains that this may or may not move back inside the `basic_dstream` class?

cmake/Modules/UseLibCXX.cmake
```diff
@@ -7,75 +7,99 @@ Define CMAKE_CXX_STANDARD before including this module.
...
+    # and the two-argument version of std::allocator::allocate
+    # Starting with 11.0, libc++ enforces these deprecations, which causes
+    # warnings from the boost headers.
+    # Boost has removed references to the two-argument version of
+    # std::allocator::allocate in 1.75, which would solve the problem, but,
+    # while the template specification deprecation is not supposed to affect
+    # existing code, libc++ 11 causes warnings for references to
+    # std::allocator<void>. Maybe a future release of libc++ will fix this.
+    # Should we ever switch to C++20 (or newer), std::allocator<void> problem
+    # will go away, as the explicit template specialization is removed in C++20.
```
alanking, line 102: Would it be worth filing an issue for this and "linking" it here?

cmake/Modules/CheckCPPMacroDefinition.cmake
```diff
@@ -83,7 +83,7 @@ function(__check_cpp_macro_def_impl macro var language)
     endif()
     file(READ ${src} content)
     file(APPEND ${CMAKE_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/CMakeError.log
-      "Determining definition of ${language} preprocessor macro ${macro} passed with the following compile output:\n${output}\n${src}:\n${content}\n\n")
+      "Determining definition of ${language} preprocessor macro ${macro} failed with the following compile output:\n${output}\n${src}:\n${content}\n\n")
```
alanking, line 86: lol

alanking (commented):
Just a couple of minor questions. Overall looks good. I'm into the template deduction changes :)

lib/core/include/dstream.hpp
SwooshyCueb: I don't think we made a decision one way or the other. Personally, I'm fine with leaving it outside the class, but I don't have a strong opinion on the matter.

cmake/Modules/UseLibCXX.cmake
SwooshyCueb: I don't know if it's worth having another issue open, but I suppose it couldn't hurt. I think it would make more sense to link it to 5317, though, as that's where this was originally addressed.

SwooshyCueb:
Gonna go ahead and mark this as ready for a review

lib/core/include/dstream.hpp
alanking: Thoughts, @korydraughn?

cmake/Modules/UseLibCXX.cmake
alanking: Okay, so long as the work is captured in an issue somewhere, I don't think we really need to refer to it in the comments here. Thanks

lib/core/include/dstream.hpp
korydraughn: I don't like them being public given that they are an implementation detail, but it is fine for now. We can always come back and research this later.

@SwooshyCueb What was the reason for this again? It was because of a GCC compilation issue, right?

korydraughn: Never mind, I just read your note about the GCC bug.

@alanking So, we can restore the location of the templates once GCC resolves the issue.

plugins/auth/native/libnative.cpp
```diff
@@ -631,32 +631,32 @@ irods::auth* plugin_factory(
         AUTH_ESTABLISH_CONTEXT,
         function<error(plugin_context&)>(
             native_auth_establish_context ) );
-    nat->add_operation<rcComm_t*,const char*>(
+    nat->add_operation(
```
korydraughn, line 634: Is this due to a bug in GCC?

CMakeLists.txt
```diff
@@ -86,28 +64,42 @@ set(CMAKE_SKIP_BUILD_RPATH FALSE)
...
 set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
-add_compile_options(-Wall -Wextra -Werror -Wno-unused-function -Wno-unused-parameter)
 
 if (IRODS_DISABLE_COMPILER_OPTIMIZATIONS)
-    add_compile_options(-O0)
+  add_compile_options(-O0)
+endif()
+
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+  add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-fpermissive>)
```
korydraughn, line 73: I'm guessing this is because of `with_durability`?

korydraughn (commented):
Nice work!

lib/core/include/dstream.hpp
SwooshyCueb: So let me make sure we're all on the same page here.  

Considering one of the goals is to not have to carry around our own compiler, if the decision is to make this scope/visibility change temporary until the parsing bug in GCC is fixed, we're going to be waiting a while before we can revert this change.

This workaround will stick around until:
1. GCC has implemented a fix for their parsing bug
2. The fix for this parsing bug has made it to distribution package repositories
3. We have dropped support for distributions that do not have a GCC with the fix in their repositories

All this taken into account...  
Should we proceed with the intention of reverting this fix however many years in the future?  
Should we consider making this workaround more permanent?  
Should we perhaps instead investigate a different solution?

lib/core/include/with_durability.hpp
```diff
@@ -49,6 +49,43 @@ namespace irods::experimental
...
+        std::chrono::milliseconds delay_;
+        float delay_multiplier_;
+
+        template <typename T>
+        constexpr auto throw_if_less_than_zero(T v) -> void
+        {
+            if (v < 0) {
+                throw std::invalid_argument{"Value must be greater than or equal to zero."};
+            }
+        }
```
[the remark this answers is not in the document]
SwooshyCueb: Some version of some compiler (I do not remember which version or which compiler, but it was a reasonable one that we might want to support) complained about not being able to find things in the private block. I figured it made more sense to move it than to add forward declarations.

CMakeLists.txt
SwooshyCueb: This is due to a number of things. I'm going to be opening a separate issue about it.

plugins/auth/native/libnative.cpp
SwooshyCueb: I couldn't tell you, as this one is impossible to google. Here's a reduced demonstration, though: https://godbolt.org/z/hr8Ysdq7E

lib/core/include/with_durability.hpp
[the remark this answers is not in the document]
SwooshyCueb: It's probably worth noting that this compiler only required moving the private block *after* `throw_if_less_than_zero` was made `constexpr`. No idea what's up with that.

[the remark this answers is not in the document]
korydraughn: Cool cool. You can mark this comment as resolved whenever you're ready.

CMakeLists.txt
korydraughn: Got it.

cmake/Modules/UseLibCXX.cmake
SwooshyCueb: Looking at it again, I think 5317 captures the issue enough that we shouldn't need to open another, so I'll mark this as resolved.

lib/core/include/dstream.hpp
SwooshyCueb: Per our previous discussion, I've moved `mandatory_openmode` into a detail namespace, so I think we can mark this as resolved

trel:
esteemed committee is pleased - please # all these.

SwooshyCueb:
PR description and title #'d, but I'm going to hold off on #\'ing the commits until 5643 is merged and I've had a chance to make sure that further changes aren't needed.

SwooshyCueb:
\#'d, ready for merge

PR #5772, opened 2021-08-13
Multiple little fixes (4-2-stable)

korydraughn:


lib/core/include/lsUtil.h
```diff
@@ -1,66 +1,82 @@
...
-lsUtil( rcComm_t *conn, rodsEnv *myRodsEnv, rodsArguments_t *myRodsArgs,
-        rodsPathInp_t *rodsPathInp );
-int
-lsDataObjUtil( rcComm_t *conn, rodsPath_t *srcPath,
-               rodsArguments_t *rodsArgs, genQueryInp_t *genQueryInp );
-int
-printLsStrShort( char *srcPath );
-int
-lsDataObjUtilLong( rcComm_t *conn, char *srcPath,
-                   rodsArguments_t *rodsArgs, genQueryInp_t *genQueryInp );
```
alanking, line 25: I realize that this file is pretty specific to the iCommand `ils`, but is it bad to change the signature of an exposed function like this when it is not in the `irods::experimental` namespace? For this case it might not matter too much, just wondering how you felt about that in general.

scripts/irods/test/test_iadmin.py
```diff
@@ -1456,6 +1456,69 @@ def test_print_full_username__issue_3170(self):
...
+
+            # While the issue was encountered via an empty string, we don't have any tools that
+            # allow us to create the empty string scenario. Instead, we know the empty string scenario
+            # is not any different then the data mode containing a non-integer value. Therefore, using
+            # an invalid string for the data mode satisfies the test.
+            for invalid_data_mode in [' ', '-', 'X']:
+                # Set the source replica's data mode to an invalid value.
+                self.admin.assert_icommand(['iadmin', 'modrepl', 'logical_path', data_object, 'replica_number', '0', 'DATA_MODE', invalid_data_mode])
+
+                # Create work for the rebalance operation by setting the second replica's status to stale.
```
alanking, line 1492: Would it be worth adding a section to this test to exercise the "create missing replicas" codepath? I realize the change occurs in a common section, but it could be worthwhile for completeness.

alanking (commented):
Overall looks good

lib/core/include/lsUtil.h
korydraughn: Yes, it is bad to do what I've done here. My first thought is that nothing except the icommands should be using these functions, but because they are already public, we can't enforce that. Therefore, it would be better to mark the old version of the functions in question as deprecated and add a new function that presents the new signature. The old function can be implemented in terms of the new one.

scripts/irods/test/test_iadmin.py
korydraughn: Sure. Adding that case will protect us from breaking things in the future.

trel (commented):
nothing additional from me

scripts/irods/test/test_iadmin.py
korydraughn: done.

lib/core/include/lsUtil.h
korydraughn: I think I've restored the signature. Instead of marking things as deprecated, I chose to keep the API as they were and just ignore unused arguments. We can clean these up in 4.3.0+.

trel:
once tests are confirmed happy - please # - and we can merge this chunk and get them into master too.

korydraughn:
Passed CI - Ubuntu 18 / Centos 7

Added #

Issue #5759, opened 2021-07-30
Creating a new data object while connected to CSC fails with apostrophe in DATA_NAME


alanking:
### iRODS Version, OS and Version
4.2.9+

### What did you try to do?
While connected to CSC:
```
$ echo yikes | istream write goo\'
```

### Expected behavior
Success

### Observed behavior (including steps to reproduce, if applicable)
```
$ echo yikes | istream write goo\'
Error: Cannot open data object.
```

This appears in the log on the connected CSC:
```
Jul 30 13:21:39 pid:446 NOTICE: [create_new_replica:417] - no replica found [path=[/tempZone/home/rods/goo'], hier=[resource1Resource]]
```
This message appears if the replica which was just registered in the catalog cannot be found via a GenQuery to obtain system metadata about said replica which are generated at registration time in the database plugin. This query is necessary because the registration API provides no way of communicating this replica information back over the wire.

The following can be seen in the CSP logs when `spLogLevel=1` is set:
```
Jul 30 13:22:06 pid:2655 LOG_SQL: pid: 2655 sql: bindVar[1]=/tempZone/home/rods
Jul 30 13:22:06 pid:2655 LOG_SQL: pid: 2655 sql: bindVar[2]=goo'' and DATA_RESC_HIER = 'resource1Resource
Jul 30 13:22:06 pid:2655 LOG_SQL: pid: 2655 sql: select distinct R_DATA_MAIN.data_owner_name ,R_DATA_MAIN.data_owner_zone ,R_DATA_MAIN.create_ts ,R_DATA_MAIN.modify_ts ,R_DATA_MAIN.data_expiry_ts ,R_DATA_MAIN.data_id  from  R_DATA_MAIN , R_COLL_MAIN  where R_COLL_MAIN.coll_name = ?  AND R_DATA_MAIN.data_name = ?  AND R_COLL_MAIN.coll_id = R_DATA_MAIN.coll_id
Jul 30 13:22:06 pid:2655 LOG_SQL: pid: 2655 result: SUCCESS
```

The second bindVar is obviously malformed due to the unescaped apostrophe.

alanking:
#5727 is the root cause, but this can be avoided by putting the `DATA_NAME` condition of the query last.

trel:
Option 2 seems best... new api plugin for 4.2.11 to be called in `rsDataObjOpen`
 - one fewer query to the database
 - works consistently on the consumer

we still need to fix the problematic parser, but that's a separate issue.

[...]

PR #5773, opened 2021-08-13
[#5759] Prevent queries in replica creation/finalize (4-2-stable)

alanking:
Tried to split up the commits to make it a bit more manageable but it's still a big change. The bulk of it is moving the contents of `rsPhyPathReg` into the new API plugin and adding some documentation, so there are only a few critical areas to look at.

CI tests passed, federation passed, topology-from-provider passed, topology-from-consumer running


scripts/irods/test/test_iput.py
```diff
@@ -513,6 +513,48 @@ def do_test_issue_5496_impl(self, file_size_in_bytes):
...
+
+        for coll in collection_names:
+            collection_path = os.path.join(self.user.session_collection, coll)
+
+            self.user.assert_icommand(['imkdir', collection_path])
+
+            for name in data_names:
+                logical_path = os.path.join(collection_path, name)
+
+                self.user.assert_icommand(['iput', local_file, logical_path])
```
trel, line 552: does `ireg` need the same exercising?

server/api/include/rs_data_object_finalize.hpp
```diff
@@ -13,12 +13,34 @@ extern "C" {
 
 /// \brief Set the state of every replica for a data object in the catalog, atomically.
 ///
-/// A complete set of the columns in their current state as well as the desired modifications
-/// is required in the form of a JSON structure.
+/// \parblock
+/// The replicas field is a complete set of the columns in their current state as well as the
+/// desired modifications is required in the form of a JSON structure. Each replica can also
```
trel, line 18: doesn't quite parse as english.

server/api/src/rsDataObjOpen.cpp
```diff
@@ -327,27 +338,94 @@ namespace
...
+                    "[{}:{}] - failed to register physical path "
                     "[error_code=[{}], path=[{}], hierarchy=[{}]",
                     __FUNCTION__, __LINE__, ec, _inp.objPath, hierarchy));
 
                 if (object_locked) {
                     if (const int unlock_ec = ill::unlock_and_publish(_comm, {_existing_replica_list->dataId, admin_op}, ill::restore_status); unlock_ec < 0) {
                         irods::log(LOG_ERROR, fmt::format("[{}:{}] - failed to unlock data object [error_code={}]", __FUNCTION__, __LINE__, unlock_ec));
                     }
+
+                    if (rst::contains(_existing_replica_list->dataId)) {
```
trel, line 362: can we get a comment here?

server/api/src/rs_register_physical_path.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+
+    if (out && out->buf) {
+        *_json_output = static_cast<char*>(out->buf);
+    }
+    std::free(out);
+    out = nullptr;
+
+    return rc;
+} // rs_data_object_finalize
```
trel, line 31: wrong name

server/core/include/irods_file_object.hpp
```diff
@@ -197,6 +199,22 @@ namespace irods {
...
     auto file_object_factory(RsComm& _comm, const rodsLong_t _data_id) -> irods::file_object_ptr;
 
+    /// \brief Factory function which takes a JSON array and creates a file_object
+    ///
+    /// \parblock
+    /// Functions very similarly to file_object_factory(RsComm&, const rodsLong_t) after the
+    /// data_object_proxy_t has been generated.
+    /// \endparblock
+    ///
+    /// \param[in] _comm
```
trel, line 209: moar

server/core/src/physPath.cpp
```diff
@@ -429,7 +429,7 @@ getchkPathPerm( rsComm_t *rsComm, dataObjInp_t *dataObjInp,
     }
 
 #if 1
-    // Possibly need to revisit this -- this is wrong
+    // TODO: this is wrong
```
trel, line 432: why?   create / link to an issue?

unit_tests/src/test_data_object_finalize.cpp
```diff
@@ -111,6 +111,8 @@ TEST_CASE("finalize", "[finalize]")
             // Get data object info, modify some fields in each replica, and stamp the catalog
             json input;
 
+            input["logical_path"] = og_op.logical_path();
```
trel, line 114: og_op, pretty funny.

trel (commented):
easier to follow than i expected when opening this tab.     good stuff.

unit_tests/src/test_data_object_finalize.cpp
alanking: I do what I can

server/core/src/physPath.cpp
alanking: Created #5775 for this

server/core/include/irods_file_object.hpp
alanking: Done

server/api/src/rs_register_physical_path.cpp
alanking: Done

server/api/src/rsDataObjOpen.cpp
alanking: Done

server/api/include/rs_data_object_finalize.hpp
alanking: Hmmm... I just copied and pasted this from `lib/api/include/data_object_finalize.hpp`. Not sure how that happened. Fixed.

scripts/irods/test/test_iput.py
alanking: I started to write tests for several iCommands (`irepl`, `icp`, etc.) but backed them out since I was just testing the apostrophe bit (didn't seem important to me at the time to try all the different iCommands). I can add these back as well as one for `ireg`.

alanking: I also noticed while contemplating this test that this does not cover all the permutations for `icp` in particular because there will be two data objects to consider, not just one: source and destination. So that test will need to cover an additional set of scenarios for the source data object.

alanking:
Addressed review comments except for adding the extra tests. I addressed all of the Codacy issues except for:

1. The docstring issue (this is a configuration issue in Codacy that needs to be resolved)
2. The `irods::at_scope_exit` being unused issue, which is just the nature of the object

Next push will contain new tests and then I'll report test results

alanking:
Added tests for `icp`, `ireg`, `irepl`, `iphymv`, and `imv`. Confirmed everything is working in topology.

scripts/irods/test/test_iput.py
alanking: Done

server/core/src/server_utilities.cpp
```diff
@@ -50,7 +50,7 @@ namespace irods
         const auto buf_size = _s.length() + 1;
 
         auto* buf = static_cast<char*>(allocate(sizeof(char) * buf_size));
-        std::strncpy(buf, _s.data(), _s.length());
+        std::snprintf(buf, _s.length(), "%s", _s.data());
```
korydraughn: This will truncate the string being copied by one byte. Replacing `_s.length()` with `buf_size` should take care of it.

server/core/src/server_utilities.cpp
```diff
@@ -39,5 +39,24 @@ namespace irods
...
+    {
+        constexpr auto allocate = [](const auto bytes) noexcept
+        {
+            return std::memset(std::malloc(bytes), 0, bytes);
+        };
+
+        const auto buf_size = _s.length() + 1;
+
+        auto* buf = static_cast<char*>(allocate(sizeof(char) * buf_size));
+        std::snprintf(buf, buf_size, "%s", _s.data());
```
korydraughn, line 53: What is the reasoning behind using `snprintf` over `strncpy` here?

korydraughn (commented):
Looking good

server/core/src/server_utilities.cpp
korydraughn: Ignore this one. I was doing the review one commit at a time.

plugins/api/src/data_object_finalize.cpp
```diff
@@ -89,6 +89,19 @@ namespace
...
+        json after;
+
+        for (const auto& i : _json_list) {
+            before.push_back(i.at("before"));
+            after.push_back(i.at("after"));
+        }
+
+        return {before, after};
+    } // split_before_and_after_info
+
```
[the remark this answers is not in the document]
alanking: I'll look into it. I think I will have to change some interfaces and add a new factory method for `data_object_proxy` as the existing interfaces operate on `nlohmann::json::array`s.

server/core/src/server_utilities.cpp
alanking: Yep, fixed.

server/core/src/server_utilities.cpp
alanking: Codacy complained via the FlawFinder tool which states: "Easily used incorrectly; doesn't always \0-terminate or check for invalid pointers [MS-banned] (CWE-120)." `std::snprintf` seems to be considered a safe alternative.

korydraughn: That makes sense. However, I bet this likely incurs more overhead. Maybe codacy would feel better about `std::memcpy`. In this case, we don't have to pay for the additional overhead given `buf` is cleared to zeros which gives us the null terminator for free.

korydraughn: Assuming my testing is correct, I'm showing strncpy as being 4.4x faster than snprintf and memcpy being 4.3x faster than snprintf. Here is the test: https://quick-bench.com/q/axHaBt3rn3B5FIX_h5QSexwCnBM

korydraughn: The only problem I see with `strncpy` is that it stops at null bytes, but given that `to_bytes_buffer` takes in a `string_view`, it is probably safe to say that is fine.

alanking: Cool, I don't mind using either one, really. I will try `std::memcpy` and see how that goes.

alanking: Codacy has deemed `std::memcpy` acceptable. I am okay to revert this to use `std::strncpy` as well. If we go that route, I may leave a comment in defense of its use for future readers of the Codacy issues.

korydraughn: Cool. It's your call.

alanking: Okay. I think I'll just revert it and leave an explanatory comment. Gracias

plugins/api/src/data_object_finalize.cpp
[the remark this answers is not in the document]
alanking: I managed to implement a version of this in which before and after are `std::vector<const nlohmann::json*>`s and it seems to be working for normal use cases. I'll run tests shortly

scripts/irods/test/test_ichmod.py
```diff
@@ -1,28 +1,74 @@
...
+
+        logical_path = os.path.join(self.collection, 'file')
+
+        # Put the data object into the inheritance structure and ensure that the user given
+        # permission to read is able to get the data object.
+        self.admin.assert_icommand(['iput', filepath, logical_path])
+        self.admin.assert_icommand(['ils', '-LA', self.collection], 'STDOUT', 'file')
+        self.user.assert_icommand(['iget', logical_path, os.path.join(self.user.local_session_dir, '')])
+
+    def test_collection_acl_inheritance_for_icp__issue_3032(self):
```
trel, line 59: please consider `imv` into the inheritance structure as well.

alanking: Done

alanking: I take it back. `imv` does not work in this scenario. That's what I get for trying to go fast

alanking:
Squashed. Running tests again

alanking:
Core tests, unit tests, timing tests, and federation tests passed

trel:
so good.    please to # when ready.

alanking:
#'d

alanking:
Just noticed that I forgot to include the new headers in the development library. I can squash those before merging

trel:
yeah, let's squash it into the right place - not pressed for time on this one.

PR #5782, opened 2021-08-20
[#5300] Server now creates PID file (master)

korydraughn:
This PR attempts to keep the irodsctl script from killing iRODS processes inside of a container by introducing a PID file.

With the PID file, the irodsctl script can confidently determine if a process is a descendant or not.

### DIFF
--- a/server/core/src/rodsServer.cpp
+++ b/server/core/src/rodsServer.cpp
@@ -34,6 +34,8 @@
 #include <sys/un.h>
 #include <netinet/in.h>
 #include <arpa/inet.h>
+#include <fcntl.h>
+#include <sys/stat.h>
 
 #include <boost/filesystem.hpp>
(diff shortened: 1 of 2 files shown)


server/core/src/rodsServer.cpp
```diff
@@ -238,7 +240,65 @@ namespace
...
+        }
+
+        // Get the current open flags for the open file descriptor.
+        const auto flags = fcntl(fd, F_GETFD);
+        if (flags == -1) {
+            ix::log::server::error("Could not retrieve open flags for PID file.");
+            return -1;
+        }
+
+        // Enable the FD_CLOEXEC option for the the open file descriptor.
```
trel, line 264: "the the"

korydraughn:
tested at the bench.

added #

PR #5783, opened 2021-08-20
Fix codacy issues

epsilon-phase:


plugins/resources/replication/irods_create_write_replicator.cpp
```diff
@@ -9,10 +9,10 @@ namespace irods {
...
         const std::string& _root_resource,
         const std::string& _current_resource,
-        const std::string& _child ) {
-        root_resource_ = _root_resource;
-        current_resource_ = _current_resource;
-        child_ = _child;
+        const std::string& _child )
+      : root_resource_(_root_resource),
+        current_resource_(_current_resource),
+        child_(_child){
```
alanking, line 15: We have generally been using list-(braced-?)initialization in our constructors by convention. Any reason in particular to switch to parentheses? I've seen some strong opinions around "universal initialization" so I think we are willing to listen to a different view on things here. Same comment for all the other spots this is happening

plugins/resources/replication/irods_repl_rebalance.cpp
```diff
@@ -299,7 +299,7 @@ namespace {
         const std::string&               _parent_resc_name,
         const std::string&               _child_resc_name,
         const size_t                     _bun_idx,
-        const std::vector<leaf_bundle_t> _bundles,
+        const std::vector<leaf_bundle_t>&_bundles,
```
alanking, line 302: Can you add a space here? I realize it throws off the alignment :/

server/re/include/irods_re_plugin.hpp
```diff
@@ -307,8 +307,8 @@ namespace irods {
 
         template<typename OP, typename... As >
         error call(
-            std::string _instance_name,
-            std::string _operation_name,
+            const std::string& _instance_name,
+            const std::string& _operation_name,
```
alanking, line 311: I don't *think* this will be an issue so there is probably not an action item here, but we may want to make sure the signature change does not break the existing rule engine plugins (or, if it does, that they are updated to reflect it). I am recalling an instance where the audit REP was surprisingly broken due to an undetected change in a framework-related signature: https://github.com/irods/irods_rule_engine_plugin_audit_amqp/commit/264fbfed26a64b7775d8d3f731df92533d37a084 However, I don't see `call` being invoked anywhere in there, so it may be fine.

@korydraughn do you know if REPs might be impacted by this?

alanking (commented):
Nice work overall. Couple of small things which may warrant further discussion

server/re/include/irods_re_plugin.hpp
alanking: Actually, after stepping away for 30 seconds, I don't think this would affect anything as we are *adding* a `const` qualifier. Please ignore :)

plugins/resources/replication/irods_create_write_replicator.cpp
epsilon-phase: I hadn't honestly thought to use it while writing the constructor

korydraughn (approved):
Looks fine to me.

trel:
awesome - please squash to a single commit(?) and point at 5781 - and we'll add the # once it's all nice and clean.

PR #5790, opened 2021-08-23
Implemented new library: scoped_permission (master)

korydraughn:
Allows operations to temporarily grant permissions on a collection or data
object for the duration of a scoped block.


korydraughn:
added #

Issue #5602, opened 2021-05-05
Add support for optional server configuration service endpoint


jasoncoposky:
## Feature

In support for the creation of a cloud-native iRODS implementation, the server configuration needs to be moved to an external service for access on agent initialization.  An environment variable `IRODS_SERVER_CONFIGURATION_ENDPOINT` will direct the agent process to an endpoint for all server configuration.  Should this environment variable not be present, the current server initialization process will ensue.  A server configuration [REST API](https://github.com/irods/irods_client_rest_cpp/issues/7) has been created to provide the necessary configuration as an array of JSON objects, one for each configuration file.

PR #5808, opened 2021-08-25
[#5602] Fix build for unit tests (master)

alanking:
Unit tests were not building due to missing include dependency for the
fmt library which was introduced when the fmt library was added as a
dependency to irods_server_properties.hpp.


trel:
ah, very good - sorry about that.

trel:
yes please # once confirmed to be good

alanking:
Unit tests passed. #'d

PR #5819, opened 2021-08-30
[5786] Add tests for itree

epsilon-phase:


scripts/irods/test/test_itree.py
```diff
@@ -0,0 +1,71 @@
...
+                 unittest.TestCase):
+
+    def setUp(self):
+        super(Test_ITree, self).setUp()
+        self.admin = self.admin_sessions[0]
+
+    def tearDown(self):
+        super(Test_ITree, self).tearDown()
+
+    @unittest.skipIf(test.settings.RUN_IN_TOPOLOGY, "Skip for Topology Testing")
```
trel, line 24: i don't know that any of these should behave differently in a topology configuration... do we need this skip? or the other three below?

trel (commented):
need to also add test_itree to core_tests_list.json so the test suite runs it automatically

scripts/irods/test/test_itree.py
epsilon-phase: Not really, but it's what the other icommand tests seem to do.

scripts/core_tests_list.json
```diff
@@ -35,6 +35,7 @@
     "test_irule",
     "test_iscan",
     "test_istream",
+    "test_itree",
```
trel, line 38: nearly alphabetical order.

trel:
these look good enough to squash and get the single commit message ready.    hold on the # sign for now.

scripts/irods/test/test_itree.py
alanking: Not to beat the dead horse, but it looks to me like this would behave fine in a topology test setup.

scripts/irods/test/test_itree.py
```diff
@@ -0,0 +1,68 @@
...
+    def test_counting(self):
+        """Can Itree count objects properly?"""
+        self.admin.assert_icommand("itree", 'STDOUT_SINGLELINE', "Found 0 collections and 0 data objects")
+        for i in range(10):
+            self.admin.assert_icommand("imkdir {}".format(i))
```
alanking, line 29: For idempotency, can we make the test collections as sub-collections of `self.admin.session_collection`? That collection gets cleaned up automatically on `tearDown()`. Maybe `setUp()` can `icd` to the session collection and then `icd` back to the original `ipwd` on `tearDown()` and that's all that would need to change

Otherwise, it looks like these will be left in the catalog after the test completes unless I'm missing something.

alanking (commented):
Just one thing I noticed but otherwise looks good

scripts/irods/test/test_itree.py
trel: oh, missed that - yes please remove this one too.

scripts/irods/test/test_itree.py
epsilon-phase: I will add that to `setUp` then

trel: this appears to be done now...

alanking (approved):

trel:
@epsilon-phase can you confirm these are now passing and are idempotent?

if so, i think we're ready for the #.

epsilon-phase:
Everything is looking good now after a bit of fiddling

trel:
let's do it - please # and we'll get it merged.

PR #5820, opened 2021-08-30
[#4083] Install IrodsCXXCompiler CMake module and fix rsPhyPathReg.cpp includes

SwooshyCueb:
In service of #4083

In libstdc++ 11, `#include <memory>` does not implicitly `#include <cstdlib>`, where `std::free( void* )` is found.
Also, I forgot to mark the `IrodsCXXCompiler` CMake module for installation.


trel:
looks good to me - please # when ready

SwooshyCueb:
\#'d

PR #5829, opened 2021-09-06
[#858] Fix typos. The file says that it is generated code, but I could find the files which were used as source. Please update the source files.

RubenGarcia:
The file says that it is generated code, but I could find the files which were used as source. Please update the source files.

The lib/api/README.txt file indicates "All functions are automatically generated by the apiGen script in the scripts directory."
But I could not find such a script in the respository.

### DIFF
--- a/lib/api/src/rcBulkDataObjPut.cpp
+++ b/lib/api/src/rcBulkDataObjPut.cpp
@@ -113,9 +113,9 @@
  *         are:
  *    \n COL_DATA_NAME - the full paths of each data object.
  *    \n COL_DATA_MODE - the createMode of each data object.
- *    \n OFFSET_INX - The accummulative size of the data objects. e.g.,
+ *    \n OFFSET_INX - The accummulative size of the data objects. E.g.,
  *          suppose we are bulk loading 2 files with sizes s1 and s2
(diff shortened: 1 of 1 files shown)


RubenGarcia:
It would probably be good to update lib/api/README.txt with the location of the mentioned script.

lib/api/src/rcBulkDataObjPut.cpp
```diff
@@ -113,9 +113,9 @@
  *         are:
  *    \n COL_DATA_NAME - the full paths of each data object.
  *    \n COL_DATA_MODE - the createMode of each data object.
- *    \n OFFSET_INX - The accummulative size of the data objects. e.g.,
+ *    \n OFFSET_INX - The accummulative size of the data objects. E.g.,
```
trel, line 116: "accummulative" -> "cumulative"

trel (commented):
please put "[#858] " at the front of your commit message and we'll get it merged.

thanks!

RubenGarcia:
Added your last change request and changed the commit message as requested.

trel:
oops, didn't notice it was two commits.   we goofed.   we'll do better next time.

Issue #5771, opened 2021-08-13
itouch -R should give better error when non-leaf resource is named


korydraughn:
- [x] master
- [x] 4-2-stable
---
### iRODS Version, OS and Version
4.2.10 (since support for `-R` was added)

### What did you try to do?
I have a resource hierarchy and I tried to create a new data object targeting the root resource of the hierarchy.
```bash
$ itouch -R pt0 foo
```
### Expected behavior
I expected a new data object to appear on one of the leaf resources.

### Observed behavior (including steps to reproduce, if applicable)
```bash
Aug 13 08:22:05 pid:15670 remote addresses: 127.0.0.1 ERROR: [rsDataObjOpen_impl:860] - [USER_INVALID_RESC_INPUT: [pt0] is not a leaf resource.

] [error_code=[-361000], path=[/tempZone/home/otherrods/2021-08-13Z12:21:58--irods-testing-Mgfmli/issue_5227], hierarchy=[]
Aug 13 08:22:05 pid:15670 remote addresses: 127.0.0.1 ERROR: iRODS Exception:
    file: /home/kory/dev/prog/cpp/irods/plugins/api/src/touch.cpp
    function: int (anonymous namespace)::rs_touch(rsComm_t *, bytesBuf_t *)
    line: 418
    code: -836000 (CAT_NOT_A_DATAOBJ_AND_NOT_A_COLLECTION)
    message:
        Logical path does not point to a collection or data object.
stack trace:
--------------

Dumping stack trace
<0>     Offset: 0x343   Address: 0x7f0d4a188053 irods::exception::exception(long, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, unsigned int, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&)
<1>     Offset:         Address: 0x7f0d4141fd4a /usr/lib/irods/plugins/api/libirods_touch_server.so(+0x85d4a) [0x7f0d4141fd4a]
<2>     Offset: 0x73    Address: 0x7f0d414efb13 int std::__1::__invoke_void_return_wrapper<int>::__call<int (*&)(RsComm*, BytesBuf*), RsComm*, BytesBuf*>(int (*&&&)(RsComm*, BytesBuf*), RsComm*&&, BytesBuf*&&)
[...]

PR #5834, opened 2021-09-08
[#5771] Server returns better error message for "itouch -R <coord_resc>" (4-2-stable)

korydraughn:


trel:
looks good to me - please # when ready.

korydraughn:
Added #

Issue #5776, opened 2021-08-16
icommands have a hidden dependency on irods-server package


korydraughn:
- [x] master
- [x] 4-2-stable
---
### What did you try to do?
I ran `itouch` in a docker container that only has the icommands installed.

### Expected behavior
I expected the icommand to execute without any issues.

### Observed behavior (including steps to reproduce, if applicable)
I got the following message: 
```bash
remote addresses: 127.0.0.1 ERROR: procApiRequest: apiTableLookup of apiNumber 20007 failed
```
This happened because the API plugins are not installed. The API plugins are provided by the `irods-server` package. Installing them fixes the issue.

## The Question
Should the icommands have a dependency on the irods-server package or should the API plugins be moved into another package (e.g. irods-runtime)?

trel:
definitely seems like they should just be in `irods-runtime` instead...  we have other precedent for this?

korydraughn:
I'm not sure. I know that the API plugins provide client-side and server-side shared libraries so it does seem that putting them in `irods-runtime` makes the most sense.

trel:
let's do that.

d-w-moore:
I'll retry this on this tip of 4-2-stable then.  But in my case the irods-server package was installed, so not sure if this is the same issue.

PR #5835, opened 2021-09-08
[#5776] API plugins are now part of the irods-runtime package (master)

korydraughn:


korydraughn:
@SwooshyCueb Do you see any potential issues with this change for master and 4-2-stable?

SwooshyCueb:
I do not think we should make this kind of change for stable, but I don't see any problem with doing it for master

SwooshyCueb (approved):

korydraughn:
Ok. Can you explain why you feel this is not appropriate for 4-2-stable? Do you expect this could negatively affect deployments or is it something else?

SwooshyCueb:
It's not *as big of a deal* now since we're not doing debian packaging properly (debian packages track symbols and complain loudly when any disappear unexpectedly), but it's still generally considered bad practice to change which package in which a binary resides in a minor update, particularly one so late in the lifecycle of the major update series.
To answer your question more directly, yes I think it *could* negatively affect deployments, but I don't know offhand if it *would*. I can't point to any specific issues I know would crop up if we make this change in stable, but it's not something I feel comfortable giving my stamp of approval.
That said, I don't feel super strongly about it, but if you think you have a really good reason for doing it in stable, feel free to disregard my opinion.

korydraughn:
That makes sense. We'll let @trel see your comment and then we can make a decision about whether we want to make this change for 4-2-stable.

trel:
We're not removing a symbol as much as moving it.

Since `irods-runtime` is always installed - this doesn't prevent anyone's future work/compatibility.    And it adds the functionality we need for `itouch`...

I think this is a safe and correct approach fr 4-2-stable.

korydraughn:
Added #

PR #5836, opened 2021-09-08
[#5776] API plugins are now part of the irods-runtime package (4-2-stable)

korydraughn:


korydraughn:
Added #

See #5835 for discussion.

PR #5853, opened 2021-09-15
dstream fixes and enhancements (master)

korydraughn:


lib/core/include/stream_factory_utility.hpp
```diff
@@ -128,7 +128,61 @@ namespace irods::experimental::io
...
+    ///                       types are supported:
+    ///                           - irods::experimental::io::replica_number
+    ///                           - irods::experimental::io::leaf_resource_name
+    ///                           - irods::experimental::io::root_resource_name
+    ///                       \endparblock
+    ///
+    /// \return A new factory function.
+    ///
+    /// \since 4.2.9
+    template <typename ReplicaIdType>
```
korydraughn, line 154: I do not like this name. I'd like the name to tell the user that this type is for identifying a specific replica.

Perhaps:
- `ReplicaIdentifyingInfo` (eww)
- `TargetReplicaType`
- Something else?

trel: `ReplicaIdentifier`?

unit_tests/src/unit_test_utils.hpp
```diff
@@ -116,13 +115,13 @@ namespace unit_test_utils
         }
 
         for (std::size_t i = 0; i < _size_in_bytes; ++i) {
-            out << _ch;
+            out.write(&_ch, 1);
```
trel, line 118: just curious... why is it '1' here?

https://www.cplusplus.com/reference/ostream/ostream/write/

korydraughn: We are writing one byte to the stream.

lib/core/include/stream_factory_utility.hpp
korydraughn: I'll use that.

unit_tests/src/unit_test_utils.hpp
trel: oh, dang, a for loop.   good lord.   carry on.

alanking (approved):

lib/core/include/stream_factory_utility.hpp
korydraughn: done

lib/core/include/stream_factory_utility.hpp
```diff
@@ -128,7 +128,61 @@ namespace irods::experimental::io
...
+    ///
+    /// \return A new factory function.
+    ///
+    /// \since 4.2.9
+    template <typename ReplicaIdentifier>
+    auto make_dstream_factory(irods::connection_pool& _conn_pool,
+                              std::string _logical_path,
+                              ReplicaIdentifier&& _replica_id)
+    {
+        return [&_conn_pool, p = std::move(_logical_path), repl_id = std::forward<ReplicaIdentifier>(_replica_id)]
```
korydraughn, line 159: should `std::forward` be used here or does a normal capture by reference maintain the type information for forwarding later.

korydraughn: through experimentation against clang and gcc, capturing by reference is the way to go.

trel:
these look good to me - # when ready

korydraughn:
Added #

PR #5846, opened 2021-09-13
Fix redirection logic and add hostname configuration to run_tests.py (4-2-stable)

alanking:
cherry-picked from work done for #5844

Issue #5774, opened 2021-08-14
itouch fails to create data object when connected to catalog consumer


alanking:
- [x] master
- [x] 4-2-stable
----------

### iRODS Version, OS and Version
4.2.10

### What did you try to do?
`itouch` while connected to catalog service consumer (CSC)

### Expected behavior
0-length file

### Observed behavior (including steps to reproduce, if applicable)
On CSC:
```bash
$ itouch foo
$ echo $?
0
$ cat log/rodsLog.2021.08.11
$ ils -l foo
remote addresses: 192.168.224.5 ERROR: lsUtil: srcPath /tempZone/home/rods/foo does not exist or user lacks access permission
```
From log on catalog service provider (CSP):
```
Aug 14 03:11:44 pid:161 remote addresses: 192.168.224.2, 192.168.224.5 ERROR: iRODS Exception:
    file: /irods/plugins/api/src/touch.cpp
    function: (anonymous namespace)::json (anonymous namespace)::parse_json(const bytesBuf_t *)
    line: 199
    code: -1084000 (INPUT_ARG_NOT_WELL_FORMED_ERR)
    message:
        Could not parse string into JSON.
stack trace:
--------------

Dumping stack trace
<0>     Offset: 0x6e    Address: 0x7f510e486a4e irods::exception::exception(long, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, unsigned int, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&)
<1>     Offset:         Address: 0x7f5105c033ea /usr/lib/irods/plugins/api/libirods_touch_server.so(+0x373ea) [0x7f5105c033ea]
<2>     Offset: 0x42    Address: 0x7f5105c3cab2 irods::api_call_adaptor<BytesBuf*>::operator()(irods::plugin_context&, RsComm*, BytesBuf*)
<3>     Offset: 0x13    Address: 0x7f5105c3ca33 std::__1::__function::__func<irods::api_call_adaptor<BytesBuf*>, std::__1::allocator<irods::api_call_adaptor<BytesBuf*> >, irods::error (irods::plugin_context&, RsComm*, BytesBuf*)>::operator()(irods::plugin_context&, RsComm*&&, BytesBuf*&&)
[...]

PR #5854, opened 2021-09-15
[#5774] Propagate errors from touch API on CSC (master)

alanking:
cherry-picked from work done in #5846 (additional commits after original cherry-pick)

Tests to be run overnight


trel:
looks good... please # when ready.

5774 is already closed... so make sure we get our cherry-picks and checkboxes correct.

alanking:
#'d

Issue #5848, opened 2021-09-14
cannot open two write connections to non unique replica using Token and Hierarchy keywords


d-w-moore:
- [x] 4-2-stable
- [x] master
---
### iRODS Version, OS and Version

iRODS 4-2-stable
Ubuntu 18

### What did you try to do?

From a Python client, open two write descriptors to a replica ( in separate server connections)  to a single data object having more than one replica on the system 

### Expected behavior

Logical locking should permit this when REPLICA_TOKEN_KW and {DEST_,}RESC_HIER_STR_KW are appropriately set.

### Observed behavior (including steps to reproduce, if applicable)

Logical locking permits it when only one replica exists in the catalog, but a LOCKED_DATA_OBJECT_ACCESS is raised if more than one replica exists.

```
 $ git clone -b conn_track.285 http://github.com/d-w-moore/python-irodsclient     ###----> test branch ( adds test and logging)
 $ python3 -m pip install ./python-irodsclient ; cd ./python-irodsclient/irods/test
 $ iadmin mkresc pt passthru
 $ iadmin mkresc leaf unixfilesystem `hostname`:/tmp/leaf
 $ iadmin addchildtoresc pt leaf
 $ truncate -s0 0m ; irm -fr 0m ; iput -R pt 0m ; python3 reproduce_285.py -R pt  --pass rods `ipwd`/0m
INFO:root:Data Object open yielding raw fd: <irods.data_object.iRODSDataObjectFileRaw object at 0x7f88e7bd0ef0>
INFO:root:                 from connection: <irods.connection.Connection object at 0x7f88e7bd0828>
INFO:root:token  = 2dd404f5-fe60-4bdf-bb00-af8090c6288a
INFO:root:hier   = pt;leaf
INFO:root:Data Object open yielding raw fd: <irods.data_object.iRODSDataObjectFileRaw object at 0x7f88e7bd0da0>
INFO:root:                 from connection: <irods.connection.Connection object at 0x7f88e7bd0748>
INFO:root:Data Object write  using  raw fd: <irods.data_object.iRODSDataObjectFileRaw object at 0x7f88e7bd0da0>
INFO:root:                 from connection: <irods.connection.Connection object at 0x7f88e7bd0748>
INFO:root:Data Object write  using  raw fd: <irods.data_object.iRODSDataObjectFileRaw object at 0x7f88e7bd0ef0>
INFO:root:                 from connection: <irods.connection.Connection object at 0x7f88e7bd0828>
```
[...]

Issue #5548, opened 2021-04-21
Voting on OPEN ignores pre-existing replicas not on default resource 


d-w-moore:
- [x] master
- [x] 4-2-stable
-————

### iRODS Version, OS and Version
iRODS 4-2-stable 69cc129bfd78281f26825966c5361d009c168890
Ubuntu 18.04

### What did you try to do?
   - create a new resource, **test_resource** , which is not the default resource, and create a data object on that resource.
   - open and write N bytes into that data object.
   - make sure the file size increases by N bytes

### Expected behavior
The pre-existing replica (zero-length at the time of creation) should now have the same number of bytes in it as were just written.

### Observed behavior (including steps to reproduce, if applicable)
   - we end up with a new replica on the default resource, which receives the written bytes, while the pre-existing replica gets none. 
   - the two replicas have different size but are both marked valid:
   ```
   rods              0 test_resource            0 2021-04-21.12:56 & test1
   rods              1 demoResc            4 2021-04-21.12:56 & test1
   ```
This means that the server's voting mechanism is not working as expected for an OPEN operation.
(In the below reproduction of the issue, we find that this behavior can be corrected by supplying a `DEST_RESC_NAME_KW` in the client's `open('w+')` call.)

To reproduce:

Copy/paste [`demo.py`](#demopy-script) from the script listed below, into the current working directory.
With PRC installed, run `python demo.py -b`

When the program halts at the python debugger (Pdb) prompt, examine the object `test1` using `ils -L` in a different shell / window. (You will see the  above-mentioned two valid replicas of different lengths).

At the interrupted program (Pdb) prompt, hit 'c' to continue the demo test.
The message indicating the failed test condition will print.
```
(Pdb) c
after writing the dummy_str of length 4
the desired condition obj.size == len(dummy_str) is False
```

On an alternative run using `-r` to cause the test program to hint the resource name of `test_resource` as an option in the OPEN call, we can see that the expected test condition will be fulfilled, and (if `-b` is used to inspect during  program pause) there will only be one replica present post-WRITE, as one would expect.
```
$ python demo.py  -r 
after writing the dummy_str of length 4
the desired condition obj.size == len(dummy_str) is True
```
---
<A id="demopy-script"></A>
### demo.py script
```
#!/usr/bin/env python
  
from __future__ import print_function
import os, sys, atexit
from irods.session import iRODSSession
import getopt
import pdb
import irods.keywords as kw

def get_session(auto_cleanup = True):
    try:
[...]

PR #5855, opened 2021-09-16
[#5548] and [#5848] Multi-1247 parallel transfer fixes (master)

alanking:
CI tests running for good faith effort but non-cleaned-up branch from before completed tests successfully

### DIFF
--- a/server/api/src/rsDataObjOpen.cpp
+++ b/server/api/src/rsDataObjOpen.cpp
@@ -212,6 +212,9 @@ namespace
             // and the operation is supposed to turn into an overwrite of the existing data object.
             // Return a value of 0 here and the caller will continue with the open operation.
             if (!creating_new_replica) {
+                irods::log(LOG_DEBUG, fmt::format(
+                    "[{}:{}] - switching to overwrite an existing replica [path=[{}], hierarchy=[{}]]",
+                    __FUNCTION__, __LINE__, _inp.objPath, hierarchy));
                 _inp.openFlags &= ~O_WRONLY;
                 _inp.openFlags |= O_RDWR;
                 cond_input[DEST_RESC_NAME_KW] = irods::hierarchy_parser{hierarchy.data()}.first_resc();
@@ -219,6 +222,10 @@ namespace
 
                 return 0;
             }
+
+            irods::log(LOG_DEBUG, fmt::format(
+                "[{}:{}] - creating a new replica [path=[{}], hierarchy=[{}]]",
+                __FUNCTION__, __LINE__, _inp.objPath, hierarchy));
 
             const auto special_collection_type = irods::get_special_collection_type_for_data_object(_comm, _inp);
             if (special_collection_type < 0) {
@@ -977,6 +984,10 @@ namespace
         // wrong and we should bail immediately. We need to reference the winning replica whereas in
         // the past the linked list was sorted with the winning replica at the head.
         auto obj = id::make_data_object_proxy(*info_head);
+
+        irods::log(LOG_DEBUG, fmt::format(
+            "[{}:{}] - path=[{}], hierarchy=[{}], replica count=[{}]",
+            __FUNCTION__, __LINE__, obj.logical_path(), hierarchy, obj.replica_count()));
 
         auto maybe_replica = id::find_replica(obj, hierarchy);
         if (!maybe_replica) {
--- a/unit_tests/src/test_logical_locking.cpp
+++ b/unit_tests/src/test_logical_locking.cpp
@@ -59,9 +59,23 @@ namespace
 
         REQUIRE(unit_test_utils::add_ufs_resource(comm, _name, "vault_for_"s + _name.data()));
     } // mkresc
+
+    auto close_replica(RcComm& _comm, const int _fd)
+    {
+        const auto close_input = nlohmann::json{
+            {"fd", _fd},
+            {"update_size", false},
+            {"update_status", false},
+            {"compute_checksum", false},
+            {"send_notifications", false},
+            {"preserve_replica_state_table", true}
+        }.dump();
+
+        return rc_replica_close(&_comm, close_input.data());
+    } // close_replica
 } // anonymous namespace
 
-TEST_CASE("open and close", "[write_lock]")
+TEST_CASE("open_and_close", "[write_lock]")
 {
     load_client_api_plugins();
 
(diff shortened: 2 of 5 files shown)


unit_tests/src/test_rc_data_obj.cpp
```diff
@@ -540,14 +552,206 @@ TEST_CASE("open,read,write,close")
             CHECK(contents.size() + 1 == static_cast<unsigned long>(replica_info.size()));
             CHECK(STALE_REPLICA == replica_info.replica_status());
         }
+
+        SECTION("create_empty_replica_on_non_default_resource_and_write_in_separate_open__issue_5548")
```
alanking, line 556: @d-w-moore: These 3 unit tests are meant to kind of emulate the python scripts you provided in #5548 (original issue and follow-up issue) and #5848. Let me know if these seem equivalent to you and/or if they could use more comments to make that clearer. Thanks!

server/api/src/rsDataObjOpen.cpp
```diff
@@ -212,6 +212,9 @@ namespace
             // and the operation is supposed to turn into an overwrite of the existing data object.
             // Return a value of 0 here and the caller will continue with the open operation.
             if (!creating_new_replica) {
+                irods::log(LOG_DEBUG, fmt::format(
+                    "[{}:{}] - switching to overwrite an existing replica [path=[{}], hierarchy=[{}]",
```
trel, line 216: add missing second closing bracket at end of hierarchy

server/api/src/rsDataObjOpen.cpp
```diff
@@ -220,6 +223,10 @@ namespace
                 return 0;
             }
 
+            irods::log(LOG_DEBUG, fmt::format(
+                "[{}:{}] - creating a new replica [path=[{}], hierarchy=[{}]",
```
trel, line 227: add missing second closing bracket at end of hierarchy

unit_tests/src/test_logical_locking.cpp
```diff
@@ -175,11 +212,26 @@ TEST_CASE("open and close", "[write_lock]")
...
+        // open existing, currently locked replica
+        fail_cond_input[RESC_HIER_STR_KW] = other_replica_resc;
+        REQUIRE(LOCKED_DATA_OBJECT_ACCESS == rcDataObjOpen(&new_comm, &fail_open_inp));
     }
 
-    SECTION("open for create")
+    SECTION("open_for_create")
     {
-        fail_open_inp.openFlags = O_CREAT | O_WRONLY | O_TRUNC;
+        fail_open_inp.openFlags = O_CREAT | O_WRONLY;
```
trel, line 234: just curious, please explain why no more trunc?

unit_tests/src/test_rc_data_obj.cpp
trel: and regardless of @d-w-moore's answer here... Dan, please make sure to write new PRC tests that box in these two issues as well.

unit_tests/src/test_rc_data_obj.cpp
```diff
@@ -540,14 +552,206 @@ TEST_CASE("open,read,write,close")
...
+                // ensure all system metadata were updated properly
+                const auto [replica_info, replica_lm] = replica::make_replica_proxy(comm, target_object, 0);
+                CHECK(replica_info.mtime() == replica_info.ctime());
+                CHECK(0 == static_cast<unsigned long>(replica_info.size()));
+                CHECK(GOOD_REPLICA == replica_info.replica_status());
+            }
+
+            {
+                // ensure all system metadata were updated properly
+                const auto [replica_info, replica_lm] = replica::make_replica_proxy(comm, target_object, 1);
```
trel, line 741: i only see the replica number is different here... they should both be marked good?   please add another few words to these two 'ensure' blocks to differentiate/confirm what should be happening.      it is because there was no writing done here?   so the original statuses are all restored?  aka both back to `GOOD_REPLICA`?

server/api/src/rsDataObjOpen.cpp
alanking: Done

server/api/src/rsDataObjOpen.cpp
alanking: Done

unit_tests/src/test_logical_locking.cpp
alanking: I added a new piece to this test where the replica token DOES allow an additional open to succeed. If `O_TRUNC` is present, the file will be truncated. In this test, the file has contents and the purpose of this test is for checking only to see if the open would succeed. Can add back, but it wasn't really necessary to begin with and, if it is added back, I think there should be some additional assertions around the file's contents.

unit_tests/src/test_rc_data_obj.cpp
alanking: Okay, I will add some clarifying comments. 

Correct, the system metadata should be unchanged save for the replica statuses - which should be restored - because no writing was done here, as you said.

unit_tests/src/test_logical_locking.cpp
trel: That's fine fine - just seemed an important change without your usual storytelling.   Resolving.

unit_tests/src/test_rc_data_obj.cpp
alanking: Done

alanking:
Oh yeah, timing tests, unit tests, core tests, and federation passed in CI

unit_tests/src/test_rc_data_obj.cpp
d-w-moore: Wrt tests, will do .... I've now verified #5848 is fixed from the PRC perspective. Now will check 5548.

d-w-moore: @alanking Also wrt 5848 and clients, I've noticed that RESC_HIER_STR_KW is the operative keyword to use in conjunction with the replica token on a second open( ) call.  As opposed to  *DEST_*REST_HIER_STR_KW.  Please correct if wrong ! ...

trel:
nice.  i'm happy to see these #'d.  but please poke the @d-w-moore one more time for confidence.

unit_tests/src/test_rc_data_obj.cpp
d-w-moore: #5548 seems good too, statuses are updated to stale for the non-modified replica as appropriate.

alanking: Cool, glad it's working.

Yes, `RESC_HIER_STR_KW` is the operative keyword for `rcDataObjOpen`. We could possibly allow `DESC_RESC_HIER_STR_KW` to serve the same purpose but that keyword exists for APIs such as `rcDataObjRepl` and `rcDataObjPhymv` where the user is dealing with a source and destination replica.

It does seem inconsistent that `RESC_NAME_KW` and `DEST_RESC_NAME_KW` both influence voting when resolving a resource hierarchy and yet `DEST_RESC_HIER_STR_KW` does not affect the behavior of `rcDataObjOpen` at all (that is, it does not bypass hierarchy resolution like `RESC_HIER_STR_KW` allows).

Thoughts?

trel: these are so poorly defined/named.  gah.

trel: perhaps a little markdown table here in the comments with keyword / api / comments...?

alanking: I will be writing an issue shortly to open up this discussion elsewhere. Resolving.

alanking:
#'d

Issue #5863, opened 2021-09-20
rxDataObjOpen unit test randomly fails on ctime/mtime assertion


korydraughn:
- [x] master
- [x] 4-2-stable
---
### iRODS Version, OS and Version
Tip of master and 4-2-stable

### What did you try to do?
I ran the unit tests.

### Expected behavior
I expected all tests to pass consistently.

### Observed behavior (including steps to reproduce, if applicable)
The following line fails sometimes. If you run the specific test case that contains this line multiple times, you will eventually see the failure. It appears the server may be using different timestamps for the ctime/mtime when creating a replica.
https://github.com/irods/irods/blob/205ee0e3bd7e3b4107ed44c8c49027d4ec21b2f9/unit_tests/src/test_rc_data_obj.cpp#L742

alanking:
This is being caused by the finalize logic of `rsDataObjRepl` always setting the mtime for the new replica to the current time which can sometimes differ from the create time. This is not the intended behavior. For comparison, `rsDataObjClose` never sets the mtime for new replicas. This should probably be the case for replication as well.

alanking:
Actually, when I said "This is not the intended behavior.", I meant to say that this is a departure from the historical behavior. I think it makes the most sense that the create time and mtime match from the time of creation for a new replica.

To illustrate the comparison I mentioned, here is the logic in `rsDataObjClose` for finalizing an ordinary open-for-write/close: https://github.com/irods/irods/blob/fdb138a6c2d7025f8de0af0308229b3c21d2b71a/server/api/src/rsDataObjClose.cpp#L367-L386
The mtime is only updated when opened for write and when bytes are written, not on create.

Meanwhile, in `rsDataObjRepl`... https://github.com/irods/irods/blob/fdb138a6c2d7025f8de0af0308229b3c21d2b71a/server/api/src/rsDataObjRepl.cpp#L235-L239
The mtime is updated no matter what when finalizing the replica.

korydraughn:
Yes. It makes sense for the mtime to match the ctime for new replicas. That is my first thought. However ...

What if the mtime of the new replica matched the source replica's mtime (for replication)? Is there benefit in that? Having the mtime match the source replica's mtime would introduce the case where the ctime is later than the mtime. Would users care about that?

alanking:
After some discussion, we have decided that the mtime and create time should be individual to the replicas (that is, neither will be inherited from the source replica).

[...]

PR #5871, opened 2021-09-20
[#5863] Update mtime for creates/non-empty writes (master)

alanking:
Historically, the mtime for a replica was only updated if it was opened
for write (that is, it must have already existed) and bytes were written
during the operation. This changes the logic such that any operation
that results in a create or a bytes-writing open-for-write will update
the mtime for the target replica. This means that the mtime and create
time for a new replica may not be equal.

___
CI tests running

### DIFF
--- a/lib/core/include/stream_factory_utility.hpp
+++ b/lib/core/include/stream_factory_utility.hpp
@@ -159,7 +159,7 @@ namespace irods::experimental::io
     {
         using openmode = std::ios_base::openmode;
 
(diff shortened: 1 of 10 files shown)


trel:
looks good - please # when tests are happy

alanking:
All tests passed except for `test_parallel_transfer_engine` unit test. Latest commit fixes this.

trel:
awaiting kory eyeballs, then please #

alanking:
#'d

Issue #5861, opened 2021-09-17
"iadmin modresc" should not accept resource names with whitespace


korydraughn:
- [x] master
- [x] 4-2-stable
---
### iRODS Version, OS and Version
Found using ZMT. Verified using `iadmin modresc`.

### What did you try to do?
I renamed an existing resource.
```bash
$ iadmin modresc ufs0 name 'ufs 0'
```

### Expected behavior
I expected an error to be returned and maybe a log message stating whitespace is not allowed in resource names.

### Observed behavior (including steps to reproduce, if applicable)
It succeeded in changing the resource's name.

trel:
this will make `modresc` consistent with the similar check as part of `mkresc`.

```
$ iadmin mkresc 'no thanks' unixfilesystem `hostname`:/tmp/nopes
Creating resource:
Name:		"no thanks"
Type:		"unixfilesystem"
Host:		"xx.xx.xx.xx"
Path:		"/tmp/nopes"
Context:	""
remote addresses: 127.0.1.1 ERROR: rcGeneralAdmin failed with error -130000 SYS_INVALID_INPUT_PARAM
```

PR #5872, opened 2021-09-20
[#5861] iadmin resource commands no longer accept resource names containing whitespace (master)

korydraughn:


korydraughn:
added #

Issue #5791, opened 2021-08-23
msiGetStderrInExecCmdOut segfaults when using failed *out as input


trel:
- [x] master
- [x] 4-2-stable
---------

### iRODS Version, OS and Version

4.2.8, as per #5778

### What did you try to do?

see the stderr produced by a failed `msiExecCmd()`

```
  *status = errorcode(msiExecCmd("bisque_paths.py", *argStr, ipc_RE_HOST, "null", "null", *out));

  if (*status != 0) {
    msiGetStderrInExecCmdOut(*out, *resp);
    fail;
  }
```

### Expected behavior

cleanly capture the stderr

### Observed behavior (including steps to reproduce, if applicable)

iRODS Agent segfault and stacktrace

alanking:
Adapting the rule code found in #5778 to a minimal example, here is the output when reproducing this issue:
```
Aug 24 13:09:00 pid:8353 NOTICE: writeLine: inString = Running hello
Aug 24 13:09:00 pid:8353 remote addresses: 127.0.0.1, 172.17.0.2 ERROR: [-]     /irods_git_repo/server/re/src/irods_re_structs.cpp:50:int actionTableLookUp(irods::ms_table_entry &, char *) :  status [PLUGIN_ERROR_MISSING_SHARED_OBJECT]  errno [] -- message [Failed to create ms plugin entry.]
        [-]     /irods_git_repo/server/re/src/irods_ms_plugin.cpp:254:irods::error irods::load_microservice_plugin(irods::ms_table &, const std::string) :  status [PLUGIN_ERROR_MISSING_SHARED_OBJECT]  errno [] -- message [Failed to create ms plugin entry.]
                [-]     /irods_git_repo/lib/core/include/irods_load_plugin.hpp:157:irods::error irods::load_plugin(PluginType *&, const std::string &, const std::string &, const std::string &, const Ts &...) [PluginType = irods::ms_table_entry, Ts = <char [4]>] :  status [PLUGIN_ERROR_MISSING_SHARED_OBJECT]  errno [] -- message [shared library does not exist [/usr/lib/irods/plugins/microservices/libUNDEFINED.so]]

Aug 24 13:09:00 pid:8353 NOTICE: writeLine: inString = Ran hello
Caught signal [11]. Dumping stacktrace and exiting

Dumping stack trace
<0>     Offset:         Address: 0x7f22da1b3123 /usr/lib/libirods_server.so.4.2.8(+0xd64123) [0x7f22da1b3123]
[...]

PR #5889, opened 2021-09-30
[#5791] msiGetStderrInExecCmdOut() no longer segfaults on bad input (4-2-stable)

korydraughn:
This PR also modifies `getStdoutInExecCmdOut()` so that it shares the same implementation as `getStderrInExecCmdOut()`.


trel:
i like the common codepath.   yes please #.    i think the centos build is failing due to out-of-date ca-certs, which should resolve as the latest packages get updated.

korydraughn:
added #

Issue #5894, opened 2021-10-01
icp between two zones results in SYS_API_NO_PRIV


alanking:
### iRODS Version, OS and Version
4.2.8 (also observed in 4.2.10)

### What did you try to do?
Run federation tests, which includes an `icp` from one zone to another

### Expected behavior
Pass

### Observed behavior (including steps to reproduce, if applicable)
Failure. Here are the results reproduced manually:

For a small file, copying from the remote zone to the local zone works:
```bash
$ ils -l /tempZone/home/zonehopper#otherZone/justin
  zonehopper        0 demoResc           0 2021-10-01.16:21 & justin
$ icp /tempZone/home/zonehopper#otherZone/justin dan
$ ils -l dan
  zonehopper        0 demoResc            0 2021-10-01.16:21 & dan
```

...but from the local zone to the remote zone does not:
```bash
$ icp dan /tempZone/home/zonehopper#otherZone/kory
remote addresses: 192.168.144.5 ERROR: cpUtil: cp error for /tempZone/home/zonehopper#otherZone/kory, status = -13000 status = -13000 SYS_NO_API_PRIV
```

For a large file, copying across the zones in either direction fails:
```bash
$ ils -l /tempZone/home/zonehopper#otherZone/kory
  zonehopper        0 demoResc     52428800 2021-10-01.16:21 & kory
$ icp /tempZone/home/zonehopper#otherZone/kory alan
remote addresses: 192.168.144.5 ERROR: cpUtil: cp error for /otherZone/home/zonehopper/alan, status = -13000 status = -13000 SYS_NO_API_PRIV
$ ils -l dave
  zonehopper        0 demoResc     52428800 2021-10-01.16:21 & dave
$ icp dave /tempZone/home/zonehopper#otherZone/terrell
remote addresses: 192.168.144.5 ERROR: cpUtil: cp error for /tempZone/home/zonehopper#otherZone/terrell, status = -13000 status = -13000 SYS_NO_API_PRIV
```

The log on the remote server says:
```
NOTICE: rsApiHandler: User has no permission for apiNumber 671 
```

The API number 671 is `L3_FILE_PUT_SINGLE_BUF_AN`, which is currently not allowed in the client API whitelist if it is being enforced: https://github.com/irods/irods/blob/07c17a8ff4b14321c49cc28b7c5bb391e24f6eee/server/core/src/client_api_whitelist.cpp#L114-L115

[...]

PR #5896, opened 2021-10-04
[#5894] Fix icp across zones (master)

alanking:
Need to run in the testing environment, works for 4-2-stable


alanking:
Rebased and #'d

PR #5897, opened 2021-10-04
[#5894] Fix icp across zones (4-2-stable)

alanking:
`test_federation` passed in the testing environment


trel:
@korydraughn we must have had a reason for hiding these at the time... what are we forgetting by opening these back up?

korydraughn:
I don't remember why we disabled these. I don't think there's any negative impact around these. I haven't spotted any APIs that invoke `L3_FILE_*_SINGLE_BUF_AN` so it is weird that we have to enable it.

@alanking Did you find which APIs are triggering these?

trel:
I notice that `DATA_GET_AN` is *not* `DATA_OBJ_GET_AN` (etc)...

Is this just federation using older, deprecated (3.x?) APIs?

alanking:
In 4.2.9, a function called `l3DataCopySingleBuf` was removed. It used `rsL3FilePutSingleBuf` and `rsL3FileGetSingleBuf` which are mapped to `L3_FILE_PUT_SINGLE_BUF_AN` and `L3_FILE_PUT_SINGLE_BUF_AN`. This was used by `rsDataObjRepl` (and by extension `rsDataObjPhymv`) and `rsDataObjCopy`. I'm going to need to try copying over federation to/from 4.2.8 and maybe earlier to see if things continue to function, but I suspect that this may be the point of redirection in the copy operation (i.e. when it redirects over federation to put data into the remote server or get data from it). I'll report back on this front once I get the machinery set up.

As for `DATA_GET_AN` and `DATA_PUT_AN`, these are used in the parallel transfer over high ports which is used in all of the large file transfers throughout the server, even up to the present. Large file copies over federation do not work in the outgoing direction (local -> remote) if `DATA_PUT_AN` is not whitelisted. Large file copies over federation do not work in the incomign direction (local <- remote) if `DATA_GET_AN` is not whitelisted.

The tests without `DATA_COPY_AN` being whitelisted pass, so we can leave this out for sure. Got a little trigger happy there :)

trel:
Looks good to me.  Please # when ready.

trel:
Please rebase as well, I think that will get you a green centos dot...

alanking:
After testing with 4.2.7, `L3_FILE_PUT_SINGLE_BUF_AN` is also needed in order for 4.2.7 to complete a copy. However, there is another issue with federation which I have documented in this comment: https://github.com/irods/irods/issues/5894#issuecomment-934698107

I vote we merge this for now to fix federation in 4.2.11 at least. Versions before 4.2.9 have a workaround.

trel:
Got it.  Merging.

PR #5911, opened 2021-10-13
[5909, 5937] Add package revision to package version strings and allow irods-runtime to replace files in older irods-server deb (4-2-stable)

SwooshyCueb:
New debian revision string follows the following format: `0+irods[revnum]~[distro]`.
- `0` because we're not debian.
- `+` because we're not a linux distribution.

Previously, `~[distro]` was part of the package filename, but not part of the version string proper.

I've also declared that some packages provide *themselves*, versioned without a debian revision string. I'm not sure if this actually works the way I intend it to (still playing around with it) but the idea is to not force all our other repos to change how they declare package relationships in 4-2-stable.

I'm marking this PR as a draft, but I would welcome feedback from someone who has more experience with funky debian package relationships.


SwooshyCueb:
I've done some more testing and I think I'm satisfied with these packaging changes. I still need to make a couple of changes to cmake stuff since we use the package filename suffix variable in other repos (oops!) but this should be mergable soon.

PR #5969, opened 2021-11-01
[5910] Move icommands packages from server package requirements to recommends/suggests (4-2-stable)

SwooshyCueb:
half of 5910 for stable


trel:
we realized today that `irodsctl` still uses/needs `irods-grid`... which is part of `irods-icommands`... so... this might not yet be ready for merge.   discush!

trel:
@alanking @d-w-moore related to this morning's conversation...

Issue #5910, opened 2021-10-13
irods-server package dependency on irods-icommands package is a tripping hazard


SwooshyCueb:
We declare a package dependency from irods-server on irods-icommands. In order to make the transition away from CPack a little easier this should be cleaned up.

PR #5970, opened 2021-11-01
[#5910, #5978, #6014] Gracefully shut down server on SIGINT/SIGTERM/SIGHUP, remove irods-grid dependency (master)

SwooshyCueb:


SwooshyCueb:
Here's my first try at tackling 5978. I haven't done *any* testing yet.

scripts/irods/controller.py
```diff
@@ -188,27 +181,42 @@ def irods_grid_shutdown(self, timeout=20, **kwargs):
...
                 try:
-                    self.irods_grid_shutdown(timeout=timeout)
+                    irods_graceful_shutdown(timeout=timeout, server_proc)
                 except Exception as e:
                     l.error('Error encountered in graceful shutdown.')
                     l.debug('Exception:', exc_info=True)
+                server_proc.terminate()
             else:
-                    l.warning('No iRODS servers running.')
+                # could not find gradpa, this message could be improved
```
korydraughn, line 213: gradpa -> grandpa

scripts/irods/controller.py
```diff
@@ -188,27 +181,42 @@ def irods_grid_shutdown(self, timeout=20, **kwargs):
...
+            if server_pid >= 0:
+                # make sure it's not a reused pid
+                server_proc = psutil.Process(server_pid)
+                if psutil.version_info >= (2, 0):
+                    server_pid_exec = server_proc.exe()
+                else:
+                    server_pid_exec = server_proc.exe
+                if server_proc.exe != self.config.server_executable:
+                    server_pid = -1
+                    server_proc = None
```
korydraughn, line 203: is `server_pid_exec` meant to be used somewhere?

server/core/src/irods_server_control_plane.cpp
```diff
@@ -198,22 +199,14 @@ namespace irods {
             return irods::error(e);
         }
 
-        std::stringstream pid_str; pid_str << svr_pid;
-        std::vector<std::string> args;
-        args.push_back( pid_str.str() );
+        // Previsouly, we ran kill_pid.py here.
```
trel, line 208: "Previsouly" -> "Previously"

server/core/src/irods_server_control_plane.cpp
```diff
@@ -198,22 +199,14 @@ namespace irods {
...
+        // This would send three signals to the process, one right after the other:
+        // SIGSTOP (suspend), SIGTERM (terminate), then SIGKILL (kill)
```
trel, line 210: but now... not sending SIGSTOP?... can you explain the difference/reasoning from before to now?

server/core/src/irods_server_control_plane.cpp
```diff
@@ -1111,53 +1149,47 @@ namespace irods {
...
 
         // the icat needs to be notified first in certain
         // cases such as RESUME where it is needed to capture
         // the host list for validation, etc
-        ret = notify_icat_and_local_servers_preop(
-                  cmd_name,
-                  cmd_option,
-                  wait_option,
-                  wait_seconds,
+        irods::error ret = notify_icat_and_local_servers_preop(
```
trel, line 1237: consider a rename here ... from "icat" to "provider"

server/core/src/irods_server_control_plane.cpp
```diff
@@ -1111,53 +1149,47 @@ namespace irods {
...
         server_state& s = server_state::instance();
         std::string the_server_state = s();
         if ( server_state::PAUSED == the_server_state &&
-                SERVER_CONTROL_RESUME != cmd_name ) {
+                SERVER_CONTROL_RESUME != _cmd_name ) {
             _output = SERVER_PAUSED_ERROR;
             return SUCCESS();
         }
 
         // the icat needs to be notified first in certain
```
trel, line 1234: icat -> provider

server/core/src/irods_server_control_plane.cpp
```diff
@@ -1205,10 +1237,10 @@ namespace irods {
         // cases such as SHUTDOWN or PAUSE  where it is
         // needed to capture the host list for validation
         ret = notify_icat_and_local_servers_postop(
```
trel, line 1288: icat -> provider

scripts/irods/controller.py
```diff
@@ -188,27 +181,42 @@ def irods_grid_shutdown(self, timeout=20, **kwargs):
...
+                server_proc = psutil.Process(server_pid)
+                if psutil.version_info >= (2, 0):
+                    server_pid_exec = server_proc.exe()
+                else:
+                    server_pid_exec = server_proc.exe
+                if server_pid_exec != self.config.server_executable:
+                    server_pid = -1
+                    server_proc = None
+
+            if server_pid >= 0:
```
trel, line 205: is this line redundant?    consider adding a comment or four in here for the next spelunker.

server/core/src/irods_server_control_plane.cpp
korydraughn: my PR takes care of this so renames can be ignored for now if we agree.

server/core/src/irods_server_control_plane.cpp
korydraughn: see other comment.

server/core/src/irods_server_control_plane.cpp
korydraughn: see other comment.

server/core/src/irods_server_control_plane.cpp
trel: ah, very nice.   didn't realize it was the same code.

server/core/src/irods_server_control_plane.cpp
SwooshyCueb: SIGSTOP would prevent any relevant signal handler from running.

trel: yep, okay - just making sure that wasn't an oversight given it was mentioned as being handled prior.

scripts/irods/controller.py
SwooshyCueb: The previous `if` block is a sanity check. Since `lib.server_get_pid()` does not have access to `self.config`, it can't check the pid from the pidfile against `self.config.server_executable`, only that a process with that pid exists. We check it here to make sure that the pid wasn't recycled for some other process. I've already left a comment about this on line 195.

The `else` block here handles the cases in which `server_pid` was negative going into the previous `if` block *and* in which `server_pid` was *set* negative in the previous `if` block. Since we need to check `server_pid` again anyway, it's neater to structure the two if blocks like this. One that double-checks that the return value from `lib.get_server_pid()` is valid and one that actually does the thing.

SwooshyCueb:
TODO: In `controller.py`, handle the case in which grandpa goes away after receiving SIGTERM but leaves children behind.
At present, `get_binary_to_pids_dict` returns an empty dict if grandpa is gone.

scripts/irods/controller.py
trel: indeed - on all counts.

scripts/irods/controller.py
```diff
@@ -251,31 +263,84 @@ def status(self):
...
+    if proc.is_running():
+        try:
+            if proc.exe():
+                return os.path.samefile(binary_path, proc.exe())
+            else:
+                return os.path.basename(binary_path) == proc.name()
+        except psutil.NoSuchProcess:
+            return False
+
+def track_proc_tree(server_proc, server_descendants, candidate_binaries=None):
```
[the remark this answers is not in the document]
trel: also consider... use all the letters... `process`?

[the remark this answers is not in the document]
SwooshyCueb: `trim` is less accurate than `track`, since if new children spawn, they are added to the list.
But I am definitely open to more suggestions.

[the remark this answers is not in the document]
korydraughn: if this function isn't actively tracking the process tree in real time, `capture_process_tree` could be a good alternative.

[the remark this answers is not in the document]
trel: mmm, regarding that sentiment... `monitor_process_tree` or `watch_process_tree`... although `watch` is already a thing.

sounds like it *is* a bit more dynamic than just 'capture'...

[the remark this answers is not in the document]
korydraughn: those words make me think the script is watching the process tree in real time. is that what's happening?

korydraughn:
I've ran this PR manually and things appear to work. Containerized iRODS servers are not killed when shutting down or restarting the iRODS server running directly on the host OS.

@SwooshyCueb Is there something you'd like me to verify other than what I just mentioned?

Overall, this PR seems functionally ready.

scripts/irods/controller.py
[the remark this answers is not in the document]
trel: looking at the code... looks like it creates or refreshes an existing process tree.

so perhaps
`capture_current_process_tree`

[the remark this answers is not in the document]
korydraughn: i don't think introducing `current` into the name adds anything to the function name, but im fine with it if others like it.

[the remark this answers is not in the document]
SwooshyCueb: I like `capture_process_tree`

SwooshyCueb:
Nah, I think we're ready to \#, unless @trel has any objections

trel:
no objections.

SwooshyCueb:
\#'d

Issue #4231, opened 2019-01-04
Disallow remote groups


alanking:
- [x] master
- [x] master (icommands)
- [x] 4-2-stable
- [x] 4-2-stable (icommands)
-------

## Feature
`iadmin mkgroup`/`igroupadmin mkgroup` should not accept remote zones in group names as remote groups are not a valid concept. An error should be returned in this case.

alanking:
`iadmin moduser` should not allow changing the type of a user to/from `rodsgroup`. If a `rodsgroup` needs to become a `rodsuser` or `rodsadmin` (or vice versa), a new user should be created for this purpose.

`iadmin moduser` should not allow changing the zone of a user. Remote users only exist as a way of managing access for a user in another zone. A new local zone user account should be created if this need arises.

alanking:
Reopening to address unintentional API changes.

PR #5971, opened 2021-11-01
[#4231] Disallow remote groups (4-2-stable)

alanking:
CI tests to run, all the new tests pass at the bench.

### DIFF
--- a/scripts/irods/lib.py
+++ b/scripts/irods/lib.py
@@ -707,3 +707,30 @@ def add_child_resource(parent_resource_name, child_resource_name, user):
 
 def remove_child_resource(parent_resource_name, child_resource_name, user):
     user.assert_icommand(['iadmin', 'rmchildfromresc', parent_resource_name, child_resource_name])
+
+def get_user_type(session, user_name, zone_name=None):
+    """Get the type field of the user indicated by user_name and zone_name."""
+    gql = "select USER_TYPE where USER_NAME = '{}' and USER_ZONE = '{}'"
+    return session.assert_icommand(['iquest', '%s',
+        gql.format(user_name, zone_name or session.zone_name)], 'STDOUT')[1].strip()
+
+
+def get_user_zone(session, user_name, zone_name=None):
+    """Get the zone field of the user indicated by user_name and zone_name."""
+    gql = "select USER_ZONE where USER_NAME = '{}' and USER_ZONE = '{}'"
+    return session.assert_icommand(['iquest', '%s',
+        gql.format(user_name, zone_name or session.zone_name)], 'STDOUT')[1].strip()
+
+
+def get_user_comment(session, user_name, zone_name=None):
+    """Get the comment field of the user indicated by user_name and zone_name."""
+    gql = "select USER_COMMENT where USER_NAME = '{}' and USER_ZONE = '{}'"
+    return session.assert_icommand(['iquest', '%s',
+        gql.format(user_name, zone_name or session.zone_name)], 'STDOUT')[1].strip()
+
+
+def get_user_info(session, user_name, zone_name=None):
+    """Get the info field of the user indicated by user_name and zone_name."""
+    gql = "select USER_INFO where USER_NAME = '{}' and USER_ZONE = '{}'"
+    return session.assert_icommand(['iquest', '%s',
+        gql.format(user_name, zone_name or session.zone_name)], 'STDOUT')[1].strip()
--- a/server/api/src/rsGeneralAdmin.cpp
+++ b/server/api/src/rsGeneralAdmin.cpp
@@ -1,5 +1,6 @@
 #include "rsGeneralAdmin.hpp"
 
+#include "administration_utilities.hpp"
 #include "generalAdmin.h"
 #include "rodsConnect.h"
 #include "icatHighLevelRoutines.hpp"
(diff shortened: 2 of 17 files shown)


server/core/include/user_validation_utilities.hpp
```diff
@@ -0,0 +1,63 @@
...
+    ///
+    /// \param[in/out] _comm Server connection handle
+    /// \param[in] _user_name Name of the user whose user type is being fetched
+    ///
+    /// \return The user type of the given user as stored in the catalog
+    ///
+    /// \throws irods::exception If _user_name does not exist
+    ///
+    /// \since 4.2.11
+    auto get_type(RsComm* _comm, const std::string_view _user_name) -> std::string;
```
korydraughn, line 23: if `_comm` is expected to be valid, we should make this a reference instead of a pointer.

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -40,6 +44,58 @@ namespace
...
+            addRErrorMsg(&_comm->rError, err, msg);
+            THROW(err, msg);
+        }
+
+        // If this point is reached, this is a rodsuser or a rodsadmin having its type changed.
+        // There are checks later on which enforce the type to which the user is being chagned.
+    } // throw_if_group_is_changing_to_user_or_user_is_changing_to_group
+
+    auto throw_if_password_is_being_set_on_a_group(
+        RsComm* _comm,
```
korydraughn, line 84: reference instead of pointer?

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -40,6 +44,58 @@ namespace
 
         return std::find_if(b, e, [](unsigned char _ch) { return ::isspace(_ch); }) != e;
     } // contains_whitespace
+
+    auto throw_if_group_is_changing_to_user_or_user_is_changing_to_group(
+        RsComm* _comm,
```
korydraughn, line 49: reference instead of pointer?

server/core/include/user_validation_utilities.hpp
alanking: Yeps

server/api/src/rsGeneralAdmin.cpp
alanking: Done

server/api/src/rsGeneralAdmin.cpp
alanking: Done

alanking:
Linking irods/irods_client_icommands#237

scripts/irods/lib.py
```diff
@@ -707,3 +707,26 @@ def add_child_resource(parent_resource_name, child_resource_name, user):
 
 def remove_child_resource(parent_resource_name, child_resource_name, user):
     user.assert_icommand(['iadmin', 'rmchildfromresc', parent_resource_name, child_resource_name])
+
+def get_user_type(session, username):
+    """Get the type field of the user indicated by username."""
+    gql = "select USER_TYPE where USER_NAME = '{}'"
```
trel, line 713: all four of these in this diff could match more than one user?   `alice#tempZone` and `alice#otherZone`?

server/core/include/administration_utilities.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+namespace irods
+{
+    /// \brief Create a user with the given name, type, and auth string
+    ///
+    /// \parblock
+    /// If _user_name does not contain the zone name using the octothorpe delimiter, the
+    /// _zone_name will be used. Both of these are available here because the GeneralAdmin
+    /// interface is flexible enough to allow the zone name to be passed through the username
+    /// string or through the generalAdninInp struct.
```
trel, line 16: Is there a test where the zone name is both *in* the username string *and* the _zone_name?   And if they don't agree... seems the _user_name string wins?

server/core/src/user_validation_utilities.cpp
```diff
@@ -0,0 +1,76 @@
...
+#include <regex>
+
+namespace irods::user
+{
+    auto get_type(RsComm& _comm, const std::string_view _user_name) -> std::string
+    {
+        char user[NAME_LEN]{};
+        parseUserName(_user_name.data(), user, nullptr);
+
+        const std::string qstr = fmt::format("SELECT USER_TYPE where USER_NAME = '{}'", user);
```
trel, line 18: same as other comment... what happens if there is both a local and a remote alice?

scripts/irods/lib.py
alanking: A fine point. CI tests caught this as well, so I'll tinker with this a bit more.

server/core/include/administration_utilities.hpp
alanking: Looking at the implementation again, we actually aren't using _zone_name at all. And `iadmin` is passing an empty string for that parameter. So I'm just going to ignore it entirely and look for the zone name in the user name.

server/core/src/user_validation_utilities.cpp
alanking: Added zone name to the query to avoid duplicates.

scripts/irods/lib.py
alanking: Added the zone names and adjusted tests. Everything is passing now.

alanking:
Force-pushed because I had done some branching gymnastics to avoid some recent commits which I have now rebased to include and just went ahead and squashed everything.

trel:
very good - seems ready for the #s to me

alanking:
#'d

Issue #5908, opened 2021-09-23
ireg --repl -R <root-random-resource> assigns random storage resource to registered replica


tedgin:
- [ ] master
- [ ] 4-2-stable
---------

I'm using iRODS 4.2.8 on CentOS 7 servers. My iRODS zone has the following resource hierarchies.

```
CyVerseRes:random
├── ds01Broker:passthru
│   └── ds01:unixfilesystem
...
├── ds20Broker:passthru
│   └── ds20:unixfilesystem
├── dustyBroker:passthru
│   └── dusty:unixfilesystem
└── snoopyBroker:passthru
    └── snoopy:unixfilesystem

taccCorralRes:passthru
└── tacc:unixfilesystem
```
I have a data object named `/iplant/home/ipc_admin/hi.txt` with its only replica on `taccCorralRes;tacc`. I also have a file on `CyVerseRes;dustyBroker;dusty` that is a copy of this replica, and I want to register it as another replica of the associated data object. I use `ireg --repl` specifying the resource as `dusty`'s root resource `CyVerseRes`. Here's the command.

```sh
$ ireg --repl -R CyVerseRes /irods_vault/dusty_vault/home/ipc_admin/hi.txt /iplant/home/ipc_admin/hi.txt
```

The command succeeds, but surprisingly, instead of registering the file as a replica on `dusty` it chooses the peer storage resource `ds20`.

```sh
$ ils -L /iplant/home/ipc_admin/hi.txt
  ipc_admin         1 taccCorralRes;tacc            3 2017-04-22.15:09 & hi.txt
    764efa883dda1e11db47671c4a3bbd9e    text    /corral-tacc/projects/iplant/Vault/home/ipc_admin/hi.txt
  ipc_admin         2 CyVerseRes;ds20Broker;ds20            3 2021-09-23.12:59 & hi.txt
    764efa883dda1e11db47671c4a3bbd9e    text    /irods_vault/dusty_vault/home/ipc_admin/hi.txt
```

I verified that it didn't copy the file to the stated location `/irods_vault/dusty_vault/home/ipc_admin/hi.txt` on `ds20`'s server.

`ireg` should have either registered the file as a replica on `dusty` or generated on error.

alanking:
After some further discussion...

[...]

PR #5983, opened 2021-11-04
Fix various issues with ireg --repl

alanking:
Open to squashing in whatever way we see fit.

CI tests running


trel:
no squashing necessary.   these hang together well enough.

bring on the #s once passing/green.

alanking:
After hitting a few snags, it has become clear that disallowing `ireg` to target a root resource is going to be a wide-reaching change that will require community input. We will punt on #5908 for now until we can answer some more questions.

I also noticed that the solution for disallowing registering multiple replicas on the same resource relies on the source replica being on the target resource for the new replica. In other words, there is a chance that the source replica could be on a different resource than the target resource for the new replica but there is already a replica on that resource. The solution for this is to check the existing replicas to make sure none of them are on the target resource. I already have this implemented locally, just need to test it.

alanking:
Tests passing. #'d

PR #5988, opened 2021-11-05
Fix various issues with ireg --repl (4-2-stable)

alanking:
cherry-picked from work for #5983 

CI tests to be run


alanking:
Tests are passing

PR #6000, opened 2021-11-09
[5748, 5758, 5918, 5962, 5963] Packaging touch-ups (4-2-stable)

SwooshyCueb:


trel:
5963 message, please update to say 'plugins' rather than plugs

rest of it looks pretty good to me.   please # when happy.

SwooshyCueb:
\#'d

Issue #6006, opened 2021-11-12
Non-package install is broken due to shared memory filename collision


korydraughn:
- [x] main
- [x] 4-2-stable
---
### iRODS Version, OS and Version
4.2.10

### What did you try to do?
Tried to run multiple iRODS servers using non-package install.

### Expected behavior
I expected the servers to stand up and run without any issues.

### Observed behavior (including steps to reproduce, if applicable)
The first server runs fine while the secondary servers experience a boost::interprocess exception.

PR #6009, opened 2021-11-16
[#6006] Stopped shared memory filename collisions from happening (main)

korydraughn:


trel:
please add an epoch seconds to this scheme somewhere - consider looking/following the shared memory pattern used for our rule engine plugins... irods_rule_language perhaps?

korydraughn:
added epoch seconds to filename.

added #

PR #6015, opened 2021-11-22
[#4231] Restore support for zone name param in admin API (4-2-stable)

alanking:
The zone name can now be provided via argument in the GeneralAdmin API
as well as the UserAdmin API. This is done to support clients which
provide the argument. iadmin will provide the desired zone name through
the octothorpe-delimited method.

---

CI tests passed. The PRC tests which began to fail are now passing at the bench.

I started to add unit tests for this API but didn't find it to be particularly beneficial at this time. Please let me know if we want to add a unit test for the APIs.


trel:
i think we're pretty well tested - no need for unit tests here for now.

alanking:
#'d

PR #6016, opened 2021-11-22
[#4231] Restore support for zone name param in admin API (master)

alanking:
The zone name can now be provided via argument in the GeneralAdmin API
as well as the UserAdmin API. This is done to support clients which
provide the argument. iadmin will provide the desired zone name through
the octothorpe-delimited method.

---

CI tests running. cherry-picked from work for #6015

### DIFF
--- a/server/api/src/rsGeneralAdmin.cpp
+++ b/server/api/src/rsGeneralAdmin.cpp
@@ -677,7 +677,8 @@ _rsGeneralAdmin( rsComm_t *rsComm, generalAdminInp_t *generalAdminInp ) {
             return irods::create_user(*rsComm,
                                       generalAdminInp->arg2 ? generalAdminInp->arg2 : "",
                                       generalAdminInp->arg3 ? generalAdminInp->arg3 : "",
-                                      generalAdminInp->arg5 ? generalAdminInp->arg5 : "");
+                                      generalAdminInp->arg5 ? generalAdminInp->arg5 : "",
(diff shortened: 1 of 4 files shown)


server/api/src/rsGeneralAdmin.cpp
```diff
@@ -677,7 +677,8 @@ _rsGeneralAdmin( rsComm_t *rsComm, generalAdminInp_t *generalAdminInp ) {
             return irods::create_user(*rsComm,
                                       generalAdminInp->arg2 ? generalAdminInp->arg2 : "",
                                       generalAdminInp->arg3 ? generalAdminInp->arg3 : "",
-                                      generalAdminInp->arg5 ? generalAdminInp->arg5 : "");
+                                      generalAdminInp->arg5 ? generalAdminInp->arg5 : "",
+                                      generalAdminInp->arg4 ? generalAdminInp->arg4 : "");
```
trel, line 681: just caught my eye that these are 'out of order'... disregard if this is confirmed to be correct.

alanking: Yeah, the order from iadmin is:
 1. action ("add")
 2. name
 3. type
 4. zone name
 5. auth string

I can change the order if desired. It's a server-only interface so it can be changed freely in the future

alanking:
#'d but can change the argument order if desired

trel:
consistency is good, but like you said, it's server only, so not worried about it - will merge when green

Issue #6022, opened 2021-11-30
`ilsresc` silently defaults to local zone when given non-existent zone.


kript:
- [x] main
- [x] 4-2-stable
---------

### iRODS Version, OS and Version

4.2.7 

### What did you try to do?

`ilsresc -z ThisZoneDoesntExist`

### Expected behavior

error message saying "ThisZoneDoesntExist"... doesn't exist.

### Observed behavior (including steps to reproduce, if applicable)

Resources from 'home' Zone displayed instead.

kjsanger:
Also present in 4.2.10 and recent HEAD of 4-2-stable.

trel:
Agreed - this should produce an error.

korydraughn:
Based on the code, the argument passed via the `-z` flag is never used. The resolved path of `.` is used as the zone hint.
https://github.com/irods/irods/blob/abb809040485326855ffffbda1883dc99a432379/lib/core/src/lsUtil.cpp#L40
Not only that, but the zone hint is changed for each path.

korydraughn:
@kript Just to clarify, both `ils` and `ilsresc` have a problem when using the `z` option?

korydraughn:
My guess is that the `ils` is a typo since it does not advertise a `-z` option.

kript:
yeah, I'm a muppet - its just `ilsresc`

trel:
updated the original text.

trel:
checkboxed.

PR #6033, opened 2021-12-02
[#6022] rxGenQuery returns an error on an unknown zone name (main)

korydraughn:


korydraughn:
Passed python test at the bench.

Added #

PR #6034, opened 2021-12-02
[#4649] Fix leaks observed on rule invocation (4-2-stable)

alanking:
CI tests to be run

PR #6042, opened 2021-12-07
[#4649] Fix leaks observed on rule invocation (master)

alanking:
cherry-picked from work done for #6034


trel:
i think we're good for master - please # em.

alanking:
Done

PR #5720, opened 2021-06-28
[5669, 5713] Use find_package for nholmann-json and update for 3.10.4

SwooshyCueb:
In service of 5669 and 5713

This PR does ~~five~~ ~~four~~ ~~two~~ three things:
1. ~~Fixes a copy-paste error in `CheckCPPMacroDefinition.cmake`~~
2. ~~Updates `lib/core/include/experimental_plugin_framework.hpp` and `plugins/api/src/experimental_api_plugin_adapotor.cpp` to use a version of `json::parse` that is not deprecated in nholmann-json 3.8.~~ (This change is backward-compatible with nholmann-json 3.7.3, which we currently use, and 3.6.1, which is in epel7)
3. Introduces a find module for nholmann-json that checks `IRODS_EXTERNALS_FULLPATH_JSON` first ~~and contains a workaround for 5713~~.
4. ~~Fixes the `#include` lines for nlohmann-json to use `<nlohmann/json.hpp>`~~ (This change is backward-compatible with nholmann-json 3.7.3/3.6.1, but *not* our package for 3.7.3)
5. Bumps the expected extern version for nohlmann-json to 3.10.4.
6. Bumps the minimum libc++ version to 9.

The first two commits can probably be backported to stable without much fuss.

PR #6050, opened 2021-12-09
Ubuntu 20.04 / Externals deps / Python 3

korydraughn:
Depends on https://github.com/irods/externals/pull/136


cmake/server.cmake
```diff
@@ -177,25 +177,19 @@ install(
 
 
 set(CPACK_DEBIAN_${IRODS_PACKAGE_COMPONENT_SERVER_NAME_UPPERCASE}_PACKAGE_NAME "irods-server")
-set(CPACK_DEBIAN_${IRODS_PACKAGE_COMPONENT_SERVER_NAME_UPPERCASE}_PACKAGE_DEPENDS "${IRODS_PACKAGE_DEPENDENCIES_STRING}, irods-runtime (= ${CPACK_DEBIAN_PACKAGE_VERSION}-${CPACK_DEBIAN_PACKAGE_RELEASE}), libc6, sudo, libssl1.0.0, libfuse2, python, openssl, python-psutil, python-requests, lsof")
-set(CPACK_DEBIAN_${IRODS_PACKAGE_COMPONENT_SERVER_NAME_UPPERCASE}_PACKAGE_RECOMMENDS "irods-icommands (>= ${CPACK_DEBIAN_PACKAGE_VERSION})")
+set(CPACK_DEBIAN_${IRODS_PACKAGE_COMPONENT_SERVER_NAME_UPPERCASE}_PACKAGE_DEPENDS "${IRODS_PACKAGE_DEPENDENCIES_STRING}, irods-runtime (= ${CPACK_DEBIAN_PACKAGE_VERSION}-${CPACK_DEBIAN_PACKAGE_RELEASE}), libc6, sudo, libssl1.1, libfuse2, python3, openssl, python3-psutil, python3-requests, python3-pyodbc, lsof")
+set(CPACK_DEBIAN_${IRODS_PACKAGE_COMPONENT_SERVER_NAME_UPPERCASE}_PACKAGE_RECOMMENDS "irods-icommands (>= ${CPACK_DEBIAN_PACKAGE_VERSION}), python3-jsonschema")
```
korydraughn, line 181: consider adding rsyslog

cmake/server.cmake
```diff
@@ -177,25 +177,19 @@ install(
...
 if (IRODS_LINUX_DISTRIBUTION_NAME STREQUAL "centos")
-  set(CPACK_RPM_${IRODS_PACKAGE_COMPONENT_SERVER_NAME}_PACKAGE_REQUIRES "${IRODS_PACKAGE_DEPENDENCIES_STRING}, irods-runtime = ${CPACK_RPM_PACKAGE_VERSION}-${CPACK_RPM_PACKAGE_RELEASE}, openssl, python, python-psutil, python-requests, python-jsonschema")
+  set(CPACK_RPM_${IRODS_PACKAGE_COMPONENT_SERVER_NAME}_PACKAGE_REQUIRES "${IRODS_PACKAGE_DEPENDENCIES_STRING}, irods-runtime = ${CPACK_RPM_PACKAGE_VERSION}-${CPACK_RPM_PACKAGE_RELEASE}, openssl, python, python-psutil, python-requests, python-jsonschema, pyodbc")
 elseif (IRODS_LINUX_DISTRIBUTION_NAME STREQUAL "almalinux" OR IRODS_LINUX_DISTRIBUTION_NAME STREQUAL "rocky")
-  set(CPACK_RPM_${IRODS_PACKAGE_COMPONENT_SERVER_NAME}_PACKAGE_REQUIRES "${IRODS_PACKAGE_DEPENDENCIES_STRING}, irods-runtime = ${CPACK_RPM_PACKAGE_VERSION}-${CPACK_RPM_PACKAGE_RELEASE}, openssl, python2, python2-psutil, python2-requests")
+  set(CPACK_RPM_${IRODS_PACKAGE_COMPONENT_SERVER_NAME}_PACKAGE_REQUIRES "${IRODS_PACKAGE_DEPENDENCIES_STRING}, irods-runtime = ${CPACK_RPM_PACKAGE_VERSION}-${CPACK_RPM_PACKAGE_RELEASE}, openssl, python3, python3-psutil, python3-requests, python3-pyodbc")
```
korydraughn, line 192: consider adding rsyslog

scripts/irods/database_connect.py
```diff
@@ -197,23 +194,41 @@ def get_database_connection(irods_config):
...
+            # TODO This needs to be tested against a PostgreSQL 10+ database.
+            # See the following for why these are needed:
+            #
+            #   - https://github.com/mkleehammer/pyodbc/issues/169#issuecomment-268665172
+            #   - https://github.com/mkleehammer/pyodbc/issues/169#issuecomment-591979062
+            #   - https://github.com/mkleehammer/pyodbc/issues/169#issuecomment-591994201
+            #
+            conn.setencoding(encoding='utf-8')
+            conn.setdecoding(pyodbc.SQL_CHAR,  encoding='utf-8')
+            conn.setdecoding(pyodbc.SQL_WCHAR, encoding='utf-8')
```
korydraughn, line 212: this is not needed if the unicode-enabled driver is used.

pyodbc must be used instead of pypyodbc because pypyodbc does not confirm to support python 3 and it does not provide functions for setting the encoding/decoding on SQL operations.

korydraughn: i'm also concerned that database code written in C/C++ may not work. running this through the testing environment will help us determine what needs to happen in that case.

cmake/server.cmake
trel: seems reasonable

cmake/server.cmake
trel: but only for the recommends?   not the requires?

plugins/api/src/experimental_api_plugin_adaptor.cpp
```diff
@@ -35,8 +35,7 @@ namespace {
 
     auto to_json(bytesBuf_t* b)
     {
-        std::string foo{(char*)b->buf, (char*)b->buf+b->len};
-        return json::parse({(char*)b->buf, (char*)b->buf+b->len});
+        return json::parse(std::string(static_cast<char*>(b->buf), b->len));
```
SwooshyCueb, line 38: Does this copy the buffer? A better solution might be to use the other overload for parse that takes pointers.

korydraughn: this is definitely worse than the solution you have in your PR.

i'll change it to match yours.

SwooshyCueb:
My PR #5720 covers updating for the new nlohmann-json package. I would suggest waiting for it to be merged before merging this.

plugins/api/src/experimental_api_plugin_adaptor.cpp
korydraughn: done

scripts/irods/database_connect.py
korydraughn: this situation may be due to locale used by the container and postgres.

standing up iRODS against a properly configured UTF-8 postgres database or official postgres container may remove the need for these lines.

cmake/server.cmake
korydraughn: done

cmake/server.cmake
korydraughn: done. see line 195.

korydraughn:
The changes in this PR allow the server to be compiled against the new externals packages.

Demonstrated on Ubuntu 18.04.

korydraughn:
Attempts to build against C++ 20 (i.e. `CXX_STANDARD 20`) results in failure due to the Boost external's use of removed library feature, `std::result_of`.

Updating the Boost external's build instructions (i.e. `-std=c++14` to `-std=c++17`) does not fix the problem. Maybe more options need to be adjusted. It is possible that explicitly defining `BOOST_ASIO_HAS_STD_INVOKE_RESULT` will fix this assuming Clang 13 supports `std::invoke_result`. See [Boost 1.77's Asio build docs](https://www.boost.org/doc/libs/1_77_0/doc/html/boost_asio/using.html#boost_asio.using.compiler_platform_feature_detection_macros) for more information.

korydraughn:
added #'s

korydraughn:
added #'s

Issue #6063, opened 2021-12-14
Do not create "rodsadmin" group on new installs


korydraughn:
## Bug Report
Having a user type and group called **rodsadmin** is very confusing and can lead users to making incorrect assumptions about the system.

The purpose of the **rodsadmin** group is not known and does not provide any special properties as far as the system is concerned.

On upgrade of an existing deployment, this group will be left in place.

trel:
It appears to never have been populated automatically - I think we can not create it on install any longer.

https://github.com/irods/irods_rule_engine_plugin_logical_quotas/issues/46#issuecomment-994985761

trel:
Yes, this is correct.   Any user can be of type 'rodsadmin'.

You can make any groups you want for the users, but being in a group never conveyed any permissions/power, directly.    This new install makes that clearer.

PR #6064, opened 2021-12-14
[#6063] "rodsadmin" group is not created on new installs (main)

korydraughn:


trel:
this seems correct...

after a clean install, we should put here:
1) output of `iadmin lu`
2) output of `iadmin lg`
3) output from a clean test run

trel:
let’s call it new install rather than vanilla.

korydraughn:
updated title and commit message.

trel:
looks good - please get output mentioned earlier here... and then # when ready/happy/confirmed.

korydraughn:
i'm guessing we want to list the users/groups following the test put?

how do we feel about the following output:
```bash
iRODS Users:
  - user 1
  - user 2
  - ...
  - user N

iRODS Groups:
  - group 1
  - group 2
  - ...
  - group N
```
is this output needed if new installs only contain the `rods` user and `public` group?

maybe this output is better:
```bash
Registered iRODS Users : rods
Registered iRODS Groups: public
```

korydraughn:
Output of requested commands following a new install.
```bash
$ iadmin lu
rods#tempZone
$ iadmin lg
public
```

trel:
good for me.

korydraughn:
I tried to run this through CI, but it fails on setup. I think we need to get the testing environment working with main.

trel:
@alanking, consider pushing this through a basic testrun?

trel:
oh, it says right there... main.    as you were.

korydraughn:
added #

Issue #6081, opened 2021-12-21
Specify which externals packages need to be installed for github actions


korydraughn:
https://github.com/irods/irods/blob/98a861b15b874c07b06187a043a3a0303563460a/.github/workflows/build-irods-centos.yml#L36
https://github.com/irods/irods/blob/98a861b15b874c07b06187a043a3a0303563460a/.github/workflows/build-irods-ubuntu.yml#L41

PR #6082, opened 2021-12-21
[#6081] Explicitly defined which externals to install for github actions (main)

korydraughn:


trel:
these look correct - awaiting two greens...

korydraughn:
added #

PR #6099, opened 2022-01-11
[multiple] CMake target housekeeping

SwooshyCueb:
Doing some prep-work for repo reorganization, CMake reorganization, and migration of external dependencies to `find_package`.

This PR does a few things:
- Removes a lingering reference to `IRODS_EXTERNALS_FULLPATH_JSON`
- Migrates usage of `OPENSSL_*` variables to `OpenSSL::` targets
- Migrates usage of `CURL_*` variables to `CURL::` targets
- Switches to `CMAKE_DL_LIBS` for linking against `libdl`
- Removes `libm` from `CMAKE_CXX_STANDARD_LIBRARIES` (was previously added in `UseLibCXX.cmake` when using `libc++` with `g++`)
- Switches many uses of `CMAKE_BINARY_DIR` and `CMAKE_SOURCE_DIR` to `CMAKE_CURRENT_BINARY_DIR` and `CMAKE_CURRENT_SOURCE_DIR`, respectively
- Cleaned up unneeded uses of `OpenSSL::` targets
- Cleaned up unneeded uses of `CURL::` targets
- Cleaned up unneeded uses of `Threads::Threads` targets
- Made some `OpenSSL::` target linkages `PUBLIC`
- Made some `CURL` target linkages `PUBLIC`
- Made `Threads::Threads` target linkages `PUBLIC`
- Made some non-import target linkages `PUBLIC`


SwooshyCueb:
\#'d

PR #6096, opened 2022-01-10
[#5927] Add checksum verification keywords to replication API and microservice (4-2-stable)

alanking:
I split this up into two parts: the API changes and the microservice changes. Willing to squash it down.

This passed in CI

Issue #5927, opened 2021-10-16
clarify msiDataObjRepl's verifyChksum when source replica already has checksum in catalog


trel:
- [x] master
- [x] 4-2-stable
-------

## Feature

`msiDataObjRepl` should be more specific/thorough in its description of how the `verifyChksum` keyword is used.   If we determine that `msiDataObjRepl` has a bug in that it is not honoring the `verifyChksum` keyword correctly, then we should turn this issue into a bug report.

There is an 'implicit' verification happening when the source replica already has a checksum.

This can be surprising/problematic when writing to a tape resource... aka, it is written to tape, then read from tape to verify the contents were written correctly... which can be very expensive/slow.

The checksum API was reworked for clarity and consistency in https://github.com/irods/irods/commit/2a0460718f7142bf494c1030b7d7617e1e60ba20 in 4.2.9.

Comments requesting this clarification issue are from:
 - https://groups.google.com/g/irod-chat/c/dKIB_jrPQkU/m/CivEkWZSAgAJ
 - https://github.com/irods/irods/issues/5592#issuecomment-842177531

trel:
@alanking Am I reading this correctly, that `verifyChksum` / `VERIFY_CHKSUM_FLAG` / `VERIFY_CHKSUM_KW` is ... not considered by `rsDataObjRepl() -> replicate_data_object() -> replicate_data() -> finalize_destination_replica()` ?

https://github.com/irods/irods/blob/f79ff76e1b3d1c677cbe54ce2aa3c5a0ad03d63b/server/api/src/rsDataObjRepl.cpp#L217-L225

It appears to only care whether the source already had a checksum?

---

If so, and we want to 'restore' the functionality of `msiDataObjRepl`'s `verifyChksum` parameter... this issue should be marked as a bug.

alanking:
I think you are reading that correctly, yes.

Will need to make (yet) another matrix for these options to make sure it is understood what they all mean.

trel:
Okay, marking as bug, and bumping to 4.2.12.

alanking:
As noted above, there are at least the following explicit ways to indicate a desire to verify the checksum for a given replica:
 - `VERIFY_CHKSUM_KW`
https://github.com/irods/irods/blob/bc6f9f1cdef6c4ec01ea14402428988892615321/lib/core/include/rodsKeyWdDef.h#L14
 - `VERIFY_CHKSUM_FLAG`
[...]

PR #6105, opened 2022-01-12
[#5927] Add checksum verification keywords to replication API and microservice (main)

alanking:
cherry-picked from work done for #6096 

I have confirmed that this builds, but haven't run through testing yet as things have moved enough such that our testing infrastructure doesn't support main right now. Workin' on it


trel:
looks good to me - awaiting your test confirmation/satisfaction, then # em.

alanking:
Tests are passing on main. #'d

PR #6119, opened 2022-01-20
Remove msiSetRescSortScheme and check error codes in db_plugin (main)

alanking:
Tests to be run

### DIFF
--- a/plugins/database/src/db_plugin.cpp
+++ b/plugins/database/src/db_plugin.cpp
@@ -429,29 +429,23 @@ irods::error determine_user_has_modify_metadata_access(
 } // user_has_modify_metadata_access
 
 
-/*
- * removeMetaMapAndAVU - remove AVU (user defined metadata) for an object,
- *   the metadata mapping information, if any.  Optionally, also remove
- *   any unused AVUs, if any, if some mapping information was removed.
- *
- */
-void removeMetaMapAndAVU( char *dataObjNumber ) {
-    char tSQL[MAX_SQL_SIZE];
-    cllBindVars[0] = dataObjNumber;
+// remove AVU (user defined metadata) for an object, the metadata mapping information, if any.
+int removeMetaMapAndAVU(const char* _id)
+{
+    char tSQL[MAX_SQL_SIZE]{};
+    cllBindVars[0] = _id;
     cllBindVarCount = 1;
-    if ( logSQL != 0 ) {
-        rodsLog( LOG_SQL, "removeMetaMapAndAVU SQL 1 " );
-    }
-    snprintf( tSQL, MAX_SQL_SIZE,
-              "delete from R_OBJT_METAMAP where object_id=?" );
-    cmlExecuteNoAnswerSql( tSQL, &icss );
-    /* Note, the status will be CAT_SUCCESS_BUT_WITH_NO_INFO (not 0) if
-       there were no rows deleted from R_OBJT_METAMAP, in which case there
-       is no need to do the SQL below.
-    */
 
-    return;
-}
+    if (logSQL) {
+        rodsLog(LOG_SQL, "removeMetaMapAndAVU SQL 1 ");
+    }
+
+    snprintf(tSQL, MAX_SQL_SIZE, "delete from R_OBJT_METAMAP where object_id=?");
+
+    const auto ec = cmlExecuteNoAnswerSql(tSQL, &icss);
+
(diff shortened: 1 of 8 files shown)


plugins/database/src/db_plugin.cpp
```diff
@@ -799,7 +793,12 @@ static int _delColl( rsComm_t *rsComm, collInfo_t *collInfo ) {
         return CATALOG_NOT_CONNECTED;
     }
 
-    splitPathByKey( collInfo->collName, logicalParentDirName, MAX_NAME_LEN, logicalEndName, MAX_NAME_LEN, '/' );
+    if (const auto ec = splitPathByKey(collInfo->collName, logicalParentDirName, MAX_NAME_LEN, logicalEndName, MAX_NAME_LEN, '/'); ec < 0) {
+        irods::log(LOG_NOTICE, fmt::format(
```
trel, line 797: just poking to make sure you meant LOG_NOTICE.

plugins/database/src/db_plugin.cpp
```diff
@@ -1149,7 +1152,12 @@ rodsLong_t checkAndGetObjectId(
     }
 
     if ( itype == 1 ) {
-        splitPathByKey( name, logicalParentDirName, MAX_NAME_LEN, logicalEndName, MAX_NAME_LEN, '/' );
+        if (const auto ec = splitPathByKey(name, logicalParentDirName, MAX_NAME_LEN, logicalEndName, MAX_NAME_LEN, '/'); ec < 0) {
+            irods::log(LOG_NOTICE, fmt::format(
```
trel, line 1156: same

plugins/database/src/db_plugin.cpp
```diff
@@ -3282,8 +3304,13 @@ irods::error db_unreg_replica_op(
         }
         status = cmlExecuteNoAnswerSql(
                      "delete from R_OBJT_ACCESS where object_id=? and not exists (select * from R_DATA_MAIN where data_id=?)", &icss );
+        // If this fails, what does it mean?
```
trel, line 3307: not sure it's actionable for us... it would be the database itself...

plugins/database/src/db_plugin.cpp
alanking: I didn't feel that the message would be particularly helpful, but maybe this does rise to the level of `LOG_ERROR`... willing to change. `splitPathByKey` emits error messages of its own, so I figured that would be enough for the rare case that something does go wrong. Thoughts?

trel: later in the commit it was treated more loudly - that is all.    i think NOTICE is probably fine, as you said, since it emits its own error messages.

alanking: Oooh, I see. Yes, I will go ahead and make these `LOG_ERROR` then. I don't expect this situation to come up very much so I suppose we ought to know about it when it does.

trel: cool cool.

plugins/database/src/db_plugin.cpp
trel: i think we'd just have a dangling avu ... pointed to a no-longer-existing-dataobj...

would this result/dangling get picked up and removed by `iadmin rum`?

trel:
found the phrase "data object count" here... might also be something to clean up?   that resc_hier, i think, can be torched... we don't keep a running total in the database anymore for that.

https://github.com/irods/irods/blob/main/plugins/database/src/db_plugin.cpp#L3198


rest looks good to me.

plugins/database/src/db_plugin.cpp
alanking: I think you're right. Of course, whatever it was trying to delete here from `R_OBJT_ACCESS` would also potentially remain, assuming the `icss` struct was not modified. Would it be appropriate to perform a rollback on error here?

trel: upon initial inspection... latest commit e4941a0 seems correct/good.

trel:
Yes, it was for when we were storing the live `object_count` in the catalog... which was always subject to race conditions and speed concerns.   We deleted that logic, but didn't clean up the variable holding the resc_hier.

Yes, please delete.

alanking:
Tests passed in CI

trel:
squash em up.

alanking:
#'d

Issue #6114, opened 2022-01-18
run_tests.py assumes python2-style filter()


alanking:
- [x] main
---
4.3.x will only support python3. `scripts/run_tests.py` is using the `filter()` function like a python2 application - that is, a `list` is generated immediately upon invocation. python3 makes `filter()` a generator instantiation, so `len()` can no longer be called on it. The `"filtered_markers"` section of the code should be updated to be python3 compatible.

PR #6121, opened 2022-01-21
[#6114] Make (almost) all tests in main pass (main)

alanking:
All tests passed. A good chunk of the change was moving some calls in the `setUp()` method of one of the test classes to reduce the amount of work done for other test cases.


trel (commented):
most impressive.

i have nothing to add.

please # when ready.

alanking:
#'d

Issue #3479, opened 2017-02-14
Investigate async safe logging of stack trace in signal handler for release builds


kellerb:

mcv21:
Hi,

Can you add the consortium_member tag to this and #3326 please? I've been looking at a race condition elsewhere in the code, and found the undefined behaviour in irods_signal.cpp. I was going to report it, but I found these two issues about it already.

Thanks :)

korydraughn:
the following code needs to be investigated as well.
https://github.com/irods/irods/blob/1467ffd6fb83c83bf4f031218900a3698fe88925/server/core/src/irodsReServer.cpp#L623-L633

PR #6131, opened 2022-01-25
[#3479] Made stacktrace logging async signal safe (main)

korydraughn:


CMakeLists.txt
```diff
@@ -134,6 +134,9 @@ set(IRODS_COMPILE_DEFINITIONS ${IRODS_COMPILE_DEFINITIONS}
                               _LARGE_FILES
                               _LARGEFILE64_SOURCE
                               BOOST_SYSTEM_NO_DEPRECATED
+                              BOOST_CHRONO_DONT_PROVIDES_DEPRECATED_IO_SINCE_V2_0_0
+                              #BOOST_STACKTRACE_USE_BACKTRACE # Cannot find <backtrace.h>
+                              BOOST_STACKTRACE_USE_ADDR2LINE # This works! Needs to be tested on other platforms/OS.
```
korydraughn, line 139: investigate which demangling lib to use

server/core/include/irods_signal.hpp
```diff
@@ -1,8 +1,8 @@
...
+#define IRODS_SIGNAL_HPP
 
-extern "C" {
-    extern void register_handlers( void );
-    extern void unregister_handlers( void );
-}
-#endif // _IRODS_SIGNAL_HPP_
+void register_handlers();
+
+void unregister_handlers();
```
korydraughn, line 6: this function is not defined and should be removed.

server/core/src/irods_signal.cpp
```diff
@@ -1,38 +1,84 @@
...
 {
-    /// @brief Signal handler for seg faults
-    static void segv_handler(int _signal)
-    {
-        using log = irods::experimental::log;
+    timespec ts;
+
+    if (clock_gettime(CLOCK_REALTIME, &ts) != 0) {
+        std::exit(_signal);
+    }
```
korydraughn, line 24: if we can't get the time, we exit.

not sure we can do anything else for this situation.

server/core/src/irods_signal.cpp
```diff
@@ -1,38 +1,84 @@
...
+                std::rename(file_path_hidden, file_path);
+            }
+        }
     }
-}
+
+    std::exit(_signal);
+} // stacktrace_signal_handler
+
+void register_handlers() // TODO Consider calling this for grandpa also.
```
korydraughn, line 76: add the word `signal` to this name.

suggestions:
- `set_signal_handlers`
- `register_signal_handlers`
- `set_unrecoverable_signal_handlers`
- `configure_unrecoverable_signal_handlers`

server/core/src/rodsServer.cpp
```diff
@@ -305,6 +312,110 @@ namespace
...
+                }
+                catch (...) {
+                    // Something happened while logging the stacktrace file.
+                    // Leaving the stacktrace file in-place for processing later.
+                    log::server::trace("Caught exception while processing stacktrace file.");
+                }
+            }
+
+            using namespace std::chrono_literals;
+            std::this_thread::sleep_for(3s); // TODO How long should we wait?
```
korydraughn, line 416: should this have a configuration options?

server/core/src/rodsServer.cpp
```diff
@@ -673,6 +796,7 @@ int serverMain(
 
             }
             ix::cron::cron::get()->run();
+            // TODO Consider letting the CRON manager handle the logging of stacktrace files.
```
korydraughn, line 799: review the CRON manager code to see if there are benefits to using it

server/core/src/rodsServer.cpp
```diff
@@ -768,6 +891,10 @@ int serverMain(
         return_code = e_.code();
     }
 
+    // Instruct the stacktrace logger background thread to stop.
+    stop_stacktrace_logger_thread.store(true);
+    irods::experimental::log::server::trace("Set stop token for stacktrace thread to true.");
```
korydraughn, line 896: remove log message

server/core/src/irods_signal.cpp
trel: could... try to write anything we have down?  without a timestamp?   if it's going to the file, the file itself may get a timestamp from the OS?

server/core/src/irods_signal.cpp
trel: don't have a strong opinion.   other programs have a consistent usage?

just want to make sure we respect / don't pollute a namespace elsewhere.

server/core/src/rodsServer.cpp
trel: yes...  now gotta name it too...

and then '3' or whatever could be the backstop default if not/poorly defined in the advanced_settings.

server/core/src/rodsServer.cpp
trel: - single loop, no watcher to watch the watcher required.
- easier to reason about later.

server/core/src/irods_signal.cpp
korydraughn: that means we have to deal with two schemes for retrieving the info.

questions:
- how do we keep filenames unique?
- do we care to keep filenames unique?
- what format should filenames use? (this impl uses: `<epoch_seconds>.<remaining_millis>.<agent_pid>`)

keep in mind that we are allowed to use a subset of the functions available to us here.

it would be better if we can design things in a way that only requires one code path.

server/core/src/rodsServer.cpp
korydraughn: do we have any suggestions for the config options?

suggestions:
- `stacktrace_logger_sleep_time_in_seconds`
- `stacktrace_file_detector_sleep_time_in_seconds`
- `stacktrace_file_processor_sleep_time_in_seconds`

server/core/src/rodsServer.cpp
korydraughn: if many stacktrace files exist on a server, processing them here will block incoming client requests from being serviced.

server/core/src/irods_signal.cpp
trel: agreed - carry on.

server/core/src/rodsServer.cpp
trel: true... how many are we planning on having?   i mean, maybe stopping and processing broken things is ... good ... at that point?

server/core/src/rodsServer.cpp
korydraughn: there aren't any big downsides to using a larger sleep time other than the possibility of more files appearing and the fact that the stacktrace would appear later in the log, in a location that is unrelated to when the stacktrace actually happened.

and yes, the files would survive server restarts.

server/core/src/rodsServer.cpp
korydraughn: hopefully none :) -- however, one file means we are touching the disk. every time we do there here, we block for the amount of work the cron manager has to handle.

i need to review the cron code tho. i can't remember how it actually works.

server/core/src/rodsServer.cpp
korydraughn: - added config option to server_config.json.template
- updated code to use value from config file
- defaults to 3 if not defined or is less than 1

server/core/src/irods_signal.cpp
korydraughn: done

lib/core/include/irods_configuration_keywords.hpp
```diff
@@ -76,6 +76,7 @@ namespace irods
     extern const std::string CFG_ADDRESS_KW;
 
     // advanced settings
+    extern const std::string DELAY_RULE_EXECUTORS_KW;
```
trel, line 79: alpha order?  do we have a convention for these things?

server/api/src/rsDataObjPut.cpp
```diff
@@ -669,6 +669,8 @@ int preProcParaPut(
     return status;
 } // preProcParaPut
 
+int crash(int* t) { return 5 + *t; }
```
trel, line 672: why 5?  give it a name?  or comment.

server/core/src/irods_signal.cpp
```diff
@@ -1,38 +1,87 @@
...
-    static void segv_handler(int _signal)
-    {
-        using log = irods::experimental::log;
+    timespec ts;
+
+    if (clock_gettime(CLOCK_REALTIME, &ts) != 0) {
+        std::exit(_signal);
+    }
+
+    const auto max_path_length = 512;
```
trel, line 26: why 512?

server/core/src/irods_signal.cpp
```diff
@@ -1,38 +1,87 @@
...
+
+    char file_path_hidden[max_path_length] = "../../";
+    std::strcat(file_path_hidden, IRODS_DEFAULT_PATH_HOMEDIR);
```
trel, line 29: why hidden?

server/core/src/rodsServer.cpp
```diff
@@ -305,6 +312,129 @@ namespace
...
+
+                if (iter != std::end(adv_settings)) {
+                    if (const auto seconds = iter->get<int>(); seconds > 0) {
+                        return seconds;
+                    }
+                }
+            }
+            catch (...) {}
+
+            return 3; // The default sleep time.
```
trel, line 347: is 3 good?  perhaps 10?  or 30?   let's leave a justfication in a comment if we have one.

lib/core/include/irods_configuration_keywords.hpp
korydraughn: We don't.

server/api/src/rsDataObjPut.cpp
korydraughn: This code will be removed. It's there to trigger a stacktrace and see if things work.

trel: ohright.  i read that.

server/core/src/irods_signal.cpp
korydraughn: No particular reason. I figured this would be large enough to hold a path. Maybe this should match the largest path Linux supports.

server/core/src/irods_signal.cpp
korydraughn: The hidden path keeps the stacktrace processor thread from touching it until the file is ready.

trel: oh, it's hidden from the processor... hmmm...   perhaps... `file_path_unprocessed`?

server/core/src/irods_signal.cpp
trel: seems defensible.

server/core/src/rodsServer.cpp
korydraughn: No reason. It can be whatever we want. The larger it is, the later it will appear in the log.

trel: right, but balanced against how often this code fires and walks the directory for files to process.

3 feels quick to me.   @kript you have a good feel for this?  what's a default value that you wouldn't feel the immediate need to change?

server/core/src/irods_signal.cpp
korydraughn: that works.

server/core/src/irods_signal.cpp
korydraughn: changed to use _POSIX_PATH_MAX which is guaranteed to be 256.

a buffer of this size is safe given that we're dealing with a relative path.

lib/core/include/irods_configuration_keywords.hpp
korydraughn: will create an issue about this file.

server/core/src/irods_signal.cpp
```diff
@@ -46,26 +48,26 @@ void stacktrace_signal_handler(int _signal)
...
 
-                char file_path[max_path_length];
-                std::strcpy(file_path, file_path_hidden);
+                char file_path[_POSIX_PATH_MAX];
+                std::strcpy(file_path, file_path_unprocessed);
                 
                 // An indicator to the thread handling the logging of these files that this file
                 // is not ready to be logged (i.e. the signal handler is still writing to it).
-                std::strcat(file_path_hidden, ".not_ready_for_logging");
+                std::strcat(file_path_unprocessed, ".not_ready_for_logging");
```
trel, line 63: should ".not_ready_for_logging" be a variable?   it's used elsewhere... the processor, I presume...

`unprocessed_stacktrace_file_suffix`?

server/core/src/irods_signal.cpp
```diff
@@ -23,19 +24,20 @@ void stacktrace_signal_handler(int _signal)
         std::exit(_signal);
     }
 
-    const auto max_path_length = 512;
+    // "_POSIX_PATH_MAX" is guaranteed to be defined as 256.
+    // This value represents the minimum path length supported by POSIX operating systems.
```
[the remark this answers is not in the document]
korydraughn: no, minimum is correct. PATH_MAX exists, but isn't guaranteed to be defined and even if it is, it can be different across OS.

server/core/src/irods_signal.cpp
korydraughn: yes it should.

korydraughn: i haven't figured out where though

korydraughn: i'll put it in the header file. it is specific to this, so it makes sense.

korydraughn: done.

server/core/src/irods_signal.cpp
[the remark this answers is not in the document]
korydraughn: the alternative is to look up the system limit using `fpathconf(stdin, _PC_PATH_MAX)`.

the issue is that it might be indeterminate. now we're back to either defining our value (which could result in `ENAMETOOLONG` errors) or using the guaranteed minimum of 256.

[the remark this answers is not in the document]
trel: portable seems good.  256 should be enough, and if not, that means someone has a non-package install way deep in their filesystem.   probably would cause other problems as well at that depth.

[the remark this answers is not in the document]
korydraughn: agreed.

server/core/src/rodsServer.cpp
```diff
@@ -305,6 +313,129 @@ namespace
...
+        boost::system::error_code ec;
+        fs::create_directories("../.." / fs::path{IRODS_DEFAULT_PATH_HOMEDIR} / "stacktrace", ec);
+    } // create_stacktrace_directory
+
+    void log_stacktrace_files(std::atomic<bool>& _stop_token)
+    {
+        namespace fs = boost::filesystem;
+
+        // The leading "../.." enables support for non-package installations.
+        const auto stacktrace_directory = "../.." / fs::path{IRODS_DEFAULT_PATH_HOMEDIR} / "stacktrace";
```
trel, line 329: consider adding...

```
// In a packaged installation, this will produce '/var/lib/irods/stacktrace'
```

server/core/src/rodsServer.cpp
```diff
@@ -305,6 +313,129 @@ namespace
 
         return 0;
     } // create_pid_file
+
+    void create_stacktrace_directory()
+    {
+        namespace fs = boost::filesystem;
+        boost::system::error_code ec;
+        fs::create_directories("../.." / fs::path{IRODS_DEFAULT_PATH_HOMEDIR} / "stacktrace", ec);
```
trel, line 321: consider 'stacktraces' ?

server/core/src/rodsServer.cpp
```diff
@@ -408,6 +539,8 @@ int main(int argc, char** argv)
         return 1;
     }
 
+    create_stacktrace_directory();
```
trel, line 542: consider 'stacktraces' ?

server/core/src/rodsServer.cpp
korydraughn: done

server/core/src/irods_signal.cpp
```diff
@@ -1,38 +1,91 @@
...
+    // Given that we're dealing with a relative path, a buffer of this size is acceptable here.
+    char file_path_unprocessed[_POSIX_PATH_MAX] = "../../";
+    std::strcat(file_path_unprocessed, IRODS_DEFAULT_PATH_HOMEDIR);
+    std::strcat(file_path_unprocessed, "/stacktrace/");
+
+    // At this point, "file_path_unprocessed" should have the following path:
+    //
+    //     ../../<install_prefix>/stacktrace/
```
korydraughn, line 36: `<install_prefix>` should be changed to `<irods_home>`

server/core/src/rodsServer.cpp
```diff
@@ -305,6 +313,130 @@ namespace
...
+            return 3; // The default sleep time.
+        }();
+
+        using log = irods::experimental::log;
+
+        while (!_stop_token.load()) {
+            for (auto&& entry : fs::directory_iterator{stacktrace_directory}) {
+                // Expected filename format:
+                //
+                //     <epoch_seconds>.<remaining_milliseconds>.<agent_pid>
```
alanking, line 358: I might have missed it... what is meant by "remaining" in `remaining_milliseconds` here?

server/core/src/rodsServer.cpp
```diff
@@ -433,10 +567,10 @@ int main(int argc, char** argv)
         key_path_t{irods::CFG_ADVANCED_SETTINGS_KW, irods::CFG_HOSTNAME_CACHE_KW, irods::CFG_EVICTION_AGE_IN_SECONDS_KW},
         irods::get_hostname_cache_eviction_age());
 
-    /* start of irodsReServer has been moved to serverMain */
+    // Start of irodsReServer has been moved to serverMain.
     signal( SIGTTIN, SIG_IGN );
     signal( SIGTTOU, SIG_IGN );
-    signal( SIGCHLD, SIG_DFL ); /* SIG_IGN causes autoreap. wait get nothing */
+    signal( SIGCHLD, SIG_DFL ); // SIG_IGN causes autoreap. wait get nothing.
```
alanking, line 573: Feel free to ignore this...

This comment seems like it's meant for the `SIG_IGN` line above? Also, what does "wait get nothing" mean?

server/core/src/rodsServer.cpp
alanking: +1 for eventually using the cron manager, assuming everything checks out

alanking (commented):
Overall looks real good. Most of my comments are just observations/questions

server/api/src/rsDataObjPut.cpp
korydraughn: lol

server/core/src/rodsServer.cpp
korydraughn: it represents partial seconds. so it provides more accuracy as to when the stacktrace happened.

see `struct timespec` in https://man7.org/linux/man-pages/man2/clock_getres.2.html

server/core/src/rodsServer.cpp
korydraughn: i didn't pay attention to that comment.

will consider adjusting or removing it.

server/core/src/rodsServer.cpp
alanking: I guess I was more wondering if it wouldn't make more sense to say `epoch_milliseconds` or somesuch. The use of the word "remaining" was mildly confusing for me. It's fine as is :)

server/core/src/rodsServer.cpp
korydraughn: updated

server/core/src/irods_signal.cpp
korydraughn: updated

server/core/src/rodsServer.cpp
korydraughn: i think the name of the function is clear.

i don't think the function name needs to include the directory name.

willing to be convinced though.

server/core/src/rodsServer.cpp
korydraughn: do we have a preferred default?

trel: 10?

that's 3x slower than '3' (so less churn), but not so slow that it's WAY far away from *when* it happened.

server/core/src/rodsServer.cpp
trel: just figured since it could hold more than one...

fine either way.

server/core/src/rodsServer.cpp
korydraughn: going with 10 seconds

server/core/src/rodsServer.cpp
korydraughn: leaving function name as is.

the actual directory name has been updated to be "stacktraces".

server/core/src/rodsServer.cpp
korydraughn: done. needs to be tested.

server/api/src/rsDataObjPut.cpp
korydraughn: will test what the stacktrace output looks like using this.

server/core/include/process_manager.hpp
```diff
@@ -1,193 +1,240 @@
...
+                  func post,
+                  std::chrono::seconds interval,
+                  bool run_pre_once,
+                  bool run_post_once,
+                  std::any data)
+            : execute_(e)
+            , pre_(pre)
+            , post_(post)
+            , interval_(interval)
+            , last_run_(std::chrono::seconds(0))
```
korydraughn, line 71: unnecessary argument

server/core/include/process_manager.hpp
```diff
@@ -5,226 +5,213 @@
...
+#include <utility>
 #include <vector>
 
-// The main use case of this facility is to provide a means of running tasks intermittently.
-// that is, in an interval.
-
 namespace irods::experimental::cron
 {
-    /// \brief a cron task, that is, one run intermittently.
+    /// A cron task, that is, run intermittently.
```
trel, line 15: lotsa commas...

consider

"A cron task which runs once every 'interval' seconds"

server/core/src/rodsServer.cpp
```diff
@@ -305,6 +313,126 @@ namespace
...
+                utc_ss << time_fmt(boost::chrono::timezone::utc, "%FT%T") << tp;
+
+                // Read the contents of the file.
+                std::ifstream file{p};
+                const auto stacktrace = boost::stacktrace::stacktrace::from_dump(file);
+                file.close();
+
+                // 3. Write the contents of the stacktrace file to syslog.
+                log::server::trace("Sending stacktrace buffer to syslog ...");
+                irods::experimental::log::server::error({
```
alanking, line 413: It just occurred to me that maybe this should go under `server::critical` instead of `server::error`. Thoughts?

trel: i'd be okay with that - we don't use it very many places (1... 2?)

korydraughn: that's reasonable.

i was planning on removing the trace log messages, but maybe we should keep them.

server/core/src/rodsServer.cpp
korydraughn: changing to `epoch_milliseconds`.

server/core/src/rodsServer.cpp
korydraughn: done

server/core/src/rodsServer.cpp
korydraughn: done

CMakeLists.txt
korydraughn: Created #6146 to address this later.

trel:
failed to build - still boost1.77 shenanigans

korydraughn:
The current externals packages need to be built and published first. I've been bench testing with a custom build of externals.

trel:
gotcha.

server/core/src/irodsReServer.cpp
```diff
@@ -721,7 +726,7 @@ int main(int argc, char** argv)
         logger::delay_server::error(e.what());
     }
 
-    logger::delay_server::info("Rule execution server exiting ...");
+    logger::delay_server::info("Rule execution server exited normally.");
```
korydraughn, line 729: this message is never printed when shutting down via the irodsctl script because the control plane sends a SIGKILL after SIGTERM.

POSIX does not allow SIGKILL to be caught.

the control plane should give systems a few seconds to respond to the SIGTERM.

korydraughn: this can be handled in a separate PR too.

server/core/src/initServer.cpp
```diff
@@ -683,44 +675,9 @@ cleanupAndExit( int status ) {
...
-#if 0
-    // RTS - rodsLog calls in signal handlers are unsafe - #3326
-    rodsLog( LOG_NOTICE,
-             "caught a signal and exiting\n" );
-#endif
     cleanupAndExit( SYS_CAUGHT_SIGNAL );
 }
 
-void
-rsPipeSignalHandler( int ) {
```
korydraughn, line 695: nothing uses this so i removed it.

any objections?

korydraughn:
any last concerns before i add the #'s?

server/core/src/initServer.cpp
trel: did you get the one at
```
./server/core/include/initServer.hpp:36:rsPipeSignalHandler( int );
```

korydraughn: good catch.

removed.

trel:
I think we're ready to see the #s...   can always tweak the next PR to get them green again re: externals.

trel:
do we want to wait on the externals work before merging this?

korydraughn:
the "externals work" being ... make all externals compile against clang 13?

trel:
i mean getting the expected versions built and posted to the web, so we get green dots here.

korydraughn:
sure. that seems fine.

.github/workflows/build-irods-centos.yml
```diff
@@ -35,8 +35,8 @@ jobs:
               sudo yum -y update
               sudo yum -y install \
                 irods-externals-avro1.11.0-0 \
```
trel, line 37: bump this to avro1.11.0-1

.github/workflows/build-irods-ubuntu.yml
```diff
@@ -40,8 +40,8 @@ jobs:
               sudo apt-get update -qq
               sudo apt-get install -qq \
                 irods-externals-avro1.11.0-0 \
```
trel, line 42: and here

CMakeLists.txt
```diff
@@ -21,12 +21,12 @@ endmacro()
 
 IRODS_MACRO_CHECK_DEPENDENCY_SET_FULLPATH_ADD_TO_IRODS_PACKAGE_DEPENDENCIES_LIST(ARCHIVE libarchive3.5.2-0)
 IRODS_MACRO_CHECK_DEPENDENCY_SET_FULLPATH_ADD_TO_IRODS_PACKAGE_DEPENDENCIES_LIST(AVRO avro1.11.0-0)
```
trel, line 23: and here

trel (commented):
getting close

.github/workflows/build-irods-centos.yml
korydraughn: done

.github/workflows/build-irods-ubuntu.yml
korydraughn: done

CMakeLists.txt
korydraughn: done

korydraughn:
the builds were successful!

trel:
woo.

Issue #844, opened 2013-11-26
change ichmod to handle the full range of the iRODS permission model


trel:
_Originally Opened: @trel (2012-06-15 15:01:01_)

in `iRODS/server/icat/src/icatSysInserts.sql`, the current model for permissions involves 18 levels of control, significantly more than the 4 provided/suggested/explained via `ichmod`.

`ichmod` grants possibility for assigning [null/read/write/own].

the full model is linear and has:

```
insert into R_TOKN_MAIN values ('access_type',1000,'null','','','','','1170000000','1170000000');
insert into R_TOKN_MAIN values ('access_type',1010,'execute','','','','','1170000000','1170000000');
insert into R_TOKN_MAIN values ('access_type',1020,'read annotation','','','','','1170000000','1170000000');
insert into R_TOKN_MAIN values ('access_type',1030,'read system metadata','','','','','1170000000','1170000000');
insert into R_TOKN_MAIN values ('access_type',1040,'read metadata','','','','','1170000000','1170000000');
insert into R_TOKN_MAIN values ('access_type',1050,'read object','','','','','1170000000','1170000000');
insert into R_TOKN_MAIN values ('access_type',1060,'write annotation','','','','','1170000000','1170000000');
insert into R_TOKN_MAIN values ('access_type',1070,'create metadata','','','','','1170000000','1170000000');
insert into R_TOKN_MAIN values ('access_type',1080,'modify metadata','','','','','1170000000','1170000000');
insert into R_TOKN_MAIN values ('access_type',1090,'delete metadata','','','','','1170000000','1170000000');
insert into R_TOKN_MAIN values ('access_type',1100,'administer object','','','','','1170000000','1170000000');
insert into R_TOKN_MAIN values ('access_type',1110,'create object','','','','','1170000000','1170000000');
insert into R_TOKN_MAIN values ('access_type',1120,'modify object','','','','','1170000000','1170000000');
insert into R_TOKN_MAIN values ('access_type',1130,'delete object','','','','','1170000000','1170000000');
insert into R_TOKN_MAIN values ('access_type',1140,'create token','','','','','1170000000','1170000000');
insert into R_TOKN_MAIN values ('access_type',1150,'delete token','','','','','1170000000','1170000000');
insert into R_TOKN_MAIN values ('access_type',1160,'curate','','','','','1170000000','1170000000');
insert into R_TOKN_MAIN values ('access_type',1200,'own','','','','','1170000000','1170000000');
```

[...]

PR #6134, opened 2022-01-26
allow all access levels from icatDefines.h

erwinvanwieringen:
This will allow all access permissions from icatDefines.h to be used when setting permissions. Since 'write' and 'read' are still valid as permission, this will be fully backward compatible.
I had some doubt about the spaces in access permission names, but it feels a bit silly to translate them all to new names without spaces. So I left them as they are.

This will work with all clients that depend on the irods api to check if a specified access permission is valid. Tested with ichmod, msiSetAcl, and the python-irodsclient.

Example usage: `ichmod "modify metadata" 'user1' /tempZone/data`

Resolves irods/irods#844

### DIFF
--- a/plugins/database/src/db_plugin.cpp
+++ b/plugins/database/src/db_plugin.cpp
@@ -10136,12 +10136,24 @@ irods::error db_mod_access_control_op(
         adminMode = 1;
     }
 
-    static const std::vector<char const *> allowed_access_levels { ACCESS_EXECUTE, ACCESS_READ_ANNOTATION, 
-        ACCESS_READ_SYSTEM_METADATA, ACCESS_READ_METADATA, ACCESS_READ_OBJECT, ACCESS_WRITE_ANNOTATION, 
-        ACCESS_CREATE_METADATA, ACCESS_MODIFY_METADATA, ACCESS_DELETE_METADATA, ACCESS_ADMINISTER_OBJECT, 
-        ACCESS_CREATE_OBJECT, ACCESS_MODIFY_OBJECT, ACCESS_DELETE_OBJECT, ACCESS_CREATE_TOKEN, 
-        ACCESS_DELETE_TOKEN, ACCESS_CURATE, ACCESS_OWN };
-    const char *myAccessLev = NULL;
+    static const std::vector<char const*> allowed_access_levels{ACCESS_EXECUTE,
+                                                                ACCESS_READ_ANNOTATION,
+                                                                ACCESS_READ_SYSTEM_METADATA,
+                                                                ACCESS_READ_METADATA,
+                                                                ACCESS_READ_OBJECT,
+                                                                ACCESS_WRITE_ANNOTATION,
+                                                                ACCESS_CREATE_METADATA,
+                                                                ACCESS_MODIFY_METADATA,
+                                                                ACCESS_DELETE_METADATA,
+                                                                ACCESS_ADMINISTER_OBJECT,
+                                                                ACCESS_CREATE_OBJECT,
+                                                                ACCESS_MODIFY_OBJECT,
+                                                                ACCESS_DELETE_OBJECT,
+                                                                ACCESS_CREATE_TOKEN,
+                                                                ACCESS_DELETE_TOKEN,
+                                                                ACCESS_CURATE,
+                                                                ACCESS_OWN};
+    const char* myAccessLev = NULL;
     int rmFlag = 0;
     int inheritFlag = 0;
     if ( strcmp( _access_level, AP_NULL ) == 0 ) {
@@ -10161,19 +10173,18 @@ irods::error db_mod_access_control_op(
         inheritFlag = 2;
     }
     else {
-        for ( char const *level: allowed_access_levels ) {
-            if ( !strcmp( _access_level, level ) ) {
+        for (char const* level : allowed_access_levels) {
+            if (!strcmp(_access_level, level)) {
                 myAccessLev = level;
                 break;
             }
         }
-        if ( myAccessLev == NULL ) {
+        if (myAccessLev == NULL) {
             char errMsg[105];
-            snprintf( errMsg, 100, "access level '%s' is invalid",
-                    _access_level );
-            addRErrorMsg( &_ctx.comm()->rError, 0, errMsg );
-            return ERROR( CAT_INVALID_ARGUMENT, errMsg );
-        }       
+            snprintf(errMsg, 100, "access level '%s' is invalid", _access_level);
+            addRErrorMsg(&_ctx.comm()->rError, 0, errMsg);
+            return ERROR(CAT_INVALID_ARGUMENT, errMsg);
+        }
     }
 
     if ( !icss.status ) {
--- a/plugins/database/src/db_plugin.cpp
+++ b/plugins/database/src/db_plugin.cpp
@@ -10136,7 +10136,12 @@ irods::error db_mod_access_control_op(
         adminMode = 1;
     }
 
-    char *myAccessLev = NULL;
+    static const std::vector<char const *> allowed_access_levels { ACCESS_EXECUTE, ACCESS_READ_ANNOTATION, 
+        ACCESS_READ_SYSTEM_METADATA, ACCESS_READ_METADATA, ACCESS_READ_OBJECT, ACCESS_WRITE_ANNOTATION, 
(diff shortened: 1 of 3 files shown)


alanking (changes requested):
First of all, thanks for the contribution!

Before merging, we will need to think through the implications of this change, but this definitely looks like the right direction.

It would be good if we could get some automated tests for the newly introduced options in order to instill confidence in the changes. One of us on the iRODS team can do this, or you can try to add some tests if you'd like. :) I think `test_ichmod.py` would be an appropriate place to add these. Here is an example for how this could be started:
```python
    def test_ichmod_modify_metadata(self):
        filename = 'test_ichmod_modify_metadata'
        local_file_path = os.path.join(self.admin.local_session_dir, filename)
        logical_path = os.path.join('/' + self.admin.zone_name, 'home', 'public', filename)

        # make a local file
        lib.make_file(local_file_path, 1)

        # put it as a data object in a collection where another user will have permissions
        self.admin.assert_icommand(['iput', local_file_path, logical_path])

        # ensure that the user cannot do anything with the data object
        self.user.assert_icommand_fail(['iget', logical_path, '-'], 'STDERR', 'CAT_NO_ACCESS_PERMISSION')

        # change permissions to "modify metadata"
        self.admin.assert_icommand(['ichmod', 'modify metadata', self.user.username, logical_path])

        # ensure that the user can modify metadata and everything below that (read, etc.)
        self.user.assert_icommand(['iget', logical_path, '-'], 'STDOUT', '0')
        # TODO: add some metadata
        # TODO: modify metadata
        # TODO: etc.

        # ensure that the user cannot do anything that requires ACL higher than "modify metadata"
        self.user.assert_icommand_fail(['iput', '-f', local_file_path, logical_path], 'STDERR', 'CAT_NO_ACCESS_PERMISSION')
        # TODO: try to irm
        # TODO: etc.
```

Also, the spaces in the permission names is actually something we are considering changing, and it would be best to do so in a .0 release (in other words, 4.3.0). The team will need to think about the implications of such a change, but we have agreed internally, at least initially, that the changes would be good to do in the same release. That said, the removal of the spaces doesn't need to happen in this commit.

plugins/database/src/db_plugin.cpp
```diff
@@ -10149,8 +10154,9 @@ irods::error db_mod_access_control_op(
     else if ( strcmp( _access_level, AP_WRITE ) == 0 ) {
         myAccessLev = ACCESS_MODIFY_OBJECT;
     }
-    else if ( strcmp( _access_level, AP_OWN ) == 0 ) {
-        myAccessLev = ACCESS_OWN;
+    else if (std::find(allowed_access_levels.begin(), allowed_access_levels.end(), _access_level) != allowed_access_levels.end()) {
+            myAccessLev = (char *)malloc(sizeof(char) * strlen(_access_level)+1);
```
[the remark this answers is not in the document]
erwinvanwieringen: I fixed this in 4cf2359. Changed `myAccesLev` to `const char*` as you suggested, and now assign a ref to a string literal form icatDefines.h

erwinvanwieringen:
I added some tests, as you suggested. I now test a couple of operations (imeta get/set/rm, iput, iget, irm, imkdir, irmdir) with various access levels. There were some suprising results (why do you need `modify object` to create an object? You would think `create object` would suffice), but that may be caused by my interpretation of how the permission model should work. I now coded the tests so they all pass.

I added a large dictionary called `test_commands` here to configure the tests. This should be easy to extend, if you want to add extra permission related tests.

Permission levels that are defined in `icatDefines.h`, but are not in use anywhere in the code (like `ACCESS_READ_ANNOTATION`) are now not included in the tests.

scripts/irods/test/test_ichmod.py
```diff
@@ -28,6 +29,180 @@ def test_ichmod_r_with_garbage_path(self):
...
+            # Test that iget needs READ OBJECT
+            'iget': {
+                'cmd': ['iget' , '-f', test_object, user_iget_file],
+                'checks': [
+                    (Access.NULL, Access.READ_METADATA,  ['STDERR', 'USER_INPUT_PATH_ERR']),
+                    (Access.READ_METADATA, Access.OWN,   ['STDERR', ''])
+                ]
+            },
+            # Test that iput needs MODIFY_OBJECT on the collection
+            # NB why is CREATE OBJECT not enough?
```
alanking, line 90: Agreed that this seems wrong. This is likely an artifact of the fact that the interface for changing permissions has been the way that it is, and the reason this change exists.

alanking (commented):
This is great!

We really like the framework presented in your test case. We reviewed this and found that it has illuminated some dark corners surrounding the permissions model, which is in large part why we have never bothered to look at all this stuff.

In particular, your paring down of the permissions to 10 from the original 18 seems good. The unused permissions will likely continue to be unused. We have some more conversations to have before this is settled.

And as far as writing the tests to make them pass, that is fine for now. We are of the opinion that the permission model may need to conform to what makes sense in the future, but this is good for now.

How do you feel about us pushing changes on top of your branch (currently enabled via Github)? You will receive credit for your work, but there are some things we may add as part of our process before merging.

I left a couple of other comments for your consideration. Thanks again for your work!

erwinvanwieringen:
Break statement has been added as suggested. Also added tests for the original **read** and **write** permission levels. 
If you want to push changes on top of this branch, that is fine with me!

alanking:
We'd like to get this change in as soon as possible, but we are working to get 4.3.0 solidified in the run up to its release. So, for now, this change looks good, but we'd like to hold off on merging it right now as we work on other tasks and evaluate the discoveries made in the permissions model. As we approach the release, we will come back to this and consider it for inclusion in 4.3.0. Thanks!

trel:
we're looping back around to get this merged this week.

trel:
ready for review - will # when blessed.

scripts/irods/test/test_ichmod.py
```diff
@@ -28,6 +29,183 @@ def test_ichmod_r_with_garbage_path(self):
...
+            Access.CREATE_METADATA : "create_metadata",
+            Access.MODIFY_METADATA : "modify_metadata",
+            Access.DELETE_METADATA : "delete_metadata",
+            Access.CREATE_OBJECT   : "create_object",
+            Access.MODIFY_OBJECT   : ["modify_object", "write"],
+            Access.DELETE_OBJECT   : "delete_object",
+            Access.OWN             : "own"
+        }
+
+        # Here the test commands we test with every acces level are defined:
```
korydraughn, line 75: acces -> access

korydraughn, line 75: acces -> access

trel: done

scripts/irods/test/test_ichmod.py
```diff
@@ -28,6 +29,183 @@ def test_ichmod_r_with_garbage_path(self):
...
+        # *arguments will be passed to the assert_icommands function
+        #
+        #
+        test_commands = {
+            # Test that iget needs READ OBJECT
+            'iget': {
+                'cmd': ['iget' , '-f', test_object, user_iget_file],
+                'checks': [
+                    (Access.NULL, Access.READ_METADATA,  ['STDERR', 'USER_INPUT_PATH_ERR']),
+                    (Access.READ_METADATA, Access.OWN,   ['STDERR', ''])
```
trel, line 88: should this be READ_OBJECT?

korydraughn: I think you're correct. And `STDERR` should be removed because `ichmod` doesn't produce output on success.

trel: well, this is `iget`...

korydraughn: Whoops. You're right. Ignore me lol

alanking:
Found an issue related to tickets while testing these changes, still investigating...

trel:
i broke the tickets with a bad clang-format - i think we're back... hoping to merge soon.

scripts/irods/test/test_ichmod.py
alanking: I think this should be `Access.READ_OBJECT`, yes. This test currently passes - I *think* - because it checks for anything or nothing being on stderr, which will always be the case. So, the suggested edit by @korydraughn should be the right thing to do regardless.

trel: done - now incorporated.

trel:
#'d

Issue #6150, opened 2022-02-02
Investigate moving `irods::experimental::authentication_base` out of `experimental` namespace


alanking:
Related to ongoing work in #5479.

Wondering whether it makes sense to leave the authentication plugin base in the `experimental` namespace as any changes to the interface or even to the flow of an individual will break a particular version's ability to authenticate for other servers and clients.

alanking:
I have found that the plugin interface itself will display as `irods::experimental::authentication_base` when resolving (based on the original implementation), so that's one more argument for moving this out of `experimental`.

trel:
it *is* a bit in-your-face.

alanking:
See comment here: https://github.com/irods/irods/pull/6151#discussion_r798966229

It invalidates my previous comment, I think (https://github.com/irods/irods/issues/6150#issuecomment-1029246519)

However, I think we should still consider this.

alanking:
Seeing as we have never *removed* anything from the experimental namespace (as far as I know), I think this can wait. Thoughts, @korydraughn / @trel?

korydraughn:
I agree.

Having it in the experimental namespace isn't hurting anything. We can circle back and move it out in a later release. On top of that, moving it out after it has been in use for a while will inform others that the API is solid.

trel:
to remove would be cleaner, less scary for readers of logs...

but... auth probably does have a pretty good chance of needing a tweak later...

agreed - let's leave it for now.

alanking:
Gonna do this as part of the work for https://github.com/irods/irods/issues/8257

Issue #5479, opened 2021-03-09
Refactor authentication plugin interface leveraging Authentication Working Group design


jasoncoposky:
A new authentication plugin interface has been created via the [AWG](https://github.com/irods-contrib/irods_working_group_authentication/tree/master/development).  This new mechanism will be ported to the main iRODS repository.

Consider backwards compatibility with older clients as well as federation.

- [x] Native
- [x] PAM
- [x] Kerberos

The following will not be carried forward into 4.3.x (see https://groups.google.com/g/iROD-Chat/c/sSlBMvI-21o):
- OS Auth
- GSI

alanking:
Just leaving a note here for tracking progress...

- [x] Pulled all code from AWG into irods/irods
- [x] Code builds
- [x] Wired native authentication plugin into `clientLogin` so that iCommands use new authentication method

Now I'm experiencing `SYS_UNMATCHED_API_NUM` when attempting to start a catalog consumer server.

trel:
That's... allow list?

trel:
ah.

aka, too late.

trel:
but also, so that it's dynamic...     hmm...

alanking:
The issue with CSC servers has been resolved. The client API table needs to be loaded when standing up the server in order to use the authentication plugins when querying the CSP for information.

Now investigating server authentication failures across federation.

alanking:
I broke the following unit tests as they are not loading the client API plugins:
```
==== begin test run results ====
-----
results for [ubuntu-1804-postgres-1012_irods-catalog-provider_1]
        test list:
                [irods_client_connection]
                [irods_connection_pool]
                [irods_query_builder]
                [irods_resource_administration]
                [irods_user_administration]
                [irods_zone_report]
        failed tests:
                [irods_client_connection]
                [irods_connection_pool]
                [irods_query_builder]
                [irods_resource_administration]
                [irods_user_administration]
                [irods_zone_report]
        return code:[1]
-----
List of failed tests:
        irods_connection_pool irods_query_builder irods_resource_administration irods_user_administration irods_zone_report
Return code:[1]
==== end of test run results ====
```

alanking:
At the most recent AWG meeting (https://github.com/irods-contrib/irods_working_group_authentication/blob/main/20220125-minutes.md), we discussed implementing the legacy PAM auth plugin using the new plugin framework. This is now done.

[...]

PR #6151, opened 2022-02-02
[#5479] Add authentication API plugin and framework (main)

alanking:
This change adds a new authentication plugin framework. The basis for
this work was originally developed for the Authentication Working Group
and its implementation and documentation can be found here:

  https://github.com/irods-contrib/irods_working_group_authentication

The framework provides a single function called authenticate_client
which allows a connected client user to be authenticated. The only
plugin which has been implemented using the new framework so far is the
iRODS native authentication. The legacy authentication plugin will
remain in place in order to support talking to older servers with newer
clients and servers.

This change also adds a new API plugin which loads an auth plugin and
calls the operation specified by "next_operation" in the JSON input.
This is used by the authentication plugins within the framework to make
requests and receive responses from the server.

The new authentication plugin framework has been integrated with
clientLogin so that all authentication requests are routed through the
new framework. Authentication requests for clients connected to servers
before 4.3.0 will use the older authentication flow.

To support this version sniffing, a method called to_version was moved
to the version library which takes a string and constructs an
irods::version instance. If the instance cannot be constructed due to a
malformed input or an error, std::nullopt is returned. Also adds unit
tests for the new function.

As a result of using the authentication framework, a new requirement for
server startup has been introduced: API plugins must be loaded at server
startup in order for the server to start. So, part of this change also
loads the client and server API tables on server startup so that any
server startup operations which require API plugins can be performed.
The delay server also requires that the client API table be loaded on
startup so that any API plugins required for communication with the
server can be loaded.

---

CI tests to be run. Confirmed that authentication between servers in the same zone and across federation is working.

Leaving in draft form for now because I'm still working on cleanup. I have a separate branch where individual chunks of work are still separated, so let me know if the single commit should be broken up.

Also, opened #6150 to address the question of namespacing.


trel:
consider the moving of 'to_version' as a separate commit.

everything else seems like a single unit of work.

alanking:
Sounds good to me. I'll force-push my two shiny commits and mark as ready for review once I can confirm that things are building

alanking:
Confirmed that changes are building and runs. Still need to run tests

CMakeLists.txt
```diff
@@ -715,6 +715,7 @@ target_link_libraries(
   ${IRODS_EXTERNALS_FULLPATH_FMT}/lib/libfmt.so
   OpenSSL::SSL
   OpenSSL::Crypto
+  ${CMAKE_DL_LIBS}
```
trel, line 718: is this in the right commit?

alanking: Hmmm... no. No, it is not.

lib/core/src/clientLogin.cpp
```diff
@@ -260,6 +303,20 @@ int clientLogin(
         } // if _scheme_override
     } // if client side auth
 
+    static constexpr auto irods_430 = irods::version{4, 3, 0};
+    const auto server_version = irods::to_version(_comm->svrVersion->relVersion);
+    if (!server_version) {
+        std::cerr << "Failed to get version from server ["
+                  << _comm->svrVersion->relVersion
+                  << "]\n";
```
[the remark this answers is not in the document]
korydraughn: should these functions be writing to stdout/stderr at all?

plugins/api/src/authenticate.cpp
```diff
@@ -0,0 +1,141 @@
...
+    const operation op{};
+    #define CALL_AUTHENTICATE NULL
+}
+
+#endif // RODS_SERVER
+
+extern "C" {
+    irods::api_entry* plugin_factory(const std::string&, const std::string&)
+    {
+        #ifdef RODS_SERVER
```
korydraughn, line 123: consider not aligning preprocessor conditionals with runtime code.

my reasoning behind this is based on the idea that preprocessor code is not part of the main logic and we should do things to make that obvious.

willing to discuss if that statement isn't clear.

server/core/src/irodsReServer.cpp
```diff
@@ -1,35 +1,36 @@
-#include "connection_pool.hpp"
+#include "catalog.hpp"
 #include "client_connection.hpp"
+#include "connection_pool.hpp"
 #include "initServer.hpp"
+#include "irodsReServer.hpp"
```
korydraughn, line 5: consider moving this header file to the top and adding a new line after it.

that helps to signal that this cpp file is the implementation of irodsReServer.hpp and everything else is a dependency.

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
```diff
@@ -0,0 +1,467 @@
+#include "authentication_plugin_framework.hpp"
+
+#include "base64.h"
+
+int get64RandomBytes( char *buf );
+void setSessionSignatureClientside( char* _sig );
+void _rsSetAuthRequestGetChallenge( const char* _c );
```
korydraughn, line 7: we should probably add a comment explaining from what header file these functions come from.

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
```diff
@@ -0,0 +1,467 @@
...
+namespace
+{
+    using json = nlohmann::json;
+
+    auto request(rcComm_t* comm, const json& msg)
+    {
+        auto str = msg.dump();
+
+        bytesBuf_t inp;
+        inp.buf = (void*)str.c_str();
```
korydraughn, line 18: consider using a c++ cast instead of a c-style cast.

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
```diff
@@ -0,0 +1,467 @@
...
+        }
+
+        // A user is allowed to authenticate itself.
+        if (std::string_view{_comm->proxyUser.userName} == _comm->clientUser.userName) {
+            return;
+        }
+
+        // A proxy user with remote priviliged user authorization can only authenticate client
+        // users from the same zone.
+        if (_proxy_priv >= REMOTE_PRIV_USER_AUTH &&
```
korydraughn, line 70: consider adding comments to the header file that defines `REMOTE_PRIV_USER_AUTH` and friends.

we really need to document what these mean.

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
```diff
@@ -0,0 +1,467 @@
...
+            std::string_view{_comm->proxyUser.rodsZone} == _comm->clientUser.rodsZone) {
+            return;
+        }
+
+        // This is a proxy user attempting to authenticate a client user with insufficient
+        // permissions to do so.
+        THROW(SYS_PROXYUSER_NO_PRIV, fmt::format(
+              "proxy user: [{}] with {} no priv to auth client user: [{}].",
```
korydraughn, line 78: this statement is a little confusing. how about:

"Proxy user [{}] is not allowed to authenticate client user [{}]."

i'm not sure how to include the privilege level. is it needed?

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
```diff
@@ -0,0 +1,467 @@
...
+namespace irods
+{
+    class native_authentication : public irods::experimental::authentication_base {
+    public:
+        native_authentication()
+        {
+            add_operation(AUTH_ESTABLISH_CONTEXT,    OPERATION(rcComm_t, native_auth_establish_context));
+            add_operation(AUTH_CLIENT_AUTH_REQUEST,  OPERATION(rcComm_t, native_auth_client_request));
+            add_operation(AUTH_CLIENT_AUTH_RESPONSE, OPERATION(rcComm_t, native_auth_client_response));
+            #ifdef RODS_SERVER
```
korydraughn, line 93: consider not aligning with runtime logic.

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
```diff
@@ -0,0 +1,467 @@
...
+            //  [50+1] obfuscated user password
+            char md5_buf[CHALLENGE_LEN + MAX_PASSWORD_LEN + 2]{};
+            strncpy(md5_buf, request_result.c_str(), CHALLENGE_LEN);
+
+            // Save a representation of some of the challenge string for use as a session signature
+            setSessionSignatureClientside(md5_buf);
+
+            // determine if a password challenge is needed, are we anonymous or not?
+            bool need_password = false;
+            if (req.at("user_name") == ANONYMOUS_USER) {
```
korydraughn, line 136: let's use an explicit cast as recommended by the nlohmann::json library

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
```diff
@@ -0,0 +1,467 @@
...
+                tcgetattr( STDIN_FILENO, &tty );
+                tcflag_t oldflag = tty.c_lflag;
+                tty.c_lflag &= ~ECHO;
+                int error = tcsetattr( STDIN_FILENO, TCSANOW, &tty );
+                int errsv = errno;
+
+                if (error) {
+                    std::cout << "WARNING: Error "
+                              << errsv
+                              << " disabling echo mode. Password will be displayed in plaintext.\n";
```
korydraughn, line 156: fmt::print?

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
```diff
@@ -0,0 +1,467 @@
...
+                }
+                printf( "Enter your current iRODS password:" );
```
korydraughn, line 158: we should replace all use of printf.

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
```diff
@@ -0,0 +1,467 @@
...
+            }
+
+            unsigned char out[RESPONSE_LEN*2];
+            unsigned long out_len{RESPONSE_LEN*2};
+            auto err = base64_encode((unsigned char*)digest, RESPONSE_LEN, out, &out_len);
+            if(err < 0) {
+                THROW(err, "base64 encoding of digest failed.");
+            }
+
+            resp["digest"] = std::string{(char*)out};
```
korydraughn, line 191: this assumes `out` is null-terminated.

if the `out_len` represents the length of the string, we should pass it to the std::string constructor.

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
```diff
@@ -0,0 +1,467 @@
...
+            resp["next_operation"] = AUTH_CLIENT_AUTH_RESPONSE;
+
+            return resp;
+        } // native_auth_establish_context
+
+        json native_auth_client_request(rcComm_t* comm, const json& req)
```
korydraughn, line 197: all functions should take the `comm` by reference if the function requires a valid object.

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
```diff
@@ -0,0 +1,467 @@
...
+            auto resp = request(comm, svr_req);
+
+            comm->loggedIn = 1;
+
+            resp["next_operation"] = ie::auth::flow_complete;
+
+            return resp;
+        } // native_auth_client_response
+
+        #ifdef RODS_SERVER
```
korydraughn, line 228: preprocessor alignment

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
```diff
@@ -0,0 +1,467 @@
...
+
+            throw_on_insufficient_privilege_for_proxy_user(comm, authCheckOut->privLevel);
+
+            rodsLog(LOG_DEBUG,
+                    "rsAuthResponse set proxy authFlag to %d, client authFlag to %d, user:%s proxy:%s client:%s",
+                    authCheckOut->privLevel,
+                    authCheckOut->clientPrivLevel,
+                    authCheckInp.username,
+                    comm->proxyUser.userName,
+                    comm->clientUser.userName );
```
korydraughn, line 422: new logger?

korydraughn (commented):
very nice!

lib/core/include/authentication_plugin_framework.hpp
```diff
@@ -0,0 +1,155 @@
...
+        std::transform(lower.begin(), lower.end(), lower.begin(), ::tolower);
+
+        const auto name = fmt::format("irods_auth_plugin-{}_{}", lower, type);
+
+        plugin_type* plugin{};
+        auto err = irods::load_plugin<plugin_type>(
+                        plugin,
+                        name,
+                        irods::PLUGIN_TYPE_AUTHENTICATION,
+                        "irods::experimental::authentication_base",
```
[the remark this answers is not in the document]
alanking: Looking at this again... I don't know that this string actually used for this plugin interface. The `load_plugin` call is attempting to load a factory method from the plugin's shared object. In this case, the factory method is provided in the auth plugins (see irods_auth_plugin-native.cpp). So, this string could actually be anything because the two `const std::string&` parameters passed to the factory method in the plugin are unused (again, see the native plugin implementation).

After a quick look around, it looks like most of the plugins are not using these parameters, but it is part of the interface for constructing the `irods::plugin_base`. We are not using that in this instance, so I think this string can be whatever.

lib/core/include/authentication_plugin_framework.hpp
```diff
@@ -0,0 +1,155 @@
...
+#ifdef RODS_SERVER
+#include "rsAuthCheck.hpp"
+#include "rsAuthRequest.hpp"
+#endif // RODS_SERVER
+
+#include <sstream>
+#include <string>
+#include <iostream>
+#include <termios.h>
+#include <unistd.h>
```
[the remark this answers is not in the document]
alanking: Like this?

```diff
diff --git a/lib/core/include/authentication_plugin_framework.hpp b/lib/core/include/authentication_plugin_framework.hpp
index 29d2eb5f1..6b4f30df8 100644
--- a/lib/core/include/authentication_plugin_framework.hpp
+++ b/lib/core/include/authentication_plugin_framework.hpp
@@ -23,17 +23,18 @@
 #include "rsAuthRequest.hpp"
 #endif // RODS_SERVER
 
+#include <iostream>
 #include <sstream>
 #include <string>
-#include <iostream>
+
 #include <termios.h>
 #include <unistd.h>
 
 #include <boost/any.hpp>
 
+#include <fmt/format.h>
 #include <nlohmann/json.hpp>
 #include <openssl/md5.h>
-#include <fmt/format.h>
 
 using json = nlohmann::json;
```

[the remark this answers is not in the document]
trel: and is boost special here?

lib/core/src/clientLogin.cpp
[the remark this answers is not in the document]
alanking: The authentication plugin should because this is where the password prompt is coming from for native authentication in the absence of an `.irodsA` file. Some of the other error messages (like this one) are only relevant should the calls be in a client (i.e. not a server).

I'll think about whether this should include a log message in addition to the error message. I think it's helpful for the user to know why the authentication failed.

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
alanking: Plot twist: these functions should actually be `extern`'d because they are declared and defined in `.cpp` files. Does this seem correct?

lib/core/include/authentication_plugin_framework.hpp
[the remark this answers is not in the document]
korydraughn: @alanking yes. i just noticed that `<openssl/md5.h>` is grouped with the headers provided by externals. the openssl header should be separated from those.

@trel i normally group boost with fmt and json when it is just a single boost header. when multiple boost headers are included, i group those together.

the scheme i use is:
1. primary header (this is the header file that shares the same name as the cpp file)
2. iRODS headers (support headers)
3. externals headers
4. headers installed via the package mgr
5. POSIX or OS headers
6a. C++ C headers (e.g. <ctime>)
6b. C++ headers

all groups are separated by a new line. 6a and 6b are grouped together (i just like to put C headers before C++ headers).

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
korydraughn: i don't think they need the extern keyword since they are declarations.

i believe extern matters when dealing with variables.

lib/core/include/authentication_plugin_framework.hpp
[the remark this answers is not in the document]
alanking: To the developer docs!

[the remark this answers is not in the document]
korydraughn: good idea

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
alanking: Indeed, it's not needed.

On that note, do you still think the comments are needed?

lib/core/include/authentication_plugin_framework.hpp
```diff
@@ -0,0 +1,155 @@
...
+    void authenticate_client(rcComm_t* comm, const rodsEnv& env)
+    {
+        // example native authentication scheme: irods-authentication_plugin-native
+        std::string scheme{env.rodsAuthScheme};
+
+        // TODO:: make some decisions about auth scheme?
+
+        auto auth = resolve_authentication_plugin(scheme, "client");
+
+        std::string next_operation{irods::AUTH_CLIENT_START};
```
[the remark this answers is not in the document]
alanking: Would a ref count?

lib/core/include/irods_plugin_base.hpp
```diff
@@ -325,14 +325,14 @@ namespace irods
         template< typename T >
         error get_property( const std::string& _key, T& _val ) {
             error ret = properties_.get< T >( _key, _val );
-            return ASSERT_PASS( ret, "Failed to get property for auth plugin." );
+            return ASSERT_PASS( ret, "Failed to get property for plugin." );
```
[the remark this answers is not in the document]
alanking: I believe I have actually addressed this in a branch wherein I am removing `ASSERT_PASS` from the codebase. I will remove this change from this PR and just handle it there. Thanks!

lib/core/include/irods_plugin_base.hpp
```diff
@@ -325,14 +325,14 @@ namespace irods
...
         } // get_property
 
         /// @brief set a property in the map
         template< typename T >
         error set_property( const std::string& _key, const T& _val ) {
             error ret = properties_.set< T >( _key, _val );
-            return ASSERT_PASS( ret, "Failed to set property in the auth plugin." );
+            return ASSERT_PASS( ret, "Failed to set property in the plugin." );
```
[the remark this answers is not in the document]
alanking: Removing this change. See other comment

lib/core/include/authentication_plugin_framework.hpp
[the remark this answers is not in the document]
korydraughn: a reference wouldn't work because once it is initialized, you can't change what it points to.

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
alanking: Perhaps something like "Requires auth level [{}] but proxy user has [{}]." ?

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
alanking: Will do

lib/core/include/authentication_plugin_framework.hpp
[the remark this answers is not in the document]
alanking: Ah, I've never used a ref in that way, but that makes sense. Will do

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
alanking: Will do

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
korydraughn: i'm kind of torn on this.

these forward declarations mean we now have to maintain them in at least two places. comments can also become stale. if the header is included, we don't have to worry about it.

however, i get why someone would do this. maybe jason wanted to avoid including a large header file or this was just easier. avoiding including of a large file isn't a good reason to do this given how fast computers are.

plugins/api/src/authenticate.cpp
alanking: Will do

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
alanking: These functions are declared *and* defined in `.cpp` files. They are not declared in header files.

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
alanking: Will do

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
alanking: I gave it my best attempt :)

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
alanking: Will do

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
alanking: Will do

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
alanking: Will do

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
alanking: Will do

server/core/src/irodsReServer.cpp
alanking: Will do

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
korydraughn: Oh no. We need to fix that.

Can handle that in a separate PR.

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
alanking: Will do

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
alanking: I created a new issue for this: #6152

lib/core/include/authentication_plugin_framework.hpp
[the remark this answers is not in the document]
alanking: What do you think about using the string "authentication_plugin_base" here? We could also just pass an empty string since it's not being used for these plugins. None of the other plugins are namespacing, but maybe there was a purpose for this in the original implementation?

lib/core/src/clientLogin.cpp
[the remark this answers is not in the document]
alanking: I'll instead add this to the `rError` stack in `_comm` and log it to capture the client and server sides of the conversation.

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
alanking: I'll just exclude the privilege lever. I don't think we need it in this message.

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
alanking: Would it do to use `req.at("user_name").get_ref<std::string&>()`?

lib/core/include/rodsUser.h
```diff
@@ -22,7 +22,16 @@
 #define PUBLIC_USER_AUTH        1       /* not authenticated yet */
 #define REMOTE_USER_AUTH        2       /* authenticated as remote user */
 #define LOCAL_USER_AUTH         3       /* authenticated as local user */
+// Authenticate as a privileged remote user. This authentication level is reserved for iRODS
+// users from a remote zone which have been given administrative privileges in the local zone.
+// The authenticated privileged remote user can perform actions on behalf of a user from the
+// same remote zone in the local zone. Users with this level of authentication are not allowed
+// to perform actions on behalf of users from other zones and cannot perform certain privileged
+// actions in the local zone, which would require authentication level LOCAL_PRIV_USER_AUTH.
```
alanking, line 30: Any thoughts on these words? I didn't do the other auth levels as I was only dealing with the last two, but I can put words for the others if desired.

lib/core/include/authentication_plugin_framework.hpp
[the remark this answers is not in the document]
korydraughn: if nothing is using it, maybe we can re-purpose it for debugging or something else.

i don't think we should pass an empty string. the string you suggested works. does it appear in the log file?

alternative ideas:
- `"authentication_plugin_framework"`: indicator for what the plugin is related to
- build a string using info about the target plugin

lib/core/src/clientLogin.cpp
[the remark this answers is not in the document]
korydraughn: sounds good to me.

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
korydraughn: having a log message is good. i think we should leave the privilege level out for now. we can always revisit it if it becomes a problem.

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
korydraughn: that is correct. using the `get<T>()` member functions is the explicit form for the library.

you'll need to add a `const` to that template argument though.

alanking: Cool, thanks!

lib/core/include/rodsUser.h
korydraughn: should the first word be "Authenticated" instead of "Authenticate"?

these values are after authentication has been done, correct?

korydraughn: other than that, i think it is very clear.

lib/core/include/authentication_plugin_framework.hpp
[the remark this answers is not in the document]
alanking: Hmm... I dug... again... and found that this is ultimately being piped into the `instance_name_` member of the `plugin_base` in instances where this is being used (e.g. the database plugin). So maybe we could just use the `name` of the plugin for this parameter as well, which would be bullet 2 from your suggestions. Thoughts?

plugins/auth/irods_auth_plugin-native/libirods_auth_plugin-native.cpp
alanking: Agreed

lib/core/include/rodsUser.h
alanking: Ah, yes. I think that's right. I'll change the word to past tense.

lib/core/include/authentication_plugin_framework.hpp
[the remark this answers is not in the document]
korydraughn: sounds fine to me.

alanking:
I think I've resolved all of the review comments so far. Tests are not passing, though, so need to figure out what that's all about.

alanking:
I've identified a few tests which are failing in main as of some recent changes so the latest commit fixes these. The tests failing due to my changes are because the client API table is not being loaded in `irule` when the rule text is executed directly. This has prompted a second PR in iCommands, found here: irods/irods_client_icommands#267.

Tests are running now

scripts/irods/test/test_iput_options.py
```diff
@@ -312,6 +312,18 @@ def tearDown(self):
...
         filename = 'test_iput_zero_length_file_with_purge_and_checksum_3883'
         lib.touch(filename)
-        self.user0.assert_icommand(['iput', '-R', 'compoundresc3883', '--purgec', '-k', filename])
-        self.user0.assert_icommand(['ils', '-L'], 'STDOUT_SINGLELINE', 'sha2:47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=')
+        logical_path = os.path.join(self.user0.session_collection, filename)
+
+        # Replicas on archive resources do not have checksums applied as of #6089 because the
+        # operation might be expensive or unfeasible depending on the underlying storage
+        # technology. It is incorrect to assume that the checksum matches that of the source
+        # replica, which has been done historically in the even of a DIRECT_ARCHIVE_ACCESS
```
korydraughn, line 320: even -> event

alanking: Yep

alanking:
The packing instructions need to be provided by the API plugin in order for XML protocol to work. I also changed the packing instructions to `BinBytesBuf_PI` from `BytesBuf_PI` to align with existing API plugins of similar signature.

A resource suite has passed so it's getting closer... Gonna mark this as draft since apparently changes are still ongoing

alanking:
All tests are passing now.

alanking:
If there aren't any other comments, I think this is good to go.

lib/core/include/authentication_plugin_framework.hpp
```diff
@@ -0,0 +1,155 @@
...
+        virtual json auth_client_start(rcComm_t& comm, const json& req) = 0;
+
+        template<typename RxComm>
+        void add_operation(const std::string& n, std::function<json(RxComm&, const json&)> f)
+        {
+            if(operations_.find(n) != operations_.end()) {
+                THROW(SYS_INTERNAL_ERR, fmt::format("operation already exists [{}]", n));
+            }
+
+            if(n.empty()) {
```
korydraughn, line 69: the if-blocks (including their bodies) on lines 65 and 69 seem like they should be swapped.

plugins/api/src/authenticate.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+        api->in_pack_key = "BinBytesBuf_PI";
+        api->in_pack_value = BytesBuf_PI;
+
+        api->out_pack_key = "BinBytesBuf_PI";
+        api->out_pack_value = BytesBuf_PI;
+
+        return api;
+    } // plugin_factory
+}; // extern "C"
```
korydraughn, line 145: semi-colon can be removed

korydraughn (approved):

plugins/api/src/authenticate.cpp
alanking: Yup

alanking:
Awaiting greens...

PR #6153, opened 2022-02-06
Enabled C++20 (main)

korydraughn:
Requires latest versions of externals packages. Specifically, fmt v8.1.1 due to https://github.com/fmtlib/fmt/commit/023c2018f78a66c9acb3b9c7522ec431a210eb8c.


server/core/src/irods_server_control_plane.cpp
```diff
@@ -139,11 +139,11 @@ namespace irods
         // copy binary encoding into a zmq message for transport
         zmq::message_t rep( data_to_send.size() );
         std::memcpy( rep.data(), data_to_send.data(), data_to_send.size() );
-        zmq_skt.send( rep );
+        zmq_skt.send( rep, zmq::send_flags::none );
 
         // wait for the server response
         zmq::message_t req;
-        zmq_skt.recv( &req );
+        [[maybe_unused]] const auto ec = zmq_skt.recv( req ); // FIXME
```
korydraughn, line 146: clean up

korydraughn:
Cleaned up commits.

alanking (approved):

plugins/api/src/atomic_apply_metadata_operations.cpp
```diff
@@ -175,7 +175,11 @@ namespace
                 break;
 
             default:
-                throw std::runtime_error{fmt::format("Invalid entity type specified [entity_type={}]", _entity_type)};
+                for (auto&& [k, v] : ic::entity_type_map) {
+                    if (_entity_type == v) {
+                        throw std::runtime_error{fmt::format("Invalid entity type specified [entity_type={}]", k)};
+                    }
```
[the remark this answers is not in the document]
korydraughn: Yes. Will update.

server/core/src/catalog_utilities.cpp
```diff
@@ -117,7 +117,11 @@ namespace irods::experimental::catalog
                 return irods::is_privileged_client(_comm);
 
             default:
-                log::database::error("Invalid entity type [entity_type => {}]", _entity_type);
+                for (auto&& [k, v] : entity_type_map) {
+                    if (_entity_type == v) {
+                        log::database::error("Invalid entity type [entity_type => {}]", k);
+                    }
+                }
```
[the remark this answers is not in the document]
korydraughn: Will update

plugins/api/src/atomic_apply_metadata_operations.cpp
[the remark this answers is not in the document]
korydraughn: done

server/core/src/catalog_utilities.cpp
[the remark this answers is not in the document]
korydraughn: done

CMakeLists.txt
```diff
@@ -34,7 +34,8 @@ IRODS_MACRO_CHECK_DEPENDENCY_SET_FULLPATH_ADD_TO_IRODS_DEVELOP_DEPENDENCIES_LIST
 string(REPLACE ";" ", " IRODS_PACKAGE_DEPENDENCIES_STRING "${IRODS_PACKAGE_DEPENDENCIES_LIST}")
 string(REPLACE ";" ", " IRODS_DEVELOP_DEPENDENCIES_STRING "${IRODS_DEVELOP_DEPENDENCIES_LIST}")
 
-set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_EXTENSIONS OFF)
```
korydraughn, line 38: should we disable compiler extensions?

what do we gain/lose by doing this?

CMakeLists.txt
```diff
@@ -136,6 +137,18 @@ set(IRODS_COMPILE_DEFINITIONS ${IRODS_COMPILE_DEFINITIONS}
...
+                              # has been removed in C++20. Verified that Boost 1.77 and Boost 1.78
+                              # do not handle this yet. Normally, defining BOOST_ASIO_HAS_ macros
+                              # should be avoided except when absolutely required as a workaround
+                              # for the latest version of a compiler or platform.
+                              #
+                              # See the following for more information:
+                              #
+                              #   https://www.boost.org/doc/libs/1_77_0/doc/html/boost_asio/using.html
+                              #
+                              BOOST_ASIO_HAS_STD_INVOKE_RESULT
```
korydraughn, line 151: this should be captured in an issue.

once boost is able to detect `std::invoke_result`, this macros should be removed.

unit_tests/CMakeLists.txt
```diff
@@ -25,9 +25,13 @@ list(INSERT CMAKE_MODULE_PATH 0 ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 include(utils)
 
 if (IRODS_UNIT_TESTS_ENABLE_ALL)
-    add_definitions(-DIRODS_ENABLE_ALL_UNIT_TESTS)
+    add_compile_definitions(IRODS_ENABLE_ALL_UNIT_TESTS)
 endif()
 
+add_compile_definitions(BOOST_ASIO_NO_DEPRECATED
+                        BOOST_SYSTEM_NO_DEPRECATED
+                        BOOST_ASIO_HAS_STD_INVOKE_RESULT)
```
korydraughn, line 33: see previous comment about `BOOST_ASIO_HAS_STD_INVOKE_RESULT`

CMakeLists.txt
korydraughn: @alanking @SwooshyCueb any thoughts on this?

korydraughn:
@trel this PR requires that the latest versions of externals packages be published to core-dev and unstable.

right now, no one can build this unless they build their own copy of the externals packages.

trel:
can you confirm only fmt8.1.1 is missing at the moment?

korydraughn:
checking ...

CMakeLists.txt
alanking: Disabling this feels like the right approach until we find that we need some extension.

The documentation states...

> Note that because most compilers enable extensions by default, this may expose portability bugs in user code or in the headers of third-party dependencies.

Found here: https://cmake.org/cmake/help/latest/manual/cmake-compile-features.7.html#availability-of-compiler-extensions

But I'm probably not the most trustworthy authority for that question :)

trel:
on it

CMakeLists.txt
korydraughn: i had the same thoughts around that.

unless @SwooshyCueb has any thoughts on it, i'm leaning towards disabling extensions. if we need them, we can re-enable them.

korydraughn:
waiting for build to succeed.

korydraughn:
builds succeeded.

added #'s

CMakeLists.txt
korydraughn: marking as resolved.

if we need extensions, we can re-enable them in a later commit.

trel:
before we merge/close this... did you capture your comments above as new issues?

korydraughn:
I was waiting until this is merged so I can reference the same code from the main branch. That is also the reason I haven't marked those as resolved.

I didn't want to link to code in the PR.

unit_tests/CMakeLists.txt
korydraughn: captured in issue.

CMakeLists.txt
SwooshyCueb: Apologies, this didin't show up in my notifications for some reason.

I agree with disabling extensions if we don't need them or have a specific reason to keep them enabled.
This isn't something I would typically do, myself, but I don't fault your logic here and I ultimately think it's the right choice for the project.

Issue #3727, opened 2017-11-10
Deprecate `ASSERT_ERROR` and `ASSERT_PASS`


alanking:
- [x] irods/irods
- [x] irods_resource_plugin_s3
- [x] irods_auth_plugin_kerberos
---
ASSERT_ERROR and ASSERT_PASS should be removed from the codebase to improve error handling and readability.

alanking:
The functions called from `ASSERT_ERROR` and `ASSERT_PASS` will actually be deprecated in 4.3.0 and then removed in a future release. I will update the issue title and make a new one for removal.

alanking:
Just need to merge S3 and Kerberos change and this can be closed.

PR #6155, opened 2022-02-07
[#3727] Remove `ASSERT_ERROR` and `ASSERT_PASS` usage (main)

alanking:
These macros and functions will be removed in the future, so their usage is removed in this change and deprecation attributed to their declarations for 4.3.0.

Tests still need to be run and plugins/clients evaluated


alanking:
All tests passed in testing environment.

plugins/auth_legacy/src/native.cpp
```diff
@@ -1,57 +1,59 @@
...
-        result = ASSERT_ERROR( proxyUserPriv >= LOCAL_PRIV_USER_AUTH ||
-                               ( proxyUserPriv >= REMOTE_PRIV_USER_AUTH &&
-                                 strcmp( rsComm->proxyUser.rodsZone, rsComm->clientUser.rodsZone ) == 0 ),
-                               SYS_PROXYUSER_NO_PRIV,
-                               "Proxyuser: \"%s\" with %d no priv to auth clientUser: \"%s\".",
-                               rsComm->proxyUser.userName, proxyUserPriv, rsComm->clientUser.userName );
-    }
+namespace
+{
+    auto check_proxy_user_privileges(rsComm_t *rsComm, const int proxyUserPriv) -> irods::error
```
alanking, line 35: Might consider moving the version of this function found in the new plugin to a common place and use it in both plugins.

plugins/network/src/ssl.cpp
```diff
@@ -26,6 +26,8 @@
 #include <openssl/x509v3.h>
 #include <openssl/err.h>
 
+#include "fmt/format.h"
```
alanking, line 29: Needs angle brackets

plugins/resources/src/unixfilesystem.cpp
```diff
@@ -20,6 +20,8 @@
 #include "irods/irods_logger.hpp"
 #include "irods/voting.hpp"
 
+#include "fmt/format.h"
```
alanking, line 23: Here, too

plugins/resources/src/unixfilesystem.cpp
```diff
@@ -302,125 +279,113 @@ irods::error unix_file_mkdir_r(
...
+irods::error unix_file_registered(irods::plugin_context& _ctx)
+{
     // Check the operation parameters and update the physical path
-    irods::error ret = unix_check_params_and_path( _ctx );
-    result = ASSERT_PASS( ret, "Invalid parameters or physical path." );
+    if (const auto err = unix_check_params_and_path(_ctx); !err.ok()) {
+        return PASSMSG("Invalid parameters or physical path.", err);
+    }
 
     // NOOP
```
alanking, line 290: How do we feel about potentially removing `unix_check_params_and_path` calls in the stub functions?

lib/core/src/irods_gsi_object.cpp
```diff
@@ -4,6 +4,8 @@
 #include "irods/irods_auth_manager.hpp"
 #include "irods/irods_auth_plugin.hpp"
 
+#include "fmt/format.h"
```
korydraughn, line 7: angle brackets

lib/core/src/irods_auth_manager.cpp
```diff
@@ -1,6 +1,34 @@
 #include "irods/irods_auth_manager.hpp"
 #include "irods/irods_auth_plugin.hpp"
 
+#include "fmt/format.h"
```
korydraughn, line 4: angle brackets

plugins/resources/src/deferred.cpp
```diff
@@ -16,6 +16,8 @@
 #include "irods/irods_kvp_string_parser.hpp"
 #include "irods/irods_random.hpp"
 
+#include "fmt/format.h"
```
alanking, line 19: Angle brackets

lib/core/src/irods_auth_manager.cpp
```diff
@@ -19,75 +47,43 @@ namespace irods {
...
-        error ret;
-        auth_ptr auth;
-
-        std::string type = _type;
-        if( CLIENT_PT == _proc_type ) {        
-            type += "_client";
-        }
-        else {
-            type += "_server";
+    auto auth_manager::init_from_type(const int&         _proc_type,
```
korydraughn, line 66: this should be copied instead of being passed by reference.

this can be handled in a separate PR.

lib/core/src/irods_krb_object.cpp
```diff
@@ -4,6 +4,8 @@
 #include "irods/irods_auth_manager.hpp"
 #include "irods/irods_auth_plugin.hpp"
 
+#include "fmt/format.h"
```
korydraughn, line 7: angle brackets

korydraughn (changes requested):
looks good

plugins/auth_legacy/src/native.cpp
korydraughn: one possibility is `irods_rs_comm_query.hpp`.

plugins/resources/src/unixfilesystem.cpp
korydraughn: what is the motivation behind removing them?

lib/core/src/irods_auth_manager.cpp
alanking: Good point. Created #6235 for this.

plugins/resources/src/unixfilesystem.cpp
korydraughn: yeah. if it wastes cycles and doesn't break anything then remove it.

alanking:
Tests passed in testing environment with new changes.

PR #6182, opened 2022-02-15
Split up `test_catalog`, package unit tests with server, fix broken unit tests (main)

alanking:
This only affects tests, and all affected tests passed.

PR #6186, opened 2022-02-16
Split up `test_catalog` and package unit tests with server (4-2-stable)

alanking:
cherry-picked from work for #6182

still need to test


alanking:
Ignoring the other Codacy issues as `self` is used by the `unittest` framework. Arguably, the underlying libraries for `assert_icommand` should not be using naked `assert`s but I'm not going to solve that here.

trel:
seems good after squashin'

trel:
please # and jam once green.

alanking:
#'d. Jammin'.

Issue #6174, opened 2022-02-14
`imeta cp` does not support relative paths


korydraughn:
- [x] main
- [x] tests for main branch
- [x] 4-2-stable
- [x] tests for 4-2-stable branch

PR #6196, opened 2022-02-21
[#6174] Added "imeta cp" test (4-2-stable)

korydraughn:


korydraughn:
passed at the bench and verified that the relative paths appear in the test output as expected.

korydraughn:
added #'s

alanking (approved):

PR #6206, opened 2022-02-24
[_6169] updates for metaclass, test generation (4-2-stable)

d-w-moore:
Use the `six` module's portable metaclass syntax (instead of \_\_metaclass\_\_ which won't work starting in Python3). More importantly we use `-r` in irule invocations to guarantee a nonzero rule code causes the tests generated from rule file directories to return the appropriate pass / fail status to the testing framework. We target:
  - native plugin instance for the [icommands rulefiles](https://github.com/irods/irods_client_icommands/tree/4-2-stable/test/rules), and
  - Python  plugin instance for the   Python rule plugin [rule files](https://github.com/irods/irods_rule_engine_plugin_python/tree/4-2-stable/python_rules)

### DIFF
--- a/scripts/irods/test/test_all_rules.py
+++ b/scripts/irods/test/test_all_rules.py
@@ -236,6 +236,7 @@ class Test_AllRules(with_metaclass(metaclass_unittest_test_case_generator.Metacl
                 "rulemsiSysReplDataObj",
                 "rulemsiNoChkFilePathPerm",
                 "rulemsiNoTrashCan",
+                "rulemsiSetRescSortScheme",
             ]
             for n in names_to_skip:
                 if n in rulefile:
--- a/scripts/irods/test/test_all_rules.py
+++ b/scripts/irods/test/test_all_rules.py
@@ -6,6 +6,7 @@ import sys
 import getpass
 import tempfile
 import json
+from six import with_metaclass
 
 if sys.version_info >= (2, 7):
     import unittest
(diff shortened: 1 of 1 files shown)


trel:
this seems fine for now - using six for 4-2-stable is good/necessary.

eventually test_all_rules.py can get refactored / go away / be absorbed by the other test suites.

trel:
so yes, please # it.

scripts/irods/test/test_all_rules.py
```diff
@@ -236,6 +236,7 @@ def filter_rulefiles(rulefile):
                 "rulemsiSysReplDataObj",
                 "rulemsiNoChkFilePathPerm",
                 "rulemsiNoTrashCan",
+                "rulemsiSetRescSortScheme.r",
```
trel, line 239: no .r ?

d-w-moore:
Uhm.... actually, there is an extra commit where incomplete tests are skipped.  (Existed on main but not here, until now - but that's how it was tested.)   Sorry about that. If that's cool , then I'll go ahead and # as requested.

scripts/irods/test/test_all_rules.py
d-w-moore: oops , ... right. Hmph.

trel:
squash it all down, get it running cleanly, and then # it.    yes please.

d-w-moore:
I'll test again with the .r taken out then.... on both branches.

PR #6241, opened 2022-03-15
[#6240] Declare package dependency on curl

### LINKED ISSUE #6240: irods runtime package needs curl dependency
Until #5884 is done, we need to declare a package dependency on libcurl from the runtime package.
For now, this should be done the same way as we do for `irods_microservice_plugins_curl`

SwooshyCueb:
Closes #6240


SwooshyCueb:
\#'d

Issue #3404, opened 2016-11-21
init_from_catalog should not error out on null genQueryOut


jasoncoposky:
- [x] If the results of the genQuery are null and no error was returned, set continueInx to 0 to exit the loop.

alanking:
I think this is done?

https://github.com/irods/irods/blob/d232c7f3e0154cacc3a115aa50e366a98617b126/server/core/src/irods_resource_manager.cpp#L310-L316

On line 315, `continueInx` is set to 0 when the genQuery results are null.

However, this been the case since before this issue was created, so I may be misunderstanding its purpose...

korydraughn:
We could replace the use of the GenQuery C API with the query iterator here.

trel:
oh yes, let's remove the ability to hold it wrong.

PR #6242, opened 2022-03-16
[#3404] Refactor resource_manager::init_from_catalog (main)

alanking:
Rewrites init_from_catalog to use the irods::query iterator for fetching
information about existing resources and populating the resource
manager.

---

All tests passed

### DIFF
--- a/server/core/include/irods/irods_resource_manager.hpp
+++ b/server/core/include/irods/irods_resource_manager.hpp
@@ -52,7 +52,7 @@ namespace irods
 
             // =-=-=-=-=-=-=-
             /// @brief  populate resource table from icat database
-            error init_from_catalog( rsComm_t* );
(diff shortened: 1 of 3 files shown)


alanking:
The main thing to focus on here is the fact that `process_init_results` was removed and a refactored implementation now exists in `init_from_catalog`.

alanking:
Hmm... maybe not that one. Perhaps a similar type of issue could be opened for it, though. I don't think that should go into 4.3.0. Thoughts?

trel:
No, not into 4.3.0... just thinking about how the data structure could be future-proofed for injection/update/overwrite later - and if that would influence this work.

Sounds like we're fine for now.

trel:
looks good to me

korydraughn (approved):

alanking:
#'d

PR #6243, opened 2022-03-16
[#4260] Enforce valid hostname for remote rule execution (main)

alanking:
Need to run full test suite

Issue #4260, opened 2019-02-08
Remote microservice does not error correctly with invalid hostname


alanking:
- [x] main
- [x] 4-2-stable
---
### iRODS Version, OS and Version
master

### What did you try to do?
```
#my_rule.r
my_rule {
    remote("badhostname", "<ZONE>tempZone</ZONE>") {
        writeLine("stdout", "Remote block");
    }
}
INPUT null
OUTPUT ruleExecOut
```

### Expected behavior
```
$ irule -F my_rule.r 
remote addresses: 172.17.0.3 ERROR: rcExecMyRule error.  status = -303000 USER_RODS_HOSTNAME_ERR
```

### Observed behavior (including steps to reproduce, if applicable)
```
$ irule -F my_rule.r
Remote block
```
In the rodsLog:
```
Feb  8 21:53:56 pid:168 remote addresses: 127.0.0.1, 172.17.0.3 ERROR: getaddrinfo_with_retry address resolution timeout [nonexistent-host] [ai_flags: [2] ai_family: [0] ai_socktype: [0] ai_protocol: [0]]
Feb  8 21:53:56 pid:168 remote addresses: 127.0.0.1, 172.17.0.3 ERROR: [-]      /irods/server/core/src/rodsConnect.cpp:138:rodsServerHost_t *mkServerHost(char *, char *) :  status [SYS_GET_HOSTNAME_ERR]  errno [] -- message [queueAddr failed]

Feb  8 21:53:56 pid:168 remote addresses: 127.0.0.1, 172.17.0.3 ERROR: resolveHost: mkServerHost error
```

alanking:
Changing expected behavior to:
```bash
$ irule -F my_rule.r 
remote addresses: 172.17.0.3 ERROR: rcExecMyRule error.  status = -17000 SYS_GET_HOSTNAME_ERR
```

trel:
only thing better might be to print out the bad hostname that was attempted?     not critical.

alanking:
As it turns out, the error code returned nowadays is actually `SYS_INVALID_SERVER_HOST`, so that seems appropriate as well.

We can add the hostname to the rError stack. Here's what I've got with my current test...
```
Assert Command: irule -r irods_rule_engine_plugin-irods_rule_language-instance -F /tmp/irods-testing-ov7y9g7f/test_remote_returns_appropriate_error_on_bad_hostname__issue_4260.r
Expecting STDERR: ['SYS_INVALID_SERVER_HOST']
[...]

PR #6244, opened 2022-03-17
[#4260] Enforce valid hostname for remote rule execution (4-2-stable)

alanking:
cherry-picked from work for #6243 

Test suite running now


trel:
i like it.

alanking:
Running tests again...

I found that the error message being returned was different from the one on main and the message I was putting on the rError stack was not coming back to the client. This is because there is some error checking happening in main: 36a0c32ca102e21f8e010182ea756e448739c006 

The failure is now occurring in the right place.

alanking:
Confirmed tests are passing. Added #, awaiting green.

Issue #3602, opened 2017-07-12
Python logging show timezone in timestamp


kellerb:
- [ ] 4-2-stable
- [ ] master

---

### 1.
Instead of setting the [converter on the global `logging.Formatter`](https://github.com/irods/irods/blob/1a60b8ec20cac97c11107038c3de6da69ac1c820/scripts/irods/log.py#L58) just set it on our instance.

`logging_handler.formatter.converter = time.gmtime`

### 2.
The `asctime` field in the [Formatter we use](https://github.com/irods/irods/blob/1a60b8ec20cac97c11107038c3de6da69ac1c820/scripts/irods/log.py#L61) does not include the `T` separator or the trailing `Z` to indicate UTC as required by ISO 8601.

Instead, create the Formatter with:

``` python
logging_handler.setFormatter(logging.Formatter(
   '%(asctime)s - %(levelname)7s - %(filename)30s:%(lineno)4d - %(message)s',
   '%Y-%m-%dT%H:%M:%SZ'))                                                                                                                                                                                                                                                                                                                                                                                                         
```

trel:
Is this something we should still concern ourselves with?

korydraughn:
seems trivial. it has a milestone of 5.0 though.

do we know why?

korydraughn:
@alan please remove the checkboxes. This issue should only target 4.3.

Including this in 4.2 wouldn't help because it would make the timestamp formats differ even more.

alanking:
Done

PR #6277, opened 2022-03-27
[#3602] Timestamps in python log output are now displayed in UTC time (main)

korydraughn:
I'm not sure we actually need this, but we can discuss how important/useful this is here.

This needs to be ran through the test suite.

### DIFF
--- a/lib/core/include/irods/irods_logger.tpp
+++ b/lib/core/include/irods/irods_logger.tpp
@@ -243,7 +243,7 @@ private:
         if (const auto ec = clock_gettime(CLOCK_REALTIME, &ts); ec != 0) {
             const auto now = clock_type::to_time_t(clock_type::now());
             std::stringstream utc_ss;
-            utc_ss << std::put_time(std::gmtime(&now), "%FT%T.000");
+            utc_ss << std::put_time(std::gmtime(&now), "%FT%T.000Z");
             return utc_ss.str();
         }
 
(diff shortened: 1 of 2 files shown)


SwooshyCueb:
IMHO if we want to be Normal and Boring:tm: we should use whatever timezone the system is configured for. 
I could see an argument for explicitly using UTC if we're printing raw seconds-since-epoch timestamps or something like that, but it doesn't look like that's what we're doing.

trel:
We do want to consider that multi-timezone iRODS Zones could begin to more easily consolidate their logging via remote rsyslogd... does that suggest we should stick to all-UTC for sanity?   Looks like rsyslogd knows/understands timezones if they're part of the timestamp correctly...

korydraughn:
Moving to UTC for all log file output is likely a good idea. Things are inconsistent today (e.g. compare timestamps of irods.log and control_log.txt). This change helps to improve on that.

I just now noticed that irods.log presents milliseconds and does not include the trailing `Z` while the python log output does not include milliseconds. I'll take care of that soon.

korydraughn:
I've updated both the python logger and server logger to use the same format. The change does make it easier to see the order of things across different log files.

d-w-moore:
The  Python bit looks good . Tried them out myself, it  aligns very well and time & other output seem all correct.  The C++ change seems correct, far as I can tell.

d-w-moore (approved):

korydraughn:
Excellent!

alanking (commented):
I'm good with it if everyone else is good with it. :)

korydraughn:
#'d

Issue #6265, opened 2022-03-23
Consider moving the `irods-grid` icommand into the server


korydraughn:
`irods-grid` is a special command.

- It doesn't have the same properties as a normal icommand.
- It doesn't authenticate with the server as an iRODS user.
- It uses a different set of credentials to connect to the server.
- It connects to the server via port 1248.

Given the points above and the fact that 4.3 doesn't require icommands to be installed, `irods-grid` should be a part of the server packages.

SwooshyCueb:
Likewise. I'm on it.

SwooshyCueb:
Should we still include irods-grid in the icommands tarball package?

SwooshyCueb:
Also, should we move it to `/usr/sbin` or just keep it in `/usr/bin`? The former would probably be easier as it wouldn't require declaring funky package relationships like we had to do when we moved the auth plugins

trel:
> Should we still include irods-grid in the icommands tarball package?

I think no.  Doesn't use the iRODS protocol.   Should only come with the server (now that we've separated these).

> should we move it to `/usr/sbin`

Feels correct.   Cleaner separation, agreed.

Issue #6106, opened 2022-01-13
CMake: tweak target names, reorganize cmake configuration


SwooshyCueb:
At present, all of our CMake targets are have names like `irods_common`, `irods_plugin_dependencies`, `irods_server`, `RodsAPIs`, etc.
In order to be Normal and Boring™, we should instead follow the same pattern everyone else does.

We can look to Boost for a good example:
Boost's CMake targets follow the pattern `Boost::<component>` where `<component>` is the bit of the library's name after `libboost_`. (`libboost_system.so` is `Boost::system`, for example.)
There is also a header-only target named `Boost::headers`.
[Relevant documentation](https://cmake.org/cmake/help/v3.21/module/FindBoost.html#imported-targets)

So `irods_common` would become `iRODS::common`, `irods_plugin_dependencies` would become `iRODS::plugin_dependencies`, and so on.
As well, projects building against iRODS will call `find_package(iRODS` instead of `find_package(IRODS`.

Additionally, this will require some reorganization of our cmake configuration files. Both Boost and Qt5 set forth good examples here.

Following Boost's and Qt5's patterns will allow projects building against iRODS to select for specific components when using `find_package`.

To ease the transition, we can create a `FindIRODS` find module that will define alias targets with the old names. Calling `find_package` with the old casing will load this find module.
*I need to double-check how CMake does case sensitivity here.*

trel:
this all sounds fantastic - we do love Normal and Boring™

do we want to run away from capital letters completely?  is that easier to hold for us and for others?  aka `irods::common` and `find_package(irods)`?

SwooshyCueb:
Given that plenty of other projects use mixed case in their cmake target names, I think we can stick with `iRODS`

trel:
cool cool.   looks official and stuff.

SwooshyCueb:
Note to self: change internal target names for shared libraries to include the `lib` prefix.

PR #6281, opened 2022-03-28
[#6265] Move irods-grid from icommands

SwooshyCueb:
This PR does two things:

1. (#6106-related) Server control plane related files are moved to a new folder `server/control_plane`. This is because `irods-grid.cpp` explicitly includes `irods_server_control_plane.hpp` and the generated `server_control_plane_command.hpp` (previously part of `server/core` and `server/main_server`, respectively). (These are the only two headers from `server/` that are `#included` from `irods-grid.cpp`.)
    - ~~An object library target `irods_server_control_plane` is created for pulling in `irods_server_control_plane.cpp`~~. Instead of moving `irods_server_control_plane.cpp`, it's been renamed to `main_server_control_plane.cpp`.
    - An interface header-only library target ~~`irods_server_control_plane_headers`~~ `irods_server_control_plane` is created to allow pulling in `irods_server_control_plane.hpp` and `server_control_plane_command.hpp` without `irods_server_control_plane.cpp` or its baggage. This is needed for `irods-grid` and `irodsServer`.
    - A top-level target `irods_server_control_plane_command_hpp` is created for generating `server_control_plane_command.hpp`. This is needed for ~~`irods_server_control_plane_headers`~~ `irods_server_control_plane`.
2. (#6265) `irods-grid` is moved from the icommands repo.
    - Companion PR: irods/irods_client_icommands#288

~~I'm keeping this as a draft for now, as I'm considering keeping `irods_server_control_plane.cpp` in `server/main_server` and part of the `irodsServer` target instead of moving it to `server/control_plane`. It's not used anywhere else, (nor could it be in its current form,) and keeping it where it is will make `server/control_plane/CMakeLists.txt` much neater. If I decide to keep it where it is, I might rename it to something along the lines of `main_server_control_plane.cpp`.~~


korydraughn:
Looks good.

What's left for this?

korydraughn (approved):
Nice work!

alanking (approved):

PR #6280, opened 2022-03-28
Handling of federated groups (4-2-stable)

alanking:
cherry-picked from work for #6245

PR #6287, opened 2022-03-30
[#858] Fix up error handling in setup_service_account (main)

alanking:
Instead of raising an IrodsError with the stderr contents, we should
handle the error case as we would with something that was executed using
execute_command. This provides more context for the error that occurred
for investigation for failures.

related to #6245 and #6280


trel:
seems good by inspection.     using the general things is generally good.

korydraughn:
agreed

PR #6291, opened 2022-03-31
[#3727] Fix auth objects resolving/loading plugins (main)

alanking:
In the event that an auth object fails to resolve an auth plugin
interface, it may be the case that the plugin hasn't been loaded yet.
The failure should indicate that an attempt to load the plugin should be
made before returning an error. On success, a resolved plugin interface
will be returned and the resolution will be considered successful.

Confirmed fix by hand for Kerberos. GSI has been nearly identical for all time so the fixes were applied there as well.

### DIFF
--- a/lib/core/src/irods_gsi_object.cpp
+++ b/lib/core/src/irods_gsi_object.cpp
@@ -29,22 +29,20 @@ namespace irods {
                 _interface.c_str()));
         }
 
-        auth_ptr auth;
-        if (const auto err = auth_mgr.resolve(AUTH_GSI_SCHEME, auth); !err.ok()) {
-            return PASSMSG("Failed to resolve the GSI auth plugin.", err);
+        const auto& type = irods::AUTH_GSI_SCHEME;
+
+        auth_ptr ap;
+        if (auto err = auth_mgr.resolve(type, ap); !err.ok()) {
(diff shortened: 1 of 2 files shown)


lib/core/src/irods_gsi_object.cpp
```diff
@@ -29,22 +29,20 @@ namespace irods {
                 _interface.c_str()));
         }
 
-        auth_ptr auth;
-        if (const auto err = auth_mgr.resolve(AUTH_GSI_SCHEME, auth); !err.ok()) {
-            return PASSMSG("Failed to resolve the GSI auth plugin.", err);
+        const auto& type = irods::AUTH_GSI_SCHEME;
+
+        auth_ptr ap;
+        if (const auto resolve_err = auth_mgr.resolve(type, ap); !resolve_err.ok()) {
```
korydraughn, line 35: You can reuse the error object 8n the inner if-stmt if the const qualifier is removed.

Not required by any means. This is totally fine as it is.

korydraughn (commented):
Anything else to do for this?

lib/core/src/irods_gsi_object.cpp
alanking: Will do

korydraughn (approved):

Issue #6292, opened 2022-04-01
Wrap-around behavior in configuration integers


d-w-moore:
### iRODS Version, OS and Version
iRODS 4.2.11

### What did you try to do?
Use a integral (intXX_t) configuration value in `server_config.json` which is outside the limits of [INT_MIN, INT_MAX].

### Expected behavior
Some warning or error stating the given value is out of range.

### Observed behavior (including steps to reproduce, if applicable)

Values which are too large or small get range limited modulo-style, ie.:
   - `(INT_MIN-1)` becomes `INT_MAX`
   - `(INT_MAX+1)` becomes `INT_MIN`

This happens without a log message to indicate the value change.

PR #6293, opened 2022-04-01
[_6292] peg and warn when configuration integer values are out-of-range

d-w-moore:


korydraughn:
do we know if the nlohmann json library handles numbers that fall outside the accepted range?

your PR appears to assume it does.

d-w-moore:
> do we know if the nlohmann json library handles numbers that fall outside the accepted range?
> 
> your PR appears to assume it does.

I suspect, based on a small amount of experimentation and code browsing, that the answer is yes.  main branch has this going for it: not having to force everything that looks integer-ish into a std::any containing a int32_t .

As we now know, based on our earlier talk, it may be moot as we may just not remedy this for 4-2-stable, and it may become a documentation exercise after all : )

d-w-moore:
I've verified that in 4.3.0 / `main` branch, a top level parameter can be recovered to its full int64_t precision if specified as an integral value and access with code like this:
```
     const long& my_value = irods::get_server_property<const long>("my_value");
     rodsLog( LOG_NOTICE, "writeLine: *** DWM ***  my_value = [%ld].", my_value);

```
 But it's also true that convert_json is still being called.  Perhaps the irods core no longer uses it, but most plugins do....  Still investigating.

korydraughn:
We shouldn't need `convert_json()` anymore. Please look into removing it.

Here is even more documentation around number handling that supports your findings: https://json.nlohmann.me/features/types/number_handling/#number-serialization

trel:
this is a 4-2-stable PR... looks like the target issue is still open.

this PR should be redirected towards main if still relevant/necessary.

alanking:
#6292 has been moved into a milestone. We can get this wrapped up as part of that release.

PR #6316, opened 2022-04-09
JSON Schema name/version and VERSION.json renaming (main)

korydraughn:
Feel free to comment on this PR.

I've modified things I'm confident about and I've left comments about the things I'm unsure about.

The important thing is how we handle the transition from VERSION.json to version.json.


preinst
```diff
@@ -31,10 +31,12 @@ if [ "$PACKAGEUPGRADE" = "true" ] ; then
     touch $UPGRADE_TEMP_FILE
 fi
 
+# TODO What should happen with these?
 if [ -f "/var/lib/irods/VERSION" ] ; then
     mv /var/lib/irods/VERSION /var/lib/irods/VERSION.previous
 fi
```
korydraughn, line 37: @trel Is this a pre 4-2-stable thing?

scripts/irods/test/test_ireg.py
```diff
@@ -318,8 +318,8 @@ def test_ireg_replica_which_does_not_exist_on_connected_host__issue_4206(self):
         """Test registering a file as a replica which does not exist on this host."""
         resource = 'reghere'
 
-        old_path_to_file = os.path.join('/var', 'lib', 'irods', 'VERSION.json.dist')
-        new_path_to_file = os.path.join('/var', 'lib', 'irods', 'VERSION.json.dist.new')
+        old_path_to_file = os.path.join('/var', 'lib', 'irods', 'version.json.dist')
+        new_path_to_file = os.path.join('/var', 'lib', 'irods', 'version.json.dist.new')
```
korydraughn, line 322: Anybody familiar with this test and what it's doing?

scripts/irods/configuration.py
```diff
@@ -248,7 +248,7 @@ def validate_configuration(self):
                 'dict': self.server_config,
                 'path': paths.server_config_path()
             },
-            'VERSION': {
+            'VERSION': { # TODO Investigate
```
korydraughn, line 251: This should probably be changed because it is part of the dict used to validate the configuration files on startup.

Thoughts?

preinst
```diff
@@ -31,10 +31,12 @@ if [ "$PACKAGEUPGRADE" = "true" ] ; then
...
 
+# TODO What should happen with these?
 if [ -f "/var/lib/irods/VERSION.json" ] ; then
     mv /var/lib/irods/VERSION.json /var/lib/irods/VERSION.json.previous
 fi
```
korydraughn, line 42: Maybe all we do here is make the second argument lowercase.

That seems to be what we would want given that this is the pre-install script. Then everything would fall in line for the startup script since it would be looking for the lowercase filename.

scripts/irods/configuration.py
korydraughn: Changed it.

scripts/irods/test/test_ireg.py
korydraughn: This change is fine. The version file's use here isn't special. Based on the test, any file could have been used.

scripts/irods/upgrade_configuration.py
```diff
@@ -40,7 +40,7 @@ def requires_upgrade(irods_config):
 def upgrade(irods_config):
     l = logging.getLogger(__name__)
 
-    if not os.path.exists('.'.join([paths.version_path(), 'previous'])) and os.path.exists(os.path.join(paths.irods_directory(), 'VERSION.previous')):
+    if not os.path.exists('.'.join([paths.version_path(), 'previous'])) and os.path.exists(os.path.join(paths.irods_directory(), 'VERSION.previous')): # TODO Investigate
```
korydraughn, line 43: No change needed here because it is specific to legacy versions of iRODS.

Notice the `VERSION.previous` reference.

scripts/irods/upgrade_configuration.py
```diff
@@ -59,7 +59,7 @@ def upgrade(irods_config):
             with open(version_previous_path, 'r') as f:
                 old_version = json.load(f)
         else :
-            old_version = {'schema_name': 'VERSION',
+            old_version = {'schema_name': 'VERSION', # TODO Investigate
```
korydraughn, line 62: Specific to legacy versions of iRODS.

Ignoring.

scripts/irods/upgrade_configuration.py
```diff
@@ -349,7 +349,7 @@ def convert_legacy_configuration_to_json(irods_config):
     legacy_irods_environment = load_legacy_file(os.getenv('irodsEnvFile',
         os.path.join(paths.home_directory, '.irods', '.irodsEnv')))
     legacy_irods_host = load_legacy_file(os.path.join(paths.config_directory(), 'irodsHost'), load_as='list')
-    legacy_version = load_legacy_file(os.path.join(paths.irods_directory(), 'VERSION.previous'), token='=')
+    legacy_version = load_legacy_file(os.path.join(paths.irods_directory(), 'VERSION.previous'), token='=') # TODO Investigate
```
korydraughn, line 352: Specific to legacy versions of iRODS.

Ignoring.

scripts/irods/upgrade_configuration.py
```diff
@@ -515,7 +515,7 @@ def convert_legacy_configuration_to_json(irods_config):
 
     #new version file
     previous_version = {
-        'schema_name': 'VERSION',
+        'schema_name': 'VERSION', # TODO Investigate
```
korydraughn, line 518: Specific to legacy versions of iRODS.

Ignoring.

preinst
korydraughn: Specific to legacy versions of iRODS.

Ignoring.

korydraughn:
I've added `schema_name` as an optional property for version.json. The schema name will be defined in version.json starting with 4.3.0.

version.json is a self-referencing schema so we can't make `schema_name` required.

version.json also uses `configuration_schema_version` instead of `schema_version`. Perhaps we want to move to `schema_version` and deprecate `configuration_schema_version`.

preinst
```diff
@@ -36,5 +36,9 @@ if [ -f "/var/lib/irods/VERSION" ] ; then
...
 
 if [ -f "/var/lib/irods/VERSION.json" ] ; then
-    mv /var/lib/irods/VERSION.json /var/lib/irods/VERSION.json.previous
+    mv /var/lib/irods/VERSION.json /var/lib/irods/version.json.previous
+fi
+
+if [ -f "/var/lib/irods/version.json" ] ; then
+    mv /var/lib/irods/version.json /var/lib/irods/version.json.previous
```
trel, line 43: consider doing the lowercase one first?   in case they *both* fire for some reason?   hmmm, maybe we should rather make the capitalized version *also* `cp` to `VERSION.json.previous`... just in case?

trel (commented):
yes, this seems pretty good and thorough.

new installs and upgrades are the only two cases I can think of, and only one/two places touch this file...

trel:
> Perhaps we want to move to schema_version and deprecate configuration_schema_version.

yes, please.

preinst
korydraughn: Yes. I think copying the file is fine. I'm not sure what that buys us, but I don't see any harm in doing that.

Perhaps admins will feel better seeing the `VERSION.json.previous` alongside the lowercase version.

korydraughn:
I'm not sure we can rename/deprecate properties because of the self-referencing schema. This will take additional research.

preinst
trel: right - we *probably* do the right thing, but in case of not... the information is not stomped/lost.

korydraughn: done

korydraughn:
I've adjusted the schema definition for version.json so that it supports the old schema and the new schema.

See schema validation results here: https://www.jsonschemavalidator.net/s/huq1M59a

The schema changes also enforce the rule that `previous_version` entries of v3 schemas can only contain properties from the v3 schemas. You can test this by modifying the JSON input at the URL previously referenced.

korydraughn:
> interesting that you needed to move from int to string. might want to talk that one out.

It wasn't required, but it makes version.json conform to the same scheme as server_config.json and service_account_environment.json.

server/delay_server/src/irodsDelayServer.cpp
```diff
@@ -635,8 +635,10 @@ int main(int argc, char** argv)
         try {
             return irods::get_advanced_setting<const int>(irods::CFG_DELAY_SERVER_SLEEP_TIME_IN_SECONDS);
         }
-        catch (const irods::exception& e) {
-            logger::delay_server::error(e.what());
+        catch (...) {
+            logger::delay_server::warn("Could not retrieve sleep time from configuration. Delay server will "
```
trel, line 639: consider showing the actual entry name that was missing - so the admin can just copy/paste to fix?

server/delay_server/src/irodsDelayServer.cpp
```diff
@@ -667,8 +669,10 @@ int main(int argc, char** argv)
         try {
             return irods::get_advanced_setting<const int>(irods::CFG_NUMBER_OF_CONCURRENT_DELAY_RULE_EXECUTORS);
         }
-        catch (const irods::exception& e) {
-            logger::delay_server::error(e.what());
+        catch (...) {
+            logger::delay_server::warn("Could not retrieve number of concurrent delay rule executors from configuration. "
```
trel, line 673: consider showing the actual entry name that was missing - so the admin can just copy/paste to fix?

server/re/src/reSysDataObjOpr.cpp
```diff
@@ -1321,10 +1323,7 @@ msiSetReServerNumProc(msParam_t* xnumProc, ruleExecInfo_t* rei)
         try {
             number_of_concurrent_executors = irods::get_advanced_setting<const int>(irods::CFG_NUMBER_OF_CONCURRENT_DELAY_RULE_EXECUTORS);
         }
-        catch (const irods::exception& e) {
-            irods::log(e);
-            return e.code();
-        }
+        catch (...) {}
```
korydraughn, line 1326: Should this microservice log that the config option isn't defined?

It makes sense for the delay server, but it's not clear whether a microservice should. This microservice will silently use the default.

I can see it going either way.

server/delay_server/src/irodsDelayServer.cpp
korydraughn: done

server/delay_server/src/irodsDelayServer.cpp
korydraughn: done

server/re/src/reSysDataObjOpr.cpp
trel: let's log as WARNING - this will nudge administrators into the future and keep consistent with the best practices as the server continues to learn new tricks.

korydraughn: done

korydraughn:
Verified that these changes work for Ubuntu 18.

- Schema validation is good
- version.json is renamed on upgrade and produces VERSION.json.previous alongside version.json.previous
- irods_environment.json is updated to use "v4" as the schema version
- server_config.json is updated to use "v4" as the schema version
- version.json is updated to use "v4" as the schema version
- version.json captures previous version information correctly
- version.json includes schema_name and schema_version (configuration_schema_version is now deprecated and not used for v4 configs)

Hiding schema_version in version.json following a successful upgrade does not result in the server failing to start. It seems the upgrade logic treats schema_version special. If it is not defined, the server prints a warning, skips validation for all files, and starts the server.

schema_name does not work like this. Not defining this property in version.json results in a hard stop on server startup. Keep in mind this only applies after the server has already been upgraded.

The startup script also results in a hard stop if either of these properties aren't present in server_config.json or irods_environment.json.

trel:
I think that's okay to do a hard stop.   Does it also say what is explicitly missing/required?  So someone can fix it pretty easily?

korydraughn:
It does.

I modified the version.json file (i.e. `schema_version` -> `_schema_version`) and this is the output I got:
```bash
irods@86d5f3eef204:~$ ./irodsctl -v start
Warning encountered in validation:
  File "/var/lib/irods/scripts/irods/configuration.py", line 212, in schema_uri_prefix
    uri_version = self.version[key]
KeyError: 'schema_version'

During handling of the above exception, another exception occurred:

  File "/var/lib/irods/scripts/irods/configuration.py", line 270, in validate_configuration
    self.schema_uri_prefix,
  File "/var/lib/irods/scripts/irods/configuration.py", line 217, in schema_uri_prefix
    (paths.version_path(), key))
irods.exceptions.IrodsWarning: /var/lib/irods/version.json did not contain 'schema_version'

During handling of the above exception, another exception occurred:

  File "/var/lib/irods/scripts/irods/controller.py", line 94, in start
    self.config.validate_configuration()
  File "/var/lib/irods/scripts/irods/configuration.py", line 277, in validate_configuration
    sys.exc_info()[2])
  File "/var/lib/irods/scripts/irods/six.py", line 671, in reraise
    raise value.with_traceback(tb)
  File "/var/lib/irods/scripts/irods/configuration.py", line 270, in validate_configuration
    self.schema_uri_prefix,
  File "/var/lib/irods/scripts/irods/configuration.py", line 217, in schema_uri_prefix
    (paths.version_path(), key))
irods.exceptions.IrodsWarning: Preflight Check problem:
  JSON Configuration Validation failed.
Ensuring catalog schema is up-to-date...
Catalog schema is up-to-date.
Starting iRODS server ...
Success
```

korydraughn:
I agree the hard stops are fine. However, schema_version is special.

Do we want to change it so that version.json results in a hard stop as well?

korydraughn:
@alanking This PR does not run in the testing environment. When you get a chance, please see if you can resolve the compatibility issues.

The output references the VERSION.json.dist file a lot. I'm pretty sure the change to lowercase for VERSION.json* is the issue.

korydraughn:
> Yes, we don't want servers running that don't have things in the right place. Every thing we can catch like this reduces the email we receive.

I'll look into it.

scripts/irods/test/test_ireg.py
alanking: Right, this test just picked something that was guaranteed to be on the other computer in a test-configured environment without yet having a supported capability for making files on the other machines in a topology. I suppose you could do some `scp`/`ssh` nonsense but then we'd have to add more dependencies, potentially.

scripts/irods/upgrade_configuration.py
```diff
@@ -230,7 +230,12 @@ def upgrade_config_file(irods_config, path, new_version, schema_name=None):
     else:
         schema_version = schema_version_as_int(new_version['previous_version']['configuration_schema_version'])
 
-    target_schema_version = schema_version_as_int(new_version['configuration_schema_version'])
+    if 'schema_version' in new_version:
+        # Given that the schema version can be greater than 9, we have to extract
+        # all bytes following the "v" prefix in "schema_version".
```
alanking, line 235: Just for my sanity, is this comment just explaining why L236 doesn't read like this?
```python
target_schema_version = schema_version_as_int(new_version['schema_version'][1])
```
I didn't understand what the comment about being greater than 9 was saying, so I wanted to make sure :)

server/delay_server/src/irodsDelayServer.cpp
```diff
@@ -635,8 +635,11 @@ int main(int argc, char** argv)
         try {
             return irods::get_advanced_setting<const int>(irods::CFG_DELAY_SERVER_SLEEP_TIME_IN_SECONDS);
         }
-        catch (const irods::exception& e) {
-            logger::delay_server::error(e.what());
+        catch (...) {
+            logger::delay_server::warn("Could not retrieve [{}] from advanced settings configuration. "
+                                       "Delay server will sleep for {} seconds.",
```
alanking, line 640: The message printed in `msiSetReServerNumProc` mentions the "default value" being used. Any thoughts on mentioning that in the rest of these messages as well?

alanking (approved):
I can look into the configuration issues in the testing environment.

Changes look good, pending issue numbers for the ?'s, of course.

scripts/irods/upgrade_configuration.py
trel: b/c numbers vs strings i think

server/delay_server/src/irodsDelayServer.cpp
trel: agreed

scripts/irods/upgrade_configuration.py
korydraughn: It is because of what @trel said.

`schema_version` is a string like `"v4"`. It could one day be something like `"v12"`. Therefore, we have to capture the bytes after "v" to form the correct integer value.

server/delay_server/src/irodsDelayServer.cpp
korydraughn: Oh, you're talking about stating that the fallback value is the default value.

Sure. Will update.

scripts/irods/upgrade_configuration.py
alanking: Cool

server/delay_server/src/irodsDelayServer.cpp
korydraughn: done.

korydraughn:
Running core tests for Ubuntu 18.

korydraughn:
Passed core tests, but I forgot to make the startup script fail if schema_version isn't defined.

Will fix that and rerun the tests.

korydraughn:
I just realized we cannot use the schema validator for existence checks of `schema_version` in version.json. We also cannot use the schema validator for existence checks of `schema_validation_base_uri` in server_config.json.

These files and properties are special to the validation code. They are used to construct the base URI for validating all config files.

So, the only (simple) way to perform a hard stop here is to change the exception type thrown when either of these properties aren't defined.

With that said, this doesn't require rerunning the test suite. We only need to agree on the error message that should be printed to the terminal.

The current error message is:
- **\<file\> did not contain '\<property\>**

Suggestions:
- **Cannot validate configuration files. [\<property\>] is a required property of [\<file\>].**
- **Cannot validate configuration files. [\<property\>] must be defined in [\<file\>].**
- **Cannot validate configuration files. [\<file\>] is missing a required property: [\<property\>].**
- **Cannot validate configuration files. Cannot find [\<property\>] in [\<file\>].**

trel:
I like your third the best...

`Cannot validate configuration files. [<file>] is missing a required property: [<property>].`

kript:
Possibly an enhancement request, and unrelated to this issue (sorry if so) but could the schema validator also be able to be called on arbitrary files? We quite often have a use case where we want to change `server_config.json` so it would be really useful (as in 4.1) if we could make a copy, make the required edits to the copy, and then run the validator against the copy - if it passes we can then use it. This would allow us to catch typos and missing/too many commas, that plague any manually edited JSON!. e.g.

```
cp server_config.json server_config.json _orig
cp server_config.json server_config.json _new
#edit server_config.json _new
validator server_config.json _new
# on success
cp server_config.json _new server_config.json
```

korydraughn:
I believe that functionality is already provided (in 4.2 as well).

There is a file called `validate_json.py` in /var/lib/irods/scripts. You can run it like so:
```bash
python scripts/validate_json.py new_server_config.json 'file:///var/lib/irods/configuration_schemas/v3/server_config.json'
```
The important thing to remember for 4.2 is ... always pass the absolute path for the JSON schema argument.

I just verified this for 4.2 using a container. See output below:
```bash
irods@e084c258d59d:~$ python scripts/validate_json.py VERSION.json file:///var/lib/irods/configuration_schemas/v3/VERSION.json
Validating [VERSION.json]... Success
irods@e084c258d59d:~$ python scripts/validate_json.py /etc/irods/server_config.json file:///var/lib/irods/configuration_schemas/v3/server_config.json
Validating [/etc/irods/server_config.json]... Success
```

Keep in mind that the 4.2 schemas do not have as many requirements as 4.3 schemas. You're free to modify the schemas for your deployment. That can be in the form of requiring more properties be defined or narrowing the acceptable values for a property, etc.

Finally, the validator supports http/https. You can point all of your servers to the same URL hosting the schemas and it should work. For example, I just ran this using a 4.2.11 container:
```bash
irods@e084c258d59d:~$ python scripts/validate_json.py /etc/irods/server_config.json https://raw.githubusercontent.com/irods/irods/4.2.11/configuration_schemas/v3/server_config.json
Validating [/etc/irods/server_config.json]... Success
```

korydraughn:
All core tests passed on Ubuntu 18.

Added #'s

trel:
so good.

korydraughn:
I forgot to mention that I synchronized the validation code with the warning message in regards to the "schema_version" property.

Before, the server_config.json and version.json files were used to construct the base URI for validation. Now, only server_config.json is used. We pull the "schema_validation_base_uri" and "schema_version" from server_config.json. We use to pull only "schema_validation_base_uri" from server_config.json.

This change makes the following statement in the validation code true:
https://github.com/irods/irods/blob/66ccb095cc47deb90f62ef92ffca5a21f4aaddea/scripts/irods/configuration.py#L263-L265

trel:
ah, nice.

and so if they choose not to validate, the server will still try to stand up without any validation....?

korydraughn:
The admin has to explicitly set schema_validation_base_uri to "off" to skip validation.

Any other value will result in a validation error.

trel:
perfect

PR #6232, opened 2022-03-11
[#6231] Server correctly cleans up proc files (main)

korydraughn:

Issue #6323, opened 2022-04-13
Delay server adds completed rules back to queue, race condition, then complains loudly


alanking:
- [x] main
- [x] 4-2-stable
---
### iRODS Version, OS and Version
main only

### What did you try to do?
```bash
$ cat rule.r
my_rule {
    for (*i = 0; *i < 1000; *i = *i + 1) {
        delay("<PLUSET>5s</PLUSET>") {
            writeLine("serverLog", "hello");
        }
    }
}

INPUT null
OUTPUT ruleExecOut
$ irule -F rule.r
```

### Expected behavior
All 1000 messages should be enqueued to be printed out on the next wake-up after 5 seconds, and then printed very quickly.

### Observed behavior (including steps to reproduce, if applicable)
After a few dozen get printed I start seeing messages like this in the log:
```
iRODS Exception:
    file: /irods_source/server/delay_server/src/irodsDelayServer.cpp
    function: ruleExecSubmitInp_t (anonymous namespace)::fill_rule_exec_submit_inp(const std::string_view)
    line: 177 
    code: -154000 (SYS_INTERNAL_ERR)
    message:
        iRODS Exception:
            file: /irods_source/server/delay_server/src/irodsDelayServer.cpp
            function: ruleExecSubmitInp_t (anonymous namespace)::fill_rule_exec_submit_inp(const std::string_view)
            line: 166 
            code: -808000 (CAT_NO_ROWS_FOUND)
            message:
                Could not find row matching rule id [rule_id=10597]
<prints a stacktrace at this point>
```
For some reason, it seems that there are rule IDs in the delay queue which do not exist in the catalog. (Note: the `SYS_INTERNAL_ERR` wrapping the `CAT_NO_ROWS_FOUND` is because the `irods::exception` thrown with the `CAT_NO_ROWS_FOUND` error is caught in a `std::exception` catch clause in the function(s) noted above)

The delay server then very slowly goes through the remaining rules with many similar messages and actually successfully executing some of the remaining delay rules (but not all).

This may be reproducible with a smaller number of delayed rules, but it doesn't seem to occur when there are fewer than a couple dozen rules to chew through.

This was discovered while running the following test from the storage tiering plugin: `test_plugin_unified_storage_tiering.TestStorageTieringContinueInxMigration.test_put_gt_max_sql_rows`

This bug does not exist in 4-2-stable.

alanking:
[...]

PR #6325, opened 2022-04-13
[#6232] Dequeue delay rules not found in catalog (main)

alanking:
When a rule in the queue is not found in the catalog for whatever
reason, it should be removed from the delay queue without doing any
additional work. This can happen due to a timing issue wherein the delay
server fetches work from the catalog that is currently being done and
then place it back on the queue after it has been removed. This is
normal and a side effect of the asynchronous producer-consumers model
employed by the delay server.

---

Tests cannot run yet due to changes in the setup script and upcoming changes to configuration files. I will confirm once these issues are resolved in the testing environment. Until then, this will be considered WIP.

### DIFF
--- a/server/delay_server/src/irodsDelayServer.cpp
+++ b/server/delay_server/src/irodsDelayServer.cpp
@@ -142,39 +142,34 @@ namespace
     {
         std::vector<std::string> exec_info;
 
-        try {
-            // The database plugin implementation is not capable of fetching the r_rule_exec.exe_context
-            // column when it contains a large string. To get around this limitation, the delay server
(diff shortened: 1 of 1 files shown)


alanking:
This does not resolve all of the issues of concern, but it does alleviate some of the symptoms. Running tests now.

alanking:
All tests passed

trel:
ship it.

korydraughn (approved):

alanking:
Should have linked this to #6323. Sorry.

trel:
branch name was right :)

Issue #6137, opened 2022-01-27
Organize and establish conventions for irods_configuration_keywords.hpp


korydraughn:
## Enhancement
Possible Conventions:
- Group related keywords together
- Group keywords together by location in sorted config files
- Keywords used in multiple sections of a config should be grouped together
- Use a prefix instead of suffix for all keywords (e.g. `CFG_`, `CFG_KW_`, etc.)
- Change data type to `const char* const` or `std::string_view` to avoid memory allocation

Plugins that will need to be updated:
- audit REP
- indexing capability
- logical quotas REP
- Python REP
- s3 resource plugin
- storage tiering capability

PR #6332, opened 2022-04-18
[#6137] Renamed variables in irods_configuration_keywords.hpp (main)

korydraughn:
- Should `KW_` be used instead of `KW_CFG_`?
- Do all variables defined in irods_configuration_keywords.hpp reference a config option?

Notice that the data type for these variables has changed from `std::string` to `const char* const`. `std::string_view` would work, but it required adjusting internal APIs and function call invocations. Either way, `const char* const` provides a lot of flexibility and doesn't break existing APIs.

### DIFF
--- a/lib/core/include/irods/authentication_plugin_framework.hpp
+++ b/lib/core/include/irods/authentication_plugin_framework.hpp
@@ -102,7 +102,7 @@ namespace irods::experimental
         auto err = irods::load_plugin<plugin_type>(
                         plugin,
                         name,
-                        irods::PLUGIN_TYPE_AUTHENTICATION,
+                        irods::KW_CFG_PLUGIN_TYPE_AUTHENTICATION,
                         name,
                         "empty_context");
         if(!err.ok()) {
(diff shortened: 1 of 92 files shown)


alanking:
- I would be fine with either. I think keeping the `CFG_` prefix in some instances would be good, but that would require more careful consideration as is being suggested in question 2.
- No. For instance, not all of the plugin types are used as configuration keywords, as near as I can tell.

Please confirm that our plugins and shipped clients still work with these changes. Otherwise, seems fine to me. Would like an opinion from @trel and/or @SwooshyCueb as well.

trel:
No strong opinions here.

Not as familiar with most of these.    Current changes appear to just move `_KW` to the front, yes?   And `const`.

korydraughn:
Cool. I'll leave this PR as is and plan to handle the plugins.

And yes, I think the use of `KW_CFG_` for configuration variables is fine. The placement of `KW_` is the real change here.

alanking (approved):
I think this is good to go, barring any other comments

korydraughn:
Passed core tests.

korydraughn:
Will add the #'s soon

korydraughn:
Added #

Issue #6192, opened 2022-02-17
Weird packaging message for CentOS 7


korydraughn:
See https://github.com/irods/irods_development_environment/issues/34

alanking:
That *is* weird. This was ported from the `ResourceBase`, so it's not new... and the code ran fine during testing on ubuntu. Not sure what that's all about.

SwooshyCueb:
This issue, #6310, and #6294 are probably the same issue at heart.
Switching gears to focus on this.

SwooshyCueb:
Does this occur in any other repositories with the 4-2-stable branch?
If not, we can close this and https://github.com/irods/irods_development_environment/issues/34

korydraughn:
I have not checked for this in any other repository.

We can leave this open for now. Once the plugins and other repos we care about go through testing, we should know whether this can be closed.

korydraughn:
I've verified that the message captured in irods/irods_development_environment#34 does not appear in 4-3-stable or 4-2-stable.

However, I did come across this CMake warning. Any thoughts @SwooshyCueb?
```bash
[1058/1060] Building CXX object unit_tests/CMakeFiles/irods_shared_memory_object.dir/src/main.cpp.o
[1059/1060] Linking CXX executable unit_tests/irods_shared_memory_object
[1059/1060] Run CPack packaging tool...                                        
CPack: Create package using RPM                                                                
CPack: Install projects               
CPack: - Install project: irods []                                     
CPack: -   Install component: irods-database-plugin-mysql
CPack: -   Install component: irods-database-plugin-oracle                                                                                                                                    
CPack: -   Install component: irods-database-plugin-postgres                                                                                                                                  
CPack: -   Install component: irods-dev                                                                                                                                                       
CPack: -   Install component: irods-runtime                  
CPack: -   Install component: irods-server
CPack: Create package                       
CPackRPM: Will use GENERATED spec file: /irods_build/_CPack_Packages/Linux/RPM/SPECS/irods-database-plugin-mysql.spec
CPackRPM: Will use GENERATED spec file: /irods_build/_CPack_Packages/Linux/RPM/SPECS/irods-database-plugin-oracle.spec
CPackRPM: Will use GENERATED spec file: /irods_build/_CPack_Packages/Linux/RPM/SPECS/irods-database-plugin-postgres.spec
CPackRPM: Will use GENERATED spec file: /irods_build/_CPack_Packages/Linux/RPM/SPECS/irods-devel.spec                                                                                         
[...]

PR #6348, opened 2022-04-20
[#6192] Ignore python byte compilaiton failures when packaging RPMs (4-2-stable)

SwooshyCueb:
Addresses #6192 and friends for this repo.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1488,6 +1488,8 @@ set(CPACK_RPM_FILE_NAME RPM-DEFAULT)
 
 set(CPACK_ARCHIVE_COMPONENT_INSTALL ON)
 
+set(CPACK_RPM_SPEC_MORE_DEFINE "%global _python_bytecompile_errors_terminate_build 0")
+
 ### TODO better place for these
 ## database plugins
 # postgres


CMakeLists.txt
```diff
@@ -1488,6 +1488,8 @@ set(CPACK_RPM_FILE_NAME RPM-DEFAULT)
 
 set(CPACK_ARCHIVE_COMPONENT_INSTALL ON)
 
+set(CPACK_RPM_SPEC_MORE_DEFINE "%global _python_bytecompile_errors_terminate_build 0")
```
korydraughn, line 1491: Are there any downsides to this?

SwooshyCueb: Ehhh, kinda, but nothing we can really do anything about while still using CPack AFAIK.

My understanding is that stable contains a mix of python 2 and python 3 code, and some .py files that are compatible with both; and that we are still primarily using python 2 in stable.

The documentation for how rpmbuild's automatic byte compilation works is pretty sparse, but based on what I could find, and what I could suss out from looking at source code, I've made the following observations:
- Without doing the byte compilation ourselves, we can only byte-compile for one version of Python.
- Automatic btye compilation is *supposed* to only happen for .py files in standard locations. Our stuff in `/var/lib/irods` is getting picked up for byte compilation, which means one of the following:
    - The documentation is lying about this.
    - The byte compiler script has a bug.
- Python will sometimes do its own byte compilation if pre-compiled .pyc/.pyo files are not found at runtime.
    - The package manager cannot track these files, which could cause problems when upgrading or uninstalling.

Since we are still primarily using Python 2 in stable, I felt the least-bad solution was to *continue using Python 2 for byte compilation* and allow the packaging to proceed if a .py file fails to compile.
This means that if Python *3* ever runs a .py file we package, it may generate .pyc/.pyo files that the package manager does not track. *However,* I feel that this will come up infrequently enough that we can deal with it on a case-by-case basis, should problems arise.

korydraughn (approved):

SwooshyCueb:
\#'d

alanking (approved):

PR #6354, opened 2022-04-21
[#6349,#6353] Fixup python package dependency declarations

### LINKED ISSUE #6349: Add python3-distro to package dependencies for server package
We now use the `distro` python module in our installed scripts, so we must declare a dependency on the `python3-distro` package from the package containing the scripts (the server package)

### LINKED ISSUE #6353: Remove python dependency declarations from runtime package
We declare dependencies on several python packages from the runtime package, but the runtime package does not use Python.

SwooshyCueb:
Fixes #6349
Fixes #6353


korydraughn:
I see. So the python references are moved from the runtime package to the server package because the server is the only package that relies on python.

And the database plugin packages don't need to have a dependency on python because the server package is normally installed along with the database plugin package.

Is that the idea here?

korydraughn:
I just read #6353. I think that answers part of my question.

SwooshyCueb:
Yeah that's the idea

alanking (approved):

korydraughn (approved):

SwooshyCueb:
\#'d

PR #6358, opened 2022-04-26
Add legacy PAM plugin with new authentication plugin framework (main)

alanking:
I'm going to leave this in draft for now as this is still being tested.

Currently, the name of this plugin is "pam". We have discussed using "pam_password" for this as the "interactive" PAM plugin being developed in the Authentication Working Group will completely replace this later, and the existing implementation is a simple password verification.

Also up for discussion: Should this be a separate plugin which exists outside of the main server?

### DIFF
--- a/lib/core/include/irods/authentication_plugin_framework.hpp
+++ b/lib/core/include/irods/authentication_plugin_framework.hpp
@@ -1,64 +1,64 @@
 #ifndef IRODS_AUTHENTICATION_PLUGIN_FRAMEWORK_HPP
 #define IRODS_AUTHENTICATION_PLUGIN_FRAMEWORK_HPP
 
-#include "irods/authCheck.h"
-#include "irods/authPluginRequest.h"
-#include "irods/authRequest.h"
-#include "irods/authResponse.h"
 #include "irods/authenticate.h"
+#include "irods/getRodsEnv.h"
 #include "irods/irods_auth_constants.hpp"
-#include "irods/irods_auth_factory.hpp"
-#include "irods/irods_auth_manager.hpp"
-#include "irods/irods_auth_plugin.hpp"
-#include "irods/irods_kvp_string_parser.hpp"
-#include "irods/irods_native_auth_object.hpp"
-#include "irods/irods_stacktrace.hpp"
-#include "irods/miscServerFunct.hpp"
-#include "irods/msParam.h"
+#include "irods/irods_exception.hpp"
+#include "irods/irods_load_plugin.hpp"
+#include "irods/irods_plugin_base.hpp"
 #include "irods/rcConnect.h"
-#include "irods/rodsDef.h"
-
-#ifdef RODS_SERVER
-#include "irods/rsAuthCheck.hpp"
-#include "irods/rsAuthRequest.hpp"
-#endif // RODS_SERVER
+#include "irods/rodsErrorTable.h"
+#include "irods/version.hpp"
 
 #include <boost/any.hpp>
 #include <fmt/format.h>
 #include <nlohmann/json.hpp>
 
-#include <openssl/md5.h>
-
-#include <termios.h>
-#include <unistd.h>
-#include <iostream>
-#include <sstream>
+#include <functional>
 #include <string>
+#include <string_view>
+#include <vector>
+
+/// \file
 
 using json = nlohmann::json;
 
-namespace irods::experimental
+namespace irods::experimental::auth
 {
-    namespace auth
-    {
-        static const std::string flow_complete{"authentication_flow_complete"};
+    static const char* const flow_complete{"authentication_flow_complete"};
+    static const char* const next_operation{"next_operation"};
 
-        static const std::string next_operation{"next_operation"};
-    }
-
+    /// \brief Base class for authentication plugin implementations.
+    ///
+    /// \since 4.3.0
     class authentication_base : public irods::plugin_base
     {
     public:
         #define OPERATION(C, F) std::function<json(C&, const json&)>([&](C& c, const json& j) -> json {return F(c, j);})
 
+        /// \brief Constructor for the authentication plugin base class.
+        ///
+        /// \since 4.3.0
         authentication_base()
             : plugin_base{"authentication_framework_plugin", "empty_context_string"}
         {
-            add_operation(AUTH_CLIENT_START, OPERATION(rcComm_t, auth_client_start));
+            add_operation(AUTH_CLIENT_START, OPERATION(RcComm, auth_client_start));
         } // ctor
 
-        virtual json auth_client_start(rcComm_t& comm, const json& req) = 0;
+        /// Pure virtual method which all authentication plugins must implement.
+        ///
+        /// \since 4.3.0
+        virtual json auth_client_start(RcComm& _comm, const json& req) = 0;
 
+        /// \brief Add the operation captured in \p f to the plugin's operation map at key \p n.
+        ///
+        /// \param[in] n Key with which the operation will be discovered in the operation map.
+        /// \param[in] f Function object which implements the operation being added.
+        ///
+        /// \throws irods::exception If \p n is empty or already exists in the operation map.
+        ///
+        /// \since 4.3.0
         template<typename RxComm>
         void add_operation(const std::string& n, std::function<json(RxComm&, const json&)> f)
         {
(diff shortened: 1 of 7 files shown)


korydraughn:
I don't see any reason why it couldn't live in its own repo. After all, it is a plugin.

We'd need to weight the pros and cons of that.

lib/core/include/irods/authentication_plugin_framework.hpp
```diff
@@ -1,123 +1,160 @@
...
-    namespace auth
-    {
-        static const std::string flow_complete{"authentication_flow_complete"};
-
-        static const std::string next_operation{"next_operation"};
-    }
+    static const std::string flow_complete{"authentication_flow_complete"};
+    static const std::string next_operation{"next_operation"};
 
+    /// \brief Pure virtual base class for authentication plugin implementations.
```
alanking, line 30: This is not purely virtual, I guess.

lib/core/include/irods/authentication_plugin_framework.hpp
```diff
@@ -1,123 +1,160 @@
...
+namespace irods::experimental::auth
 {
-    namespace auth
-    {
-        static const std::string flow_complete{"authentication_flow_complete"};
-
-        static const std::string next_operation{"next_operation"};
-    }
+    static const std::string flow_complete{"authentication_flow_complete"};
+    static const std::string next_operation{"next_operation"};
```
alanking, line 28: Should these be `char*`s instead?

lib/core/include/irods/authentication_plugin_framework.hpp
```diff
@@ -1,123 +1,160 @@
...
+                                                   "empty_context");
         if(!err.ok()) {
             THROW(err.code(), err.result());
         }
 
         return plugin;
-    } // resolve_authentication_plugin
-
-    void authenticate_client(rcComm_t& comm, const rodsEnv& env)
+    } // resolve_plugin
```
alanking, line 130: resolve_plugin -> resolve_authentication_plugin

plugins/auth/src/native.cpp
```diff
@@ -233,7 +253,7 @@ namespace irods
 
             // need to do NoLogin because it could get into inf loop for cross zone auth
             rodsServerHost_t *rodsServerHost;
-            auto zone_name{get<std::string>("zone_name", req)};
+            auto zone_name = req.at("zone_name").get<std::string>();
```
alanking, line 256: Investigate using `get_ref` or `get_ptr`

plugins/auth/src/pam.cpp
```diff
@@ -0,0 +1,373 @@
...
+
+        bytesBuf_t* resp{};
+        auto ec = procApiRequest(&comm, 110000, static_cast<void*>(&inp), nullptr, reinterpret_cast<void**>(&resp), nullptr);
+
+        if (ec < 0) {
+            THROW(ec, "failed to perform request");
+        }
+
+        return json::parse(static_cast<char*>(resp->buf), static_cast<char*>(resp->buf) + resp->len);
+    } // request
```
alanking, line 46: This duplicates a function in the native authentication plugin. Not 100% sure how useful it is, but it feels this could be included in the framework?

plugins/auth/src/pam.cpp
```diff
@@ -0,0 +1,373 @@
...
+    auto verify(const json& p, std::vector<std::string> n)
+    {
+        for(auto&& x : n) {
+            if(!p.contains(x)) {
+                return std::make_tuple(false, x);
+            }
+        }
+
+        return std::make_tuple(true, std::string{});
+    } // verify
```
alanking, line 58: This duplicates a function in the native authentication plugin. Not 100% sure how useful it is, but it feels this could be included in the framework?

plugins/auth/src/pam.cpp
```diff
@@ -0,0 +1,373 @@
...
+            }
+
+            resp["request_result"] = password_out;
+
+            if (comm.auth_scheme) {
+                free(comm.auth_scheme);
+            }
+
+            static constexpr char* auth_scheme_pam = "pam";
+            comm.auth_scheme = strdup(auth_scheme_pam);
```
alanking, line 361: Most of this was lifted from the legacy plugin and adapted to the new framework.

plugins/auth/src/pam.cpp
```diff
@@ -0,0 +1,373 @@
...
+        char new_password[MAX_PASSWORD_LEN + 2]{};
+        strncpy(new_password, password.c_str(), MAX_PASSWORD_LEN);
+        printf("\n");
+        tty.c_lflag = oldflag;
+        if (tcsetattr(STDIN_FILENO, TCSANOW, &tty)) {
+            printf( "Error reinstating echo mode." );
+        }
+
+        return new_password;
+    } // get_password_from_client_stdin
```
alanking, line 84: Almost unmodified from original implementation

plugins/auth/src/pam.cpp
```diff
@@ -0,0 +1,373 @@
...
+                }
+
+                const auto end_ssl = irods::at_scope_exit{[host] { sslEnd(host->conn); }};
+
+                return request(*host->conn, req);
+            }
+
+            json resp{req};
+
+            log_auth::debug("getting TTL param");
```
alanking, line 319: Maybe this should be trace

plugins/auth/src/pam.cpp
```diff
@@ -0,0 +1,373 @@
...
+                                                         "zone_name",
+                                                         irods::AUTH_PASSWORD_KEY};
+            if (auto [e, n] = verify(req, required_keys); !e) {
+                THROW(SYS_INVALID_INPUT_PARAM,
+                      fmt::format("[{}:{}] missing [{}] in request", __func__, __LINE__, n));
+            }
+
+            rodsServerHost_t* host = nullptr;
+
+            log_auth::debug("connecting to catalog provider");
```
alanking, line 292: Maybe this should be trace

lib/core/src/clientLogin.cpp
```diff
@@ -290,14 +289,14 @@ int clientLogin(rcComm_t* _comm, const char* _context, const char* _scheme_overr
...
-    static constexpr auto minimum_version_for_auth_plugin_framework = irods::version{4, 3, 0};
-    const auto server_version = irods::to_version(_comm->svrVersion->relVersion);
-    if (!server_version) {
-        static constexpr auto ec = VERSION_EMPTY_IN_STRUCT_ERR;
-
-        const std::string msg = fmt::format(
-            "Failed to get version from server [{}]\n",
-            _comm->svrVersion->relVersion);
+    try {
+        if (!irods::experimental::auth::use_legacy_authentication(*_comm)) {
```
alanking, line 293: I'm not even sure whether this is necessary given the direction we are moving with respect to newer clients and older servers (that is, they will not be made compatible).

alanking (commented):
Some things I noticed and... pre-defensive comments :)

lib/core/include/irods/authentication_plugin_framework.hpp
korydraughn: It isn't if any of the virtual functions have a body.

I believe pure virtual means all member functions have no body and have a trailing `= 0;` on the function name.

lib/core/include/irods/authentication_plugin_framework.hpp
korydraughn: What is the reasoning?

Are we talking about avoiding memory allocation?

plugins/auth/src/pam.cpp
korydraughn: I don't quite understand what is being asked.

alanking: It's not really important, just duplicate code:

https://github.com/irods/irods/blob/f6fa01c972cd36a8f56c8caae5cbe82e1041b90e/plugins/auth/src/native.cpp#L19-L35

plugins/auth/src/pam.cpp
alanking: And here

https://github.com/irods/irods/blob/f6fa01c972cd36a8f56c8caae5cbe82e1041b90e/plugins/auth/src/native.cpp#L38-L47

plugins/auth/src/pam.cpp
korydraughn: Seems like the right choice to make that code available to other auth plugins.

plugins/auth/src/pam.cpp
korydraughn: If it is duplicated, we should adjust the code to avoid that where it makes sense.

So yes, I agree.

lib/core/include/irods/authentication_plugin_framework.hpp
alanking: Yep

korydraughn: `const char* const` is good. Using `const std::string_view` works as well. It's up to you.

lib/core/src/clientLogin.cpp
alanking: We should leave support for the older authentication flow until we determine it is no longer needed, at which point we should probably just stop using `clientLogin`.

alanking:
Need to run this through tests a couple of times and try the PAM plugin manually, but I think this is ready for review now.

alanking:
By the way, still okay with throwing this into its own repository as a separate plugin. We should at least have a Great Naming for the plugin because it *cannot change* once we ship it. The work done in this branch leading up to the actual PAM plugin needs to be merged as soon as it's ready. We can always remove the PAM plugin from the main repository before we ship 4.3.0.

alanking:
I've seen all of the tests pass at this point, but I'm now seeing some failures when the tests are run in a certain order, so I'm looking into this.

alanking:
The tests which were previously failing are now passing.

It seems that some iCommands may be using various methods of specifying an authentication scheme and so we need to guard against this in the client library (`clientLogin`) wrapping the call to the new authentication framework library call.

korydraughn:
Do you see potential issues around the handling of the auth scheme across various icommands

Are they blockers?

alanking:
No, the handling should now be at least as robust as things have ever been as of the latest commit.

korydraughn:
Ok. Given that all of the tests are now passing, I think we're good to merge this.

korydraughn (approved):

alanking:
Okay, confirmed that all the tests passed (in 2 hours 45 minutes!) and squashed commits are okay. Adding #'s and will merge on green unless there are objections/conflicts

PR #6361, opened 2022-04-27
3newbuilders

trel:


korydraughn (commented):
Excellent!

Consider switching almalinux to use `dnf` instead of `yum`.

Do we want to go on and add Rocky Linux 8 to this? I imagine it is just a matter of adding it to the matrix.container list.

trel:
will do the dnf - thanks.

and yes, i'll play with building rocky8 too, i'll add it to the alma file for now.

alanking (commented):
Looks good to me. I think it's ready to squash once we see the greens?

trel:
yes, the last set was a bigger swipe, so i expect at least one to hiccup.   but yes, will squash on greens.

korydraughn:
The workflow name is a little confusing for alma and rocky.

We should change that. Can't be el8.

Perhaps enterprise-linux?

korydraughn:
I'm suggesting changing things so that the builder list doesn't show something like:
```
build-irods-almalinux / almalinux:8 (pull_request)
build-irods-almalinux / rockylinux:8 (pull_request) 
```

korydraughn:
I like option 1 better, but either is fine.

trel:
all green with better naming, planning to # em.

korydraughn (approved):

alanking (approved):

Issue #6372, opened 2022-05-02
Add old submodule directory to .gitignore


SwooshyCueb:
In order to make development easier when switching between branches, we should add the `irods_schema_messaging` directory to `.gitignore` in the main branch. This will prevent it from being accidentally included in the contents of a commit on the main branch.

trel:
related to https://github.com/irods/irods/issues/2768

PR #6373, opened 2022-05-02
[#6372] add irods_schema_messaging to gitignore

SwooshyCueb:


.gitignore
```diff
@@ -11,3 +11,6 @@ packaging/unattended_setup.json
 .ycm_extra_conf.py
 compile_commands.json
 CMakeLists.txt.user
+
+# old submodule
+/irods_schema_messaging
```
korydraughn, line 16: Can the use of a leading forward slash be misinterpreted by `git`?

trel: seems okay?

https://stackoverflow.com/a/38433961

SwooshyCueb: The leading forward slash matches the root of the tree, which is what we want.

alanking (approved):

SwooshyCueb:
\#'d

Issue #6322, opened 2022-04-12
Centos 7 requires pyodbc externals package


SwooshyCueb:
We currently use pyodbc in one script:
https://github.com/irods/irods/blob/26ee011a37ec75dfcafc6e1ea3915b9603b14aa9/scripts/irods/database_connect.py#L17

The only available pyodbc RPM package for Centos 7 is for Python 2:
https://centos.pkgs.org/7/epel-x86_64/pyodbc-3.0.10-1.el7.x86_64.rpm.html
We cannot declare relationships from an RPM package onto a pip-installed package.

Therefore we must either package pyodbc ourselves (can't go in `/opt`!) or remove the dependency.

trel:
Tough to remove the dependency.

If we know that Python3 is installed on the host (which... I think will be required for iRODS 4.3.0... right?)...

then, can we make the control scripts smart enough to complain loudly if pyodbc isn't installed, and we assume a README-based solution is sufficient?

and then, only if not... we look into packaging pyodbc for cen7 (ew.)

korydraughn:
Before 4.3, iRODS contained a complete copy of pypyodbc. It was a single file that lived in this repo.

I'm against keeping a copy of pyodbc in the repo, but if that is easy to do and we can't resolve this through packaging, then that might be our only option.

Are we just getting lucky that pyodbc is compatible with Python 3? I believe it is, but we should confirm that.

The commit containing the switch to pyodbc (including packaging edits) is here: https://github.com/irods/irods/commit/9215fb2204df1050b2a97f602ba839ea9c36d3c5

SwooshyCueb:
pyodbc is compatible with Python 3. We currently install it with pip in the development environment.

Related: irods/irods_development_environment#45

SwooshyCueb:
Both AlmaLinux 8 and Rocky Linux 8 have `python3-pyodbc`.
I do not like the idea of having users manually install pyodbc from pip, but I like the idea of automating a pip package installation even less.
I think in the interest of getting 4.3 out the door, we can check-and-scream for now, and look into rolling our own package later.

trel:
Maybe we can include pyodbc *in our server package* for cen7 only?

SwooshyCueb:
pyodbc is a native library. I think that's probably a bad idea.

korydraughn:
After some discussion, we've decided that this will be handled via documentation for 4.3.0.

The plan is to provide a proper package for 4.3.1.

PR #6374, opened 2022-05-02
[#6322] Make missing pyodbc errors slightly more useful

SwooshyCueb:


scripts/setup_irods.py
```diff
@@ -62,6 +62,7 @@ def wrap_if_necessary():
 import irods.log
 from irods.password_obfuscation import maximum_password_length
 from irods.logging_infrastructure import setup_rsyslog_and_logrotate, rsyslog_config_path, logrotate_config_path
+from irods import database_interface
```
trel, line 65: so a setup run that is not a provider does not need the database_interface - which is why we didn't have it up top...

agreed that earlier is better - but can we preserve the 'only if needed' behavior?   not sure how, since we're pivoting on one of their answers...

scripts/irods/database_connect.py
```diff
@@ -14,7 +14,12 @@
 from . import password_obfuscation
 from .exceptions import IrodsError, IrodsWarning
 
-import pyodbc
+try:
+    import pyodbc
+except ImportError as e:
+    print('Fatal error! pyodbc module not could not be imported. Perhaps it is missing?', file=sys.stderr)
```
korydraughn, line 20: "not could not" -> "could not"

scripts/setup_irods.py
alanking: Perhaps a try-except which prints the warning here (along with a note saying that it may not be required) and then actually raises the exception at the point(s) where it is needed? The import should be attempted again if it happens further down.

I would prefer if this happened on an "only if needed" basis as well.

Is there a reason to do this outside of the current implementation being annoying?

SwooshyCueb: Not that I'm aware.
I'll wrap this in a try-except.

korydraughn (approved):

scripts/setup_irods.py
```diff
@@ -63,6 +63,12 @@ def wrap_if_necessary():
 from irods.password_obfuscation import maximum_password_length
 from irods.logging_infrastructure import setup_rsyslog_and_logrotate, rsyslog_config_path, logrotate_config_path
 
+try:
+    import pyodbc
+except:
+    print('WARNING: pyodbc module could not be imported.', file=sys.stderr)
+    print('The pyodbc module is required for setup of provider servers.', file=sys.stderr)
```
trel, line 70: "provider servers" -> "catalog service providers" ?

korydraughn: I like this wording. It's a little easier to read imo.

trel (approved):

SwooshyCueb:
\#'d

PR #6376, opened 2022-05-02
Renamed Functions and Setup Github workflow for Clang-Format

korydraughn:


.github/workflows/build-irods-clang-format-linter.yml
```diff
@@ -0,0 +1,29 @@
...
+                  echo "deb [arch=amd64] https://unstable.irods.org/apt/ $(lsb_release -sc) main" | sudo tee /etc/apt/sources.list.d/renci-irods-unstable.list
+                  sudo apt-get update -qq
+                  sudo apt-get install -qq irods-externals-clang13.0.0-0
+            - name: Run Clang-Format
+              run: |
+                  export PATH=/opt/irods-externals/clang13.0.0-0/bin:$PATH
+                  git config --global clangFormat.binary clang-format
+                  git config --global clangFormat.style file
+                  git config --global clangFormat.extensions 'h,c,hpp,cpp,tpp'
+                  output=`git clang-format HEAD^ --diff`
```
alanking, line 28: It looks like this is only going to run the check on the last commit against the previous commit, which in many cases will also be part of the branch for the PR.

To demonstrate this problem, you can add a commit with a formatting violation and then add another commit on top of that with no violations. I suspect the Github Action will give a green checkmark in spite of the formatting violation.

Is there a way to run this against the whole set of commits and diff it against the base commit? Is it worth adding this if we can't run it against the complete diff?

korydraughn: Good catch and yes, we should be able to run this against a series of commits.

We are free to derive the commits however we want. If bash does not make this trivial, then we can use python or whatever else we feel is a good tool.

If we can't handle multiple commits, then no, we cannot support this workflow.

trel: perhaps the approach here... https://github.com/KhronosGroup/SPIRV-Tools/commit/235ef34571dc734fa63a27aa672244b1548e5758 ?

korydraughn: I think I got it.

After some testing, it appears that `git clang-format` can be passed the name of the branch of the target repo and it will diff/format only those lines that were changed in the PR.

See example with multiple commits here: https://github.com/korydraughn/irods/pull/3

trel: nice.   so the output of clang format will tell you what you 'need to do' aka what you 'should have done'.

.github/workflows/build-irods-clang-format-linter.yml
```diff
@@ -0,0 +1,47 @@
...
+                  # Make clang-format available.
+                  export PATH=/opt/irods-externals/clang13.0.0-0/bin:$PATH
+
+                  # Configure Git so that "git clang-format" can be run.
+                  git config --global clangFormat.binary clang-format
+                  git config --global clangFormat.style file
+                  git config --global clangFormat.extensions 'h,c,hpp,cpp,tpp'
+
+                  # Diff the target branch with all changes made within the pull request.
+                  diff_output=`git clang-format --diff origin/$GITHUB_BASE_REF`
```
alanking, line 38: That's pretty sweet :)

Would it be better to use the `$()` syntax rather than the tickmarks? Feel free to ignore this, just have a vague memory of that being preferable/"safer".

alanking (approved):

.github/workflows/build-irods-clang-format-linter.yml
trel: ah, right - yes, please try with `$()`.

korydraughn: Will update

.github/workflows/build-irods-clang-format-linter.yml
korydraughn: That's correct.

alanking (approved):
Looks good

.github/workflows/build-irods-clang-format-linter.yml
korydraughn: done

korydraughn:
Added #'s

Issue #6352, opened 2022-04-21
chown_directories_for_postinstall.py stomps on msiExecCmd_bin ownership


SwooshyCueb:
The post-installation script for the server is careful to avoid stomping on file/dir ownership in `msiExecCmd_bin`:
https://github.com/irods/irods/blob/bb30c55ff041564daa84938ebba121a6e801d7a9/packaging/postinstall.sh#L59

But `chown_directories_for_postinstall.py` doesn't care.

trel:
yes, needs to be more... discriminating.     probably by just skipping this one subdirectory.

PR #6377, opened 2022-05-02
[#6352] Skip msiExecCmd_bin when chown-ing directories

SwooshyCueb:


scripts/chown_directories_for_postinstall.py
```diff
@@ -6,5 +6,10 @@
 if os.path.exists(paths.service_account_file_path()):
     service_account_uid = paths.irods_uid()
     service_account_gid = paths.irods_gid()
-    for dirpath, _, _ in os.walk(paths.irods_directory()):
+    is_top = True
+    for dirpath, dirnames, _ in os.walk(paths.irods_directory()):
+        if is_top:
```
alanking, line 11: Just for my own understanding, this check is here to ensure that we are only skipping the "official" msiExecCmd_bin and not one that may be deeper down and was added by the administrators for whatever reason?

korydraughn (approved):

scripts/chown_directories_for_postinstall.py
SwooshyCueb: Correct

alanking: Cool, thanks!

alanking (approved):

SwooshyCueb:
\#'d

Issue #2902, opened 2015-09-22
password visible in debug log while using PAM and SSL


trel:
- [x] main
- [x] 4-2-stable
---
found by @michael-conway 

> In rodsLog...
> 
> ```
> received header: len = 135
> <MsgHeader_PI><type>RODS_API_REQ</type><msgLen>132</msgLen><errorLen>0</errorLen><bsLen>0</bsLen><intInfo>1201</intInfo></MsgHeader_PI>
> received msg: 
> <authPlugReqInp_PI><auth_scheme_>PAM</auth_scheme_>
> <context_>a_user=mconway;a_pw=mypasswordhere;a_ttl=0</context_>
> </authPlugReqInp_PI>
> ```

trel:
This is intended behavior when debugging is turned on.

Pete-C:
May we request that this not be the case for auditing purposes. It would be unfortunate to be in the position of having to pass on logs that may include debug and hence password details within our logs :)Pete 

    On Tuesday, 16 February 2016, 18:08, Terrell Russell <notifications@github.com> wrote:
 

 This is intended behavior when debugging is turned on.—

trel:
Fair enough - reopening.

jasoncoposky:
We will skip logging XML which include `authPlugReqInp_PI` and `PAM`.  If you find any other thing that need skipped, please open another ticket.

Pete-C:
Thanks for looking at this 

    On Wednesday, 17 February 2016, 15:19, Jason M. Coposky <notifications@github.com> wrote:
 

 We will skip logging XML which include authPlugReqInp_PI and PAM. If you find any other thing that need skipped, please open another ticket.—

trel:
Need to confirm this is still happening, and if so - remove the XML logging with any `authPlugReqInp_PI` or `PAM` information.

korydraughn:
Confirmed this information is still shown on the client and server side.

Reproduced using the following steps:
1. configure SSL
2. configure PAM
3. set `irods_log_level` in irods_environment.json to 8
4. call `irodsProt=1 init` and enter password

korydraughn:
@alanking Please add the checkboxes for this issue. Thanks!

alanking:
Done. Please confirm that this is in the right milestone and close

PR #6380, opened 2022-05-09
[#2902] Passwords are no longer leaked while running in debug mode (main)

korydraughn:


lib/core/src/packStruct.cpp
```diff
@@ -3362,7 +3363,12 @@ int packStruct(const void *inStruct,
         /* add a NULL termination */
         *static_cast<char*>(outPtr) = '\0';
         if ( getRodsLogLevel() >= LOG_DEBUG9 ) {
-            printf( "packed XML: \n%s\n", ( char * ) packedOutput.bBuf.buf );
+            const auto& bbuf = packedOutput.bBuf;
+            const std::string_view text(static_cast<char*>(bbuf.buf), bbuf.len);
+
+            if (text.find("<authPlugReqInp_PI>") == std::string_view::npos) {
```
trel, line 3369: same as 4-2-stable, please add comment x5

trel (approved):

alanking (approved):

plugins/network/src/tcp.cpp
```diff
@@ -444,8 +446,14 @@ irods::error read_bytes_buf(
     _buffer->len = bytes_read;
 
     // log transaction if requested
-    if ( _protocol == XML_PROT ) {
-        rodsLog(LOG_DEBUG8, "received msg: \n%.*s\n", _buffer->len, ( char* )_buffer->buf );
+    if (_protocol == XML_PROT) {
+        const std::string_view text(static_cast<char*>(_buffer->buf), _buffer->len);
+
+        // Do not print authentication request input when using the XML protocol
```
alanking, line 452: Looks like there's some trailing space at the end of this line, which is probably why clang-format is upset...

alanking:
Let's # it

korydraughn:
Will do. Just verifying that the changes actually fix the problem.

Will also run through the testing environment to make sure things are still happy.

alanking (approved):
Seems fine to me. Hadn't considered the `\since` tag as it relates to 4.3.0 vs 4.2.12 (considering 4.3.0 will likely come out before 4.2.12).

korydraughn:
I feel the approach is correct, but willing to be convinced to change `\since` to something else.

alanking:
I think it's the right way to go, yeah.

korydraughn:
Core tests pass.

Added #

PR #6381, opened 2022-05-09
[#2902] Passwords are no longer leaked while running in debug mode (4-2-stable)

korydraughn:


alanking (approved):
Let's # it

lib/core/include/rcMisc.h
```diff
@@ -349,6 +349,22 @@ int get_canonical_name(const char* _hostname, char* _buf, size_t _len);
...
+///
+/// \param[in] _buffer      The buffer.
+/// \param[in] _buffer_size The size of the buffer.
+///
+/// \return A boolean value.
+/// \retval true  If the buffer might contain sensitive data.
+/// \retval false Otherwise.
+///
+/// \since 4.2.12
+bool might_contain_sensitive_data(const char* _buffer, size_t _buffer_size);
```
trel, line 366: might -> may ?

not convinced, just a suggestion...

lib/core/src/packStruct.cpp
```diff
@@ -3391,8 +3393,12 @@ int packStruct(const void *inStruct,
 
         /* add a NULL termination */
         *static_cast<char*>(outPtr) = '\0';
-        if ( getRodsLogLevel() >= LOG_DEBUG9 ) {
-            printf( "packed XML: \n%s\n", ( char * ) packedOutput.bBuf.buf );
+        if (getRodsLogLevel() >= LOG_DEBUG9) {
```
trel, line 3396: there are only two LOG_DEBUG9's in this diff - the rest are LOG_DEBUG8.

not sure... if that's important.

lib/core/include/rcMisc.h
korydraughn: updated

lib/core/src/packStruct.cpp
korydraughn: According to the docs, a log level of 8 causes the wireline XML packing instructions to be logged.

My guess is that LOG_DEBUG9 is a typo and is meant to be LOG_DEBUG8. Regardless, people outside of the core dev team have been able to implement the XML protocol without changing this.

Overall, it might not be that big of a deal to change it for 4.3.0.

trel: Seems fine - just didn't see an obvious reason for the difference.   Happy to leave it alone.

korydraughn: I'm going to leave it as is.

Issue #5219, opened 2020-10-30
iRODS does not follow XDG Base Directory Specification


SwooshyCueb:
iRODS does not follow the XDG Base Directory Specification

Biggest offender is that configuration files go in `~/.irods` by default instead of `$XDG_CONFIG_HOME/irods` (`~/.config/irods`).

trel:
This the one?    https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

SwooshyCueb:
That's the one

PR #6384, opened 2022-05-10
[#6383,#6386,#6392] Correct some bad path assumptions

### LINKED ISSUE #6383: paths.py does not use install directories set in CMake
We missed paths.py when addressing #5220. As a result, anything relying on this script will fail for non-package installs and for installs with non-standard directories.

For reference, irods_default_paths.cpp got similar treatment in ab9eb54b233973e7522168fc91f191301645f853

Required for #6284

### LINKED ISSUE #6386: More bad path assumptions
While working on 6383, I noticed a few other places we're making potentially bad assumptions about paths.

Hardcoded `/var/lib/irods` in `irods_logger.cpp`:
https://github.com/irods/irods/blob/0830ad132243bfc3d9c793eb8525a71fefbf9c8c/lib/core/src/irods_logger.cpp#L34

Hardcoded `var/lib/irods` in `procLog.cpp`:
https://github.com/irods/irods/blob/0830ad132243bfc3d9c793eb8525a71fefbf9c8c/server/core/src/procLog.cpp#L29

Hardcoded `/var/lib/irods` in `irods_resource_manager.cpp`:
https://github.com/irods/irods/blob/0830ad132243bfc3d9c793eb8525a71fefbf9c8c/server/core/src/irods_resource_manager.cpp#L44

Hardcoded `var/lib/irods` in `osauth.h`:
https://github.com/irods/irods/blob/0830ad132243bfc3d9c793eb8525a71fefbf9c8c/lib/core/include/irods/osauth.h#L9

Hardcoded `etc` in `osauth.h`:
https://github.com/irods/irods/blob/0830ad132243bfc3d9c793eb8525a71fefbf9c8c/lib/core/include/irods/osauth.h#L11

Hardcoded `var/lib/irods` in `execCmd.h`:
https://github.com/irods/irods/blob/0830ad132243bfc3d9c793eb8525a71fefbf9c8c/lib/api/include/irods/execCmd.h#L8

(macro from `execCmd.h` is used in `rsExecCmd.cpp`):
https://github.com/irods/irods/blob/0830ad132243bfc3d9c793eb8525a71fefbf9c8c/server/api/src/rsExecCmd.cpp#L372-L374

This will cause problems with non-package installs and installations with non-standard paths.
Required for #6284

### LINKED ISSUE #6392: install(DIRECTORY) for scripts picks up __pycache__ and friends, and clobbers permissions
For installing the scripts directory, we use a call to [`install(DIRECTORY`](https://cmake.org/cmake/help/v3.12/command/install.html#installing-directories):
https://github.com/irods/irods/blob/0830ad132243bfc3d9c793eb8525a71fefbf9c8c/server/CMakeLists.txt#L201-L205
The way we're doing this here picks up *everything* in that directory, *including any precompiled Python*. It also does not preserve the permissions of the files.

SwooshyCueb:
The goal of this PR is to give our python scripts the #5573 treatment and correct a few other bad path assumptions.

I'm submitting it this early so I can put a tickbox list here to keep myself organized.

- [x] Use `__file__` rather than `inspect` to derive scripts path
- [x] Make `paths.root_directory` less public and eliminate usages outside of `paths.py`
- [x] Create CMake-generated python file containing paths such as `CMAKE_INSTALL_LIBDIR` and `IRODS_HOME_DIRECTORY`
- [x] Use CMake-generated paths file in `paths.py` to derive paths in a similar manner as we do in `irods_default_paths.cpp`
- [x] Implement handling for script execution in irods source tree (i.e. when irods is not installed).
- [x] In `controller.py`, check `CMAKE_INSTALL_RUNSTATEDIR` in addition to standard locations for shm files.
- [x] everything in 6386
- [x] Testing

Fixes #6383
Fixes #6386
Fixes #6392


SwooshyCueb:
I'm changing `path.py` to use [`pathlib`](https://docs.python.org/3.6/library/pathlib.html) for constructing paths so that we can have similar code to that in `irods_default_paths.cpp`. I've also implemented caching of return values, as deriving the same paths over and over again can get expensive.
The "public" methods still return strings, but honestly the whole situation in there is a mess and we should consider a small refactor once we've got 4.3 out the door.
We should also consider looking at the paths we use in our tests, as when grepping for stuff like `var/`, they showed up quite a lot. I would imagine we would want to be able to run tests on a non-package install.

It's late, so I haven't done a whole lot of testing just yet, but if I haven't broken anything, I think the python changes are mostly finished.

korydraughn (commented):
Looking real good.

scripts/irods/paths.py
```diff
@@ -1,139 +1,452 @@
...
+# TODO: is this what we really want here?
+# Not cached, as irods_user_and_group_entries return value could change
+def _home_directory():
+    return pathlib.Path(irods_user_and_group_entries()[0].pw_dir)
+
+# TODO: is this what we really want here?
+# Not cached, as irods_user_and_group_entries return value could change
 def home_directory():
-    return irods_user_and_group_entries()[0][5]
+    return irods_user_and_group_entries()[0].pw_dir
```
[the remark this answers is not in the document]
SwooshyCueb: `home_directory()` always returns the home directory of the irods service account, rather than the home directory of the user executing the script. Therefore, `default_client_environment_path()` and friends always return the path to the irods service account's files, which may not always be desirable.
I think we may want to consider renaming these methods to convey that the paths returned are specifically for the irods service account.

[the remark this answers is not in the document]
SwooshyCueb: I've changed `# TODO: is this what we really want here?` to be more specific about this

SwooshyCueb:
@korydraughn you may want to take another look at this now that I've pushed the commits for 6386

server/core/include/irods/irods_signal.hpp
```diff
@@ -1,6 +1,8 @@
 #ifndef IRODS_SIGNAL_HPP
 #define IRODS_SIGNAL_HPP
 
+#include <string>
+
```
SwooshyCueb, line 5: Remove this change

server/core/src/irods_signal.cpp
```diff
@@ -26,14 +29,12 @@ void stacktrace_signal_handler(int _signal)
 
     // "_POSIX_PATH_MAX" is guaranteed to be defined as 256.
     // This value represents the minimum path length supported by POSIX operating systems.
-    // Given that we're dealing with a relative path, a buffer of this size is acceptable here.
-    char file_path_unprocessed[_POSIX_PATH_MAX] = "../../";
-    std::strcat(file_path_unprocessed, IRODS_DEFAULT_PATH_HOMEDIR);
-    std::strcat(file_path_unprocessed, "/stacktraces/");
-
+    char file_path_unprocessed[_POSIX_PATH_MAX];
```
SwooshyCueb, line 32: Since we're not making assumptions about the working directory (and therefore not using relative paths), I don't know if `_POSIX_PATH_MAX` makes sense to use any more.

lib/core/include/irods/osauth.h
```diff
@@ -6,9 +6,7 @@
 #define OS_AUTH_H__
 
 
-#define OS_AUTH_CMD      "var/lib/irods/clients/bin/genOSAuth"
 #define OS_AUTH_ENV_USER "OSAUTH_USERNAME"
-#define OS_AUTH_KEYFILE "etc/irods/irods.key"
```
korydraughn, line 11: Hmm, these appear to be part of the public API. We need to decide whether these macros can be moved. Based on the values, they look like they shouldn't be used on the client-side.

@trel @alanking
- What kind of support will the OS auth plugin receive in 4.3.0?
- Is it just not being ported to the new auth framework?
- Are we still supporting the legacy auth plugin?

scripts/irods/paths.py
```diff
@@ -1,139 +1,458 @@
...
+def _irods_install_directory():
+    global _irods_install_directory_cache
+    if _irods_install_directory_cache is None:
+        if no_paths_cmake:
+            raise IrodsError('Running from source tree. No install directories.')
+        _irods_install_directory_cache = pathlib.PurePath(os.path.normpath(paths_cmake.cmake_dir_irodshome))
+    return _irods_install_directory_cache
+
+_sysconf_install_directory_cache = None
+def _sysconf_install_directory():
```
korydraughn, line 35: We should document these for others less familiar with them (i.e. me lol).

Same for `_localstate_install_directory()`. Or we reference other material that provides good explanations of these directory names.

korydraughn, line 35: We should document these for others less familiar with them (i.e. me lol).

Same for `_localstate_install_directory()`. Or we reference other material that provides good explanations of these directory names.

scripts/irods/paths.py
```diff
@@ -1,139 +1,458 @@
...
 
 def host_access_control_config_path():
-    return os.path.join(
-        config_directory(),
-        'host_access_control_config.json')
+    return str(_host_access_control_config_path())
 
+# TODO: consider renaming to convey that this is specifically the service account's userconf directory
+# TODO: cache this if we start caching _home_directory()
+def _userconf_directory():
```
korydraughn, line 260: Same as `_localstate_*`

korydraughn, line 260: Same as `_localstate_*`

server/core/src/irods_signal.cpp
```diff
@@ -12,9 +12,12 @@
 #include <ctime>
 #include <system_error>
 #include <charconv>
+#include <string>
 
 #include <unistd.h>
 
+static const std::string STACKTRACE_DIR(irods::get_irods_stacktrace_directory().string());
```
korydraughn, line 19: We need to think about when this line is initialized. It must be initialized on program startup to avoid violating async-signal safety.

Can the value returned by `irods::get_irods_stacktrace_directory()` be available at compile-time?

server/core/src/irods_signal.cpp
korydraughn: We have to be careful in this function since it must be async-signal-safe.

`_POSIX_PATH_MAX` was chosen because it is consistent across POSIX implementations. If we don't use that, what would you recommend the size of the buffer be?

korydraughn: We have to be careful in this function since it must be async-signal-safe.

`_POSIX_PATH_MAX` was chosen because it is consistent across POSIX implementations. If we don't use that, what would you recommend the size of the buffer be?

server/api/src/rsExecCmd.cpp
```diff
@@ -368,8 +371,8 @@ execCmd( execCmd_t *execCmdInp, int stdOutFd, int stdErrFd ) {
     char *av[LONG_NAME_LEN];
     int status;
 
-
-    snprintf( cmdPath, LONG_NAME_LEN, "%s/%s", CMD_DIR, execCmdInp->cmd );
+    auto cmd_path = irods::get_irods_home_directory() / CMD_DIR / execCmdInp->cmd;
```
korydraughn, line 374: Is anything else using `CMD_DIR`?

If not, maybe we just delete the macro and use its value directly.

Also, `cmd_path` can be marked as const.

server/re/src/reIn2p3SysRule.cpp
```diff
@@ -119,7 +123,8 @@ int rodsMonPerfLog( char *serverName, char *resc, char *output, ruleExecInfo_t *
         pthread_mutex_lock( &my_mutex );
 #endif
         /* append to the output log file */
-        FILE *foutput = fopen( fname, "a" );
+        auto log_fname = irods::get_irods_home_directory() / "log" / fname;
```
korydraughn, line 126: Can be marked as const.

server/re/src/reIn2p3SysRule.cpp
```diff
@@ -571,8 +576,9 @@ int msiServerMonPerf( msParam_t *verb, msParam_t *ptime, ruleExecInfo_t *rei ) {
     /* read the config file or the iCAT to know the servers list to monitor */
     nresc = 0;
 
+    auto mon_cfg_file = irods::get_irods_home_directory() / MON_CFG_FILE;
```
korydraughn, line 579: Can be marked as const.

server/core/src/procLog.cpp
```diff
@@ -26,7 +27,8 @@ initAndClearProcLog() {
 
 int
 initProcLog() {
-    std::snprintf(ProcLogDir, MAX_NAME_LEN, "../../var/lib/irods/log/proc");
+    auto proc_log_dir = irods::get_irods_home_directory() / "log" / "proc";
```
korydraughn, line 30: Can be marked as const.

lib/core/include/irods/osauth.h
trel: I think we are... not shipping osauth with 4.3.0.   If anyone wants it, we can discuss a plugin later - or just configure PAM to provide the same effect (require a local authenticated user... like our PAM training).   @alanking is that right?

server/core/src/irods_signal.cpp
SwooshyCueb: No, for a couple of different reasons.
1. `irods::get_irods_stacktrace_directory()` returns a `boost::filesystem::path`. Since these objects are mutable, a new one must be constructed each time.
2. The values returned by `irods::get_irods_stacktrace_directory()` and friends are dependent on where `libirods_common.so.4.3.0` is located at runtime. This is required for relocatable installations.

scripts/irods/paths.py
SwooshyCueb: All of the `*_install_directory()` methods are wrappers around CMake-provided values, and outside of `path.py` their usage should generally be avoided, as they are not guaranteed to provide usable paths on their own.

scripts/irods/paths.py
SwooshyCueb: This one is going to change once we start properly following the XDG specifications.
Right now it's just `~/.irods`, but application-specific dotfolders violate the spec, so this will eventually *evaluate to* `~/.config/irods` under normal circumstances, though setting a few environment variables can change the result.

server/core/src/irods_signal.cpp
SwooshyCueb: I've made `STACKTRACE_DIR` non-static and moved it into the `irods` namespace. I've also added a little logging to `irods::set_unrecoverable_signal_handlers()` that references the string so it is certain to be initialized before `stacktrace_signal_handler` is ever called.

scripts/irods/paths.py
SwooshyCueb: Relevant: #5219

lib/core/include/irods/osauth.h
korydraughn: Ok. Since we aren't planning to carry OS auth forward into 4.3, these changes are fine.

korydraughn (approved):
Nice work.

We're going to need to review our signal handling logic. It looks like only the agent factory and agents can produce crash reports for the unrecoverable signals.

Not a blocker for 4.3.0 though.

korydraughn:
Oh, don't forget to format the code changes using `git clang-format`

scripts/irods/paths.py
```diff
@@ -1,139 +1,469 @@
...
+def _irods_install_directory():
+    global _irods_install_directory_cache
+    if _irods_install_directory_cache is None:
+        if no_paths_cmake:
+            raise IrodsError('Running from source tree. No install directories.')
+        _irods_install_directory_cache = pathlib.PurePath(os.path.normpath(paths_cmake.cmake_dir_irodshome))
+    return _irods_install_directory_cache
+
+# Returns a normalized pathlib.PurePath of CMake-provided CMAKE_INSTALL_SYSCONFDIR
+# Not guaranteed to return a usable value. Use _sysconf_directory() instead.
```
alanking, line 32: The following functions call their `*_install_*` equivalents, so it feels like there's no way around the situation in which no usable value is returned...
- `_sysconf_directory`
- `_localstate_directory`
- `_plugins_directory`
- `_server_bin_directory`
- `_runstate_directory`

`_irods_directory` does, however, do something, so that comment is still valid. Am I misunderstanding?

lib/core/include/irods/osauth.h
alanking: Yes, that's right. Should we need to ship it in the future (unlikely) we can always find a different way forward.

alanking (dismissed):

SwooshyCueb:
> We're going to need to review our signal handling logic. It looks like only the agent factory and agents can produce crash reports for the unrecoverable signals.

Yes. However, that is outside the scope of this PR.

> Oh, don't forget to format the code changes using `git clang-format`

I've already made all the changes suggested by `clang-format` that do not clash with the code style of the individual files I worked on.
If the files themselves need to be re-formatted to make `clang-format` happy, that is outside the scope of this PR.

scripts/irods/paths.py
SwooshyCueb: Those functions call their `*_install_directory()` equivalents to derive usable values from them.
`_irods_directory` does not rely on an `*_install_directory()`, which is how we are able to do this.

It's the same logic as what's in `irods_default_paths.cpp`. The values from the macros are not usable on their own.

alanking: Okay, I guess I just find the comment here confusing, then:

 > Not guaranteed to return a usable value. Use _sysconf_directory() instead.

We derive a usable value from this function but it's not guaranteed to return a usable value. In that case, we recommend calling `_sysconf_directory()` instead of `_sysconf_install_directory()`, which will just call `_sysconf_install_directory()` anyway. So why are we recommending it as an alternative? Am I overthinking this?

This is not a blocker, I just don't understand what this means (and the equivalent comments for the functions listed above).

alanking: Gotcha. Thanks!

alanking (approved):
Let's ship it!

trel:
so we're ready for clicking resolve on discussions, optional clang-formatting, optional squashing, and #'ing.

SwooshyCueb:
I still have a little testing left to do before I'm comfortable \#'ing

SwooshyCueb:
\#'d

trel:
i disagree.

Issue #6382, opened 2022-05-09
Remove support for password argument from `iinit`


alanking:
- [x] main
---
`iinit` allows the user to supply a password as an argument which will be used in lieu of the password prompt. This should be removed as it is insecure and does not add anything that cannot be accomplished through other means (i.e. scripting).

kript:
Could the docs be updated with an example/suggested way of providing the password now this is removed, please?

alanking:
The password is still taken via stdin at a prompt for the native and pam_password plugins as per usual and can be piped in like so:
```bash
$ echo $MY_IRODS_PASSWORD | iinit
```
I can add this to the help text.

PR #6391, opened 2022-05-12
Various authentication plugin fixes (main)

alanking:
Companion PR: https://github.com/irods/irods_client_icommands/pull/300

Two commits in the middle seem contradictory because they are. I wasn't sure if we wanted to support the password keyword in the pam_password plugin or not. I think this is only being used by `iinit` and that support is being removed in the referenced issue, #6382... but just in case I left support for it there.

Should we support the password key in the request message to the PAM plugin? If so, I will squash the two commits referencing #6382. If not, I will delete the second commit.

Still need to run through tests, but works when tested by hand

### DIFF
--- a/lib/core/include/irods/authentication_plugin_framework.hpp
+++ b/lib/core/include/irods/authentication_plugin_framework.hpp
@@ -163,14 +163,10 @@ namespace irods::experimental::auth
 
         const std::string* next_operation = &irods::AUTH_CLIENT_START;
 
-        json req{}, resp{};
+        json req{_ctx}, resp{};
 
         req["scheme"] = scheme;
         req[auth::next_operation] = *next_operation;
-
-        for (const auto& [k, v] : _ctx.items()) {
-            req[k] = v;
-        }
 
         while (true) {
             resp = auth->call(_comm, *next_operation, req);
--- a/lib/core/include/irods/authentication_plugin_framework.hpp
+++ b/lib/core/include/irods/authentication_plugin_framework.hpp
@@ -28,6 +28,7 @@ namespace irods::experimental::auth
 {
     static const char* const flow_complete{"authentication_flow_complete"};
     static const char* const next_operation{"next_operation"};
+    static const char* const force_password_prompt{"force_password_prompt"};
 
     /// \brief Base class for authentication plugin implementations.
     ///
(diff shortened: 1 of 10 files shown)


lib/core/include/irods/authentication_plugin_framework.hpp
```diff
@@ -28,6 +28,7 @@ namespace irods::experimental::auth
 {
     static const char* const flow_complete{"authentication_flow_complete"};
     static const char* const next_operation{"next_operation"};
+    static const char* const force_password_prompt{"force_password_prompt"};
```
alanking, line 31: This change should be retained regardless of the decision about whether to keep this commit or not.

lib/core/src/irods_server_properties.cpp
```diff
@@ -214,13 +214,17 @@ namespace irods
...
-            return;
+            THROW(ret.code(),
+                  fmt::format("[{}:{}] - Failed to find server config file [{}]", __func__, __LINE__, ret.result()));
         }
 
         std::ifstream svr{svr_fn};
-        if (svr.is_open()) {
-            config_props_ = json::parse(svr);
+
+        if (!svr || !svr.is_open()) {
```
korydraughn, line 223: `!svr.is_open()` is not needed because `!svr` covers that case.

`!svr` is basically asking if the stream is in a bad state.

korydraughn (commented):
Code looks good.

Will think about the password key commit and share thoughts later.

lib/core/src/irods_server_properties.cpp
alanking: Oh, cool. Shows how much I know about the streaming library :p

alanking: Done

alanking:
All tests passed, but I observed a weird failure in `test_resource_types.Test_Resource_Passthru.test_ssl_iput_small_and_large_files` when it tries to restart after the test is over. I ran the test again and everything worked fine, but I want to make sure this doesn't become a tripping hazard on our way to the release...

lib/core/include/irods/authentication_plugin_framework.hpp
alanking: After discussion, we have decided that the plugin will no longer support the password key. The latest commit removes this change and retains the good stuff.

alanking:
I'm just going to run this through tests one more time and barring any incidents I'll feel comfortable merging on approval.

trel (approved):

alanking:
Tests have passed again, so I guess this is good to go?

korydraughn (approved):
Ship it!

alanking:
Added #'s, merging on green.

PR #6397, opened 2022-05-15
[5277] updates for full permission model (main)

trel:
code, tests, and database tokens


trel:
@korydraughn i didn't add any new unit tests around these permissions.   but it builds!

@alanking needs a testrun alongside/atop https://github.com/irods/irods/pull/6134

korydraughn (commented):
Code changes look fine.

However, the change involving the use of underscores breaks backwards compatibility. We either need to continue using spaces or support both schemes.

korydraughn:
@trel I'm not sure if you manually formatted the code according to the output of the clang-format github action, but if so, you can simplify that step by configuring support for clang-format in git and then running `git clang-format <branch_to_rebase_on>` from the feature branch. That will format all of the code you changed. Then you can just commit the changes.

trel:
Yeah, I did it manually - was just a little copy/paste.    But your workflow sounds better - please get that documented in the developer docs if it's not already there.

trel:
Will need to discuss the spaces->underscores compatibility.  I was thinking it's only in the token entry in the database, so changing it there is the source... but could be missing a scenario/client/usecase (or five).

korydraughn (approved):
After some discussion, the changes do not appear to break backwards compatibility. We'll confirm this by running it through the test suite.

With this PR, the expected behavior in regard to backwards compatibility is ...
- Newly exposed permission levels will fail when passed to a 4.2 server
    - 4.2 servers only accept null, read, write, and own
- *modify_object* is always translated to *write* when passed through the filesystem API
- When using `ichmod`, always use `read`/`write` instead of `read_object`/`modify_object` to avoid compatibility issues with 4.2 servers

trel:
> modify_object is always translated to write when passed through the filesystem API

and

_read_object_ is always translated to _read_ when passed through the filesystem API

trel:
ready to # when blessed.

korydraughn:
Has this been run through the testing environment?

korydraughn:
Has this been run through the testing environment?

trel:
i have not.

scripts/irods/database_upgrade.py
```diff
@@ -107,6 +107,16 @@ def run_update(irods_config, cursor):
...
+        database_connect.execute_sql_statement(cursor, "update R_TOKN_MAIN set token_name = 'delete_object' where token_name = 'delete object';")
+        database_connect.execute_sql_statement(cursor, "update R_TOKN_MAIN set token_name = 'modify_object' where token_name = 'modify object';")
+        database_connect.execute_sql_statement(cursor, "update R_TOKN_MAIN set token_name = 'create_object' where token_name = 'create object';")
+        database_connect.execute_sql_statement(cursor, "update R_TOKN_MAIN set token_name = 'delete_metadata' where token_name = 'delete metadata';")
+        database_connect.execute_sql_statement(cursor, "update R_TOKN_MAIN set token_name = 'modify_metadata' where token_name = 'modify metadata';")
+        database_connect.execute_sql_statement(cursor, "update R_TOKN_MAIN set token_name = 'create_metadata' where token_name = 'create metadata';")
+        database_connect.execute_sql_statement(cursor, "update R_TOKN_MAIN set token_name = 'read_object' where token_name = 'read object';")
+        database_connect.execute_sql_statement(cursor, "update R_TOKN_MAIN set token_name = 'read_metadata' where token_name = 'read metadata';")
```
alanking, line 118: Should we note this particular upgrade change in the documentation for 4.3.0? Won't this take a very, very long time for a deployment with a lot of objects?

server/icat/include/irods/icatDefines.h
```diff
@@ -44,26 +44,26 @@
...
+#define ACCESS_DELETE_OBJECT                       "delete_object"
+#define ACCESS_CREATE_TOKEN                        "create_token"
+#define ACCESS_DELETE_TOKEN                        "delete_token"
+#define ACCESS_CURATE                              "curate"
+#define ACCESS_OWN                                 "own"
 
-#define ACCESS_INHERIT              "inherit"
-#define ACCESS_NO_INHERIT           "noinherit"
+#define ACCESS_INHERIT                             "inherit"
+#define ACCESS_NO_INHERIT                          "noinherit"
```
alanking, line 67: I dunno about all those spaces. :p Maybe add some `// clang-format off` for this section... Feel free to disregard.

alanking (commented):
Overall, looks good. I can try to run through testing shortly.

scripts/irods/database_upgrade.py
trel: no, this is a single update in the token table.   one per line.

trel: at least... that's the intent.

server/icat/include/irods/icatDefines.h
trel: let it breathe!    trust the clang-format!

scripts/irods/database_upgrade.py
korydraughn: The time needed for updating these is constant. Object permissions are linked via ids.

As for documentation, I'm not sure if this needs to be documented outside of the release notes. If we did document this, what would you imagine the docs would say?

plugins/database/src/general_query.cpp
```diff
@@ -1683,7 +1701,16 @@ genqAppendAccessCheck() {
...
             cllBindVars[cllBindVarCount++] = sessionTicket;
             cllBindVars[cllBindVarCount++] = sessionTicket;
-            if ( !rstrcat( whereSQL, "( R_DATA_MAIN.data_id in (select object_id from R_TICKET_MAIN TICK where TICK.ticket_string=?) OR R_COLL_MAIN.coll_id in (select object_id from R_TICKET_MAIN TICK where TICK.ticket_string=?) OR R_COLL_MAIN.coll_name LIKE (select (coll_name || '%') from R_COLL_MAIN where coll_id in (select object_id from R_TICKET_MAIN TICK where TICK.ticket_string=?)))", MAX_SQL_SIZE_GQ ) ) { return USER_STRLEN_TOOLONG; }
+            if (!rstrcat(whereSQL,
+                         "R_COLL_MAIN.coll_id in (select object_id from R_OBJT_ACCESS OA, R_USER_GROUP UG, R_USER_MAIN "
+                         "UM, R_TOKN_MAIN TM where UM.user_name=? and UM.zone_name=? and "
+                         "UM.user_type_name!='rodsgroup' and UM.user_id = UG.user_id and UG.group_user_id = OA.user_id "
+                         "and OA.object_id = R_COLL_MAIN.coll_id and OA.access_type_id >= TM.token_id and "
+                         "TM.token_namespace ='access_type' and TM.token_name = 'read_object')",
+                         MAX_SQL_SIZE_GQ))
```
alanking, line 1710: This breaks tickets. Please revert this hunk and I think the rest should be good to go.

trel: but it's just a space -> underscore change?    discush!

trel: oh... no!   it's not just an underscore...   have i copy/pasted clang-format incorrectly?!

trel: appears so!   hold please.

trel: okay, repushed - will await the great clang-format once more.     apologies.

alanking (approved):
Confirmed that tests are passing. Let's squash this down

PR #6405, opened 2022-05-19
Rename new pam plugin to pam_password (main)

alanking:
companion PR: https://github.com/irods/irods_client_icommands/pull/305

---

Confirmed that tests pass in testing environment.

### DIFF
--- a/plugins/auth/src/pam_password.cpp
+++ b/plugins/auth/src/pam_password.cpp
@@ -1,3 +1,4 @@
+#include "irods/plugins/auth/pam_password.hpp"
 #include "irods/authentication_plugin_framework.hpp"
 
 #define USE_SSL 1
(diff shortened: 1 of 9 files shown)


plugins/auth/src/pam_password.cpp
```diff
@@ -113,28 +116,31 @@ namespace
...
 {
-    class pam_authentication : public irods_auth::authentication_base {
+    class pam_password_authentication : public irods_auth::authentication_base
+    {
     private:
         static constexpr char* perform_native_auth = "perform_native_auth";
 
     public:
-        pam_authentication()
+        // clang-format off
```
korydraughn, line 134: I'm guessing these are for the calls to `add_operation()`. If so, let's move them directly around those calls instead of having them around the constructor.

If more code is ever added inside of the constructor, we don't want it to be ignored.

plugins/auth/src/pam_password.cpp
```diff
@@ -113,28 +116,31 @@ namespace
         //
         // The list of arguments must be terminated by a NULL pointer, and, since these are
         // variadic functions, this pointer must be cast (char *) NULL.
-        execl(PAM_AUTH_CHECK_PROG, PAM_AUTH_CHECK_PROG, _username.c_str(), (char*)NULL);
+        execl(PAM_AUTH_CHECK_PROG, PAM_AUTH_CHECK_PROG, _username.c_str(), (char*) NULL);
```
korydraughn, line 119: NULL -> nullptr

trel:
clang appears green

korydraughn (commented):
Looks good overall.

plugins/auth/src/pam_password.cpp

plugins/auth/src/pam_password.cpp
alanking: Yep, done.

korydraughn (approved):

alanking:
Awaiting greens, then squashing, #'ing, and merging.

Issue #6117, opened 2022-01-20
Killing the agent factory (`irodsServer`) results in leaked processes when restarted


alanking:
- [x] main
---
### iRODS Version, OS and Version
main

### What did you try to do?
Kill the agent factory process

### Expected behavior
The agent factory is started again and continues to run as normal (as seen in #4977)

### Observed behavior (including steps to reproduce, if applicable)
The agent factory is started again and spawns new agents to service requests, but the agent processes are never cleaned up. The agent processes are cleaned up appropriately by the original agent factory process.

This is the `ps` output after killing the agent factory process and running `ils` a few times:
```bash
$ ps -ef
UID        PID  PPID  C STIME TTY          TIME CMD
<snip>
irods    11055 11050  0 22:09 pts/0    00:00:00 ./irodsServer -uv
irods    11112 11055  0 22:10 pts/0    00:00:00 irodsReServer
irods    11135 11055  0 22:11 pts/0    00:00:00 ./irodsServer -uv
irods    11143 11135  0 22:11 pts/0    00:00:00 ./irodsServer -uv
irods    11145 11135  0 22:11 pts/0    00:00:00 ./irodsServer -uv
irods    11147 11135  0 22:11 pts/0    00:00:00 ./irodsServer -uv
irods    11149 11135  0 22:11 pts/0    00:00:00 ./irodsServer -uv
irods    11151 11135  0 22:11 pts/0    00:00:00 ./irodsServer -uv
irods    11153 11135  0 22:11 pts/0    00:00:00 ./irodsServer -uv
irods    11155 11135  0 22:11 pts/0    00:00:00 ./irodsServer -uv
irods    11157 11135  0 22:12 pts/0    00:00:00 ./irodsServer -uv
irods    11159 11135  0 22:12 pts/0    00:00:00 ./irodsServer -uv
irods    11161 11135  0 22:12 pts/0    00:00:00 ./irodsServer -uv
irods    11163 11135  0 22:12 pts/0    00:00:00 ./irodsServer -uv
irods    11165 11135  0 22:12 pts/0    00:00:00 ./irodsServer -uv
irods    11167 11135  0 22:12 pts/0    00:00:00 ./irodsServer -uv
irods    11169 11135  0 22:12 pts/0    00:00:00 ./irodsServer -uv
irods    11170 11135  0 22:12 pts/0    00:00:00 ./irodsServer -uv
<snip>
```
Stopping the server will forcefully kill all of these processes.

Investigation ongoing.

[...]

Issue #6116, opened 2022-01-20
Killing the delay server (`irodsReServer`) kills the main server


alanking:
- [x] main
---
### iRODS Version, OS and Version
main

### What did you try to do?
Kill the `irodsReServer` process

### Expected behavior
The `irodsReServer` is started again and continues to run as normal (as seen in #4977)

### Observed behavior (including steps to reproduce, if applicable)
The main server dies when it tries to start `irodsReServer` again. When the server is run in the foreground (`-u`), "Segmentation fault" appears in the output.

Here is the line which trips the segfault:
https://github.com/irods/irods/blob/6383ea596b90fa0162c0ba2ecf82aa99ffc36225/server/core/src/rodsServer.cpp#L1411

This is closing the socket opened on the server's zone port and retained through the `fork()` in order to cease receiving communications. The initial startup of the delay server does not experience any issues, but closing the socket on the second invocation of the `start_delay_server` lambda expression will result in a segmentation fault.

At this time, we don't think the call here to `close` the socket is necessary and so it should be removed.

alanking:
This may be resolved by https://github.com/irods/irods/pull/6272. Testing.

alanking:
Verified that this is no longer happening, and we even have a test for it.

PR #6406, opened 2022-05-19
[#6116][#6117] Fortify process spawner test (main)

alanking:
Confirmed that killing delay server and killing agent factory do not result in killing the main server or leaked processes, respectively. Confirmed that modified test passes in testing environment, but I'm running it through the whole gauntlet just in case it raises issues in the run up to release.

### DIFF
--- a/scripts/irods/test/test_misc.py
+++ b/scripts/irods/test/test_misc.py
@@ -198,7 +198,6 @@ class Test_Misc(session.make_sessions_mixin([('otherrods', 'rods')], []), unitte
                 self.assertFalse(os.path.exists(fn))
 
     @unittest.skipIf(test.settings.RUN_IN_TOPOLOGY, "skip for topology testing")
-    @unittest.skip('delete this line on resolution of #6116 and #6117 - kills grandpa and cannot come back')
     def test_server_respawns_processes__issue_4977(self):
         import psutil
         import time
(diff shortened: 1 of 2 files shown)


scripts/irods/test/test_misc.py
```diff
@@ -198,47 +198,85 @@ def test_server_removes_leftover_temporary_rulebase_pid_files_on_startup__issue_
...
+                    process_has_at_least_two_child_processes_with_expected_names(main_server))
+
+            # Double-check that there is exactly one delay server and exactly one agent factory
+            # instance. As the agent factory process shares a name with the main server process
+            # as well as iRODS agent processes, we check for the 'irodsServer' process name.
+            seen_process_names = defaultdict(int)
+            for i in main_server.children():
+                seen_process_names[i.name()] += 1
+            self.assertEqual(1, seen_process_names['irodsDelayServer'])
+            # TODO: Is it possible that a connection can open to spawn an agent (timing issue)?
```
trel, line 256: yes, it is - you could make the 'wake up interval' really big at the beginning of this test... reducing the window/chance?

alanking: We may need to wrap this in a `delayAssert`. If we widen the sleep window, it *could* still occur (plus the test would take way longer to complete), but if we just check again via `delayAssert`, there's an almost 0 chance that there will be extra child processes *100 times* unless something is actually wrong. Thoughts?

trel: your assessment seems correct to me.

alanking: Cool. Done

trel:
so good.

korydraughn (approved):

Issue #6404, opened 2022-05-18
iRODS server process needs to remove leftover shared memory files on startup


korydraughn:
## Refactor
Today, the startup scripts (i.e. `irodsctl`, etc.) are responsible for cleaning up leftover shared memory files. This responsibility should belong to the iRODS server process.

korydraughn:
This is handled by the iRODS 5 server.

See #7988 and #7229.

alanking:
Do we need to un-skip this test? https://github.com/irods/irods/commit/e874106bbb1b67d7b59e7cbd112d37d200c2710a

trel:
mmm, appears likely...    any others in a grep through the tests?

korydraughn:
Doesn't hurt to remove the skip and see what happens.

alanking:
Okay. Re-opening so we can remove the skip. If it proves non-trivial, we should probably reword the `msg` in the `unittest.skip` and open a new issue to address it.

SwooshyCueb:
> This is handled by the iRODS 5 server.
> 
> See [#7988](https://github.com/irods/irods/pull/7988) and [#7229](https://github.com/irods/irods/issues/7229).

Can you show me in the code where this is happening? I see where the dns/hostname caches ensure there isn't already files using the names they want, but those are pid and timestamp dependant:
https://github.com/irods/irods/blob/main/lib/core/src/dns_cache.cpp#L181-L186
https://github.com/irods/irods/blob/main/lib/core/src/hostname_cache.cpp#L94-L99

korydraughn:
The lines you've highlighted are mostly it (although the names are unlikely to ever collide with anything due to the timestamp). Most shared memory files are handled by an at_scope_exit object at the initialization call site in main.cpp and agent_main.cpp.

With that said, not all shared memory files are handled that way. We still have to handle the case of resource plugins (e.g. #8072).

And in the case of a crash, we can't assume anything about shared memory files unless we track more information or change our management scheme. There's also the situation of running multiple instances on the same computer.

PR #6407, opened 2022-05-19
[#6404] Server process removes leftover shared memory files on startup (main)

korydraughn:


server/main_server/src/rodsServer.cpp
```diff
@@ -271,6 +271,68 @@ namespace
...
+                    }
+
+                    paths_to_remove.clear();
+
+                    // Build up a list of files to remove.
+                    // This avoids potential issues with the directory iterator.
+                    for (auto&& shm_file : fs::directory_iterator{shm_dir}) {
+                        const auto filename = shm_file.path().filename();
+                        const std::string_view s = filename.c_str();
+                        const std::string_view t = "irods";
```
trel, line 306: consider "irods_" instead... seems a little safer / more precise.

server/main_server/src/rodsServer.cpp
```diff
@@ -271,6 +271,68 @@ namespace
...
+                        const auto e = std::end(s);
+
+                        if (std::search(std::begin(s), e, std::begin(t), std::end(t)) != e) {
```
trel, line 309: does this assert that it begins with "irods_"... or just that it is in there somewhere?

i think we want to assert 'begins with'.

server/main_server/src/rodsServer.cpp
```diff
@@ -945,6 +1007,11 @@ int main(int argc, char** argv)
         return 1;
     }
 
+    // This must always be called after successfully creating the PID file, but before any
+    // libraries that handle shared memory. Violating these requirements can result in incorrect
+    // behavior depending on the whether an iRODS instance is already running.
```
trel, line 1012: "on the whether" -> "on whether"

server/main_server/src/rodsServer.cpp
```diff
@@ -945,6 +1007,11 @@ int main(int argc, char** argv)
...
+    remove_leftover_shared_memory_files();
```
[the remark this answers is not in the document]
korydraughn: We have logic for cleaning up things on server stop, but they aren't guaranteed to always run.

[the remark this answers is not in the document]
trel: right - cool.

trel:
This will remove all the shm files that match - including ones that might be in use by another iRODS server running on the same machine... (non-package-install)...

So, we're going to hold on this for now to better determine why we are occasionally seeing collisions with these shm files - and solve that more directly, eventually.

korydraughn:
This PR is no longer necessary now that #7988 has been merged.

Linking to #7229 so we can find this PR later.

Cleanup of shared memory files has been absorbed into the server directly. Plugins such as the S3 resource plugin can now use the setup/teardown functions for managing shared memory.

Closing.

korydraughn:
I mispoke. The setup/teardown operations only apply to rule engine plugins. We'll need to provide similar functionality for resource plugins.

So, two issues are needed. One for the iRODS 5 server to grow support for setup/teardown for resource plugins. Another issue for the S3 resource plugin to implement those operations.

korydraughn:
Hmm, @JustinKyleJames do all s3 resources share the same shared memory instance on a single iRODS server?

korydraughn:
Thanks. I'll write up issues for this soon.

alanking:
Issues have been filed...

Audit AMQP: https://github.com/irods/irods_rule_engine_plugin_audit_amqp/issues/160
Metadata guard: https://github.com/irods/irods_rule_engine_plugin_metadata_guard/issues/89
Logical quotas: https://github.com/irods/irods_rule_engine_plugin_logical_quotas/issues/121
PREP: https://github.com/irods/irods_rule_engine_plugin_python/issues/230
Indexing: https://github.com/irods/irods_capability_indexing/issues/166
Storage Tiering: https://github.com/irods/irods_capability_storage_tiering/issues/283
Resource plugins (investigation): https://github.com/irods/irods/issues/8072

Issue #6284, opened 2022-03-29
Fix non-package installations for 4.3.1


korydraughn:
- [x] main
- [x] 4-3-stable
---
The following things need to be verified/updated for non-package installations:
- [x] `docs.irods.org` documents the correct steps for a successful installation
- [x] Schema validation works
- [x] Servers stand up and shut down without error
- [x] icommands work
- [x] The logs capture all activity

4.2.11 documentation for reference: https://docs.irods.org/4.2.11/getting_started/installation/#non-package-install-run-in-place

trel:
And consumers.

SwooshyCueb:
I think the targets list in the main CMakeLists.txt will need to be updated at the very least

korydraughn:
Can you elaborate on that a little more?

SwooshyCueb:
https://github.com/irods/irods/blob/4087cef3caf89ba94643202729298f19f8f3bd1a/CMakeLists.txt#L310-L351

This is the list of targets that must be built before the install commands are run in the `non-package-install-*` targets. It does not include every target that would be installed by the install commands, so unless someone does `make all` or whatever the ninja equivalent is before running `make non-package-install-postgres`, CMake will throw errors due to missing files.

korydraughn:
We've detected issues with shared library resolution.

Bumping to 4.3.1.

PR #6408, opened 2022-05-20
[#6284] Updated for non-package installs (main)

korydraughn:
I was able to compile the server with this PR, but there are still some issues involving the linker and `libboost_atomic.so`.

Investigating that now. Will update PR with new info as I go.


trel:
eventually, probably not in this PR - we should think about getting the non-package-install into github actions as well

korydraughn:
That sounds like a really good idea.

korydraughn:
Although this compiles, I noticed the following in the output:
```bash
[ 79%] Linking CXX shared module libirods_data_object_finalize_client.so                                                                                                                                                                                                                                                                                                                     
[ 79%] Built target irods_api_plugin-register_physical_path_client                                                                                                                                                                                                                                                                                                                           
[ 79%] Linking CXX shared module libirods_get_file_descriptor_info_client.so                                                                                                                                                                                                                                                                                                                 
[ 79%] Built target irods_api_plugin-set_delay_server_migration_info_client                                                                                                                                                                                                                                                                                                                  
[ 79%] Built target irods_api_plugin-data_object_finalize_client                                                                                                                                                                                                                                                                                                                             
[ 79%] Built target irods_api_plugin-get_file_descriptor_info_client                                                                                                                                                                                                                                                                                                                         
[ 79%] Linking CXX shared module libirods_touch_client.so                                                                                                                                                                                                                                                                                                                                    
[ 79%] Built target irods_api_plugin-touch_client                                                                                                                                                                                                                                                                                                                                            
[ 80%] Linking CXX shared module libhelloworld_client.so                                                                                                                                                                                                                                                                                                                                     
[ 80%] Built target irods_api_plugin-helloworld_client                                                                                                                                                                                                                                                                                                                                       
[ 80%] Linking CXX executable irodsTestPutGet                                                                                                                                                                                                                                                                                                                                                
/usr/bin/ld: warning: libboost_atomic.so.1.78.0, needed by /externals/boost1.78.0-0/lib/libboost_filesystem.so.1.78.0, not found (try using -rpath or -rpath-link)                                                                                                                                                                                                                           
[ 80%] Linking CXX shared module libirods_data_object_modify_info_client.so                                                                                                                                                                                                                                                                                                                  
[ 80%] Linking CXX shared module libosauth_client.so
```
Will have to investigate what caused this.

korydraughn:
After more inspection, `hostname_resolves_to_local_address` does not link against Boost.Atomic. This means we must be dealing with a transitive dependency.

Using `readelf` and a tool called `lddtree` shows the problem.
```bash
root@cecbff525a6c:/irods_npi_root/var/lib/irods# lddtree /externals/boost1.78.0-0/lib/libboost_filesystem.so
libboost_filesystem.so => /externals/boost1.78.0-0/lib/libboost_filesystem.so (interpreter => none)
    libboost_atomic.so.1.78.0 => not found
    librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1
    libc++.so.1 => not found
    libc++abi.so.1 => not found
    libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6
        ld-linux-x86-64.so.2 => /lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
    libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1
    libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0
    libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6
root@cecbff525a6c:/irods_npi_root/var/lib/irods# readelf -d /externals/boost1.78.0-0/lib/libboost_filesystem.so

Dynamic section at offset 0x1dd00 contains 33 entries:
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libboost_atomic.so.1.78.0]
 0x0000000000000001 (NEEDED)             Shared library: [librt.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libc++.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libc++abi.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libpthread.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x000000000000000e (SONAME)             Library soname: [libboost_filesystem.so.1.78.0]
 0x000000000000001d (RUNPATH)            Library runpath: [/opt/irods-externals/clang-runtime13.0.0-0/lib:/opt/irods-externals/boost1.78.0-0/lib]
```
The `RUNPATH` for externals is incorrect for non-package installs. I believe this is expected and therefore, it's just a matter of updating the `LD_LIBRARY_PATH` in the documentation.

Including the following step allows us to make progress.
```bash
$ export LD_LIBRARY_PATH=/externals/boost1.78.0-0/lib:$LD_LIBRARY_PATH
```

korydraughn:
Seems we still have some hard-coded paths to take care of.
```bash
root@cecbff525a6c:/irods_npi_root/var/lib/irods# python3  scripts/setup_irods.py < /localhost_setup_postgres.input.modified 
WARNING: pyodbc module could not be imported.
The pyodbc module is required for setup of catalog service providers.
Warning: Hostname `cecbff525a6c` should be a fully qualified domain name.
iRODS user [irods]: 
iRODS group [irods]: 
  File "scripts/setup_irods.py", line 529, in <module>
    sys.exit(main())
  File "scripts/setup_irods.py", line 515, in main
    setup_server(irods_config,
  File "scripts/setup_irods.py", line 100, in setup_server
    setup_rsyslog_and_logrotate(register_tty=False)
  File "/irods_npi_root/var/lib/irods/scripts/irods/logging_infrastructure.py", line 32, in setup_rsyslog_and_logrotate
    shutil.copyfile('/var/lib/irods/packaging/irods.rsyslog', dst)
  File "/usr/lib/python3.8/shutil.py", line 264, in copyfile
    with open(src, 'rb') as fsrc, open(dst, 'wb') as fdst:
FileNotFoundError: [Errno 2] No such file or directory: '/var/lib/irods/packaging/irods.rsyslog'
```
Perhaps the answer to this is that iRODS does not attempt to configure syslog.

Given that iRODS does not currently declare a dependency on rsyslog, it seems reasonable to not have setup do any configuration for it. iRODS messages will still appear in the default syslog file.

Now, dealing with syslog is just a matter of reading documentation.

scripts/setup_irods.py
```diff
@@ -150,13 +148,10 @@ def setup_server(irods_config, json_configuration_file=None, test_mode=False):
...
-            'The locations of the log file and configuration files are listed below.\n\n'
-            '  Log File Path               : ' + irods.paths.server_log_path() + '\n'
-            '  Rsyslog Configuration Path  : ' + rsyslog_config_path() + '\n'
-            '  Logrotate Configuration Path: ' + logrotate_config_path() + '\n\n'
-            'iRODS will never touch these configuration files again.\n'
-            'If you need to make adjustments, you must do so manually.'))
+    l.info(('iRODS uses syslog for logging. If your OS has a running syslog service, messages\n'
+            'will appear in the system log file (normally located at /var/log/syslog).\n\n'
+            'See the following for more information:\n\n'
+            f'  https://docs.irods.org/{version_json["irods_version"]}/system_overview/troubleshooting/'))
```
korydraughn, line 154: Any thoughts on the wording?

The location of the docs is temporary (because I haven't wrote the docs yet).

korydraughn:
The current issue I'm experiencing has to do with schema validation, `$ref`, and `CMAKE_INSTALL_PREFIX`.

The original code couldn't handle the inclusion of `CMAKE_INSTALL_PREFIX`. Changing the code to use `urllib.parse.urlparse` fixes that. The only issue is that the values for `$id` in the configuration schemas will look weird.

For example, supporting schema validation means doing something along the lines of this:
```diff
-"$id": "file:///@IRODS_HOME_DIRECTORY@/configuration_schemas/v4/server_config.json",
+"$id": "file:///@CMAKE_INSTALL_PREFIX@/@IRODS_HOME_DIRECTORY@/configuration_schemas/v4/server_config.json",
```
In iRODS, `CMAKE_INSTALL_PREFIX` is configured to default to `/`. This means the change above leads to schema files having `file://///<EXPANDED_IRODS_HOME_DIRECTORY>/rest/of/path` for the `$id` property.

For non-package installations, users are likely to specify a value for `CMAKE_INSTALL_PREFIX`. Say a user defines it as `/irods_npi_root`, that means the `$id` property will expand to `file:////irods_npi_root/<EXPANDED_IRODS_HOME_DIRECTORY>/rest/of/path`.

What we want is for the value of `$id` to always result in the following format:
```
file:///[non-package-install-root/]path/to/schema.json
```

I think the real solution is to introduce a new CMake variable that is defined using logic similar to the following:
```bash
# Make sure we have an absolute path!
CANONICAL_PATH = canonical_path(CMAKE_INSTALL_PREFIX)

# This is for packaged versions of iRODS.
if CANONICAL_PATH == '/' {
  IRODS_SCHEMA_ROOT = ''
}
# This is for non-package installs.
else if not empty(CANONICAL_PATH) {
  if not ends_with(CANONICAL_PATH, '/') {
    IRODS_SCHEMA_ROOT = CANONICAL_PATH + '/'
  }
}
else {
  # Fail the build!
}
```
This would happen before the call to CMake's `configure_file()`.

Then we update the input files in the following way:
```diff
-"$id": "file:///@IRODS_HOME_DIRECTORY@/configuration_schemas/v4/server_config.json",
+"$id": "file:///@IRODS_SCHEMA_ROOT@@IRODS_HOME_DIRECTORY@/configuration_schemas/v4/server_config.json",
```

The goal is to do this without the python validation code needing to adjust paths. More important than that, the schemas should reflect the actual location on disk regardless of how things were installs.

If there's a better way to do this, then let me know.

korydraughn:
The changes up to this point appear to work and produce the results we want for non-package installs and packaged installs.

CMake appears to expand/handle tildes too. Even wrapping the tilde in single quotes didn't bother CMake.

scripts/setup_irods.py
trel: Wording is ... good enough for now I think (we haven't heard anyone who is unhappy seeing it yet, or having trouble).

I still think this is probably the right approach - but we definitely want to make it as painless/obvious as possible for people to get their iRODS logs in one place, rather than the system log file.   We will certainly get email about this unless the process is smooth / obvious / good.

Perhaps something like...
"See the following for more information about configuring an iRODS-specific logfile:"

korydraughn: Agreed.

korydraughn: Done.

server/CMakeLists.txt
```diff
@@ -91,6 +91,28 @@ install(
...
 )
 
+# Initialize IRODS_CONFIGURATION_SCHEMA_PREFIX used by configuration schemas.
+# This variable helps to produce the correct URI for the "$id" JSON property.
+if (CMAKE_INSTALL_PREFIX STREQUAL "/")
+  # This branch is primarily for packaged installations, but can work for non-package
+  # installs as well.
+  set(IRODS_CONFIGURATION_SCHEMA_PREFIX "")
+elseif (NOT CMAKE_INSTALL_PREFIX STREQUAL "")
+  # This branch is for non-package installations.
```
SwooshyCueb, line 101: This is not a good way to detect non-package installs, but I think the behavior here is probably correct.

server/CMakeLists.txt
```diff
@@ -91,6 +91,28 @@ install(
   NAMELINK_COMPONENT ${IRODS_PACKAGE_COMPONENT_DEVELOPMENT_NAME}
 )
 
+# Initialize IRODS_CONFIGURATION_SCHEMA_PREFIX used by configuration schemas.
+# This variable helps to produce the correct URI for the "$id" JSON property.
+if (CMAKE_INSTALL_PREFIX STREQUAL "/")
```
SwooshyCueb, line 96: I think this case should also apply if `IRODS_HOME_DIRECTORY` is absolute.

korydraughn: You're talking about the case where someone defines a value for `IRODS_HOME_DIRECTORY` on the command line, right?

SwooshyCueb: > You're talking about the case where someone defines a value for `IRODS_HOME_DIRECTORY` on the command line, right?

Correct.

> Should the handling of `IRODS_HOME_DIRECTORY` be done here or is it unrelated to non-package installs?

I'm not sure I understand the question. If `IRODS_HOME_DIRECTORY` is absolute, we don't want to prepend `CMAKE_INSTALL_PREFIX` at all in the schemas.

korydraughn: I see.

Should anyone ever set `IRODS_HOME_DIRECTORY`?
Can you think of cases where someone may want to set that variable?

SwooshyCueb: Pretty much any case in which it is undesirable to have the home directory in `/var`

korydraughn: Done.

korydraughn:
Someone else needs to try doing a non-package install using this PR.

If you do it in a container, you'll need to make sure your container runs with an init process. This is easy with Ubuntu containers as they only require that you pass `--init` when launched.

I've only tested this on Ubuntu 20.04.

You'll need to follow the updated instructions located here:
- https://github.com/korydraughn/irods_docs/blob/non_package_install/docs/getting_started/installation.md#non-package-install-run-in-place

korydraughn:
Verified that the following changes work for CentOS 7.

However, I did run into a situation that requires a little thought.

Basically, if setup fails for any reason, the next attempt at running setup_irods.py will also fail unless the version.json file (that was copied during the first run of setup) is removed.

This is because setup defines a specific local variable only when version.json does not exist.

I think the solution is to simply read the version.json file before printing the message about syslog, etc.

korydraughn:
The change to make setup read the version.json.dist file was a success.

Adding #'s ...

korydraughn:
Added #'s

korydraughn:
There's likely more work to be done following this work.

Once merged, two things need to happen:
- Verify the test put works for non-package installs
- Figure out how to make it so that the setup script can shutdown the server

Both of these issues most likely share the same problem. And they may be specific to the fact that I tested this in a CentOS 7 docker container.

Ubuntu 20.04 didn't appear to have a shutdown problem.

trel:
`./irodsctl stop` wouldn't work?

korydraughn:
Not for centos. D-bus issues. Likely because I didn't launch the container with the proper volume mounts.

korydraughn:
I experienced a similar issue with Ubuntu, but it was solved by launching the container with `--init` which causes the container to run with an init service.

Centos is not as easy to please.

trel:
but... irods_control.py doesn't use sysctl/dbus?    hmm.

korydraughn:
Hmm. I'll try again,. If you can spin up a centos VM for me, that would be better.

trel:
done.

korydraughn:
I've verified that the issues with CentOS 7 (in docker) were due to how the container was launched. I forgot to pass the following on the call to `docker run`:
- `-v /tmp/$(mktemp -d):/run`

I've also verified that the test put, shutdown, and logging via rsyslog work as expected.

Non-package install is in good shape.

trel:
awesome - good stuff.

alanking (approved):

Issue #5715, opened 2021-06-24
Add full utf8 support to mysql


peterverraedt:
## Feature

- Add full support for utf8 encoding of mysql database, by changing the database scheme slightly.

## Current problems

- When setting the database charset to utf8mb4, irods will fail to set up properly because the odbc configuration is not set up for utf-8.
- Even if the above is fixed, database creation will fail because table column sizes or index sizes are exceeding the maximum supported values.

## Fixes to implement

We do, after a clean installation of irods, the following fixes that create full support for mysql with a utf8 charset. Some indexes are changed from layout, but that should be ok in light of the list of all indices.

```bash
# Fix mysql indices to use with utf8 encoding
SQL=/var/lib/irods/packaging/sql/icatSysTables.sql

# remove coll_name from index and don't create a unique index
sed -i '/idx_coll_main2/d' $SQL
echo "create index idx_coll_main2 on R_COLL_MAIN (parent_coll_name (767));" >> $SQL

sed -i '/idx_data_main2/d' $SQL
echo "create unique index idx_data_main2 on R_DATA_MAIN (coll_id,data_name (515),data_repl_num,data_version);" >> $SQL

sed -i '/idx_data_main4/d' $SQL
echo "create index idx_data_main4 on R_DATA_MAIN (data_name (515));" >> $SQL

sed -i '/idx_grid_configuration/d' $SQL
echo "create unique index idx_grid_configuration on R_GRID_CONFIGURATION (namespace (383), option_name (383));" >> $SQL

# Replace varchar(2700) by TEXT in R_RULE_MAIN
sed -i '8s/varchar(2700)/TEXT/g' $SQL

# Set Charset=UTF8 in /var/lib/irods/.odbc.ini
sed -i "s/'FOUND_ROWS': '1',/'FOUND_ROWS': '1', 'Charset': 'UTF8',/" /var/lib/irods/scripts/irods/database_connect.py
```

It i an easy exercise to adapt the corresponding source files.

korydraughn:
This is great! We'll need to test all the things to make sure there aren't any regressions or behavior issues.

Are there any drawbacks/concerns around these changes?

peterverraedt:
Unicity of file names (data_name) is only checked for the first 515 characters now, instead of the current 767, but that is the only real change in indexing behaviour. 

For R_COLL_MAIN, the indices
```
create unique index idx_coll_main3 on R_COLL_MAIN (coll_name);
[...]

PR #6427, opened 2022-05-31
[#5715] Enabled full support for UTF-8 in MySQL 8 (main)

korydraughn:


korydraughn:
So far, the following test suites pass:
- Ubuntu 20.04 / MySQL 8.0.29
    - unit tests
    - federation tests
- Ubuntu 18.04 / MySQL 5.7
    - unit tests

korydraughn:
The following test suites pass:
- Ubuntu 20.04 / MySQL 8.0.29
    - unit tests
    - federation tests
    - core tests (except test_misc.py)
- Ubuntu 18.04 / MySQL 5.7
    - unit tests

korydraughn:
It is. I've seen it fail in Postgres as well.

I'm now looking into why.

korydraughn:
Please eyeball the changes carefully and make sure I didn't mistype anything outlined here https://github.com/irods/irods/issues/5715#issue-929193063

alanking:
Only difference I spotted was the capitalization of the word "text" (in the linked issue it is "TEXT"). Otherwise, looks like everything is in there.

korydraughn:
Here are some questions to consider.

Q. What does this change mean for people with existing deployments?
Q. Are we responsible for providing database upgrade logic?
Q. What are the administrator's options?

korydraughn:
Here are my thoughts regarding these questions ...

> Q. What does this change mean for people with existing deployments?

Based on the following code samples, this PR doesn't appear to apply to existing deployments.
https://github.com/irods/irods/blob/0cd9df0328c5e84e5b21cb8c2bf3459eeaba84f6/scripts/irods/database_interface.py#L26
https://github.com/irods/irods/blob/0cd9df0328c5e84e5b21cb8c2bf3459eeaba84f6/scripts/irods/database_connect.py#L332-L365

> Q. Are we responsible for providing database upgrade logic?

Given my answer to the previous question, no.

> Q. What are the administrator's options?

- Admins can alter the existing database so that it matches the new database definitions.
    - Given that there are many changes, this is probably the best path forward.
    - Of course, admins should backup their existing database before doing this.
- Alternatively, admins can stand up a new MySQL database and provider. The setup script will create the needed tables. Then, they can export the data from the old database and import it into the new database.

korydraughn:
I performed the following steps to verify if these changes are safe:
1. I stood up an iRODS 4.2.11 instance backed by MySQL 5.7 (without these changes).
2. I verified that the MySQL database was configured with a latin1 character set and collation.
3. I verified that the database tables touched in this PR matched the previous definitions defined by iRODS 4.2.11.
4. I then compiled this PR and upgraded the installation to 4.3.0.
5. I ran a few commands to verify things are still functional (`ils`, `itouch`, `istream`, `irm`, `irmtrash`, and `imiscsvrinfo`).
6. I then ran the following test suites: `test_ils.py` and `test_misc.py` (both suites passed).
7. I checked the database table definitions again and verified that they had not been touched.

With that said, I am confident these changes are only applied on new installations of iRODS.

They are never executed during an upgrade.

This test was carried out using the testing environment.

korydraughn:
I'm now running the unit tests, core tests, and federation tests for Ubuntu 20 / MySQL 8 to see what happens.

korydraughn:
All unit tests and federation tests passed on the first run.

The majority of the core tests passed. The tests that failed only needed to be rerun again in order to pass.

I think this is ready.

Any concerns with adding the #?

alanking (approved):
Looks good to me. Are we waiting on +1 from @trel to include this in 4.3.0, or has that been decided?

PR #6431, opened 2022-06-02
Fix next_executor() in delay server + various test fixes/skips (main)

alanking:


scripts/irods/test/test_delay_queue.py
```diff
@@ -652,20 +653,30 @@ def test_delay_server_implicit_remote__issue_4429(self):
...
         }}
         '''.format(rep_name, object_name)
-        with lib.file_backed_up(config.server_config_path):
-            self.admin.assert_icommand(['itouch', object_name])
-            # Use the consumers as rule executors
-            config.server_config['advanced_settings']['delay_rule_executors'] = hostnames_expected
-            config.server_config['advanced_settings']['delay_server_sleep_time_in_seconds'] = 1
-            lib.update_json_file_from_dict(config.server_config_path, config.server_config)
+        try:
+            self.admin.assert_icommand(['iadmin', 'set_delay_server', lib.get_hostname()])
```
korydraughn, line 657: Why is this needed again?

Is it just trying to make sure the provider is running the delay server?

alanking: Yes, in the future, we will need to move the delay server to the local host in order for the server configurations changes to have an effect (i.e. adding the executor hosts to the list). At this time, this has no effect as we are skipping the test when executing on a host that is not the catalog service provider, but it will once we remove the dependency on an ODBC configuration in the delay server someday.

korydraughn (approved):
If the tests are passing with this change, pound it.

PR #6433, opened 2022-06-04
Removed mysql-connector-odbc dependency for almalinux-8 (main)

korydraughn:
This change allows iRODS to be installed.

Verified that all unit tests pass with this change.


korydraughn:
Q. Is relying on the mariadb package guaranteed to be compatible and safer than downloading and using a package from the MySQL website?

Q. The testing environment downloads the mysql-connector-odbc tar.gz file from the MySQL site. Should the testing environment do this at all?

Q. Should iRODS be responsible for declaring odbc dependencies other than unixODBC at all? Given that ODBC is the interface used by iRODS, should we consider letting admins pick the driver themselves? Does the mariadb driver support both MySQL 5 and 8?

korydraughn:
After some discussion, we arrived at the following:
- We'll use #6436 instead of this PR
- Given that iRODS relies on ODBC, we shouldn't need to declare package dependencies on database specific ODBC implementations such as mysql-connector-odbc
    - This means we'd document our odbc and database requirements

Issue #6448, opened 2022-06-22
Scope error for constants in `irods/apiNumber.h`


dworkin:
- [x] main
- [x] 4-3-stable
- [x] 4-2-stable
---
### iRODS Version, OS and Version

irods-dev 4.3.0, CentOS7

### What is wrong?

in `irods/apiNumber.h`, constants are defined which should be `static`. This matters when the file is included (for example, through `irods/rodsClient.h`) in multiple C source files, leading to a clash at link time.

dworkin:
Fixed in pull request #6449.

alanking:
I will take care of cherry-picking to the other branches. Thanks!

PR #6449, opened 2022-06-22
Fix scope issue in apiNumber.h.

dworkin:
Fixes issue #6448.

### DIFF
--- a/lib/api/include/irods/apiNumber.h
+++ b/lib/api/include/irods/apiNumber.h
@@ -7,7 +7,7 @@
 #ifndef API_NUMBER_H__
 #define API_NUMBER_H__
 
-#define API_NUMBER(NAME, VALUE) const int NAME = VALUE;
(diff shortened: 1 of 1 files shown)


korydraughn (approved):

korydraughn:
This change will likely result in larger binaries since it instructs the compiler to create a unique variable for each translation unit.

We need to investigate an alternative approach to handling API numbers in the future if the size of the binaries see a significant increase.

korydraughn:
Sure. We just compile everything with this change and compare the size of each binary with the released version.

It is going to depend on how many times this header is included.

dworkin:
> This change will likely result in larger binaries since it instructs the compiler to create a unique variable for each translation unit.

Well, it instructs the compiler to do that in both cases, `static` or not. It may make a difference in the compilation stage where they are optimized away, but integer constants are inlined and nothing will refer to them by the symbol.

I don't expect a change in size.

alanking (approved):

Issue #5821, opened 2021-08-31
Add facility for manually refreshing server_config in running irodsServer


alanking:
## Feature

Similar to `apachectl --graceful`, it would be nice to have a way to reload the server config in memory for a running `irodsServer` (grandpa) without restarting the process (and subsequently its children).

Most applications/services respond to `SIGHUP`. The original `irodsServer` could be targeted with this signal which is then forwarded to the agent factory (and the delay server?) so that future agents will have the updated configuration.

trel:
The server already handles `SIGHUP`, but just exits...

https://github.com/irods/irods/blob/fa596d4aba8ca0b18db8089de332e41a9946a641/server/core/src/rodsServer.cpp#L396

kript:
In 4.1.x, couldn't we just 'touch' `core. re` and it would reload? Or am I confusing this with the config re-read now done by each agent on start? I guess I am saying I expected that behaviour to persist in 4.2.x - does it not?

trel:
You are correct - and yes, it still works that way.   But only for the settings that are read by each new Agent.

There are a few settings that the AgentFactory itself (or perhaps the delay server) read earlier and it's 'too late'.

This SIGHUP response/reload would allow the main server to be refreshed without a restart (which today requires killing/bouncing the server or gracefully waiting for all open connections to close on their own).

An additional bonus of this is that containers running iRODS could use the main server as their `ENTRYPOINT`, which makes docker-compose projects much cleaner to build and maintain.

this is also related to yesterday's https://github.com/irods/irods/issues/5816

closing this issue would also fix https://github.com/irods/irods/issues/5290

kript:
> You are correct - and yes, it still works that way. But only for the settings that are read by each new Agent.
> 
> There are a few settings that the AgentFactory itself (or perhaps the delay server) read earlier and it's 'too late'.

Can you elaborate (apart from the SSL one)? I'm thinking user/group (and perhaps..databases, but surely that's a step too far!)?

> An additional bonus of this is that containers running iRODS could use the main server as their `ENTRYPOINT`, which makes docker-compose projects much cleaner to build and maintain.
> 

:heart:  As I've had to fettle with containerised iRODS recently I definitely approve of this plan :+1:

trel:
Yes, enumerating which ones 'matter' in this regard has been a long-standing... opportunity.

korydraughn:
This seems like it can be closed now that #6456 has been merged.

trel:
[...]

PR #6456, opened 2022-06-23
Configuration reload [#5821] 

epsilon-phase:
This is meant to address the need for a configuration reload facility in iRODS. 

It seems like this will lead to ongoing projects to bring its capabilities up to expected:
* handling auth changes
* making the delay server able to adjust the number of workers

Then there are fun questions here that need testing, such as changing delay server executors(the remote ones) which may or may not need additional changes to work properly.

There's probably more I'm not thinking about of course, and there is of course the positively terrifying possibility of adding an `msi_set_server_property()` microservice.

But primarily, this introduces the configuration reloading to the code and attempts to provide a facility that permits it to notify structures of settings changing in a relatively extensible manner.

### DIFF
--- a/lib/core/include/irods/irods_server_properties.hpp
+++ b/lib/core/include/irods/irods_server_properties.hpp
@@ -8,6 +8,8 @@
 #include "irods/irods_error.hpp"
 #include "irods/irods_exception.hpp"
 
+#include <mutex>
+#include <shared_mutex>
 #include <string>
 #include <string_view>
 
@@ -54,14 +56,20 @@ namespace irods
         /// @brief Read server configuration and fill server_properties::properties
         void capture();
 
+        /// \brief Read server configuration, replacing existing keys instead of deleting them.
+        /// \returns a json array containing a json patch
+        nlohmann::json reload();
+
         auto contains(const std::string_view _key) -> bool
         {
+            auto lock = acquire_read_lock();
             return config_props_.find(_key) != config_props_.end();
         }
 
         template<typename T>
         T get_property(const std::string& _key)
         {
+            auto lock = acquire_read_lock();
             auto prop = config_props_.find(_key);
             if (prop != config_props_.end()) {
                 return !prop->empty() ? prop->get<T>() : T{};
@@ -73,6 +81,7 @@ namespace irods
         template<typename T>
         T get_property(const configuration_parser::key_path_t& _keys)
         {
+            auto lock = acquire_read_lock();
             auto* tmp = &config_props_;
 
             for (auto&& k : _keys) {
@@ -89,6 +98,7 @@ namespace irods
         template<typename T>
         T set_property(const std::string& _key, const T& _val)
         {
+            auto lock = acquire_write_lock();
             nlohmann::json tmp;
             if (config_props_.contains(_key)) {
                 tmp = config_props_.at(_key);
@@ -102,6 +112,7 @@ namespace irods
         template<typename T>
         T set_property(const configuration_parser::key_path_t& _keys, const T& _val)
         {
+            auto lock = acquire_write_lock();
             auto* tmp = &config_props_;
 
             for (auto&& k : _keys) {
@@ -124,6 +135,7 @@ namespace irods
         template<typename T>
         T remove( const std::string& _key )
         {
+            auto lock = acquire_write_lock();
             nlohmann::json tmp;
 
             if (config_props_.contains(_key)) {
@@ -136,18 +148,41 @@ namespace irods
 
         void remove( const std::string& _key );
 
+        /// \brief Get a reference to the underlying json object. Writing
+        /// in particular should be done with caution as there is no guarantee
+        /// that this will be safe.
+        /// In threaded/asynchronous contexts, this function bypasses safety guarantees.
         nlohmann::json& map() noexcept
         {
             return config_props_;
         }
 
     private:
+        /// \brief Acquire a read lock, that is, a lock that is not exclusive, that is why it returns a shared_lock
+        /// here.
+        std::shared_lock<std::shared_mutex> acquire_read_lock() const
+        {
+            return std::shared_lock<std::shared_mutex>(property_mutex_);
+        }
+
+        /// \brief Acquire an exclusive lock for writing. While as many threads may read the properties object,
+        /// writing to the underlying std::map(or close enough, nlohmann uses red black trees) is not guaranteed to be
+        /// thread safe
+        ///
+        /// The unique lock is necessary here over scoped_lock or move_lock because the unique lock
+        /// supports move semantics.
+        std::unique_lock<std::shared_mutex> acquire_write_lock() const
+        {
+            return std::unique_lock(property_mutex_);
+        }
+
         server_properties();
 
         server_properties(const server_properties&) = delete;
         server_properties& operator=(const server_properties&) = delete;
 
         nlohmann::json config_props_;
+        mutable std::shared_mutex property_mutex_;
     }; // class server_properties
 
     template<>
--- a/lib/core/src/irods_server_properties.cpp
+++ b/lib/core/src/irods_server_properties.cpp
@@ -1,6 +1,7 @@
 #include "irods/irods_server_properties.hpp"
 
 #include "irods/irods_get_full_path_for_config_file.hpp"
+#include "irods/irods_logger.hpp"
 #include "irods/irods_stacktrace.hpp"
 #include "irods/rodsLog.h"
 
@@ -36,7 +37,10 @@ namespace irods
 
     namespace
     {
+        // clang-format off
         using json = nlohmann::json;
+        using log_server = irods::experimental::log::server;
+        // clang-format on
 
         struct string_t
         {
@@ -51,7 +55,7 @@ namespace irods
             void* tmp_ptr{};
 
             if (!stream) {
-                rodsLog (LOG_ERROR, "%s", "irodsCurl::write_string: NULL destination stream.");
+                log_server::error("irodsCurl::write_string: NULL destination stream.");
                 return 0;
             }
 
@@ -63,7 +67,7 @@ namespace irods
             // Add an extra byte for terminating null char.
             tmp_ptr = realloc(string->ptr, new_len + 1);
             if (!tmp_ptr) {
-                rodsLog(LOG_ERROR, "%s", "irodsCurl::write_string: realloc failed.");
+                log_server::error("irodsCurl::write_string: realloc failed.");
                 return -1;
             }
 
@@ -170,6 +174,7 @@ namespace irods
     template<>
     const nlohmann::json& server_properties::get_property<const nlohmann::json&>(const std::string& _key)
     {
+        auto lock = acquire_read_lock();
         if (auto prop = config_props_.find(_key); prop != config_props_.end()) {
             return *prop;
         }
@@ -180,6 +185,7 @@ namespace irods
     template<>
     const nlohmann::json& server_properties::get_property<const nlohmann::json&>(const configuration_parser::key_path_t& _keys)
     {
+        auto lock = acquire_read_lock();
         const json* tmp = &config_props_;
 
         for (auto&& k : _keys) {
@@ -206,6 +212,7 @@ namespace irods
 
     void server_properties::capture()
     {
+        auto lock = acquire_write_lock();
         if (use_remote_configuration()) {
             config_props_ = process_remote_configuration(capture_remote_configuration());
             return;
@@ -227,8 +234,53 @@ namespace irods
         config_props_ = json::parse(svr);
     } // capture
 
+    nlohmann::json server_properties::reload()
+    {
+        nlohmann::json new_values;
+        if (use_remote_configuration()) {
+            new_values = process_remote_configuration(capture_remote_configuration());
+        }
+        else {
+            std::string svr_fn;
+            irods::error ret = irods::get_full_path_for_config_file(SERVER_CONFIG_FILE, svr_fn);
+            if (!ret.ok()) {
+                THROW(
+                    ret.code(),
+                    fmt::format("[{}:{}] - Failed to find server config file [{}]", __func__, __LINE__, ret.result()));
+            }
+
+            std::ifstream svr{svr_fn};
+
+            if (!svr) {
+                THROW(FILE_OPEN_ERR, fmt::format("[{}:{}] - Failed to open server config file", __func__, __LINE__));
+            }
+            new_values = json::parse(svr);
+        }
+
+        auto patch = json::diff(config_props_, new_values);
+        for (auto i = patch.cbegin(); i != patch.cend(); ++i) {
+            // If we remove the runtime values the server will not be happy
+
+            // It may be prudent to have a list of runtime properties that need to be preserved instead, but
+            // I'm not entirely decided on that yet.
+            if (i.value().at("op").get_ref<const std::string&>() == "remove") {
+                i = patch.erase(i);
+                if (i == patch.cend()) {
+                    break;
+                }
+            }
+        }
+        auto lock = acquire_write_lock();
+        // TODO uncomment when issue #6470 is fixed
+        // log_server::error("Before patch '{}'", config_props_.dump());
+        config_props_.patch(patch);
+        // log_server::error("After patch '{}'", config_props_.dump());
+        return patch;
+    } // reload
+
     void server_properties::remove(const std::string& _key)
     {
+        auto lock = acquire_write_lock();
         if (config_props_.contains(_key)) {
             config_props_.erase(_key);
         }
@@ -249,14 +301,16 @@ namespace irods
                 return bytes;
             }
 
-            rodsLog(LOG_ERROR, "Invalid shared memory size for DNS cache [size=%d].", bytes);
+            log_server::error("Invalid shared memory size for DNS cache [size={}].", bytes);
         }
         catch (...) {
-            rodsLog(LOG_DEBUG, "Could not read server configuration property [%s.%s.%s].",
-                    KW_CFG_ADVANCED_SETTINGS, KW_CFG_DNS_CACHE, KW_CFG_SHARED_MEMORY_SIZE_IN_BYTES);
+            log_server::debug("Could not read server configuration property [{}.{}.{}].",
+                              KW_CFG_ADVANCED_SETTINGS,
+                              KW_CFG_DNS_CACHE,
+                              KW_CFG_SHARED_MEMORY_SIZE_IN_BYTES);
         }
 
-        rodsLog(LOG_DEBUG, "Returning default shared memory size for DNS cache [default=5000000].");
+        log_server::debug("Returning default shared memory size for DNS cache [default=5000000].");
 
         return 5'000'000;
     } // get_dns_cache_shared_memory_size
@@ -271,14 +325,16 @@ namespace irods
                 return seconds;
             }
 
-            rodsLog(LOG_ERROR, "Invalid eviction age for DNS cache [seconds=%d].", seconds);
+            log_server::error("Invalid eviction age for DNS cache [seconds={}].", seconds);
         }
         catch (...) {
-            rodsLog(LOG_DEBUG, "Could not read server configuration property [%s.%s.%s].",
-                    KW_CFG_ADVANCED_SETTINGS, KW_CFG_DNS_CACHE, KW_CFG_EVICTION_AGE_IN_SECONDS);
+            log_server::debug("Could not read server configuration property [{}.{}.{}].",
+                              KW_CFG_ADVANCED_SETTINGS,
+                              KW_CFG_DNS_CACHE,
+                              KW_CFG_EVICTION_AGE_IN_SECONDS);
         }
 
-        rodsLog(LOG_DEBUG, "Returning default eviction age for DNS cache [default=3600].");
+        log_server::debug("Returning default eviction age for DNS cache [default=3600].");
 
         return 3600;
     } // get_dns_cache_eviction_age
@@ -293,14 +349,16 @@ namespace irods
                 return bytes;
             }
 
-            rodsLog(LOG_ERROR, "Invalid shared memory size for Hostname cache [size=%d].", bytes);
+            log_server::error("Invalid shared memory size for Hostname cache [size={}].", bytes);
         }
         catch (...) {
-            rodsLog(LOG_DEBUG, "Could not read server configuration property [%s.%s.%s].",
-                    KW_CFG_ADVANCED_SETTINGS, KW_CFG_HOSTNAME_CACHE, KW_CFG_SHARED_MEMORY_SIZE_IN_BYTES);
+            log_server::debug("Could not read server configuration property [{}.{}.{}].",
+                              KW_CFG_ADVANCED_SETTINGS,
+                              KW_CFG_HOSTNAME_CACHE,
+                              KW_CFG_SHARED_MEMORY_SIZE_IN_BYTES);
         }
 
-        rodsLog(LOG_DEBUG, "Returning default shared memory size for Hostname cache [default=2500000].");
+        log_server::debug("Returning default shared memory size for Hostname cache [default=2500000].");
 
         return 2'500'000;
     } // get_hostname_cache_shared_memory_size
@@ -315,14 +373,16 @@ namespace irods
                 return seconds;
             }
 
-            rodsLog(LOG_ERROR, "Invalid eviction age for Hostname cache [seconds=%d].", seconds);
+            log_server::error("Invalid eviction age for Hostname cache [seconds={}].", seconds);
         }
         catch (...) {
-            rodsLog(LOG_DEBUG, "Could not read server configuration property [%s.%s.%s].",
-                    KW_CFG_ADVANCED_SETTINGS, KW_CFG_HOSTNAME_CACHE, KW_CFG_EVICTION_AGE_IN_SECONDS);
+            log_server::debug("Could not read server configuration property [{}.{}.{}].",
+                              KW_CFG_ADVANCED_SETTINGS,
+                              KW_CFG_HOSTNAME_CACHE,
+                              KW_CFG_EVICTION_AGE_IN_SECONDS);
         }
 
-        rodsLog(LOG_DEBUG, "Returning default eviction age for Hostname cache [default=3600].");
+        log_server::debug("Returning default eviction age for Hostname cache [default=3600].");
 
         return 3600;
     } // get_hostname_cache_eviction_age
--- a/plugins/microservices/src/get_server_property.cpp
+++ b/plugins/microservices/src/get_server_property.cpp
@@ -1,0 +1,105 @@
+#include "irods/irods_server_properties.hpp"
+#include "irods/irods_ms_plugin.hpp"
+#include "irods/irods_re_structs.hpp"
+#include "irods/msParam.h"
+#include "irods/rodsErrorTable.h"
+#include "irods/irods_error.hpp"
+#include "irods/irods_logger.hpp"
+
+#include <exception>
+#include <functional>
+#include <string>
+
+namespace
+{
+    using log_msi = irods::experimental::log::microservice;
+
+    auto msi_impl(msParam_t* _in_name, msParam_t* _out_value, ruleExecInfo_t* rei) -> int
+    {
+        if (!rei || !rei->rsComm) {
+            log_msi::error("msi_get_server_property: input rei or rsComm is NULL.");
+            return SYS_INTERNAL_NULL_INPUT_ERR;
+        }
+
+        // ensure that this user is an admin
+        if (rei->uoic->authInfo.authFlag < LOCAL_PRIV_USER_AUTH) {
+            log_msi::error("msi_get_server_property: User {} is not local admin. status = {}",
+                           rei->uoic->userName,
+                           CAT_INSUFFICIENT_PRIVILEGE_LEVEL);
+            return CAT_INSUFFICIENT_PRIVILEGE_LEVEL;
+        }
+
+        try {
+            const std::string name = parseMspForStr(_in_name);
+            const nlohmann::json* json_value = nullptr;
+            const auto& config = irods::server_properties::server_properties::instance().map();
+
+            try {
+                if (name.find("/") != std::string::npos) {
+                    json_value = &config.at(nlohmann::json::json_pointer(name));
+                }
+                else {
+                    json_value = &config.at(name);
+                }
+            }
+            catch (nlohmann::json::out_of_range& e) {
+                log_msi::error("Property not found. An error occurred {}", e.what());
+                return KEY_NOT_FOUND;
+            }
+            catch (nlohmann::json::parse_error& e) {
+                log_msi::error("Parse error in key. {}", e.what());
+                return KEY_NOT_FOUND;
+            }
+
+            const std::string value = json_value->dump();
+            fillStrInMsParam(_out_value, value.c_str());
+            return 0;
+        }
+        catch (const irods::exception& e) {
+            log_msi::error(e.what());
+            return e.code();
+        }
+        catch (const std::exception& e) {
+            log_msi::error(e.what());
+            return SYS_INTERNAL_ERR;
+        }
+        catch (...) {
+            log_msi::error("An unknown error occurred while processing the request.");
+            return SYS_UNKNOWN_ERROR;
+        }
+    } // msi_impl
+
+    template <typename... Args, typename Function>
+    auto make_msi(const std::string& name, Function func) -> irods::ms_table_entry*
+    {
+        auto* msi = new irods::ms_table_entry{sizeof...(Args)};
+        msi->add_operation(name, std::function<int(Args..., ruleExecInfo_t*)>(func));
+        return msi;
+    } // make_msi
+} // anonymous namespace
+
+extern "C" auto plugin_factory() -> irods::ms_table_entry*
+{
+    return make_msi<msParam_t*, msParam_t*>("msi_get_server_property", msi_impl);
+}
+
+#ifdef IRODS_FOR_DOXYGEN
+/// \brief Gets the value of a value in the server_properties map. Requires administrative
+/// permissions.
+///
+/// \param[in]      in_name   The name of the property to retrieve, may be a json pointer or a simple
+///                           name.
+/// \param[out]     out_value The value of the property. It is in serialized form if it is an object,
+///                           otherwise it is converted to a string.
+/// \param[in,out]  rei       The rule execution context. Ignore this parameter as a user.
+///
+/// \return An integer.
+/// \retval 0        On success
+/// \retval Non-zero On failure
+/// \retval CAT_INSUFFICIENT_PRIVILEGE_LEVEL The calling user isn't an administrator
+/// \retval KEY_NOT_FOUND The key provided does not have a set value in the configuration
+/// \retval INVALID_OBJECT_NAME The key provided is an invalid json pointer
+///
+/// \since 4.3.1
+auto msi_get_server_property(msParam_t* in_name, msParam_t out_value, ruleExecInfo_t* rei) -> int;
+#endif // IRODS_FOR_DOXYGEN
--- a/scripts/core_tests_list.json
+++ b/scripts/core_tests_list.json
@@ -3,6 +3,7 @@
     "test_auth",
     "test_catalog",
     "test_collection_mtime",
+    "test_configuration_reload",
     "test_control_plane",
     "test_delay_queue",
     "test_dynamic_peps",
--- a/scripts/irods/test/test_configuration_reload.py
+++ b/scripts/irods/test/test_configuration_reload.py
@@ -1,0 +1,85 @@
+import unittest
+
+import json
+import os
+
+from . import session
+from ..controller import IrodsController
+from .. import lib
+from .. import paths
+
+SessionsMixin = session.make_sessions_mixin([('otherrods', 'rods')], [])
+
+
+class TestConfigurationReload(SessionsMixin, unittest.TestCase):
+    def setUp(self):
+        super(TestConfigurationReload, self).setUp()
+        self.admin = self.admin_sessions[0]
+        self.testdir = os.path.join(self.admin.session_collection, 'testdir')
+
+    @staticmethod
+    def prepare_rule(property_name):
+        """Prepare a query about the server settings."""
+        return f"""
+        msi_get_server_property("{property_name}",*thing);
+        writeLine("stdout","*thing");
+        """
+
+    # Test to see that it recognizes the  configuration's change
+    def test_server_correctly_detects_change_in_server_config(self):
+        irodsctl = IrodsController()
+        server_config_filename = paths.server_config_path()
+        PROP_NAME = "is_issue_6456_done"
+        with lib.file_backed_up(server_config_filename):
+            with open(server_config_filename) as f:
+                svr_cfg = json.load(f)
+            svr_cfg[PROP_NAME] = "No?"
+            with open(server_config_filename, 'w') as f:
+                f.write(json.dumps(svr_cfg, sort_keys=True,
+                        indent=4, separators=(',', ': ')))
+
+            irodsctl.reload_configuration()
+            self.admin.assert_icommand(['irule', TestConfigurationReload.prepare_rule(
+                f"/{PROP_NAME}"), 'null', 'ruleExecOut'], "STDOUT", ["?"])
+            svr_cfg[PROP_NAME] = "#6456 is done"
+            with open(server_config_filename, 'w') as f:
+                f.write(json.dumps(svr_cfg, sort_keys=True,
+                        indent=4, separators=(',', ': ')))
+
+            irodsctl.reload_configuration()
+            self.admin.assert_icommand(['irule', TestConfigurationReload.prepare_rule(
+                f"/{PROP_NAME}"), 'null', 'ruleExecOut'], "STDOUT", ["#6456 is done"])
+
+    def test_how_long_reloading_takes(self):
+        test_count = 100
+        server_config_filename = paths.server_config_path()
+        with open(server_config_filename) as f:
+            svr_config = json.load(f)
+        property_name = "Hello"
+        with lib.file_backed_up(server_config_filename):
+            rule = TestConfigurationReload.prepare_rule(property_name)
+            for i in range(test_count):
+                svr_config[property_name] = i
+                with open(server_config_filename, 'w') as f:
+                    f.write(json.dumps(svr_config))
+
+                IrodsController().reload_configuration()
+                self.admin.assert_icommand(
+                    ['irule', rule, 'null', 'ruleExecOut'], "STDOUT", [str(i)])
+
+    def test_how_long_restarting_takes(self):
+        test_count = 100
+        server_config_filename = paths.server_config_path()
+        with open(server_config_filename) as f:
+            svr_config = json.load(f)
+        property_name = "Hello"
+        with lib.file_backed_up(server_config_filename):
+            rule = TestConfigurationReload.prepare_rule(property_name)
+            for i in range(test_count):
+                svr_config[property_name] = i
+                with open(server_config_filename, 'w') as f:
+                    f.write(json.dumps(svr_config))
+
+                IrodsController().restart(test_mode=True)
+                self.admin.assert_icommand(
+                    ['irule', rule, 'null', 'ruleExecOut'], "STDOUT", [str(i)])
--- a/scripts/irods/test/test_delay_queue.py
+++ b/scripts/irods/test/test_delay_queue.py
@@ -319,7 +319,7 @@ class Test_Delay_Queue(session.make_sessions_mixin([('otherrods', 'rods')], [('a
                 f.write(new_server_config)
 
             # Bounce server to apply setting
-            irodsctl.restart(test_mode=True)
+            irodsctl.reload_configuration()
 
             # Fire off rule and wait for message to get written out to serverLog
             self.admin.assert_icommand(['irule', '-F', rule_file], 'STDOUT_SINGLELINE', "rule queued")
@@ -370,8 +370,7 @@ class Test_Delay_Queue(session.make_sessions_mixin([('otherrods', 'rods')], [('a
                         config.client_environment_path,
                         {'irods_connection_pool_refresh_time_in_seconds' : 2})
 
-                    # Bounce server to apply setting
-                    irodsctl.restart(test_mode=True)
+                    irodsctl.reload_configuration()
 
                     # Fire off rule and ensure the delay queue is correctly populated
                     self.admin.assert_icommand(['irule', '-F', rule_file])
@@ -400,7 +399,7 @@ class Test_Delay_Queue(session.make_sessions_mixin([('otherrods', 'rods')], [('a
 
         finally:
             os.remove(rule_file)
-            irodsctl.restart(test_mode=True)
+            irodsctl.reload_configuration()
 
     @unittest.skipIf(plugin_name == 'irods_rule_engine_plugin-python', 'Delete this line on resolution of #4094')
     def test_sigpipe_in_delay_server(self):
@@ -432,8 +431,7 @@ class Test_Delay_Queue(session.make_sessions_mixin([('otherrods', 'rods')], [('a
                 with open(server_config_filename, 'w') as f:
                     f.write(new_server_config)
 
-                # Bounce server to apply setting
-                irodsctl.restart(test_mode=True)
+                irodsctl.reload_configuration()
 
                 # Fire off rule and wait for message to get written out to serverLog
                 self.admin.assert_icommand(['irule', '-F', rule_file], 'STDOUT_SINGLELINE', "rule queued")
@@ -459,7 +457,7 @@ class Test_Delay_Queue(session.make_sessions_mixin([('otherrods', 'rods')], [('a
         finally:
             os.remove(rule_file)
             self.admin.run_icommand(['iqdel', '-a'])
-            irodsctl.restart(test_mode=True)
+            irodsctl.reload_configuration()
 
     @unittest.skipIf(plugin_name == 'irods_rule_engine_plugin-python', 'Delete this line on resolution of #4094')
     @unittest.skipIf(test.settings.TOPOLOGY_FROM_RESOURCE_SERVER, 'odbc.ini file does not exist on Catalog Service Consumer')
@@ -500,8 +498,7 @@ class Test_Delay_Queue(session.make_sessions_mixin([('otherrods', 'rods')], [('a
                         config.client_environment_path,
                         {'irods_connection_pool_refresh_time_in_seconds' : 2})
 
-                    # Bounce server to apply setting
-                    irodsctl.restart(test_mode=True)
+                    irodsctl.reload_configuration()
                     with lib.file_backed_up(odbc_ini_file):
                         self.admin.assert_icommand(['irule', '-F', rule_file])
                         time.sleep(2)
@@ -536,7 +533,7 @@ class Test_Delay_Queue(session.make_sessions_mixin([('otherrods', 'rods')], [('a
             # Lower the delay server's sleep time so that rules are executed quicker.
             config.server_config['advanced_settings']['delay_server_sleep_time_in_seconds'] = 1
             lib.update_json_file_from_dict(config.server_config_path, config.server_config)
-            IrodsController().restart(test_mode=True)
+            IrodsController().reload_configuration()
 
             core_re_path = os.path.join(config.core_re_directory, 'core.re')
 
@@ -568,7 +565,7 @@ class Test_Delay_Queue(session.make_sessions_mixin([('otherrods', 'rods')], [('a
                 lib.delayAssert(lambda: assert_metadata_exists)
 
         # Restore the server's configuration settings.
-        IrodsController().restart(test_mode=True)
+        IrodsController().reload_configuration()
 
     @unittest.skipIf(plugin_name == 'irods_rule_engine_plugin-python', 'Skip for PREP and Topology Testing')
     def test_session_variables_can_be_used_in_delay_rules_indirectly(self):
@@ -578,7 +575,7 @@ class Test_Delay_Queue(session.make_sessions_mixin([('otherrods', 'rods')], [('a
             # Lower the delay server's sleep time so that rules are executed quicker.
             config.server_config['advanced_settings']['delay_server_sleep_time_in_seconds'] = 1
             lib.update_json_file_from_dict(config.server_config_path, config.server_config)
-            IrodsController().restart(test_mode=True)
+            IrodsController().reload_configuration()
 
             core_re_path = os.path.join(config.core_re_directory, 'core.re')
 
@@ -613,7 +610,7 @@ class Test_Delay_Queue(session.make_sessions_mixin([('otherrods', 'rods')], [('a
                 lib.delayAssert(lambda: assert_metadata_exists)
 
         # Restore the server's configuration settings.
-        IrodsController().restart(test_mode=True)
+        IrodsController().reload_configuration()
 
     @unittest.skipIf(plugin_name == 'irods_rule_engine_plugin-python', 'Skip for PREP and Topology Testing')
     def test_delay_rule_priority_level_defaults_to_5_when_no_priority_level_is_specified__issue_2759(self):
--- a/scripts/irods/test/test_icommands_file_operations.py
+++ b/scripts/irods/test/test_icommands_file_operations.py
@@ -63,7 +63,7 @@ class Test_ICommands_File_Operations(resource_suite.ResourceBase, unittest.TestC
             self.assertTrue(occur[1].group(1) == b'logical_path'    and occur[1].group(2).startswith(b'/'))
             self.assertTrue(occur[2].group(1) == b'proxy_user_name' and occur[2].group(2).decode('utf-8') == self.admin.username)
         finally:
-            IrodsController().restart()
+            IrodsController().reload_configuration()
 
     @unittest.skipUnless(plugin_name == 'irods_rule_engine_plugin-python', 'only applicable for python REP')
     def test_re_serialization__prep_55(self):
@@ -83,7 +83,7 @@ class Test_ICommands_File_Operations(resource_suite.ResourceBase, unittest.TestC
             self.assertTrue(1 == len(occur))
             self.assertTrue(occur[0].group(1) == b'user_rods_zone' and occur[0].group(2).decode('utf-8') == self.admin.zone_name)
         finally:
-            IrodsController().restart()
+            IrodsController().reload_configuration()
 
     def iput_r_large_collection(self, user_session, base_name, file_count, file_size):
         local_dir = os.path.join(self.testing_tmp_dir, base_name)
@@ -218,7 +218,7 @@ class Test_ICommands_File_Operations(resource_suite.ResourceBase, unittest.TestC
                 with open(server_config_filename, 'w') as f:
                     f.write(new_server_config)
 
-                IrodsController().restart()
+                IrodsController().reload_configuration()
 
                 # get log offset
                 initial_size_of_server_log = lib.get_file_size_by_path(IrodsConfig().server_log_path)
--- a/server/main_server/src/main_server_control_plane.cpp
+++ b/server/main_server/src/main_server_control_plane.cpp
@@ -547,7 +547,8 @@ namespace irods
     void server_control_executor::operator()()
     {
         signal( SIGINT, ctrl_plane_handle_shutdown_signal );
-        signal( SIGHUP, ctrl_plane_handle_shutdown_signal );
+        // SIGHUP is the signal to reload the configuration. This shouldn't interact with it
+        signal(SIGHUP, SIG_IGN);
         signal( SIGTERM, ctrl_plane_handle_shutdown_signal );
 
         int port, num_hash_rounds;
--- a/server/main_server/src/rodsServer.cpp
+++ b/server/main_server/src/rodsServer.cpp
@@ -5,6 +5,7 @@
 #include "irods/irods_configuration_keywords.hpp"
 #include "irods/irods_configuration_parser.hpp"
 #include "irods/irods_get_full_path_for_config_file.hpp"
+#include "irods/json_events.hpp"
 #include "irods/rcMisc.h"
 #include "irods/rodsErrorTable.h"
 #include "irods/rodsServer.hpp"
@@ -85,6 +86,8 @@ namespace hnc  = irods::experimental::net::hostname_cache;
 namespace dnsc = irods::experimental::net::dns_cache;
 // clang-format on
 
+using log_server = irods::experimental::log::server;
+
 struct sockaddr_un local_addr{};
 int agent_conn_socket{};
 bool connected_to_agent{};
@@ -98,6 +101,8 @@ int SvrSock;
 
 std::atomic<bool> is_control_plane_accepting_requests = false;
 int failed_delay_server_migration_communication_attempts = 0;
+
+std::atomic<bool> reload_server_config = false;
 
 agentProc_t* ConnectedAgentHead{};
 agentProc_t* ConnReqHead{};
@@ -134,21 +139,21 @@ namespace
         // Should only ever set the cache salt once.
         try {
             const auto& existing_salt = irods::get_server_property<const std::string>(irods::KW_CFG_RE_CACHE_SALT);
-            rodsLog(LOG_ERROR, "createAndSetRECacheSalt: salt already set [%s]", existing_salt.c_str());
+            log_server::error("createAndSetRECacheSalt: salt already set [{}]", existing_salt.c_str());
             return ERROR(SYS_ALREADY_INITIALIZED, "createAndSetRECacheSalt: cache salt already set");
         }
         catch (const irods::exception&) {
             irods::buffer_crypt::array_t buf;
             irods::error ret = irods::buffer_crypt::generate_key(buf, RE_CACHE_SALT_NUM_RANDOM_BYTES);
             if (!ret.ok()) {
-                rodsLog(LOG_ERROR, "createAndSetRECacheSalt: failed to generate random bytes");
+                log_server::error("createAndSetRECacheSalt: failed to generate random bytes");
                 return PASS(ret);
             }
 
             std::string cache_salt_random;
             ret = irods::buffer_crypt::hex_encode(buf, cache_salt_random);
             if (!ret.ok()) {
-                rodsLog(LOG_ERROR, "createAndSetRECacheSalt: failed to hex encode random bytes");
+                log_server::error("createAndSetRECacheSalt: failed to hex encode random bytes");
                 return PASS(ret);
             }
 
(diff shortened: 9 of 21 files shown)


scripts/irods/test/test_configuration_reload.py
```diff
@@ -0,0 +1,127 @@
...
+    # Test to see that it recognizes the  configuration's change
+    def test_configuration_change_seen(self):
+        irodsctl = IrodsController()
+        server_config_filename = paths.server_config_path()
+        with lib.file_backed_up(server_config_filename):
+            with open(server_config_filename) as f:
+                svr_cfg = json.load(f)
+            # Here I should make a few different data objects and set their metadata
+            # but for now I'm just curious that it's being seen.
+            svr_cfg["I_am_a_valid_property_name"] = 2
```
korydraughn, line 62: Consider pulling the property name out into a variable so that it can be used in other places. The same goes for the property value.

Consider using a property value other than `2`. Using `2` could lead to inaccurate test results. Use a string that helps with debugging tests. Many of the iRODS tests will include things in strings like the issue number or a phrase that's related to what is being tested.

scripts/irods/test/test_configuration_reload.py
```diff
@@ -0,0 +1,127 @@
...
+            TestConfigurationReload.signal_server(irodsctl.get_server_proc().pid,signals.SIGUSR1)
+            rc, stdout, stderr = self.admin.assert_icommand(['irule',
+                                                             #'-r',
+                                                             #'irods_rule_engine_plugin-irods_rule_language-instance',
+                                                             TestConfigurationReload.prepare_rule("I_am_a_valid_property_name"),
+                                                             'null',
+                                                             'ruleExecOut'],
+                                                            "STDOUT",
+                                                            ["2"])
+            print(stdout)
```
korydraughn, line 75: Seems this should be something like: `self.assertEqual(stdout, property_value)`

scripts/irods/test/test_configuration_reload.py
```diff
@@ -0,0 +1,127 @@
...
+
+            TestConfigurationReload.signal_server(irodsctl.get_server_proc().pid,signals.SIGUSR1)
+            rc, stdout, stderr = self.admin.assert_icommand(['irule',
+                                                             #'-r',
+                                                             #'irods_rule_engine_plugin-irods_rule_language-instance',
+                                                             TestConfigurationReload.prepare_rule("I_am_a_valid_property_name"),
+                                                             'null',
+                                                             'ruleExecOut'],
+                                                            "STDOUT",
+                                                            ["2"])
```
korydraughn, line 74: Indentation

server/core/include/irods/configuration_hook_manager.hpp
```diff
@@ -0,0 +1,65 @@
+#ifndef CONFIGURATION_HOOK_MANAGER_HPP
+#define CONFIGURATION_HOOK_MANAGER_HPP
```
korydraughn, line 2: Add `IRODS_` prefix

server/main_server/src/rodsServer.cpp
```diff
@@ -454,6 +455,12 @@ namespace
         }
     } // log_stacktrace_files
 
+
+    std::atomic<bool> will_reload_config = false;
```
korydraughn, line 459: Move this closer to the beginning of the file so that all globals are grouped in a general location.

server/core/src/configuration_hook_manager.cpp
```diff
@@ -0,0 +1,22 @@
+#include "irods/configuration_hook_manager.hpp"
+#include <fmt/format.h>
+#include "irods/irods_logger.hpp"
```
korydraughn, line 3: Reorder headers into groups.

server/core/src/configuration_hook_manager.cpp
```diff
@@ -0,0 +1,22 @@
...
+  configuration_hook_manager::configuration_hook_manager(){}
+  configuration_hook_manager &configuration_hook_manager::instance(){
+    static configuration_hook_manager i;
+    return i;
+  }
+  void configuration_hook_manager::add_hook(hook a){
+    hooks_.push_back(a);
+  }
+  void configuration_hook_manager::run_hooks(const std::vector<nlohmann::json>& changes){
+    irods::experimental::log::server::info(fmt::format("Found {} changes", changes.size()));
```
korydraughn, line 14: `fmt::format` isn't necessary.

The logger supports the same interface as fmtlib (mostly).

server/core/include/irods/configuration_hook_manager.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+#include <vector>
+#include <string>
+#include <nlohmann/json.hpp>
+/// \file
```
korydraughn, line 7: This directive should be placed above `#include` directives to match other iRODS files.

server/core/include/irods/configuration_hook_manager.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+namespace irods::experimental::hooks{
```
korydraughn, line 9: Why is the `hooks` namespace needed?
What else would go in this namespace?

server/core/include/irods/configuration_hook_manager.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+   */
+  using hook =
+    std::function<void(const std::string&,const nlohmann::json&,const nlohmann::json&)>;
+  ///
+  ///Provides a mechanism to trigger the execution of code
+  ///when a value in the configuration changes.
+  ///
+  ///Hooks are responsible for determining 
+  class configuration_hook_manager{
```
korydraughn, line 20: Why use a singleton class?
Why not use free functions?

For example, see the following:
- https://github.com/irods/irods/blob/4.3.0/lib/core/include/irods/hostname_cache.hpp
- https://github.com/irods/irods/blob/4.3.0/server/core/include/irods/replica_access_table.hpp

server/core/include/irods/configuration_hook_manager.hpp
```diff
@@ -0,0 +1,65 @@
...
+  public:
+    static configuration_hook_manager& instance();
+    void add_hook(hook a);
+    void run_hooks(const std::vector<nlohmann::json> &changes);
+  };
+  /// CRTP base class to add some hook behavior.
+  /// 
+  /// Classes that derive from this should implement a function called
+  /// configuration_value_changed that takes a string and the old value and new value.
+  template<typename T> class hooked_resource{
```
korydraughn, line 32: Where is this used?

server/core/include/irods/configuration_hook_manager.hpp
```diff
@@ -0,0 +1,65 @@
...
+    /// Non-owning pointers, to call into;
+    static inline std::vector<T*> instances_;
+    /// Has this particular hooked resource been added to the configuration_hook_manager yet?
+    static inline bool initialized_ = false;
+  public:
+    hooked_resource(){
+      if( !hooked_resource<T>::initialized_ ){
+        hooked_resource<T>::initialized_ = true;
+        configuration_hook_manager::instance()
+          .add_hook([](auto a, auto b,auto c){
```
korydraughn, line 42: This indicates copying of `a`, `b`, and `c`.

Is that intentional?

server/main_server/src/rodsServer.cpp
```diff
@@ -1920,7 +1943,6 @@ int initServerMain(rsComm_t *svrComm,
             migrate_delay_server(enable_test_mode, write_to_stdout);
         });
     ix::cron::cron::instance().add_task(delay_server.build());
-
```
korydraughn, line 1923: Let's restore this line.

scripts/irods/test/test_configuration_reload.py
```diff
@@ -0,0 +1,127 @@
...
+            # but for now I'm just curious that it's being seen.
+            svr_cfg["I_am_a_valid_property_name"] = 2
+            with open(server_config_filename, 'w') as f:
+                f.write(json.dumps(svr_cfg, sort_keys=True, indent=4, separators=(',',': ')))
+
+            TestConfigurationReload.signal_server(irodsctl.get_server_proc().pid,signals.SIGUSR1)
+            rc, stdout, stderr = self.admin.assert_icommand(['irule',
+                                                             #'-r',
+                                                             #'irods_rule_engine_plugin-irods_rule_language-instance',
+                                                             TestConfigurationReload.prepare_rule("I_am_a_valid_property_name"),
```
korydraughn, line 70: I'm wondering how clear this is?

Hmm ...

korydraughn (changes requested):
Looking good.

Try to use more empty lines to separate things. Adding additional whitespace helps improve readability.

Look at some of the newer libraries and see how they are formatted/structured. The following repo lists the majority of the modern iRODS libraries. Give their implementations and usage a look.
- https://github.com/irods/irods_library_examples

server/core/include/irods/configuration_hook_manager.hpp
epsilon-phase: I always forget I can just have some variables tucked away in a namespace. :)

server/core/include/irods/configuration_hook_manager.hpp
epsilon-phase: It isn't, I suppose that this should also probably be removed with the singleton class anyway :)

scripts/irods/test/test_configuration_reload.py
```diff
@@ -0,0 +1,127 @@
...
+    @staticmethod
+    def prepare_rule(property_name):
+        "Prepare a query about the server settings."
+        return f"""
+        msi_get_server_property("{property_name}",*thing);
+        writeLine("stdout","*thing");
+        """
+    @staticmethod
+    def signal_server(server_pid, signal):
+        os.kill(server_pid, signal)
```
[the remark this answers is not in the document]
epsilon-phase: This made more sense back before I added .reload_configuration to the irodscontroller. The tests should  just use that function anyway :)

plugins/microservices/src/get_server_property.cpp
```diff
@@ -93,7 +102,9 @@ auto plugin_factory() -> irods::ms_table_entry*
 /// \retval 0        On success
 /// \retval Non-zero On failure
 /// \retval CAT_INSUFFICIENT_PRIVILEGE_LEVEL The calling user isn't an administrator
+/// \retval KEY_NOT_FOUND The key provided does not have a set value in the confiuguration
```
trel, line 105: confiuguration -> configuration

scripts/core_tests_list.json
```diff
@@ -76,6 +76,7 @@
     "test_resource_types.Test_Resource_ReplicationWithinReplication",
     "test_resource_types.Test_Resource_Unixfilesystem",
     "test_resource_types.Test_Resource_WeightedPassthru",
+    "test_configuration_reload",
```
trel, line 79: please place in alpha order in the list

scripts/irods/test/test_configuration_reload.py
```diff
@@ -0,0 +1,107 @@
...
+        """
+
+
+    def test_signal(self):
+        irodsctl = IrodsController()
+        irodsctl.reload_configuration()
+        lib.delayAssert(lambda:\
+            lib.log_message_occurrences_greater_than_count(msg="Configuration Reloaded", count=0))
+
+    # Test to see that it recognizes the  configuration's change
```
trel, line 47: two spaces

scripts/irods/test/test_delay_queue.py
```diff
@@ -319,7 +320,8 @@ def test_delay_block_with_output_param__3906(self):
                 f.write(new_server_config)
 
             # Bounce server to apply setting
-            irodsctl.restart(test_mode=True)
+            #irodsctl.restart(test_mode=True)
```
trel, line 323: can remove once confirmed functional/working

scripts/irods/test/test_delay_queue.py
```diff
@@ -371,7 +373,8 @@ def test_delay_queue_connection_refresh(self):
                         {'irods_connection_pool_refresh_time_in_seconds' : 2})
 
                     # Bounce server to apply setting
-                    irodsctl.restart(test_mode=True)
+                    #irodsctl.restart(test_mode=True)
```
trel, line 376: same

scripts/irods/test/test_delay_queue.py
```diff
@@ -400,7 +403,8 @@ def test_delay_queue_connection_refresh(self):
 
         finally:
             os.remove(rule_file)
-            irodsctl.restart(test_mode=True)
+            #irodsctl.restart(test_mode=True)
```
trel, line 406: same

server/core/CMakeLists.txt
```diff
@@ -63,6 +63,7 @@ add_library(
   "${CMAKE_CURRENT_SOURCE_DIR}/src/specColl.cpp"
   "${CMAKE_CURRENT_SOURCE_DIR}/src/user_validation_utilities.cpp"
   "${CMAKE_CURRENT_SOURCE_DIR}/src/voting.cpp"
+  "${CMAKE_CURRENT_SOURCE_DIR}/src/configuration_hook_manager.cpp"
```
trel, line 66: alpha order

server/main_server/src/rodsServer.cpp
```diff
@@ -17,6 +17,7 @@
 #include "irods/irods_network_factory.hpp"
 #include "irods/irods_re_plugin.hpp"
 #include "irods/irods_server_properties.hpp"
+#include "irods/configuration_hook_manager.hpp"
```
trel, line 20: alpha order

server/main_server/src/rodsServer.cpp
```diff
@@ -1186,21 +1202,27 @@ int serverMain(const bool enable_test_mode = false, const bool write_to_stdout =
             if (irods::server_state::server_state::running != state) {
                 ix::log::server::info("Invalid iRODS server state [{}].", to_string(state));
             }
-
+            if(will_reload_config){
```
trel, line 1205: space after 'if'

lib/core/include/irods/irods_server_properties.hpp
```diff
@@ -10,6 +10,8 @@
 
 #include <string>
 #include <string_view>
+#include <mutex>
+#include <shared_mutex>
```
trel, line 14: alpha order

lib/core/src/irods_server_properties.cpp
```diff
@@ -3,6 +3,7 @@
 #include "irods/irods_get_full_path_for_config_file.hpp"
 #include "irods/irods_stacktrace.hpp"
 #include "irods/rodsLog.h"
+#include "irods/irods_logger.hpp"
```
trel, line 6: alpha order

server/core/src/configuration_hook_manager.cpp
```diff
@@ -0,0 +1,22 @@
+#include "irods/configuration_hook_manager.hpp"
+#include <fmt/format.h>
+#include "irods/irods_logger.hpp"
+namespace irods::experimental::hooks{
+  configuration_hook_manager::configuration_hook_manager(){}
```
trel, line 5: i think we're four spaces for indentation.... does the clang formatter pick this up?

korydraughn: Yes. It's captured in the clang-format github action output.

epsilon-phase: I didn't notice `git clang-format` until I was a few commits in.

server/core/include/irods/configuration_hook_manager.hpp
epsilon-phase: I wasn't sure how much infrastructure it was going to need, so putting it in its own namespace seemed like not a bad first decision

Probably doesn't need that with the current code

trel:
still a number of unaddressed/unresolved comments - make sure to flip open the accordion and mark things that are handled as 'resolved'.

plugins/microservices/src/get_server_property.cpp
```diff
@@ -0,0 +1,109 @@
...
+        if (rei->uoic->authInfo.authFlag < LOCAL_PRIV_USER_AUTH) {
+            log::microservice::error("msi_get_server_property: User {} is not local admin. status = {}",
+                                     rei->uoic->userName,
+                                     CAT_INSUFFICIENT_PRIVILEGE_LEVEL);
+            return CAT_INSUFFICIENT_PRIVILEGE_LEVEL;
+        }
+
+        try {
+            std::string name = parseMspForStr(_in_name);
+            nlohmann::json json_value;
```
korydraughn, line 36: Currently, we're copying the JSON object stored in the config object.

We can avoid that if this were a pointer.

plugins/microservices/src/get_server_property.cpp
```diff
@@ -0,0 +1,109 @@
...
+                    if (name[0] != '/') {
+                        name.insert(0, "/");
+                    }
+                    else {
+                        return KEY_NOT_FOUND;
+                    }
+                    continue;
+                }
+            }
+            while (false);
```
korydraughn, line 61: Let's add a new line after this loop.

scripts/irods/test/test_configuration_reload.py
```diff
@@ -0,0 +1,134 @@
...
+
+SessionsMixin = session.make_sessions_mixin([('otherrods','rods')],[])
+
+class TestConfigurationReload(SessionsMixin, unittest.TestCase):
+    def setUp(self):
+        super(TestConfigurationReload, self).setUp()
+        self.admin = self.admin_sessions[0]
+        self.testdir = os.path.join(self.admin.session_collection,'testdir')
+
+
```
korydraughn, line 19: Remove extra line.

scripts/irods/test/test_configuration_reload.py
```diff
@@ -0,0 +1,134 @@
...
+    @staticmethod
+    def prepare_rule(property_name):
+        """Prepare a query about the server settings."""
+        return f"""
+        msi_get_server_property("{property_name}",*thing);
+        writeLine("stdout","*thing");
+        """
+
+
```
korydraughn, line 28: Remove extra line.

scripts/irods/test/test_configuration_reload.py
```diff
@@ -0,0 +1,134 @@
...
+            irodsctl.reload_configuration()
+            self.admin.assert_icommand(['irule',
+                                        TestConfigurationReload.prepare_rule("/advanced_settings/delay_server_sleep_time_in_seconds"),
+                                        'null',
+                                        'ruleExecOut'],
+                                       "STDOUT",
+                                       ["1"])
+
+
```
korydraughn, line 91: Remove extra line.

scripts/irods/test/test_configuration_reload.py
```diff
@@ -0,0 +1,134 @@
...
+                IrodsController().reload_configuration()
+                self.admin.assert_icommand(['irule',
+                                            rule,
+                                            'null',
+                                            'ruleExecOut'],
+                                           "STDOUT",
+                                           [str(i)])
+
+
```
korydraughn, line 113: Remove extra line.

scripts/irods/test/test_delay_queue.py
```diff
@@ -157,6 +157,7 @@ def test_delay_queue_with_long_job(self):
                 # Bounce server to apply setting
                 irodsctl.restart(test_mode=True)
 
+
```
korydraughn, line 160: Remove extra line.

scripts/irods/test/test_delay_queue.py
```diff
@@ -432,8 +434,8 @@ def test_sigpipe_in_delay_server(self):
                 with open(server_config_filename, 'w') as f:
                     f.write(new_server_config)
 
-                # Bounce server to apply setting
-                irodsctl.restart(test_mode=True)
+                # Tell the server to apply setting
```
korydraughn, line 437: This comment is unnecessary now. Your `reload_configuration()` function is clear.

server/core/src/configuration_hook_manager.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include "irods/irods_logger.hpp"
+#include <unordered_map>
+
+namespace irods::experimental
+{
+    namespace
+    {
+        std::vector<hook> hooks;
+        std::unordered_multimap<std::string, size_t> edge_sets;
+    } // namespace
```
korydraughn, line 12: Add new line.

server/core/src/configuration_hook_manager.cpp
```diff
@@ -0,0 +1,33 @@
...
+    void add_hook(const auto& hook_triggers, hook&& hook)
+    {
+        size_t id = hooks.size();
+        // Take ownership, we don't need to deal with any missing closures.
+        // those can be nasty to debug.
+        hooks.emplace_back(std::move(hook));
+        for (const auto& name : hook_triggers) {
+            edge_sets.insert(name, id);
+        }
+    }
```
korydraughn, line 22: Add new line.

server/core/include/irods/configuration_hook_manager.hpp
```diff
@@ -0,0 +1,17 @@
+#ifndef CONFIGURATION_HOOK_MANAGER_HPP
+#define CONFIGURATION_HOOK_MANAGER_HPP
+
+#include <vector>
+#include <string>
+#include <nlohmann/json.hpp>
```
korydraughn, line 6: Please reorder to:
```cpp
#include <nlohmann/json.hpp>

#include <vector>
#include <string>
```

We include things in the following order:
1. iRODS headers
2. Third-party headers (grouped for readability)
3. System headers
4. C headers (`<cstring>`)
5. C++ STL headers

server/core/include/irods/configuration_hook_manager.hpp
```diff
@@ -0,0 +1,17 @@
...
+/// \file
+
+namespace irods::experimental
+{
+    /// \brief Hooks are functions called with the pathname of the json value that has changed, as well as the new
+    /// operation performed, removal versus replacement versus insertion etc
+    using hook = std::function<void(const std::string&, const nlohmann::json&)>;
+    void add_hook(const auto& hooks, hook&& hook);
+    void run_hooks(const nlohmann::json& changes);
+} // namespace irods::experimental
```
korydraughn, line 16: Add new line.

server/core/include/irods/configuration_hook_manager.hpp
```diff
@@ -0,0 +1,17 @@
...
+#endif
```
korydraughn, line 17:
```cpp
#endif // IRODS_CONFIGURATION_HOOK_MANAGER_HPP
```

server/core/src/configuration_hook_manager.cpp
```diff
@@ -0,0 +1,33 @@
...
+    void run_hooks(const nlohmann::json& changes)
+    {
+        for (const auto& operation : changes) {
+            std::string name = operation["path"];
```
korydraughn, line 26: Implicit casts should be avoided.
```cpp
const auto& name = operation.at("path").get_ref<const std::string&>();
```

server/main_server/src/rodsServer.cpp
```diff
@@ -1131,7 +1161,7 @@ int serverMain(const bool enable_test_mode = false, const bool write_to_stdout =
                 }
 
                 using namespace std::chrono_literals;
-                std::this_thread::sleep_for(100ms);
+                std::this_thread::sleep_for(10ms);
```
korydraughn, line 1164: What is the motivation behind this change?

server/main_server/src/rodsServer.cpp
```diff
@@ -1186,24 +1217,28 @@ int serverMain(const bool enable_test_mode = false, const bool write_to_stdout =
...
 
-                rodsLog(LOG_NOTICE, "serverMain: select() error, errno = %d", errno);
+                ix::log::server::info("serverMain: select() error, errno = {}", errno);
                 return -1;
             }
 
+            // Busy wait for 'latency'
+            while (will_reload_config) {
+            }
+
```
korydraughn, line 1241: Tagging this for later discussion.

server/main_server/src/rodsServer.cpp
```diff
@@ -2222,19 +2280,18 @@ int procSingleConnReq(agentProc_t* connReq)
...
+                              status);
     }
     else {
-        rodsLog(LOG_DEBUG,
-                "Agent process %d started for puser=%s and cuser=%s from %s",
-                connReq->pid,
-                connReq->startupPack.proxyUser,
-                connReq->startupPack.clientUser,
-                inet_ntoa(connReq->remoteAddr.sin_addr));
+        ix::log::server::debug("Agent process {} started for puser={} and cuser={} from {}",
```
korydraughn, line 2290: We should change all instances of `ix::log::server` to `log_server`.

server/main_server/src/rodsServer.cpp
```diff
@@ -462,11 +469,11 @@ namespace
         signal(SIGPIPE, SIG_IGN);
 #ifdef osx_platform
         signal(SIGINT, (sig_t) serverExit);
-        signal(SIGHUP, (sig_t) serverExit);
+        signal(SIGHUP, (sig_t) reload_next_loop);
```
korydraughn, line 472: Consider changing `reload_next_loop`.

For example: `set_reload_server_config_flag`

server/main_server/src/rodsServer.cpp
korydraughn: `will_reload_config` -> `reload_server_config`

server/core/include/irods/configuration_hook_manager.hpp
```diff
@@ -0,0 +1,17 @@
...
+#include <nlohmann/json.hpp>
+/// \file
+
+namespace irods::experimental
+{
+    /// \brief Hooks are functions called with the pathname of the json value that has changed, as well as the new
+    /// operation performed, removal versus replacement versus insertion etc
+    using hook = std::function<void(const std::string&, const nlohmann::json&)>;
+    void add_hook(const auto& hooks, hook&& hook);
+    void run_hooks(const nlohmann::json& changes);
```
korydraughn, line 15: Needs documentation.

korydraughn (changes requested):

plugins/microservices/src/get_server_property.cpp
```diff
@@ -0,0 +1,109 @@
...
+                    return KEY_NOT_FOUND;
+                }
+                catch (nlohmann::json::parse_error& e) {
+                    if (name[0] != '/') {
+                        name.insert(0, "/");
+                    }
+                    else {
+                        return KEY_NOT_FOUND;
+                    }
+                    continue;
```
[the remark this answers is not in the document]
epsilon-phase: I suppose that should just log an error instead of trying to correct the user's json path syntax :)

server/main_server/src/rodsServer.cpp
epsilon-phase: This was a hack before I got around to writing the spinlock. It also happens to negate the single benefit of the spinlock that I had in mind anyway, the latency, so I will replace the spinlock with a barrier.

lib/core/src/configuration_hook_manager.cpp
```diff
@@ -0,0 +1,44 @@
...
+        for (const auto& name : hook_triggers) {
+            edge_sets.insert(name, id);
+        }
+    }
+
+    void run_hooks(const nlohmann::json& changes)
+    {
+        // These are hooks that run for every key, seems useful for plugins that
+        // want to be especially nosy about configuration changes
+        auto catch_all_hooks = edge_sets.equal_range("*CATCHALL*");
```
korydraughn, line 29: Please explain the idea behind `*CATCHALL*`

lib/core/src/configuration_hook_manager.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+        for (const auto& operation : changes) {
+            const std::string& name = operation["path"].get_ref<const std::string&>();
+            // The catchall stuff is given priority here on a hunch that it makes more sense.
```
korydraughn, line 33: Please elaborate more on this.

If `*CATCHALL*` is special, we need to document it well. Otherwise, no one will remember what it is used for.

lib/core/src/irods_server_properties.cpp
```diff
@@ -38,6 +38,7 @@ namespace irods
     {
         using json = nlohmann::json;
 
+        using log_server = irods::experimental::log::server;
         struct string_t
```
korydraughn, line 42: Let's change this to the following:
```cpp
// clang-format off
using json       = nlohmann::json;
using log_server = irods::experimental::log::server;
// clang-format on

struct string_t
```

plugins/microservices/src/get_server_property.cpp
```diff
@@ -0,0 +1,105 @@
...
+                return KEY_NOT_FOUND;
+            }
+
+            const std::string value = json_value->dump();
+            fillStrInMsParam(_out_value, value.c_str());
+            return 0;
+        }
+        catch (const irods::exception& e) {
+            log_msi::error(e.what());
+            return SYS_INTERNAL_ERR;
```
korydraughn, line 60: `return e.code();`

plugins/microservices/src/get_server_property.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+#ifdef IRODS_FOR_DOXYGEN
+/// \brief Gets the value of a value in the server_properties map. Requires administrative
+/// permissions
+///
+/// \param[in]      in_name   The name of the property to retrieve, may be a json pointer or a simple
+/// name
+/// \param[out]     out_value The value of the property. It is in serialized form if it is an object,
+/// otherwise it is converted to a string
+/// \param[in,out]  rei       The rule execution context. Ignore this parameter as a user.
```
korydraughn, line 94: Format the comments so that they are easier to read in code.
```cpp
/// \param[in]     in_name   The name of the property to retrieve, may be a json pointer or a simple
///                          name.
/// \param[out]    out_value The value of the property. It is in serialized form if it is an object,
///                          otherwise it is converted to a string.
/// \param[in,out] rei       The rule execution context. Ignore this parameter as a user.
```

plugins/microservices/src/get_server_property.cpp
```diff
@@ -0,0 +1,105 @@
...
+} // anonymous namespace
+
+extern "C" auto plugin_factory() -> irods::ms_table_entry*
+{
+    return make_msi<msParam_t*, msParam_t*>("msi_get_server_property", msi_impl);
+}
+
+#ifdef IRODS_FOR_DOXYGEN
+/// \brief Gets the value of a value in the server_properties map. Requires administrative
+/// permissions
```
korydraughn, line 88: Missing period on second sentence.

scripts/irods/test/test_configuration_reload.py
```diff
@@ -0,0 +1,98 @@
...
+            with open(server_config_filename, 'w') as f:
+                f.write(json.dumps(svr_cfg, sort_keys=True, indent=4, separators=(',',': ')))
+
+            irodsctl.reload_configuration()
+            self.admin.assert_icommand(['irule',
+                                        TestConfigurationReload.prepare_rule(f"/{PROP_NAME}"),
+                                        'null',
+                                        'ruleExecOut'],
+                                       "STDOUT",
+                                       ["?"])
```
korydraughn, line 45: Indentation

scripts/irods/test/test_configuration_reload.py
```diff
@@ -0,0 +1,98 @@
...
+            with open(server_config_filename, 'w') as f:
+                f.write(json.dumps(svr_cfg, sort_keys=True, indent=4, separators=(',',': ')))
+
+            irodsctl.reload_configuration()
+            self.admin.assert_icommand(['irule',
+                                        TestConfigurationReload.prepare_rule(f"/{PROP_NAME}"),
+                                        'null',
+                                        'ruleExecOut'],
+                                       "STDOUT",
+                                       ["#6456 is done"])
```
korydraughn, line 56: Indentation

scripts/irods/test/test_configuration_reload.py
```diff
@@ -0,0 +1,98 @@
...
+                with open(server_config_filename,'w') as f:
+                    f.write(json.dumps(svr_config))
+
+                IrodsController().reload_configuration()
+                self.admin.assert_icommand(['irule',
+                                            rule,
+                                            'null',
+                                            'ruleExecOut'],
+                                           "STDOUT",
+                                           [str(i)])
```
korydraughn, line 77: Indentation

scripts/irods/test/test_configuration_reload.py
```diff
@@ -0,0 +1,98 @@
...
+                with open(server_config_filename,'w') as f:
+                    f.write(json.dumps(svr_config))
+
+                IrodsController().restart(test_mode=True)
+                self.admin.assert_icommand(['irule',
+                                            rule,
+                                            'null',
+                                            'ruleExecOut'],
+                                           "STDOUT",
+                                           [str(i)])
```
korydraughn, line 98: Indentation

scripts/irods/test/test_delay_queue.py
```diff
@@ -319,7 +319,8 @@ def test_delay_block_with_output_param__3906(self):
                 f.write(new_server_config)
 
             # Bounce server to apply setting
-            irodsctl.restart(test_mode=True)
+            #irodsctl.restart(test_mode=True)
```
korydraughn, line 322: Remove

scripts/irods/test/test_icommands_file_operations.py
```diff
@@ -218,7 +218,8 @@ def test_iput_r_with_kw(self):
                 with open(server_config_filename, 'w') as f:
                     f.write(new_server_config)
 
-                IrodsController().restart()
+                #IrodsController().restart()
```
korydraughn, line 221: Remove

server/main_server/src/rodsServer.cpp
```diff
@@ -1119,6 +1130,18 @@ int serverMain(const bool enable_test_mode = false, const bool write_to_stdout =
...
     }
 
+    {
+        ix::cron::cron_builder configuration_reloader;
+        configuration_reloader.interval(0).task([]() {
+            if (reload_server_config) {
+                reload_server_config = false;
+                auto diff_list = irods::server_properties::server_properties::instance().reload();
+                ix::run_hooks(diff_list);
+                log_server::info("Configuration Reloaded");
```
korydraughn, line 1140: Add period to message.

server/main_server/src/rodsServer.cpp
```diff
@@ -1184,26 +1208,32 @@ int serverMain(const bool enable_test_mode = false, const bool write_to_stdout =
...
-                rodsLog(LOG_NOTICE, "serverMain: select() error, errno = %d", errno);
+                log_server::info("serverMain: select() error, errno = {}", errno);
                 return -1;
             }
 
+            if (reload_server_config) {
+                // As a side effect of reading the server properties object, we acquire a lock
+                // that cannot be had while the configuration is in the middle of reloading.
+                irods::server_properties::instance().contains("");
+            }
```
korydraughn, line 1235: What happens if we don't wait for the server to finish reloading?

Are there any risks?

Is it a bad idea to block until the server finishes reloading its config?

server/main_server/src/main_server_control_plane.cpp
```diff
@@ -547,7 +547,8 @@ namespace irods
     void server_control_executor::operator()()
     {
         signal( SIGINT, ctrl_plane_handle_shutdown_signal );
-        signal( SIGHUP, ctrl_plane_handle_shutdown_signal );
+        // SIGHUP is the signal to reload the configuration. This shouldn't interact with it
+        signal(SIGHUP, SIG_IGN);
```
korydraughn, line 551: How did you arrive as this being the solution?

plugins/microservices/src/get_server_property.cpp
```diff
@@ -0,0 +1,105 @@
...
+            log_msi::error("msi_get_server_property: User {} is not local admin. status = {}",
+                           rei->uoic->userName,
+                           CAT_INSUFFICIENT_PRIVILEGE_LEVEL);
+            return CAT_INSUFFICIENT_PRIVILEGE_LEVEL;
+        }
+
+        try {
+            std::string name = parseMspForStr(_in_name);
+            const nlohmann::json* json_value = nullptr;
+            const auto& config = irods::server_properties::server_properties::instance().map();
```
korydraughn, line 35: We have to be careful with `.map()` moving forward now that the server properties use locks.

`.map()` bypasses the locks and returns a reference to the underlying JSON object. Given that the agents are mostly single threaded, this is fine. However, we need to document the pitfalls of using `.map()` in an asynchronous context where the server properties are being updated.

lib/core/src/configuration_hook_manager.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include "irods/irods_logger.hpp"
+
+#include <unordered_map>
+
+namespace irods::experimental
+{
+    namespace
+    {
+        std::vector<hook> hooks;
+        std::unordered_multimap<std::string, size_t> edge_sets;
```
korydraughn, line 11: Please document these variables.

These don't require doxygen style comments.

Make sure to explain any details needed for proper maintenance of this library. Without good explanations of these variables, someone will be tempted to re-implement this 6 to 12 months from now.

lib/core/include/irods/irods_server_properties.hpp
```diff
@@ -142,12 +154,26 @@ namespace irods
...
 
     private:
+        /// @brief Acquire a read lock, that is, a lock that is not exclusive, that is why it returns a shared_lock
+        /// here.
+        std::shared_lock<std::shared_mutex> acquire_read_lock() const
+        {
+            return std::shared_lock<std::shared_mutex>(property_mutex_);
+        }
+
+        /// @brief Acquire an exclusive lock for writing(Hence the unique_lock)
```
korydraughn, line 164: Please elaborate more on the reason why `std::unique_lock` is required for correctness.

It is important for the next reader.

korydraughn (changes requested):
Looking good!

Consider writing a unit test that proves `irods::server_properties` correctly protects its data in a multi-threaded context.

For the python tests, if we removed the blocking logic for reloads (in rodsServer.cpp), we can use `lib.delayAssert()` to handle the asynchronous nature of the reloads. This, of course, depends on whether we decide the blocking logic should be removed.

server/main_server/src/main_server_control_plane.cpp
epsilon-phase: The main server kept on having trouble with the control plane being terminated.

The control plane forwards the termination right back to the main server as it exits.

plugins/microservices/src/get_server_property.cpp
epsilon-phase: I'll add a note about that to the function in the irods_server_properties

server/main_server/src/rodsServer.cpp
epsilon-phase: > What happens if we don't wait for the server to finish reloading?

> Are there any risks?

There aren't any risks other than inconsistency, 

> Is it a bad idea to block until the server finishes reloading its config?

as to the blocking, well, how bad are we assuming disk latency is for our users? Is it high enough to make that a problem?

lib/core/include/irods/configuration_hook_manager.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+namespace irods::experimental::configuration_reload
+{
+    /// \brief Hooks are functions called with the pathname of the json value that has changed, as well as the new
+    /// operation performed, removal versus replacement versus insertion etc
+    using hook = std::function<void(const std::string&, const nlohmann::json&)>;
+
+    /// \brief adds a hook to be run whenever a key on the list of hooks changes.
+    /// \param hooks A list of strings to match
+    /// \param hook The function to run when the string matches
```
korydraughn, line 19: We forgot to include the `[in]` part of the `\param` directive.

lib/core/include/irods/irods_server_properties.hpp
```diff
@@ -136,18 +148,37 @@ namespace irods
 
         void remove( const std::string& _key );
 
+        /// \brief Get a reference to the underlying json object. Writing
+        /// in particular should be done with caution as there is no guarantee
+        /// that this will be safe.
```
korydraughn, line 153: Add a line stating use of this function bypasses safety guarantees when called in an asynchronous context.

lib/core/include/irods/irods_server_properties.hpp
```diff
@@ -136,18 +148,37 @@ namespace irods
...
+        {
+            return std::shared_lock<std::shared_mutex>(property_mutex_);
+        }
+
+        /// @brief Acquire an exclusive lock for writing. While as many threads may read the properties object,
+        /// writing to the underlying std::map(or close enough, nlohmann uses red black trees) is not guaranteed to be
+        /// thread safe
+        std::unique_lock<std::shared_mutex> acquire_write_lock() const
+        {
+            return std::unique_lock(property_mutex_);
```
korydraughn, line 172: Add a comment stating `std::unique_lock` is used instead of `std::scoped_lock` and `std::lock_guard` because it supports move semantics.

While using the other types could work (because of RVO), we don't want to rely on that, therefore `std::unique_lock` is the correct choice.

lib/core/src/configuration_hook_manager.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+    void run_hooks(const nlohmann::json& changes)
+    {
+        // These are hooks that run for every key, seems useful for plugins that
+        // want to be especially nosy about configuration changes
+        auto catch_all_hooks = edge_sets.equal_range("*CATCHALL*");
+
+        for (const auto& operation : changes) {
+            const std::string& name = operation["path"].get_ref<const std::string&>();
+            // The catchall stuff is given priority here on a hunch.
```
korydraughn, line 37: What is the motivation behind this comment?

korydraughn (changes requested):

unit_tests/src/test_server_properties.cpp
```diff
@@ -29,3 +31,38 @@ TEST_CASE("server_properties", "[exceptions]")
...
+        auto& instance = irods::server_properties::instance();
+        const auto limit = 1000;
+
+        std::thread a([&]() {
+            std::string key;
+            for (auto i = 0; i < limit; i++) {
+                fmt::format_to(std::back_inserter(key), "{}", i);
+                instance.set_property(key, i);
+                key.clear();
+            }
```
korydraughn, line 48: Why not use `std::to_string(i)` instead of `fmt::format_to`?

_This also applies to the other thread._

unit_tests/src/test_server_properties.cpp
```diff
@@ -29,3 +31,38 @@ TEST_CASE("server_properties", "[exceptions]")
                                  irods::KW_CFG_DB_PASSWORD));
     }
 }
+
+TEST_CASE("server_property_threadsafe", "[exceptions]")
+{
+    SECTION("Threads don't blow up")
+    {
```
korydraughn, line 38: Consider consolidating these since there's only one test.

e.g. `TEST_CASE("server_properties is thread-safe", "[thread-safety]")`

unit_tests/src/test_server_properties.cpp
```diff
@@ -29,3 +31,38 @@ TEST_CASE("server_properties", "[exceptions]")
...
+                    REQUIRE(instance.get_property<int>(key) == i);
+                    i++;
+                }
+                else {
+                    continue;
+                }
+            }
+        });
+        a.join();
+        b.join();
```
korydraughn, line 66: Consider using `irods::at_scope_exit` for each of these to guard against exceptions.

unit_tests/src/test_server_properties.cpp
```diff
@@ -29,3 +31,38 @@ TEST_CASE("server_properties", "[exceptions]")
...
+            }
+        });
+        std::thread b([&]() {
+            auto i = 0;
+            std::string key;
+            while (i < limit) {
+                key.clear();
+                fmt::format_to(std::back_inserter(key), "{}", i);
+                if (instance.contains(key)) {
+                    REQUIRE(instance.get_property<int>(key) == i);
```
korydraughn, line 57: This may require `DYNAMIC_REQUIRE` since it's in a loop.

unit_tests/src/test_server_properties.cpp
```diff
@@ -29,3 +31,38 @@ TEST_CASE("server_properties", "[exceptions]")
...
+    SECTION("Threads don't blow up")
+    {
+        auto& instance = irods::server_properties::instance();
+        const auto limit = 1000;
+
+        std::thread a([&]() {
+            std::string key;
+            for (auto i = 0; i < limit; i++) {
+                fmt::format_to(std::back_inserter(key), "{}", i);
+                instance.set_property(key, i);
```
korydraughn, line 46: Prefer using the free functions over getting a reference to the underlying instance.

Using the free functions covers more code and is more inline with how the server properties will be used.

lib/core/include/irods/irods_server_properties.hpp
```diff
@@ -151,8 +151,7 @@ namespace irods
         /// \brief Get a reference to the underlying json object. Writing
         /// in particular should be done with caution as there is no guarantee
         /// that this will be safe.
-        /// In threaded/asynchronous contexts, this function
-        /// bypasses the safety guarantees provided by the locking.
+        /// In threaded/asynchronous contexts, this function bypasses safety guarantees
```
korydraughn, line 154: Needs a period.

korydraughn (changes requested):
Because the test involves threads, add comments that explain what the test is trying to accomplish.

unit_tests/src/test_server_properties.cpp
epsilon-phase: As far as I can tell it works fine

lib/core/src/configuration_hook_manager.cpp
epsilon-phase: `*CATCHALL*` hooks are triggered on every configuration value change

The reason it looks like that is that I don't think anyone is likely to stumble into a namespace collision with that.

lib/core/src/irods_server_properties.cpp
```diff
@@ -18,7 +19,6 @@
 #if LIBCURL_VERSION_NUM < 0x071503
     #include <curl/types.h>
 #endif
-
```
korydraughn, line 21: Let's restore this line.

lib/core/include/irods/configuration_hook_manager.hpp
```diff
@@ -0,0 +1,28 @@
...
+namespace irods::experimental::configuration_reload
+{
+    /// \brief Hooks are functions called with the pathname of the json value that has changed, as well as the new
+    /// operation performed, removal versus replacement versus insertion etc
+    using hook = std::function<void(const std::string&, const nlohmann::json&)>;
+
+    /// \brief adds a hook to be run whenever a key on the list of hooks changes.
+    /// \param [in] hooks A list of strings to match
+    /// \param [in] hook The function to run when the string matches
+    void add_hook(const auto& hooks, hook&& hook);
```
korydraughn, line 20: `const auto& hooks` is going to need an example.

What is meant by `"A list of strings to match"`?
We need to be clear on the requirements of this function, else no one will know how to use it properly.

Put yourself in the shoes of a developer who is seeing this for the first time.
What do you think they would think after reading the documentation?

We want to remove as much guesswork as possible.

lib/core/src/configuration_hook_manager.cpp
```diff
@@ -0,0 +1,47 @@
...
+        hooks.emplace_back(std::move(hook));
+        for (const auto& name : hook_triggers) {
+            edge_sets.insert(name, id);
+        }
+    }
+
+    void run_hooks(const nlohmann::json& changes)
+    {
+        // The *CATCHALL* hook is triggered on every change in configuration value.
+        auto catch_all_hooks = edge_sets.equal_range("*CATCHALL*");
```
korydraughn, line 32: I'm still very confused by the use of `"*CATCHALL*"`.

Is it used somewhere else?
Is this something the caller is supposed to know about?

I don't see any other references to it in the PR.

lib/core/src/configuration_hook_manager.cpp
```diff
@@ -0,0 +1,47 @@
...
+    {
+        // This is where the function objects are stored, given that they will not be removed or reordered, their index
+        // is what edge_sets' value refers to.
+        std::vector<hook> hooks;
+
+        // Edge set, as in edge-triggered, contains an association between the changed property and an index into hooks.
+        std::unordered_multimap<std::string, size_t> edge_sets;
+    } // namespace
+
+    void add_config_reload_hook(const auto& hook_triggers, hook&& hook)
```
korydraughn, line 18: Seems this should be called `add_hook`?

korydraughn (changes requested):
Take a moment and put yourself in the next developer's shoes. Imagine you've never seen the configuration reload interface.

Now ask yourself what is this missing? Not from the standpoint of features, but from the standpoint of someone being able to understand how it is supposed to work and be used.

How can it be improved?

lib/core/include/irods/configuration_hook_manager.hpp
epsilon-phase: Okay, the idea is that they can pass it something like an array or std::vector, or std::list I guess.

You're definitely right that need to expand on this

server/main_server/src/rodsServer.cpp
```diff
@@ -1184,26 +1208,32 @@ int serverMain(const bool enable_test_mode = false, const bool write_to_stdout =
...
                 }
 
-                rodsLog(LOG_NOTICE, "serverMain: select() error, errno = %d", errno);
+                log_server::info("serverMain: select() error, errno = {}", errno);
                 return -1;
             }
 
+            if (reload_server_config) {
+                // As a side effect of reading the server properties object, we acquire a lock
+                // that cannot be had while the configuration is in the middle of reloading.
```
alanking, line 1233: I'm a little slow, but I think it may be helpful to say explicitly *why* we are making this call here in addition to the *how*. In other words, we are attempting to read the server properties object *because* we wish to block here until configuration reload is completed so that all following connections will receive the updated configuration, whereas attempting to acquire the lock - as described here - is *how* we are blocking.

If this seems extraneous or incorrect to others, just ignore me :)

lib/core/include/irods/configuration_hook_manager.hpp
```diff
@@ -0,0 +1,43 @@
...
+    /// of addition.
+    using hook = std::function<void(const std::string&, const nlohmann::json&)>;
+
+    /// \brief Adds a hook to be run whenever a key on the list of hooks changes.
+    ///        \see hook
+    /// \param [in] hook_triggers An array or container of json paths to match.
+    ///
+    ///             The special value "*CATCHALL*" matches any property that has changed and gets run first
+    /// \param [in] hook The function to run when the json path matches
+    void add_hook(const auto& hook_triggers, hook&& hook);
```
alanking, line 35: Should we add a unit test or three for this hook system? I don't see where it is being used (maybe I missed it), so we should at least test it if we are going to include it. It would be fine if we just create an issue for this so we can move along since this is a main-only change.

lib/core/src/configuration_hook_manager.cpp
```diff
@@ -0,0 +1,47 @@
...
+        hooks.emplace_back(std::move(hook));
+        for (const auto& name : hook_triggers) {
+            edge_sets.insert(name, id);
+        }
+    }
+
+    void run_hooks(const nlohmann::json& changes)
+    {
+        // The *CATCHALL* hook is triggered on every change in configuration value.
+        auto catch_all_hooks = edge_sets.equal_range("*CATCHALL*");
```
alanking, line 32: Should we capture the `"*CATCHALL*"` string in a constant somewhere?

alanking (commented):
Seems real nice

Sorry if I'm retreading old ground with some of the comments, I have not been "in on" everything so far :)

lib/core/include/irods/configuration_hook_manager.hpp
korydraughn: Yes, we should add tests for this. However, due to time, let's capture that in a new issue.

server/main_server/src/rodsServer.cpp
trel: confirmed - comments should capture *why*.

lib/core/include/irods/configuration_hook_manager.hpp
korydraughn, line 35: Add an example showing usage of this function.

You can use this as a reference:
- https://github.com/irods/irods/blob/4.3.0/lib/api/include/irods/dataObjChksum.h

And here is that same documentation rendered as HTML:
- https://docs.irods.org/4.3.0/doxygen/dataObjChksum_8h.html#a1e907fe7c1f3353786302dee490d7301

If you want to see render your documentation to see how it will look, I can help with that. We have a repo for doing just that. Just let me know.

lib/core/include/irods/configuration_hook_manager.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+    /// \brief Run the hooks, the changes are a list of json patches.
+    /// \param [in] changes The list of changes in json patch format, see
+    /// RFC6902 for more details about that format.
+    void run_hooks(const nlohmann::json& changes);
```
korydraughn, line 40: Add an example for this function too.

lib/core/include/irods/configuration_hook_manager.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+#include <vector>
+#include <string>
+
+namespace irods::experimental::configuration_reload
+{
+    /// \brief Hooks are functions called with the pathname of the json value that has changed, as well as the new
+    /// operation performed, removal versus replacement versus insertion etc
+    ///
+    /// Hooks are invoked in two stages for each entry in a list of json patches
```
korydraughn, line 16: Missing periods.

lib/core/include/irods/configuration_hook_manager.hpp
```diff
@@ -0,0 +1,43 @@
...
+    /// For example, if you change the delay server's sleep duration, the
+    /// "*CATCHALL*" hooks will be run first in order of addition, followed by
+    /// the hooks associated with
+    /// "/advanced_settings/delay_server_sleep_time_in_seconds", also in order
+    /// of addition.
+    using hook = std::function<void(const std::string&, const nlohmann::json&)>;
+
+    /// \brief Adds a hook to be run whenever a key on the list of hooks changes.
+    ///        \see hook
+    /// \param [in] hook_triggers An array or container of json paths to match.
```
korydraughn, line 31: What are the requirements on the container? Support for begin/end iterators?

We need to add a statement stating the requirements on the container.

unit_tests/src/test_server_properties.cpp
```diff
@@ -29,3 +32,27 @@ TEST_CASE("server_properties", "[exceptions]")
                                  irods::KW_CFG_DB_PASSWORD));
     }
 }
+
+// The goal here is to ensure that the server_properties object is locked and unlocked
+// safely while using the normal api
+TEST_CASE("server_properties is threadsafe", "[thread-safety]")
```
korydraughn, line 38: Have we seen this compile/run?

korydraughn (changes requested):

unit_tests/src/test_server_properties.cpp
epsilon-phase: Yeah, it runs successfully

trel:
> still a number of unaddressed/unresolved comments - make sure to flip open the accordion and mark things that are handled as 'resolved'.

bump :)

lib/core/include/irods/configuration_hook_manager.hpp
epsilon-phase: Honestly it wasn't too hard to write the test for this

epsilon-phase:
I think this is ready for merge.

trel:
still a lot of unresolved accordian comments above.    please resolve those and then consider how you want to squash these 65 commits down to the 'right' number.

trel:
All are resolved!    Good stuff.

I see a few clang-tidy opinions from the checker.  @korydraughn do we have time to lean into those?   Or shall we begin the great squashing?

lib/core/include/irods/configuration_hook_manager.hpp
```diff
@@ -0,0 +1,117 @@
...
+        // Edge set, as in edge-triggered, contains an association between the changed property and an index into hooks.
+        std::unordered_multimap<std::string, size_t> edge_sets;
+    } // namespace
+
+    /// \brief Adds a hook to be run whenever a key on the list of hooks changes.
+    ///        \see hook
+    /// \param [in] hook_triggers An array or container of json paths to match, this
+    ///             container must support begin and end iterators.
+    ///
+    ///             The special value CATCHALL_KEYWORD matches any property that has changed and gets run first
```
korydraughn, line 49: `CATCHALL_KEYWORD` -> `\p KW_CATCHALL`

The `\p` applies special formatting so that it stands out more. This tells the developer reading the docs that this symbol is special.

lib/core/include/irods/configuration_hook_manager.hpp
```diff
@@ -0,0 +1,117 @@
...
+    /// \param [in] hook The function to run when the json path matches.
+    ///
+    /// \b Example
+    /// \code{.cpp}
+    /// namespace cr = irods::experimental::configuration_reload;
```
korydraughn, line 54: Add a new line after this.

lib/core/include/irods/configuration_hook_manager.hpp
```diff
@@ -0,0 +1,117 @@
...
+    /// cr::add_hook({cr::CATCHALL_KEYWORD}, [&](const std::string& json_path,
+    ///                                          const nlohmann::json& json_patch_op)
+    /// {
+    ///    if(json_path.substr(0,29) != "/example_configuration_object") {
+    ///         return;
+    ///    }
+    ///    handle_update(json_patch_op);
+    /// });
```
korydraughn, line 62: Add a new line after this.

lib/core/include/irods/configuration_hook_manager.hpp
```diff
@@ -0,0 +1,117 @@
...
+    ///    // Update the "PI" global value in the static lua state.
+    ///    if(json_patch_op.at("op")=="insert"||json_patch_op.at("op")=="replace"){
+    ///        lua_pushnumber( Lua_state, json_patch_op.at("value").get<double>());
+    ///        lua_setglobal( Lua_state, "PI");
+    ///    }
+    /// });
+    /// \endcode
+    void add_hook(const auto& hook_triggers, hook&& hook)
+    {
+        size_t id = hooks.size();
```
korydraughn, line 74: `size_t` -> `std::size_t`

lib/core/include/irods/configuration_hook_manager.hpp
```diff
@@ -0,0 +1,117 @@
...
+    using hook = std::function<void(const std::string&, const nlohmann::json&)>;
+
+    namespace
+    {
+        // This is where the function objects are stored, given that they will not be removed or reordered, their index
+        // is what edge_sets' value refers to.
+        std::vector<hook> hooks;
+
+        // Edge set, as in edge-triggered, contains an association between the changed property and an index into hooks.
+        std::unordered_multimap<std::string, size_t> edge_sets;
```
korydraughn, line 41: `size_t` -> `std::size_t`

lib/core/include/irods/configuration_hook_manager.hpp
```diff
@@ -0,0 +1,117 @@
...
+        // those can be nasty to debug.
+        hooks.emplace_back(std::move(hook));
+        for (const auto& name : hook_triggers) {
+            edge_sets.insert(std::make_pair(name, id));
+        }
+    }
+
+    /// \brief Run the hooks, the changes are a list of json patches.
+    /// \param [in] changes The list of changes in json patch format, see
+    /// RFC6902 for more details about that format.
```
korydraughn, line 85: Add new line after this.

lib/core/include/irods/configuration_hook_manager.hpp
```diff
@@ -0,0 +1,117 @@
...
+        // This is where the function objects are stored, given that they will not be removed or reordered, their index
+        // is what edge_sets' value refers to.
+        std::vector<hook> hooks;
+
+        // Edge set, as in edge-triggered, contains an association between the changed property and an index into hooks.
+        std::unordered_multimap<std::string, size_t> edge_sets;
+    } // namespace
+
+    /// \brief Adds a hook to be run whenever a key on the list of hooks changes.
+    ///        \see hook
```
korydraughn, line 45: Add a new line after this.

lib/core/include/irods/configuration_hook_manager.hpp
```diff
@@ -0,0 +1,117 @@
...
+        size_t id = hooks.size();
+        // Take ownership, we don't need to deal with any missing closures.
+        // those can be nasty to debug.
+        hooks.emplace_back(std::move(hook));
+        for (const auto& name : hook_triggers) {
+            edge_sets.insert(std::make_pair(name, id));
+        }
+    }
+
+    /// \brief Run the hooks, the changes are a list of json patches.
```
korydraughn, line 83: Add a new line after this.

lib/core/include/irods/configuration_hook_manager.hpp
```diff
@@ -0,0 +1,117 @@
...
+    /// \param [in] changes The list of changes in json patch format, see
+    /// RFC6902 for more details about that format.
+    /// \code{.cpp}
+    /// namespace cr = irods::experimental::configuration_reload;
+    /// cr::add_hook({cr::CATCHALL_KEYWORD}[](const std::string& json_path, const nlohmann::json& patch){
+    ///    std::cout<<patch<<std::endl;
+    /// });
```
korydraughn, line 90: Missing call to `run_hooks()`.

lib/core/include/irods/configuration_hook_manager.hpp
```diff
@@ -0,0 +1,117 @@
...
+    /// \endcode
+    /// Change the delay_server_sleep_time_in_seconds in the configuration and send a
+    /// SIGHUP signal to the toplevel irods process. The server should write
+    /// something like
+    /// `{ "path":"/advanced_settings/delay_server_sleep_time_in_seconds",
+    ///    "op":"replace", "value":<something> }` to stdout.
```
korydraughn, line 96: These lines can be embedded in the example to help explain `run_hooks()`.

lib/core/include/irods/configuration_hook_manager.hpp
```diff
@@ -0,0 +1,117 @@
...
+            for (auto it = catch_all_hooks.first; it != catch_all_hooks.second; ++it) {
+                hooks[it->second](name, operation);
+            }
+
+            auto range = edge_sets.equal_range(name);
+            for (auto it = range.first; it != range.second; ++it) {
+                hooks[it->second](name, operation);
+            }
+        }
+    }
```
korydraughn, line 114: The definition of this function needs to move to the cpp file to avoid ODR violations

lib/core/include/irods/configuration_hook_manager.hpp
```diff
@@ -0,0 +1,117 @@
+#ifndef IRODS_CONFIGURATION_HOOK_MANAGER_HPP
```
korydraughn, line 1: This library is not actually specific to just configuration. We need to give this a more general name.

Discush!

unit_tests/CMakeLists.txt
```diff
@@ -45,6 +45,7 @@ set(
   fixed_buffer_resource
   get_file_descriptor_info
   hierarchy_parser
+  hooks
```
korydraughn, line 48: `hooks` is too general.

Waiting for library rename discussion results.

korydraughn (changes requested):

lib/core/include/irods/configuration_hook_manager.hpp
```diff
@@ -0,0 +1,117 @@
...
+
+    namespace
+    {
+        // This is where the function objects are stored, given that they will not be removed or reordered, their index
+        // is what edge_sets' value refers to.
+        std::vector<hook> hooks;
+
+        // Edge set, as in edge-triggered, contains an association between the changed property and an index into hooks.
+        std::unordered_multimap<std::string, size_t> edge_sets;
+    } // namespace
```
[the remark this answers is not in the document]
epsilon-phase: That complicates the implementation of add_hook, that needs to be in the header so long as I'm using auto parameters, if I understand that bit right.

I'm undecided on how to resolve that

lib/core/include/irods/configuration_hook_manager.hpp
epsilon-phase: I mean, it's not *really* meant to be run by the user normally :p 

I should mention that somewhere shouldn't I?

lib/core/include/irods/configuration_hook_manager.hpp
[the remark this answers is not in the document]
korydraughn: I see. What you can do is give the anonymous namespace a name (i.e. `detail`).

The `detail` namespace is private to the implementation and provides signal to the developer that they shouldn't use anything inside of it. Inside that namespace would be two functions that simply return references to each of those variables.

Now your template can use the functions inside of the `detail` namespace without issue.

The functions also allow us to change things without having to touch `add_hook`.

lib/core/include/irods/configuration_hook_manager.hpp
korydraughn: Absolutely, but consider developers who are writing clients. They may want to use this. For those cases, they would have to work out when `run_hooks()` should be invoked in their code.

Or perhaps I'm misunderstanding something?

epsilon-phase: Fair enough, I hadn't considered there being legitimate uses for it on the client side.

epsilon-phase: Also putting it in the examples makes them way easier to write

lib/core/include/irods/configuration_hook_manager.hpp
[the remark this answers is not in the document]
epsilon-phase: Hmm, I'm running into a double free segfault at exit now, I'm not quite sure what to do about that.

lib/core/include/irods/configuration_hook_manager.hpp
korydraughn: Has this been resolved?

korydraughn:
The squashed commit is good. Splitting the commit into more focused chunks of work would be better (i.e. reload logic vs log messages), but I don't feel the main work is lost or difficult to spot within its current form.

Change the first line of the commit message to something like:
```bash
[5816] Main server now reloads its config on SIGHUP signal.
```
And in the commit message body, do the following:
- Replace the back ticks with double quotes or single quotes.
- The use of "propagate events" feels incorrect. Try rewording that statement.

@trel @alanking Any thoughts?

trel:
If there is more than one conceptual piece here... then yes, more commits is good.   But I'm okay with a single one.

Commit message needs to be thorough.

epsilon-phase:
Hmm, how does that look?

trel:
very close!

first paragraph is not just about tests... it's about allowing a running server to be refreshed without a restart - which is a big deal for servers with constant load (a restart breaks their transfers, requires a maintenance window, etc.).

"this could allows" -> "this could allow"

epsilon-phase:
Yeah, split the stuff about the tests getting faster into a second paragraph.

trel:
yeps, i'm good with the commit message.

i'll defer on any other code changes before the # gets added.

hooray.

korydraughn:
Commit message looks good. I think this is ready for the `#`.

korydraughn:
Has this comment been resolved? https://github.com/irods/irods/pull/6456#discussion_r931102178

lib/core/include/irods/configuration_hook_manager.hpp
epsilon-phase: Yeah

epsilon-phase:
Yeah, it has

korydraughn:
Due to time, we can ignore the clang-tidy checks. It hasn't reported anything horrible.

lib/core/include/irods/irods_server_properties.hpp
```diff
@@ -136,18 +148,41 @@ namespace irods
...
+        /// in particular should be done with caution as there is no guarantee
+        /// that this will be safe.
+        /// In threaded/asynchronous contexts, this function bypasses safety guarantees.
         nlohmann::json& map() noexcept
         {
             return config_props_;
         }
 
     private:
+        /// @brief Acquire a read lock, that is, a lock that is not exclusive, that is why it returns a shared_lock
```
alanking, line 161: `@brief` -> `\brief`

Same on 168

alanking (commented):
I think it looks good.

Saw one thing which we can fix later if necessary

alanking (approved):

trel:
ship it.

Issue #6459, opened 2022-06-29
`setup_irods.py` assumes `default_resource_name` to be present in `server_config`


dworkin:
- [x] main
- [x] 4-3-stable
---
### iRODS Version, OS and Version

iRODS 4.3.0, CentOS 7

### What did you try to do?

Configure iRODS with `json_configuration_file`

### Expected behavior

iRODS is configured

### Observed behavior (including steps to reproduce, if applicable)

Configuration fails on line 110 of `setup_irods.py`:
```
        default_resource_name = json_configuration_dict['default_resource_name']
```
This looks for `default_resource_name` at the top level. However, it is present within json object `server_config`.

*Edit*: my assumption was incorrect. It should be present at the top level, but *not* in `server_config`. However, `setup_irods.py` looks for it in both locations.

dworkin:
My assumption that `default_resource_name` should not be at the top level turns out to be incorrect.  However, on line 147 of `setup_irods.py` it is assumed to be present in `server_config`.

PR #6460, opened 2022-06-29
[#6459] look for `default_resource_name` in the correct location.

dworkin:
`default_resource_name` is found in `server_config`.

### DIFF
--- a/scripts/setup_irods.py
+++ b/scripts/setup_irods.py
@@ -107,7 +107,7 @@ def setup_server(irods_config, json_configuration_file=None, test_mode=False):
         # role
         irods_config.commit(json_configuration_dict['server_config'], irods.paths.server_config_path())
         # default resource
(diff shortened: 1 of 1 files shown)


dworkin:
> Thanks for the PR, however this change is incorrect based on the following schemas:
> 
>     * https://github.com/irods/irods/blob/main/schemas/configuration/v4/unattended_installation.json.in
> 
>     * https://github.com/irods/irods/blob/main/schemas/configuration/v4/server_config.json.in

In that case, there is an error on line 147 of `setup_irods.py`, where it is assumed that `default_resource_name` is present in `server_config`.  Removing `default_resource_name` from `server_config` does indeed cause an error there during unattended installation.

korydraughn:
Excellent catch.

Please create an issue for that so that we don't forget to fix it.

trel:
well, we've already got #6459, can we just repurpose/reword it?

PR #6466, opened 2022-07-08
Prevent inappropriate downgrade of a rodsadmin in the same zone

phdavis1027:
Prevents a rodsadmin from downgrading either themselves or any other rodsadmin in the local zone.


server/api/src/rsGeneralAdmin.cpp
```diff
@@ -53,56 +53,55 @@ namespace
...
-                     _new_value != "rodsadmin"
-                   )
-                {
-                   THROW(SYS_NOT_ALLOWED, "Rodsadmin cannot downgrade their own privileges.");
-                }
+            if ( admins_in_zone.find(static_cast<std::string>(_user_name)) != admins_in_zone.end() &&
+                    _new_value != "rodsadmin"
+                )
+            {
+                THROW(SYS_NOT_ALLOWED, "Cannot downgrade a rodsadmin running another server in your zone.");
```
trel, line 96: "another" -> "a"

"your" -> "this"

Do we want to log this as well?

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -53,56 +53,55 @@ namespace
...
+            }
+
+            if ( std::string(rsComm.clientUser.userName) == _user_name &&
+                    _new_value != "rodsadmin"
+                )
+            {
+                THROW(SYS_NOT_ALLOWED, "Rodsadmin cannot downgrade their own privileges.");
```
trel, line 103: "A rodsadmin account cannot downgrade its own privileges."

Do we want to log this as well?

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -52,6 +53,57 @@ namespace
         return std::find_if(b, e, [](unsigned char _ch) { return ::isspace(_ch); }) != e;
     } // contains_whitespace
 
+     auto throw_if_inappropriate_rodsadmin_downgrade(
+        RsComm& rsComm,
+        const std::string_view& _new_value,
+        const std::string_view& _user_name
+    ) -> void {
+            std::set<std::string> admins_in_zone;
```
trel, line 61: consider...  "server_administrators_in_zone" ?

server/api/src/rsGeneralAdmin.cpp
phdavis1027: For both of these, would it be acceptable to follow the pattern of the other ```throw_if_...``` functions here and add to the connection's error message?

trel: I think so.   Consistency is good.

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -52,6 +53,63 @@ namespace
...
+
+            if (const auto ec = rsZoneReport(&rsComm, &bbuf); ec < 0 ) {
+                const auto msg = fmt::format("[{}-{}] - Error obtaining zone report", __func__, __LINE__);
+                addRErrorMsg(&rsComm.rError, ec, msg.c_str());
+                THROW(ec, msg);
+            }
+
+            nlohmann::json zone_report = nlohmann::json::parse(
+                    std::string(static_cast<char *>(bbuf->buf), bbuf->len)
+            )["zones"];
```
korydraughn, line 74: I'm pretty sure this can be rewritten as:
```cpp
const auto* buf = static_cast<char*>(bbuf->buf);
const auto zone_report = nlohmann::json::parse(buf, buf + bbuf->len).at("zones");
```

This avoids the construction of `std::string` which may result in a memory allocation. And if nothing modifies this object, then we can mark it as `const`.

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -52,6 +53,63 @@ namespace
...
+                    );
+
+                    for ( const auto& server : zone["servers"] ) {
+                        server_administrators_in_zone.insert(
+                            static_cast<std::string>(server["service_account_environment"]["irods_user_name"])
+                        );
+                    }
+
+                }
+            }
```
korydraughn, line 92: I see a few things that would improve this code.

- When using the JSON library, they recommend avoiding implicit conversions.
    - This recommendation is good in general.
- You perform the same lookup multiple times instead of caching it.
- While `static_cast` probably results in the same machine code, using `.get<T>()` is more readable.

With that said, this code can be rewritten like so:
```cpp
const auto* const local_zone_name = getLocalZoneName();

for (const auto& zone : zone_report) {
    const auto& svc_acct_env = zone.at("icat_server").at("service_account_environment");

    if (svc_acct_env.at("irods_zone_name").get_ref<const std::string&>() != local_zone_name) {
        continue;
    }

    server_administrators_in_zone.insert(svc_acct_env.at("irods_user_name").get<std::string>());

    for (const auto& server : zone.at("servers")) {
        server_administrators_in_zone.insert(server.at("service_account_environment").at("irods_user_name").get<std::string>());
    }
}
```

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -52,6 +53,63 @@ namespace
...
+
+            if ( server_administrators_in_zone.find(static_cast<std::string>(_user_name)) != server_administrators_in_zone.end() &&
+                    _new_value != "rodsadmin"
+                )
+            {
+                const auto msg = fmt::format("[{}-{}] - Cannot downgrade another rodsadmin running another server in this zone.", __func__, __LINE__);
+                addRErrorMsg(&rsComm.rError, SYS_NOT_ALLOWED, msg.c_str());
```
korydraughn, line 99: Remove this line.

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -52,6 +53,63 @@ namespace
         return std::find_if(b, e, [](unsigned char _ch) { return ::isspace(_ch); }) != e;
     } // contains_whitespace
 
+     auto throw_if_inappropriate_rodsadmin_downgrade(
+        RsComm& rsComm,
```
korydraughn, line 57: This can be marked as `const` if it isn't modified by this function.

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -52,6 +53,63 @@ namespace
...
+            if ( server_administrators_in_zone.find(static_cast<std::string>(_user_name)) != server_administrators_in_zone.end() &&
+                    _new_value != "rodsadmin"
+                )
+            {
+                const auto msg = fmt::format("[{}-{}] - Cannot downgrade another rodsadmin running another server in this zone.", __func__, __LINE__);
+                addRErrorMsg(&rsComm.rError, SYS_NOT_ALLOWED, msg.c_str());
+                THROW(SYS_NOT_ALLOWED, msg);
+            }
+
+            if ( std::string(rsComm.clientUser.userName) == _user_name &&
```
korydraughn, line 103: `std::string_view` supports `operator==()` for string literals, so the use of `std::string` is unnecessary.
```cpp
if (rsComm.clientUser.userName == _user_name && _new_value != "rodsadmin") {
```

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -52,6 +53,63 @@ namespace
         return std::find_if(b, e, [](unsigned char _ch) { return ::isspace(_ch); }) != e;
     } // contains_whitespace
 
+     auto throw_if_inappropriate_rodsadmin_downgrade(
```
korydraughn, line 56: After reading through the function, I suspect we can add a check at the beginning that causes it to return immediately if `_new_value` is `"rodsadmin"`.

If `_new_value` is `"rodsadmin"`, then there's nothing to do.

This also means the checks near the end of the function handling `"rodsadmin"` shouldn't be needed anymore because, at that point, we know we're doing a downgrade of the user's type.

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -52,6 +53,63 @@ namespace
...
+        RsComm& rsComm,
+        const std::string_view& _new_value,
```
korydraughn, line 58: Consider renaming this parameter. `_new_value` is a little too general.

Perhaps, `_new_user_type`?

korydraughn (changes requested):

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -52,6 +53,63 @@ namespace
...
+        const std::string_view& _user_name
+    ) -> void {
+            std::set<std::string> server_administrators_in_zone;
+
+            BytesBuf* bbuf = nullptr;
+            irods::at_scope_exit free_buf { [&bbuf] { std::free(bbuf); }  };
+
+            if (const auto ec = rsZoneReport(&rsComm, &bbuf); ec < 0 ) {
+                const auto msg = fmt::format("[{}-{}] - Error obtaining zone report", __func__, __LINE__);
+                addRErrorMsg(&rsComm.rError, ec, msg.c_str());
```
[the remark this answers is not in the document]
phdavis1027: Replacing with call to rodsLog with level LOG_ERROR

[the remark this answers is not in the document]
korydraughn: Logging isn't necessary because you're throwing an exception. The information should be logged in the catch clause.

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -52,6 +53,59 @@ namespace
...
+    ) -> void {
+            if ("rodsadmin" == _new_user_type) return;
+
+            std::set<std::string> server_administrators_in_zone;
+
+            BytesBuf* bbuf = nullptr;
+            irods::at_scope_exit free_buf { [&bbuf] { std::free(bbuf); }  };
+
+            if (const auto ec = rsZoneReport(&rsComm, &bbuf); ec < 0 ) {
+                const auto msg = fmt::format("[{}-{}] - Error obtaining zone report", __func__, __LINE__);
```
korydraughn, line 69: Remove the use of the function name and line number from all error messages.

Take a look at the other `throw_if_*` functions for examples.

For this particular case, we probably want to change the message to something more generic because the use of `rsZoneReport()` is an implementation detail that will likely confuse the user.

Perhaps ... `"The server experienced an unexpected error while processing the request. Please report to the administrator."`

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -52,6 +53,59 @@ namespace
         return std::find_if(b, e, [](unsigned char _ch) { return ::isspace(_ch); }) != e;
     } // contains_whitespace
 
+     auto throw_if_user_manages_an_irods_server(
+        RsComm& rsComm,
+        const std::string_view _new_user_type,
+        const std::string_view _user_name
```
korydraughn, line 59: Let's swap the order of these parameters as well.

I mentioned this in a different comment, but we need to consider what the client is trying to change.

Therefore, this function should accept 4 parameters.
```cpp
RsComm& _comm,
const std::string_view _option,
const std::string_view _user_name,
const std::string_view _new_user_type
```
Then, if `_option` is not equal to `"type"`, we return early.

korydraughn (changes requested):
With these changes, I think all we'll need is a few tests to verify things work.

That includes making sure the error messages are reported and are what we expect on the client-side.

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -52,6 +53,59 @@ namespace
         return std::find_if(b, e, [](unsigned char _ch) { return ::isspace(_ch); }) != e;
     } // contains_whitespace
 
+     auto throw_if_user_manages_an_irods_server(
```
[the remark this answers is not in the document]
korydraughn: Ignore this comment.

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -52,6 +53,59 @@ namespace
...
+                    );
+                }
+
+            }
+            if ( server_administrators_in_zone.find(_user_name.data()) != server_administrators_in_zone.end() ) {
+                const auto msg = fmt::format("[{}-{}] - Cannot downgrade another rodsadmin running another server in this zone.", __func__, __LINE__);
+                rodsLog(LOG_ERROR, msg.c_str());
+                THROW(SYS_NOT_ALLOWED, msg);
+            }
+            if ( std::string(rsComm.clientUser.userName) == _user_name ) {
```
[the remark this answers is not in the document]
phdavis1027: Would it be worth just rearranging the checks for the sake of distinguishing two different ways that the user could have issued a bad command?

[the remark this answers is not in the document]
korydraughn: Please elaborate.

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -52,6 +53,59 @@ namespace
...
+
+            std::set<std::string> server_administrators_in_zone;
+
+            BytesBuf* bbuf = nullptr;
+            irods::at_scope_exit free_buf { [&bbuf] { std::free(bbuf); }  };
+
+            if (const auto ec = rsZoneReport(&rsComm, &bbuf); ec < 0 ) {
+                const auto msg = fmt::format("[{}-{}] - Error obtaining zone report", __func__, __LINE__);
+                rodsLog(LOG_ERROR, msg.c_str());
+                THROW(ec, msg);
```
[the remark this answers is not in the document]
korydraughn: Has this actually been resolved?

I still see use of `rodsLog` instead of `addRErrorMsg`.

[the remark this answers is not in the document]
phdavis1027: Oops, meant to mark this as in-progress.

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -52,6 +53,58 @@ namespace
...
+
+                for ( const auto& server : zone.at("servers") ) {
+                    server_administrators_in_zone.insert(
+                        server.at("service_account_environment").at("irods_user_name").get<std::string>()
+                    );
+                }
+            }
+
+            if ( server_administrators_in_zone.find(_user_name.data()) != server_administrators_in_zone.end() ) {
+                const auto msg = "ERROR: Cannot downgrade another rodsadmin running another server in this zone.";
```
trel, line 102: consider putting the attempted username in the error message?  and the other servername... in case that helps the admin of a large zone.

trel:
and i see a comma at the end of the commit message

server/api/src/rsGeneralAdmin.cpp
phdavis1027: Funnily enough, I actually did that before second-guessing myself. Whoops.

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -52,6 +53,51 @@ namespace
...
+        }
+
+        const auto buf = static_cast<char*> (bbuf->buf);
+        const nlohmann::json zone_report = nlohmann::json::parse(buf, buf + bbuf->len);
+        const auto& zones = zone_report.at("zones");
+
+        const auto * local_zone_name = getLocalZoneName();
+
+        for ( const auto& zone : zones ) {
+            const auto icat_server_env = zone.at("icat_server").at("service_account_environment");
```
korydraughn, line 82: This results in a copy.

Make `icat_server_env` a reference.

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -52,6 +53,51 @@ namespace
...
+
+            if ( icat_server_env.at("irods_zone_name").get_ref<const std::string&>() != local_zone_name ) {
+                continue;
+            }
+
+            for ( const auto& server : zone.at("servers") ) {
+                const auto server_admin = server.at("service_account_environment").at("irods_user_name").get<std::string>();
```
korydraughn, line 89: `server_admin` is never modified. This line also results in a copy.

- Make `server_admin` a reference
- Use `.get_ref<const std::string&>()`

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -52,6 +53,51 @@ namespace
...
+                if ( server_admin  == _user_name ) {
+                    const auto host_of_target_user = server.at("host_system_information").at("hostname").get<std::string>();
```
korydraughn, line 91: Same situation as `server_admin`.

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -52,6 +53,51 @@ namespace
...
+                    const auto msg = fmt::format( "ERROR: Cannot downgrade another rodsadmin ({}) running another server in this zone. On server: {}", _user_name, host_of_target_user );
```
korydraughn, line 92: Remove `ERROR: ` prefix.

The client most likely covers that.

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -52,6 +53,51 @@ namespace
...
+                    addRErrorMsg(&rsComm.rError, SYS_NOT_ALLOWED, msg.c_str());
+                    THROW(SYS_NOT_ALLOWED, msg);
+                }
+            }
+
```
korydraughn, line 97: Remove empty line.

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -801,6 +847,13 @@ _rsGeneralAdmin( rsComm_t *rsComm, generalAdminInp_t *generalAdminInp ) {
                 // it will be used multiple times. Note: subject to TOCTOU problem.
                 const auto current_user_type = irods::user::get_type(*rsComm, user_name);
 
+                throw_if_user_type_can_result_in_non_functional_irods_server(
+                    *rsComm,
+                    new_value,
+                    user_name,
+                    option
```
korydraughn, line 854: Incorrect order of arguments.

korydraughn (changes requested):
We need to run `git clang-format` on this change and add testing around this.

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -52,6 +53,50 @@ namespace
...
+
+        for ( const auto& zone : zones ) {
+            const auto& icat_server_env = zone.at("icat_server").at("service_account_environment");
+
+            if ( icat_server_env.at("irods_zone_name").get_ref<const std::string&>() != local_zone_name ) {
+                continue;
+            }
+
+            for ( const auto& server : zone.at("servers") ) {
+                const auto server_admin = server.at("service_account_environment").at("irods_user_name").get_ref<const std::string&>();
```
korydraughn, line 89: `const auto&`

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -52,6 +53,51 @@ namespace
...
+             return;
+         }
+
+        BytesBuf* bbuf = nullptr;
+        irods::at_scope_exit free_buf { [&bbuf] { std::free(bbuf); }  };
+
+        if (const auto ec = rsZoneReport(&rsComm, &bbuf); ec < 0 ) {
+            const auto msg = "The server has experienced an unexpected error while processing the request. Please "
+                             "report to the administrator.";
+            addRErrorMsg(&rsComm.rError, SYS_NOT_ALLOWED, msg);
```
alanking, line 72: `SYS_NOT_ALLOWED` -> `ec`

Should the message we return be more specific about what happened? After all, we are in `rsGeneralAdmin`, so the administrator is (hopefully) the one who is going to see the message anyway, right?

korydraughn: That is a very good point.

Perhaps the error message becomes ... `"Failed to gather rodsadmin users managing a server in the local zone."`

unit_tests/src/test_user_administration.cpp
```diff
@@ -8,6 +8,36 @@
 #include <algorithm>
 #include <iterator>
 
+TEST_CASE("rodsadmin downgrade")
```
korydraughn, line 11: Writing a C++ unit test is great, but you're going to run into some challenges for this particular case.

Your best option is to switch to the python test suite.

See scripts/irods/test/test_iadmin.py.

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -52,6 +53,55 @@ namespace
...
+
+                if ( icat_server_env.at("irods_zone_name").get_ref<const std::string&>() != local_zone_name ) {
+                    continue;
+                }
+
+                for ( const auto& server : zone.at("servers") ) {
+                    const auto& server_admin = server.at("service_account_environment").at("irods_user_name").get_ref<const std::string&>();
+                    if ( server_admin  == _user_name ) {
+                        const auto& host_of_target_user = server.at("host_system_information").at("hostname").get_ref<const std::string&>();
+                        const auto msg = fmt::format( "Cannot downgrade another rodsadmin ({}) running another server in this zone. On server: {}", _user_name, host_of_target_user );
```
trel, line 93: We've been using square brackets to hold values in error messages...

so...
"rodsadmin ({})" -> "rodsadmin [{}]"

and then... the later part could be...
"another server [{}] in this zone."

scripts/irods/test/test_iadmin.py
```diff
@@ -2140,6 +2144,14 @@ def test_moduser_type_invalid_type(self):
                                    'STDERR_SINGLELINE', 'CAT_INVALID_USER_TYPE')
         self.assertEqual('rodsuser', lib.get_user_type(self.admin, self.username))
 
+    def test_moduser_bad_downgrade(self):
+
+        remote_zone = 'somezone'
+        remote_zone_admin = 'zone_admin'
+
+        self.assertEqual('rodsadmin', lib.get_user_type(self.admin, 'rods'))
+        self.admin.assert_icommand(['admin', 'moduser', 'rods', 'type', 'roduser'])
```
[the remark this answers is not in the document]
alanking: Also, this test name is identical to the one on 2135.

scripts/irods/test/test_iadmin.py
```diff
@@ -2133,8 +2133,14 @@ def test_moduser_type_rodsgroup(self):
         self.assertEqual('rodsuser', lib.get_user_type(self.admin, self.username))
 
     def test_moduser_bad_downgrade(self):
+        SYS_NOT_ALLOWED = -169000
         self.assertEqual('rodsadmin', lib.get_user_type(self.admin, 'rods'))
-        self.admin.assert_icommand(['iadmin', 'moduser', 'rods', 'type', 'rodsuser'])
+        res, _ = lib.execute(['iadmin', 'moduser', 'rods', 'type', 'rodsuser'])
```
trel, line 2138: i think it can just be...

```
self.admin.assert_icommand(['iadmin', 'moduser', 'rods', 'type', 'rodsuser'], 'STDERR_SINGLELINE', 'SYS_NOT_ALLOWED')
```

?

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -52,6 +53,58 @@ namespace
...
+            }
+
+            const auto buf = static_cast<char*> (bbuf->buf);
+            const nlohmann::json zone_report = nlohmann::json::parse(buf, buf + bbuf->len);
+            const auto& zones = zone_report.at("zones");
+
+            const auto * local_zone_name = getLocalZoneName();
+
+            for ( const auto& zone : zones ) {
+                const auto icat_server_env = zone.at("icat_server").at("serivce_account_environment");
```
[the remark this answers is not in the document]
korydraughn: Has this been resolved?

scripts/irods/test/test_iadmin.py
```diff
@@ -2132,6 +2132,10 @@ def test_moduser_type_rodsgroup(self):
                                    'STDERR', 'SYS_NOT_ALLOWED')
         self.assertEqual('rodsuser', lib.get_user_type(self.admin, self.username))
 
+    def test_moduser_bad_downgrade(self):
```
korydraughn, line 2135: Let's use more words to describe the test. For example:
```python
def test_downgrade_of_service_account_irods_user_is_not_allowed__issue_NNNN(self):
```
Where `NNNN` should match the issue number.

scripts/irods/test/test_iadmin.py
```diff
@@ -2132,6 +2132,10 @@ def test_moduser_type_rodsgroup(self):
...
+        self.assertEqual('rodsadmin', lib.get_user_type(self.admin, 'rods'))
+        self.admin.assert_icommand(['iadmin', 'moduser', 'rods', 'type', 'rodsuser'], 'STDERR_SINGLELINE', 'SYS_NOT_ALLOWED')
```
korydraughn, line 2137: `SYS_NOT_ALLOWED` could appear in another string in the future which could lead to the test producing incorrect results.

Adjust the assertion so that `SYS_NOT_ALLOWED` can't lead to incorrect results.

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -52,6 +53,55 @@ namespace
...
+                        addRErrorMsg(&rsComm.rError, SYS_NOT_ALLOWED, msg.c_str());
+                        THROW(SYS_NOT_ALLOWED, msg);
+                    }
+                }
+            }
+        }
+        catch (const nlohmann::json::exception& e) {
+            THROW(SYS_LIBRARY_ERROR, e.what());
+        }
+     } // throw_if_user_type_can_result_in_non_functional_irods_server
```
korydraughn, line 103: Update comment.

korydraughn (changes requested):
We're close.

scripts/irods/test/test_iadmin.py
trel: oh, you mean... make it more specific?   so that it catches *just* this particular error?

korydraughn: Correct.

trel:
looking good - get that clang-format cleaned up.

scripts/irods/test/test_iadmin.py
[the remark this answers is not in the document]
korydraughn: @phdavis1027 Has this been resolved?

korydraughn:
You can run `git clang-format main` from the branch associated with this PR to format the things you touched. If anything is modified, you'll need to add those changes as another commit.

trel:
i confirm a green check on clang-format... and i only see one unresolved comment above.

scripts/irods/test/]
```diff
@@ -0,0 +1,2419 @@
+from __future__ import print_function
```
korydraughn:1]: A random file has appeared!

scripts/irods/test/test_iadmin.py
```diff
@@ -2132,6 +2132,11 @@ def test_moduser_type_rodsgroup(self):
                                    'STDERR', 'SYS_NOT_ALLOWED')
         self.assertEqual('rodsuser', lib.get_user_type(self.admin, self.username))
 
+    def test_downgrade_of_service_account_user_is_not_allowed__6127(self):
+        self.assertEqual('rodsadmin', lib.get_user_type(self.admin, 'rods'))
+        desired_error_msg = "Cannot downgrade another rodsadmin"
+        self.admin.assert_icommand(['iadmin', 'moduser', 'rods', 'type', 'rodsuser'], 'STDERR_SINGLELINE', desired_error_msg)
```
korydraughn, line 2138: Add another assertion that tests downgrade to `groupadmin` too.

That completes testing for all user types.

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -52,6 +53,62 @@ namespace
...
+            const auto msg =
+                fmt::format("[{}:{}] - Failed to gather rodsadmin users managing a server in the local zone.",
+                            __func__,
+                            __LINE__);
+            addRErrorMsg(&rsComm.rError, ec, msg.c_str());
+            THROW(ec, msg);
+        }
+
+        const auto buf = static_cast<char*>(bbuf->buf);
+        try {
```
korydraughn, line 78: Swap lines so that the variable is inside the try-block. `buf` is only used by the call to `nlohmann::json::parse`.

korydraughn (changes requested):
Once the changes have been made, I think we just run the core test suite and unit test suite to make sure things work.

korydraughn:
Has this been rebased on top of tip of main and tested?

phdavis1027:
> Has this been rebased on top of tip of main and tested?
At the moment, working through some likely-Docker-related problems with running the testing script.

phdavis1027:
> Has this been rebased on top of tip of main and tested?

test_iadmin.py has failed. The test report contains the following failed assertions:
```
 --- IrodsSession: icommand executed by [otherrods] [iadmin moduser rods type rodsuser] ---                             
Assert Command: iadmin moduser rods type rodsuser                                                                       
Expecting STDERR_SINGLELINE: ['Cannot downgrade another rodsadmin']                                                     
stdout:                                                                                                               
|                                                                                                                   
stderr:                                                                                                               
|                                                                                                                   
Output not found                                                                                                        
                                                                                                                      
FAILED TESTING ASSERTION

[...]

 --- IrodsSession: icommand executed by [moduser_user] [iinit] ---                                                      
Assert Command: iinit                                                                                                   
Expecting STDOUT_SINGLELINE: ['']                                                                                       
stdout:                                                                                                               
| WARNING: Error 25 disabling echo mode. Password will be displayed in plaintext.                                   
| Enter your current iRODS password:                                                                                
stderr:                                                                                                               
| Error occurred while authenticating user [moduser_user] [CAT_INVALID_USER: failed to perform request              
|                                                                                                                   
| ] [ec=-827000] failed with error -827000 CAT_INVALID_USER                                                         
Unexpected output on stderr                                                                                             
                                                                                                                         
FAILED TESTING ASSERTION 

[...]

  --- IrodsSession: icommand executed by [moduser_user] [iinit] ---                                                      
Assert Command: iinit                                                                                                   
Expecting STDOUT_SINGLELINE: ['']                                                                                       
stdout:                                                                                                               
| WARNING: Error 25 disabling echo mode. Password will be displayed in plaintext.                                   
| Enter your current iRODS password:                                                                                
stderr:                                                                                                               
| Error occurred while authenticating user [moduser_user] [CAT_INVALID_AUTHENTICATION: failed to perform request    
|                                                                                                                   
| ] [ec=-826000] failed with error -826000 CAT_INVALID_AUTHENTICATION                                               
Unexpected output on stderr                                                                                             
                                                                                                                         
FAILED TESTING ASSERTION

[...]

 --- IrodsSession: icommand executed by [otherrods] [irmtrash] ---                                                      
Assert Command: irmtrash                                                                                                
Expecting EMPTY: ['']                                                                                                   
stdout:                                                                                                               
|                                                                                                                   
stderr:                                                                                                               
|                                                                                                                   
Output found                                                                                                            
                                                                                                                        
--- IrodsSession: icommand executed by [rods] [iadmin rmuser otherrods] ---                                            
Assert Command: iadmin rmuser otherrods                                                                                 
Expecting EMPTY: ['']                                                                                                   
stdout:                                                                                                               
|                                                                                                                   
stderr:                                                                                                               
| remote addresses: 172.27.0.3 ERROR: rcGeneralAdmin failed with error -13000 SYS_NO_API_PRIV                       
Unexpected output on stderr                                                                                             
                                                                                                                       
FAILED TESTING ASSERTION 
```

Sanity checked by running same testing command against the main branch, which passed.

korydraughn:
What test in test_iadmin.py failed?

phdavis1027:
> What test in test_iadmin.py failed?

Based on these results, looks like test_moduser_type_rodsadmin_rodsuser and test_downgrade_of_service_account_user_is_not_allowed__6127

scripts/irods/test/test_iadmin.py
```diff
@@ -2132,6 +2132,14 @@ def test_moduser_type_rodsgroup(self):
...
+        self.assertEqual('rodsadmin', lib.get_user_type(self.admin, 'rods'))
+
+        self.admin.assert_icommand(['iadmin', 'moduser', 'rods', 'type', 'groupadmin', 'STDERR_SINGLELINE', desired_error_msg])
+        self.assertEqual('rodsadmin', lib.get_user_type(self.admin, 'rods'))
```
korydraughn, line 2142: These test cases overlap with a few assertions in `test_moduser_type_rodsadmin_rodsuser`.

Let's remove your new test and fix the failed assertions in `test_moduser_type_rodsadmin_rodsuser`.

korydraughn:
This work has been continued in PR #6774

PR #6467, opened 2022-07-09
[_6465] allow remote(...) hints to target INST_NAME

d-w-moore:

Issue #4164, opened 2018-10-26
Add method for reading remote logs in test infrastructure


alanking:
## Feature

Certain actions generate messages in log files on remote servers in our topology tests but there is no way to access those logs. Please investigate adding this. Many tests are being skipped if they must access a log file. In some cases, reading the remote log is required in order to have a valid test.

Currently, we can only read logs from the local machine on which the tests are being run. Here is an example usage of the current method for reading the server log:
```
initial_log_size = lib.get_file_size_by_path(paths.server_log_path())
# ... do some stuff ...
log_output_count = lib.count_occurrences_of_string_in_log(paths.server_log_path(), expected_value, start_index=initial_log_size)
self.assertTrue(expected_count == log_output_count)
```

PR #6468, opened 2022-07-11
[_6465] allow remote(...) hints to target INST_NAME

d-w-moore:
Replaces https://github.com/irods/irods/pull/6467

### DIFF
--- a/plugins/rule_engines/irods_rule_language/src/functions.cpp
+++ b/plugins/rule_engines/irods_rule_language/src/functions.cpp
@@ -1907,7 +1907,12 @@ Res *smsi_remoteExec( Node** paramsr, int, Node* node, ruleExecInfo_t* rei, int,
 
     try {
         auto taggedValues = getTaggedValues(params[1]->text);
-        auto it = taggedValues.find("ZONE");
+        auto it = taggedValues.find("INST_NAME");
+        if (it != taggedValues.end()) {
+            addKeyVal(&execMyRuleInp.condInput, INSTANCE_NAME_KW, it->second.front().c_str());
+            taggedValues.erase(it);
+        }
+        it = taggedValues.find("ZONE");
         if ( it != taggedValues.end() ) {
             strncpy(execMyRuleInp.addr.zoneName, it->second.front().c_str(), NAME_LEN);
             taggedValues.erase(it);
--- a/scripts/irods/test/test_rule_engine_plugin_framework.py
+++ b/scripts/irods/test/test_rule_engine_plugin_framework.py
@@ -15,6 +15,7 @@ import subprocess
 import mmap
 import shutil
 import re
+import textwrap
 
 from .. import lib
 from .. import paths
@@ -23,8 +24,11 @@ from . import settings
 from .resource_suite import ResourceBase
 from ..configuration import IrodsConfig
 from ..controller import IrodsController
-from ..core_file import temporary_core_file
+from ..core_file import (temporary_core_file,
+                         IRODS_RULE_LANGUAGE_RULE_ENGINE_PLUGIN_NAME,
+                         PYTHON_RULE_ENGINE_PLUGIN_NAME)
 from .rule_texts_for_tests import rule_texts
+from .test_python_rule_engine_plugin import append_native_re_to_server_config
 from ..test.command import assert_command
 from . import session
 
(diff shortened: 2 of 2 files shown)


trel:
can `getTaggedValues` throw?

d-w-moore:
Yes indeed, if the argument is not well-formed. thanks.

trel:
right, looks good to me - please touch up the clang-formatting and await alan/kory eyes.

korydraughn:
The logic looks good.

Are you planning on adding a test? That should be doable using the topology settings and metadata.

d-w-moore:
@korydraughn  I'm definitely interested in writing the test and was going to ask about how to do it. You mention the topology settings and metadata, is there a guide or example? Thanks!

d-w-moore:
@trel Not sure what's meant by the clang formatting

plugins/rule_engines/irods_rule_language/src/functions.cpp
```diff
@@ -1907,7 +1907,12 @@ Res *smsi_remoteExec( Node** paramsr, int, Node* node, ruleExecInfo_t* rei, int,
 
     try {
         auto taggedValues = getTaggedValues(params[1]->text);
-        auto it = taggedValues.find("ZONE");
+        auto it = taggedValues.find("INST_NAME");
+        if ( it != taggedValues.end() ) {
+            addKeyVal( &execMyRuleInp.condInput, "instance_name", it->second.front().c_str());
```
alanking, line 1912: "instance_name" -> `irods::KW_CFG_INSTANCE_NAME`

alanking (commented):
Overall looks good. I agree with clang-format and test sentiments. As @korydraughn suggested, if you use the `irods.test.settings.HOSTNAME_X` variables in your `remote()` call's target host(s), it can run in and outside of topology tests without having to selectively skip the tests and exercise your code.

plugins/rule_engines/irods_rule_language/src/functions.cpp
d-w-moore: noticed that, will use the symbol.

d-w-moore: hm, there are several keywords that would all seem to fit, but the simplest seems to be INSTANCE_NAME_KW just as it is here, in a function called from _delayExec: https://github.com/irods/irods/blob/fc0653187ec7a308a466529aa76917c723c06995/server/re/src/irods_re_structs.cpp#L519

alanking: Oh, I didn't know that existed, or that we had multiple of that keyword. I think what you used is better, personally. Let's go with `INSTANCE_NAME_KW`.

alanking:
One more clang-format thing and I think this will be good to go

d-w-moore:
> One more clang-format thing and I think this will be good to go

If i understand correctly, do i need to add some clang format directives ?

alanking:
You can configure `clang-format` with `git` by following the instructions in the GitHub Actions workflow: https://github.com/irods/irods/blob/fc0653187ec7a308a466529aa76917c723c06995/.github/workflows/linter-irods-clang-format.yml#L29-L35

*Or* you can apply the diff directly using `git apply` since it's pretty small:
```bash
$ echo '
>  index c1c7b2d8f..4b4c9bf77 100644
> --- a/plugins/rule_engines/irods_rule_language/src/functions.cpp
> +++ b/plugins/rule_engines/irods_rule_language/src/functions.cpp
> @@ -1908,8 +1908,8 @@ Res *smsi_remoteExec( Node** paramsr, int, Node* node, ruleExecInfo_t* rei, int,
>      try {
>          auto taggedValues = getTaggedValues(params[1]->text);
>          auto it = taggedValues.find("INST_NAME");
> -        if ( it != taggedValues.end() ) {
> -            addKeyVal( &execMyRuleInp.condInput, INSTANCE_NAME_KW, it->second.front().c_str());
> +        if (it != taggedValues.end()) {
> +            addKeyVal(&execMyRuleInp.condInput, INSTANCE_NAME_KW, it->second.front().c_str());
>              taggedValues.erase(it);
>          }
>          it = taggedValues.find("ZONE");
> ' > /tmp/applyme
$ git apply /tmp/applyme
```

**OR** you can just delete the two spaces that it wants you to delete. :p

alanking (approved):
Overall seems good to me. Shall we squash? Leave off the # in case other thoughts come in

trel:
let’s squash em, no #

d-w-moore:
Done!

korydraughn:
I didn't see any topology-based tests in the PR.

Did we conclude they aren't necessary?

alanking:
I think we determined that that's going to be hard because we would have to configure remote servers from inside the test and/or read from a remote log (see https://github.com/irods/irods/issues/4164).

trel:
Right, this was tested by hand as I understand.

alanking:
Got verbal approval, so let's go ahead and add #'s.

d-w-moore:
Will do. I assume that goes for the irods_docs (https://github.com/irods/irods_docs/pull/164) and Python RE  plugin (https://github.com/irods/irods_rule_engine_plugin_python/pull/128) PRs against `main` as well ?

alanking:
No, just this one. We can address the other PRs in turn.

alanking:
If you feel that they must all be merged together, then hold off on #'s until all of them are approved.

d-w-moore:
coolio. done....

Issue #6273, opened 2022-03-26
Update `get_script_output_single_line()` to use Boost.Process


korydraughn:
- [x] main
- [x] 4-3-stable
---
https://github.com/irods/irods/blob/a86b0ed40bcb2bd23cf526a52244286b78285468/server/core/src/miscServerFunct.cpp#L2974-L2978

alanking:
@epsilon-phase - Since we have some more time, go ahead and cherry-pick this to 4-3-stable. Thanks

PR #6474, opened 2022-07-15
Reimplement get_script_output_single_line with boost::process

epsilon-phase:
This replaces the old popen code with the much more concise and modern feeling boost::process. This addresses #6273

### DIFF
--- a/lib/core/include/irods/rodsErrorTable.h
+++ b/lib/core/include/irods/rodsErrorTable.h
@@ -239,6 +239,7 @@ NEW_ERROR(NOT_A_COLLECTION,                            -170000)
 NEW_ERROR(NOT_A_DATA_OBJECT,                           -171000)
 NEW_ERROR(JSON_VALIDATION_ERROR,                       -172000)
 NEW_ERROR(DATABASE_TYPE_NOT_SUPPORTED,                 -173000)
+NEW_ERROR(RESOLUTION_ERROR,                            -174000)
 
 /** @} */
 
--- a/server/core/src/miscServerFunct.cpp
+++ b/server/core/src/miscServerFunct.cpp
@@ -32,6 +32,12 @@
 #include <boost/thread/thread.hpp>
 #include <boost/thread/scoped_thread.hpp>
 #include <boost/lexical_cast.hpp>
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wc++11-narrowing"
+#include <boost/process.hpp>
+#pragma clang diagnostic pop
+
 #include <openssl/md5.h>
 
 
@@ -59,6 +65,7 @@ char *__loc1;
 #include <cstring>
 #include <iomanip>
 #include <fstream>
+#include <filesystem>
 #include <string>
 #include <vector>
 
@@ -2976,53 +2983,63 @@ irods::error get_script_output_single_line(
     const std::string&              script_name,
     const std::vector<std::string>& args,
     std::string&                    output ) {
-    output.clear();
-    std::stringstream exec;
+    namespace bp = boost::process;
+
     try {
-    exec << script_language
-         << " " << irods::get_irods_home_directory().string()
-         << "/scripts/" << script_name;
-    } catch (const irods::exception& e) {
-        irods::log(e);
-        return ERROR(-1, "failed to get irods home directory");
-    }
-    for ( std::vector<std::string>::size_type i = 0; i < args.size(); ++i ) {
-        exec << " " << args[i];
-    }
+        auto resolved_program = bp::search_path(script_language);
+        if (resolved_program.empty()) {
+            return ERROR(RESOLUTION_ERROR, fmt::format("Could not resolve {} to an executable", script_language));
+        }
 
-    FILE *fp = popen( exec.str().c_str(), "r" );
-    if ( fp == NULL ) {
-        return ERROR( SYS_FORK_ERROR, "popen() failed" );
+        auto script_path = fmt::format("{}/scripts/{}", irods::get_irods_home_directory().c_str(), script_name);
+        if (!std::filesystem::exists(script_path)) {
+            return ERROR(RESOLUTION_ERROR, fmt::format("Script file not found! {}", script_path));
+        }
+
+        auto script_path_and_args = args;
+        script_path_and_args.insert(script_path_and_args.begin(), script_path);
+
+        bp::ipstream program_output;
+        bp::ipstream program_stderr;
+        bp::child c(resolved_program, script_path_and_args, bp::std_out > program_output, bp::std_err > program_stderr);
+        output.clear();
+        c.wait();
+        const auto ec = c.exit_code();
+
+        std::string stderr_output;
+        if (ec != 0 || (program_stderr && std::getline(program_stderr, stderr_output) && !stderr_output.empty())) {
+            irods::experimental::log::server::error(
+                "Received stderr output [{}] from script process [{} {}] which exited with code {}",
+                stderr_output,
+                script_language,
+                fmt::join(script_path_and_args, " "),
+                ec);
+        }
+
+        if (!program_output) {
+            return ERROR(SYS_PIPE_ERROR,
+                         fmt::format("stdout was unable to be opened, process exited with code {}.", ec));
+        }
+
+        if (!std::getline(program_output, output)) {
+            return ERROR(SYS_PIPE_ERROR,
+                         fmt::format("Unable to read line from child process, exited with code {}.", ec));
+        }
+
+        return SUCCESS();
     }
-
-    std::vector<char> buf( 1000 );
-    const char* fgets_ret = fgets( &buf[0], buf.size(), fp );
-    if ( fgets_ret == NULL ) {
-        std::stringstream msg;
-        msg << "fgets() failed. feof["
-            << std::feof( fp )
-            << "] ferror["
-            << std::ferror( fp ) << "]";
-        const int pclose_ret = pclose( fp );
-        msg << " pclose[" << pclose_ret << "]";
-        return ERROR( FILE_READ_ERR, msg.str() );
+    catch (const bp::process_error& e) {
+        // In this case, the server is most likely unable to find a suitable
+        // executable to run with a given name
+        irods::experimental::log::server::error("Error starting process, {}", e.what());
+        return ERROR(SYS_LIBRARY_ERROR, e.what());
     }
-
-    const int pclose_ret = pclose( fp );
-    if ( pclose_ret == -1 ) {
-        return ERROR( SYS_FORK_ERROR,
-                      "pclose() failed." );
+    catch (const irods::exception& e) {
+        return ERROR(e.code(), e.what());
     }
-
-    output = &buf[0];
-    // Remove trailing newline
-    const std::string::size_type size = output.size();
-    if ( size > 0 && output[size - 1] == '\n' ) {
-        output.resize( size - 1 );
+    catch (const std::exception& e) {
+        return ERROR(SYS_INTERNAL_ERR, e.what());
     }
-
-    return SUCCESS();
-
 }  // get_script_output_single_line
 
 irods::error add_global_re_params_to_kvp_for_dynpep(


server/core/src/miscServerFunct.cpp
```diff
@@ -65,6 +71,7 @@ char *__loc1;
 #include <boost/filesystem.hpp>
 
 using leaf_bundle_t = irods::resource_manager::leaf_bundle_t;
+namespace bp = boost::process;
```
korydraughn, line 74: Given that Boost.Process is used in only one function, we can move this into the function.

server/core/src/miscServerFunct.cpp
```diff
@@ -2976,53 +2983,12 @@ irods::error get_script_output_single_line(
...
-    }
-
-    output = &buf[0];
-    // Remove trailing newline
-    const std::string::size_type size = output.size();
-    if ( size > 0 && output[size - 1] == '\n' ) {
-        output.resize( size - 1 );
-    }
-
+    std::getline(output_stream, output);
```
korydraughn, line 2989: The examples indicate we should check `output_stream` before using it, which makes sense.

korydraughn (changes requested):
That's a good looking PR.

We need to run this through the testing environment to make sure nothing breaks.

We also need to test this manually and verify if the log file is quiet.

If Boost.Process throws exceptions, we need to guard against those as well.

server/core/src/miscServerFunct.cpp
epsilon-phase: what do you mean by checking it?

korydraughn: We check the stream to make sure it isn't in a bad state.
```cpp
if (!output_stream) {
    // Stream is in a bad state.
    // Handle error.
}

std::getline(output_stream, output);
```

server/core/src/miscServerFunct.cpp
```diff
@@ -2983,12 +2982,24 @@ irods::error get_script_output_single_line(
...
-    bp::ipstream output_stream;
-    bp::child c(script_language, args, bp::std_out > output_stream);
-    output.clear();
-    std::getline(output_stream, output);
-    c.wait();
-    return SUCCESS();
+    namespace bp = boost::process;
+    try {
+        auto tmp = args;
+        tmp.insert(tmp.begin(), fmt::format("{}/scripts/{}", irods::get_irods_home_directory().string(), script_name));
```
korydraughn, line 2988: `.string()` -> `.c_str()`

server/core/src/miscServerFunct.cpp
```diff
@@ -2983,12 +2982,24 @@ irods::error get_script_output_single_line(
...
+        bp::ipstream output_stream;
+        bp::child c(script_language, tmp, bp::std_out > output_stream);
```
korydraughn, line 2990: We need to consider stderr and what that means to this function.

server/core/src/miscServerFunct.cpp
```diff
@@ -2983,12 +2982,24 @@ irods::error get_script_output_single_line(
...
+        output.clear();
+        if (!output_stream) {
+            return ERROR(FILE_READ_ERR, "Bad stream");
```
korydraughn, line 2993: The error message needs to include more information about what happened so that the admin can determine what to do.

`FILE_READ_ERR` doesn't express what happened very well. We should look for a different error code or create a new one.

server/core/src/miscServerFunct.cpp
```diff
@@ -2983,12 +2982,24 @@ irods::error get_script_output_single_line(
...
+        }
+        std::getline(output_stream, output);
+        c.wait();
+        return SUCCESS();
```
korydraughn, line 2997: We need to consider the exit code of the script and what that means to this function.
```cpp
const int ec = c.exit_code();
```

server/core/src/miscServerFunct.cpp
```diff
@@ -2983,12 +2982,24 @@ irods::error get_script_output_single_line(
...
-    bp::child c(script_language, args, bp::std_out > output_stream);
-    output.clear();
-    std::getline(output_stream, output);
-    c.wait();
-    return SUCCESS();
+    namespace bp = boost::process;
+    try {
+        auto tmp = args;
+        tmp.insert(tmp.begin(), fmt::format("{}/scripts/{}", irods::get_irods_home_directory().string(), script_name));
+        bp::ipstream output_stream;
```
korydraughn, line 2989: `output_stream` -> `input_stream`

server/core/src/miscServerFunct.cpp
```diff
@@ -2983,12 +2982,24 @@ irods::error get_script_output_single_line(
...
+        if (!output_stream) {
+            return ERROR(FILE_READ_ERR, "Bad stream");
+        }
+        std::getline(output_stream, output);
+        c.wait();
+        return SUCCESS();
+    }
+    catch (bp::process_error& e) {
+        irods::experimental::log::server::error("Error starting process, {}", e.what());
+        return ERROR(SYS_FORK_ERROR, e.what());
```
korydraughn, line 3001: I think we need to use a different error code here.

Do we know what `e.what()` says in this case?

korydraughn (changes requested):

server/core/src/miscServerFunct.cpp
korydraughn: You're probably right. Try passing it different inputs and see what `e.what()` contains.

server/core/src/miscServerFunct.cpp
epsilon-phase: Well, the old version of the function doesn't do anything with stderr, so it's new territory if we want to do that.

server/core/src/miscServerFunct.cpp
```diff
@@ -2976,53 +2982,28 @@ irods::error get_script_output_single_line(
...
-        msg << " pclose[" << pclose_ret << "]";
-        return ERROR( FILE_READ_ERR, msg.str() );
-    }
-
-    const int pclose_ret = pclose( fp );
-    if ( pclose_ret == -1 ) {
-        return ERROR( SYS_FORK_ERROR,
-                      "pclose() failed." );
+        auto tmp = args;
+        auto resolved_program = bp::search_path(script_language);
```
alanking, line 2988: Should we make sure that this returns something? Or is it okay that this is empty?

server/core/src/miscServerFunct.cpp
```diff
@@ -2976,53 +2982,28 @@ irods::error get_script_output_single_line(
...
-        const int pclose_ret = pclose( fp );
-        msg << " pclose[" << pclose_ret << "]";
-        return ERROR( FILE_READ_ERR, msg.str() );
-    }
-
-    const int pclose_ret = pclose( fp );
-    if ( pclose_ret == -1 ) {
-        return ERROR( SYS_FORK_ERROR,
-                      "pclose() failed." );
+        auto tmp = args;
```
alanking, line 2987: Maybe instead of `tmp` we could call this `script_path_and_args` or something to that effect?

alanking (commented):
Lookin good.

server/core/src/miscServerFunct.cpp
epsilon-phase: It is as it turns out mostly about not being able to find the program, so I'll handle it as such

server/core/src/miscServerFunct.cpp
epsilon-phase: Well, I'm not sure what it should do in the new function, but in the old implementation it was checked for an error closing the stream, but otherwise discarded.

server/core/src/miscServerFunct.cpp
korydraughn: True, but we want to make sure things run and operate as expected. In 99% of the cases, we expect stderr to be empty. However, if it's not, we need to check it (and log it).

As for whether stderr should result in this function returning an error, that is something that needs to be discussed as a group. That also means we need to know what scripts are being invoked via this function. Knowing that will help us decide on what to do when stderr contains information.

server/core/src/miscServerFunct.cpp
korydraughn: That doesn't sound like the right approach. A bad exit code indicates we need to check things out.

Again, this is something that requires discussion as a group.

server/core/src/miscServerFunct.cpp
korydraughn: What did the messages look like?

Post one of the outputs you received.

lib/core/include/irods/rodsErrorTable.h
```diff
@@ -100,6 +100,7 @@ NEW_ERROR(SYS_UNRECOGNIZED_REMOTE_FLAG,                -20000)
 NEW_ERROR(SYS_INVALID_SERVER_HOST,                     -21000)
 NEW_ERROR(SYS_SVR_TO_SVR_CONNECT_FAILED,               -22000)
 NEW_ERROR(SYS_BAD_FILE_DESCRIPTOR,                     -23000)
+NEW_ERROR(SYS_RESOLUTION_ERROR,                        -23001)
```
korydraughn, line 103: Append to the end of group.

Make the integer value follow the same pattern as its siblings.

server/core/src/miscServerFunct.cpp
```diff
@@ -2976,53 +2982,41 @@ irods::error get_script_output_single_line(
...
-        const int pclose_ret = pclose( fp );
-        msg << " pclose[" << pclose_ret << "]";
-        return ERROR( FILE_READ_ERR, msg.str() );
-    }
-
-    const int pclose_ret = pclose( fp );
-    if ( pclose_ret == -1 ) {
-        return ERROR( SYS_FORK_ERROR,
-                      "pclose() failed." );
+        auto script_path_and_args = args;
```
korydraughn, line 2987: `script_path_and_args` is not used until L2996.

Let's move it closer to where it's used.

server/core/src/miscServerFunct.cpp
```diff
@@ -2976,53 +2982,41 @@ irods::error get_script_output_single_line(
...
+        if (resolved_program.empty()) {
+            return ERROR(SYS_RESOLUTION_ERROR, fmt::format("Could not resolve {} to an executable", script_language));
+        }
+        auto script_path = fmt::format("{}/scripts/{}", irods::get_irods_home_directory().c_str(), script_name);
+        if (!std::filesystem::exists(script_path)) {
+            return ERROR(SYS_RESOLUTION_ERROR, fmt::format("Script file not found! {}", script_path));
+        }
+        script_path_and_args.insert(
+            script_path_and_args.begin(),
+            fmt::format("{}/scripts/{}", irods::get_irods_home_directory().c_str(), script_name));
```
korydraughn, line 2998: Duplicate operation on L2992

server/core/src/miscServerFunct.cpp
```diff
@@ -2976,53 +2982,41 @@ irods::error get_script_output_single_line(
...
-    output = &buf[0];
-    // Remove trailing newline
-    const std::string::size_type size = output.size();
-    if ( size > 0 && output[size - 1] == '\n' ) {
-        output.resize( size - 1 );
+    catch (bp::process_error& e) {
+        // In this case, the server is most likely unable to find a suitable
+        // executable to run with a given name
+        irods::experimental::log::server::error("Error starting process, {}", e.what());
+        return ERROR(SYS_RESOLUTION_ERROR, e.what());
```
korydraughn, line 3018: Boost.Process can throw any number of exceptions so we need to use a more general error code here.

For this particular case, `SYS_LIBRARY_ERROR` is likely the best choice. This let's us know that there was a problem using one of the irods-externals libraries.

server/core/src/miscServerFunct.cpp
```diff
@@ -2993,29 +2992,39 @@ irods::error get_script_output_single_line(
...
         }
         if (!std::getline(program_output, output)) {
-            return ERROR(ACTION_FAILED_ERR,
+            return ERROR(SYS_PIPE_ERROR,
                          fmt::format("Unable to read line from child process, exited with code {}", ec));
         }
+        std::string stderr_output;
+        if (program_stderr && std::getline(program_stderr, stderr_output) && !stderr_output.empty()) {
+            irods::experimental::log::server::error(
+                "Received stderr output '{}' from script process '{} {}' which exited with code",
```
korydraughn, line 3015: Missing fourth format placeholder.

Use square brackets around the placeholders. That is the normal iRODS convention. It helps with identifying whether the value was an empty string or not.

lib/core/include/irods/rodsErrorTable.h
```diff
@@ -239,6 +239,7 @@ NEW_ERROR(NOT_A_COLLECTION,                            -170000)
 NEW_ERROR(NOT_A_DATA_OBJECT,                           -171000)
 NEW_ERROR(JSON_VALIDATION_ERROR,                       -172000)
 NEW_ERROR(DATABASE_TYPE_NOT_SUPPORTED,                 -173000)
+NEW_ERROR(SYS_RESOLUTION_ERROR,                        -174000)
```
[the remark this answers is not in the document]
epsilon-phase: I am inclined to agree with you on that, RESOLUTION_ERROR seems sufficient

server/core/src/miscServerFunct.cpp
```diff
@@ -2976,53 +2983,50 @@ irods::error get_script_output_single_line(
...
+            return ERROR(RESOLUTION_ERROR, fmt::format("Could not resolve {} to an executable", script_language));
+        }
+        auto script_path = fmt::format("{}/scripts/{}", irods::get_irods_home_directory().c_str(), script_name);
+        if (!std::filesystem::exists(script_path)) {
+            return ERROR(RESOLUTION_ERROR, fmt::format("Script file not found! {}", script_path));
+        }
+        auto script_path_and_args = args;
+        script_path_and_args.insert(script_path_and_args.begin(), script_path);
+
+        bp::ipstream program_output, program_stderr;
```
korydraughn, line 2999: Let's put `program_stderr` on its own line.
```cpp
bp::ipstream program_output;
bp::ipstream program_stderr;
```
This improves readability.

server/core/src/miscServerFunct.cpp
```diff
@@ -2976,53 +2983,50 @@ irods::error get_script_output_single_line(
...
-    // Remove trailing newline
-    const std::string::size_type size = output.size();
-    if ( size > 0 && output[size - 1] == '\n' ) {
-        output.resize( size - 1 );
+    catch (bp::process_error& e) {
+        // In this case, the server is most likely unable to find a suitable
+        // executable to run with a given name
+        irods::experimental::log::server::error("Error starting process, {}", e.what());
+        return ERROR(SYS_LIBRARY_ERROR, e.what());
     }
```
korydraughn, line 3029: Let's add a catch-block for `irods::exception` and `std::exception` for completeness.

korydraughn (changes requested):
I think this last change request puts us in a good spot.

Please run your PR in the irods_testing_environment (core tests and unit tests) so that we know we didn't break anything.

If the tests pass, add a comment to your PR stating that the tests passed.

server/core/src/miscServerFunct.cpp
```diff
@@ -3027,6 +3028,12 @@ irods::error get_script_output_single_line(
         irods::experimental::log::server::error("Error starting process, {}", e.what());
         return ERROR(SYS_LIBRARY_ERROR, e.what());
     }
+    catch (irods::exception& e) {
+        return ERROR(SYS_INTERNAL_ERROR, e.what());
```
korydraughn: `SYS_INTERNAL_ERROR` -> `e.code()`

epsilon-phase:
It looks like all the tests are passing

korydraughn:
Also, did you see what the original implementation produced in the log file before your fixes?

It would cause an error in the log even though the operation succeeded. We want to verify that your change doesn't result in the same behavior. That is, the operation succeeds and the logs are quiet.

epsilon-phase:
Yeah, it only logs errors when stderr is written to or when something else went wrong with running the program. Maybe it should print a notice for stderr not being empty?

korydraughn:
I thought your implementation already logged the contents of stderr as an error if it were present. Does it not?

server/core/src/miscServerFunct.cpp
```diff
@@ -2976,53 +2983,57 @@ irods::error get_script_output_single_line(
...
+        }
+        auto script_path_and_args = args;
+        script_path_and_args.insert(script_path_and_args.begin(), script_path);
+
+        bp::ipstream program_output;
+        bp::ipstream program_stderr;
+        bp::child c(resolved_program, script_path_and_args, bp::std_out > program_output, bp::std_err > program_stderr);
+        output.clear();
+        c.wait();
+        const auto ec = c.exit_code();
```
korydraughn, line 3004: I think we need to log an error if `ec` is non-zero as well.

We should also check `program_stderr` before `program_output` so we capture what happened.

In both of these cases, we only need to log the error since that is what the original code did?

epsilon-phase:
The tests appear to have all passed.

epsilon-phase:
> 



> I thought your implementation already logged the contents of stderr as an error if it were present. Does it not?

It does, but I meant if the log level should be reduced to notice rather than error.

korydraughn (changes requested):
Once these changes are applied and squashed, add the # and we'll get this merged.

alanking:
I think the commit message is off by 10. Should be #6273?

epsilon-phase:
Yeah, I think it's done now

Issue #6492, opened 2022-07-22
Add github workflow for clang-tidy


korydraughn:
- [x] main
- [x] 4-3-stable

korydraughn:
Leaving open until the clang-tidy config is nailed down.

korydraughn:
Closing this as the work to add a github workflow for clang-tidy is complete.

If we need to tweak the clang-tidy config, we can open new issues for that.

PR #6494, opened 2022-07-22
[#6492] Added github workflow for clang-tidy (main)

korydraughn:


trel:
Seems good to me.

Let's leave the issue open until we see it in use a couple times.

korydraughn:
I think we can go ahead and merge this. This gives us a baseline to work with.

We can tweak the config as we gain more experience.

trel:
agreed

korydraughn:
#'d

Issue #6511, opened 2022-08-02
UseLibCXX: CMake version and LINK_LANGUAGE generator expression


SwooshyCueb:
- [x] main
- [x] 4-3-stable
- [x] 4-2-stable
---
This affects 4-2-stable, 4-3-stable, and main.

Related issue: #6500 

See the following lines:
https://github.com/irods/irods/blob/e99cf5579134ee776f6291509b7b03b0b771b5ef/cmake/Modules/UseLibCXX.cmake#L20-L26

This statement about `LINK_LANGUAGE` is incorrect. `LINK_LANGUAGE` wasn't added until 3.18. The version guards here need to be fixed.

alanking:
I added checkboxes but don't want to make any assumptions. Please checkbox and close if finished. Thanks!

PR #6512, opened 2022-08-02
[#6511] UseLibCXX: Fix LINK_LANGUAGE version guard (main)

SwooshyCueb:
Takes care of #6511 for main.
Will be cherry-picked into 4-3-stable if github actions builds succeed.


alanking (approved):
Nice

@korydraughn's and/or @trel ?

korydraughn (approved):
@SwooshyCueb Do you know if merging this PR will close the issue or not?

Given that this fix applies to multiple branches, if merging this PR causes the issue to be closed before the other PRs are merged, then we need to avoid using the auto-close features for issues and PRs.

Is there a way to make Github understand that multiple PRs need to be merged before closing the issue?

SwooshyCueb:
Yes, merging this PR will close the issue. I'm not aware of a way to link multiple PRs to a single issue. But given that I have PRs in for all three relevant branches, I think it's a non-issue in this particular case, provided all three are merged together.

korydraughn:
Ok. Let's remove the use of "Fixes" from the PRs so none of them close the issue. That let's gives us a chance to check the boxes and have the closing of the issue be the final step.

At this point, removal of "Fixes" from the comments may be pointless, but we'll at least see out what github does when attempting to unlink an issue.

trel:
Yeah, we'd just keep them tight in time... or just re-open.    Or... of course, don't use the magic words that GitHub knows...

SwooshyCueb:
PR unlinked from the issue.

Found some more information on PR-issue linking: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

Unless I missed it, it does not say what the behavior is when multiple PRs are linked to a single issue.
Regardless, I do not have permission to manually link or un-link issues in the way described in this article.

alanking:
It appears that none of the PRs is linked to the issue now, thanks to changing the wording in the original message. I'll merge shortly barring any objections

Issue #3397, opened 2016-11-15
Set RUNPATH instead of RPATH


kellerb:
- [ ] main
- [ ] 4-3-stable
---
https://cmake.org/pipermail/cmake/2011-August/045927.html
Specifically, `SET(CMAKE_EXE_LINKER_FLAGS "-Wl,--enable-new-dtags")`

Also, use `set(CMAKE_BUILD_WITH_INSTALL_RPATH TRUE)`

SwooshyCueb:
We're already passing `--enable-new-dtags` to the linker.

I'm going to be looking into whether or not it makes sense for us to define `CMAKE_BUILD_WITH_INSTALL_RPATH` when I start working on #6052.

alanking:
Let me know if we consider this task done and I can check the boxes. It seems that the two parts of the OP are addressed now.

SwooshyCueb:
Yeah it's done, check away

PR #6515, opened 2022-08-03
[#3397] Introduce IrodsRunpathDefaults CMake module (main)

SwooshyCueb:
Addresses #3397 for main for this repository.

This module handles setting overridable defaults for some `RPATH`-/`RUNPATH`-related variables.

We will not provide a default for `CMAKE_BUILD_WITH_INSTALL_RPATH`.
This value can instead be defined on the command line if needed.

Will be cherry-picked into 4-3-stable.


korydraughn:
When would someone want to set `CMAKE_BUILD_WITH_INSTALL_RPATH`?

If there are specific reasons for doing this, we should find a place to document that. I guess the cmake files make the most sense, my only concern is how do people learn about the option.

alanking:
As I understand it, `CMAKE_BUILD_WITH_INSTALL_RPATH` is useful to avoid re-linking binaries when installing if you know that the linked dependencies will be identical between the two. It instructs the linker to use the install `RPATH` at build time: https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_WITH_INSTALL_RPATH.html Does that seem right?

I think I agree that if it's worth mentioning explicitly in the commit message (and the original issue) that a comment or three in the CMake module may be worth writing to explain why we don't provide a default. In the original issue, it simply says:

> Also, use `set(CMAKE_BUILD_WITH_INSTALL_RPATH TRUE)` 

So, more context would be nice.

Otherwise, this seems good

korydraughn:
We also need to provide information on how to compile iRODS for people who land at the repo. That can be provided on the top-level README or via a separate markdown file.

That effort should be handled by a separate PR though.

trel:
Yes, let's keep the main README.md pretty clean...

I think a front page link that points to a different BUILDING.md file seems best?   I've seen that convention in other projects.

korydraughn:
Sure. I'll write up a new issue for that.

korydraughn:
Issue created: #6519

korydraughn (approved):

korydraughn:
What does the new CMake module mean for plugins, clients, etc?

SwooshyCueb:
> What does the new CMake module mean for plugins, clients, etc?

Not much. When I next do a consistency sweep, they will be changed to inherit/override the values from this file rather than manually defining the variables themselves.

SwooshyCueb:
\#'d

alanking (approved):

Issue #6387, opened 2022-05-12
Remote psql server connection over SSL


scimerman:
Hi, I would like to share my workaround/ask for a feedback/have a feature request.

Recently I had to connect our irods server (v4.2.11) to remote PostgreSQL server over SSL. The default irods options were not working, so I ended up with fixing the issue manually. I add a postgresql variables into `irods/scripts/irods/database_connect.py`, inside `return` of the `get_odbc_entry` function:
```
'SSL': 'on',                                      
'SSLMode': 'require',                             
'SSLRootCert': '/etc/irods/usertrust_ca.crt',
'SSLKey': '/etc/irods/localhost.key',      
'SSLCert': '/etc/irods/localhost.crt'
```
where `SSLRootCert` is CA certificate of remote psql server, `SSLKey` is icat server's SSL key file, and SSLCert is it's corresponding certificate.  The values were hardcoded, as I don't know how to reuse them from already defined values of the irods.

I also added to the `get_connection_string` of the same file, the corresponding lines:
```
odbc_dict['sslkey'] = '/etc/irods/localhost.key'   
odbc_dict['sslcert'] = '/etc/irods/localhost.crt'
odbc_dict['sslrootcert'] = '/etc/irods/usertrust_ca.crt'  
odbc_dict['sslmode'] = 'require'                          
odbc_dict['ssl'] = 'true'
```

After the restart of irods service, the client connects to the remote server and it updates the `~/.odbc.ini` so that includes the lines:

```
$ cat .odbc.ini
[iRODS Catalog]
...
SSLCert=/etc/irods/localhost.crt
Servername=my.remote.psql.server.com
SSLMode=require
SSLRootCert=/etc/irods/usertrust_ca.crt
SSLKey=/etc/irods/localhost.key
...
```

I understand that there is little that the irods community can do for the creation of the local `~/.postgresql` and associated files (see below). But I would love that the postgresql would (like it does for the cockroachdb) automatically configure everything that was already provided. It should use the existing certificates (those that irods is already using) and automatically configure them in the `~/.odbc.ini` so that the connection to the remote postgres can be established.

### Appendix:

Beside this, the `irods` user needs to have in his home a folder named `~/.postgresql` with (according to the https://www.postgresql.org/docs/current/libpq-ssl.html) following files
- `~/.postgresql/postgresql.key` the clients key
- `~/.postgresql/postgresql.crt` the clients certificate
- `~/.postgresql/root.crt` the CA certificate of the remote psql server
[...]

PR #6522, opened 2022-08-08
Postgresql: added SSL communication

scimerman:
Hi everyone,
this is my first contribution, so try to :see_no_evil:

My team is using the postgresql service from external provider, so I have created a few lines within the `database_connect.py` in order that the database communication was using TLS. That was few months ago for the v4.2.11 and I reported this in the issue #6387. Back then I have simply hard-linked our configuration values into the python script and I was happy about it.

Since then we have updated our iCAT to v4.3.0. And there has been some updates in the `database_connect.py` script, so I had to do it again. This time I used irods calls in the script, so that the `database_connect.py` should work for everyone. I would be happy is this can be made part of the irods, as communication without TLS is asking for troubles :fearful:

There are two things that should be mentioned:

**First**, it should be written in the documentation that the linux user `irods` (one that runs the irods service) needs to have inside home a subfolder named `~/.postgresql` with the (according to https://www.postgresql.org/docs/current/libpq-ssl.html) following 4 files:
- `~/.postgresql/postgresql.key` the clients (the iCAT machine) key
- `~/.postgresql/postgresql.crt` the clients (iCAT) certificate 
- `~/.postgresql/root.crt` the CA certificate of the remote psql server
- `~/.postgresql/root.crl` corresponding revocation list (I just created an empty file)
While I understand that this is not exactly *irods's issue*, I think that (based on the number of postgresql users) it should be mentioned.

**Secondly**, I have tested the script and it works, but my testing environment is limited.

Regards,
Sandi

### DIFF
--- a/scripts/irods/database_connect.py
+++ b/scripts/irods/database_connect.py
@@ -60,7 +60,31 @@ def dump_odbc_ini(odbc_dict, f):
         print('', file=f)
 
 def get_odbc_entry(db_config, catalog_database_type):
-    if catalog_database_type == 'postgres' or catalog_database_type == 'cockroachdb':
+    if catalog_database_type == 'postgres' and 'db_sslmode' in db_config and db_config['db_sslmode'] == 'require':
+        return {
+            'Description': 'iRODS Catalog',
+            'Driver': db_config['db_odbc_driver'],
+            'Trace': 'No',
+            'Debug': '0',
+            'CommLog': '0',
+            'TraceFile': '',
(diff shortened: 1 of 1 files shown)


korydraughn:
Thanks for your contribution. We'll review this and let you know what needs to happen next.

korydraughn:
Closing in favor of more flexible solution. See  https://github.com/irods/irods/issues/6387#issuecomment-1217199715 for details.

PR #6525, opened 2022-08-11
[4421] Added detached mode for unixfilesystem resource (main)

JustinKyleJames:
Adding detached mode for unixfilesystem resources.

Notes:

1. Right now, all servers in the zone would have to have a mount at the same location, as the location of the vault path is not server dependent.
2. If we wanted to limit the servers which can serve the request, we could add some type of list in the context string meaning only the servers in the list can serve it.
3. If we really wanted to make it configurable, the list from 2 could have two parts - the server name and the vault path.

For 3 we could do something like:
 ```"HOST_MODE=detached;SERVER_LIST=server1.irods.org:/lustre_mount/resource1,server2.irods.org:/mounts/lustre_mount/resource1"```

in the context string.

I haven't researched this fully to determine if it is possible or could be done efficiently.  Just throwing it out there as an idea.

### DIFF
--- a/plugins/resources/unixfilesystem/CMakeLists.txt
+++ b/plugins/resources/unixfilesystem/CMakeLists.txt
@@ -1,0 +1,76 @@
+cmake_minimum_required(VERSION 3.12.0 FATAL_ERROR)
+# target_link_libraries for object targets
+
+# anything being unit tested needs to be split out into an object library.
+add_library(
+  irods_resource_plugin_obj-unixfilesystem
+  OBJECT
+  "${CMAKE_CURRENT_SOURCE_DIR}/src/irods_is_in_host_list.cpp"
+)
+target_link_libraries(
+  irods_resource_plugin_obj-unixfilesystem
+  PUBLIC
+  irods_common
+)
+target_include_directories(
+  irods_resource_plugin_obj-unixfilesystem
+  PUBLIC
+  "${IRODS_EXTERNALS_FULLPATH_BOOST}/include"
+  "${IRODS_EXTERNALS_FULLPATH_FMT}/include"
+  "${IRODS_EXTERNALS_FULLPATH_SPDLOG}/include"
+  "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
+)
+target_compile_definitions(
+  irods_resource_plugin_obj-unixfilesystem
+  PRIVATE
+  ${IRODS_COMPILE_DEFINITIONS_PRIVATE}
+  IRODS_ENABLE_SYSLOG
+)
+
+add_library(
+  irods_resource_plugin-unixfilesystem
+  MODULE
+  "${CMAKE_CURRENT_SOURCE_DIR}/src/unixfilesystem.cpp"
+)
+set_property(TARGET irods_resource_plugin-unixfilesystem PROPERTY LIBRARY_OUTPUT_NAME "unixfilesystem")
+target_link_objects(
+  irods_resource_plugin-unixfilesystem
+  PRIVATE
+  irods_resource_plugin_obj-unixfilesystem
+)
+target_link_libraries(
+  irods_resource_plugin-unixfilesystem
+  PRIVATE
+  irods_common
+  irods_server
+  "${IRODS_EXTERNALS_FULLPATH_ARCHIVE}/lib/libarchive.so"
+  "${IRODS_EXTERNALS_FULLPATH_BOOST}/lib/libboost_filesystem.so"
+  "${IRODS_EXTERNALS_FULLPATH_BOOST}/lib/libboost_system.so"
+  "${IRODS_EXTERNALS_FULLPATH_FMT}/lib/libfmt.so"
+  OpenSSL::Crypto
+)
+target_include_directories(
+  irods_resource_plugin-unixfilesystem
+  PRIVATE
+  "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
+  "${IRODS_EXTERNALS_FULLPATH_ARCHIVE}/include"
+  "${IRODS_EXTERNALS_FULLPATH_BOOST}/include"
+  "${IRODS_EXTERNALS_FULLPATH_FMT}/include"
+  "${IRODS_EXTERNALS_FULLPATH_SPDLOG}/include"
+)
+target_compile_definitions(
+  irods_resource_plugin-unixfilesystem
+  PRIVATE
+  ${IRODS_COMPILE_DEFINITIONS_PRIVATE}
+  ENABLE_RE
+  IRODS_ENABLE_SYSLOG
+)
+
+add_dependencies(all-plugins-resources irods_resource_plugin-unixfilesystem)
+install(
+  TARGETS
+  irods_resource_plugin-unixfilesystem
+  LIBRARY
+  DESTINATION "${IRODS_PLUGINS_DIRECTORY}/resources"
+  COMPONENT ${IRODS_PACKAGE_COMPONENT_SERVER_NAME}
+)
--- a/plugins/resources/unixfilesystem/include/irods/private/irods_is_in_host_list.hpp
+++ b/plugins/resources/unixfilesystem/include/irods/private/irods_is_in_host_list.hpp
@@ -1,0 +1,13 @@
+#ifndef IRODS_IS_IN_HOST_LIST_HPP
+#define IRODS_IS_IN_HOST_LIST_HPP
+
+#include "irods/irods_lookup_table.hpp"
+
+#include <string_view>
+
+namespace irods
+{
+    auto is_host_in_host_list(irods::plugin_property_map& prop_map, const std::string_view& resource_hostname) -> bool;
+} // namespace irods
+
+#endif // IRODS_IS_IN_HOST_LIST_HPP
--- a/plugins/resources/unixfilesystem/src/irods_is_in_host_list.cpp
+++ b/plugins/resources/unixfilesystem/src/irods_is_in_host_list.cpp
@@ -1,0 +1,45 @@
+#include "irods/private/irods_is_in_host_list.hpp"
+
+#include "irods/irods_error.hpp"
+#include "irods/irods_string_tokenize.hpp"
+#include "irods/filesystem/path.hpp"
+
+// boost includes
+#include <boost/algorithm/string/predicate.hpp>
+
+// stl includes
+#include <vector>
+#include <string_view>
+
+namespace irods
+{
+    const std::string HOST_LIST("host_list");
+
+    // returns true if there is no host_list or if the current host is in
+    // the list
+    auto is_host_in_host_list(irods::plugin_property_map& prop_map, const std::string_view& resource_hostname) -> bool
+    {
+        std::string host_list_str;
+
+        if (!prop_map.get<std::string>(HOST_LIST, host_list_str).ok()) {
+            // no HOST_LIST parameter, all hosts assumed to be able to handle request
+            // and original vault path used for all hosts
+            return true;
+        }
+
+        // There is a host_list=x,y,z in the resource's context string.
+        // Check to see if the resource_hostname is in the comma delimited list.
+
+        // split parameter by delimiter (,)
+        const std::string delimiter = ",";
+        std::vector<std::string> tokens;
+        irods::string_tokenize(host_list_str, delimiter, tokens);
+
+        for (const std::string& token : tokens) {
+            if (boost::iequals(token, resource_hostname)) {
+                return true;
+            }
+        }
+        return false;
+    } // is_host_in_host_list
+} // namespace irods
--- a/plugins/resources/unixfilesystem/src/unixfilesystem.cpp
+++ b/plugins/resources/unixfilesystem/src/unixfilesystem.cpp
@@ -7,6 +7,7 @@
 
 // =-=-=-=-=-=-=-
 #include "irods/irods_resource_plugin.hpp"
+#include "irods/irods_resource_backport.hpp"
 #include "irods/irods_file_object.hpp"
 #include "irods/irods_physical_object.hpp"
 #include "irods/irods_collection_object.hpp"
@@ -19,6 +20,8 @@
 #include "irods/irods_kvp_string_parser.hpp"
 #include "irods/irods_logger.hpp"
 #include "irods/voting.hpp"
+#include "irods/private/irods_is_in_host_list.hpp"
+#include "irods/filesystem/path.hpp"
 
 // =-=-=-=-=-=-=-
 // stl includes
@@ -26,6 +29,7 @@
 #include <sstream>
 #include <vector>
 #include <string>
+#include <optional>
 
 // =-=-=-=-=-=-=-
 // boost includes
@@ -83,7 +87,90 @@
 const std::string DEFAULT_VAULT_DIR_MODE( "default_vault_directory_mode_kw" );
 const std::string HIGH_WATER_MARK( "high_water_mark" ); // no longer used
 const std::string REQUIRED_FREE_INODES_FOR_CREATE("required_free_inodes_for_create"); // no longer used
+const std::string HOST_MODE("host_mode");
 
+namespace
+{
+    auto is_operating_in_detached_mode(irods::plugin_property_map& prop_map) -> bool
+    {
+        std::string host_mode_str;
+
+        irods::error ret = prop_map.get<std::string>(HOST_MODE, host_mode_str);
+        if (ret.ok()) {
+            return boost::iequals(host_mode_str.c_str(), "detached");
+        }
+
+        // default is attached mode
+        return false;
+    }
+} // namespace
+
+// The return value is always SUCCESS() because the errors encountered are to
+// be treated as warnings and no fatal errors.
+auto unix_file_start_operation(irods::plugin_property_map& prop_map) -> irods::error
+{
+    namespace logger = irods::experimental::log;
+
+    if (!is_operating_in_detached_mode(prop_map)) {
+        // Detached mode is disabled.  Return early.
+        return SUCCESS();
+    }
+
+    // check if we are in the host list
+    char local_hostname[MAX_NAME_LEN];
+    gethostname(local_hostname, MAX_NAME_LEN);
+
+    auto resource_hostname_optional = resolve_hostname(local_hostname, hostname_resolution_scheme::match_preferred);
+
+    std::string resource_hostname =
+        resource_hostname_optional ? std::move(*resource_hostname_optional) : local_hostname;
+
+    if (!is_host_in_host_list(prop_map, resource_hostname)) {
+        // Host is not in host list.  Return early as detached mode is disabled on this resource.
+        return SUCCESS();
+    }
+
+    // Set the RESOURCE_HOST of this resource to the rodsHostAddr_t for this server.
+    // This server will now act as the owner of this resource for any processing on
+    // this server.
+
+    // Note:  On error conditions below, detached mode will remain disabled and a warning
+    // will be logged.
+
+    std::string resource_name;
+    irods::error ret = prop_map.get<std::string>(irods::RESOURCE_NAME, resource_name);
+    if (!ret.ok()) {
+        // Logging a warning as returning an error here simply logs the error
+        // and continues.
+        logger::resource::warn(
+            "Detached mode for unixfilesystem resource failed to set RESOURCE_HOST to {}."
+            "  Failed to get irods::RESOURCE_NAME property.",
+            local_hostname);
+        return SUCCESS();
+    }
+
+    // Look up the rodsHostAddr_t for this server and set this as the RESOURCE_HOST of the resource.
+    // Note that the rodsHostAddr_t entry is shared throughout the code (not just used for this
+    // resource) so you can't adjust the existing entry attached to the resource.
+    rodsHostAddr_t addr{};
+    std::strncpy(addr.hostAddr, local_hostname, LONG_NAME_LEN);
+    const auto& zone_name = irods::get_server_property<const std::string>(irods::KW_CFG_ZONE_NAME);
+    std::strncpy(addr.zoneName, zone_name.c_str(), NAME_LEN);
+
+    rodsServerHost_t* local_host = nullptr;
+    if (resolveHost(&addr, &local_host) < 0) {
+        logger::resource::warn(
+            "[resource_name={}] Detached mode failed to set RESOURCE_HOST to {}."
+            "  Failed to call resolveHost.",
+            resource_name,
+            local_hostname);
+        return SUCCESS();
+    }
+
+    irods::set_resource_property<rodsServerHost_t*>(resource_name, irods::RESOURCE_HOST, local_host);
+
+    return SUCCESS();
+} // unix_file_start_operation
 // =-=-=-=-=-=-=-
 // NOTE: All storage resources must do this on the physical path stored in the file object and then update
 //       the file object's physical path with the full path
@@ -280,7 +367,32 @@ irods::error unix_file_mkdir_r(const std::string& path, mode_t mode)
 /// @brief interface to notify of a file registration
 irods::error unix_file_registered(irods::plugin_context& _ctx)
 {
-    // NOOP
+    if (!is_operating_in_detached_mode(_ctx.prop_map())) {
+        return SUCCESS();
+    }
+
+    irods::file_object_ptr fco = boost::dynamic_pointer_cast<irods::file_object>(_ctx.fco());
+    std::string file_path = fco->physical_path();
+
+    std::string vault_path;
+    _ctx.prop_map().get<std::string>(irods::RESOURCE_PATH, vault_path);
+
+    if (!has_prefix(file_path.c_str(), vault_path.c_str())) {
+        // registered path not in vault, give warning
+        std::string resource_name;
+        _ctx.prop_map().get<std::string>(irods::RESOURCE_NAME, resource_name);
+
+        auto warning_message = fmt::format(
+            "Warning:  The registration was performed on a "
+            "unixfilesystem resource [{}] currently configured as 'host_mode=detached'.  "
+            "Verify that the path [{}] is available on all servers that may serve "
+            "requests for this resource.",
+            resource_name.c_str(),
+            file_path.c_str());
+
+        addRErrorMsg(&_ctx.comm()->rError, 0, warning_message.c_str());
+    }
+
     return SUCCESS();
 } // unix_file_registered
 
@@ -1174,6 +1286,8 @@ class unixfilesystem_resource : public irods::resource {
                         itr->first,
                         itr->second );
                 } // for itr
+
+            set_start_operation(unix_file_start_operation);
 
         } // ctor
 
--- a/scripts/irods/paths.py
+++ b/scripts/irods/paths.py
@@ -535,3 +535,16 @@ def possible_shm_locations():
     if _possible_shm_locations_str_cache is None:
         _possible_shm_locations_str_cache = frozenset(str(shm_dir) for shm_dir in _possible_shm_locations())
     return _possible_shm_locations_str_cache
+
+# This is a common mount point used by topology tests when testing detached mode
+# behavior for unixfilesystem resources.  In detached mode the servers are expected
+# to have a common mount where the vault exists.
+_test_mount_directory_cache = None
+def _test_mount_directory():
+    global _test_mount_directory_cache
+    if _test_mount_directory_cache is None:
+        _test_mount_directory_cache = pathlib.Path(os.sep, 'irods_testing_environment_mount_dir')
+    return _test_mount_directory_cache
+
+def test_mount_directory():
+    return str(_test_mount_directory())
--- a/unit_tests/CMakeLists.txt
+++ b/unit_tests/CMakeLists.txt
@@ -76,6 +76,7 @@ set(
   json_apis_from_client
   zone_report
   ticket_administration
+  host_list_context_string
 )
 
 foreach(test IN LISTS IRODS_UNIT_TESTS)
--- a/unit_tests/src/test_host_list_context_string.cpp
+++ b/unit_tests/src/test_host_list_context_string.cpp
@@ -1,0 +1,74 @@
+#include <catch2/catch.hpp>
+
+#include <string_view>
+#include <sstream>
+#include <vector>
+
+#include "irods/irods_plugin_base.hpp"
+#include "irods/filesystem/path.hpp"
+#include "irods/private/irods_is_in_host_list.hpp"
+
+TEST_CASE("host list tests", "[detached_mode_host_list]")
+{
+    const std::string_view resource_hostname = "myhost";
+    irods::plugin_property_map prop_map;
+
+    SECTION("not in host list")
+    {
+        prop_map.set<std::string>("host_list", "host2,host3");
+        REQUIRE_FALSE(is_host_in_host_list(prop_map, resource_hostname));
+    }
+
+    SECTION("not in host list but hostname with common prefix")
+    {
+        prop_map.set<std::string>("host_list", "myhost1,myhost2,host2,host3,myhost3");
+        REQUIRE_FALSE(is_host_in_host_list(prop_map, resource_hostname));
+    }
+
+    SECTION("at beginning of host list")
+    {
+        prop_map.set<std::string>("host_list", "myhost,host2,host3");
+        REQUIRE(is_host_in_host_list(prop_map, resource_hostname));
+    }
+
+    SECTION("at end of host list no path")
+    {
+        prop_map.set<std::string>("host_list", "host2,host3,myhost");
+        REQUIRE(is_host_in_host_list(prop_map, resource_hostname));
+    }
+
+    SECTION("in middle of host list no path")
+    {
+        prop_map.set<std::string>("host_list", "host2,myhost,host3");
+        REQUIRE(is_host_in_host_list(prop_map, resource_hostname));
+    }
+
+    SECTION("at beginning of host list not case match")
+    {
+        prop_map.set<std::string>("host_list", "MYHOST,host2,host3");
+        REQUIRE(is_host_in_host_list(prop_map, resource_hostname));
+    }
+
+    SECTION("at end of host list not case match")
+    {
+        prop_map.set<std::string>("host_list", "host2,host3,MYHOST");
+        REQUIRE(is_host_in_host_list(prop_map, resource_hostname));
+    }
+
+    SECTION("in middle of host list not case match")
+    {
+        prop_map.set<std::string>("host_list", "host2,MYHOST,host3");
+        REQUIRE(is_host_in_host_list(prop_map, resource_hostname));
+    }
+
+    SECTION("empty host list")
+    {
+        prop_map.set<std::string>("host_list", "");
+        REQUIRE_FALSE(is_host_in_host_list(prop_map, resource_hostname));
+    }
+
+    SECTION("no host list - all hosts in list")
+    {
+        REQUIRE(is_host_in_host_list(prop_map, resource_hostname));
+    }
+}
--- a/plugins/resources/CMakeLists.txt
+++ b/plugins/resources/CMakeLists.txt
@@ -6,6 +6,7 @@ add_custom_target(all-plugins-resources)
 add_dependencies(all-plugins_no_database all-plugins-resources)
 
 add_subdirectory(replication)
+add_subdirectory(unixfilesystem)
 
 set(
   IRODS_RESOURCE_PLUGINS
--- a/scripts/irods/test/test_resource_types.py
+++ b/scripts/irods/test/test_resource_types.py
@@ -600,6 +600,496 @@ OUTPUT ruleExecOut
         # local cleanup
         os.remove(filepath)
 
+    ###################
+    # detached mode
+    ###################
+    def test_detached_mode__issue_4421(self):
+        try:
+            file1 = "f1"
+            file2 = "f2"
+            resource_host = "irods.org"
+            resource_name = 'detached_resource'
+            resource_context = 'host_mode=detached'
+
+            lib.create_ufs_resource(resource_name, self.admin, resource_host)
+            self.admin.assert_icommand("iadmin modresc %s context %s" %
+                                   (resource_name, resource_context))
+
+            # create file to put
+            lib.make_file(file1, 100)
+
+            # put small file
+            self.admin.assert_icommand("iput -R %s %s" % (resource_name, file1))  # iput
+
+            # get file
+            self.admin.assert_icommand("iget %s %s" % (file1, file2))  # iget
+
+            # make sure the file that was put and got are the same
+            self.admin.assert_icommand("diff %s %s " % (file1, file2))
+        finally:
+            # local cleanup
+            self.admin.assert_icommand("irm -f " + file1)
+
+            if os.path.exists(file1):
+                os.unlink(file1)
+            if os.path.exists(file2):
+                os.unlink(file2)
+
+            # cleanup
+            lib.remove_resource(resource_name, self.admin)
+
+    def test_detached_mode_in_host_list__issue_4421(self):
+        try:
+            file1 = "f1"
+            file2 = "f2"
+            resource_host = "irods.org"
+            resource_name = 'detached_resource'
+            hostname = lib.get_hostname()
+            resource_context = 'host_mode=detached;host_list=irods.org,%s' % hostname
+
+            lib.create_ufs_resource(resource_name, self.admin, resource_host)
+            self.admin.assert_icommand("iadmin modresc %s context %s" %
+                                   (resource_name, resource_context))
+
+            # create file to put
+            lib.make_file(file1, 100)
+
+            # put small file
+            self.admin.assert_icommand("iput -R %s %s" % (resource_name, file1))  # iput
+
+            # get file
+            self.admin.assert_icommand("iget %s %s" % (file1, file2))  # iget
+
+            # make sure the file that was put and got are the same
+            self.admin.assert_icommand("diff %s %s " % (file1, file2))
+        finally:
+            # local cleanup
+            self.admin.assert_icommand("irm -f " + file1)
+
+            if os.path.exists(file1):
+                os.unlink(file1)
+            if os.path.exists(file2):
+                os.unlink(file2)
+
+            # cleanup
+            lib.remove_resource(resource_name, self.admin)
+
+    def test_detached_mode_in_host_list_not_case_matching__issue_4421(self):
+        try:
+            file1 = "f1"
+            file2 = "f2"
+            resource_host = "irods.org"
+            resource_name = 'detached_resource'
+            hostname = lib.get_hostname().upper()
+            resource_context = 'host_mode=detached;host_list=irods.org,%s' % hostname
+
+            lib.create_ufs_resource(resource_name, self.admin, resource_host)
+            self.admin.assert_icommand("iadmin modresc %s context %s" %
+                                   (resource_name, resource_context))
+
+            # create file to put
+            lib.make_file(file1, 100)
+
+            # put small file
+            self.admin.assert_icommand("iput -R %s %s" % (resource_name, file1))  # iput
+
+            # get file
+            self.admin.assert_icommand("iget %s %s" % (file1, file2))  # iget
+
+            # make sure the file that was put and got are the same
+            self.admin.assert_icommand("diff %s %s " % (file1, file2))
+        finally:
+            # local cleanup
+            self.admin.assert_icommand("irm -f " + file1)
+
+            if os.path.exists(file1):
+                os.unlink(file1)
+            if os.path.exists(file2):
+                os.unlink(file2)
+
+            # cleanup
+            lib.remove_resource(resource_name, self.admin)
+
+    def test_detached_mode_host_not_in_host_list__issue_4421(self):
+        try:
+            file1 = "f1"
+            resource_host = "irods.org"
+            resource_name = 'detached_resource'
+            resource_context = 'host_mode=detached;host_list=irods.org,some_other_host'
+
+            lib.create_ufs_resource(resource_name, self.admin, resource_host)
+            self.admin.assert_icommand("iadmin modresc %s context %s" %
+                                   (resource_name, resource_context))
+
+            # create file to put
+            lib.make_file(file1, 100)
+
+            # put small file
+            self.admin.assert_icommand("iput -R %s %s" % (resource_name, file1), 'STDERR', 'USER_SOCK_CONNECT_ERR')  # iput
+            self.admin.assert_icommand("ils -L %s" % file1, 'STDERR_SINGLELINE', "does not exist")  # should not be listed
+        finally:
+            if os.path.exists(file1):
+                os.unlink(file1)
+
+            # cleanup
+            lib.remove_resource(resource_name, self.admin)
+
+    def test_attached_mode_default_setting_invalid_host__issue_4421(self):
+        try:
+            file1 = "f1"
+            resource_host = "irods.org"
+            resource_name = 'attached_resource'
+
+            lib.create_ufs_resource(resource_name, self.admin, resource_host)
+
+            # create file to put
+            lib.make_file(file1, 100)
+
+            # put small file
+            self.admin.assert_icommand("iput -R %s %s" % (resource_name, file1), 'STDERR', 'USER_SOCK_CONNECT_ERR')  # iput
+            self.admin.assert_icommand("ils -L %s" % file1, 'STDERR_SINGLELINE', "does not exist")  # should not be listed
+        finally:
+            if os.path.exists(file1):
+                os.unlink(file1)
+
+            # cleanup
+            lib.remove_resource(resource_name, self.admin)
+
+    def test_attached_mode_explicit_setting_invalid_host__issue_4421(self):
+        try:
+            file1 = "f1"
+            resource_host = "irods.org"
+            resource_name = 'attached_resource'
+            resource_context = 'host_mode=attached'
+
+            lib.create_ufs_resource(resource_name, self.admin, resource_host)
+            self.admin.assert_icommand("iadmin modresc %s context %s" %
+                                   (resource_name, resource_context))
+
+            # create file to put
+            lib.make_file(file1, 100)
+
+            # put small file
+            self.admin.assert_icommand("iput -R %s %s" % (resource_name, file1), 'STDERR', 'USER_SOCK_CONNECT_ERR')  # iput
+            self.admin.assert_icommand("ils -L %s" % file1, 'STDERR_SINGLELINE', "does not exist")  # should not be listed
+        finally:
+            if os.path.exists(file1):
+                os.unlink(file1)
+
+            # cleanup
+            lib.remove_resource(resource_name, self.admin)
+
+    @unittest.skip("Currently the warning only displays if there is no redirect")
+    def test_detached_mode_register_outside_vault__issue_4421(self):
+        try:
+            file1 = "f1"
+
+            # create file to put
+            filepath = lib.create_local_testfile(file1)
+            registered_path = os.path.join(self.admin.session_collection, file1)
+
+            resource_host = lib.get_hostname().upper()
+            resource_name = 'detached_resource'
+            resource_context = 'host_mode=detached'
+
+            lib.create_ufs_resource(resource_name, self.admin, resource_host)
+            self.admin.assert_icommand("iadmin modresc %s context %s" %
+                                   (resource_name, resource_context))
+
+            self.admin.assert_icommand("ireg -R %s %s %s" % (resource_name, filepath, registered_path),
+                    'STDOUT_SINGLELINE',
+                    "Warning:  The registration was performed on a unixfilesystem resource [%s] currently configured as 'host_mode=detached'."
+                    % resource_name)
+
+        finally:
+            # cleanup
+            self.admin.assert_icommand("irm -f " + registered_path)
+            if os.path.exists(filepath):
+                os.unlink(filepath)
+            lib.remove_resource(resource_name, self.admin)
+
+    def test_detached_mode_register_inside_vault__issue_4421(self):
+        try:
+            resource_host = lib.get_hostname().upper()
+            resource_name = 'detached_resource'
+            resource_context = 'host_mode=detached'
+
+            lib.create_ufs_resource(resource_name, self.admin, resource_host)
+            self.admin.assert_icommand("iadmin modresc %s context %s" %
+                                   (resource_name, resource_context))
+
+            # create file to register
+            file1 = "f1"
+            registered_path = os.path.join(self.admin.session_collection, file1)
+            vault_path = self.admin.get_vault_path(resource=resource_name)
+            os.mkdir(vault_path)
+            filepath = '%s/%s' % (vault_path, file1)
+            with open(filepath, 'w+') as f:
+                print('I AM A TESTFILE -- [' + self.testfile + ']', file=f, end='')
+
+            self.admin.assert_icommand("ireg -R %s %s %s" % (resource_name, filepath, registered_path))
+
+        finally:
+            # cleanup
+            self.admin.assert_icommand("irm -f " + registered_path)
+            if os.path.exists(filepath):
+                os.unlink(filepath)
+            lib.remove_resource(resource_name, self.admin)
+
+    @unittest.skipUnless(test.settings.RUN_IN_TOPOLOGY, "Only run in topology")
+    def test_detached_mode_in_topology__issue_4421(self):
+        try:
+            file1 = "f1"
+            file2 = "f2"
+            resource_name = 'detached_resource'
+            resource_context = 'host_mode=detached'
+            shared_mount = paths.test_mount_directory()
+            vault_name = '%s_vault' % resource_name
+            vault_path = os.path.join(shared_mount, os.path.basename(self.admin.local_session_dir),\
+                    vault_name)
+
+            # make the resource
+            self.admin.assert_icommand(['iadmin', "mkresc", resource_name, 'unixfilesystem',\
+                    test.settings.HOSTNAME_2 + ":" + vault_path],\
+                    'STDOUT_SINGLELINE', 'unixfilesystem')
+
+            # create file to put
+            lib.make_file(file1, 100)
+
+            # put small file onto HOSTNAME_2
+            self.admin.assert_icommand("iput -R %s %s" % (resource_name, file1))
+
+            # modify resource to make it detached
+            self.admin.assert_icommand("iadmin modresc %s context %s" %
+                                   (resource_name, resource_context))
+
+            # pause HOSTNAME_2
+            with session.make_session_for_existing_admin() as admin_session:
+                admin_session.environment_file_contents = IrodsConfig().client_environment
+                admin_session.assert_icommand("irods-grid pause --hosts=%s" % test.settings.HOSTNAME_2,\
+                        'STDOUT_SINGLELINE', 'pausing')
+
+            # get file
+            self.admin.assert_icommand("iget %s %s" % (file1, file2))
+
+            # make sure the file that was put and got are the same
+            self.admin.assert_icommand("diff %s %s " % (file1, file2))
+
+        finally:
+            # cleanup
+            if os.path.exists(file1):
+                os.unlink(file1)
+            if os.path.exists(file2):
+                os.unlink(file2)
+
+            # resume HOSTNAME_2
+            with session.make_session_for_existing_admin() as admin_session:
+                admin_session.environment_file_contents = IrodsConfig().client_environment
+                admin_session.assert_icommand("irods-grid resume --hosts=%s" % test.settings.HOSTNAME_2,\
+                        'STDOUT_SINGLELINE', 'resuming')
+
+            self.admin.assert_icommand("irm -f " + file1)
+            lib.remove_resource(resource_name, self.admin)
+
+    @unittest.skipUnless(test.settings.RUN_IN_TOPOLOGY, "Only run in topology")
+    def test_detached_mode_in_topology_resource_assigned_to_icat__issue_4421(self):
+        # see irods_resource_plugin_s3 issue #2082
+        try:
+            file1 = "f1"
+            file2 = "f2"
+            resource_name = 'detached_resource'
+            resource_context = 'host_mode=detached'
+            shared_mount = paths.test_mount_directory()
+            vault_name = '%s_vault' % resource_name
+            vault_path = os.path.join(shared_mount, os.path.basename(self.admin.local_session_dir),\
+                    vault_name)
+
+            # make the resource
+            self.admin.assert_icommand(['iadmin', "mkresc", resource_name, 'unixfilesystem',\
+                    test.settings.ICAT_HOSTNAME + ":" + vault_path],\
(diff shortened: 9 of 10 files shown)


trel:
I like the flexibility of note 3 above... perhaps consider `SERVER_LIST=server1.irods.org:/lustre_mount/resource1` (using a colon instead of a comma... like we do for `iadmin mkresc`.    Would also need to consider default behavior if the vaultpath isn't specified for one of those servers (I suppose it would just revert to the path defined for the entire resource (note 1)).

scripts/irods/test/resource_suite.py
```diff
@@ -1167,3 +1167,105 @@ def test_itrim_no_op(self):
...
+            file2 = "f2"
+            resource_host = "irods.org"
+            resource_name = 'detached_resource'
+
+            hostuser = getpass.getuser()
+
+            resource_context = 'HOST_MODE=detached'
+            testvault = "/tmp/" + hostuser + "/" + resource_name
+
+            self.admin.assert_icommand("iadmin mkresc %s unixfilesystem %s:/%s %s" %
```
trel, line 1188: we have a lib call for this

scripts/irods/test/resource_suite.py
```diff
@@ -1167,3 +1167,105 @@ def test_itrim_no_op(self):
...
+                                   (resource_name, resource_host, testvault, resource_context), 'STDOUT_SINGLELINE', "Creating")
+
+            # create file to put
+            lib.make_file(file1, 100)
+
+            # put small file
+            self.admin.assert_icommand("iput -R %s %s" % (resource_name, file1))  # iput
+
+            # get file
+            self.admin.assert_icommand("iget %s %s" % (file1, file2))  # iput
```
trel, line 1198: "# iput" -> "# iget"

JustinKyleJames:
> I like the flexibility of note 3 above... perhaps consider `SERVER_LIST=server1.irods.org:/lustre_mount/resource1` (using a colon instead of a comma... like we do for `iadmin mkresc`. Would also need to consider default behavior if the vaultpath isn't specified for one of those servers (I suppose it would just revert to the path defined for the entire resource (note 1)).

Yes, I think flipping the colons and commas makes far more sense.

If the vault path isn't specified, I think using the existing one makes sense.

scripts/irods/test/resource_suite.py
```diff
@@ -1167,3 +1167,105 @@ def test_itrim_no_op(self):
...
+        try:
+            file1 = "f1"
+            resource_host = "irods.org"
+            resource_name = 'detached_resource'
+
+            hostuser = getpass.getuser()
+
+            testvault = "/tmp/" + hostuser + "/" + resource_name
+
+            self.admin.assert_icommand("iadmin mkresc %s unixfilesystem %s:/%s" %
```
trel, line 1256: lib call for this

scripts/irods/test/resource_suite.py
```diff
@@ -1167,3 +1167,105 @@ def test_itrim_no_op(self):
...
+            # put small file
+            self.admin.assert_icommand_fail("iput -R %s %s" % (resource_name, file1))  # iput
+
+        finally:
+
+            if os.path.exists(file1):
+                os.unlink(file1)
+
+            # cleanup
+            self.admin.assert_icommand("iadmin rmresc %s" % resource_name, 'EMPTY')
```
trel, line 1271: lib call for this?

scripts/irods/test/resource_suite.py
```diff
@@ -1167,3 +1167,105 @@ def test_itrim_no_op(self):
...
```
trel, line 1243: lib call for this?

scripts/irods/test/resource_suite.py
```diff
@@ -1167,3 +1167,105 @@ def test_itrim_no_op(self):
...
+            file1 = "f1"
+            resource_host = "irods.org"
+            resource_name = 'detached_resource'
+
+            hostuser = getpass.getuser()
+
+            resource_context = 'HOST_MODE=attached'
+            testvault = "/tmp/" + hostuser + "/" + resource_name
+
+            self.admin.assert_icommand("iadmin mkresc %s unixfilesystem %s:/%s %s" %
```
trel, line 1228: lib call for this

scripts/irods/test/resource_suite.py
```diff
@@ -1167,3 +1167,105 @@ def test_itrim_no_op(self):
...
+                                   (resource_name, resource_host, testvault, resource_context), 'STDOUT_SINGLELINE', "Creating")
+
+            # create file to put
+            lib.make_file(file1, 100)
+
+            # put small file
+            self.admin.assert_icommand_fail("iput -R %s %s" % (resource_name, file1))  # iput
```
trel, line 1235: can you specify more clearly the failure received - aka use `assert_icommand` with checked error code, rather than `assert_icommand_fail`

plugins/resources/src/unixfilesystem.cpp
```diff
@@ -83,6 +84,80 @@
...
+
+    bool detached_mode = is_detached_mode(prop_map);
+
+    if (detached_mode) {
+
+        bool error = false;
+
+        // update host to new host
+        char resource_location[MAX_NAME_LEN];
+        gethostname(resource_location, MAX_NAME_LEN);
```
trel, line 119: do we have an `irods::` way to do this?   i don't remember, but i thought we were getting away from calling this directly.

trel (commented):
seems so simple when it's laid out.   i like it.

trel:
> > I like the flexibility of note 3 above... perhaps consider `SERVER_LIST=server1.irods.org:/lustre_mount/resource1` (using a colon instead of a comma... like we do for `iadmin mkresc`. Would also need to consider default behavior if the vaultpath isn't specified for one of those servers (I suppose it would just revert to the path defined for the entire resource (note 1)).
> 
> Yes, I think flipping the colons and commas makes far more sense.
> 

cool, feels more correct.

> If the vault path isn't specified, I think using the existing one makes sense.
> 

yeps

> Also if SERVER_LIST (or whatever name we call it) doesn't exist we would simply assume all servers can serve the request (since detached mode is enabled).

Definitely, empty or undefined list means you get 'all servers' behavior.

Alternate name could be `HOST_LIST`... discush...

JustinKyleJames:
> Alternate name could be `HOST_LIST`... discush...

That sounds good to me.

kript:
> > Also if SERVER_LIST (or whatever name we call it) doesn't exist we would simply assume all servers can serve the request (since detached mode is enabled).
> 
> Definitely, empty or undefined list means you get 'all servers' behavior.

Excited to see this, can I ask that the detached behaviour is made explicit in the documentation, so that its clear that enabling this on a resource (can it be done after creation? or only at creation time?) means that _every_ server (provider *and* consumer?) would start to answer calls for this resource?

I can see 'every server' playing havoc with a pass-thru tree structure that has been set up to minimise traffic to one side of the tree, for example, if consumers held on servers that hold resources from both sides of the tree 'suddenly' started answering calls for the resource - whether they had it mounted or not!

Perhaps some links to best practices around mounting file systems across several systems might be a good idea (e..g pros and cons of using the automounter etc). As a consortium partner, NetApp seem to have some good ones they might allow you to point at e.g. [this one](https://docslib.org/doc/12830439/automount-implementation-guide-best-practices-tuning-and-troubleshooting-tips)

scripts/irods/test/resource_suite.py
JustinKyleJames: I see lib.create_ufs_resource().  However, it does not set a context.  Would you prefer I use that to create a resource and then modify the resource with assert_icommand()?

I get that you don't want it in /tmp.

scripts/irods/test/resource_suite.py
JustinKyleJames: done

trel:
it's just a context string setting, so it can be set/manipulated after creation time (same as S3, and every other resource context string).

scripts/irods/test/resource_suite.py
trel: > However, it does not set a context  

ohhh, i see... that seems like a missing bit of functionality...

we should probably open a new issue for adding context string to that function.

should just be able to add a new parameter, right?   `context=None`?

https://github.com/irods/irods/blob/aea8436ad097e561222cba5f101fa5788d01e1fa/scripts/irods/lib.py#L593

plugins/resources/src/unixfilesystem.cpp
JustinKyleJames: Still looking for this.

scripts/irods/test/resource_suite.py
```diff
@@ -1167,3 +1167,96 @@ def test_itrim_no_op(self):
...
+    # detached mode
+    ###################
+    def test_detached_mode(self):
+
+        try:
+            file1 = "f1"
+            file2 = "f2"
+            resource_host = "irods.org"
+            resource_name = 'detached_resource'
+
```
korydraughn, line 1182: Why an empty line here? The next line seems like it can be grouped with the ones above.

Same thing applies to other tests.

scripts/irods/test/resource_suite.py
```diff
@@ -1167,3 +1167,96 @@ def test_itrim_no_op(self):
 
         # try to trim down to repl_count
         self.admin.assert_icommand("itrim -N {repl_count} {filename}".format(**locals()), 'STDOUT_SINGLELINE', "Total size trimmed = 0.000 MB. Number of files trimmed = 0.")
+
+
+    ###################
+    # detached mode
+    ###################
+    def test_detached_mode(self):
```
korydraughn, line 1175: Missing issue number in test name.

scripts/irods/test/resource_suite.py
```diff
@@ -1167,3 +1167,96 @@ def test_itrim_no_op(self):
...
+
```
korydraughn, line 1176: Remove empty line.

scripts/irods/test/resource_suite.py
```diff
@@ -1167,3 +1167,96 @@ def test_itrim_no_op(self):
...
+            self.admin.assert_icommand("iput -R %s %s" % (resource_name, file1))  # iput
+
+            # get file
+            self.admin.assert_icommand("iget %s %s" % (file1, file2))  # iget
+
+            # make sure the file that was put and got are the same
+            self.admin.assert_icommand("diff %s %s " % (file1, file2), 'EMPTY')
+
+        finally:
+
```
korydraughn, line 1202: Remove empty line.

scripts/irods/test/resource_suite.py
```diff
@@ -1167,3 +1167,96 @@ def test_itrim_no_op(self):
...
+            if os.path.exists(file1):
+                os.unlink(file1)
+            if os.path.exists(file2):
+                os.unlink(file2)
+
+            # cleanup
+            lib.remove_resource(resource_name, self.admin)
+
+    def test_attached_mode_default_setting_invalid_host(self):
```
korydraughn, line 1214: Missing issue number in test name.

scripts/irods/test/resource_suite.py
```diff
@@ -1167,3 +1167,96 @@ def test_itrim_no_op(self):
...
+
```
korydraughn, line 1215: Remove empty line.

scripts/irods/test/resource_suite.py
```diff
@@ -1167,3 +1167,96 @@ def test_itrim_no_op(self):
...
+            lib.create_ufs_resource(resource_name, self.admin, resource_host)
+
+            # create file to put
+            lib.make_file(file1, 100)
+
+            # put small file
+            self.admin.assert_icommand("iput -R %s %s" % (resource_name, file1), 'STDERR', 'USER_SOCK_CONNECT_ERR')  # iput
+
+        finally:
+
```
korydraughn, line 1230: Remove empty line.

scripts/irods/test/resource_suite.py
```diff
@@ -1167,3 +1167,96 @@ def test_itrim_no_op(self):
...
+            if os.path.exists(file1):
+                os.unlink(file1)
+
+            # cleanup
+            lib.remove_resource(resource_name, self.admin)
+
+    def test_attached_mode_explicit_setting_invalid_host(self):
```
korydraughn, line 1237: Missing issue number.

scripts/irods/test/resource_suite.py
```diff
@@ -1167,3 +1167,96 @@ def test_itrim_no_op(self):
...
+
```
korydraughn, line 1238: Remove empty line.

scripts/irods/test/resource_suite.py
```diff
@@ -1167,3 +1167,96 @@ def test_itrim_no_op(self):
...
+                                   (resource_name, resource_context), 'EMPTY')
+
+            # create file to put
+            lib.make_file(file1, 100)
+
+            # put small file
+            self.admin.assert_icommand("iput -R %s %s" % (resource_name, file1), 'STDERR', 'USER_SOCK_CONNECT_ERR')  # iput
+
+        finally:
+
```
korydraughn, line 1257: Remove empty line.

scripts/irods/test/resource_suite.py
```diff
@@ -1167,3 +1167,96 @@ def test_itrim_no_op(self):
...
+            file1 = "f1"
+            file2 = "f2"
+            resource_host = "irods.org"
+            resource_name = 'detached_resource'
+
+            resource_context = 'HOST_MODE=detached'
+
+            lib.create_ufs_resource(resource_name, self.admin, resource_host)
+            self.admin.assert_icommand("iadmin modresc %s context %s" %
+                                   (resource_name, resource_context), 'EMPTY')
```
korydraughn, line 1187: I think it's safe to not specify `'EMPTY'` when you don't expect any output.

scripts/irods/test/resource_suite.py
```diff
@@ -1167,3 +1167,96 @@ def test_itrim_no_op(self):
...
+            lib.make_file(file1, 100)
+
+            # put small file
+            self.admin.assert_icommand("iput -R %s %s" % (resource_name, file1))  # iput
+
+            # get file
+            self.admin.assert_icommand("iget %s %s" % (file1, file2))  # iget
+
+            # make sure the file that was put and got are the same
+            self.admin.assert_icommand("diff %s %s " % (file1, file2), 'EMPTY')
```
korydraughn, line 1199: Remove `'EMPTY'` argument.

scripts/irods/test/resource_suite.py
```diff
@@ -1167,3 +1167,96 @@ def test_itrim_no_op(self):
...
+
+        finally:
+
+            # local cleanup
+            self.admin.assert_icommand("irm -f " + file1, 'EMPTY')
```
korydraughn, line 1204: Remove `'EMPTY'` argument.

scripts/irods/test/resource_suite.py
```diff
@@ -1167,3 +1167,96 @@ def test_itrim_no_op(self):
...
+        try:
+            file1 = "f1"
+            resource_host = "irods.org"
+            resource_name = 'attached_resource'
+
+            resource_context = 'HOST_MODE=attached'
+
+            lib.create_ufs_resource(resource_name, self.admin, resource_host)
+            self.admin.assert_icommand("iadmin modresc %s context %s" %
+                                   (resource_name, resource_context), 'EMPTY')
```
korydraughn, line 1248: Remove `'EMPTY'` argument.

plugins/resources/src/unixfilesystem.cpp
```diff
@@ -83,6 +84,80 @@
 const std::string DEFAULT_VAULT_DIR_MODE( "default_vault_directory_mode_kw" );
 const std::string HIGH_WATER_MARK( "high_water_mark" ); // no longer used
 const std::string REQUIRED_FREE_INODES_FOR_CREATE("required_free_inodes_for_create"); // no longer used
+const std::string HOST_MODE("HOST_MODE");
```
korydraughn, line 87: What's the conventions around resource context string? Are they defined?

korydraughn:
> > If the vault path isn't specified, I think using the existing one makes sense.
> 
> yeps

Q. Are there use-cases for not specifying a vault path while in detached mode?
Q. What happens if the admin means to include a vault path, but forgets to include it? Can this lead to unwanted behavior?

scripts/irods/test/resource_suite.py
trel: That's true - 'EMPTY' is the default... but being explicit is okay too?

We haven't written our preference for either anywhere I don't think...

trel:
> Q. Are there use-cases for not specifying a vault path while in detached mode?

yes, we want a subset of hosts, (aka HOST_LIST), but they will each be provided with a common/identical lustre mountpoint

> Q. What happens if the admin means to include a vault path, but forgets to include it? Can this lead to unwanted behavior?

Nothing horrible - any request serviced by that host would look in the default / common location, and probably hit a UNIX file error since it wouldn't exist or it wouldn't have correct permissions to read/write.   Same as any other misconfiguration.  And our error message would have the attempted path included, so should be easy enough to debug.    Perhaps with a bonus note of '(default)'.

plugins/resources/src/unixfilesystem.cpp
korydraughn: Does this need to consider host resolution info (i.e. hosts_config.json in 4-2-stable)?

trel: right - i feel like we made an official frontdoor for this type of check/lookup...

plugins/resources/src/unixfilesystem.cpp
```diff
@@ -102,6 +106,91 @@ bool is_detached_mode(irods::plugin_property_map& prop_map) {
...
+    // have a HOST_LIST parameter
+
+    // split parameter by delimiter (,)
+    char delimiter = ',';
+    std::string token;
+
+    std::string resource_hostname_with_colon = resource_hostname + ":";
+    std::string resource_hostname_with_comma = resource_hostname + ",";
+
+    std::vector<std::string> tokens = split(host_list_str, delimiter);
```
trel, line 156: should we trim any remaining whitespace?  do we do that anywhere else?

unit_tests/cmake/test_config/irods_test_host_list_context_string.cmake
```diff
@@ -0,0 +1,14 @@
...
+                            ${CMAKE_CURRENT_SOURCE_DIR}/src/test_host_list_context_string.cpp)
+
+set(IRODS_TEST_INCLUDE_PATH ${IRODS_EXTERNALS_FULLPATH_BOOST}/include
+                            ${IRODS_EXTERNALS_FULLPATH_FMT}/include)
+
+set(IRODS_TEST_LINK_LIBRARIES irods_common
+                              irods_client
+                              irods_plugin_dependencies
+                              #irods_resource_plugin-unixfilesystem
+                              /usr/lib/irods/plugins/resources/libunixfilesystem.so
```
trel, line 13: we still want the first flavor, right?

trel (commented):
tests look good.   will be interesting to see how to allow logging via unit test calls.

plugins/resources/src/unixfilesystem.cpp
JustinKyleJames: I added resolve_hostname().

unit_tests/cmake/test_config/irods_test_host_list_context_string.cmake
JustinKyleJames: Yes, but that doesn't work.  I am still looking into how to get that to work.

plugins/resources/src/unixfilesystem.cpp
```diff
@@ -83,6 +84,80 @@
...
+            error = true;
+
+        } else {
+
+            rodsLong_t resc_id = 0;
+
+            ret = resc_mgr.hier_to_leaf_id(resource_name, resc_id);
+            if( !ret.ok() ) {
+                error = true;
+            } else {
```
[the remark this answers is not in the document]
JustinKyleJames: Are you ok with repeated log entries?  I suppose you want more explicit log messages for each case which has a side effect of not needing the error flag and simpler flow control.  Correct?

plugins/resources/src/unixfilesystem.cpp
```diff
@@ -83,6 +84,80 @@
...
 const std::string REQUIRED_FREE_INODES_FOR_CREATE("required_free_inodes_for_create"); // no longer used
+const std::string HOST_MODE("HOST_MODE");
+
+bool is_detached_mode(irods::plugin_property_map& prop_map) {
+
+    std::string host_mode_str;
+
+    irods::error ret = prop_map.get< std::string >(HOST_MODE, host_mode_str);
+    if (ret.ok()) {
+
```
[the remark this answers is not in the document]
JustinKyleJames: done

plugins/resources/src/unixfilesystem.cpp
```diff
@@ -83,6 +84,80 @@
...
+        if (boost::iequals(host_mode_str.c_str(), "detached"))  {
+            return true;
+        }
+    }
+
+    // default is attached mode
+    return false;
+}
+
+irods::error unixfilesystem_start_operation(irods::plugin_property_map& prop_map)
```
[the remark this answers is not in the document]
JustinKyleJames: done

plugins/resources/src/unixfilesystem.cpp
```diff
@@ -83,6 +84,80 @@
...
+{
+    using logger = irods::experimental::log;
+
+    irods::error ret = SUCCESS();
+
+    bool detached_mode = is_detached_mode(prop_map);
+
+    if (detached_mode) {
+
+        bool error = false;
```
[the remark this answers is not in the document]
JustinKyleJames: done

plugins/resources/src/unixfilesystem.cpp
```diff
@@ -83,6 +84,80 @@
...
+
+        // update host to new host
+        char resource_location[MAX_NAME_LEN];
+        gethostname(resource_location, MAX_NAME_LEN);
+
+        std::string resource_name;
+        ret = prop_map.get<std::string>(irods::RESOURCE_NAME, resource_name);
+        if ( !ret.ok() ) {
+
```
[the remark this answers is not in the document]
JustinKyleJames: done

plugins/resources/src/unixfilesystem.cpp
```diff
@@ -83,6 +84,80 @@
...
+            error = true;
+
+        } else {
+
+            rodsLong_t resc_id = 0;
+
+            ret = resc_mgr.hier_to_leaf_id(resource_name, resc_id);
```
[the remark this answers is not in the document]
JustinKyleJames: done

plugins/resources/src/unixfilesystem.cpp
```diff
@@ -83,6 +84,80 @@
...
+            if( !ret.ok() ) {
+                error = true;
+            } else {
+
+                rodsServerHost_t *resource_host = nullptr;
+                ret = irods::get_resource_property< rodsServerHost_t*& >(resc_id, irods::RESOURCE_HOST, resource_host);
+
+                if (!ret.ok() || !resource_host) {
+                     error = true;
+                } else {
```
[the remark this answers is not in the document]
JustinKyleJames: done

plugins/resources/src/unixfilesystem.cpp
```diff
@@ -83,6 +84,80 @@
...
+        }
+
+        if (error) {
+            // log a warning but continue
+            logger::resource::warn("[resource_name={}] Attached mode failed to set RESOURCE_HOST to {}.",
+                    resource_name.c_str(), resource_location);
+        }
+    }
+
+    return ret;
```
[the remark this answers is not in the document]
JustinKyleJames: done, although I am not returning the error because I am afraid the system may treat it more severely than the warning I intended.  It is always returning SUCCESS() now.

scripts/irods/test/resource_suite.py
JustinKyleJames: done

scripts/irods/test/resource_suite.py
JustinKyleJames: done

scripts/irods/test/resource_suite.py
JustinKyleJames: done

scripts/irods/test/resource_suite.py
JustinKyleJames: done

scripts/irods/test/resource_suite.py
JustinKyleJames: done

unit_tests/CMakeLists.txt
```diff
@@ -75,6 +75,7 @@ set(
   with_durability
   json_apis_from_client
   zone_report
+  test_host_list_context_string
```
SwooshyCueb, line 78: For consistency with other unit tests, this should be `host_list_context_string`, and `irods_test_host_list_context_string.cmake` should be named `irods_host_list_context_string.cmake`. Our current naming scheme only puts `test_` in the `.cpp` filename.

plugins/resources/src/unixfilesystem.cpp
```diff
@@ -83,6 +87,198 @@
...
+            result.push_back(token);
+        }
+    }
+    return result;
+}
+
+// returns a tuple with the following:
+//   bool - true iff the resource_hostname is in HOST_LIST string
+//   std::optional<std::string> - the new vault path if it is defined
+std::tuple<bool, std::optional<std::string>> get_detached_mode_vault_path(
```
SwooshyCueb, line 127: You might want to add a comment here about the forward declaration in the unit test, as if the function signature is ever changed here, it will also need to be changed in the unit test.

unit_tests/src/test_host_list_context_string.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+#include <catch2/catch.hpp>
+
+#include <string>
+#include <sstream>
+#include <vector>
+#include <tuple>
+#include <optional>
+
+#include <irods/irods_plugin_base.hpp>
```
SwooshyCueb, line 23: Use quotes instead of angle brackets for headers that are in the same repository.

unit_tests/cmake/test_config/irods_test_host_list_context_string.cmake
```diff
@@ -0,0 +1,13 @@
+set(IRODS_TEST_TARGET host_list_context_string)
```
SwooshyCueb, line 1: To be consistent with other unit tests, the target name should start with `irods_`.

SwooshyCueb (changes requested):

plugins/resources/src/unixfilesystem.cpp
```diff
@@ -83,6 +87,198 @@
...
+
+    // split parameter by delimiter (,)
+    char delimiter = ',';
+    std::string token;
+
+    std::string resource_hostname_with_colon = resource_hostname + ":";
+
+    std::vector<std::string> tokens = split(host_list_str, delimiter);
+
+    for (std::string token : tokens) {
```
SwooshyCueb, line 158: `clazy` suggests a change here:

> Missing reference in range-for with non trivial type (`std::string`) [`clazy-range-for-loop-reference`]

```suggestion
    for (std::string& token : tokens) {
```

plugins/resources/src/unixfilesystem.cpp
```diff
@@ -83,6 +87,198 @@
...
+                return true;
+            }
+        }
+
+        // default is attached mode
+        return false;
+    }
+}
+
+static std::vector<std::string> split(const std::string str, char delim)
```
SwooshyCueb, line 111: Is there a particular reason to define a static `split` function here rather than using the existing `irods::string_tokenize`?

If explicitly using a `char`-typed delimiter is desired, I would suggest overloading `irods::string_tokenize`.
If `irods::string_tokenize` is in need of modernization, then let's do that.

SwooshyCueb:
I have a solution for the unit test linking problem: Split the code to be tested out into an object library target, and link to that instead.

I've done this myself over here: https://github.com/SwooshyCueb/irods/commit/50648f47536814b8ea13b4eb54bb42a96ce15704
Once #6539 is merged, you can cherry-pick this into your branch.
It removes the `get_detached_mode_vault_path` declaration from the unit test in favor of including the new header file, so you won't have to worry about adding the comment I suggested.
Additionally, it also contains my suggested change to the range-for loop.
It does not include any of my other included changes. Where possible, I left the code untouched from how it is in your branch.

plugins/resources/src/unixfilesystem.cpp
JustinKyleJames: As far as things like semicolon delimiting different keys and equal separating key from value, that is all handled in irods::parse_kvp_string() which is called in the constructor.

We did talk about case and unlike the S3 plugin the unixfilesystem plugin uses lowercase for keys so I will update it to lowercase.

unit_tests/CMakeLists.txt
JustinKyleJames: done

unit_tests/src/test_host_list_context_string.cpp
JustinKyleJames: done

plugins/resources/src/unixfilesystem.cpp
JustinKyleJames: Done.  Also added a const qualifier.

unit_tests/cmake/test_config/irods_test_host_list_context_string.cmake
JustinKyleJames: Done

plugins/resources/src/unixfilesystem.cpp
JustinKyleJames: done

plugins/resources/src/unixfilesystem.cpp
JustinKyleJames: I didn't know about it or didn't think of it.  That is much better than creating my own.  Thanks for pointing it out.  It was seamless to swap that out.

unit_tests/CMakeLists.txt
JustinKyleJames: done

unit_tests/cmake/test_config/irods_test_host_list_context_string.cmake
JustinKyleJames: This now has the magic that Markus added in.

trel:
clang-format has some opinions.

plugins/resources/unixfilesystem/src/irods_get_detached_mode_vault_path.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include <string>
+#include <tuple>
+
+const std::string HOST_LIST("host_list");
+
+// returns a tuple with the following:
+//   bool - true iff the resource_hostname is in HOST_LIST string
+//   std::optional<std::string> - the new vault path if it is defined
+//
+//   Note:  There is a forward declaration of this in unit_tests/src/test_host_list_context_string.cpp
```
SwooshyCueb, line 20: Not needed any more since we have a header for this now

SwooshyCueb (changes requested):

plugins/resources/unixfilesystem/CMakeLists.txt
```diff
@@ -15,6 +15,10 @@ target_link_libraries(
 target_include_directories(
   irods_resource_plugin_obj-unixfilesystem
   PUBLIC
+  "${IRODS_EXTERNALS_FULLPATH_ARCHIVE}/include"
+  "${IRODS_EXTERNALS_FULLPATH_BOOST}/include"
+  "${IRODS_EXTERNALS_FULLPATH_FMT}/include"
+  "${IRODS_EXTERNALS_FULLPATH_SPDLOG}/include"
```
SwooshyCueb, line 21: I do not think these lines are needed for the object library target.

unit_tests/CMakeLists.txt
```diff
@@ -75,7 +75,7 @@ set(
   with_durability
   json_apis_from_client
   zone_report
-  ticket_administration
```
SwooshyCueb, line 78: Are we removing this test?

JustinKyleJames: That was probably my mistake.  I will add back in.

plugins/resources/unixfilesystem/src/irods_get_detached_mode_vault_path.cpp
JustinKyleJames: Yes, it is no longer there.  I just forgot to remove the comment.  Will remove.

plugins/resources/unixfilesystem/src/unixfilesystem.cpp
```diff
@@ -202,101 +206,89 @@ auto unix_file_start_operation(irods::plugin_property_map& prop_map) -> irods::e
...
-            if (outFd > 0) close(outFd);
-        }
-    };
-
-    inFd = open( srcFileName, O_RDONLY, 0 );
+    const irods::at_scope_exit close_fds{[&]() {
+        if (inFd > 0)
+            close(inFd);
+        if (outFd > 0)
+            close(outFd);
```
trel, line 229: please add open/close braces around this... seems dangerous...

plugins/resources/unixfilesystem/src/unixfilesystem.cpp
```diff
@@ -36,153 +42,253 @@
...
+        ret = prop_map.get<std::string>(irods::RESOURCE_NAME, resource_name);
+        if (!ret.ok()) {
+            logger::resource::warn("[resource_name={}] Detached mode failed to set RESOURCE_HOST to {}."
+                                   "  Failed to get irods::RESOURCE_NAME property.",
+                                   resource_name.c_str(),
+                                   local_hostname);
+            return SUCCESS();
+        }
+
+        std::tie(in_host_list, new_vault_path_optional) =
```
alanking, line 142: Could use structured bindings here instead, i.e.:
```c++
auto [in_host_list, new_vault_path_optional] = get_detached_mode_vault_path(prop_map, resource_hostname, resource_name);
```

plugins/resources/unixfilesystem/src/unixfilesystem.cpp
```diff
@@ -36,153 +42,253 @@
...
+        std::optional<std::string> new_vault_path_optional;
+
+        // update host to new host
+        char local_hostname[MAX_NAME_LEN];
+        gethostname(local_hostname, MAX_NAME_LEN);
+
+        auto resource_hostname_optional = resolve_hostname(local_hostname, hostname_resolution_scheme::match_preferred);
+
+        std::string resource_hostname;
+        resource_hostname = resource_hostname_optional ? std::move(*resource_hostname_optional) : local_hostname;
```
alanking, line 130: Could combine these two lines

plugins/resources/unixfilesystem/src/unixfilesystem.cpp
```diff
@@ -36,153 +42,253 @@
...
+                return SUCCESS();
+            }
+
+            if (!resource_host) {
+                logger::resource::warn("[resource_name={}] Detached mode failed to set RESOURCE_HOST to {}."
+                                       "  The resource_host from resource property irods::RESOURCE_HOST was null.",
+                                       resource_name.c_str(),
+                                       local_hostname);
+                return SUCCESS();
+            }
```
alanking, line 174: This is dead code. We should either remove this or remove the `nullptr` check on 160.

scripts/irods/test/resource_suite.py
```diff
@@ -1167,3 +1167,181 @@ def test_itrim_no_op(self):
...
+
+    def test_detached_mode_vault_path_updated__issue_4421(self):
+        try:
+            file1 = "f1"
+            file2 = "f2"
+            resource_host = "irods.org"
+            resource_name = 'detached_resource'
+            vault_path = '/tmp/detached_resource'
+            hostname = lib.get_hostname()
+            resource_context = 'host_mode=detached;host_list=irods.org:/some/path,%s:%s;some_other_host:/some/other/path' % (hostname, vault_path)
```
alanking, line 1218: Should the semicolon right before `some_other_host` be a comma?

scripts/irods/test/resource_suite.py
```diff
@@ -1167,3 +1167,181 @@ def test_itrim_no_op(self):
...
+
+            lib.create_ufs_resource(resource_name, self.admin, resource_host)
+            self.admin.assert_icommand("iadmin modresc %s context %s" %
+                                   (resource_name, resource_context))
+
+            # create file to put
+            lib.make_file(file1, 100)
+
+            # put small file
+            self.admin.assert_icommand("iput -R %s %s" % (resource_name, file1), 'STDERR', 'USER_SOCK_CONNECT_ERR')  # iput
```
alanking, line 1298: Should we make sure that the data object doesn't exist here as well?

unit_tests/src/test_host_list_context_string.cpp
```diff
@@ -0,0 +1,175 @@
...
+    SECTION("not in host list")
+    {
+        irods::plugin_property_map prop_map;
+        prop_map.set<std::string>("host_list", "host2:/host2/path,host3");
+
+        bool in_host_list;
+        std::optional<std::string> new_vault_path_optional;
+
+        std::tie(in_host_list, new_vault_path_optional) =
+            get_detached_mode_vault_path(prop_map, resource_hostname, resource_name);
```
alanking, line 42: Structured bindings could be used here as well, it looks like. Same for other uses of `std::tie` in this file.

unit_tests/src/test_host_list_context_string.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+        REQUIRE(in_host_list == false);
+        REQUIRE(new_vault_path_optional.has_value() == false);
+    }
+
+    /* The following test causes the unixfilesystem plugin to log a warning
+     * using the new logger.
+     * This causes a SEGV because the unit test does not have the
+     * environment to do logging.
+     * Commenting this out for now.
```
alanking, line 67: Do the changes made by @SwooshyCueb make this better? Just wondering

alanking (commented):
In topology testing, we have a catalog service provider and 3 catalog service consumers, and the tests can be run from the provider or consumer 1 (i.e. `irods.test.settings.HOSTNAME_1`). Consider adding a test (or tests) that have a detached resource with catalog service consumers 2 and 3 (i.e. `irods.test.settings.HOSTNAME_2/3`) in the `host_list`. What happens when we put a file to that resource?

scripts/irods/test/resource_suite.py
JustinKyleJames: No, the entire string is part of the host_list KVP.  The comma separates host/paths which is not the same as semicolon which separates different KVP's.

trel: mmm, i think alan's right?

unit_tests/src/test_host_list_context_string.cpp
JustinKyleJames: No, it still fails with SEGV.

plugins/resources/unixfilesystem/src/irods_get_detached_mode_vault_path.cpp
```diff
@@ -0,0 +1,78 @@
...
+        // and original vault path used for all hosts
+        in_host_list = true;
+        return std::make_tuple(in_host_list, std::nullopt);
+    }
+
+    // have a HOST_LIST parameter
+
+    // split parameter by delimiter (,)
+    const std::string delimiter = ",";
+    std::string token;
```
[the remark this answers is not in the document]
alanking: I don't see where `token` is being used. Line 49 defines a `const std::string& token` as part of the range-based for loop and that is being used, but I don't see where *this* `token` is being used. I'm a little surprised that this doesn't generate a warning. Anyways, I was suggesting that `token` on line 42 can be deleted, I think. Is that right?

Played around with it here, too: https://godbolt.org/z/hf3qcx7TY

plugins/resources/unixfilesystem/CMakeLists.txt
JustinKyleJames: The archive was not but the others seem to be required.

plugins/resources/unixfilesystem/src/unixfilesystem.cpp
JustinKyleJames: Done

plugins/resources/unixfilesystem/src/unixfilesystem.cpp
JustinKyleJames: done

scripts/irods/test/resource_suite.py
JustinKyleJames: It should be a comma.  Will change.

plugins/resources/unixfilesystem/src/unixfilesystem.cpp
JustinKyleJames: done

scripts/irods/test/resource_suite.py
JustinKyleJames: done

scripts/irods/test/resource_suite.py
JustinKyleJames: done

JustinKyleJames:
We have decided that allowing the user to set the vault path in the host_list string is not advisable at this time so we've removed this functionality.

The reason is that the entire path is stored in the database.  That will cause issues when a server that has the mount point at a different location tries to read the object.

Example:

1. Server A has the mount point as /mnt1 and sets the vault path to /mnt1/resource1.
2. Server B has the mount point as /mnt2 and sets the vault path to /mnt2/resource1.
3. User is connected to server A and puts the file into /tempZone/home/rods/file1.  This file will go to /mnt1/resource1/tempZone/home/rods/file1.  This path will be written to the database.
4.  If the user connects to server B and tries to retrieve the file, it will try to read /mnt1/resource1/tempZone/home/rods/file1. This file does not exist on server B but instead will be under /mnt2.

For now we will require that all servers that are in the host_list have the exact same mount point.  We will also not allow registrations as that could pose other problems.

trel:
The registrations issue is very subtle...  do we just disallow non-vault registrations into unixfilesystem resources configured as `detached`?   That seems... a brittle solution.

Or... do we allow it to happen and the expect any error messages to be helpful enough to the administrator later?    Only admins can `ireg` by default... is that enough?

JustinKyleJames:
> The registrations issue is very subtle... do we just disallow non-vault registrations into unixfilesystem resources configured as detached? That seems... a brittle solution.

> Or... do we allow it to happen and the expect any error messages to be helpful enough to the administrator later? Only admins can ireg by default... is that enough?

Upon testing we can't reject the registration in the plugin.  There are registration operations but they run after registration is complete.  Returning an warning type error on the registration is possible.  That would mean that doing what you describe in the second paragraph would be the easiest solution as it can be done in the plugin.

plugins/resources/unixfilesystem/src/irods_is_in_host_list.cpp
```diff
@@ -0,0 +1,52 @@
+#include "irods/private/irods_is_in_host_list.hpp"
+
+#include "irods/filesystem/path.hpp"
+#include "irods/irods_error.hpp"
+#include "irods/irods_logger.hpp"
```
korydraughn, line 5: This header can be removed since it's not used in the implementation.

plugins/resources/unixfilesystem/src/irods_is_in_host_list.cpp
```diff
@@ -0,0 +1,52 @@
+#include "irods/private/irods_is_in_host_list.hpp"
+
+#include "irods/filesystem/path.hpp"
```
korydraughn, line 3: This header seems like it can be removed as well since nothing appears to use it.

plugins/resources/unixfilesystem/src/irods_is_in_host_list.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include "irods/irods_string_tokenize.hpp"
+
+// boost includes
+#include <boost/algorithm/string/predicate.hpp>
+
+// stl includes
+#include <vector>
+#include <sstream>
+#include <string>
+#include <tuple>
```
korydraughn, line 16: Remove since it's not used.

plugins/resources/unixfilesystem/src/irods_is_in_host_list.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include "irods/irods_logger.hpp"
+#include "irods/irods_lookup_table.hpp"
+#include "irods/irods_string_tokenize.hpp"
+
+// boost includes
+#include <boost/algorithm/string/predicate.hpp>
+
+// stl includes
+#include <vector>
+#include <sstream>
```
korydraughn, line 14: Remove since it's not used.

plugins/resources/unixfilesystem/include/irods/private/irods_is_in_host_list.hpp
```diff
@@ -0,0 +1,15 @@
...
+#include "irods/irods_lookup_table.hpp"
+
+#include <optional>
+#include <string>
+#include <tuple>
+
+auto is_host_in_host_list(
+    irods::plugin_property_map& prop_map,
+    const std::string& resource_hostname,
+    const std::string& resource_name) -> bool;
```
korydraughn, line 13: This function declaration and its implementation should be placed inside of the `irods` namespace.

korydraughn (changes requested):
Looking good.

plugins/resources/unixfilesystem/include/irods/private/irods_is_in_host_list.hpp
```diff
@@ -0,0 +1,15 @@
+#ifndef IRODS_IS_IN_HOST_LIST_HPP
+#define IRODS_IS_IN_HOST_LIST_HPP
+
+#include "irods/irods_lookup_table.hpp"
+
+#include <optional>
+#include <string>
+#include <tuple>
```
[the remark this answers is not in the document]
JustinKyleJames: done

plugins/resources/unixfilesystem/src/irods_is_in_host_list.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include <tuple>
+
+const std::string HOST_LIST("host_list");
+
+// returns true if there is no host_list or if the current host is in
+// the list
+auto is_host_in_host_list(
+    irods::plugin_property_map& prop_map,
+    const std::string& resource_hostname,
+    const std::string& resource_name) -> bool
```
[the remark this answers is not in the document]
JustinKyleJames: done

plugins/resources/unixfilesystem/src/irods_is_in_host_list.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+    // have a host_list parameter
+
+    // split parameter by delimiter (,)
+    const std::string delimiter = ",";
+    std::vector<std::string> tokens;
+    irods::string_tokenize(host_list_str, delimiter, tokens);
+
+    for (const std::string& token : tokens) {
+        // see if this token begins with our resource location but with no path
```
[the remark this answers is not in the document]
JustinKyleJames: Out-of-date comment.  Removed.

plugins/resources/unixfilesystem/src/irods_is_in_host_list.cpp
```diff
@@ -0,0 +1,52 @@
...
+// the list
+auto is_host_in_host_list(
+    irods::plugin_property_map& prop_map,
+    const std::string& resource_hostname,
+    const std::string& resource_name) -> bool
+{
+    std::string host_list_str;
+
+    irods::error ret = prop_map.get<std::string>(HOST_LIST, host_list_str);
+    if (!ret.ok()) {
```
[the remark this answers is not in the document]
JustinKyleJames: done

plugins/resources/unixfilesystem/src/irods_is_in_host_list.cpp
```diff
@@ -0,0 +1,52 @@
...
+        // no HOST_LIST parameter, all hosts assumed to be able to handle request
+        // and original vault path used for all hosts
+        return true;
+    }
+
+    // have a host_list parameter
```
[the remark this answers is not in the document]
JustinKyleJames: done

unit_tests/src/test_host_list_context_string.cpp
```diff
@@ -0,0 +1,109 @@
...
+//   - Creative Commons Attribution-Sharealike 3.0 Unported License (CC-BY-SA)
+//   - GNU Free Documentation License (GFDL)
+//
+// For more information about these licenses, visit:
+//
+//   - https://en.cppreference.com/w/Cppreference:FAQ
+//   - https://en.cppreference.com/w/Cppreference:Copyright/CC-BY-SA
+//   - https://en.cppreference.com/w/Cppreference:Copyright/GDFL
+//
+
```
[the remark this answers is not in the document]
JustinKyleJames: done

unit_tests/src/test_host_list_context_string.cpp
```diff
@@ -0,0 +1,109 @@
...
+    std::string resource_name = "my_resc";
+
+    SECTION("not in host list")
+    {
+        irods::plugin_property_map prop_map;
+        prop_map.set<std::string>("host_list", "host2,host3");
+
+        bool in_host_list = is_host_in_host_list(prop_map, resource_hostname, resource_name);
+
+        REQUIRE(in_host_list == false);
```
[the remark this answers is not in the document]
JustinKyleJames: done

plugins/resources/unixfilesystem/src/irods_is_in_host_list.cpp
JustinKyleJames: done

plugins/resources/unixfilesystem/src/irods_is_in_host_list.cpp
JustinKyleJames: done

plugins/resources/unixfilesystem/src/irods_is_in_host_list.cpp
JustinKyleJames: done

plugins/resources/unixfilesystem/src/irods_is_in_host_list.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include "irods/irods_lookup_table.hpp"
+#include "irods/irods_string_tokenize.hpp"
+
+// boost includes
+#include <boost/algorithm/string/predicate.hpp>
+
+// stl includes
+#include <vector>
+#include <sstream>
+#include <string>
```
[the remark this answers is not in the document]
JustinKyleJames: done

scripts/irods/paths.py
```diff
@@ -535,3 +535,13 @@ def possible_shm_locations():
...
+
+_test_mount_directory_cache = None
+def _test_mount_directory():
+    global _test_mount_directory_cache
+    if _test_mount_directory_cache is None:
+        _test_mount_directory_cache = '/irods_testing_environment_mount_dir'
+    return _test_mount_directory_cache
+
+def test_mount_directory():
+    return str(_test_mount_directory())
```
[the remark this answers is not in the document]
JustinKyleJames: done

plugins/resources/unixfilesystem/include/irods/private/irods_is_in_host_list.hpp
JustinKyleJames: done

alanking (commented):
The parts I looked over look pretty good

I'm finding it difficult to review the unixfilesystem resource changes due to the clang-format changes on the whole file (not your fault, just how things shook out). Maybe we can do an "offline" review of that bit wherein you walk us through the salient changes. Thoughts?

scripts/irods/test/resource_suite.py
```diff
@@ -1167,3 +1167,451 @@ def test_itrim_no_op(self):
...
+            # create file to register
+            file1 = "f1"
+            registered_path = '/%s/home/%s/%s/%s' % (self.admin.zone_name, self.admin.username, self.admin._session_id, file1)
+            vault_path = self.admin.get_vault_path(resource=resource_name)
+            os.mkdir(vault_path)
+            filepath = '%s/%s' % (vault_path, file1)
+            with open(filepath, 'w+') as f:
+                print('I AM A TESTFILE -- [' + self.testfile + ']', file=f, end='')
+
+            self.admin.assert_icommand("ireg -R %s %s %s" % (resource_name, filepath, registered_path))
```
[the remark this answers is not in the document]
trel: in detached mode, where the host is in the host_list... yes.

scripts/irods/paths.py
```diff
@@ -535,3 +535,16 @@ def possible_shm_locations():
...
     return _possible_shm_locations_str_cache
+
+# This is a common mount point used by topology tests when testing detached mode
+# behavior for unixfilesystem resources.  In detached mode the servers are expected
+# to have a common mount where the vault exists.
+_test_mount_directory_cache = None
+def _test_mount_directory():
+    global _test_mount_directory_cache
+    if _test_mount_directory_cache is None:
+        _test_mount_directory_cache = '/irods_testing_environment_mount_dir'
```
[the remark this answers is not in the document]
JustinKyleJames: done

unit_tests/src/test_host_list_context_string.cpp
```diff
@@ -0,0 +1,70 @@
...
+        prop_map.set<std::string>("host_list", "host2,host3,MYHOST");
+        REQUIRE(is_host_in_host_list(prop_map, resource_hostname));
+    }
+
+    SECTION("in middle of host list not case match")
+    {
+        irods::plugin_property_map prop_map;
+        prop_map.set<std::string>("host_list", "host2,MYHOST,host3");
+        REQUIRE(is_host_in_host_list(prop_map, resource_hostname));
+    }
```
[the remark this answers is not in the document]
JustinKyleJames: done

scripts/irods/test/resource_suite.py
```diff
@@ -1167,3 +1167,451 @@ def test_itrim_no_op(self):
...
+            # cleanup
+            lib.remove_resource(resource_name, self.admin)
+
+    def test_detached_mode_register_outside_vault__issue_4421(self):
+        try:
+            file1 = "f1"
+
+            # create file to put
+            filepath = lib.create_local_testfile(file1)
+            registered_path = '/%s/home/%s/%s/%s' % (self.admin.zone_name, self.admin.username, self.admin._session_id, file1)
```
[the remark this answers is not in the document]
JustinKyleJames: done

scripts/irods/test/resource_suite.py
```diff
@@ -1167,3 +1167,451 @@ def test_itrim_no_op(self):
...
+            resource_name = 'detached_resource'
+            resource_context = 'host_mode=detached'
+
+            lib.create_ufs_resource(resource_name, self.admin, resource_host)
+            self.admin.assert_icommand("iadmin modresc %s context %s" %
+                                   (resource_name, resource_context))
+
+            # create file to register
+            file1 = "f1"
+            registered_path = '/%s/home/%s/%s/%s' % (self.admin.zone_name, self.admin.username, self.admin._session_id, file1)
```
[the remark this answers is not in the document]
JustinKyleJames: done

scripts/irods/test/resource_suite.py
```diff
@@ -1167,3 +1167,451 @@ def test_itrim_no_op(self):
...
+            resource_name = 'detached_resource'
+            resource_context = 'host_mode=detached'
+
+            lib.create_ufs_resource(resource_name, self.admin, resource_host)
+            self.admin.assert_icommand("iadmin modresc %s context %s" %
+                                   (resource_name, resource_context))
+
+            self.admin.assert_icommand("ireg -R %s %s %s" % (resource_name, filepath, registered_path),
+                    'STDOUT_SINGLELINE',
+                    "Warning:  The registration was performed on a unixfilesystem resource [%s] currently configured as 'host_mode=detached'."
```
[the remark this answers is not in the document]
JustinKyleJames: Possibly, depending on if the registered path is a common mount point.

SwooshyCueb (approved):

plugins/resources/unixfilesystem/src/unixfilesystem.cpp
```diff
@@ -280,7 +367,29 @@ irods::error unix_file_mkdir_r(const std::string& path, mode_t mode)
...
 {
-    // NOOP
+    if (is_operating_in_detached_mode(_ctx.prop_map())) {
+        irods::file_object_ptr fco = boost::dynamic_pointer_cast<irods::file_object>(_ctx.fco());
+        std::string file_path = fco->physical_path();
+
+        std::string vault_path;
+        _ctx.prop_map().get<std::string>(irods::RESOURCE_PATH, vault_path);
+
+        if (!file_path.starts_with(vault_path)) {
```
korydraughn, line 377: This can lead to incorrect results. Prefer `has_prefix()` when checking paths for prefixes.
https://github.com/irods/irods/blob/02b03c67ff428959a758a17fdbcbd25001b4e39d/lib/core/include/irods/rodsPath.h#L77-L90

plugins/resources/unixfilesystem/src/unixfilesystem.cpp
```diff
@@ -83,7 +86,91 @@
...
+
+    // check if we are in the host list
+    char local_hostname[MAX_NAME_LEN];
+    gethostname(local_hostname, MAX_NAME_LEN);
+
+    auto resource_hostname_optional = resolve_hostname(local_hostname, hostname_resolution_scheme::match_preferred);
+
+    std::string resource_hostname =
+        resource_hostname_optional ? std::move(*resource_hostname_optional) : local_hostname;
+
```
korydraughn, line 126: Remove line.

plugins/resources/unixfilesystem/src/irods_is_in_host_list.cpp
```diff
@@ -0,0 +1,44 @@
...
+        irods::string_tokenize(host_list_str, delimiter, tokens);
+
+        for (const std::string& token : tokens) {
+            if (boost::iequals(token, resource_hostname)) {
+                return true;
+            }
+        }
+        return false;
+    } // is_host_in_host_list
+
```
korydraughn, line 43: Remove empty line.

plugins/resources/unixfilesystem/src/irods_is_in_host_list.cpp
```diff
@@ -0,0 +1,44 @@
...
+// stl includes
+#include <vector>
+
+namespace irods
+{
+    const std::string HOST_LIST("host_list");
+
+    // returns true if there is no host_list or if the current host is in
+    // the list
+    auto is_host_in_host_list(irods::plugin_property_map& prop_map, const std::string& resource_hostname) -> bool
```
korydraughn, line 18: The data type of `resource_hostname` can be a `const std::string_view`.

korydraughn (changes requested):

plugins/resources/unixfilesystem/src/unixfilesystem.cpp
JustinKyleJames: If you want I can remove every empty line in the entire file.

plugins/resources/unixfilesystem/src/unixfilesystem.cpp
JustinKyleJames: I did that because that is what Codacy said to do.  I'll just ignore Codacy from now on.

plugins/resources/unixfilesystem/src/unixfilesystem.cpp
trel: oh dear.

plugins/resources/unixfilesystem/src/irods_is_in_host_list.cpp
JustinKyleJames: I like that empty line.  Micromanage much?

plugins/resources/unixfilesystem/src/unixfilesystem.cpp
JustinKyleJames: done

plugins/resources/unixfilesystem/src/irods_is_in_host_list.cpp
JustinKyleJames: Looks horrible but done.

plugins/resources/unixfilesystem/src/irods_is_in_host_list.cpp
JustinKyleJames: done

plugins/resources/unixfilesystem/src/irods_is_in_host_list.cpp
JustinKyleJames: done

plugins/resources/unixfilesystem/src/unixfilesystem.cpp
JustinKyleJames: I get what you are saying but I figured the paths would have been normalized before this point.  Change has been made.

korydraughn:
Other than clang-format and clang-tidy, is this PR complete? Has it been tested?

trel:
review has happened - changes made - a claim of tested... i think it can be squashed, but not yet #.

korydraughn:
I agree.

unit_tests/src/test_host_list_context_string.cpp
```diff
@@ -0,0 +1,83 @@
...
+#include "irods/filesystem/path.hpp"
+#include "irods/private/irods_is_in_host_list.hpp"
+
+TEST_CASE("host list tests", "[detached_mode_host_list]")
+{
+    const std::string_view resource_hostname = "myhost";
+
+    SECTION("not in host list")
+    {
+        irods::plugin_property_map prop_map;
```
alanking, line 17: Optional: Instantiating the property map could happen at the top of the `TEST_CASE` since each `SECTION` is instantiating and using it.

alanking (approved):
Couple of optional improvements. I think we should merge irods/irods_testing_environment#111 as it is required for testing after this change.

plugins/resources/unixfilesystem/src/unixfilesystem.cpp
```diff
@@ -280,7 +367,29 @@ irods::error unix_file_mkdir_r(const std::string& path, mode_t mode)
 /// @brief interface to notify of a file registration
 irods::error unix_file_registered(irods::plugin_context& _ctx)
 {
-    // NOOP
+    if (is_operating_in_detached_mode(_ctx.prop_map())) {
```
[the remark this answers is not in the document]
JustinKyleJames: done

unit_tests/src/test_host_list_context_string.cpp
JustinKyleJames: done

alanking (approved):
Nothing else from me. Any last thoughts, @korydraughn / @trel / @SwooshyCueb?

trel (approved):

alanking:
I think this is good to go!

korydraughn (approved):

SwooshyCueb:
LGTM

PR #6530, opened 2022-08-12
[6529] Remote configuration reload

epsilon-phase:
Adding support to trigging the configuration reload to
the control plane enables easier use in clusters or
multiple instance setups.

The current command doesn't support any options other than
the standard ones, and thus is invoked like
"irods-grid reload"

### DIFF
--- a/scripts/irods/test/test_configuration_reload.py
+++ b/scripts/irods/test/test_configuration_reload.py
@@ -83,3 +83,35 @@ class TestConfigurationReload(SessionsMixin, unittest.TestCase):
                 IrodsController().restart(test_mode=True)
                 self.admin.assert_icommand(
                     ['irule', rule, 'null', 'ruleExecOut'], "STDOUT", [str(i)])
+
+    def test_reload_command__issue_6529(self):
+        server_config_filename = paths.server_config_path()
+        PROP_NAME = "is_issue_6529_done"
+        with lib.file_backed_up(server_config_filename):
+            # Make the configuration changes
+            with open(server_config_filename) as f:
+                svr_cfg = json.load(f)
+            svr_cfg[PROP_NAME] = "Yes?"
+            with open(server_config_filename, 'w') as f:
+                f.write(json.dumps(svr_cfg, sort_keys=True,
+                        indent=4, separators=(',', ': ')))
+
+            # Invoke irods-grid reload --all
+            _, out, _ = self.admin.assert_icommand(
+                "irods-grid reload --all", 'STDOUT_MULTILINE')
+
+            # Load the output of the command
+            obj = json.loads(out)
+            changes = obj["hosts"][0]["configuration_changes_made"]
+
+            # And check the properties such that they are what we expect
+            self.assertTrue(len(changes) == 1)
+            self.assertTrue(changes[0]["path"] == f"/{PROP_NAME}")
+            self.assertTrue(changes[0]["op"] == "add")
+            self.assertTrue(changes[0]["value"] == "Yes?")
+
+            # Also make sure that the actual value is correct, as far as the
+            # rule engine is concerned.
+            rule = self.prepare_rule(PROP_NAME)
+            self.admin.assert_icommand(
+                ["irule", rule, 'null', 'ruleExecOut'], 'STDOUT', ["Yes?"])
--- a/server/main_server/src/main_server_control_plane.cpp
+++ b/server/main_server/src/main_server_control_plane.cpp
@@ -8,6 +8,7 @@
 #include "irods/irods_server_properties.hpp"
 #include "irods/irods_server_state.hpp"
 #include "irods/irods_stacktrace.hpp"
+#include "irods/json_events.hpp"
 #include "irods/miscServerFunct.hpp"
 #include "irods/rcMisc.h"
 #include "irods/rodsServer.hpp"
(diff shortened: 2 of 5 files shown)


korydraughn:
You can use the testing environment to launch clusters of iRODS servers. That will allow you to verify the behavior of this code change.

`irods-grid` has options that allow you to target a specific set of hosts or all hosts. We need to verify that those options are honored when used.

server/main_server/src/main_server_control_plane.cpp
```diff
@@ -567,9 +567,7 @@ namespace irods
 
     void server_control_executor::operator()()
     {
-        signal( SIGINT, ctrl_plane_handle_shutdown_signal );
-        // SIGHUP is the signal to reload the configuration. This shouldn't interact with it
-        signal(SIGHUP, SIG_IGN);
+        signal(SIGINT, ctrl_plane_handle_shutdown_signal);
         signal( SIGTERM, ctrl_plane_handle_shutdown_signal );
```
korydraughn, line 571: Remove the outer spaces around the arguments so that things line up with L570.

scripts/irods/test/test_configuration_reload.py
```diff
@@ -83,3 +84,27 @@ def test_how_long_restarting_takes(self):
                 IrodsController().restart(test_mode=True)
                 self.admin.assert_icommand(
                     ['irule', rule, 'null', 'ruleExecOut'], "STDOUT", [str(i)])
+
+    def test_reload_command(self):
```
trel, line 88: please name test as 'test_reload_command__issue_6529"

scripts/irods/test/test_configuration_reload.py
```diff
@@ -83,3 +84,27 @@ def test_how_long_restarting_takes(self):
...
+                    "irods-grid reload --all", 'STDOUT_MULTILINE')
+                obj = json.loads(out)
+                changes = obj["hosts"][0]["configuration_changes_made"]
+                self.assertTrue(len(changes) == 1)
+                self.assertTrue(changes[0]["path"] == f"/{PROP_NAME}")
+                self.assertTrue(changes[0]["op"] == "add")
+                self.assertTrue(changes[0]["value"] == "No?")
+            rule = self.prepare_rule(PROP_NAME)
+            self.admin.assert_icommand(
+                ["irule", rule, 'null', 'ruleExecOut'], 'STDOUT', ["No?"])
```
korydraughn, line 110: Let's add some empty lines and comments so that it is easy to follow.

scripts/irods/test/test_control_plane.py
```diff
@@ -84,6 +84,27 @@ def test_shutdown_local_server(self):
             finally:
                 IrodsController().start()
 
+    def test_reload_command(self):
```
trel, line 87: is this test... identical to the one in test_configuration_reload.py?

trel (commented):
consider renaming 'replace' with 'update'?   i don't see that code in this PR though...

korydraughn (changes requested):
Awesome catch on the call to `config_props_.patch`.

Try to think of ways in which this could break. We need to make sure we test the durability of the reload functionality.

korydraughn:
> consider renaming 'replace' with 'update'? i don't see that code in this PR though...

Where did you see `replace` appear?

trel:
In the manual demo - an entry that was changed in the config appeared in the output as ‘replace’ I think…

epsilon-phase:
add/replace/remove is defined by the json patch specification I think, so it'd be unwise to change that.

korydraughn:
Correct. Those words are coming from the nlohmann JSON library.

trel:
ooooh, i see.    carry on.

scripts/irods/test/test_control_plane.py
epsilon-phase: Not exactly, it tests fewer things, so it should probably just live over there now

scripts/irods/test/test_configuration_reload.py
```diff
@@ -83,3 +84,37 @@ def test_how_long_restarting_takes(self):
...
+            with open(server_config_filename) as f:
+                svr_cfg = json.load(f)
+            svr_cfg[PROP_NAME] = "No?"
+            with open(server_config_filename, 'w') as f:
+                f.write(json.dumps(svr_cfg, sort_keys=True,
+                        indent=4, separators=(',', ': ')))
+
+            # Invoke irods-grid reload --all
+            with session.make_session_for_existing_admin() as admin_session:
+                admin_session.environment_file_contents = IrodsConfig().client_environment
```
alanking, line 102: I'm not sure that I understand the purpose of this line. Did we modify the client environment?

scripts/irods/test/test_configuration_reload.py
```diff
@@ -83,3 +84,37 @@ def test_how_long_restarting_takes(self):
...
                     ['irule', rule, 'null', 'ruleExecOut'], "STDOUT", [str(i)])
+
+    def test_reload_command__issue_6529(self):
+        server_config_filename = paths.server_config_path()
+        PROP_NAME = "is_issue_6529_done"
+        with lib.file_backed_up(server_config_filename):
+            # Make the configuration changes
+            with open(server_config_filename) as f:
+                svr_cfg = json.load(f)
+            svr_cfg[PROP_NAME] = "No?"
```
alanking, line 95: "No?" -> "Yes?" :p

scripts/irods/test/test_configuration_reload.py
```diff
@@ -83,3 +84,37 @@ def test_how_long_restarting_takes(self):
...
+                obj = json.loads(out)
+                changes = obj["hosts"][0]["configuration_changes_made"]
+
+                # And check the properties such that they are what we expect
+                self.assertTrue(len(changes) == 1)
+                self.assertTrue(changes[0]["path"] == f"/{PROP_NAME}")
+                self.assertTrue(changes[0]["op"] == "add")
+                self.assertTrue(changes[0]["value"] == "No?")
+
+            # Also make sure that the actual
```
alanking, line 117: The suspense is killing me...

alanking (commented):
Looks good! Just one real comment

alanking (approved):
Looks good to me. @trel?

trel:
clang-tidy has a couple opinions that look real - should those be addressed here?

otherwise, please # it.

epsilon-phase:
I addressed those concerns as best as I can, the missing file it complains seems like it probably isn't correct about, since that wouldn't compile otherwise

trel:
I think we did it!

alanking:
I agree. clang-tidy is a work in progress from a configuration perspective so I think the remaining issues are invalid. Let's # it.

alanking:
I see that this has already been done. Merging

Issue #6538, opened 2022-08-18
Can't link using irods_resource_plugin-unixfilesystem


JustinKyleJames:
- [x] main
- [x] 4-3-stable
---
When linking my unit tests using irods_resource_plugin-unixfilesystem in the IRODS_TEST_LINK_LIBRARIES, I get the following error:

```
CMake Error at unit_tests/CMakeLists.txt:87 (target_link_libraries):
  Target "irods_resource_plugin-unixfilesystem" of type MODULE_LIBRARY may
  not be linked into another target.  One may link only to INTERFACE, OBJECT,
  STATIC or SHARED libraries, or to executables with the ENABLE_EXPORTS
  property set.
```

Markus asked me to write this up.

Here is my full configuration:

```
set(IRODS_TEST_TARGET host_list_context_string)

set(IRODS_TEST_SOURCE_FILES ${CMAKE_CURRENT_SOURCE_DIR}/src/main.cpp
                            ${CMAKE_CURRENT_SOURCE_DIR}/src/test_host_list_context_string.cpp)

set(IRODS_TEST_INCLUDE_PATH ${IRODS_EXTERNALS_FULLPATH_BOOST}/include
                            ${IRODS_EXTERNALS_FULLPATH_FMT}/include)

set(IRODS_TEST_LINK_LIBRARIES irods_common
                              irods_client
                              irods_plugin_dependencies
                              irods_resource_plugin-unixfilesystem
                              #/${IRODS_PLUGINS_DIRECTORY}/resources/libunixfilesystem.so
                              ${IRODS_EXTERNALS_FULLPATH_FMT}/lib/libfmt.so)
```

alanking:
I believe this task is done, but awaiting confirmation from @SwooshyCueb

SwooshyCueb:
will close once 6525 is merged

alanking:
#6525 is merged

korydraughn:
This should be part of the 4.3.1 milestone now.

korydraughn:
Has the work been done for this? Are we ready for checkboxes?

SwooshyCueb:
Yeah, this can be ticked and closed

PR #6539, opened 2022-08-18
[#6538] Provide mechanism for linking object libraries to unit tests (main)

SwooshyCueb:
This needs to be merged before #6525.

Plugins are built as module library targets. This tells CMake that they are not meant to be installed or used in the same way as shared library targets. This is what we want. However, this presents a hurdle when it comes to unit testing code in our plugins, as module library targets cannot be linked against in CMake.

What we must do instead is split out the code that is to be tested into an object library target, and link the unit test against *that* instead. This PR seeks to make this trivially possible, by adding a `IRODS_TEST_LINK_OBJLIBRARIES` CMake variable that can be defined with a list of object library targets to be linked against.


trel:
similar/related/identical to #6538

alanking (approved):
Seems good to me.

I think the nice writeup in the PR message would go nicely in the commit message as well. What do you think?

Any thoughts @trel? Otherwise, I think this is good to go pending commit message discussion

SwooshyCueb:
I just went ahead and added the commit message. \#'d.

trel (approved):

PR #6541, opened 2022-08-19
Add nice stuff to 4.3 stable

epsilon-phase:
This contains:

* The original configuration reload work
* The `get_script_output_single_line` rewrite
* The `irods-grid reload` command

### DIFF
--- a/scripts/irods/test/test_configuration_reload.py
+++ b/scripts/irods/test/test_configuration_reload.py
@@ -1,0 +1,85 @@
+import unittest
+
+import json
+import os
+
+from . import session
+from ..controller import IrodsController
+from .. import lib
+from .. import paths
+
+SessionsMixin = session.make_sessions_mixin([('otherrods', 'rods')], [])
+
+
+class TestConfigurationReload(SessionsMixin, unittest.TestCase):
+    def setUp(self):
+        super(TestConfigurationReload, self).setUp()
+        self.admin = self.admin_sessions[0]
+        self.testdir = os.path.join(self.admin.session_collection, 'testdir')
+
+    @staticmethod
+    def prepare_rule(property_name):
+        """Prepare a query about the server settings."""
+        return f"""
+        msi_get_server_property("{property_name}",*thing);
+        writeLine("stdout","*thing");
+        """
+
+    # Test to see that it recognizes the  configuration's change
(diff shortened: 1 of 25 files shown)


scripts/irods/test/test_configuration_reload.py
```diff
@@ -97,18 +98,20 @@ def test_reload_command__issue_6529(self):
...
-            obj = json.loads(out)
-            changes = obj["hosts"][0]["configuration_changes_made"]
-
-            # And check the properties such that they are what we expect
-            self.assertTrue(len(changes) == 1)
-            self.assertTrue(changes[0]["path"] == f"/{PROP_NAME}")
-            self.assertTrue(changes[0]["op"] == "add")
-            self.assertTrue(changes[0]["value"] == "Yes?")
+            with session.make_session_for_existing_admin() as admin_session:
+                admin_session.environment_file_contents = IrodsConfig().client_environment
```
alanking, line 102: Still not sure that I understand the reason for this. Does this not work if we don't load the `IrodsConfig().client_environment` into the `admin_session.environment_file_contents`?

scripts/irods/test/test_configuration_reload.py
```diff
@@ -97,18 +98,20 @@ def test_reload_command__issue_6529(self):
...
+
+                # Load the output of the command
+                obj = json.loads(out)
+                changes = obj["hosts"][0]["configuration_changes_made"]
+
+                # And check the properties such that they are what we expect
+                self.assertTrue(len(changes) == 1)
+                self.assertTrue(changes[0]["path"] == f"/{PROP_NAME}")
+                self.assertTrue(changes[0]["op"] == "add")
+                self.assertTrue(changes[0]["value"] == "Yes?")
```
alanking, line 114: Since we're in here touching this again, these could be `assertEqual`s.

alanking (commented):
Cherry-picks look clean to me after eyeballing a diff with main. Had a couple comments on the test thing. Good catch on that.

scripts/irods/test/test_configuration_reload.py
epsilon-phase: Yeah, irods-grid doesn't work without this line

alanking (approved):

Issue #4367, opened 2019-05-02
Replace rodsLog calls with new logger


alanking:
Calls to `rodsLog` currently map to `irods::experimental::log::legacy` but in most instances they are in need of a proper category to fall in line with the new rsyslog system.

korydraughn:
Related to #8003.

PR #6549, opened 2022-08-23
[#4367] rodsLog cleanup (main)

epsilon-phase:
I haven't gotten everything, but at least there'll be a few less :)

### DIFF
--- a/lib/core/include/irods/irods_logger.hpp
+++ b/lib/core/include/irods/irods_logger.hpp
@@ -53,6 +53,7 @@ namespace irods::experimental
 
         struct category
         {
+            // clang-format off
             struct legacy {};
             struct server {};
             struct agent_factory {};
@@ -65,6 +66,8 @@ namespace irods::experimental
             struct microservice {};
             struct network {};
             struct rule_engine {};
+            struct sql {};
+            // clang-format on
         }; // struct category
 
         template <typename Category> class logger_config;
@@ -83,6 +86,7 @@ namespace irods::experimental
         using microservice   = logger<category::microservice>;
         using network        = logger<category::network>;
         using rule_engine    = logger<category::rule_engine>;
+        using sql            = logger<category::sql>;
         // clang-format on
 
         log() = delete;
--- a/plugins/api/src/register_physical_path.cpp
+++ b/plugins/api/src/register_physical_path.cpp
@@ -21,6 +21,7 @@
 #include "irods/irods_at_scope_exit.hpp"
 #include "irods/irods_get_l1desc.hpp"
 #include "irods/irods_hierarchy_parser.hpp"
+#include "irods/irods_logger.hpp"
 #include "irods/irods_resource_backport.hpp"
 #include "irods/irods_resource_redirect.hpp"
 #include "irods/key_value_proxy.hpp"
@@ -54,22 +55,20 @@
 #include "irods/server_utilities.hpp"
 #include "irods/specColl.hpp"
 #include "irods/stringOpr.h"
-
 #define IRODS_FILESYSTEM_ENABLE_SERVER_SIDE_API
 #include "irods/filesystem.hpp"
-
 #define IRODS_REPLICA_ENABLE_SERVER_SIDE_API
 #include "irods/replica_proxy.hpp"
-
 #include <boost/lexical_cast.hpp>
 #include <fmt/format.h>
-
 #include <iostream>
 
 /* holds a struct that describes pathname match patterns
    to exclude from registration. Needs to be global due
    to the recursive dirPathReg() calls. */
 static pathnamePatterns_t *ExcludePatterns = NULL;
+
+using log_api = irods::experimental::log::api;
 
 namespace
 {
@@ -120,8 +119,8 @@ namespace
         status = rsFileStat( _comm, &fileStatInp, &stbuf );
         if ( status != 0 ) {
             if ( status != UNIX_FILE_STAT_ERR - ENOENT ) {
-                rodsLog( LOG_DEBUG, "readPathnamePatternsFromFile: can't stat %s. status = %d",
-                         fileStatInp.fileName, status );
+                log_api::debug(
+                    "readPathnamePatternsFromFile: can't stat {}. status = {}", fileStatInp.fileName, status);
             }
             return NULL;
         }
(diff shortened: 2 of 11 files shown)


korydraughn:
Ignore files such as `lsUtil.cpp`. Those are specific to the icommands and should not contain references to the logging API (because the logging API is not designed for client-side use).

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -428,7 +428,9 @@ int _addChildToResource(generalAdminInp_t* _generalAdminInp, rsComm_t* _rsComm)
 
     resc_input[irods::RESOURCE_CHILDREN] = rescChildren;
 
-    log_api::info("rsGeneralAdmin add child \"{}\" to resource \"{}\"", _generalAdminInp->arg3, _generalAdminInp->arg2);
+    log_api::info(R"stuff(rsGeneralAdmin add child "{}" to resource "{}")stuff",
```
trel, line 431: stuff?

epsilon-phase:
Raw string literals are funny huh? Does "delimiter()delimiter" work better?

-------- Original Message --------

korydraughn:
The delimiters are optional. If you prefer to use them, perhaps using 2 or more underscores is a good option since underscores tend to be less noisy.

lib/core/include/irods/irods_logger.hpp
```diff
@@ -43,6 +43,7 @@ namespace irods::experimental
 
         enum class level
         {
+            sql,
```
korydraughn, line 46: What is the motivation behind these changes to the logging API?

epsilon-phase: The sql loglevel is distinct in the old log api, I would guess because it's even more prone to filling the logs than trace.

korydraughn: The log levels are pretty standard, so we don't want to introduce one people are unfamiliar with.

A better approach would be to introduce a new category.

trel:
this is looking pretty good.   the main use case for the LOG_SQL thing was to see the bind variables and sequel strings being sent to the database - that should probably be debug rather than trace...  if you haven't already touched that area.

plugins/api/src/register_physical_path.cpp
```diff
@@ -21,56 +21,59 @@
...
-#define IRODS_REPLICA_ENABLE_SERVER_SIDE_API
-#include "irods/replica_proxy.hpp"
-
-#include <boost/lexical_cast.hpp>
-#include <fmt/format.h>
-
-#include <iostream>
+#  include "irods/irods_logger.hpp"
+#  include "irods/irods_resource_backport.hpp"
+#  include "irods/irods_resource_redirect.hpp"
```
korydraughn, line 26: Let's revert the formatting on all these include directives.

Right now, they look weird in relation to the include directives that weren't touched.

server/api/src/rsDataObjRename.cpp
```diff
@@ -47,8 +47,6 @@
 #include <cstring>
 #include <chrono>
 
-using logger = irods::experimental::log;
```
korydraughn, line 50: Shouldn't this be `using log_api = irods::experimental::log::api;` instead of being removed?

server/api/src/rsGetRescQuota.cpp
```diff
@@ -12,6 +13,8 @@
 
 #include <cstring>
 
+using log_server = irods::experimental::log::server;
```
korydraughn, line 16: Why the `server` category instead of the `api` category?

korydraughn (changes requested):
Excellent! This is going to make dealing with the log so much better.

server/api/src/rsDataObjRename.cpp
epsilon-phase: There was a name clash that developed for some reason. I'll change the style over :)

server/api/src/rsDataObjRename.cpp
```diff
@@ -47,7 +47,7 @@
...
+using log_api = log_api;
```
korydraughn, line 50: I see now. Some other header is defining this alias, right?

Let's explicitly define the alias so there's no dependency on a header providing it.
```cpp
```

korydraughn:
If the changes compile for each OS (under the github actions) then squash the commits and leave off the pound.

Have you verified that the core tests pass with these changes?

alanking (commented):
Looks real good. Agreed with @korydraughn's comment regarding tests. If this is building and passing tests then I think it's pretty much there. Any thoughts from others on keeping the addition of the new sql category in its own commit separate from the `rodsLog` replacements?

Along those lines, we should probably make an issue to update our documentation with respect to `spLogSql` since it will no longer be the way to enable SQL logging. Also, the new category should be represented in the docs anyway.

korydraughn:
> Any thoughts from others on keeping the addition of the new sql category in its own commit separate from the rodsLog replacements?

That's not a bad idea. Let's do that.

> Along those lines, we should probably make an issue to update our documentation with respect to `spLogSql` since it will no longer be the way to enable SQL logging. Also, the new category should be represented in the docs anyway.

Yes, we need an issue to update the docs, but I'm not sure this PR obsoletes `spLogSql` just yet. I didn't see anything in there that removed the need for that option. I only saw changes regarding the operations (e.g. `log_sql::debug("chlSomeOperation() 1")`). Did you spot the code that removes the need for that option? If so, I'm sure some tests are going to fail.

trel:
Two eventual commits seems good (one for new sql category, one for changing all the things).

alanking:
> Did you spot the code that removes the need for that option?

We added a new category for SQL logging and then replaced many calls, but it turns out that we haven't replaced all of the calls in this PR (and that's okay). Once all calls to `rodsLog(LOG_SQL, ...)` and `rodsLogSql` are replaced, the option will not be useful anymore. It only exists to allow for `LOG_SQL` to have an effect in `rodsLog` because `LOG_SQL` is level 11 (the highest available log level). Was just mentioning that we should probably update the documentation once this has happened, and maybe that's just part of #4367.

korydraughn:
Let's capture the documentation effort for `spLogSql` in a new issue.

@epsilon-phase Please use the issue number 4367 for the sql category commit as well.

alanking (approved):

Issue #6591, opened 2022-09-13
Allow identity of user attached to connection/agent to change in real time


korydraughn:
- [x] main
- [x] 4-3-stable
---
## Feature
iRODS connections tie the identity of a user to the socket. This is fine for one-off commands, but not for situations where there can be hundreds to thousands of concurrent users. Creating a new connection for every user will quickly drain resources.

Therefore, iRODS should provide a way to change the user identity tied to the connection object. This would lead to huge improvements regarding performance, scalability, and resource management.

This would also improve support for client applications because the client libraries would finally be able to implement real connection pooling for iRODS connections.

### Requirements
The only API requirement is ... switching the user tied to the connection object requires **rodsadmin** level privileges.

### Challenges
The primary challenge with implementing this has to do with long running agents.

When an agent is forked, depending on the operation being carried out, the agent may choose to use the catalog or local memory. If information is fetched from local memory, then we run the risk of using out-of-date information. This situation applies to resource hierarchies, rule engine plugins, and likely other things.

So, to support this feature means iRODS must guarantee that long running agents see important changes. The question now becomes, when do iRODS agents sync with the catalog and how?

### Other
Add support to the following libraries:
- client_connection.hpp
- connection_pool.hpp

---
Related issues: #5060

korydraughn:
See PR #6593

PR #6593, opened 2022-09-13
[#6591] Implemented new API plugin: rc_switch_user (main)

korydraughn:
The name of the plugin is not final.


lib/api/src/rc_switch_user.cpp
```diff
@@ -0,0 +1,71 @@
...
+    // Not doing this is good for benchmarking.
+    //
+    // Should we do this check at all?
+    // How do we know the RcComm and RsComm represent the same user?
+    //
+    //if (std::strcmp(_username, static_cast<const char*>(client.userName)) == 0 &&
+    //    std::strcmp(_zone, static_cast<const char*>(client.rodsZone)) == 0)
+    //{
+    //    return 0;
+    //}
```
korydraughn, line 31: Add comments explaining why we shouldn't check this.

lib/api/src/rc_switch_user.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+    // TODO
+    // Does the .irodsA file matter at all?
```
korydraughn, line 34: Investigate whether this matters.

plugins/api/include/irods/plugins/api/switch_user_types.h
```diff
@@ -0,0 +1,27 @@
+#ifndef IRODS_SWITCH_USER_TYPES_H
+#define IRODS_SWITCH_USER_TYPES_H
+
+typedef struct SwitchUserInput // NOLINT(modernize-use-using)
+{
+    // NOLINTBEGIN(modernize-avoid-c-arrays)
+    char username[64];
+    char zone[64];
+    // NOLINTEND(modernize-avoid-c-arrays)
+} switchUserInp_t;
```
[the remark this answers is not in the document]
korydraughn: Until we know there's a need for a key value object, we shouldn't add one.

Adding it to the API can always be done at a later time.

lib/api/src/rc_switch_user.cpp
korydraughn: The .irodsA file only matters when authenticating with the server. This API plugin is designed to be used only after authentication.

lib/api/src/rc_switch_user.cpp
korydraughn: done

lib/api/src/rc_switch_user.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+        // NOLINTNEXTLINE(readability-implicit-bool-conversion)
+        if (ec == 0) {
+            // On success, we always assume that the output pointer is valid.
+            // If the server-side implementation is solid, then we shouldn't need to check the pointer.
+
+            // NOLINTBEGIN(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+            std::strncpy(client.userName, _username, sizeof(UserInfo::userName));
+            std::strncpy(client.rodsZone, _zone, sizeof(UserInfo::rodsZone));
+            std::strncpy(client.userType, output->user_type, sizeof(UserInfo::userType));
```
korydraughn, line 42: Investigate whether the client connection requires `userType`.

If it isn't set following `rcConnect()` or `clientLogin()` then it is probably unused on the client-side. That also means we CAN short-ciruit this API call.

plugins/api/src/switch_user.cpp
```diff
@@ -0,0 +1,184 @@
...
+#  include "irods/irods_rs_comm_query.hpp"
+#  include "irods/rodsErrorTable.h"
+#  include "irods/irods_logger.hpp"
+#  include "irods/server_utilities.hpp"
+
+#  define IRODS_USER_ADMINISTRATION_ENABLE_SERVER_SIDE_API
+#  include "irods/user_administration.hpp"
+
+#  include <cstring>
+#  include <string>
```
korydraughn, line 28: Missing `<string_view>`

lib/core/include/irods/library_support.h
```diff
@@ -0,0 +1,15 @@
...
+
+/// \file
+/// \brief iRODS Feature Test Macros
+///
+/// This file defines macros that can be used to test whether one or more features
+/// are supported by the iRODS library.
+///
+/// \since 4.3.1
+
+#define IRODS_HAS_API_ENDPOINT_SWITCH_USER 4003001L
```
SwooshyCueb, line 12: Should be date-based as discussed during standup

SwooshyCueb (changes requested):

lib/core/include/irods/library_support.h
korydraughn: Done.

plugins/api/src/switch_user.cpp
korydraughn: Done

lib/api/src/rc_switch_user.cpp
korydraughn: Done, but we need to be sure we are happy with the changes.

lib/api/src/rc_switch_user.cpp
```diff
@@ -11,25 +11,22 @@
...
+    if (!_comm || !_username || !_zone) {
         return SYS_INVALID_INPUT_PARAM;
     }
 
     auto& client = _comm->clientUser;
 
-    // Attempting to verify that the user is already the target user cannot be done without
-    // going over the wire. That's because the client cannot determine the user type of the
-    // target user. For that reason, this implementation always results in a network request.
+    if (std::strcmp(_username, client.userName) == 0 && std::strcmp(_zone, client.rodsZone) == 0) {
```
trel, line 20: please add a comment about why this is allowed / good / efficient.

plugins/api/src/switch_user.cpp
```diff
@@ -0,0 +1,170 @@
...
+        }
+
+        const auto* username = _input->username;
+        const auto* zone = _input->zone;
+        bool is_local_zone = false;
+
+        // Passing an empty string as the zone argument is equivalent to explicitly setting
+        // the zone parameter to the local zone.
+        if (const std::string_view z{_input->zone}; z.empty() || z == local_zone) {
+            zone = local_zone;
```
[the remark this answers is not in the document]
korydraughn: I'm thinking of changing the API so that it forces users to pass a non-empty zone.

The optimization where we check the username and zone to avoid unnecessary network interaction assumes the client app always passes a non-empty zone, yet the API plugin allows clients to pass empty zone strings.

The issue is that the rc_switch_user function copies the client-provided zone argument into the zone member variable of the RcComm following success. If the zone argument was empty, the client will now be in a situation.

Therefore, we can't accept empty zone names.

Thoughts?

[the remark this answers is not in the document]
korydraughn: The more I think about this, the more I feel requiring the client to pass a zone is the correct thing to do.

If we don't require the zone, clients can simply provide a username and let the server become whoever it wants.

lib/api/src/rc_switch_user.cpp
korydraughn: Done

plugins/api/src/switch_user.cpp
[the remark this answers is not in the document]
trel: this seems very reasonable.   there is no legacy code to worry about / support.

no way to hold it wrong is a good thing.

lib/administration/user/CMakeLists.txt
```diff
@@ -52,6 +54,7 @@ foreach(variant IN ITEMS client server)
     "$<BUILD_INTERFACE:${CMAKE_IRODS_BINARY_DIR}/lib/core/include>"
     "$<BUILD_INTERFACE:${CMAKE_IRODS_SOURCE_DIR}/lib/core/include>"
     "$<BUILD_INTERFACE:${CMAKE_IRODS_SOURCE_DIR}/lib/api/include>"
+    #"$<BUILD_INTERFACE:${CMAKE_IRODS_SOURCE_DIR}/lib/administration/zone/include>"
```
trel, line 57: TODO/remove

lib/administration/user/include/irods/user_administration.hpp
```diff
@@ -29,58 +30,68 @@
...
+        /// \return A user_type enumeration.
+        ///
+        /// \since 4.3.1
+        auto to_user_type(const std::string_view user_type_string) -> user_type;
+
+        /// \brief Converts a user_type enumeration to a string.
+        ///
+        /// \param[in] user_type The enumeration to convert.
+        ///
+        /// \throws user_management_error If the string cannot be converted to an enumeration.
```
trel, line 74: flip it and reverse it

lib/administration/zone/include/irods/zone_administration.hpp
```diff
@@ -0,0 +1,289 @@
...
+        /// Holds the ACL and entity to attach to a zone. Primarily used to modify a zone.
+        ///
+        /// \since 4.3.1
+        struct zone_collection_acl_property
+        {
+            // clang-format off
+            zone_collection_acl acl; ///< The permission for a zone collection.
+            std::string name;        ///< The name of a user or group.
+            // clang-format on
+        }; // struct comment_property
```
trel, line 95: zone_collection_acl_property

lib/api/src/rc_switch_user.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+    auto& client = _comm->clientUser;
+
+    // clang-format off
+    // NOLINTBEGIN(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+    //
+    // Return immediately if the RcComm object already represents the target user.
+    //
+    // This is allowed because the RcComm only cares about the username and zone. If the RcComm
+    // considered the user's type or authentication level (i.e. LOCAL_PRIV_USER_AUTH), then the
```
trel, line 34: ", then the" -> ", then"

lib/api/src/rc_switch_user.cpp
```diff
@@ -0,0 +1,71 @@
...
+    // this optimitization would not be possible.
```
trel, line 35: "optimitization" -> "optimization"

unit_tests/src/test_zone_administration.cpp
```diff
@@ -0,0 +1,126 @@
...
+    CHECK_NOTHROW(adm::client::modify_zone(conn, test_zone_name_2, adm::zone_name_property{test_zone_name_1}));
+    CHECK(adm::client::zone_exists(conn, test_zone_name_1));
+
+    // Show that the connection information can be adjusted.
+    auto zone_info = adm::client::zone_info(conn, test_zone_name_1);
+    REQUIRE(zone_info.has_value());
+    CHECK(zone_info->type == adm::zone_type::remote);
+    CHECK(zone_info->name == test_zone_name_1);
+
+    const auto* const conn_info_2 = "unit-test.irods.org:1247";
```
trel, line 94: let's use example.org rather than irods.org for tests/documentation... unless we have a strong reason not to.

trel (commented):
very nice.

lib/administration/user/CMakeLists.txt
korydraughn: done

lib/administration/user/include/irods/user_administration.hpp
korydraughn: done

lib/administration/zone/include/irods/zone_administration.hpp
korydraughn: done

lib/api/src/rc_switch_user.cpp
korydraughn: done

lib/api/src/rc_switch_user.cpp
korydraughn: done

unit_tests/src/test_zone_administration.cpp
korydraughn: done

lib/administration/user/CMakeLists.txt
```diff
@@ -17,6 +18,7 @@ target_include_directories(
   PUBLIC
   "$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>"
   "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
+  "$<BUILD_INTERFACE:${CMAKE_IRODS_SOURCE_DIR}/lib/administration/zone/include>"
```
korydraughn, line 21: This was added because of the externalization of the `zone_type` enumeration.

While this works, I wonder if this is the correct cmake change for this.

plugins/api/src/switch_user.cpp
```diff
@@ -0,0 +1,198 @@
...
+#  include "irods/rodsErrorTable.h"
+#  include "irods/irods_logger.hpp"
+#  include "irods/server_utilities.hpp"
+
+#  define IRODS_USER_ADMINISTRATION_ENABLE_SERVER_SIDE_API
+#  include "irods/user_administration.hpp"
+
+#  include <cstring>
+#  include <string>
+#  include <string_view>
```
korydraughn, line 29: Not a fan of the indentation here because this preprocessor block contains more than just preprocessor directives. If these directives were closer to the `#ifdef`, I'd be happier.

I wonder what others think about this?

korydraughn:
Here's something to think about ...

What happens if you have a long running agent that has performed redirection prior to a call to `rc_switch_user`?

We need to investigate that and see if the `RcComm` used for redirection needs to be updated. The API plugin may need to update all active connections used for redirection.

trel:
Wouldn't this ALWAYS happen before redirection?   Hmm...

korydraughn:
The client-to-server connection is fine. The potential issue has to do with how server-to-server connections are handled.

We need to determine if server-to-server connections are cached or torn down. I'd imagine that connections are cached for performance. If yes, then the `rc_switch_user` API plugin must update the server-to-server connections as well.

This may affect federation as well.

lib/administration/user/CMakeLists.txt
SwooshyCueb: This might need to be propagated from further up. I'll take a look later

plugins/api/src/switch_user.cpp
SwooshyCueb: Honestly I think I've decided that I don't like the way `clang-format` does preprocessor indentation at all. I wouldn't be opposed to switching it back off.

lib/administration/user/CMakeLists.txt
korydraughn: Sure thing.

plugins/api/src/switch_user.cpp
korydraughn: I haven't reached a decision about the formatting of PP directives just yet. Will think about it a bit more as this PR nears completion.

plugins/api/src/switch_user.cpp
```diff
@@ -73,6 +79,27 @@ namespace
         return 0;
     } // check_input
 
+    auto update_server_to_server_connections(RsComm& _comm) -> void
+    {
+        const auto update_zone_connection = [&client = _comm.clientUser](rodsServerHost& _host)
+        {
+            if (_host.conn) {
+                const auto ec = rc_switch_user(_host.conn, client.userName, client.rodsZone);
+                log_api::info("rc_switch_user ec = [{}]", ec);
```
trel, line 88: debug?

korydraughn:
Turns out that iRODS does not disconnect connections created as a result of redirection. The plugin now iterates over the connections (used for redirection) and invokes `rc_switch_user` on each one. This only occurs on the call to the API plugin.

plugins/api/src/switch_user.cpp
```diff
@@ -0,0 +1,234 @@
...
+            }
+
+            // iRODS agents do not disconnect from other nodes following a redirect. For long running agents,
+            // this means the API plugin must invoke rc_switch_user() on each connection.
+            //
+            // TODO What happens if this fails half-way through updating the connections?
+            // Should things be rolled back?
+            // Instead, should the server-to-server connections be refreshed instead of calling rc_switch_user()?
+            // Should iRODS agents hold on to connections after redirecting?
+            update_server_to_server_connections(*_comm);
```
trel, line 164: would this catch any errors from 'inside'?     yes, very interested in the test that triggers this kind of update (and possible failure)...

plugins/api/src/switch_user.cpp
```diff
@@ -0,0 +1,234 @@
...
+        return 0;
+    } // check_input
+
+    auto update_server_to_server_connections(RsComm& _comm) -> void
+    {
+        const auto update_zone_connection = [&client = _comm.clientUser](rodsServerHost& _host)
+        {
+            if (_host.conn) {
+                const auto ec = rc_switch_user(_host.conn, client.userName, client.rodsZone);
+                log_api::info("rc_switch_user ec = [{}]", ec);
```
korydraughn, line 88: See comments around the call to `update_server_to_server_connection`.

plugins/api/src/switch_user.cpp
korydraughn, line 164: Think about these questions.

korydraughn: As currently implemented, this does not guard against failures of `rc_switch_user` on connections created by a redirect.

If an error occurs at any point, the connections could be left in an invalid state. This invalid state can be fixed by calling `rc_switch_user` again.

I doubt we have any tests that trigger these cases as they require the client to keep the agents running. I've demonstrated this using the testing environment and a custom client that reuses the same connection.

.clang-tidy
```diff
@@ -32,7 +32,8 @@ Checks:
     portability-std-allocator-const,
     portability-simd-intrinsics,
     readability-*,
-    -readability-avoid-const-params-in-decls
+    -readability-avoid-const-params-in-decls,
+    -readability-function-cognitive-complexity
```
SwooshyCueb, line 36: I would prefer to leave this check enabled, but as a warning rather than an error

lib/administration/user/CMakeLists.txt
SwooshyCueb: This seems fine to me

plugins/api/src/switch_user.cpp
korydraughn: Current plan:
- If dealing with an older server (no `rc_switch_user` available), call `rcDisconnect`
- If dealing with a newer server and an error occurs, call `rcDisconnect`

.clang-tidy
korydraughn: I'm not sure this check offers much to us, but I'm willing to be convinced.

unit_tests/src/test_rc_switch_user.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+    // Become the test user.
+    auto* conn_ptr = static_cast<RcComm*>(conn);
+    REQUIRE(rc_switch_user(conn_ptr, alice.name.c_str(), alice.zone.c_str()) == 0);
+    // NOLINTBEGIN(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+    CHECK(conn_ptr->clientUser.userName == alice.name);
+    CHECK(conn_ptr->clientUser.rodsZone == alice.zone);
+    CHECK(std::strcmp(conn_ptr->clientUser.userType, adm::to_c_str(adm::user_type::rodsuser)) == 0);
+    // NOLINTEND(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+    CHECK(conn_ptr->clientUser.authInfo.authFlag == LOCAL_USER_AUTH);
```
korydraughn, line 132: No longer needed because the api plugin does not change this member variable.

unit_tests/src/test_rc_switch_user.cpp
```diff
@@ -0,0 +1,146 @@
...
+    // Give the administrator OWN permissions on the data object created by the test user.
+    // This allows the administrator to remove the sandbox collection without issues.
+    fs::client::permissions(conn, data_object, env.rodsUserName, fs::perms::own);
+
+    REQUIRE(rc_switch_user(conn_ptr, env.rodsUserName, env.rodsZone) == 0);
+    CHECK(std::strcmp(conn_ptr->clientUser.userName, env.rodsUserName) == 0);
+    CHECK(std::strcmp(conn_ptr->clientUser.rodsZone, env.rodsZone) == 0);
+    CHECK(std::strcmp(conn_ptr->clientUser.userType, adm::to_c_str(adm::user_type::rodsadmin)) == 0);
+    // NOLINTEND(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+    CHECK(conn_ptr->clientUser.authInfo.authFlag == LOCAL_PRIV_USER_AUTH);
```
korydraughn, line 87: No longer needed because the api plugin does not change this member variable.

unit_tests/src/test_rc_switch_user.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+    // Become the test user.
+    auto* conn_ptr = static_cast<RcComm*>(conn);
+    REQUIRE(rc_switch_user(conn_ptr, alice.name.c_str(), alice.zone.c_str()) == 0);
+    // NOLINTBEGIN(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+    CHECK(conn_ptr->clientUser.userName == alice.name);
+    CHECK(conn_ptr->clientUser.rodsZone == alice.zone);
+    CHECK(std::strcmp(conn_ptr->clientUser.userType, adm::to_c_str(adm::user_type::rodsuser)) == 0);
+    // NOLINTEND(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+    CHECK(conn_ptr->clientUser.authInfo.authFlag == LOCAL_USER_AUTH);
```
korydraughn, line 66: No longer needed because the api plugin does not change this member variable.

unit_tests/src/test_rc_switch_user.cpp
```diff
@@ -0,0 +1,146 @@
...
+    CHECK_FALSE(fs::client::exists(conn, sandbox));
+    CHECK_FALSE(fs::client::exists(conn, data_object));
+
+    // Become the administrator so that the test can clean up properly.
+    REQUIRE(rc_switch_user(conn_ptr, env.rodsUserName, env.rodsZone) == 0);
+    CHECK(std::strcmp(conn_ptr->clientUser.userName, env.rodsUserName) == 0);
+    CHECK(std::strcmp(conn_ptr->clientUser.rodsZone, env.rodsZone) == 0);
+    CHECK(std::strcmp(conn_ptr->clientUser.userType, adm::to_c_str(adm::user_type::rodsadmin)) == 0);
+    // NOLINTEND(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+    CHECK(conn_ptr->clientUser.authInfo.authFlag == LOCAL_PRIV_USER_AUTH);
```
korydraughn, line 144: No longer needed because the api plugin does not change this member variable.

plugins/api/src/switch_user.cpp
```diff
@@ -0,0 +1,234 @@
...
+        };
+
+        for (auto* zone_ptr = ZoneInfoHead; zone_ptr; zone_ptr = zone_ptr->next) {
+            for (auto* host_ptr = zone_ptr->primaryServerHost; host_ptr; host_ptr = host_ptr->next) {
+                update_zone_connection(*host_ptr);
+            }
+
+            for (auto* host_ptr = zone_ptr->secondaryServerHost; host_ptr; host_ptr = host_ptr->next) {
+                update_zone_connection(*host_ptr);
+            }
```
korydraughn, line 99: Consider what happens to the entries in `ZoneInfoHead` upon invoking `rcDisconnect` on them.

Perhaps, we don't have to do anything.

lib/core/include/irods/library_support.h
```diff
@@ -0,0 +1,23 @@
...
+
+/// Defined if the development library supports ::rc_switch_user.
+///
+/// \since 4.3.1
+#define IRODS_HAS_API_ENDPOINT_SWITCH_USER    202209L
+
+/// Defined if the development library supports the C++ zone administration library.
+///
+/// \since 4.3.1
+#define IRODS_HAS_LIBRARY_ZONE_ADMINISTRATION 202209L
```
korydraughn, line 20: Update dates.

lib/administration/zone/include/irods/zone_administration.hpp
```diff
@@ -0,0 +1,289 @@
...
+            /// \param[in] _property  An object containing the required information for applying the change.
+            ///
+            /// \throws irods::exception If an error occurs.
+            ///
+            /// \since 4.3.1
+            template <typename ZoneProperty>
+                requires std::same_as<ZoneProperty, zone_name_property> ||
+                    std::same_as<ZoneProperty, connection_info_property> ||
+                    std::same_as<ZoneProperty, comment_property> ||
+                    std::same_as<ZoneProperty, zone_collection_acl_property>
```
korydraughn, line 179: Should probably give this a name instead of inlining it here.

lib/administration/zone/include/irods/zone_administration.hpp
```diff
@@ -0,0 +1,289 @@
...
+        /// Holds various pieces of information that uniquely identify a zone.
+        ///
+        /// \since 4.3.1
+        struct zone_info
+        {
+            int id;
+            zone_type type;
+            std::string name;
+            std::string connection_info;
+            std::string comment;
```
korydraughn, line 106: Consider moving smaller data types to the bottom.

lib/administration/zone/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
...
+  IRODS_ZONE_ADMINISTRATION_ENABLE_SERVER_SIDE_API
+)
+
+install(
+  FILES
+  "${CMAKE_CURRENT_SOURCE_DIR}/include/irods/zone_administration.hpp"
+  "${CMAKE_CURRENT_SOURCE_DIR}/include/irods/zone_type.hpp"
+  DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/irods"
+  COMPONENT ${IRODS_PACKAGE_COMPONENT_DEVELOPMENT_NAME}
+)
```
korydraughn, line 58: Add new line for unix tools

lib/CMakeLists.txt
```diff
@@ -159,4 +162,4 @@ install(
   ARCHIVE
   DESTINATION "${CMAKE_INSTALL_LIBDIR}"
   COMPONENT ${IRODS_PACKAGE_COMPONENT_DEVELOPMENT_NAME}
-)
\ No newline at end of file
+)
```
korydraughn, line 165: Add new line for unix tools

unit_tests/cmake/test_config/irods_rc_switch_user.cmake
```diff
@@ -0,0 +1,10 @@
+set(IRODS_TEST_TARGET irods_rc_switch_user)
+
+set(IRODS_TEST_SOURCE_FILES ${CMAKE_CURRENT_SOURCE_DIR}/src/main.cpp
+                            ${CMAKE_CURRENT_SOURCE_DIR}/src/test_rc_switch_user.cpp)
+
+set(IRODS_TEST_INCLUDE_PATH ${IRODS_EXTERNALS_FULLPATH_BOOST}/include
+                            ${IRODS_EXTERNALS_FULLPATH_FMT}/include)
+ 
+set(IRODS_TEST_LINK_LIBRARIES irods_client
+                              ${IRODS_EXTERNALS_FULLPATH_FMT}/lib/libfmt.so)
```
korydraughn, line 10: Add new line for unix tools

unit_tests/cmake/test_config/irods_zone_administration.cmake
```diff
@@ -0,0 +1,10 @@
+set(IRODS_TEST_TARGET irods_zone_administration)
+
+set(IRODS_TEST_SOURCE_FILES ${CMAKE_CURRENT_SOURCE_DIR}/src/main.cpp
+                            ${CMAKE_CURRENT_SOURCE_DIR}/src/test_zone_administration.cpp)
+
+set(IRODS_TEST_INCLUDE_PATH ${IRODS_EXTERNALS_FULLPATH_BOOST}/include)
+ 
+set(IRODS_TEST_LINK_LIBRARIES irods_common
+                              irods_client
+                              ${IRODS_EXTERNALS_FULLPATH_FMT}/lib/libfmt.so)
```
korydraughn, line 10: Add new line for unix tools

unit_tests/src/test_zone_administration.cpp
```diff
@@ -0,0 +1,126 @@
...
+    CHECK_NOTHROW(adm::client::modify_zone(conn, test_zone_name_1, adm::comment_property{comment_2}));
+
+    zone_info = adm::client::zone_info(conn, test_zone_name_1);
+    REQUIRE(zone_info.has_value());
+    CHECK(zone_info->type == adm::zone_type::remote);
+    CHECK(zone_info->name == test_zone_name_1);
+    CHECK(zone_info->connection_info == conn_info_2);
+    CHECK(zone_info->comment == comment_2);
+
+    // Show that the handling of the zone collection in regards to string-acls can be modified.
```
korydraughn, line 114: `string-acls` -> `strict-acls`

korydraughn:
This API plugin needs to consider rule engine plugins (e.g. NREP and rulebase changes).

korydraughn:
I'm considering renaming `library_support.h` to one of the following (order does not indicate preference):
- library_features.h
- features.h

I don't think `library_support.h` is bad, but having `features` in the filename feels a little clearer? I'm not completely sold on it yet.

Should these feature test macros be available at runtime? See the code sample from cppreference (located at the bottom of the page).
- https://en.cppreference.com/w/cpp/feature_test

plugins/api/src/switch_user.cpp
korydraughn: The alternative to optionally disconnecting is to ALWAYS disconnect connections created via a redirect.

This can be thought of as a reset. This will always be the case for older servers. For servers that support this API plugin, disconnecting is pretty wasteful.

Perhaps, the plugin should check the version of the server it is connected to and disconnect if it is dealing with an older server. That would avoid unnecessary network traffic and keep the logs of the remote server clean.

I'm starting to feel like this is the right approach, assuming we can sniff the version of the remote server.

plugins/api/src/switch_user.cpp
korydraughn: This no longer applies due to code changes.

plugins/api/src/switch_user.cpp
korydraughn: Implementation now checks the remote server's version and if it is less than 4.3.1, the plugin calls `rcDisconnect`. Otherwise, `rc_switch_user` is called.

unit_tests/src/test_rc_switch_user.cpp
korydraughn: done

unit_tests/src/test_rc_switch_user.cpp
korydraughn: done

unit_tests/src/test_rc_switch_user.cpp
korydraughn: done

unit_tests/src/test_rc_switch_user.cpp
korydraughn: done

lib/core/include/irods/library_support.h
korydraughn: done

lib/administration/zone/include/irods/zone_administration.hpp
korydraughn: done

lib/administration/zone/include/irods/zone_administration.hpp
korydraughn: done

unit_tests/src/test_zone_administration.cpp
korydraughn: done

plugins/api/src/switch_user.cpp
trel: This seems correct - no surprises with older servers, and the nice things with the new ones.

I think the only combinations are
- client -> new -> old ... (disconnect)
- client -> new -> new ... (switch user)
- client -> old -> old ... (n/a)
- client -> old -> new (n/a)

.clang-tidy
korydraughn: After discussion, the plan is to look at configuring this option and seeing how it behaves in the future.

plugins/api/src/switch_user.cpp
korydraughn: Created issue: #6644

.clang-tidy
korydraughn: The more I think about this one, the more I feel keeping it is the right thing to do. It's only an issue when looking at old C code, which means it is likely doing its job.

unit_tests/cmake/test_config/irods_rc_switch_user.cmake
korydraughn: done

unit_tests/cmake/test_config/irods_zone_administration.cmake
korydraughn: done

.clang-tidy
trel: this seems very reasonable.    can always change it later if you still find you don't like it.

korydraughn: I agree. Closing #6594 as a result of these conversations.

korydraughn: Done.

lib/administration/zone/CMakeLists.txt
korydraughn: This is no longer an issue for this PR since the zone admin lib was moved into a separate PR. See #6647

korydraughn:
Moved smaller/completed chunks of work to #6647

plugins/api/src/switch_user.cpp
```diff
@@ -16,108 +16,181 @@
...
-        // Update the user identity associated with the RsComm.
-        std::strncpy(client.userName, username, sizeof(UserInfo::userName));
-        std::strncpy(client.rodsZone, zone, sizeof(UserInfo::rodsZone));
-        std::strncpy(client.userType, user_type_string, sizeof(UserInfo::userType));
+    auto update_server_to_server_connections(RsComm& _comm) -> void
+    {
+        const auto switch_user_or_disconnect = [&client = _comm.clientUser](rodsServerHost& _host)
+        {
+            if (!_host.conn) {
+                log_api::warn("No connection to remote host."); // TODO Consider including the host's name.
```
trel, line 88: don't see a downside - could help the admin discover the misconfiguration more easily.

lib/api/src/rc_switch_user.cpp
```diff
@@ -15,11 +15,31 @@ auto rc_switch_user(RcComm* _comm, const char* _username, const char* _zone) ->
...
-    if (std::strcmp(_username, client.userName) == 0 && std::strcmp(_zone, client.rodsZone) == 0) {
+    // clang-format off
+    // NOLINTBEGIN(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+    //
+    // Return immediately if the RcComm object already represents the target user.
+    //
+    // This is allowed because the RcComm only cares about the username and zone. If the RcComm
+    // considered the user's type or authentication level (i.e. LOCAL_PRIV_USER_AUTH), then this
+    // optimization would not be possible.
+    if (std::strncmp(_username, client.userName, sizeof(UserInfo::userName)) == 0 &&
```
trel, line 36: the use of `strncmp` is to be more specific/safe since it only checks `n` characters?

is there any other reason here?

plugins/api/src/switch_user.cpp
```diff
@@ -0,0 +1,251 @@
...
+            // Remember, only iRODS 4.3.1 and later support the switch user API plugin.
+            if (*irods::to_version(_host.conn->svrVersion->relVersion) < irods::version{4, 3, 1}) {
+                rcDisconnect(_host.conn);
+                return;
+            }
+
+            // At this point, we know the remote server supports this API plugin.
+            if (const auto ec = rc_switch_user(_host.conn, client.userName, client.rodsZone); ec != 0) {
+                log_api::error(
+                    "rc_switch_user failed on remote server [error_code=[{}]]. Disconnecting from remote server.", ec);
```
trel, line 103: same here?   helpful to write remote server's hostname if we already have it?

plugins/api/src/switch_user.cpp
korydraughn: Will include the host info.

plugins/api/src/switch_user.cpp
```diff
@@ -0,0 +1,251 @@
...
+                return SYS_INTERNAL_ERR;
+            }
+
+            namespace adm = irods::experimental::administration;
+
+            // This call covers the existence check too.
+            const auto user_type = adm::server::type(*_comm, adm::user{_input->username, _input->zone});
+
+            if (!user_type) {
+                log_api::error("[{}#{}] is not a user in the local zone.", _input->username, _input->zone);
```
trel, line 150: could also mention the local zone's name

trel (commented):
why is removing the output / SwitchUserOut_PI the right answer?    just not used/needed for switch user at all - because no feedback is required?

plugins/api/src/switch_user.cpp
korydraughn: Sure thing.

lib/api/src/rc_switch_user.cpp
korydraughn: Correct.

The struct, `UserInfo`, defines the exact size of a username. Therefore, we check up to that many characters.

Thinking about this more, we should probably check `sizeof(UserInfo::userName) - 1` bytes. That takes the null byte into account.

trel: definitely poke at it with the edge cases and document any magical `- 1`s in the code please.

korydraughn: Assuming the value stored in `UserInfo::userName` is expected to be a null-terminated string. My guess is yes.

korydraughn:
> why is removing the output / SwitchUserOut_PI the right answer? just not used/needed for switch user at all - because no feedback is required?

The output type was only needed if setting the user type and privilege level on the client side. We've since learned that `RcComm`s only care about the username and zone, removing the need for an output type.

We already know who the client is trying to become so we use that info if the operation succeeds. The error code provides enough info for us to make that decision.

lib/CMakeLists.txt
korydraughn: done

trel:
sold - thanks.

plugins/api/src/switch_user.cpp
korydraughn: done

lib/api/src/rc_switch_user.cpp
korydraughn: done

plugins/api/src/switch_user.cpp
korydraughn: done

plugins/api/src/switch_user.cpp
korydraughn: done

trel:
no preference from me on 'renaming library_support.h' - i defer.

would suggest you consider how you'll reference it in casual discussion / talking / explaining to others... and use that name.

korydraughn:
Based on your comment, I think `library_features.h` is the right way to go.

The primary use of the header is to check if a particular library feature is available, so I think the rename is clear.

lib/api/src/rc_switch_user.cpp
```diff
@@ -0,0 +1,74 @@
...
+        return SYS_INVALID_INPUT_PARAM;
+    }
+
+    // Accepting empty strings for either parameter is not allowed.
+    // This is especially true for the "_zone" parameter because the implementation of this
+    // function copies the string represented by "_zone" into the RcComm. Copying an empty
+    // string into the RcComm's rodsZone member variables makes the RcComm invalid.
+    if (std::strlen(_username) == 0 || std::strlen(_zone) == 0) {
+        return SYS_INVALID_INPUT_PARAM;
+    }
```
korydraughn, line 24: We need to check that the strings passed also fit in the buffers.

korydraughn: done

korydraughn:
This PR now has a dependency on #6647

korydraughn:
Remember to add a test case that verifies `rc_switch_user` detects username and zone strings which violate the documented requirements.

lib/api/src/rc_switch_user.cpp
```diff
@@ -0,0 +1,92 @@
...
+        std::strncpy(static_cast<char*>(input.username), _username, sizeof(SwitchUserInput::username));
+        std::strncpy(static_cast<char*>(input.zone), _zone, sizeof(SwitchUserInput::zone));
+
+        const int ec = procApiRequest(_comm, SWITCH_USER_APN, &input, nullptr, nullptr, nullptr);
+
+        // NOLINTNEXTLINE(readability-implicit-bool-conversion)
+        if (ec == 0) {
+            // On success, we always assume that the output pointer is valid.
+            // If the server-side implementation is solid, then we shouldn't need to check the pointer.
+
```
korydraughn, line 76: remove

lib/api/include/irods/switch_user.h
```diff
@@ -0,0 +1,50 @@
...
+/// - It must be null-terminated
+/// - It must be non-empty
+/// \endparblock
+///
+/// \return An integer.
+/// \retval 0        On success.
+/// \retval non-zero On failure.
+///
+/// \since 4.3.1
+int rc_switch_user(struct RcComm* _comm, const char* _username, const char* _zone);
```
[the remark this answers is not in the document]
korydraughn: done

lib/api/src/rc_switch_user.cpp
korydraughn: done

korydraughn:
> This API plugin needs to consider rule engine plugins (e.g. NREP and rulebase changes).

We need to think about whether long running agents need to do anything for this case. Perhaps they do nothing. They operate as if the policy they started with never changes.

Thoughts?

plugins/api/src/switch_user.cpp
korydraughn: Based on what I've read, it looks like we don't have to do anything more in regard to `ZoneInfoHead`. `cleanupAndExit()` calls `disconnectAllSvrToSvrConn()` which iterates over the linked list pointed to by `ServerHostHead` and invokes `rcDisconnect()` on each entry.

This may also mean that the code snippet referenced here can iterate over the `ServerHostHead` instead of the `ZoneInfoHead`.

trel:
> They operate as if the policy they started with never changes.

Yes, not sure how we would demarcate exactly WHEN to use the 'new' policy/config vs the 'old'...  Seems per-connection is granular enough, and if we make that clear in the user/developer docs... then users/admins can plan accordingly.  The unit of work is the connection/agent lifetime.

korydraughn:
I think that is a fine trade off.

If you want speed and don't modify your policy often, `rc_switch_user` is perfect. Otherwise, it is a matter of picking what things can take advantage of the new API plugin.

I was considering handling the NREP files following a successful call to the API plugin, but that would likely kill any performance gained by using long running agents.

With that said, I need to update the API plugin documentation to include a few lines about agents and policy updates.

plugins/api/src/switch_user.cpp
```diff
@@ -0,0 +1,256 @@
...
+        }
+
+        return 0;
+    } // check_input
+
+    auto update_server_to_server_connections(RsComm& _comm) -> void
+    {
+        const auto switch_user_or_disconnect = [&client = _comm.clientUser](rodsServerHost& _host) {
+            if (!_host.conn) {
+                log_api::warn("No connection to remote host [{}].", _host.hostName->name);
```
korydraughn, line 89: Change to debug or trace.

plugins/api/src/switch_user.cpp
korydraughn: I wrote a little program that does the following:
1. Connects to consumer 1 as alice.
2. Creates a data object on otherResc under alice's home collection.
3. Switches user identity to bobby via `rc_switch_user`.
4. Creates a data object on the default resource under bobby's home collection.

Next, I used the testing environment to launch a zone with the following config:
- 1 provider
- 2 consumers
- 3 users (rods, alice, and bobby)
- 2 resources (demoResc and otherResc)

With this setup, I then ran the test program from consumer 1.

This test verifies that `rc_switch_user` updates the server-to-server connections as expected.

It seems that things are working as expected.

Do we feel more testing is needed?

Having documentation around `ZoneInfoHead` and friends would be beneficial here. I'll create an issue for this.

plugins/api/src/switch_user.cpp
korydraughn: done

plugins/api/src/switch_user.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+#ifdef RODS_SERVER
+    irods::client_api_allowlist::instance().add(SWITCH_USER_APN);
+#endif // RODS_SERVER
+
+    // clang-format off
+    irods::apidef_t def{SWITCH_USER_APN,            // API number
+                        RODS_API_VERSION,           // API version
+                        NO_USER_AUTH,               // Client auth
+                        NO_USER_AUTH,               // Proxy auth
```
korydraughn, line 240: Consider these.

Right now, privilege checks for handled by the API plugin directly. However, setting these allows us to remove that check.

The current implementation is easier to reason about. Hmm ...

unit_tests/src/test_rc_switch_user.cpp
```diff
@@ -0,0 +1,202 @@
...
+        // NOLINTEND(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+    }
+
+    const std::vector<std::pair<const char*, const char*>> bad_inputs{
+        {"", ""},
+        {"alice", ""},
+        {"", "tempZone"},
+        {"_________________________64_byte_string_________________________", ""},
+        {"", "_________________________64_byte_string_________________________"}
+    };
```
korydraughn, line 188: clang-format off/on

plugins/api/src/switch_user.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+#endif // RODS_SERVER
+
+// The plugin factory function must always be defined.
+extern "C" auto plugin_factory(
+    const std::string& _instance_name, // NOLINT(bugprone-easily-swappable-parameters)
+    const std::string& _context) -> irods::api_entry*
+{
+    static_cast<void>(_instance_name);
+    static_cast<void>(_context);
```
korydraughn, line 230: [[maybe_unused]]

unit_tests/src/test_rc_switch_user.cpp
korydraughn: done

plugins/api/src/switch_user.cpp
korydraughn: done

plugins/api/src/switch_user.cpp
trel: The error message on 129 is very good - will explain clearly when someone is holding it wrong...   is there a way to get that level of explanation here?

I think it's probably fine the way it is now if you leave a C++ comment near 236/237 saying what you've got in the discussion comment in this thread - "the 'real' check is above"

korydraughn: Newer code has used the style of L129 because it's easy/flexible, but it side steps the design.

With L236/237, we can't report error messages like with L129. However, we could modify the framework and apidef_t type to hold **static** error messages.

I'm fine with L129 though, it works. I just remembered the other way of doing this and it made me wonder.

I'm going to stick with L129.

trel: Seems good.

plugins/api/src/switch_user.cpp
korydraughn: Oh snap! @alanking already documented the relationship between `ZoneInfoHead` and `ServerHostHead`. No issue needed.

https://github.com/irods/irods/blob/54c2d73cab1edaff93ae61bc008eeb4361ff4b98/server/core/include/irods/rodsConnect.h#L115-L128

Based on the comments, we're good to mark this discussion as resolved.

korydraughn: This does infer that changing things to use `ServerHostHead` is acceptable.

This is very helpful. Found one typo in last statement.
https://github.com/irods/irods/blob/a86b0ed40bcb2bd23cf526a52244286b78285468/server/core/include/irods/rsGlobalExtern.hpp#L34-L47

korydraughn: I'm going to leave the code as is.

`secondaryServerHost` is never populated in iRODS so it is essentially a NOP. However, if it suddenly begins to be populated, then the API plugin will update those connections too.

korydraughn:
> With that said, I need to update the API plugin documentation to include a few lines about agents and policy updates.

This feels more like general information that should appear in the main docs rather than this API plugin.

Do we document the relationshipo between long-running agents and NREP policy?

lib/api/include/irods/switch_user.h
```diff
@@ -0,0 +1,54 @@
...
+
+// NOLINTBEGIN(modernize-use-trailing-return-type)
+
+/// Switches the identity of the user associated with the connection.
+///
+/// Requires that the proxy user be a \p rodsadmin.
+///
+/// On success, the following things will be true:
+/// - All API operations invoked via the \p RcComm will be carried out as the user that was just switched to.
+/// - Server-to-server connections will be updated or disconnected based on the remote server's version.
```
korydraughn, line 20: Add statement about what happens on failure.

plugins/api/src/switch_user.cpp
trel: makes sense to me!

korydraughn:
Add a unit test that does the following:
- Create a remote zone
- Create a user for the remote zone
- Verify that `rc_switch_user` can become the remote user

plugins/api/src/switch_user.cpp
```diff
@@ -0,0 +1,253 @@
...
+            // Set the appropriate privilege level based on whether the user is local or remote
+            // to this zone.
+            if (std::strncmp(_input->zone, local_zone, sizeof(SwitchUserInput::zone)) == 0) {
+                client.authInfo.authFlag =
+                    (*user_type == adm::user_type::rodsadmin) ? LOCAL_PRIV_USER_AUTH : LOCAL_USER_AUTH;
+            }
+            else {
+                client.authInfo.authFlag =
+                    (*user_type == adm::user_type::rodsadmin) ? REMOTE_PRIV_USER_AUTH : REMOTE_USER_AUTH;
+            }
```
korydraughn, line 176: If the server-to-server property is not set, the plugin needs to update the proxy user info as well.

If the server-to-server property is set, then do not update the proxy user info because the proxy user is the iRODS user administering the server.

plugins/api/src/switch_user.cpp
```diff
@@ -0,0 +1,253 @@
...
+        }
+    } // update_server_to_server_connections
+
+    auto rs_switch_user(RsComm* _comm, SwitchUserInput* _input) -> int
+    {
+        try {
+            // Only administrators are allowed to invoke this API.
+            // We check the proxy user because it covers clients and server-to-server redirects.
+            // Remember, the client and proxy user information is identical until there's a redirect.
```
alanking, line 128: Is this always true?

plugins/api/src/switch_user.cpp
```diff
@@ -0,0 +1,253 @@
...
+            }
+
+            namespace adm = irods::experimental::administration;
+
+            // This call covers the existence check too.
+            const auto user_type = adm::server::type(*_comm, adm::user{_input->username, _input->zone});
+
+            if (!user_type) {
+                log_api::error(
+                    "[{}#{}] is not a user in the local zone [{}].", _input->username, _input->zone, local_zone);
```
alanking, line 155: This wording makes me think that remote users are not allowed or considered here. Maybe this could say:
```
User type for [{}#{}] could not be identified.
```
Wording is fine as is, just offering this for clarity

unit_tests/src/test_rc_switch_user.cpp
```diff
@@ -0,0 +1,204 @@
...
+        // NOLINTEND(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+    }
+
+    // clang-format off
+    const std::vector<std::pair<const char*, const char*>> bad_inputs{
+        {"", ""},
+        {"alice", ""},
+        {"", "tempZone"},
+        {"_________________________64_byte_string_________________________", ""},
+        {"", "_________________________64_byte_string_________________________"}
```
alanking, line 188: Should we include tests for non-existent users/zones (but are valid inputs) and usernames which include zone names? Or are these equivalent to passing empty strings?

unit_tests/src/test_rc_switch_user.cpp
```diff
@@ -0,0 +1,204 @@
...
+    _getRodsEnv(env);
+
+    irods::experimental::client_connection conn;
+
+    auto* conn_ptr = static_cast<RcComm*>(conn);
+
+    SECTION("null pointers")
+    {
+        REQUIRE(rc_switch_user(conn_ptr, nullptr, nullptr) == SYS_INVALID_INPUT_PARAM);
+        // Become the administrator so that the test can clean up properly.
```
alanking, line 161: Is it necessary to switch back to the admin user in these cases? I thought the switch user call would fail and we would continue to be the admin user.

unit_tests/src/test_rc_switch_user.cpp
```diff
@@ -0,0 +1,204 @@
...
+    //
+    // IMPORTANT: This test requires access to a rodsadmin user!
+    //
+
+    load_client_api_plugins();
+
+    rodsEnv env;
+    _getRodsEnv(env);
+
+    irods::experimental::client_connection conn;
```
alanking, line 154: Would it make sense to include a test which creates a client connection using a non-rodsadmin user and attempt to invoke this API to ensure that only rodsadmins can invoke it?

alanking (commented):
Looks good. Just some clarifying comments, mostly

plugins/api/src/switch_user.cpp
korydraughn: Yes, unless you've seen something different.

plugins/api/src/switch_user.cpp
korydraughn: Will consider it.

unit_tests/src/test_rc_switch_user.cpp
korydraughn: Sure. Will add those cases.

unit_tests/src/test_rc_switch_user.cpp
korydraughn: That comment is incorrect. The user will not be changed on an error.

unit_tests/src/test_rc_switch_user.cpp
korydraughn: Absolutely.

lib/api/include/irods/switch_user.h
```diff
@@ -0,0 +1,54 @@
...
+///
+/// \param[in] _comm     A pointer to a RcComm.
+/// \param[in] _username \parblock
+/// The name of the iRODS user to become.
```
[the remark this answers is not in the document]
korydraughn: Nice catch. Will update.

plugins/api/src/switch_user.cpp
alanking: Okay. I may be getting authentication and connection information conflated in my mental model.

korydraughn: I believe there is a comment in the server that explicitly states this. Not sure where I found that. I know rcConnect sets both to the same value depending on how it is invoked.

unit_tests/src/test_rc_switch_user.cpp
korydraughn: Will remove the unnecessary calls to rc_switch_user too.

lib/api/include/irods/switch_user.h
korydraughn: done

lib/api/src/rc_switch_user.cpp
```diff
@@ -0,0 +1,89 @@
...
+    // NOLINTBEGIN(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+    //
+    // Return immediately if the RcComm object already represents the target user.
+    //
+    // This is allowed because the RcComm only cares about the username and zone. If the RcComm
+    // considered the user's type or authentication level (i.e. LOCAL_PRIV_USER_AUTH), then this
+    // optimization would not be possible.
+    //
+    // The substraction of 1 on the final argument helps to ensure there is a null byte. The
+    // server assumes the strings passed to the API are null-terminated strings.
```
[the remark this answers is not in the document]
korydraughn: done

plugins/api/src/switch_user.cpp
korydraughn: done

lib/api/include/irods/switch_user.h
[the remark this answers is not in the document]
korydraughn: done

unit_tests/src/test_rc_switch_user.cpp
korydraughn: done

unit_tests/src/test_rc_switch_user.cpp
korydraughn: done

unit_tests/src/test_rc_switch_user.cpp
korydraughn: done

lib/core/include/irods/library_features.h
```diff
@@ -18,5 +18,10 @@
 /// \since 4.3.1
 #define IRODS_HAS_LIBRARY_ZONE_ADMINISTRATION 202210L
 
+/// Defined if the development library supports ::rc_switch_user.
+///
+/// \since 4.3.1
+#define IRODS_HAS_API_ENDPOINT_SWITCH_USER 202210L
```
korydraughn, line 24: 202211L

korydraughn: done

alanking:
Looks good... Is there anything else to do?

My only remaining question/concern: Should we make this a separate plugin (i.e. doesn't ship with the server) and pull it in? I' spent a little time thinking through pitfalls and security concerns and couldn't think of anything bad, but once it's in, it's in...

trel:
for inclusion
 - no additional wiring needed, it's just in there
 - we're not going to introduce it to 4.2.x... so we're already in the 'right timeframe' to add something like this
 - we want it for client/gateway performance

as a plugin (at least at first)
 - we don't know what we don't know... but it seems okay from our own thinking out loud so far

...

so... i'm leaning ... for inclusion ...

alanking:
Yeah, I think it's probably going to get pulled into the server anyway as we start looking into applications (e.g. connection pooling). Just wanted to take the opportunity to lift up that stone before it was too late. :)

>    How do we test it? (the unit tests are probably good enough for this plugin)

I agree. The API tests you wrote are all we've got since nothing is using this yet.

>    How does it interact with library_features.h?
      Does it interact with library_features.h at all?

I feel that it shouldn't interact. A true plugin is something that will not always be available for a given server, so I'm not sure what the macro would do for the developer in that case.

>    Where does it live? In its own repo or a repo with other plugins?

I would think it would be its own repo like we do for our other plugins.

>    Is there an incubation period for things like this?

I don't recall there being any process declared around this. Only precedent I can think of is update_collection_mtime, and that isn't quite the same thing.

trel:
Correct, update_collection_mtime is all I can remember being external, then getting pulled in...

Definitely no formal process for incubation today - just part of the generic development deliberation/process by the consortium (what we're doing now).

korydraughn:
I agree. I'll think about the pending review comment and hopefully we'll see this land tomorrow.

plugins/api/src/switch_user.cpp
korydraughn: How does this sound?
```
User type for [{}#{}] could not be identified in local zone [{}].
```

lib/api/src/rc_switch_user.cpp
```diff
@@ -0,0 +1,87 @@
...
+    // This is especially true for the "_zone" parameter because the implementation of this
+    // function copies the string represented by "_zone" into the RcComm. Copying an empty
+    // string into the RcComm's rodsZone member variables makes the RcComm invalid.
+    if (username_length == 0 || zone_length == 0) {
+        return SYS_INVALID_INPUT_PARAM;
+    }
+
+    // These compile-time assertion guarantee that we always keep these data structures in sync.
+    static_assert(sizeof(SwitchUserInput::username) == sizeof(UserInfo::userName));
+    static_assert(sizeof(SwitchUserInput::zone) == sizeof(UserInfo::rodsZone));
```
korydraughn, line 31: Move to the plugin impl file.

plugins/api/src/switch_user.cpp
```diff
@@ -0,0 +1,290 @@
...
+            return SYS_INVALID_INPUT_PARAM;
+        }
+
+        if (std::strlen(_input->zone) == 0) {
+            log_api::error("Invalid zone argument: empty string");
+            return SYS_INVALID_INPUT_PARAM;
+        }
+
+        return 0;
+    } // check_input
```
korydraughn, line 93: The input validation performed in the rc function need to be performed here.

Consider removing some of the checks from the rc function if it makes sense.

plugins/api/src/switch_user.cpp
trel: that seems good and specific/thorough

korydraughn: done

lib/api/src/rc_switch_user.cpp
korydraughn: done

plugins/api/src/switch_user.cpp
korydraughn: done

alanking (approved):

korydraughn:
I removed one last comment (no code changes).

Will run all of the unit tests one last time. If all tests pass, I'll add the #'s.

korydraughn:
All unit tests passed.

Added #'s.

Issue #6569, opened 2022-08-29
Investigate test failures in test_configuration_reload.py


korydraughn:
- [x] main
- [x] 4-3-stable
---
## Bug Report

### iRODS Version, OS and Version
main

### What did you try to do?
I ran the test_configuration_reload.py test suite using the testing environment.

### Expected behavior
I expected all tests to pass.

### Observed behavior (including steps to reproduce, if applicable)
```bash
 --- IrodsSession: icommand executed by [rods] [iinit] ---                                                                                                                                                                                                                                                                                                                                   
Assert Command: iinit                                                                                                                                                                                                                                                                                                                                                                        
Expecting STDOUT_SINGLELINE: ['']                                                                                                                                                             
  stdout:                                                                                                                                                                                                                                                                                                                                                                                    
    | WARNING: Error 25 disabling echo mode. Password will be displayed in plaintext.                                                                                                         
    | Enter your current iRODS password:
  stderr:
    | remote addresses: 172.25.0.3 ERROR: _rcConnect: connectToRhost timed out, retrying
    | remote addresses: 172.25.0.3 ERROR: _rcConnect: connectToRhost error, server on 76aaa30cffb6:1247 is probably down status = -115000 SYS_SOCK_READ_TIMEDOUT
    | remote addresses: 172.25.0.3 ERROR: Saved password, but failed to connect to server 76aaa30cffb6
Unexpected output on stderr

FAILED TESTING ASSERTION


<xmlrunner.xmlrunner._XMLTestResult run=4 errors=3 failures=0>

2022-08-29 07:59:12,988 ERROR - 
Running tests...
----------------------------------------------------------------------
  test_how_long_reloading_takes (irods.test.test_configuration_reload.TestConfigurationReload) ... OK (17.666s)
  test_how_long_restarting_takes (irods.test.test_configuration_reload.TestConfigurationReload) ... OK (312.079s)
  test_reload_command__issue_6529 (irods.test.test_configuration_reload.TestConfigurationReload) ... FAIL (1.952s)
  test_server_correctly_detects_change_in_server_config (irods.test.test_configuration_reload.TestConfigurationReload) ... FAIL (200.236s)

======================================================================
ERROR [0.000s]: test_reload_command__issue_6529 (irods.test.test_configuration_reload.TestConfigurationReload)
----------------------------------------------------------------------
Traceback (most recent call last):
[...]

PR #6605, opened 2022-09-26
[#6569] Fixed issues with configuration reload tests (main)

korydraughn:
Running core tests ...

### DIFF
--- a/lib/core/src/irods_server_properties.cpp
+++ b/lib/core/src/irods_server_properties.cpp
@@ -11,6 +11,7 @@
 #include <cstring>
 #include <fstream>
 #include <unordered_map>
+#include <algorithm>
 
 #include <curl/curl.h>
 #include <curl/easy.h>
@@ -236,7 +237,10 @@ namespace irods
 
     nlohmann::json server_properties::reload()
     {
-        nlohmann::json new_values;
+        using json = nlohmann::json;
+
+        json new_values;
+
         if (use_remote_configuration()) {
             new_values = process_remote_configuration(capture_remote_configuration());
         }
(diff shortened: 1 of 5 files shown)


lib/core/src/irods_server_properties.cpp
```diff
@@ -257,24 +261,42 @@ namespace irods
...
+            if (std::any_of(std::begin(runtime_properties), std::end(runtime_properties), pred)) {
+                op = "add";
+                entry["value"] = config_props_.at(json::json_pointer{path});
             }
         }
-        auto lock = acquire_write_lock();
-        // TODO uncomment when issue #6470 is fixed
-        // log_server::error("Before patch '{}'", config_props_.dump());
+
+        log_server::debug("Before patch ==> {}", [this] { return std::make_tuple(config_props_.dump()); });
```
korydraughn, line 296: Verify that this is correct and doesn't invoke UB.

server/main_server/src/rodsServer.cpp
```diff
@@ -139,21 +139,21 @@ namespace
         // Should only ever set the cache salt once.
         try {
             const auto& existing_salt = irods::get_server_property<const std::string>(irods::KW_CFG_RE_CACHE_SALT);
-            log_server::error("createAndSetRECacheSalt: salt already set [{}]", existing_salt.c_str());
-            return ERROR(SYS_ALREADY_INITIALIZED, "createAndSetRECacheSalt: cache salt already set");
+            log_server::debug("createAndSetRECacheSalt: Cache salt already set [{}]", existing_salt.c_str());
```
alanking, line 142: Is this changing to `debug` because the returned `irods::error` will get logged somewhere?

trel: `debug` because this will be the normal case on a reload.

alanking: Okay

lib/core/src/irods_server_properties.cpp
korydraughn: Verified that the code works as intended.

Learned that the main server process does not honor the log level following a reload. Created issue at #6608.

alanking (approved):

korydraughn:
All unit/core tests passed.

korydraughn:
Added #'s

PR #6615, opened 2022-09-28
[4421] Added detached mode for unixfilesystem resource (4-3-stable)

JustinKyleJames:
All unixfilesystem tests were run and passed.

### DIFF
--- a/plugins/resources/CMakeLists.txt
+++ b/plugins/resources/CMakeLists.txt
@@ -6,6 +6,7 @@ add_custom_target(all-plugins-resources)
 add_dependencies(all-plugins_no_database all-plugins-resources)
 
 add_subdirectory(replication)
+add_subdirectory(unixfilesystem)
 
 set(
   IRODS_RESOURCE_PLUGINS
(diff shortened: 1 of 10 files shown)


trel:
one conflict now

JustinKyleJames:
> one conflict now

This was the one I referred to earlier.  The previous change must have been merged in the meantime.  Will resolve.

alanking:
For reference...

cherry-picked from https://github.com/irods/irods/pull/6525

JustinKyleJames:
> one conflict now

Resolved it and pushed.

alanking (approved):
Identical to main, so seems good to me

korydraughn (approved):
I just realized that the `is_host_in_host_list` function takes a `const std::string_view&`. We need to remove the ampersand from that parameter.

Let's get this in as-is so that main and 4-3-stable line up. And then let's handle the removal of the ampersand in a separate PR.

alanking:
Are we ready for #'s on this?

JustinKyleJames:
> I just realized that the `is_host_in_host_list` function takes a `const std::string_view&`. We need to remove the ampersand from that parameter.
> 
> Let's get this in as-is so that main and 4-3-stable line up. And then let's handle the removal of the ampersand in a separate PR.

Yes, there is an issue on that.  I can fix and open another PR for main and 4-3-stable ASAP.

JustinKyleJames:
Added pound.

Issue #6597, opened 2022-09-21
Ubuntu 22.04 Version Support


kript:
- [x] main (irods/irods and irods/irods_client_icommands)
- [x] 4-3-stable (irods/irods and irods/irods_client_icommands)
- [x] GitHub Actions (main)
- [x] GitHub Actions (4-3-stable)
- [x] development environment
- [x] testing environment

----

## Feature

My apologies if this is in one of the development updates, but I couldn't find it and I didn't see an issue so I thought I'd ask; since Ubuntu 22.04 is now released, what is the roadmap for it being supported on iRODS (taking into account #4637 ) on both 4.3.x and 4.2.x?

Thanks!

John

trel:
We will support Ubuntu22.04 in an upcoming 4.3.x release.  No known blockers at this time.

We will not support Ubuntu22.04 for 4.2.x. - see earlier discussion at https://github.com/irods/irods/issues/4883.

alanking:
I have demonstrated that iRODS 4.3.x can build and run on Ubuntu 22.04, but it will require several changes across a few repositories (including, probably, all of the plugins). These mainly have to do with the version bump to OpenSSL 3.0 and a lot of deprecation warnings.

korydraughn:
We have not confirmed anything yet. That effort has taken a backseat to other issues. Once those are taken care of, we'll be able to focus on Ubuntu 22.

Is support for Ubuntu 22 urgent?

ccacciari:
We offer iRODS to our users in a cloud environment deployed on our infrastructure. iRODS is one of the default components, therefore the missing Ubuntu 22 iRODS packages prevent us to support Ubuntu 22 in that environment, unless we remove iRODS from the default software stack. We would like to offer Ubuntu 22 before the summer.
However, it would be enough to have the icommands client package as a start, not everything. Would providing just the icommands for Ubuntu 22 makes the things easier for you? Or it does not change much?

alanking:
I was able to build and install iRODS (future 4.3.1) and iCommands on Ubuntu 22.04 with the following changes:
irods/externals: https://github.com/irods/externals/pull/178
irods/irods: https://github.com/irods/irods/pull/6620
irods/irods_client_icommands: https://github.com/irods/irods_client_icommands/pull/323

I have not tried building the plugins yet, so those will likely require changes to how they are being built.

For completeness, the development and testing environments have changes in the wings as well:
irods_development_environment: https://github.com/irods/irods_development_environment/pull/72
[...]

PR #6620, opened 2022-10-05
[#6597] Specify OpenSSL versions (main)

alanking:
On newer platforms, some of the functions and types in use by the SSL networking plugin have been deprecated or hidden so we need to set a compiler definition to indicate the version of SSL we would like to be compatible with until we update these calls.

Specifically, OpenSSL 3.0 deprecated many low-level APIs and types in favor of a new kind of interface.

The version of libssl in our declared dependencies needs to be changed conditionally based on the version of OpenSSL detected by find_package.

---

This is required to build for Ubuntu 22.04 / Jammy Jellyfish.

We will need to actually update our OpenSSL usage in the future. For more information about the macro used, see: https://www.openssl.org/docs/man3.0/man7/OPENSSL_API_COMPAT.html


korydraughn (approved):
The changes look good to me.

alanking:
Rebased and marking Ready for Review

alanking:
Will need to investigate the build failures...

CMakeLists.txt
```diff
@@ -227,6 +227,7 @@ set(
   BOOST_CHRONO_DONT_PROVIDES_DEPRECATED_IO_SINCE_V2_0_0
   BOOST_STACKTRACE_USE_ADDR2LINE
   BOOST_ASIO_NO_DEPRECATED
+  OPENSSL_API_COMPAT=0x10100000L
```
korydraughn, line 230: Consider adding a comment above this line explaining the reason for this.

alanking: Good point

alanking: Done

alanking:
I confirmed that the platforms which were failing to build are now building with the latest change. We'll see how the workflows respond...

korydraughn:
Ok. Will await confirmation of success.

CMakeLists.txt
```diff
@@ -227,6 +227,10 @@ set(
   BOOST_CHRONO_DONT_PROVIDES_DEPRECATED_IO_SINCE_V2_0_0
   BOOST_STACKTRACE_USE_ADDR2LINE
   BOOST_ASIO_NO_DEPRECATED
+  # Historically, iRODS has used OpenSSL API version 1.1.x. OpenSSL 3.x deprecated and even removed some API calls
+  # in use by iRODS. This requires a compile definition which specifies that the API version with which the target
```
alanking, line 231:
```suggestion
  # in use by iRODS. This requires a compile definition which specifies the API version with which the target
```

alanking:
Looks like that did the trick. Unfortunately, this form of the `OPENSSL_API_COMPAT` macro and the one used in https://github.com/irods/libs3/pull/22 are not consistent, so I'm looking into why that is. If we don't care about consistency between these, then I have a version that seems to work for all platforms.

korydraughn:
How are they not consistent?

They use the same shared library, right?

alanking:
https://www.openssl.org/docs/man3.0/man7/OPENSSL_API_COMPAT.html describes two forms of defining `OPENSSL_API_COMPAT`:

1. **0xMNNFF000L**
2. **mmnnpp**

The second is the newer, recommended form, and the one which is being used in the `CMakeLists.txt` for this PR.

The first is the only one which works for https://github.com/irods/libs3/pull/22 for all platforms. I got the following message on Ubuntu 22.04 when attempting to compile libs3 with the second form:
```
error: #error "OPENSSL_API_COMPAT expresses an impossible API compatibility level"
```

So, the first form works fine for all platforms when compiling libs3. The second form works for all platforms when compiling iRODS. Not sure why yet.

korydraughn:
Let's capture that in an issue (perhaps in a investigative like issue).

If testing for 4.3.1 results in weird behavior in regard to SSL, libs3, etc, then we can circle back and discuss whether we feel this is related.

alanking:
Created https://github.com/irods/irods/issues/7175 for investigating this issue.

CMakeLists.txt
alanking: Done

korydraughn:
Anything left to do for this?
Are we waiting on other PRs before merging this?

alanking:
Thinking on it, I guess this could be merged independently of the other changes. The existing platforms build with these changes and without needing to change the externals, and Ubuntu 22.04 builds with the changes here as well as to externals.

So, no, I think there is nothing left to do for this change.

korydraughn:
Ok. Let's wait until you're happy with everything that is ub22 and then we'll merge everything together.

korydraughn (approved):
Let's add the pound and merge.

Issue #6622, opened 2022-10-06
closeAllL1desc should not call PEPs


dworkin:
- [x] main
- [x] 4-3-stable
- [x] 4-2-stable
---
`msiExecCmd` sets up communication pipes, forks, closes excess file descriptors and executes a program.

The file descriptors are closed by `closeAllL1desc`, which is called from nowhere else. It starts at `L1` to discover the descriptors, and then goes back up to `L3` to close them. However, this has all sorts of effects -- apart from closing the file descriptors -- that are unwanted at this point. It is not appropriate that `pep_resource_close_pre` is called because the relevant resource is still open in the parent process, and it's not appropriate to run _any_ rules in a forked process.

We ran into this because `msiExecCmd` was called from python code, and then `invoke_policy_enforcement_point` would call `rule_exists` in the python plugin, which deadlocks because it tries to do
```
std::lock_guard<std::recursive_mutex> lock{python_mutex};
```
from a forked process.

dworkin:
[This code](https://github.com/irods/irods/blob/main/server/api/src/rsFileClose.cpp#L93-L100) tries to deal with an earlier version of this problem and may be superfluous after the pull request is merged, assuming that it really only deals with the `msiExecCmd` case.

dworkin:
An earlier incarnation of this issue is #1713.

trel:
*much* earlier :)

PR #6623, opened 2022-10-06
[#6622]  closeAllL1desc should not call PEPs

dworkin:
closeAllL1desc is (only) called from a forked process which is preparing to execute an external program.  It should not do anything other than close the open resources. See issue #6622.

Applies to main, 4-3-stable, 4-2-stable.

### DIFF
--- a/server/core/src/objDesc.cpp
+++ b/server/core/src/objDesc.cpp
@@ -111,7 +111,7 @@ closeAllL1desc( rsComm_t *rsComm ) {
     for ( i = 3; i < NUM_L1_DESC; i++ ) {
         if ( L1desc[i].inuseFlag == FD_INUSE &&
                 L1desc[i].l3descInx > 2 ) {
-            l3Close( rsComm, i );
+            close( FileDesc[L1desc[i].l3descInx].fd );
         }
     }
     return 0;


trel:
We'll check it out - thanks!

Should have a test to trap/exercise this as well...

alanking:
I think that this change is okay. The forked process does not act as an iRODS agent, so closing these resources in this way seems fine. I agree with @dworkin's statement in #6622:
> The file descriptors are closed by closeAllL1desc, which is called from nowhere else. It starts at L1 to discover the descriptors, and then goes back up to L3 to close them. However, this has all sorts of effects -- apart from closing the file descriptors -- that are unwanted at this point. It is not appropriate that pep_resource_close_pre is called because the relevant resource is still open in the parent process, and it's not appropriate to run any rules in a forked process.

My only concern would be whether non-`unixfilesystem`-type storage resources with open L3 descriptors would respond appropriately to the `close` call in all cases. @JustinKyleJames - Do you know if it's appropriate to call `close` on the L3 descriptors opened in the S3 resource plugin, if indeed there are any to close?

@trel - I also agree that we should consider writing a test for this.

JustinKyleJames:
> iate that pep_resource_close_pre is called because the relevant resource is still open in the parent process, and it's not appropriate to run any ru

I would definitely have to test that out.

alanking:
I'll write a test for this today to be sure, and I'll hand it off to @JustinKyleJames for testing with the S3 resource plugin.

JustinKyleJames:
I executed the test Alan describes above against an S3 resource in both cache and cacheless mode.  Everything worked fine.

alanking:
Excellent, thanks! I'll take care of cherry-picking and whatnot soon.

Issue #6579, opened 2022-09-06
`ichmod` should not be allowed to bypass the permission model


korydraughn:
- [x] main
- [x] 4-3-stable
---
## Improvement
`ichmod` is currently allowed to bypass the permission model when the user adjusting the permissions matches the original owner. See example below.
```bash
alice@machine $ itouch foo
alice@machine $ ichmod null alice foo
alice@machine $ ichmod own alice foo   # This should result in an error.
```
Only users with `own` permissions or an admin should be allowed to restore `alice`'s permissions.

This makes the behavior around permissions clearer and makes all clients behave the same.

darthtux:
I am a big fan of removing ichmod's ability to bypass the permissions model when used by the original object owner.   Its presence causes some issues for our workflows and, as already stated, removing this ability would make permissions clearer, especially from an audit standpoint.

mstfdkmn:
Would it be good to think about keeping at least one user ownership on an object somehow? This could be either by not allowing the 'owner' to remove his/her access completely (not lower than read) or by locking the last remaining acl on an object. Because we are seeing some 'orphaned' objects (no acl exists on) and only admins can search these and make them again visible in the logical namespace by giving permission to the original owner.

A rods user can/cant do:
```
ils -A alice1.txt
  /kuleuven_tier1_pilot/home/vsc33586/alice1.txt
        ACL - vsc33586#kuleuven_tier1_pilot:own
```
`ichmod null vsc33586 alice1.txt
`
```
iquest "select COLL_NAME, DATA_NAME, DATA_PATH, DATA_OWNER_NAME where DATA_NAME = 'alice1.txt'"
CAT_NO_ROWS_FOUND: Nothing was found matching your query
```
A rodsadmin can do:
 ```
iquest "select DATA_NAME, DATA_PATH, DATA_OWNER_NAME where DATA_NAME = 'alice1.txt'"
DATA_NAME = alice1.txt
DATA_PATH = /irods/a/home/vsc33586/alice1.txt
DATA_OWNER_NAME = vsc33586

```
```
ils -A /kuleuven_tier1_pilot/home/vsc33586/data.tar
  /kuleuven_tier1_pilot/home/vsc33586/data.tar
        ACL -
```
[...]

PR #6625, opened 2022-10-07
[#6579] Fix `ichmod` permission bypass (4-3-stable)

MartinFlores751:
User should now not be able to `ichmod` data objects and collections back to `own`, after setting to a lower permission.


trel:
you probably got it in one - but we want to hold off on the initial # on the cherrypicks until they're approved too.

leave this one as is - no need for extra churn.

PR #6632, opened 2022-10-11
Add where clause in delete from R_ObjectId_seq_tbl (main)

peterverraedt:
The mysql function R_ObjectId_nextval() is not safe for parallel database updates, that some of our users can trigger by opening a large number of parallel connections. This can be simulated by

```
(for i in $(seq 0 10); do echo 'echo "SELECT R_ObjectID_nextval();" |
mysql -h (...) -u(...) -p(...) irods'; done) | parallel -j 10
```

This results in "Deadlock found when trying to get lock; try restarting transaction". After restricting the delete to only remove the row that that instance of the function has inserted, the deadlock is avoided.

Signed-off-by: Peter Verraedt <peter.verraedt@kuleuven.be>

### DIFF
--- a/plugins/database/src/mysql_functions.sql.in
+++ b/plugins/database/src/mysql_functions.sql.in
@@ -5,7 +5,7 @@ returns bigint
 begin
    insert into R_ObjectId_seq_tbl values (NULL) ;
    set @R_ObjectId_val=LAST_INSERT_ID() ;
-   delete from R_ObjectId_seq_tbl ;
+   delete from R_ObjectId_seq_tbl where nextval = @R_ObjectId_val ;
    return @R_ObjectId_val ;
 end
 %%


alanking (commented):
This looks good to me. Is there an issue we can link here?

trel:
seems safe - localized to mysql.

korydraughn:
Sounds good to me.

korydraughn (approved):

korydraughn:
I guess the only thing is, can we capture this in a test?

alanking:
Test suite passed

trel:
Nice.   Next steps?

alanking:
Sorry, I guess I didn't say anything. If we are all in agreement that there is no test needed for this, then we can add the issue number to the commit message and get this in. If we think there should be a test, please suggest one and why it is necessary.

Awaiting feedback :)

korydraughn:
I think we're good. The test suite passed so there's no known issues at this time.

trel:
Agreed - # it up.

alanking:
Oh, I forgot about my own question:

> Something I noticed is that the mysql_functions.sql.in file is used to generate a file which is only used when creating database tables when setting up the catalog. Does that mean that this will only apply to new catalog setups? If not, is there a way to deploy this change on an existing system?

I may be misunderstanding the role of this file, but I wanted to know whether or not we needed to consider upgrades here, or if this is something that can be done by the administrator after setting up the database.

korydraughn:
Based on the input file, this could be run for upgrades and I think it should. We need to confirm if the upgrade logic already do that though.

trel:
From my reading, all runs other than the initial run, will find the iRODS tables already existing, and skip the rest of `create_database_tables`... not affecting the previously defined mysql_functions.

https://github.com/irods/irods/blob/f6eb6c72786288878706e2562a370b91b7d0802e/scripts/irods/database_connect.py#L335-L354

We could have an upgrade codepath check this defined value and update the function... or just re-run this mysql_functions.sql file on each upgrade...

korydraughn:
We can add a file for the upgrade path (e.g. mysql_upgrade.sql) which contains this one fix.

trel:
rather than just re-run this same mysql_functions.sql file with the 'drop function if exists'?

single source of truth?

korydraughn:
That could work, but that file touches other things. By using a separate file, we get to be explicit. This also protects databases from weird behavior around replacing things in MySQL.

alanking:
After discussion, we have determined that the best course of action would be to call this logic on package upgrades (which is run every time the server is started to determine whether it has been upgraded) instead of calling it on initial installation (`setup_catalog`). If we find that the MySQL functions are necessary during installation, we will leave the call where it is.

We will go ahead and merge this and address the upgrade changes before closing #6670.

alanking:
@peterverraedt - Please adjust commit message title to look like this:
```
[#6670] Add where clause in delete from R_ObjectId_seq_tbl
```
I'm not allowed to push on this branch :)

alanking:
Thanks!

Issue #5964, opened 2021-10-29
Fix delay hints parser for "DOUBLE" directives


korydraughn:
- [x] main
- [ ] 4-3-stable
---
## Bug Report
The delay hint `DOUBLE` directives produce non-deterministic results.
https://github.com/irods/irods/blob/970a4286ef2096a59707f415a00bf7e07f2dbbeb/lib/core/src/rcMisc.cpp#L1666-L1708

MartinFlores751:
Closing. Merged in main and 4-3-stable.

PR #6633, opened 2022-10-11
[#5964] Correct `DOUBLE` directive tests (main)

MartinFlores751:
It seems that for some `DOUBLE` directive tests, we were using the same string object passed to string views. This meant that when we set the expected results string, we also modified the input string. This may have been causing the non-deterministic results.

### DIFF
--- a/unit_tests/src/test_delay_hints_parser.cpp
+++ b/unit_tests/src/test_delay_hints_parser.cpp
@@ -5,77 +5,79 @@
 #include <fmt/format.h>
 #include <fmt/chrono.h>
 
-#include <cstring>
 #include <map>
 #include <string>
 #include <string_view>
+#include <array>
 
 struct input
 {
-    std::string_view current_time;
-    std::string_view frequency;
+    std::string current_time;
+    std::string frequency;
 };
 
 struct output
 {
     int expected_action_code;
-    std::string_view expected_next_time;
-    std::string_view expected_frequency;
+    std::string expected_next_time;
+    std::string expected_frequency;
 };
 
 auto test_execution_frequency_handler(const input& _input, const output& _output)
 {
-    char current_time[TIME_LEN]{};
-    std::strcpy(current_time, _input.current_time.data());
+    std::array<char, TIME_LEN> current_time{};
+    std::copy(_input.current_time.cbegin(), _input.current_time.cend(), current_time.begin());
 
-    char frequency[128]{};
-    std::strcpy(frequency, _input.frequency.data());
+    constexpr int frequency_length{128};
+    std::array<char, frequency_length> frequency{};
+    std::copy(_input.frequency.cbegin(), _input.frequency.cend(), frequency.begin());
 
-    char next_time[TIME_LEN]{};
+    std::array<char, TIME_LEN> next_time{};
 
-    CHECK(getNextRepeatTime(current_time, frequency, next_time) == _output.expected_action_code);
-    CHECK(next_time == _output.expected_next_time);
-    CHECK(frequency == _output.expected_frequency);
+    CHECK(getNextRepeatTime(current_time.data(), frequency.data(), next_time.data()) == _output.expected_action_code);
+    CHECK(next_time.data() == _output.expected_next_time);
+    CHECK(frequency.data() == _output.expected_frequency);
 }
 
-template <typename Timepoint>
-auto make_repeat_until_time_string(int _nnnn,
-                                   const std::string_view _U,
-                                   const Timepoint _tp) -> std::string
+auto make_simple_directive_string(
+    int _delay_duration,
+    const std::string_view _delay_time_unit,
+    const std::string_view _directive) -> std::string
 {
-    return fmt::format("{}{} REPEAT UNTIL {}", _nnnn, _U, std::chrono::system_clock::to_time_t(_tp));
+    return fmt::format("{}{} {}", _delay_duration, _delay_time_unit, _directive);
 }
 
-template <typename Timepoint>
-auto make_repeat_until_success_or_until_time_string(int _nnnn,
-                                                    const std::string_view _U,
-                                                    const Timepoint _tp) -> std::string
+auto make_time_directive_string(
+    int _delay_duration,
+    const std::string_view _delay_time_unit,
+    const std::string_view _directive,
+    const rodsLong_t _time) -> std::string
 {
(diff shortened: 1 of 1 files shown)


unit_tests/src/test_delay_hints_parser.cpp
```diff
@@ -262,36 +256,35 @@ TEST_CASE("#5503: getNextRepeatTime parses frequency correctly")
...
-                .expected_action_code = 0,
-                .expected_next_time   = "20",
-                .expected_frequency   = double_until_success_or_n_times
+                .expected_action_code = 4,
+                .expected_next_time   = "10",
+                .expected_frequency   = "20s DOUBLE UNTIL SUCCESS OR 19 TIMES. ORIGINAL TIMES=20"
            }
         }},
-        {"DOUBLE UNTIL SUCCESS OR UPTO <TIME>", {
+        {"DOUBLE UNTIL SUCCESS UPTO <TIME>", {
```
korydraughn, line 275: What lead to this change?

If `OR` is incorrect, then we will need to update the docs as well. See https://docs.irods.org/4.3.0/plugins/pluggable_rule_engine/#delay-execution

unit_tests/src/test_delay_hints_parser.cpp
```diff
@@ -75,7 +70,14 @@ auto make_double_until_success_or_until_time_string(int _nnnn,
...
     return fmt::format("{}{} DOUBLE UNTIL SUCCESS OR UNTIL {}", _nnnn, _U, std::chrono::system_clock::to_time_t(_tp));
 }
-#endif
+
+template <typename Timepoint>
+auto make_double_until_success_or_upto_time_string(int _nnnn,
+                                                   const std::string_view _U,
+                                                   const Timepoint _tp) -> std::string
+{
+    return fmt::format("{}{} DOUBLE UNTIL SUCCESS UPTO {}", _nnnn, _U, std::chrono::system_clock::to_time_t(_tp));
```
korydraughn, line 79: No `OR UPTO`?

The function name includes `_or_` and the delay directive above this uses it as well.

korydraughn (changes requested):
Very nice.

So the issue was with the tests rather than `getNextRepeatTime()`?

unit_tests/src/test_delay_hints_parser.cpp
```diff
@@ -75,7 +70,14 @@ auto make_double_until_success_or_until_time_string(int _nnnn,
 {
     return fmt::format("{}{} DOUBLE UNTIL SUCCESS OR UNTIL {}", _nnnn, _U, std::chrono::system_clock::to_time_t(_tp));
 }
-#endif
+
+template <typename Timepoint>
+auto make_double_until_success_or_upto_time_string(int _nnnn,
+                                                   const std::string_view _U,
```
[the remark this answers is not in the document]
MartinFlores751: Along with better parameter names, I think we could also end up refactoring all of the functions that make the strings into a single function too. They all effectively do the same thing, we just have to make the directive a parameter.

unit_tests/src/test_delay_hints_parser.cpp
MartinFlores751: The current implementation and the test were in disagreement. What ended up happening was I 'sided' with the existing implementation, probably cause I thought that it was an error in the test since I was already working on the previous ones.

I think I also looked at the 4.3.0 doc you provided too, and didn't see any reference to an `UPTO` directive. For consistency though, it probably should have an `OR`.

unit_tests/src/test_delay_hints_parser.cpp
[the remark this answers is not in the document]
korydraughn: If you feel that makes the code clearer, then please share what you have in mind here and others can post their thoughts.

unit_tests/src/test_delay_hints_parser.cpp
```diff
@@ -262,36 +256,35 @@ TEST_CASE("#5503: getNextRepeatTime parses frequency correctly")
...
+                .frequency    = "10s DOUBLE UNTIL SUCCESS OR 20 TIMES"
            },
             // Output
            {
-                .expected_action_code = 0,
-                .expected_next_time   = "20",
-                .expected_frequency   = double_until_success_or_n_times
+                .expected_action_code = 4,
+                .expected_next_time   = "10",
+                .expected_frequency   = "20s DOUBLE UNTIL SUCCESS OR 19 TIMES. ORIGINAL TIMES=20"
```
alanking, line 272: Is there a reason we do not have `double_until_success_or_n_times` and `double_until_success_or_n_times_result` strings here, similar to all the other cases?

alanking (commented):
The writeup in the OP (or something like it) would be good to include in your commit message, IMO

unit_tests/src/test_delay_hints_parser.cpp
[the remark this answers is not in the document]
MartinFlores751: No problem, I was thinking of making a function like the following to replace the existing `make_..._string` functions:

```C++
template <typename Timepoint>
auto make_directive_string(int _delay_duration,
                           std::string_view _delay_time_unit,
                           std::string_view _directive,
                           const Timepoint _time) -> std::string
{
  return fmt::format("{}{} {} {}", _delay_duration, _delay_time_unit, _directive, std::chrono::system_clock::to_time_t(_time));
}
```

It should help reduce some duplicate code.

unit_tests/src/test_delay_hints_parser.cpp
MartinFlores751: In this case, I didn't write a function to generate this type of directive string, so no intermediate variable was needed.
We should probably add a function to generate these types of strings too though.

I'm also thinking it might also be worth avoiding our use of `std::string_view` in our input and output structs too, just to avoid excess intermediate variables as a result of object lifetime considerations.

korydraughn: Sure, that seems fine.

korydraughn:
Anything else to do for this PR?

MartinFlores751:
This should be it for this PR. If you feel anything else should be done, just let me know.

korydraughn:
If you're happy with the changes, then squash the commits, but leave off the pound sign so we can review it one last time.

trel (approved):

korydraughn (approved):
Add the #

Issue #6596, opened 2022-09-21
Should iRODS enforce always having one permission on collections and data objects 


korydraughn:
See original conversation starting here: https://github.com/irods/irods/issues/6579#issuecomment-1249208434

PR #6636, opened 2022-10-12
[#6596] Keep in-memory resource manager in sync with catalog (main)

korydraughn:


lib/administration/resource/include/irods/resource_administration.hpp
```diff
@@ -176,25 +224,89 @@ namespace irods::experimental::administration
...
+        template <typename ResourceProperty>
+            requires std::same_as<ResourceProperty, resource_type_property> ||
+                std::same_as<ResourceProperty, resource_type_property> ||
+                std::same_as<ResourceProperty, host_name_property> ||
+                std::same_as<ResourceProperty, vault_path_property> ||
+                std::same_as<ResourceProperty, resource_status_property> ||
+                std::same_as<ResourceProperty, resource_comments_property> ||
+                std::same_as<ResourceProperty, resource_info_property> ||
+                std::same_as<ResourceProperty, free_space_property> ||
+                std::same_as<ResourceProperty, context_string_property>
```
korydraughn, line 254: Consider making this a concept.

lib/administration/resource/include/irods/resource_administration.hpp
```diff
@@ -41,25 +53,24 @@ namespace irods::experimental::administration
...
+        inline constexpr std::string_view deferred       = "deferred";
+        inline constexpr std::string_view load_balanced  = "load_balanced";
+        inline constexpr std::string_view mockarchive    = "mockarchive";
+        inline constexpr std::string_view nonblocking    = "nonblocking";
+        inline constexpr std::string_view passthrough    = "passthru";
+        inline constexpr std::string_view random         = "random";
+        inline constexpr std::string_view replication    = "replication";
+        inline constexpr std::string_view struct_file    = "structfile";
+        inline constexpr std::string_view universal_mss  = "univmss";
+        inline constexpr std::string_view unixfilesystem = "unixfilesystem";
```
korydraughn, line 73: Consider making these `const char* const`.

string_view is great for certain things, but I'm not leaning more to them only being useful as function parameter types and in the impls of functions.

lib/administration/resource/include/irods/resource_administration.hpp
```diff
@@ -126,6 +136,46 @@ namespace irods::experimental::administration
...
+
+    struct free_space_property
+    {
+        std::string value;
+    }; // struct free_space_property
+
+    struct context_string_property
+    {
+        std::string value;
+    }; // struct context_string_property
```
korydraughn, line 177: Document these.

lib/administration/resource/include/irods/resource_administration.hpp
```diff
@@ -176,25 +224,89 @@ namespace irods::experimental::administration
...
+            }
+            else if constexpr (std::is_same_v<ResourceProperty, context_string_property>) {
+                input.arg3 = "context";
+                input.arg4 = _property.value.data();
+            }
+
+            if (const auto ec = rxGeneralAdmin(&_comm, &input); ec < 0) {
+                constexpr char* msg = "Could not change [{}] to [{}] for resource [{}]";
+                // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+                THROW(ec, fmt::format(msg, input.arg3, input.arg4, _name));
```
korydraughn, line 307: Introduce a compile-time `to_string` function for each property type. This error message isn't very helpful without knowing what kind of property is being changed.

lib/administration/resource/include/irods/resource_administration.hpp
```diff
@@ -176,25 +224,89 @@ namespace irods::experimental::administration
...
+            requires std::same_as<ResourceProperty, resource_type_property> ||
+                std::same_as<ResourceProperty, resource_type_property> ||
+                std::same_as<ResourceProperty, host_name_property> ||
+                std::same_as<ResourceProperty, vault_path_property> ||
+                std::same_as<ResourceProperty, resource_status_property> ||
+                std::same_as<ResourceProperty, resource_comments_property> ||
+                std::same_as<ResourceProperty, resource_info_property> ||
+                std::same_as<ResourceProperty, free_space_property> ||
+                std::same_as<ResourceProperty, context_string_property>
+        auto modify_resource(RxComm& _comm, const resource_name_type _name, const ResourceProperty& _property) -> void
```
korydraughn, line 255: Don't forget to remove all other free functions if we go with style.

plugins/database/src/db_plugin.cpp
```diff
@@ -3997,6 +3997,7 @@ irods::error db_del_child_resc_op(
 
     irods::sql_logger logger( "chlDelChildResc", logSQL );
     std::string child_string( resc_input[irods::RESOURCE_CHILDREN] );
+    log_db::info("child_string=[{}]", child_string);
```
korydraughn, line 4000: Remove or make it debug/trace.

plugins/database/src/db_plugin.cpp
```diff
@@ -4005,6 +4006,7 @@ irods::error db_del_child_resc_op(
               parent_name,
               parent_resource_id,
               parent_parent_resource_id);
+    log_db::info("parent_name=[{}], parent_resource_id=[{}], parent_parent_resource_id=[{}]", parent_name, parent_resource_id, parent_parent_resource_id);
```
korydraughn, line 4009: Remove or make it debug/trace.

plugins/database/src/db_plugin.cpp
```diff
@@ -4014,18 +4016,21 @@ irods::error db_del_child_resc_op(
 
     std::string child_name;
     parser.first_child( child_name );
+    log_db::info("child_name=[{}]", child_name);
```
korydraughn, line 4019: Remove or make it debug/trace.

plugins/database/src/db_plugin.cpp
```diff
@@ -4014,18 +4016,21 @@ irods::error db_del_child_resc_op(
...
 
     std::string child_resource_id;
     std::string child_parent_resource_id;
     ret = extract_resource_properties_for_operations(
               child_name,
               child_resource_id,
               child_parent_resource_id);
+    log_db::info("child_name=[{}], child_resource_id=[{}], child_parent_resource_id=[{}]", child_name, child_resource_id, child_parent_resource_id);
```
korydraughn, line 4027: Remove or make it debug/trace.

plugins/database/src/db_plugin.cpp
```diff
@@ -4014,18 +4016,21 @@ irods::error db_del_child_resc_op(
...
     if(!ret.ok()) {
            return PASS(ret);
     }
 
     if (child_parent_resource_id != parent_resource_id) {
+        log_db::error("child_parent_resource_id=[{}], parent_resource_id=[{}]", child_parent_resource_id, parent_resource_id);
```
korydraughn, line 4033: Remove or make it debug/trace.

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -431,18 +432,27 @@ int _addChildToResource(generalAdminInp_t* _generalAdminInp, rsComm_t* _rsComm)
     log_api::info(
         R"__(rsGeneralAdmin add child "{}" to resource "{}")__", _generalAdminInp->arg3, _generalAdminInp->arg2);
 
-    if ( ( result = chlAddChildResc( _rsComm, resc_input ) ) != 0 ) {
+    if ((result = chlAddChildResc(_rsComm, resc_input)) == 0) {
+        try {
+            log_api::info("ADDING CHILD [{}] TO RESOURCE [{}] ::: rescChildren = [{}]", _generalAdminInp->arg3, _generalAdminInp->arg2, rescChildren);
```
korydraughn, line 437: Remove

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -431,18 +432,27 @@ int _addChildToResource(generalAdminInp_t* _generalAdminInp, rsComm_t* _rsComm)
...
+            resc_mgr.add_child_to_resource(_generalAdminInp->arg2, _generalAdminInp->arg3);
```
korydraughn, line 438: What should happen if this fails? Should everything be rolled back?

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -468,29 +478,37 @@ _removeChildFromResource(
...
 
-    rodsLog( LOG_NOTICE, "rsGeneralAdmin remove child \"%s\" from resource \"%s\"", resc_input[irods::RESOURCE_CHILDREN].c_str(),
-             resc_input[irods::RESOURCE_NAME].c_str() );
+    log_api::info("Removing child [{}] from resource [{}]", resc_input[irods::RESOURCE_CHILDREN],
+             resc_input[irods::RESOURCE_NAME]);
 
-    if ( ( result = chlDelChildResc( _rsComm, resc_input ) ) != 0 ) {
+    if ((result = chlDelChildResc(_rsComm, resc_input)) == 0) {
+        try {
+            log_api::info("REMOVING CHILD [{}] FROM RESOURCE [{}]", _generalAdminInp->arg3, _generalAdminInp->arg2);
```
korydraughn, line 486: Remove

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -468,29 +478,37 @@ _removeChildFromResource(
...
+            resc_mgr.remove_child_from_resource(resc_input.at(irods::RESOURCE_NAME),
+                                                resc_input.at(irods::RESOURCE_CHILDREN));
```
korydraughn, line 488: What should happen if this fails? Should everything be rolled back?

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -588,39 +598,110 @@ _addResource(
...
-    // =-=-=-=-=-=-=-
-    // register resource with the metadata catalog, roll back on an error
-    else if ( ( result = chlRegResc( _rsComm, resc_input ) ) != 0 ) {
-        chlRollback( _rsComm );
+        log_api::error("rsGeneralAdmin: acPreProcForCreateResource error for {}, stat={}",
+            resc_input[irods::RESOURCE_NAME], result);
     }
+    // register resource with the metadata catalog, roll back on an error
+    else if ((result = chlRegResc(_rsComm, resc_input)) == 0) {
+        log_api::info("ADDING NEW RESOURCE [{}]", resc_input.at(irods::RESOURCE_NAME));
```
korydraughn, line 612: Remove

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -588,39 +598,110 @@ _addResource(
...
+        if (const auto err = resc_mgr.init_from_catalog(*_rsComm, resc_input.at(irods::RESOURCE_NAME)); !err.ok()) {
+            log_api::error(err.result());
+        }
```
korydraughn, line 615: What should happen if this fails? Should everything be rolled back?

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -588,39 +598,110 @@ _addResource(
...
+
+        if (0 == ec) {
+            log_api::info("DRYRUN REMOVING RESOURCE [{}] :: SUCCESS", _input.arg2);
+        }
+        else {
+            log_api::info("DRYRUN REMOVING RESOURCE [{}] :: FAILURE", _input.arg2);
+        }
+
+        return ec;
+    }
```
korydraughn, line 662: Is this correct?

We return before making any call to rollback. Perhaps the impl is relying on the ODBC driver to automatically issue a rollback? I hope that isn't a thing.

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -588,39 +598,110 @@ _addResource(
...
+
+        log_api::error("rsGeneralAdmin: acPreProcForDeleteResource error for {}, stat={}", resc_name, i);
+
+        return i;
+    }
+
+    const auto ec = chlDelResc(&_comm, resc_name.data());
+
+    if (ec == 0) {
+        log_api::info(">>>>>>>>>>>>>>>>> SHUTTING DOWN RESOURCE [{}]", resc_name);
```
korydraughn, line 680: Remove

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -588,39 +598,110 @@ _addResource(
...
+        resc_mgr.shut_down_resource(resc_name);
+        log_api::info(">>>>>>>>>>>>>>>>> ERASING RESOURCE [{}] FROM RESC_MGR", resc_name);
```
korydraughn, line 682: Remove

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -588,39 +598,110 @@ _addResource(
...
+        resc_mgr.erase_resource(resc_name);
```
korydraughn, line 683: What happens if either of these fail? Should things be rolled back?

server/core/include/irods/irods_resource_manager.hpp
```diff
@@ -1,296 +1,289 @@
...
+        /// \param[in] _comm
+        /// \param[in] _resc_name
+        ///
+        /// \return An iRODS error object containing the status of the operation.
+        error init_from_catalog(RsComm& _comm, const std::string_view _resc_name);
+
+        /// call shutdown on resources before destruction
+        error shut_down_resources();
+
+        /// TODO
```
korydraughn, line 68: Documentation

server/core/include/irods/irods_resource_manager.hpp
```diff
@@ -1,296 +1,289 @@
...
+        error shut_down_resource(const std::string_view _resc_name);
+
+        /// TODO
```
korydraughn, line 71: Documentation

server/core/include/irods/irods_resource_manager.hpp
```diff
@@ -1,296 +1,289 @@
...
+                return SUCCESS();
+            }
 
-                    }
+            constexpr char* msg = "failed to find resource for property [{}] and value";
+            return ERROR(SYS_RESC_DOES_NOT_EXIST, fmt::format(msg, _key));
+        } // resolve_from_property
 
-                } // for itr
+        /// TODO
```
korydraughn, line 217: Documentation

server/core/include/irods/irods_resource_manager.hpp
```diff
@@ -1,296 +1,289 @@
...
+        void remove_child_from_resource(rodsLong_t _parent_resc_id, rodsLong_t _child_resc_id);
+
+        /// TODO
+        void remove_child_from_resource(const std::string_view _parent_resc_name, const std::string_view _child_resc_name);
+
+        /// TODO
+        void erase_resource(rodsLong_t _resc_id);
+
+        /// TODO
+        void erase_resource(const std::string_view _resc_name);
```
korydraughn, line 233: Documentation

server/core/include/irods/irods_resource_manager.hpp
```diff
@@ -1,296 +1,289 @@
...
-            /// @brief given a resource name get the parent name from the id
-            error get_parent_name( resource_ptr, std::string& );
-
-            // =-=-=-=-=-=-=-
-            // Attributes
-            lookup_table< resource_ptr >                        resource_name_map_;
-            lookup_table< resource_ptr, long, std::hash<long> > resource_id_map_;
-            std::vector< std::vector< pdmo_type > > maintenance_operations_;
+    private:
+        /// TODO
```
korydraughn, line 246: Documentation

server/core/include/irods/irods_resource_manager.hpp
```diff
@@ -1,296 +1,289 @@
...
+        void init_resource_from_database_row(const std::vector<std::string>& _row);
 
+        /// Initialize the child map from the resources lookup table
+        error init_parent_child_relationships();
+
+        /// TODO
```
korydraughn, line 252: Documentation

server/core/include/irods/irods_resource_manager.hpp
```diff
@@ -1,296 +1,289 @@
...
+        error init_parent_child_relationship(const std::string_view _resc_name);
+
+        void init_parent_child_relationship_impl(const std::string_view _resc_name, resource_ptr _resc_ptr);
+
+        /// top level function to gather the post disconnect maintenance 
+        /// operations from the resources, in breadth first order
+        error gather_operations();
+
+        /// TODO
```
korydraughn, line 261: Documentation

server/core/include/irods/irods_resource_manager.hpp
```diff
@@ -1,296 +1,289 @@
...
+        error gather_operations(const std::string_view _resc_name);
+
+        /// top level function to call the start operation on the resource plugins
+        error start_resource_plugins();
+
+        /// TODO
```
korydraughn, line 267: Documentation

server/core/src/irods_resource_manager.cpp
```diff
@@ -13,226 +14,213 @@
...
-        else {
-            std::stringstream msg;
-            msg << "failed to find resource for path [";
-            msg << _physical_path;
-            msg << "]";
-            return ERROR( SYS_INVALID_INPUT_PARAM, msg.str() );
-        }
 
+        // TODO
+        return ERROR(SYS_INVALID_INPUT_PARAM, fmt::format("failed to find resource for path [{}]", _physical_path));
```
korydraughn, line 163: Handle this

server/core/src/irods_resource_manager.cpp
```diff
@@ -13,226 +14,213 @@
...
+        resc_ptr->set_property<std::string>(RESOURCE_TYPE,           type);
+        resc_ptr->set_property<std::string>(RESOURCE_CLASS,          _row.at(4));
+        resc_ptr->set_property<std::string>(RESOURCE_PATH,           _row.at(6));
+        resc_ptr->set_property<std::string>(RESOURCE_INFO,           _row.at(8));
+        resc_ptr->set_property<std::string>(RESOURCE_COMMENTS,       _row.at(9));
+        resc_ptr->set_property<std::string>(RESOURCE_CREATE_TS,      _row.at(10));
+        resc_ptr->set_property<std::string>(RESOURCE_MODIFY_TS,      _row.at(11));
+        resc_ptr->set_property<std::string>(RESOURCE_CHILDREN,       _row.at(13));
+        resc_ptr->set_property<std::string>(RESOURCE_CONTEXT,        context);
+        resc_ptr->set_property<std::string>(RESOURCE_PARENT,         _row.at(15));
```
korydraughn, line 216: We need to find a better way to deal with the resource id and parent. They are both integer values, but one is stored as an integer while the other as a string.

Perhaps we should move away from map-like representations for certain things.

The following is easier to handle as compared to a map-like type.
```cpp
struct resource
{
    int id;
    optional<int> parent_id;
    string name;
    string location;
    string vault_path;
};
```
This makes even more sense given that this information is strictly for server-side use.

unit_tests/cmake/test_config/irods_resource_administration.cmake
```diff
@@ -3,7 +3,9 @@ set(IRODS_TEST_TARGET irods_resource_administration)
...
                             ${CMAKE_CURRENT_SOURCE_DIR}/src/test_resource_administration.cpp)
 
-set(IRODS_TEST_INCLUDE_PATH ${IRODS_EXTERNALS_FULLPATH_BOOST}/include)
+set(IRODS_TEST_INCLUDE_PATH ${IRODS_EXTERNALS_FULLPATH_BOOST}/include
+                            ${IRODS_EXTERNALS_FULLPATH_FMT}/include)
  
 set(IRODS_TEST_LINK_LIBRARIES irods_common
-                              irods_client)
+                              irods_client
+                              ${IRODS_EXTERNALS_FULLPATH_FMT}/lib/libfmt.so)
```
korydraughn, line 11: Add new line for unix tools

unit_tests/cmake/test_config/irods_zone_report.cmake
```diff
@@ -3,7 +3,8 @@ set(IRODS_TEST_TARGET irods_zone_report)
 set(IRODS_TEST_SOURCE_FILES ${CMAKE_CURRENT_SOURCE_DIR}/src/main.cpp
                             ${CMAKE_CURRENT_SOURCE_DIR}/src/test_zone_report.cpp)
 
-set(IRODS_TEST_INCLUDE_PATH ${IRODS_EXTERNALS_FULLPATH_BOOST}/include)
+set(IRODS_TEST_INCLUDE_PATH ${IRODS_EXTERNALS_FULLPATH_BOOST}/include
+                            ${IRODS_EXTERNALS_FULLPATH_FMT}/include)
  
 set(IRODS_TEST_LINK_LIBRARIES irods_common
                               irods_client)
```
korydraughn, line 10: Add new line for unix tools

unit_tests/src/unit_test_utils.hpp
```diff
@@ -78,7 +78,9 @@ namespace unit_test_utils
         ufs_info.host_name = host_name;
         ufs_info.vault_path = vault_path;
 
-        return adm::client::add_resource(_comm, ufs_info).value() == 0;
+        adm::client::add_resource(_comm, ufs_info);
+
+        return true; // TODO
```
korydraughn, line 83: Handle this

.clang-tidy
```diff
@@ -28,6 +28,7 @@ Checks:
     hicpp-signed-bitwise,
     misc-*,
     modernize-*,
+    -modernize-use-trailing-return-type,
```
korydraughn, line 31: We need to consider whether this is important or not.

My immediate thought is, trailing return types are good for keeping your function names aligned. There are certainly reasons for using trailing return types in generic code sometimes. However, it's nothing wrong with leading return types. It's better to maintain a consistent style for each library/API IMO.

Will think about this more.

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -588,39 +598,110 @@ _addResource(
...
+    }
+
+    if (contains_whitespace(_input.arg2)) {
+        log_api::error("Whitespace is not allowed in resource names [{}].", _input.arg2);
+        return CAT_INVALID_RESOURCE_NAME;
+    }
+
+    const std::string_view resc_name = _input.arg2;
+
+    // add a dry run option to idamin rmresc.
```
trel, line 647: idamin -> iadmin

server/api/src/rsGeneralAdmin.cpp
trel: or we're not actually doing any work?

definitely don't want to rely on ODBC driver for a rollback, agreed.

server/core/src/irods_resource_manager.cpp
trel: agreed... but not quite sure what you mean with server-side use.

we definitely send this information to things like ilsresc and izonereport

server/core/src/irods_resource_manager.cpp
```diff
@@ -444,497 +405,575 @@ namespace irods
...
+        log_resc::info(err.result());
+        err = child_resc_ptr->set_property<std::string>(RESOURCE_PARENT, std::to_string(_parent_resc_id)); // TODO Can throw!!!
+        log_resc::info(err.result());
+    } // add_child_to_resource
+
+    void resource_manager::add_child_to_resource(
+        const std::string_view _parent_resc_name,
+        const std::string_view _child_resc_name)
+    {
+        log_resc::info(">>>>>>>>>>>>> IN {}", __func__);
```
trel, line 627: ew.

trel (commented):
a lot of work - so many return values no longer getting returned!

server/core/src/irods_resource_manager.cpp
korydraughn: The interface used to set/get info about a specific resource requires that you know the type of the object and requires verbose function calls to access it. For example:
```cpp
// (1) Lookup the resource pointer by resource name.
auto resc_ptr = resource_name_map[resc_name_string];

// (2) Extract the resource id.
rodsLong_t resc_id;
auto error = resc_ptr->get<rodsLong_t>(irods::RESOURCE_ID, resc_id);
if (!error.ok()) {
    // Something bad happened. Can't move forward.
}

// (3) We have the resource id. Do work!
```
The step of interest is (2). It takes several lines to fetch this information and can lead to time spent in the debugger if the type passed to `get<T>` is incorrect. This design allows arbitrary objects to be attached to the resource, but leads to bugs.

Given that there is a set of properties known to be attached to every resource, we can replace this generic design with a simpler type. Hence the example in my previous comment. The types are concrete and the dev doesn't have to risk getting stuck trying to figure out the type of the stored property.

This can be achieved without having to replace the original design. We can introduce a set of free functions that do the work for use. For example:
```cpp
auto resource_id(const resource& _r) -> rodsLong_t;
auto resource_parent_id(const resource& _r) -> std::optional<std::string>;
```
All of the work is now hidden behind a clean interface. I think this is the quickest and simplest solution for cleaning up resource manager/pointer code.

I'll go ahead and implement this and adjust the resource manager implementation to make use of it.

By server-side use, I mean the iRODS data types that store this info are only available and used in server-side code. We need to make accessing that info less error prone and verbose.

server/core/src/irods_resource_manager.cpp
korydraughn: Yes, a known leftover from debugging. These will not be part of the final PR.

lib/administration/resource/include/irods/resource_administration.hpp
korydraughn: done

lib/administration/resource/include/irods/resource_administration.hpp
korydraughn: Changed to `const char* const` for better flexibility.

lib/administration/resource/include/irods/resource_administration.hpp
korydraughn: Done

lib/administration/resource/include/irods/resource_administration.hpp
korydraughn: Done

lib/administration/resource/include/irods/resource_administration.hpp
korydraughn: Done

plugins/database/src/db_plugin.cpp
korydraughn: Done

plugins/database/src/db_plugin.cpp
korydraughn: Done

plugins/database/src/db_plugin.cpp
korydraughn: Done

plugins/database/src/db_plugin.cpp
korydraughn: Done

plugins/database/src/db_plugin.cpp
korydraughn: Done

server/api/src/rsGeneralAdmin.cpp
korydraughn: Done

server/api/src/rsGeneralAdmin.cpp
korydraughn: Done

server/api/src/rsGeneralAdmin.cpp
korydraughn: Done

server/api/src/rsGeneralAdmin.cpp
korydraughn: Done

server/api/src/rsGeneralAdmin.cpp
korydraughn: Done

server/core/src/irods_resource_manager.cpp
korydraughn: Done

unit_tests/src/unit_test_utils.hpp
korydraughn: done

korydraughn:
I need to review the unit tests I changed very carefully.

I also learned that a few of them expect the irods_environment.json file to define a default resource. May need to hard code those tests to `demoResc`.

lib/administration/resource/include/irods/resource_administration.hpp
```diff
@@ -126,7 +136,130 @@ namespace irods::experimental::administration
...
+    template <typename T>
+    concept ResourceProperty = std::same_as<T, resource_type_property> ||
+        std::same_as<T, resource_type_property> ||
+        std::same_as<T, host_name_property> ||
+        std::same_as<T, vault_path_property> ||
+        std::same_as<T, resource_status_property> ||
+        std::same_as<T, resource_comments_property> ||
+        std::same_as<T, resource_info_property> ||
+        std::same_as<T, free_space_property> ||
+        std::same_as<T, context_string_property>;
```
alanking, line 228: Concept!

Docs?

lib/administration/resource/include/irods/resource_administration.hpp
```diff
@@ -176,139 +307,110 @@ namespace irods::experimental::administration
...
         ///
         /// \returns A std::error_code.
-        auto rebalance_resource(RxComm& _comm, const resource_name_type _name) -> std::error_code;
+        auto rebalance_resource(RxComm& _comm, const resource_name_type _name) -> void;
 
-        /// \brief Retrieves the name of a resource given a resource ID.
+        /// Retrieves the name of a resource given a resource ID.
         ///
         /// \param[in] _comm The communication object.
         /// \param[in] _name The ID of the resource.
```
alanking, line 410:
```suggestion
        /// \param[in] _id The ID of the resource.
```
Needs some `\throws irods::exception` as well, I think

server/api/src/rsGeneralAdmin.cpp
alanking: It looks to me like it should.

alanking (commented):
Skimmed and caught a couple things, but we probably should go over this live at some point. Looks promising so far

korydraughn:
Thinking about keeping the resc_mgr in sync with the catalog has raised some concerns.

Primarily, is this problem solvable when agents aren't aware of resource hierarchy changes that happen via a sibling agent?

Agents always operate on a snapshot of the resource hierarchy. Some APIs check the catalog for resource info while others check the resc_mgr. It isn't always clear which source of truth should be used in a given operation.

Perhaps the answer is to make all agents consult a centralized resource service (which presents other issues - agents being blocked due to hierarchy updates; policy changing while transferring data).

korydraughn:
This PR is dead. There are too many challenges surrounding this effort and the latest improvements to the connection pool _(shipped in 4.3.1 and motivated by the HTTP API)_ prove to be a much better/safer approach.

See the following:
- https://github.com/irods/irods/issues/7287
- https://github.com/irods/irods/issues/7303
- https://github.com/irods/irods/issues/7304

Closing.

Issue #6624, opened 2022-10-07
iadmin fails to list user in case of particular user name length combinations


stsnel:
- [x] main (irods/irods)
- [x] 4-3-stable (irods/irods)
- [x] 4-2-stable (irods/irods)
- [x] main (irods/irods_client_icommands)
- [x] 4-3-stable (irods/irods_client_icommands)
- [x] 4-2-stable (irods/irods_client_icommands)
---
### iRODS Version, OS and Version

iRODS 4.2.11, with both CentOS 7.9.2009 and Ubuntu 18.04.6 LTS

### What did you try to do?

List users in the system:

```
iadmin lu
```

### Expected behavior

* iadmin lists all users
* no error messages in the rodsLog.

### Observed behavior (including steps to reproduce, if applicable)

One user account is missing from the iadmin output. 

rodsLog says:

```
Oct  7 15:05:34 pid:20271 remote addresses: 10.0.2.15, 127.0.0.1 ERROR: rstrcat not enough space in dest, slen:33, dlen:991, maxLen:1024, stacktrace:

Dumping stack trace
<0>	Offset:       	Address: 0x7ff8c9194baf	/usr/lib/libirods_common.so.4.2.11(rstrcat+0x6f) [0x7ff8c9194baf]
<1>	Offset: 0xd65 	Address: 0x7ff8bef53bd5	db_simple_query_op_vector(irods::plugin_context&, char const*, std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > > >, int, int*, char*, int)
<2>	Offset: 0x497 	Address: 0x7ff8bef55287	db_simple_query_op(irods::plugin_context&, char const*, char const*, char const*, char const*, char const*, int, int*, char*, int)
[...]

Issue #5247, opened 2020-11-13
remove simplequery


trel:
- [x] irods/irods (main)
- [x] python-irodsclient (confirmed no usage of SimpleQuery)
- [x] iCommands (confirmed no usage of SimpleQuery)
- [x] Jargon (created https://github.com/DICE-UNC/jargon/issues/503)
-----

little or nothing in the server still uses the simpleQuery service.

audit and remove all usage of simpleQuery - check core, icommands, python, java...

d-w-moore:
looking...

d-w-moore:
#6638 (fixed by @alanking ) is the same issue, correct?

d-w-moore:
...Or does this issue still exist (mentioned in description) that simplequery should be removed from other official repo's ?

korydraughn:
SimpleQuery has been deprecated.

This issue is about removing it from the codebase. We can wait a while before actually doing that though.

alanking:
As this was deprecated in 4.3.1, I think we should remove this in the next *major* release. Bumping to 5.0.0.

alanking:
Jargon has a lot of SimpleQuery stuff in it, so we will need to look into that.

python-irodsclient has the API number defined, but appears to offer no interface for SimpleQuery (which may be enough to call it done).

No mention of SimpleQuery in iCommands.

The docs only mention SimpleQuery in the now-removed API and DB plugin PEPs. Those will go away once we re-generate those.

alanking:
Created https://github.com/DICE-UNC/jargon/issues/503 for dealing with Jargon. Closing this one.

PR #6637, opened 2022-10-13
Add token create/modify times to GenQuery (main)

alanking:
Addresses #6624 

Companion PR: https://github.com/irods/irods_client_icommands/pull/331

Once merged, I think `rcSimpleQuery` can be deprecated. Should we handle this in a separate PR?


alanking:
clang-format basically wants to rewrite the entirety of `rodsGenQuery.h` and `rodsGenQueryNames.h`, so I've been ignoring it. Willing to run it if that's what we want to do.

korydraughn:
Ignore clang-format for this one.

> Once merged, I think `rcSimpleQuery` can be deprecated. Should we handle this in a separate PR?

Sure.

trel:
mark `rcSimpleQuery` with `[[deprecated]]`?

korydraughn:
That attribute form may not be supported by current C compilers. Most likely will have to use the old attribute form.

alanking:
Probably `__attribute__((deprecated("Use rcGenQuery or rcSpecificQuery instead.")))`?

I'm willing to do it here, might have to go to draft mode while I get that worked out. Should that at least be a separate issue, or should it just be lumped in with #6624 since it kind of brought this to our attention?

korydraughn:
The attribute form you referenced is correct.

Let's create a separate issue so it's easy to find later.

korydraughn:
You can link it to #6624 so we know they are related.

alanking:
I created https://github.com/irods/irods/issues/6638 for deprecation.

In the process, I found https://github.com/irods/irods/issues/5247 which will delete it.

trel:
Hey, that #5247 looking real smart.

scripts/irods/test/test_iadmin.py
```diff
@@ -1612,7 +1640,7 @@ def test_non_admins_are_not_allowed_to_invoke_iadmin_lg__issue_6188(self):
         self.user0.assert_icommand(['iadmin', 'lg'], 'STDERR', ['-830000 CAT_INSUFFICIENT_PRIVILEGE_LEVEL'])
 
     def test_non_admins_are_not_allowed_to_invoke_iadmin_set_delay_server__issue_5249(self):
-        self.user0.assert_icommand(['iadmin', 'set_delay_server', lib.get_hostname()], 'STDERR', ['-13000 SYS_NO_API_PRIV'])
+        self.user0.assert_icommand(['iadmin', 'set_delay_server', lib.get_hostname()], 'STDERR', ['-830000 CAT_INSUFFICIENT_PRIVILEGE_LEVEL'])
```
korydraughn, line 1643: Was this test just out of date?

korydraughn:
Is the plan to handle the deprecation in a separate PR? If so, then this PR seems ready to go.

scripts/irods/test/test_iadmin.py
alanking: The error code returned is actually changing here because `iadmin` is checking the user's privilege level and preventing `rodsuser`s from using it. This was necessary to maintain the "security" afforded by SimpleQuery since everything was changed to GenQuery.

alanking:
For purposes of getting something in, I think the deprecation can happen at another time.

PR #6643, opened 2022-10-14
[#6579] Mimic old `ichmod` behavior in filesystem unit test (main)

MartinFlores751:
Fix unit test relying on previous `ichmod` behavior by adding admin command where old behavior is expected.


trel:
i really like that this was as easy as adding the flag.

trel:
is clang-tidy just... wrong?

korydraughn:
I'm pretty sure it's because we don't run cmake with the option that instructs it to build the unit tests. A very easy tweak.

This PR can ignore the clang-tidy failure since it's not useful.

korydraughn (approved):
The change looks good to me.

alanking (approved):
Please add #

PR #6651, opened 2022-10-19
[5964] Correct DOUBLE directive unit tests (4-3-stable)

MartinFlores751:
It seems that for some DOUBLE directive tests, we were using the same string object passed to string views. This meant that when we set the expected results string, we also modified the input string. This may have been causing the non-deterministic results.


korydraughn (approved):

trel:
when cherry-picking - please leave the # off the first attempt - so we can see/approve it before it creates a backlink from the issue itself.

not a real problem, just a bit of polish on the process.

looks good to me.

alanking (approved):
Mergin on green

Issue #6638, opened 2022-10-13
Deprecate SimpleQuery


alanking:
- [x] main
- [x] 4-3-stable
---
To be removed in https://github.com/irods/irods/issues/5247

Includes a client API (`rcSimpleQuery`), server-side "shortcut" API call (`rsSimpleQuery`), and database high-level operation (`chlSimpleQuery`).

See #6624 for more details about how we got here...

alanking:
I think deprecating the database high-level operation is a bit much.

Do I need to add anything to do the docs, or will our note about this deprecation just be part of the release notes? There is not currently any mention of it in the docs that I have been able to find.

korydraughn:
Why do you feel deprecating the high-level op is too much?
Is it not specific to the simple query api?

I think a line in the release notes is good enough until we formalize more rules around deprecation.

trel:
We have had a deprecation section in the release notes for the last few cycles - so work that is done in an issue will have a home in the notes.

This is that issue - so I think we're good on that front.

alanking:
I guess I'm not sure where the line is for deprecating an API. I'm okay to deprecate `chlSimpleQuery`, but then I was wondering what other bits of machinery would need to be deprecated? I guess it all feeds back into the discussion around our deprecation model.

I figured that deprecating the user API and the internal server shortcut function would cover all known existing cases, and precisely *because* the `chl` operation is specific to SimpleQuery, it is kind of tied into the deprecation of the other bits. However, I can see the case for the other way.

Here's what I can find related to SimpleQuery:
 - `rcSimpleQuery`
 - `rsSimpleQuery`
 - `chlSimpleQuery`
 - `simpleQueryInp_t`
 - `simpleQueryOut_t`
 - `db_simple_query_op`
 - `DATABASE_OP_SIMPLE_QUERY`
 - `call_simpleQueryInp_simpleQueryOut`
 - `SIMPLE_QUERY_AN`(??)

Not sure if API numbers can be deprecated, but the rest of it seems like fair game. Thoughts?

korydraughn:
Marking all types and functions that are highly specific to an API seems appropriate to me. The line gets a bit more blurry when things start to appear in different APIs. For the things you listed, I think adding a deprecation attribute is good.

I'm not sure you can mark a macro as deprecated. If we can't, we can leave a comment stating it is and @trel can highlight that in the release notes.

trel:
Agreed

PR #6657, opened 2022-10-21
[#6638] Deprecate SimpleQuery (main)

alanking:
This change deprecates the use of SimpleQuery as well as its client API endpoint and server-side shortcut API.

---

The deprecated functions are no longer being used, so I didn't run the test suite.

### DIFF
--- a/lib/api/include/irods/simpleQuery.h
+++ b/lib/api/include/irods/simpleQuery.h
@@ -25,6 +25,7 @@ typedef struct {
 #ifdef __cplusplus
 extern "C"
 #endif
-int rcSimpleQuery( rcComm_t *conn, simpleQueryInp_t *simpleQueryInp, simpleQueryOut_t **simpleQueryOut );
+__attribute__((deprecated("SimpleQuery is deprecated. Use GenQuery or SpecificQuery instead.")))
+int rcSimpleQuery(rcComm_t* conn, simpleQueryInp_t* simpleQueryInp, simpleQueryOut_t** simpleQueryOut);
 
 #endif
--- a/server/api/src/rsSimpleQuery.cpp
+++ b/server/api/src/rsSimpleQuery.cpp
@@ -1,20 +1,56 @@
-/*** Copyright (c), The Regents of the University of California            ***
- *** For more information please refer to files in the COPYRIGHT directory ***/
-/* This is script-generated code (for the most part).  */
-/* See simpleQuery.h for a description of this API call.*/
-
-#include "irods/simpleQuery.h"
-#include "irods/rodsConnect.h"
-#include "irods/icatHighLevelRoutines.hpp"
-#include "irods/miscServerFunct.hpp"
-#include "irods/irods_configuration_keywords.hpp"
 #include "irods/rsSimpleQuery.hpp"
+
+#include "irods/apiNumber.h"
+#include "irods/icatHighLevelRoutines.hpp"
+#include "irods/irods_configuration_keywords.hpp"
+#include "irods/miscServerFunct.hpp"
+#include "irods/rodsConnect.h"
+
+namespace
+{
+    auto _rsSimpleQuery(rsComm_t* rsComm, simpleQueryInp_t* simpleQueryInp, simpleQueryOut_t** simpleQueryOut) -> int
+    {
+        int control = simpleQueryInp->control;
+        int maxBufSize = simpleQueryInp->maxBufSize;
+        char* outBuf = static_cast<char*>(malloc(maxBufSize));
+
+        int status = chlSimpleQuery(
+            rsComm,
+            simpleQueryInp->sql,
+            simpleQueryInp->arg1,
+            simpleQueryInp->arg2,
+            simpleQueryInp->arg3,
+            simpleQueryInp->arg4,
+            simpleQueryInp->form,
+            &control,
+            outBuf,
+            maxBufSize);
+
+        if (status < 0) {
+            if (status != CAT_NO_ROWS_FOUND) {
+                rodsLog(LOG_NOTICE, "_rsSimpleQuery: simpleQuery for %s, status = %d", simpleQueryInp->sql, status);
+            }
+            return status;
+        }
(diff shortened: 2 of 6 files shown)


server/api/src/rsSimpleQuery.cpp
```diff
@@ -42,8 +75,13 @@ rsSimpleQuery( rsComm_t *rsComm, simpleQueryInp_t *simpleQueryInp,
         }
     }
     else {
-        status = rcSimpleQuery( rodsServerHost->conn,
-                                simpleQueryInp, simpleQueryOut );
+        status = procApiRequest(
```
alanking, line 78: I can either turn off the deprecation warning here, or do this to get around having to use the `rc` interface. The question was raised in my mind at this point whether we should also be putting a warning message in the log and/or the `rError` stack?

lib/api/include/irods/simpleQuery.h
```diff
@@ -25,6 +25,8 @@ typedef struct {
 #ifdef __cplusplus
 extern "C"
 #endif
-int rcSimpleQuery( rcComm_t *conn, simpleQueryInp_t *simpleQueryInp, simpleQueryOut_t **simpleQueryOut );
+    int
+    rcSimpleQuery(rcComm_t* conn, simpleQueryInp_t* simpleQueryInp, simpleQueryOut_t** simpleQueryOut)
+        __attribute__((deprecated("SimpleQuery is deprecated. Use GenQuery or SpecificQuery instead.")));
```
alanking, line 30: Not sure about this formatting

server/api/src/rsSimpleQuery.cpp
trel: `rError` back to the user seems... helpful?    As long as we're still sending the same return code, their code shouldn't break...

Not convinced logging it on the server is as helpful.

alanking: After discussion, we want to just turn off the deprecation warning (a la apiTable).

Will add warning message to `rError`.

lib/api/include/irods/simpleQuery.h
alanking: After discussion, we decided that ignoring clang-format here is the way to go.

lib/api/include/irods/simpleQuery.h
```diff
@@ -25,8 +25,12 @@ typedef struct {
...
-    int
-    rcSimpleQuery(rcComm_t* conn, simpleQueryInp_t* simpleQueryInp, simpleQueryOut_t** simpleQueryOut)
-        __attribute__((deprecated("SimpleQuery is deprecated. Use GenQuery or SpecificQuery instead.")));
+
+    // clang-format off
+
+    __attribute__((deprecated("SimpleQuery is deprecated. Use GenQuery or SpecificQuery instead.")))
+    int rcSimpleQuery(rcComm_t* conn, simpleQueryInp_t* simpleQueryInp, simpleQueryOut_t** simpleQueryOut);
+
+    // clang-format on
```
alanking, line 34: Another formatting note... not sure why clang-format only wants one of these comments to be indented (namely, the one on 29). I think it should be both or neither.

korydraughn: Let's ignore indentation due to `extern "C"` here.

I really hate that clang-format wants to indent the function declaration.

alanking: Okay. In that case, I'm just going to remove the clang-format off/on comments and ignore clang-format :p

lib/api/include/irods/simpleQuery.h
```diff
@@ -25,6 +25,7 @@ typedef struct {
 #ifdef __cplusplus
 extern "C"
 #endif
-int rcSimpleQuery( rcComm_t *conn, simpleQueryInp_t *simpleQueryInp, simpleQueryOut_t **simpleQueryOut );
+__attribute__((deprecated("SimpleQuery is deprecated. Use GenQuery or SpecificQuery instead.")))
+int rcSimpleQuery(rcComm_t* conn, simpleQueryInp_t* simpleQueryInp, simpleQueryOut_t** simpleQueryOut);
```
korydraughn, line 29: I'm not sure of the order of `extern "C"` and attributes.

Consider using the braced form of `extern "C"` to make things clear. For example:
```cpp
#ifdef __cplusplus
extern "C" {
#endif

__attribute(...)
int rcFunc();

#ifdef __cplusplus
} // extern "C"
#endif
```

I guess one way to verify the `extern "C"` is honored is to use `readelf` or `objdump` to see if the function name is mangled. If it's not mangled, then we're good.

korydraughn (commented):
Looks good. Just need to verify the function isn't mangled.

The rError message was the right call too.

lib/api/include/irods/simpleQuery.h
alanking: Seems okay to me
```
$ objdump -t ~/hdd/builds/irods/main/ubuntu-18.04/irods/lib/libirods_client.so | grep rcSimpleQuery
0000000000000000 l    df *ABS*  0000000000000000              rcSimpleQuery.cpp
000000000016f710 g     F .text  000000000000003e              rcSimpleQuery
$ readelf -s ~/hdd/builds/irods/main/ubuntu-20.04/irods/lib/libirods_client.so | grep rcSimpleQuery
  5899: 00000000001700e0    62 FUNC    GLOBAL DEFAULT   11 rcSimpleQuery
  2631: 0000000000000000     0 FILE    LOCAL  DEFAULT  ABS rcSimpleQuery.cpp
 10977: 00000000001700e0    62 FUNC    GLOBAL DEFAULT   11 rcSimpleQuery
```

korydraughn: Good enough for me.

korydraughn:
Anything left for this?

alanking:
I think we're good to go

korydraughn:
Squash and add the #'s

Issue #6665, opened 2022-10-26
Administration libraries do not pass down include dirs properly


SwooshyCueb:

trel:
adding checked checkboxes, for easier forensics later.

PR #6666, opened 2022-10-26
[#6665] Pass down proper include dirs from administration libraries (main)

SwooshyCueb:
Addresses #6665 for main


korydraughn (approved):

alanking (approved):
Mergin on green

PR #6667, opened 2022-10-26
[#6665] Pass down proper include dirs from administration libraries (4-3-stable)

SwooshyCueb:
Addresses #6665 for 4-3-stable


korydraughn (approved):
Looks good to me.

If the github actions come back green, I think we can go ahead and merge these.

alanking (approved):
Mergin on green

Issue #6672, opened 2022-10-31
Remove responsibility of freeing heap allocated memory from packstruct on API response


korydraughn:
- [x] main
- [x] 4-3-stable
---
## Feature
The use of `FREE_POINTER` instructs packstruct to free the internal memory of `myOutStruct` (not the object pointed to by `myOutStruct`). However, this doesn't always work because of issues with the packstruct implementation.
https://github.com/irods/irods/blob/4cf8a7d3f4ce436648c6cfd7a7dc586407efb0ea/server/core/src/rsApiHandler.cpp#L312-L332
https://github.com/irods/irods/blob/4cf8a7d3f4ce436648c6cfd7a7dc586407efb0ea/server/core/src/rsApiHandler.cpp#L268-L270

For this reason, this feature of packstruct should not be used. Instead, freeing memory should happen outside of packstruct.

This means adjusting `apidef_t` and `api_entry` types so that they can hold a cleanup function that knows how to handle the output type. See the use of `clearInStruct` for example. This also means all API definitions must be updated to hold the correct cleanup function.
https://github.com/irods/irods/blob/4cf8a7d3f4ce436648c6cfd7a7dc586407efb0ea/lib/core/include/irods/apiHandler.hpp#L34-L62

The following location is where the cleanup function should be called.
https://github.com/irods/irods/blob/4cf8a7d3f4ce436648c6cfd7a7dc586407efb0ea/lib/core/include/irods/apiHandler.hpp#L347

The following models how the cleanup function would be invoked.
https://github.com/irods/irods/blob/4cf8a7d3f4ce436648c6cfd7a7dc586407efb0ea/server/core/src/rsApiHandler.cpp#L242-L250

By removing this responsibility from packstruct, this gets iRODS closer to supporting pluggable encoders/decoders (see #6489).

Issue #6429, opened 2022-06-02
Agent Factory respawn leads to orphaned agent factory and delay server


korydraughn:
### iRODS Version, OS and Version
pre-4.3.0, Ubuntu 20.04

### What did you try to do?
```bash
kill -9 <agent_factory_pid> # or kill -15

# Wait for the agent factory to respawn.
# Everything is fine. icommands work, etc.

# Try to shutdown the server.
./irodsctl -v stop
```

### Expected behavior
I expected the server to shutdown without any issues.

### Observed behavior (including steps to reproduce, if applicable)
All iRODS processes are shutdown except the agent factory. The agent factory is left as an orphan.

It gets worst.

If you do the following, not only is the agent factory orphaned on shutdown, the delay server becomes orphaned as well!
```bash
kill -9 <agent_factory_pid> # or kill -15

# Wait for the agent factory to respawn and kill it again.
kill -9 <respawned_agent_factory_pid>

# Shutting down the server now only terminates the main iRODS process
# leaving the agent factory and delay server as orphans.
./irodsctl -v stop
```

## Solution
Use `_exit(1)` instead of `exit(1)` in the agent factory's signal handler.

According to POSIX, if a child process does not call one of the exec functions or the exec function fails, `_exit()` should be used because `exit()` can result in modifications to the parent's address space.

Making this change resolves the issue entirely.

PR #6673, opened 2022-10-31
Address sanitizer options and fixes (main)

korydraughn:
TODOs:
- [x] Set ASan option to NO/OFF
- [x] Verify correct form of `exit()` to use
- [x] Verify main server process leaks are reported _(no known leaks in main server process)_
- [x] Verify delay server memory leaks are reported _(Highly depends on where the leak occurs)_
- [x] Run against test suite _(seen all unit/core tests pass)_
- [x] Assign issue numbers

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,6 +45,19 @@ set(CMAKE_MODULE_LINKER_FLAGS_INIT "-Wl,--enable-new-dtags -Wl,--as-needed")
 set(CMAKE_SHARED_LINKER_FLAGS_INIT "-Wl,--enable-new-dtags -Wl,--as-needed")
 
 if (IRODS_ENABLE_ADDRESS_SANITIZER)
+  # The following options are compiled into all binaries and allow them to start even
+  # when an ODR violation is detected.
+  #
+  # Make sure the correct llvm-symbolizer binary is available to Address Sanitizer. This binary
+  # allows debug symbols to be reported appropriately. There are two ways to do this:
+  #
+  #     export PATH=/opt/irods-externals/clang13.0.0-0/bin:$PATH
+  #
+  # - or -
+  #
+  #     export ASAN_SYMBOLIZER_PATH=/opt/irods-externals/clang13.0.0-0/bin/llvm-symbolizer
+  #
+  add_compile_definitions(IRODS_ADDRESS_SANITIZER_DEFAULT_OPTIONS="log_path=/tmp/irods_asan_output:detect_odr_violation=1")
   add_compile_options(
     -fsanitize=address
     -fno-omit-frame-pointer
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,7 @@ list(INSERT CMAKE_MODULE_PATH 0 "${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules")
 include(RequireOutOfSourceBuild)
 
 option(IRODS_DISABLE_COMPILER_OPTIMIZATIONS "Disables compiler optimizations by setting -O0." OFF)
+option(IRODS_ENABLE_ADDRESS_SANITIZER "Enables detection of memory leaks and other features provided by Address Sanitizer." OFF)
 
 if (NOT CMAKE_CONFIGURATION_TYPES AND NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE Release CACHE STRING "Choose the type of build {Debug, Release}." FORCE)
@@ -37,10 +38,31 @@ string(REPLACE ";" ", " IRODS_DEVELOP_DEPENDENCIES_STRING "${IRODS_DEVELOP_DEPEN
 set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_EXTENSIONS OFF)
 set(IRODS_CXX_STANDARD "${CMAKE_CXX_STANDARD}")
+
 # export-dynamic so stacktrace entries from executables have function names
-set(CMAKE_EXE_LINKER_FLAGS_INIT "-Wl,--export-dynamic -Wl,--enable-new-dtags -Wl,--as-needed -Wl,-z,defs")
-set(CMAKE_MODULE_LINKER_FLAGS_INIT "-Wl,--enable-new-dtags -Wl,--as-needed -Wl,-z,defs")
-set(CMAKE_SHARED_LINKER_FLAGS_INIT "-Wl,--enable-new-dtags -Wl,--as-needed -Wl,-z,defs")
+set(CMAKE_EXE_LINKER_FLAGS_INIT "-Wl,--export-dynamic -Wl,--enable-new-dtags -Wl,--as-needed")
+set(CMAKE_MODULE_LINKER_FLAGS_INIT "-Wl,--enable-new-dtags -Wl,--as-needed")
+set(CMAKE_SHARED_LINKER_FLAGS_INIT "-Wl,--enable-new-dtags -Wl,--as-needed")
+
+if (IRODS_ENABLE_ADDRESS_SANITIZER)
+  add_compile_options(
+    -fsanitize=address
+    -fno-omit-frame-pointer
+    -fno-optimize-sibling-calls
+    -O1
+  )
+  add_link_options(
+    -fsanitize=address
+    -fno-omit-frame-pointer
+    -fno-optimize-sibling-calls
+    -O1
+  )
+else()
+  set(CMAKE_EXE_LINKER_FLAGS_INIT "${CMAKE_EXE_LINKER_FLAGS_INIT} -Wl,-z,defs")
+  set(CMAKE_MODULE_LINKER_FLAGS_INIT "${CMAKE_MODULE_LINKER_FLAGS_INIT} -Wl,-z,defs")
+  set(CMAKE_SHARED_LINKER_FLAGS_INIT "${CMAKE_SHARED_LINKER_FLAGS_INIT} -Wl,-z,defs")
+endif()
+
 set(CMAKE_EXE_LINKER_FLAGS_RELEASE_INIT "-Wl,--gc-sections -Wl,-z,combreloc")
 set(CMAKE_MODULE_LINKER_FLAGS_RELEASE_INIT "-Wl,--gc-sections -Wl,-z,combreloc")
 set(CMAKE_SHARED_LINKER_FLAGS_RELEASE_INIT "-Wl,--gc-sections -Wl,-z,combreloc")
--- a/lib/api/include/irods/data_object_finalize.h
+++ b/lib/api/include/irods/data_object_finalize.h
@@ -3,8 +3,6 @@
 
 /// \file
 
-const char* SET_TINE_TO_NOW_KW = "set time to now";
-
 struct RcComm;
 
 #ifdef __cplusplus
--- a/lib/api/include/irods/fileStat.h
+++ b/lib/api/include/irods/fileStat.h
@@ -7,7 +7,8 @@
 struct RcComm;
 struct rodsStat;
 
-typedef struct FileStatInp {
+typedef struct FileStatInp
+{
     rodsHostAddr_t addr;
     char fileName[MAX_NAME_LEN];
     char rescHier[MAX_NAME_LEN];
@@ -18,8 +19,13 @@ typedef struct FileStatInp {
 #define fileStatInp_PI "struct RHostAddr_PI; str fileName[MAX_NAME_LEN]; str rescHier[MAX_NAME_LEN]; str objPath[MAX_NAME_LEN]; double rescId;"
 
 #ifdef __cplusplus
-extern "C"
+extern "C" {
 #endif
+
 int rcFileStat(struct RcComm* conn, fileStatInp_t* fileStatInp, struct rodsStat** fileStatOut);
+
+#ifdef __cplusplus
+} // extern "C"
+#endif
 
 #endif // IRODS_FILE_STAT_H
--- a/lib/api/include/irods/fileStat.h
+++ b/lib/api/include/irods/fileStat.h
@@ -1,21 +1,25 @@
-#ifndef FILE_STAT_H__
-#define FILE_STAT_H__
+#ifndef IRODS_FILE_STAT_H
+#define IRODS_FILE_STAT_H
 
 #include "irods/rodsDef.h"
-#include "irods/rcConnect.h"
+#include "irods/rodsType.h"
 
-typedef struct {
+struct RcComm;
+struct rodsStat;
+
+typedef struct FileStatInp {
     rodsHostAddr_t addr;
     char fileName[MAX_NAME_LEN];
     char rescHier[MAX_NAME_LEN];
     char objPath[MAX_NAME_LEN];
     rodsLong_t rescId;
 } fileStatInp_t;
+
 #define fileStatInp_PI "struct RHostAddr_PI; str fileName[MAX_NAME_LEN]; str rescHier[MAX_NAME_LEN]; str objPath[MAX_NAME_LEN]; double rescId;"
 
 #ifdef __cplusplus
 extern "C"
 #endif
-int rcFileStat( rcComm_t *conn, fileStatInp_t *fileStatInp, rodsStat_t **fileStatOut );
+int rcFileStat(struct RcComm* conn, fileStatInp_t* fileStatInp, struct rodsStat** fileStatOut);
 
-#endif
+#endif // IRODS_FILE_STAT_H
--- a/lib/api/include/irods/getMiscSvrInfo.h
+++ b/lib/api/include/irods/getMiscSvrInfo.h
@@ -21,8 +21,13 @@ typedef struct MiscSvrInfo {
 #define MiscSvrInfo_PI "int serverType; int serverBootTime; str relVersion[NAME_LEN]; str apiVersion[NAME_LEN]; str rodsZone[NAME_LEN];"
 
 #ifdef __cplusplus
-extern "C"
+extern "C" {
 #endif
-int rcGetMiscSvrInfo(rcComm_t *conn, miscSvrInfo_t **outSvrInfo );
+
+int rcGetMiscSvrInfo(rcComm_t* conn, miscSvrInfo_t** outSvrInfo);
+
+#ifdef __cplusplus
+} // extern "C"
+#endif
 
 #endif // IRODS_GET_MISC_SVR_INFO_H
--- a/lib/api/include/irods/getMiscSvrInfo.h
+++ b/lib/api/include/irods/getMiscSvrInfo.h
@@ -1,5 +1,5 @@
-#ifndef GET_MISC_SVR_INFO_H__
-#define GET_MISC_SVR_INFO_H__
+#ifndef IRODS_GET_MISC_SVR_INFO_H
+#define IRODS_GET_MISC_SVR_INFO_H
 
 #include "irods/rcConnect.h"
 
@@ -17,11 +17,12 @@ typedef struct MiscSvrInfo {
     char apiVersion[NAME_LEN];    // the API version number
     char rodsZone[NAME_LEN];      // the zone of this server
 } miscSvrInfo_t;
+
 #define MiscSvrInfo_PI "int serverType; int serverBootTime; str relVersion[NAME_LEN]; str apiVersion[NAME_LEN]; str rodsZone[NAME_LEN];"
 
 #ifdef __cplusplus
 extern "C"
 #endif
-int rcGetMiscSvrInfo( rcComm_t *conn, miscSvrInfo_t **outSvrInfo );
+int rcGetMiscSvrInfo(rcComm_t *conn, miscSvrInfo_t **outSvrInfo );
 
-#endif
+#endif // IRODS_GET_MISC_SVR_INFO_H
--- a/lib/core/include/irods/apiHandler.hpp
+++ b/lib/core/include/irods/apiHandler.hpp
@@ -1,5 +1,5 @@
-#ifndef API_HANDLER_HPP
-#define API_HANDLER_HPP
+#ifndef IRODS_API_HANDLER_HPP
+#define IRODS_API_HANDLER_HPP
 
 #include "irods/rods.h"
 #include "irods/packStruct.h"
@@ -28,9 +28,6 @@
 
 namespace irods
 {
-    // NOOP function for clearInStruct
-    void clearInStruct_noop( void* );
-
     struct apidef_t {
         // =-=-=-=-=-=-=-
         // attributes
@@ -56,7 +53,8 @@ namespace irods
 
         const char *   operation_name;
 
-        std::function<void( void* )> clearInStruct;	// free input struct function
+        std::function<void(void*)> clearInStruct;	// free input struct function
+        std::function<void(void*)> clearOutStruct;	// free output struct function
 
         int(*call_wrapper)(...);        // wraps the api call for type casting
     }; // struct apidef_t
@@ -306,17 +304,14 @@ namespace irods
             return 0;
         } // call_handler_without_policy
 
-        // =-=-=-=-=-=-=-
         // ctors
         api_entry( apidef_t& );
 
         api_entry( const api_entry& );
 
-        // =-=-=-=-=-=-=-
         // operators
         api_entry& operator=( const api_entry& );
 
-        // =-=-=-=-=-=-=-
         // attributes
         int            apiNumber;      /* the API number */
         char*          apiVersion;     /* The API version of this call */
@@ -344,7 +339,8 @@ namespace irods
 
         lookup_table< std::string>   extra_pack_struct;
 
-        std::function<void( void* )> clearInStruct;		//free input struct function
+        std::function<void(void*)> clearInStruct;   // free input struct function
+        std::function<void(void*)> clearOutStruct;  // free output struct function
 
     private:
 #ifdef ENABLE_RE
@@ -425,6 +421,6 @@ namespace irods
         api_entry_table&  _api_tbl,    // table holding api entries
         pack_entry_table& _pack_tbl,   // table for pack struct ref
         bool              _cli_flg = true ); // default to client
-}; // namespace irods
+} // namespace irods
 
-#endif          /* API_HANDLER_H */
+#endif // IRODS_API_HANDLER_HPP
--- a/lib/core/include/irods/apiHandler.hpp
+++ b/lib/core/include/irods/apiHandler.hpp
@@ -53,8 +53,8 @@ namespace irods
 
         const char *   operation_name;
 
-        std::function<void(void*)> clearInStruct;	// free input struct function
-        std::function<void(void*)> clearOutStruct;	// free output struct function
+        std::function<void(void*)> clearInStruct; // free input struct function
+        std::function<void(void*)> clearOutStruct; // free output struct function
 
         int(*call_wrapper)(...);        // wraps the api call for type casting
     }; // struct apidef_t
@@ -71,11 +71,12 @@ namespace irods
     }
 
     template <typename... types_t>
-    class api_call_adaptor {
-    private:
+    class api_call_adaptor
+    {
+      private:
         std::function<int(rsComm_t*, types_t...)> fcn_;
 
-    public:
+      public:
         api_call_adaptor( std::function<int(rsComm_t*, types_t...)> _fcn )
             : fcn_(_fcn)
         {
@@ -134,10 +135,10 @@ namespace irods
 
     class api_entry : public irods::plugin_base
     {
-    private:
+      private:
         using rule_engine_context_manager_type = rule_engine_context_manager<unit, ruleExecInfo_t*, AUDIT_RULE>;
 
-    public:
+      public:
         /// =-=-=-=-=-=-=-
         /// @brief adaptor from old api sig to new plugin sig
         template<typename... types_t>
@@ -339,10 +340,10 @@ namespace irods
 
         lookup_table< std::string>   extra_pack_struct;
 
-        std::function<void(void*)> clearInStruct;   // free input struct function
-        std::function<void(void*)> clearOutStruct;  // free output struct function
+        std::function<void(void*)> clearInStruct; // free input struct function
+        std::function<void(void*)> clearOutStruct; // free output struct function
 
-    private:
+      private:
 #ifdef ENABLE_RE
         template<typename... types_t>
         error invoke_policy_enforcement_point(
@@ -405,13 +406,13 @@ namespace irods
     class api_entry_table
         : public lookup_table<api_entry_ptr, size_t, boost::hash<size_t>>
     {
-    public:
+      public:
         api_entry_table(apidef_t defs[], size_t size);
 
         auto is_plugin_loaded(std::string_view plugin_name) -> bool;
         auto mark_plugin_as_loaded(std::string_view plugin_name) -> void;
 
-    private:
+      private:
         std::vector<std::string> loaded_plugins_;
     }; // class api_entry_table
 
--- a/lib/core/include/irods/authentication_plugin_framework.hpp
+++ b/lib/core/include/irods/authentication_plugin_framework.hpp
@@ -281,4 +281,3 @@ namespace irods::experimental::auth
 } // namespace irods::experimental::auth
 
 #endif // IRODS_AUTHENTICATION_PLUGIN_FRAMEWORK_HPP
-
--- a/lib/core/include/irods/authentication_plugin_framework.hpp
+++ b/lib/core/include/irods/authentication_plugin_framework.hpp
@@ -3,6 +3,7 @@
 
 #include "irods/authenticate.h"
 #include "irods/getRodsEnv.h"
+#include "irods/irods_at_scope_exit.hpp"
 #include "irods/irods_auth_constants.hpp"
 #include "irods/irods_exception.hpp"
 #include "irods/irods_load_plugin.hpp"
@@ -16,6 +17,7 @@
 #include <nlohmann/json.hpp>
 
 #include <functional>
+#include <memory>
 #include <string>
 #include <string_view>
 #include <vector>
@@ -110,7 +112,7 @@ namespace irods::experimental::auth
     /// \returns The authentication plugin object (descendant of \p authentication_base).
     ///
     /// \since 4.3.0
-    auto resolve_authentication_plugin(const std::string& _scheme, const std::string& _type)
+    inline auto resolve_authentication_plugin(const std::string& _scheme, const std::string& _type)
     {
         using plugin_type = authentication_base;
 
@@ -156,15 +158,16 @@ namespace irods::experimental::auth
     /// \endparblock
     ///
     /// \since 4.3.0
-    void authenticate_client(RcComm& _comm, const RodsEnvironment& _env, const json& _ctx)
+    inline void authenticate_client(RcComm& _comm, const RodsEnvironment& _env, const json& _ctx)
     {
         std::string scheme = _env.rodsAuthScheme;
 
-        auto auth = resolve_authentication_plugin(scheme, "client");
+        std::unique_ptr<authentication_base> auth{resolve_authentication_plugin(scheme, "client")};
 
         const std::string* next_operation = &irods::AUTH_CLIENT_START;
 
-        json req{_ctx}, resp{};
+        json req{_ctx};
+        json resp;
 
         req["scheme"] = scheme;
         req[auth::next_operation] = *next_operation;
@@ -200,7 +203,7 @@ namespace irods::experimental::auth
     /// \retval false If server version is greater than or equal to 4.3.0
     ///
     /// \since 4.3.0
-    auto use_legacy_authentication(const RcComm& _comm) -> bool
+    inline auto use_legacy_authentication(const RcComm& _comm) -> bool
     {
         static constexpr auto minimum_version_for_auth_plugin_framework = irods::version{4, 3, 0};
 
@@ -228,7 +231,7 @@ namespace irods::experimental::auth
     /// \return JSON response from the server which may be built upon for the next request
     ///
     /// \since 4.3.0
-    auto request(rcComm_t& _comm, const json& _msg)
+    inline auto request(rcComm_t& _comm, const json& _msg)
     {
         constexpr int authentication_api_number = 110000;
 
@@ -249,6 +252,8 @@ namespace irods::experimental::auth
         if (ec < 0) {
             THROW(ec, "failed to perform request");
         }
+
+        at_scope_exit free_memory{[resp] { freeBBuf(resp); }};
 
         return json::parse(static_cast<char*>(resp->buf),
                            static_cast<char*>(resp->buf) + resp->len);
@@ -276,3 +281,4 @@ namespace irods::experimental::auth
 } // namespace irods::experimental::auth
 
 #endif // IRODS_AUTHENTICATION_PLUGIN_FRAMEWORK_HPP
+
--- a/lib/core/include/irods/rcMisc.h
+++ b/lib/core/include/irods/rcMisc.h
@@ -132,10 +132,10 @@ int addStrArray(strArray_t* strArray, char* value);
 
 int resizeStrArray(strArray_t* strArray, int newSize);
 
-int queDataObjInfo(dataObjInfo_t** dataObjInfoHead,
-                   dataObjInfo_t* dataObjInfo,
-                   int singleInfoFlag,
-                   int topFlag);
+int queueDataObjInfo(dataObjInfo_t** dataObjInfoHead,
+                     dataObjInfo_t* dataObjInfo,
+                     int singleInfoFlag,
+                     int topFlag);
 
 int dequeDataObjInfo(dataObjInfo_t** dataObjInfoHead, dataObjInfo_t* dataObjInfo); // JMC - backport 4590
 
--- a/lib/core/include/irods/rcMisc.h
+++ b/lib/core/include/irods/rcMisc.h
@@ -31,6 +31,24 @@ void clearModDataObjMetaInp(void*);
 
 void clearRegReplicaInp(void* voidInp);
 
+void clearGeneralAdminInput(void* _p);
+
+void clearOpenedDataObjInp(void* _p);
+
+void clearSpecificQueryInp(void* _p);
+
+void clearTicketAdminInp(void* _p);
+
+void clearExecMyRuleInp(void* _p);
+
+void clearRuleExecModifyInput(void* _p);
+
+void clearExecCmd(void* _p);
+
+void clearExecRuleExpressionInput(void* _p);
+
+void clearRuleExecSubmitInput(void* _p);
+
 int isPath(char* path);
 
 rodsLong_t getFileSize(char* path);
--- a/lib/core/include/irods/rcMisc.h
+++ b/lib/core/include/irods/rcMisc.h
@@ -49,6 +49,25 @@ void clearExecRuleExpressionInput(void* _p);
 
 void clearRuleExecSubmitInput(void* _p);
 
+void clearAuthRequestOut(void* _p);
+
+void clearAuthCheckOut(void* _p);
+
+void clearPamAuthRequestOut(void* _p);
+
+void clearRodsObjStat(void* _p);
+
+void clearExecCmdOut(void* _p);
+
+void clearDataObjInfo(void* _p);
+
+void clearBytesBuffer(void* _p);
+
+// clang-format off
+__attribute__((deprecated("SimpleQuery is deprecated. Use GenQuery or SpecificQuery instead.")))
+void clearSimpleQueryOut(void* _p);
+// clang-format on
+
 int isPath(char* path);
 
 rodsLong_t getFileSize(char* path);
--- a/lib/core/include/irods/rcMisc.h
+++ b/lib/core/include/irods/rcMisc.h
@@ -132,10 +132,7 @@ int addStrArray(strArray_t* strArray, char* value);
 
 int resizeStrArray(strArray_t* strArray, int newSize);
 
-int queueDataObjInfo(dataObjInfo_t** dataObjInfoHead,
-                     dataObjInfo_t* dataObjInfo,
-                     int singleInfoFlag,
-                     int topFlag);
+int queueDataObjInfo(dataObjInfo_t** dataObjInfoHead, dataObjInfo_t* dataObjInfo, int singleInfoFlag, int topFlag);
 
 int dequeDataObjInfo(dataObjInfo_t** dataObjInfoHead, dataObjInfo_t* dataObjInfo); // JMC - backport 4590
 
--- a/lib/core/include/irods/rcMisc.h
+++ b/lib/core/include/irods/rcMisc.h
@@ -118,6 +118,8 @@ int copyKeyVal(const keyValPair_t* srcCondInput, keyValPair_t* destCondInput);
 
 int replDataObjInp(dataObjInp_t* srcDataObjInp, dataObjInp_t* destDataObjInp);
 
+int replDataObjInfo(struct DataObjInfo* _dst, struct DataObjInfo* _src);
+
 int replSpecColl(specColl_t* inSpecColl, specColl_t** outSpecColl);
 
 int addKeyVal(keyValPair_t* condInput, const char* keyWord, const char* value);
--- a/lib/core/include/irods/rodsGenQueryNames.h
+++ b/lib/core/include/irods/rodsGenQueryNames.h
@@ -10,11 +10,13 @@
 #include "irods/rodsGenQuery.h"
 
 // NOLINTNEXTLINE(modernize-use-using)
-typedef struct GenQueryColumnName {
+typedef struct GenQueryColumnName
+{
     int columnId;
     char *columnName;
 } columnName_t;
 
+// clang-format off
 // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays, modernize-avoid-c-arrays)
 static const columnName_t columnNames[] = {
     { COL_ZONE_ID,          "ZONE_ID", },
@@ -386,8 +388,8 @@ static const columnName_t columnNames[] = {
     { COL_TICKET_OWNER_NAME,               "TICKET_OWNER_NAME", },
     { COL_TICKET_OWNER_ZONE,               "TICKET_OWNER_ZONE", },
 };
+// clang-format on
 
 static const int NumOfColumnNames = sizeof(columnNames) / sizeof(columnName_t);
 
-#endif	// IRODS_GEN_QUERY_NAMES_H
-
+#endif // IRODS_GEN_QUERY_NAMES_H
--- a/lib/core/include/irods/rodsGenQueryNames.h
+++ b/lib/core/include/irods/rodsGenQueryNames.h
@@ -1,21 +1,22 @@
-/*** Copyright (c), The Regents of the University of California            ***
- *** For more information please refer to files in the COPYRIGHT directory ***/
+#ifndef IRODS_GEN_QUERY_NAMES_H
+#define IRODS_GEN_QUERY_NAMES_H
 
-/* rodsGenQueryNames.h - common header file for the generalized query names.
- * Maps the user specified strings to the corresponding #define values.
- */
-
-#ifndef GEN_QUERY_NAMES_H__
-#define GEN_QUERY_NAMES_H__
+/// \file
+///
+/// \brief Common header file for the generalized query names.
+///
+/// Maps the user specified strings to the corresponding #define values.
 
 #include "irods/rodsGenQuery.h"
 
-typedef struct {
+// NOLINTNEXTLINE(modernize-use-using)
+typedef struct GenQueryColumnName {
     int columnId;
     char *columnName;
 } columnName_t;
 
-columnName_t columnNames[] = {
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays, modernize-avoid-c-arrays)
+static const columnName_t columnNames[] = {
     { COL_ZONE_ID,          "ZONE_ID", },
     { COL_ZONE_NAME,        "ZONE_NAME", },
 
@@ -384,9 +385,9 @@ columnName_t columnNames[] = {
     { COL_TICKET_COLL_NAME,                "TICKET_COLL_NAME", },
     { COL_TICKET_OWNER_NAME,               "TICKET_OWNER_NAME", },
     { COL_TICKET_OWNER_ZONE,               "TICKET_OWNER_ZONE", },
-
 };
 
-int NumOfColumnNames = sizeof( columnNames ) / sizeof( columnName_t );
+static const int NumOfColumnNames = sizeof(columnNames) / sizeof(columnName_t);
 
-#endif	/* GEN_QUERY_NAMES_H__ */
+#endif	// IRODS_GEN_QUERY_NAMES_H
+
--- a/lib/core/src/apiHandler.cpp
+++ b/lib/core/src/apiHandler.cpp
@@ -18,34 +18,34 @@ namespace
 
 namespace irods
 {
-    api_entry::api_entry( apidef_t& _def )
-        : plugin_base( "api_instance", "api_context" )
-        , apiNumber( _def.apiNumber )
-        , apiVersion( _def.apiVersion )
-        , clientUserAuth( _def.clientUserAuth )
-        , proxyUserAuth( _def.proxyUserAuth )
-        , inPackInstruct( _def.inPackInstruct )
-        , inBsFlag( _def.inBsFlag )
-        , outPackInstruct( _def.outPackInstruct )
-        , outBsFlag( _def.outBsFlag )
+    api_entry::api_entry(apidef_t& _def)
+        : plugin_base("api_instance", "api_context")
+        , apiNumber(_def.apiNumber)
+        , apiVersion(_def.apiVersion)
+        , clientUserAuth(_def.clientUserAuth)
+        , proxyUserAuth(_def.proxyUserAuth)
+        , inPackInstruct(_def.inPackInstruct)
+        , inBsFlag(_def.inBsFlag)
+        , outPackInstruct(_def.outPackInstruct)
+        , outBsFlag(_def.outBsFlag)
         , call_wrapper(_def.call_wrapper)
         , operation_name(_def.operation_name)
-        , clearInStruct( _def.clearInStruct )
-        , clearOutStruct( _def.clearOutStruct )
+        , clearInStruct(_def.clearInStruct)
+        , clearOutStruct(_def.clearOutStruct)
     {
         operations_[ _def.operation_name ] = _def.svrHandler;
     } // constructor
 
-    api_entry::api_entry( const api_entry& _rhs )
-        : plugin_base( _rhs )
-        , apiNumber( _rhs.apiNumber )
-        , apiVersion( _rhs.apiVersion )
-        , clientUserAuth( _rhs.clientUserAuth )
-        , proxyUserAuth( _rhs.proxyUserAuth )
-        , inPackInstruct( _rhs.inPackInstruct )
-        , inBsFlag( _rhs.inBsFlag )
-        , outPackInstruct( _rhs.outPackInstruct )
-        , outBsFlag( _rhs.outBsFlag )
+    api_entry::api_entry(const api_entry& _rhs)
+        : plugin_base(_rhs)
+        , apiNumber(_rhs.apiNumber)
+        , apiVersion(_rhs.apiVersion)
+        , clientUserAuth(_rhs.clientUserAuth)
+        , proxyUserAuth(_rhs.proxyUserAuth)
+        , inPackInstruct(_rhs.inPackInstruct)
+        , inBsFlag(_rhs.inBsFlag)
+        , outPackInstruct(_rhs.outPackInstruct)
+        , outBsFlag(_rhs.outBsFlag)
         , call_wrapper(_rhs.call_wrapper)
         , in_pack_key(_rhs.in_pack_key)
         , out_pack_key(_rhs.out_pack_key)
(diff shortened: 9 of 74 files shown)


CMakeLists.txt
```diff
@@ -5,6 +5,7 @@ list(INSERT CMAKE_MODULE_PATH 0 "${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules")
 include(RequireOutOfSourceBuild)
 
 option(IRODS_DISABLE_COMPILER_OPTIMIZATIONS "Disables compiler optimizations by setting -O0." OFF)
+option(IRODS_ENABLE_ADDRESS_SANITIZER "Enables detection of memory leaks and other features provided by Address Sanitizer." ON)
```
korydraughn, line 8: Set to OFF or NO

CMakeLists.txt
```diff
@@ -37,10 +38,30 @@ string(REPLACE ";" ", " IRODS_DEVELOP_DEPENDENCIES_STRING "${IRODS_DEVELOP_DEPEN
...
+    -fno-omit-frame-pointer
+    -fno-optimize-sibling-calls
+    -O1
+  )
+  add_link_options(
+    -fsanitize=address
+    -fno-omit-frame-pointer
+    -fno-optimize-sibling-calls
+    -O1
+  )
```
korydraughn, line 55: Consider if additional options are needed.
Consider whether -O0 would be better.

lib/core/src/msParam.cpp
```diff
@@ -43,113 +43,148 @@ addIntParamToArray( msParamArray_t *msParamArray, char *label, int inpInt ) {
...
+        auto* msp = out->msParam[i] = static_cast<MsParam*>(std::malloc(sizeof(MsParam)));
+        std::memset(msp, 0, sizeof(MsParam));
+        if (const auto ec = replMsParam(in->msParam[i], msp); ec < 0) {
+            rodsLogError(LOG_ERROR, ec, "Error when calling replMsParam in %s", __func__);
+            return ec;
         }
     }
+
     return 0;
 }
```
korydraughn, line 189: Need others to review this section carefully.

Thanks.

plugins/rule_engines/irods_rule_language/src/nre.reLib1.cpp
```diff
@@ -294,8 +302,9 @@ execMyRuleWithSaveFlag( char * ruleDef, msParamArray_t *inMsParamArray, const ch
         reDebug( EXEC_MY_RULE_BEGIN, -1, &param, NULL, NULL, rei );
     }
 
-
     char *outParamNames[MAX_PARAMS_LEN];
+    // FIXME extractVarNames can produce a buffer overflow!
+    // It doesn't consider the size of outParamNames!
     int n = extractVarNames( outParamNames, outParamsDesc );
```
korydraughn, line 308: Capture in an issue.

server/api/src/rsExecMyRule.cpp
```diff
@@ -1,152 +1,143 @@
...
-            irods::log( PASS( ret ) );
-            return ret.code();
+        if (!ret.ok()) {
+            log_api::error(ret.result());
+            return ret.code(); // NOLINT(bugprone-narrowing-conversions, cppcoreguidelines-narrowing-conversions)
         }
 
         std::string ret_string = "Available rule engine plugin instances:\n";
-        for ( auto& name : instance_names ) {
+        for (auto& name : instance_names) {
```
korydraughn, line 35: const auto&

server/api/src/rsExecMyRule.cpp
```diff
@@ -1,152 +1,143 @@
...
-                           out_param_desc);
-    if(inst_name.empty()) {
-        freeRErrorContent(&rei.rsComm->rError);
+    // Need to have a non zero inpParamArray for execMyRule to work.
+    if (!_exec_inp->inpParamArray) { // NOLINT(readability-implicit-bool-conversion)
+        // NOLINTNEXTLINE(cppcoreguidelines-owning-memory, cppcoreguidelines-no-malloc)
+        _exec_inp->inpParamArray = static_cast<MsParamArray*>(std::malloc(sizeof(MsParamArray)));
+        std::memset(_exec_inp->inpParamArray, 0, sizeof(MsParamArray));
     }
+    rei.msParamArray = _exec_inp->inpParamArray;
```
korydraughn, line 83: This is just some questionable code.

This feels like the caller should be required to set `_exec_inp->inpParamArray` instead of the server doing this.

server/api/src/rsExecMyRule.cpp
```diff
@@ -1,152 +1,143 @@
...
+
+    // TODO Copy the msparam array
+#if 0
+    *_out_param_arr = rei.msParamArray;
+#else
+    // NOLINTNEXTLINE(cppcoreguidelines-owning-memory, cppcoreguidelines-no-malloc)
+    *_out_param_arr = static_cast<MsParamArray*>(std::malloc(sizeof(MsParamArray)));
+    std::memset(*_out_param_arr, 0, sizeof(MsParamArray));
+    replMsParamArray(rei.msParamArray, *_out_param_arr);
+#endif
```
korydraughn, line 116: `*_out_param_arr` points to the same `MsParamArray` as `_exec_inp->inpParamArray`. This is so weird, hence the else-block.

The else block gives the pointers different copies of the data. The input parameter is free'd, but the output pointer leaks memory.

Can't do anything about this until #6672.

server/core/src/objDesc.cpp
```diff
@@ -28,10 +29,84 @@
...
+
+    // TODO Should these be copied?
+    _dst.pluginData = _src.pluginData;
+    _dst.replica_token = _src.replica_token;
+
+    // NOLINTBEGIN(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+    // TODO Should these be copied?
+    std::memcpy(_dst.chksum, _src.chksum, sizeof(l1desc::chksum));
+    std::memcpy(_dst.in_pdmo, _src.in_pdmo, sizeof(l1desc::in_pdmo));
+    // NOLINTEND(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
```
korydraughn, line 107: Should think about these.

server/core/src/rodsAgent.cpp
```diff
@@ -194,7 +194,7 @@ irodsAgentSignalExit( int ) {
     // Because the agent factory does not call any of the exec-family functions,
     // the POSIX standard recommends using _exit() instead of exit() to keep the
     // child from corrupting the parent's memory.
-    _exit(1);
+    exit(1);
```
korydraughn, line 197: Perhaps this should change based on whether ASan is enabled. Or maybe this form of `exit` is correct.

server/core/src/rsApiHandler.cpp
```diff
@@ -233,63 +234,58 @@ int rsApiHandler(rsComm_t*   rsComm,
...
-    }
+
+    return retVal;
 }
 
-int
-sendAndProcApiReply( rsComm_t * rsComm, int apiInx, int status,
-                     void * myOutStruct, bytesBuf_t * myOutBsBBuf ) {
-    int retval;
+int sendAndProcApiReply(rsComm_t* rsComm, int apiInx, int status, void*& myOutStruct, bytesBuf_t* myOutBsBBuf)
```
korydraughn, line 258: Notice `void*&`.

This allows other function calls to see when `myOutStruct` is changed (set to nullptr).

server/core/src/rsApiHandler.cpp
```diff
@@ -233,63 +234,58 @@ int rsApiHandler(rsComm_t*   rsComm,
...
+        rsComm->portalOpr = nullptr;
     }
 
     return retval;
 }
 
-int
-sendApiReply( rsComm_t * rsComm, int apiInx, int retVal,
-              void * myOutStruct, bytesBuf_t * myOutBsBBuf ) {
+int sendApiReply(rsComm_t* rsComm, int apiInx, int retVal, void*& myOutStruct, bytesBuf_t* myOutBsBBuf)
```
korydraughn, line 282: Notice `void*&`. Same reason as previous comment.

server/main_server/src/rodsServer.cpp
```diff
@@ -1025,11 +1025,11 @@ int main(int argc, char** argv)
 
                 try {
                     // Collapse non-zero error codes into one.
-                    _exit(runIrodsAgentFactory(local_addr) == 0 ? 0 : 1);
+                    exit(runIrodsAgentFactory(local_addr) == 0 ? 0 : 1);
                 }
                 catch (...) {
                     // If an exception is thrown for any reason, terminate the agent factory.
-                    _exit(1);
+                    exit(1);
```
korydraughn, line 1032: Consider whether this is correct.

Running this in the test suite should help answer this question since the introduction of `_exit` was driven by docker.

korydraughn (commented):
Got some things to consider.

korydraughn:
I tried running this PR (with `exit()` and asan disabled) in the testing environment and unfortunately, the use of `_exit()` is still necessary for proper operation. Run the steps from #6429 in a container created via the testing environment to reproduce this issue.

Perhaps this behavior is due to the testing environment not launching containers with an init process. @alanking Does the testing environment launch with an init process? Based on output from `ps`, it doesn't appear to do that.

Running this same code on the host OS does not present any issues. So if modifications to the testing environment don't fix this, then our only option is to use `exit()` when the macro set by asan is defined.

Ultimately, this really sucks. It means we can't rely on just asan for memory issues. Using valgrind along with asan should cover all memory issues.

CMakeLists.txt
korydraughn: done

korydraughn:
After tweaking the testing environment to use an init process, the issue regarding agent factory respawn and `exit()` was not resolved.

I'll do a bit more testing around `exit()` before taking this PR out of draft.

If nothing changes, then we'll adjust iRODS' use of `exit` based on whether asan is used.

server/main_server/src/rodsServer.cpp
```diff
@@ -1173,33 +1177,34 @@ int main(int argc, char** argv)
             // When the server is starting up, agent_spawning_pid will be zero. Therefore,
             // we skip checking for the agent factory on startup.
             if (agent_spawning_pid > 0 && waitpid(agent_spawning_pid, nullptr, WNOHANG) != -1) {
+                log_server::info("Agent factory [{}] still exists.", agent_spawning_pid);
```
korydraughn, line 1180: info -> trace

server/core/src/rodsAgent.cpp
```diff
@@ -39,144 +39,173 @@
...
-    char ack_buffer[256]{};
-    snprintf( ack_buffer, 256, "OK" );
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-member-init, cppcoreguidelines-avoid-magic-numbers)
+    std::array<char, 256> ack_buffer; // NOLINT(readability-magic-numbers)
+    std::snprintf(ack_buffer.data(), ack_buffer.size(), "OK"); // NOLINT(cppcoreguidelines-pro-type-vararg)
 
     while (!data_complete) {
-        memset( in_buf, 0, 1024 );
-        num_bytes = recv( conn_tmp_socket, &in_buf, 1024, 0 );
+        std::fill(std::begin(in_buf), std::end(in_buf), 0);
```
korydraughn, line 104: Add missing `<algorithm>` header.

server/core/src/rodsAgent.cpp
```diff
@@ -192,437 +221,499 @@ void cleanup()
...
+    int agent_status{};
+
+    while ((agent_pid = waitpid(-1, &agent_status, WNOHANG)) > 0) {
+        rmProcLog(agent_pid);
     }
 
+#if __has_feature(address_sanitizer) || defined(__SANITIZE_ADDRESS__)
+    // Calling this function is likely not async-signal-safe, but it is okay because
+    // if the code has been compiled with Address Sanitizer enabled. For that reason,
+    // we can assume that the binary is not running in a production environment.
```
korydraughn, line 249: Fix these ASan comments.

Don't forget to touch other locations as well.

plugins/rule_engines/irods_rule_language/src/nre.reLib1.cpp
korydraughn: See #6743

server/api/src/rsExecMyRule.cpp
korydraughn: Done

server/api/src/rsExecMyRule.cpp
korydraughn: With #6672 being addressed by this PR, I've confirmed that the else-block is appropriate and results in no memory leaks.

server/core/src/rsApiHandler.cpp
korydraughn: This is a good thing.

server/core/src/rsApiHandler.cpp
korydraughn: This is a good thing.

lib/core/src/apiHandler.cpp
```diff
@@ -122,8 +120,9 @@ namespace irods
 
                 // Skip the API plugin if it was loaded before.
                 if (_api_tbl.is_plugin_loaded(path.c_str())) {
-                    const auto* msg = "init_api_table :: API plugin [%s] has already been loaded. Skipping ...";
-                    rodsLog(LOG_DEBUG, msg, path.stem().c_str());
+                    // TODO Do these calls to the logger require guards (i.e. ProcessType != CLIENT_PT)?
+                    log_agent::debug(
+                        "init_api_table :: API plugin [{}] has already been loaded. Skipping ...", path.stem());
```
korydraughn, line 125: TODO

lib/core/src/apiHandler.cpp
```diff
@@ -154,12 +153,9 @@ namespace irods
...
                 if (ret.ok() && entry) {
-                    rodsLog(
-                        LOG_DEBUG,
-                        "init_api_table :: adding %d - [%s] - [%s]",
-                        entry->apiNumber,
-                        entry->operation_name.c_str(),
-                        name.c_str() );
+                    // TODO Do these calls to the logger require guards (i.e. ProcessType != CLIENT_PT)?
+                    log_agent::debug(
+                        "init_api_table :: adding {} - [{}] - [{}].", entry->apiNumber, entry->operation_name, name);
```
korydraughn, line 158: TODO

plugins/rule_engines/irods_rule_language/src/nre.reLib1.cpp
```diff
@@ -294,8 +302,9 @@ execMyRuleWithSaveFlag( char * ruleDef, msParamArray_t *inMsParamArray, const ch
         reDebug( EXEC_MY_RULE_BEGIN, -1, &param, NULL, NULL, rei );
     }
 
-
     char *outParamNames[MAX_PARAMS_LEN];
+    // FIXME extractVarNames can produce a buffer overflow!
+    // It doesn't consider the size of outParamNames!
```
korydraughn, line 307: Remove

server/core/src/rodsAgent.cpp
```diff
@@ -192,437 +224,529 @@ void cleanup()
...
 
-    irods::server_properties::instance().capture();
-    log::agent_factory::set_level(log::get_level_from_config(irods::KW_CFG_LOG_LEVEL_CATEGORY_AGENT_FACTORY));
+        if (WIFEXITED(agent_status)) { // NOLINT(hicpp-signed-bitwise)
+            const int exit_status = WEXITSTATUS(agent_status); // NOLINT(hicpp-signed-bitwise)
+            if (exit_status != 0) {
+                log_agent_factory::error("Agent process [{}] exited with status [{}].", agent_pid, exit_status);
+            }
+            else {
+                log_agent_factory::info("Agent process [{}] exited with status [{}].", agent_pid, exit_status); // TODO restore to debug
```
korydraughn, line 276: TODO

server/core/src/rodsAgent.cpp
```diff
@@ -192,437 +224,529 @@ void cleanup()
...
+                log_agent_factory::error("close(conn_socket) failed with errno = [{}]: {}", errno, strerror(errno));
+            }
 
-                return SYS_FORK_ERROR;
+            if (close(listen_socket) < 0) {
+                log_agent_factory::error("close(listen_socket) failed with errno = [{}]: {}", errno, strerror(errno));
             }
+
+            return SYS_FORK_ERROR;
+        }
```
korydraughn, line 543: TODO

server/core/src/rodsAgent.cpp
```diff
@@ -192,437 +224,529 @@ void cleanup()
...
+
+    log_agent::trace("Agent forked. Initializing ...");
+
+    close(listen_socket);
+
+    // This seems to have fixed the issue with agents not being reaped by the agent factory.
+    // However, if we don't restore the disposition of SIGABRT, every agent will produce a stacktrace on exit.
+    // The current question is, should agents have custom signal handlers?
+    std::signal(SIGABRT, SIG_DFL);
+    //std::signal(SIGABRT, [](int _signum) { log_agent::error("Agent caught SIGABRT [{}]!", _signum); _exit(_signum); }); // Agents are definitely hitting this!
```
korydraughn, line 558: TODO

server/core/src/rodsAgent.cpp
```diff
@@ -192,437 +224,529 @@ void cleanup()
...
-    rodsLog( log_level, "Agent [%d] exiting with status = %d", getpid(), status );
-    return status;
+    (0 == status) ? log_agent::debug("Agent [{}] exiting with status = {}", getpid(), status)
+                  : log_agent::error("Agent [{}] exiting with status = {}", getpid(), status);
+
+#if __has_feature(address_sanitizer) || defined(__SANITIZE_ADDRESS__)
+    __lsan_do_leak_check();
+#endif
+
+    _exit((0 == status) ? 0 : 1);
```
korydraughn, line 746: Add a comment explaining why `_exit` is used here.

server/main_server/src/rodsServer.cpp
```diff
@@ -1178,28 +1260,31 @@ int main(int argc, char** argv)
 
             log_server::info("Forking agent factory ...");
 
-            const auto pid = fork();
+            // FIXME If the agent factory terminates, we need to make sure that all mutexes are
+            // unlocked while forking the new agent factory. Ultimately, this means we must introduce
+            // some signaling mechanism to instruct threads to wait until the new agent factory appears.
+            agent_spawning_pid = fork();
```
korydraughn, line 1266: While true, this is fixed by the use of `_exit` in rodsAgent.cpp.

server/main_server/src/rodsServer.cpp
```diff
@@ -1529,7 +1559,14 @@ serverExit()
     // Wake and terminate agent spawning process
     kill(agent_spawning_pid, SIGTERM);
 
-    exit(1);
+#if __has_feature(address_sanitizer) || defined(__SANITIZE_ADDRESS__)
+    // Calling this function is likely not async-signal-safe, but it is okay because
+    // if the code has been compiled with Address Sanitizer enabled. For that reason,
+    // we can assume that the binary is not running in a production environment.
+    __lsan_do_leak_check();
```
korydraughn, line 1566: Fix this comment.

server/main_server/src/rodsServer.cpp
```diff
@@ -2250,7 +2287,7 @@ void task_spawn_manager()
         }
 
         boost::unique_lock<boost::mutex> spwn_req_lock(SpawnReqCondMutex);
-        SpawnReqCond.wait(spwn_req_lock);
+        SpawnReqCond.wait(spwn_req_lock); // TODO This form of wait is not recommended due to spurious wake ups, etc.
```
korydraughn, line 2290: While not the best, iRODS has used this form of `wait` for quite some time. Therefore, we'll leave this as is.

server/main_server/src/rodsServer.cpp
korydraughn: Removed.

server/main_server/src/rodsServer.cpp
korydraughn: Leave a comment about `runIrodsAgentFactory` never returning on shutdown.

server/main_server/src/rodsServer.cpp
korydraughn: Create an issue for this form of `wait`.

server/core/src/rodsAgent.cpp
```diff
@@ -192,437 +224,529 @@ void cleanup()
...
+        log::set_server_type("agent");
+
+        // Reload server_config.json into memory for the newly forked agent process.
+        irods::environment_properties::instance().capture();
+
+        // should we check this return code??
+        status = receiveDataFromServer(conn_tmp_socket);
+        if (status < 0) {
+            log_agent::error("receiveDataFromServer failed [error_code=[{}]].", status);
+            //return err.code(); // TODO Should we return here?
```
alanking, line 578: Looks like this hasn't changed, but there is a new TODO.

alanking (commented):
Overall, everything looks like it's going in the right direction!

lib/api/include/irods/data_object_finalize.h
```diff
@@ -3,10 +3,10 @@
 
 /// \file
 
-const char* SET_TINE_TO_NOW_KW = "set time to now";
-
 struct RcComm;
 
+static const char* const SET_TINE_TO_NOW_KW = "set time to now";
```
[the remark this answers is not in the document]
trel: N vs M…

SET_TINE_TO_NOW_KW

server/api/src/rsExecMyRule.cpp
korydraughn: Leaving as is. The current code does not report any leaks stemming from this code. And it's not that questionable. It seems it is more about convenience than anything else.

server/core/src/objDesc.cpp
korydraughn: These should be copied because that is what the function is supposed to do and doing anything else feels more incorrect.

lib/core/src/msParam.cpp
korydraughn: Good.

server/core/src/objDesc.cpp
korydraughn: A quick use of grep shows that the following member variables are not used:
- `otherDataObjInfo`
- `replDataObjInfo`
- `pluginData`

I'm going to ignore these variables and create an issue to deprecate them. They aren't used and their meaning is clear.

server/main_server/src/rodsServer.cpp
korydraughn: Done.

CMakeLists.txt
```diff
@@ -80,6 +101,7 @@ include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/InstallDirs.cmake)
 
 set(CMAKE_POSITION_INDEPENDENT_CODE TRUE)
 
+add_compile_options(-O0 -ggdb)
```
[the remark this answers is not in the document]
korydraughn: This is temporary and will be removed.

lib/api/include/irods/data_object_finalize.h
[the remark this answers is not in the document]
korydraughn: Great. I really didn't like seeing the `_TINE_` in there.

For some reason, I didn't ask questions when I came across it. I just tried to resolve the ODR violation reported by ASan.

Will remove.

lib/api/include/irods/fileStat.h
```diff
@@ -4,7 +4,7 @@
 #include "irods/rodsDef.h"
```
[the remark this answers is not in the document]
korydraughn: Got a little tired of touching all the things lol.

Will summon the strength to update those :-)

lib/api/include/irods/getMiscSvrInfo.h
```diff
@@ -12,7 +12,7 @@
 
 typedef struct MiscSvrInfo {
     int serverType;     // RCAT_ENABLED or RCAT_NOT_ENABLED
-    uint serverBootTime;
+    int serverBootTime; // TODO uint (non-standard) to int - ABI break!
```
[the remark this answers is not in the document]
korydraughn: I'm not exactly sure. My guess is that it's an ABI break. However, `uint` is not a standard data type and we don't claim to provide a stable ABI AFAIK.

If that is true, then this is safe. Please think about it and let me know if you feel we shouldn't do this.

Also, take a look at the packing instruction for that member.

lib/core/include/irods/apiHandler.hpp
```diff
@@ -56,7 +57,8 @@ namespace irods
 
         const char *   operation_name;
 
-        std::function<void( void* )> clearInStruct;	// free input struct function
+        std::function<void(void*)> clearInStruct;	// free input struct function
+        std::function<void(void*)> clearOutStruct;	// free output struct function
```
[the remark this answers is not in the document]
korydraughn: I don't understand the question. Can you elaborate?

As for the internal use part of the question, I think these are internal only and are mainly used by the api table and plugins.

lib/core/include/irods/rcMisc.h
```diff
@@ -31,6 +31,39 @@ void clearModDataObjMetaInp(void*);
...
+void clearPamAuthRequestOut(void* _p);
+
+void clearRodsObjStat(void* _p);
+
+void clearExecCmdOut(void* _p);
+
+void clearDataObjInfo(void* _p);
+
+__attribute__((deprecated("SimpleQuery is deprecated. Use GenQuery or SpecificQuery instead.")))
+void clearSimpleQueryOut(void* _p);
```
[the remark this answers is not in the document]
korydraughn: Well, we already have rcMisc.h and that's where the other clear* functions live. I really didn't want to have to think through another file lol (for this PR).

We need to discuss how we want to handle splitting up rcMisc.h. @MartinFlores751 is refactoring it now. That can be part of that effort.

plugins/api/src/data_object_modify_info.cpp
```diff
@@ -143,6 +144,7 @@ auto plugin_factory(const std::string& _instance_name,
                         op,                              // Operation
                         "api_data_object_modify_info",   // Operation name
                         clearModDataObjMetaInp,          // Null clear function
+                        irods::clearOutStruct_noop,
```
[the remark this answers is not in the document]
korydraughn: No, I agree. Many of the things such as this are because I'm afraid of the work existing on my machine.

I'm going to clean up all of these changes as this PR comes to a close.

server/api/src/rsDataObjRename.cpp
```diff
@@ -309,6 +311,11 @@ namespace
         srcDataObjInp = &dataObjRenameInp->srcDataObjInp;
         destDataObjInp = &dataObjRenameInp->destDataObjInp;
 
+        irods::at_scope_exit free_cond_inputs{[srcDataObjInp, destDataObjInp] {
+            clearKeyVal(&srcDataObjInp->condInput);
+            clearKeyVal(&destDataObjInp->condInput);
```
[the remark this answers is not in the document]
korydraughn: Will revisit this. I haven't run any of these changes through the testing environment. If that comes back clean, then I think we're good.

server/core/src/dataObjOpr.cpp
```diff
@@ -379,10 +380,11 @@ getDataObjInfo(
...
         dataObjInfo->writeFlag = writeFlag;
 
-        dataObjInfo->next = 0;
-
-        queDataObjInfo( dataObjInfoHead, dataObjInfo, 1, 0 );
+        dataObjInfo->next = nullptr;
 
+        constexpr int is_single_object = 1;
+        constexpr int prepend_object = 0;
+        queDataObjInfo(dataObjInfoHead, dataObjInfo, is_single_object, prepend_object);
```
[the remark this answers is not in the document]
korydraughn: Will fix that right up.

server/main_server/src/rodsServer.cpp
```diff
@@ -1178,28 +1260,31 @@ int main(int argc, char** argv)
...
 
                 try {
                     // Collapse non-zero error codes into one.
-                    _exit(runIrodsAgentFactory(local_addr) == 0 ? 0 : 1);
+                    std::exit(runIrodsAgentFactory(local_addr) == 0 ? 0 : 1);
                 }
                 catch (...) {
                     // If an exception is thrown for any reason, terminate the agent factory.
-                    _exit(1);
+                    std::exit(1);
```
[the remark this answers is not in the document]
korydraughn: I think I'm going to have to restore this to `_exit` for the case where the agent factory fails to fork an agent.

I'll add a comment explaining why.

lib/core/src/msParam.cpp
```diff
@@ -43,113 +45,148 @@ addIntParamToArray( msParamArray_t *msParamArray, char *label, int inpInt ) {
...
+
+        // Make the new set of pointers point to elements in the msParamArray.
+        // The new set of pointers will own the memory referenced by msParamArray.
+        for (int i = 0; i < len; ++i) {
+            msp_pointers[i] = msParamArray->msParam[i];
         }
-        if ( msParamArray->msParam != NULL ) {
-            free( msParamArray->msParam );
+
+        if (msParamArray->msParam) {
```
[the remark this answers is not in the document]
korydraughn: The number of things being reported was too much for clangd so it just stops and I don't see the rest until the others are resolved.

lib/core/src/msParam.cpp
```diff
@@ -43,113 +45,148 @@ addIntParamToArray( msParamArray_t *msParamArray, char *label, int inpInt ) {
...
-        newLen = msParamArray->len + PTR_ARRAY_MALLOC_LEN;
-        newParam = ( msParam_t ** ) malloc( newLen * sizeof( *newParam ) );
-        memset( newParam, 0, newLen * sizeof( *newParam ) );
-        for ( int i = 0; i < len; i++ ) {
-            newParam[i] = msParamArray->msParam[i];
+    // Resize the MsParam array if necessary.
+    if ((msParamArray->len % PTR_ARRAY_MALLOC_LEN) == 0) {
+        const auto element_count = msParamArray->len + PTR_ARRAY_MALLOC_LEN;
+        const auto byte_count = element_count * sizeof(MsParam*);
+        auto** msp_pointers = static_cast<MsParam**>(std::malloc(byte_count));
```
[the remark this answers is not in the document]
korydraughn: See https://github.com/irods/irods/pull/6673#discussion_r1037752157

lib/core/src/rcMisc.cpp
```diff
@@ -1436,6 +1459,232 @@ clearDataObjCopyInp( void* voidInp ) {
...
+    }
+
+    auto* q = static_cast<DataObjInfo*>(_p);
+
+    clearKeyVal(&q->condInput);
+    free_pointer(q->specColl);
+    freeAllDataObjInfo(q->next);
+}
+
+void clearSimpleQueryOut(void* _p)
```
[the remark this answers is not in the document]
korydraughn: `const` pointers aren't supported.

These functions match the signature needed by `apidef_t` and `api_entry` (the `clearOutStruct` function wrapper).

lib/core/src/msParam.cpp
```diff
@@ -43,113 +45,148 @@ addIntParamToArray( msParamArray_t *msParamArray, char *label, int inpInt ) {
...
+            if (!type || !msp->type) {
+                log_msi::warn("{}: Type argument for label [{}] is a null pointer.", __func__, label);
                 continue;
             }
-            /***  Jan 28 2010 to make it not given an error ***/
-            if ( !strcmp( msParamArray->msParam[i]->type, STR_MS_T ) &&
-                    !strcmp( type, STR_MS_T ) &&
-                    !strcmp( ( char * ) inOutStruct, ( char * ) msParamArray->msParam[i]->inOutStruct ) ) {
+
+            // clang-format off
```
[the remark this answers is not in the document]
korydraughn: It's because of the opening curly brace on L87.

I feel clang-format should never allow this:
```cpp
if (cond_1 &&
    cond_2) {
    stmt_1;
    stmt_2;
}
```
I just find that formatting to be unpleasant and clang-format doesn't provide controls for handling it.

lib/core/src/msParam.cpp
```diff
@@ -493,49 +529,65 @@ rmMsParamByLabel( msParamArray_t *msParamArray, const char *label, int freeStruc
...
+    if (msParam->label) {
+        std::free(msParam->label);
     }
-    if ( msParam->inOutStruct != NULL && ( freeStruct > 0 ||
-                                           ( msParam->type != NULL && strcmp( msParam->type, STR_MS_T ) == 0 ) ) ) {
-        free( msParam->inOutStruct );
+
+    if (msParam->inOutStruct) {
+        if (msParam->type) {
+            if (msParam->type && std::strcmp(msParam->type, STR_MS_T) == 0) {
```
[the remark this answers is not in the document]
korydraughn: Nice catch.

server/core/include/irods/objDesc.hpp
```diff
@@ -28,8 +23,13 @@
 #define REG_CHKSUM      1
 #define VERIFY_CHKSUM   2
 
-/* values in l1desc_t is the desired value. values in dataObjInfo are
- * the values in rcat */
+// Values in l1desc_t is the desired value.
+// Values in dataObjInfo are the values in rcat.
+//
+// If you're adding a non-trivial type to this structure, you must also update
+// init_l1desc()'s implementation.
```
korydraughn, line 30: Update to include `copy_l1desc`.

server/core/src/irods_signal.cpp
```diff
@@ -80,14 +80,20 @@ namespace irods
...
-        rodsLog(LOG_DEBUG, "Stacktraces will be dumped to \"%s\"", irods::STACKTRACE_DIR.c_str());
-        struct sigaction action{};
+        using log_server = irods::experimental::log::server;
+
+        log_server::debug("Setting stacktrace dump signal handler for process [{}].", getpid());
+        log_server::debug("Stacktraces will be dumped to [{}].", irods::STACKTRACE_DIR);
+
+        struct sigaction action;
+        action.sa_flags = 0;
+        sigemptyset(&action.sa_mask);
```
korydraughn, line 92: @SwooshyCueb Did you leave `action.sa_mask` untouched on purpose?

I wasn't sure if there was a reason for not calling `sigemptyset` or `sigfillset` on it.

server/core/src/rodsAgent.cpp
korydraughn: Changed to `std::memset`.

lib/core/src/apiHandler.cpp
korydraughn: Done

lib/core/src/apiHandler.cpp
korydraughn: Done

server/core/src/rodsAgent.cpp
korydraughn: Done

server/core/src/rodsAgent.cpp
korydraughn: Done

lib/api/include/irods/data_object_finalize.h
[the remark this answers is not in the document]
korydraughn: Done

server/core/src/rodsAgent.cpp
korydraughn: Done.

server/core/src/objDesc.cpp
korydraughn: Done.

server/core/src/rodsAgent.cpp
korydraughn: Changed to `_exit` because `std::exit` is not async-signal-safe.

server/main_server/src/rodsServer.cpp
korydraughn: The agent factory is normally shut down via SIGTERM. However, if the agent factory fails to fork an agent, it will return SYS_FORK_ERROR up the call chain.

If the agent factory crashes for whatever reason, respawned by the main server process, and then fails to fork an agent, the agent factory will most likely exit with a SIGABRT due to a failed assertion in boost::mutex or boost::condition_variable.

The only way around this is to continue using `_exit` in the cron manager task for the agent factory.

Basically the change captured here needs to be reverted to its original form.

I'll add this as the comment to the code.

lib/core/src/rcMisc.cpp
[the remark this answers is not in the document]
alanking: Understood

lib/core/src/rcMisc.cpp
```diff
@@ -1436,6 +1459,232 @@ clearDataObjCopyInp( void* voidInp ) {
...
+    free_const_pointer(q->arg1);
+    free_const_pointer(q->arg2);
+    free_const_pointer(q->arg3);
+    free_const_pointer(q->arg4);
+    free_const_pointer(q->arg5);
+    free_const_pointer(q->arg6);
+    free_const_pointer(q->arg7);
+    free_const_pointer(q->arg8);
+    free_const_pointer(q->arg9);
+}
```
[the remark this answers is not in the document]
korydraughn: This is actually ok. The data referenced by the pointer doesn't mean the data is actually const. The const-ness can be casted away as demonstrated in the code above.

See https://gcc.godbolt.org/z/Ecxxb3rPj

In case the link dies, the code and compiler explorer options are below.
```cpp
#include <cstdlib>

void no_leaks_or_undefined_behavior()
{
    const auto* p = static_cast<char*>(std::malloc(100));
    auto* q = const_cast<char*>(p);
    std::free(q);
    p = nullptr;
    q = nullptr;
}

int main()
{
    no_leaks_or_undefined_behavior();
}
```
Compiler options: `-std=c++20 -Wall -Wextra -Wpedantic -Wconversion -O0 -g -fsanitize=address,undefined`

lib/api/include/irods/getMiscSvrInfo.h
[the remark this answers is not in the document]
alanking: haha Oh, I see. Yes, this is probably safe. The packing instruction convinced me.

server/main_server/src/rodsServer.cpp
korydraughn: Done.

server/core/src/rodsAgent.cpp
korydraughn: Done.

server/main_server/src/rodsServer.cpp
korydraughn: Done.

server/main_server/src/rodsServer.cpp
korydraughn: Done.

CMakeLists.txt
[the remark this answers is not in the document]
korydraughn: Done.

lib/api/include/irods/fileStat.h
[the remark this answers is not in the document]
korydraughn: Done.

lib/api/include/irods/getMiscSvrInfo.h
```diff
@@ -12,7 +12,7 @@
```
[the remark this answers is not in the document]
korydraughn: Done.

server/core/src/dataObjOpr.cpp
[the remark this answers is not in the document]
korydraughn: Done.

server/core/src/rsApiHandler.cpp
```diff
@@ -298,51 +303,56 @@ sendApiReply( rsComm_t * rsComm, int apiInx, int retVal,
...
     if ( RsApiTable[apiInx]->outBsFlag == 0 ) {
-        myOutBsBBuf = NULL;
+        myOutBsBBuf = nullptr;
     }
 
     if ( rsComm->rError.len > 0 ) {
-        status = pack_struct( ( char * ) &rsComm->rError, &rErrorBBuf,
-                             "RError_PI", RodsPackTable, 0, rsComm->irodsProt, rsComm->cliVersion.relVersion );
+        status = pack_struct(
+            (char*) &rsComm->rError,
```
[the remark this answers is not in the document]
korydraughn: Done.

server/main_server/src/rodsServer.cpp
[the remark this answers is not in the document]
korydraughn: Done.

lib/core/src/msParam.cpp
```diff
@@ -218,22 +255,21 @@ replBytesBuf( const bytesBuf_t* in) {
...
+{
     if ( label != NULL ) {
         msParam->label = strdup( label );
     }
 
     msParam->type = type ? strdup( type ) : NULL;
-    if ( inOutStruct != NULL && msParam->type != NULL &&
-            strcmp( msParam->type, STR_MS_T ) == 0 ) {
+
+    if (inOutStruct != NULL && msParam->type != NULL && strcmp(msParam->type, STR_MS_T) == 0) {
```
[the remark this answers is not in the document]
korydraughn: Done.

lib/core/src/msParam.cpp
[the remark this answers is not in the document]
korydraughn: Done.

server/core/src/rodsAgent.cpp
korydraughn: Done.

server/core/include/irods/objDesc.hpp
korydraughn: Done.

lib/api/include/irods/getMiscSvrInfo.h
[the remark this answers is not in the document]
korydraughn: Turns out that we define this type alias!
https://github.com/irods/irods/blob/2f602874739f3d0d15551127cddbac181fb88e27/lib/core/include/irods/rodsType.h#L24
Will restore this value and create an issue to deprecate it.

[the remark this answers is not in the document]
korydraughn: Done. See #6756

plugins/api/src/data_object_modify_info.cpp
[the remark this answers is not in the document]
korydraughn: Done

plugins/api/src/get_grid_configuration_value.cpp
```diff
@@ -116,7 +117,8 @@ auto plugin_factory(const std::string& _instance_name,
                         "GridConfigurationOut_PI", 0,         // Out PI / bs flag
                         op,                                   // Operation
                         "api_get_grid_configuration_value",   // Operation name
-                        nullptr,                              // Null clear function
+                        irods::clearInStruct_noop,                              // Null clear function
+                        irods::clearOutStruct_noop,
```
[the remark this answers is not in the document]
korydraughn: Done

server/api/src/rsExecMyRule.cpp
```diff
@@ -1,152 +1,142 @@
...
+    // NOLINTBEGIN(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+    const std::string my_rule_text = _exec_inp->myRule;
+    const std::string out_param_desc = _exec_inp->outParamDesc;
+    // NOLINTEND(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+
+    irods::rule_engine_context_manager<irods::unit, RuleExecInfo*, irods::AUDIT_RULE> re_ctx_mgr(
+        irods::re_plugin_globals->global_re_mgr, &rei);
+    irods::error err = re_ctx_mgr.exec_rule_text(inst_name, my_rule_text, _exec_inp->inpParamArray, out_param_desc);
+
+    // If the client didn't specify a target REP, clear all error information. This is by design.
```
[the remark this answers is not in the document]
korydraughn: Done

plugins/api/src/register_physical_path.cpp
```diff
@@ -1346,16 +1350,19 @@ auto plugin_factory(const std::string& _instance_name,
...
+        NO_USER_AUTH,                   // Client auth
+        NO_USER_AUTH,                   // Proxy auth
+        "DataObjInp_PI", 0,             // In PI / bs flag
+        "BinBytesBuf_PI", 0,            // Out PI / bs flag
+        op,                             // Operation
+        "api_register_physical_path",   // Operation name
+        clearDataObjInp,                // Clear input function // TODO Why was this set to NOP?
+        [](void* _p) { clearBBuf(static_cast<BytesBuf*>(_p)); }, // Clear output function
+        (funcPtr) CALL_REGISTER_PHYSICAL_PATH
+    };
```
korydraughn, line 1365: @alanking Notice the TODO.

Is it correct for this API plugin to not clear the incoming `DataObjInp`?

server/core/src/irods_signal.cpp
SwooshyCueb: I have no memory of this, so probably not

server/core/src/irods_signal.cpp
```diff
@@ -80,14 +80,20 @@ namespace irods
 
     const std::string STACKTRACE_DIR(irods::get_irods_stacktrace_directory().string());
 
-    void set_unrecoverable_signal_handlers() // TODO Consider calling this for grandpa also.
+    void set_unrecoverable_signal_handlers() // TODO Consider calling this for grandpa also. <<<
```
korydraughn, line 83: Handle TODO

server/main_server/src/rodsServer.cpp
```diff
@@ -1517,7 +1571,7 @@ void
 #if defined(linux_platform) || defined(aix_platform) || defined(solaris_platform) || defined(osx_platform)
 serverExit( int sig )
 #else
-serverExit()
+serverExit() // TODO What OS would hit this code path? Windows?
```
korydraughn, line 1574: Handle TODO

plugins/api/src/register_physical_path.cpp
korydraughn: This is fine as the clear function is only triggered when the rc-call is invoked.

lib/core/include/irods/rcMisc.h
korydraughn, line 65: All of these functions need to call `std::memset` as the final step.

plugins/api/src/atomic_apply_acl_operations.cpp
```diff
@@ -569,17 +569,22 @@ auto plugin_factory(const std::string& _instance_name,
     irods::client_api_allowlist::add(ATOMIC_APPLY_ACL_OPERATIONS_APN);
 #endif // RODS_SERVER
 
+    const auto clear_bbuf_adapter = [](void* _p) { clearBBuf(static_cast<BytesBuf*>(_p)); };
```
korydraughn, line 572: I feel there's a library here ...

Given a callable, return a new callable matching the supplied signature.

Is this possible to implement?

server/core/src/irods_signal.cpp
korydraughn: Marking as resolved since we can infer from the original code that no signals should be blocked due to how `action` was originally initialized.

server/core/src/rsApiHandler.cpp
```diff
@@ -233,63 +234,67 @@ int rsApiHandler(rsComm_t*   rsComm,
...
+    clearBBuf(myOutBsBBuf);
+#if 0
+    if (myOutStruct) {
+        std::free(myOutStruct);
+        myOutStruct = nullptr;
     }
-    freeRErrorContent( &rsComm->rError );
-
-    /* check for portal operation */
+#endif
```
korydraughn, line 277: Remove this block.

server/main_server/src/rodsServer.cpp
```diff
@@ -673,7 +678,11 @@ namespace
             execv(args[0], args.data());
 
             // If execv() fails, the POSIX standard recommends using _exit() instead of
-            // exit() to keep the child from corrupting the parent's memory.
+            // exit() to avoid flushing stdio buffers and handlers registered by the parent.
+            //
+            // In the case of C++, this is necessary to avoid triggering destructors as there may
+            // be invariants that could be violated due to the use of fork() (e.g. assertion made
+            // by the boost::mutex implementation regarding pthread_mutex_destroy()).
```
korydraughn, line 685: Improve this wording.

server/api/src/rsDataObjRename.cpp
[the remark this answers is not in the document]
korydraughn: Yeah, this is incorrect.

If invoked by a client, the API framework will free this memory.
If invoked internally, it is the caller's responsibility to free this memory.

Therefore, this code must be removed.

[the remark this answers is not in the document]
korydraughn: Removed.

server/main_server/src/rodsServer.cpp
korydraughn: Removed.

server/core/src/rsApiHandler.cpp
korydraughn: Done

server/main_server/src/rodsServer.cpp
korydraughn: Done

lib/core/include/irods/rcMisc.h
korydraughn: Done.

server/core/src/irods_signal.cpp
korydraughn: Captured in #6760.

lib/core/include/irods/authentication_plugin_framework.hpp
```diff
@@ -276,3 +281,4 @@ namespace irods::experimental::auth
 } // namespace irods::experimental::auth
 
 #endif // IRODS_AUTHENTICATION_PLUGIN_FRAMEWORK_HPP
+
```
korydraughn: remove

lib/core/include/irods/rodsGenQueryNames.h
```diff
@@ -384,9 +385,9 @@ columnName_t columnNames[] = {
...
     { COL_TICKET_OWNER_ZONE,               "TICKET_OWNER_ZONE", },
-
 };
 
-int NumOfColumnNames = sizeof( columnNames ) / sizeof( columnName_t );
+static const int NumOfColumnNames = sizeof(columnNames) / sizeof(columnName_t);
+
+#endif	// IRODS_GEN_QUERY_NAMES_H
 
-#endif	/* GEN_QUERY_NAMES_H__ */
```
korydraughn, line 392: Remove trailing empty line.

lib/core/include/irods/authentication_plugin_framework.hpp
korydraughn: done

lib/core/include/irods/rodsGenQueryNames.h
korydraughn: done

plugins/api/src/atomic_apply_acl_operations.cpp
korydraughn: Done

server/core/src/rodsAgent.cpp
```diff
@@ -192,437 +224,529 @@ void cleanup()
...
+    log::set_error_object(&rsComm.rError);
+    irods::at_scope_exit release_error_stack{[] { log::set_error_object(nullptr); }};
 
-    memset( &rsComm, 0, sizeof( rsComm ) );
-    rsComm.thread_ctx = ( thread_context* )malloc( sizeof( thread_context ) );
+    //std::memset(&rsComm, 0, sizeof(RsComm));
+    rsComm.thread_ctx = static_cast<thread_context*>(std::malloc(sizeof(thread_context)));
 
-    status = initRsCommWithStartupPack( &rsComm, nullptr );
+    status = initRsCommWithStartupPack(&rsComm, nullptr);
```
korydraughn, line 589: Apparently, this function treats a socket fd of 0 as invalid which is incorrect. 0 is in fact a valid socket fd.

The socket fd is retrieved via an environment variable and parsed using `std::atoi()`. `std::atoi()` can return 0 if it cannot parse the input.

Because of this, the server never stores the IP address in the `RsComm` causing  `iticket` tests to fail.

korydraughn: The values stored in the environment need to be logged as debug.

We should consider logging inputs and outputs of functions for debug purposes. Perhaps as compile-time only log messages.

CMakeLists.txt
korydraughn: Marking as resolved as these options work just fine.

We can always tweak later.

korydraughn:
I've run these code changes against the core tests and unit tests. All tests passed.

Due to the tweak to the NREP, I now need to verify the memory leak situation is better than before. That ultimately means the main server process and the delay server process must not leak memory.

I'm also going to try running this PR through the testing environment with ASan enabled to see what happens.

server/core/src/json_serialization.cpp
```diff
@@ -15,6 +15,8 @@ namespace
     {
         auto* s = parseMspForStr(&_p);
 
+        // TODO Returning nullptr isn't incorrect. However, parseMspForStr() will return
+        // a nullptr if the string value is "null". This feels incorrect.
         if (!s) {
             THROW(SYS_INVALID_INPUT_PARAM, "Failed to convert microservice argument to string.");
         }
```
korydraughn, line 22: Handle TODO

`nullptr` in this case can mean success or failure.

server/core/src/finalize_utilities.cpp
```diff
@@ -223,25 +223,18 @@ namespace irods
...
+            auto* rest = dest.dataObjInfo->next;
+            freeAllDataObjInfo(rest);
+            dest.dataObjInfo->next = nullptr;
         }
 
-        // TODO: need duplication logic
-        dest.remoteZoneHost = nullptr;
-        dest.otherDataObjInfo = nullptr;
-        dest.replDataObjInfo = nullptr;
-
```
korydraughn, line 244: @alanking I changed the code to do a deep copy of the source l1desc object. Alternatively, I could copy construct the source l1desc object and keep the rest of the code as it was originally. See #6792 for the alternative solution.

Using the copy constructor seems better in the sense that it is likely faster than using `copy_l1desc()`.

Any thoughts? Primarily in regard to correctness.

_I'm going to rename `copy_l1desc()` to `deep_copy_l1desc()` or something that includes `deep` in the name._

server/core/src/dataObjOpr.cpp
```diff
@@ -347,7 +347,8 @@ getDataObjInfo(
         snprintf( dataObjInfo->objPath, MAX_NAME_LEN, "%s/%s", tmpCollName, tmpDataName );
         rstrcpy( dataObjInfo->rescHier, tmpHierString, MAX_NAME_LEN );
         irods::error ret = resc_mgr.hier_to_leaf_id( tmpHierString, dataObjInfo->rescId );
-        if( !ret.ok() ) {
+        if (!ret.ok()) {
+            std::free(dataObjInfo);
```
alanking, line 351: Should this be `freeDataObjInfo`?

server/core/src/finalize_utilities.cpp
alanking: I think that the copy constructor in #6792 is fine/bettter, as you said. It will have the same effect as the existing code without the UB risk.

The deep copy would be super-extra-safe since the source could go away and leave us with dangling pointers in the destination L1 descriptor. That apparently hasn't been a problem for anything so far since we weren't making a deep copy before.

alanking (commented):
Haven't looked at everything yet, just responding to a couple things

server/core/src/dataObjOpr.cpp
alanking: Or maybe even `freeAllDataObjInfo`

korydraughn: Seems reasonable. I'm not sure if there was a reason I used `std::free` here.

Will investigate.

server/core/src/finalize_utilities.cpp
korydraughn: Same here. Both implementations would likely work. And the original code seems to only care about the first `DataObjInfo`.

What makes you think we can dangle? The original implementation appears to allocate and copy the data it cares about from the source (i.e. L230 and L236 in the original impl).

alanking: Well, maybe the concern is unfounded as #6754 deprecates the unused pointers.

server/core/src/json_serialization.cpp
korydraughn: Created separate issue for this. See #6810.

This will be handled by a different PR.

korydraughn: Removed comment now that it's captured in an issue.

server/core/src/finalize_utilities.cpp
korydraughn: Restored some of the implementation to its original form due to the rebase and this discussion.

server/core/src/dataObjOpr.cpp
korydraughn: Done.

server/api/src/rsExecMyRule.cpp
```diff
@@ -1,152 +1,144 @@
 #include "irods/rsExecMyRule.hpp"
 
+#include "irods/objInfo.h"
+#include "irods/msParam.h"
+#include "irods/rcConnect.h"
 #include "irods/execMyRule.h"
 #include "irods/irods_logger.hpp"
 #include "irods/irods_re_plugin.hpp"
 #include "irods/miscServerFunct.hpp"
 #include "irods/rcMisc.h"
```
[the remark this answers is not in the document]
korydraughn: Done.

server/main_server/src/rodsServer.cpp
```diff
@@ -1371,8 +1446,14 @@ int main(int argc, char** argv)
...
             if (irods::server_state::server_state::stopped == state) {
                 procChildren(&ConnectedAgentHead);
-                // Wake up the agent factory process so it can clean up and exit
+
+                // Shut down the agent factory.
                 kill(agent_spawning_pid, SIGTERM);
+                int agent_factory_status = 0;
+                waitpid(agent_spawning_pid, &agent_factory_status, 0);
+                log_server::info(
+                    "Agent factory has completed shutdown [exit_code={}].", WEXITSTATUS(agent_factory_status));
```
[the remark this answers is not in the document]
korydraughn: This code is fine. The call to `waitpid` will cause the server to wait until the target process terminates.

server/main_server/src/rodsServer.cpp
```diff
@@ -1016,13 +1025,89 @@ namespace
...
+    } // launch_purge_lock_file_thread
+
+    void join_purge_lock_file_thread(std::string_view _server_role)
+    {
+        if (irods::KW_CFG_SERVICE_ROLE_PROVIDER == _server_role) {
+            try {
+                PurgeLockFileThread->join();
+            }
+            catch (const boost::thread_resource_error&) {
+                log_server::error("boost encountered a thread_resource_error during join.");
```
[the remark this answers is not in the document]
korydraughn: Done.

server/main_server/src/rodsServer.cpp
```diff
@@ -1016,13 +1025,89 @@ namespace
...
+
+    void join_spawn_manager_thread()
+    {
+        SpawnReqCond.notify_all();
+
+        try {
+            SpawnManagerThread->join();
+        }
+        catch (const boost::thread_resource_error&) {
+            log_server::error("boost encountered a thread_resource_error during spawn manager thread.");
```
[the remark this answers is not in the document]
korydraughn: Done.

server/main_server/src/rodsServer.cpp
```diff
@@ -1016,13 +1025,89 @@ namespace
...
+    void join_read_worker_threads()
+    {
+        for (int i = 0; i < NUM_READ_WORKER_THR; ++i) {
+            ReadReqCond.notify_all();
+
+            try {
+                ReadWorkerThread[i]->join();
+            }
+            catch (const boost::thread_resource_error&) {
+                log_server::error("boost encountered a thread_resource_error during read worker thread.");
```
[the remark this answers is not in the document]
korydraughn: Done.

server/core/src/rodsAgent.cpp
korydraughn: There's nothing actionable here.

Marking as resolved.

korydraughn:
All unit tests passed.

Running core tests.

korydraughn:
All core tests passed.

All that's left is to add issue numbers and do one last pass over the changes.

server/api/src/rsDataObjPut.cpp
```diff
@@ -576,6 +577,8 @@ namespace
 
         try {
             dataObjInfo_t* dataObjInfoHead{};
+            irods::at_scope_exit free_data_object_info{[&dataObjInfoHead] { freeAllDataObjInfo(dataObjInfoHead); }};
```
alanking, line 580: It looks like we might be able to remove `dataObjInfoHead` since it is not being used. `file_object_factory` frees the `DataObjInfo` internally if `nullptr` is passed as the 4th parameter (`nullptr` is also the default value). Up to you - this change has the same effect.

alanking (commented):
Very nice. Just noticed one thing that can be ignored

server/api/src/rsDataObjPut.cpp
korydraughn: Will take a look.

korydraughn: I think I added this because a memory leak was detected when putting a directory which involved overwriting existing data objects.

I'm going to leave it as everything is happy at this time. We can continue to iterate on this after it is merged.

korydraughn:
I've assigned issue numbers to each commit.

I believe the issue numbers line up correctly.

If there are no further issues, let me know and I'll add the pounds.

alanking (dismissed):
I think it's ready... perhaps one more approval and we can do it?

trel (approved):
looks good to me

korydraughn:
Added #'s

alanking (approved):

Issue #4094, opened 2018-09-02
Implement several rule tests for PREP


alanking:
Remove skipIf lines for the following tests and provide a python rule engine plugin implementation:

- [ ] test_msiServerMonPerf_default_3736
- [ ] test_msiCheckAccess_3309
- [ ] test_acPreProcForExecCmd__3867
- [ ] test_unix_filesystem_free_space_on_root__3928
- [ ] test_ichksum_fails_with_dyn_pep__issue_3485
- [ ] test_icommands_segfault_when_ssl_cert_hostname_not_matching__issue_3609
- [ ] test_open_write_close_for_repl__3909
- [ ] test_delay_block_with_output_param__3906

korydraughn:
PR #8193 added tests which need to be considered too.

Please search the test suites for `TODO(#4094)` to find them.

PR #6676, opened 2022-11-01
[#6622] Add test for closeAllL1Desc (main)

alanking:
Adds a test for the change which address #6622

Companion PR to #6623

---

I tried to add a test for PREP but found that `msiExecCmd` does not work in PREP right now. Instead of calling across to a native rule, I think that the current test is sufficient as it is.


scripts/irods/test/rule_texts_for_tests.py
```diff
@@ -606,6 +606,25 @@
...
+'''
+rule_texts['irods_rule_engine_plugin-irods_rule_language']['Test_Rulebase']['test_msiExecCmd_closeAllL1Desc__issue_6623_pep'] ='''
+pep_resource_close_post(*inst, *ctx, *out) {{
+    msiGetSystemTime(*time_str, '');
+    msiAddKeyVal(*key_val_pair,"{attr}","*time_str");
+    msiAssociateKeyValuePairsToObj(*key_val_pair,"{logical_path}","-d");
+    # Sleep to ensure that the actual close doesn't occur in the same second
+    msiSleep("1", "0");
+}}
+'''
```
korydraughn, line 627: Consider moving these into their respsective test.

Having part of the test logic in a separate file introduces more hurdles to understanding the test.

If you see benefits to continuing to use this file, I'd be interested in hearing them.

alanking: I created #6677 to remove this file. Might as well just go ahead and do it. ;p

alanking:
I have confirmed that all the tests pass in isolation but I'm running it through one more time to be sure.

alanking:
For some reason, `test_stacktrace` falls over when run along with the rest of the suite, but it passes when run individually. I think this may be a fluke, so I'm going to say the tests are passing.

trel:
sounds like a cleanup situation in the test prior to test_stacktrace...

korydraughn:
Well, the test is relying on the log file so I agree it is likely a fluke/timing issue. We need a better way to confirm these things.

alanking:
One test suite failed for PREP because it is only implemented for the native rule language. I have skipped it and referenced #4094. All the rest of the tests passed.

alanking:
Finally made Codacy happy, so I'm gonna run tests just one more time. :)

alanking:
All tests passed

korydraughn (approved):
Let's add the #'s.

PR #6685, opened 2022-11-03
[#6638] SimpleQuery: Deprecate types, constants, and functions (main)

alanking:
Addresses #6638 (the rest of it)

Follow-up to #6657 

Just left a comment for the API number, but the rest has now received a deprecation attribute. Had to wrap some bits to ignore the deprecated declarations warnings that are now generated.

### DIFF
--- a/server/api/src/rsSimpleQuery.cpp
+++ b/server/api/src/rsSimpleQuery.cpp
@@ -6,6 +6,9 @@
 #include "irods/miscServerFunct.hpp"
 #include "irods/rodsConnect.h"
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+
 namespace
 {
     auto _rsSimpleQuery(rsComm_t* rsComm, simpleQueryInp_t* simpleQueryInp, simpleQueryOut_t** simpleQueryOut) -> int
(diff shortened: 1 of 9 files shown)


korydraughn (approved):
Ship it!

korydraughn:
Let's wait for another approval before the pounds though.

server/api/src/rsSimpleQuery.cpp
```diff
@@ -87,3 +90,5 @@ rsSimpleQuery( rsComm_t *rsComm, simpleQueryInp_t *simpleQueryInp,
     }
     return status;
 }
+
+#pragma clang diagnostic pop // ignored "-Wdeprecated-declarations"
```
trel, line 94: why does this one get a comment?  the others didn't...

alanking: I put a comment on this one because the `clang diagnostic push` occurs at the top of the file (there's a lot of deprecated stuff being used in here :p). The other `push`es and `pop`s were close enough to one another that they could be visually inspected very easily to understand what is happening.

That said, I'm happy to add comments to the others, or remove this comment

trel: all good, don't feel strongly - just noticed the difference.

alanking: In that case, I am going to leave things how they are. I'm hoping the next time we look at any of this is when SimpleQuery gets removed.

trel (approved):

PR #6694, opened 2022-11-04
Adjusted log message structure (main)

korydraughn:


korydraughn:
Well, that's not good. Got some compilation issues to work.

alanking (dismissed):

lib/core/include/irods/irods_logger.hpp
```diff
@@ -687,6 +689,13 @@ namespace irods::experimental::log
                 object[tag::server::pid] = getpid();
                 object[tag::server::timestamp] = utc_timestamp();
 
+                if (const auto* name = irods::find_server_property<std::string>(irods::KW_CFG_ZONE_NAME); name) {
```
korydraughn, line 692: I'm a little concerned with this.

The server properties impl now uses a mutex to make sure access is synchronized across threads. However, server properties also calls into the logger on error. This means we may have a deadlock situation here.

How do we get the zone name without producing this situation?
How do we handle `irods-grid --reload`?
How do we handle admins changing the zone name?

If zone name changes require a restart of the server, then the solution is to attach the zone name to the logger on program start. _(I feel this may be the answer as processes like the delay server may not see zone name changes.)_

Thoughts?

alanking: A fine point...

I've confirmed with a bench test that the local iRODS Zone can be renamed without restarting the server, surprisingly. However, as you've stated, the delay server and the delay server migration logic fail to authenticate with the server until it is restarted. So, technically, in order for 4.3.x to be fully functional, a restart is required to rename the local zone.

That said, I agree with your assessment that getting/storing the zone name at program start is the right answer.

korydraughn: Excellent. I'll adjust it.

Thanks for verifying.

lib/core/include/irods/irods_server_properties.hpp
```diff
@@ -66,6 +66,18 @@ namespace irods
...
+        {
+            auto lock = acquire_read_lock();
+
+            if (const auto iter = config_props_.find(_key); iter != std::end(config_props_)) {
+                return iter->get_ptr<const T*>();
+            }
+
+            return nullptr;
+        } // find
+
```
korydraughn, line 80: remove

lib/core/include/irods/irods_server_properties.hpp
```diff
@@ -196,6 +208,12 @@ namespace irods
         return irods::server_properties::instance().contains(_prop);
     }
 
+    template <typename T>
+    const T* find_server_property(const std::string& _prop)
+    {
+        return irods::server_properties::instance().find<T>(_prop);
+    }
+
```
korydraughn, line 216: remove

lib/core/include/irods/irods_server_properties.hpp
korydraughn: done

lib/core/include/irods/irods_server_properties.hpp
korydraughn: done

lib/core/include/irods/irods_logger.hpp
korydraughn: Updated.

alanking (approved):
Add #'s when ready

korydraughn:
Added #'s

korydraughn:
Looks like the ubuntu containers for github are experiencing some issues (outside of our control). Let's ignore them for now.

Issue #6619, opened 2022-10-04
Refactor resource administration API to throw exceptions instead of return error codes


korydraughn:
- [x] main
- [x] 4-3-stable
---
## Refactor / Enhancement
The current design requires developers to check each operation resulting in very noisy code. Moving to exceptions allows developers to write simpler/clearer code.

This change may be backported to 4-2-stable if it is trivial.

PR #6703, opened 2022-11-06
[#6619] Refactored resource administration lib to prefer exceptions (main)

korydraughn:
All unit tests passed.

~Running core tests.~

We can skip running the core tests because nothing in the server outside of unit tests use the resource administration library.

### DIFF
--- a/lib/administration/resource/src/resource_administration.cpp
+++ b/lib/administration/resource/src/resource_administration.cpp
@@ -3,12 +3,12 @@
 #undef rxGeneralAdmin
 
 #ifdef IRODS_RESOURCE_ADMINISTRATION_ENABLE_SERVER_SIDE_API
-    #include "irods/rsGeneralAdmin.hpp"
-    #define rxGeneralAdmin  rsGeneralAdmin
-    #define IRODS_QUERY_ENABLE_SERVER_SIDE_API
+#  include "irods/rsGeneralAdmin.hpp"
+#  define rxGeneralAdmin rsGeneralAdmin
+#  define IRODS_QUERY_ENABLE_SERVER_SIDE_API
 #else
-    #include "irods/generalAdmin.h"
-    #define rxGeneralAdmin  rcGeneralAdmin
+#  include "irods/generalAdmin.h"
+#  define rxGeneralAdmin rcGeneralAdmin
 #endif // IRODS_RESOURCE_ADMINISTRATION_ENABLE_SERVER_SIDE_API
 
 #include "irods/rodsErrorTable.h"
@@ -18,7 +18,7 @@
 
 namespace irods::experimental::administration::NAMESPACE_IMPL
 {
-    auto add_resource(RxComm& _comm, const resource_registration_info& _info) -> std::error_code
+    auto add_resource(RxComm& _comm, const resource_registration_info& _info) -> void
     {
         std::string location;
 
@@ -36,13 +36,13 @@ namespace irods::experimental::administration::NAMESPACE_IMPL
         input.arg6 = ""; // Zone name (cannot be null).
 
         if (const auto ec = rxGeneralAdmin(&_comm, &input); ec < 0) {
-            return std::error_code{ec, std::generic_category()};
+            constexpr char* msg = "Could not create resource [name=[{}], type=[{}], location=[{}], context=[{}]]";
+            // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+            THROW(ec, fmt::format(msg, _info.resource_name, _info.resource_type, location, _info.context_string));
         }
-
-        return {};
     }
 
-    auto remove_resource(RxComm& _comm, const resource_name_type _name) -> std::error_code
+    auto remove_resource(RxComm& _comm, const resource_name_type _name) -> void
     {
         generalAdminInp_t input{};
         input.arg0 = "rm";
@@ -51,16 +51,16 @@ namespace irods::experimental::administration::NAMESPACE_IMPL
         input.arg3 = ""; // Dryrun flag (cannot be null).
 
         if (const auto ec = rxGeneralAdmin(&_comm, &input); ec < 0) {
-            return std::error_code{ec, std::generic_category()};
+            // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+            THROW(ec, fmt::format("Could not remove resource [{}]", _name));
         }
-
-        return {};
     }
 
-    auto add_child_resource(RxComm& _comm,
-                            const resource_name_type _parent,
-                            const resource_name_type _child,
-                            const std::string_view _context_string) -> std::error_code
+    auto add_child_resource(
+        RxComm& _comm,
+        const resource_name_type _parent,
+        const resource_name_type _child,
+        const std::string_view _context_string) -> void
     {
         generalAdminInp_t input{};
         input.arg0 = "add";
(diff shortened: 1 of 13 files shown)


lib/administration/resource/src/resource_administration.cpp
```diff
@@ -87,51 +84,48 @@ namespace irods::experimental::administration::NAMESPACE_IMPL
...
+    auto resource_exists(RxComm& _comm, const resource_name_type _name) -> bool
     {
         try {
-            const auto gql = fmt::format("select RESC_ID where RESC_NAME = '{}'", _name);
-            return {std::error_code{}, query_builder{}.build(_comm, gql).size() > 0};
+            return query_builder{}.build(_comm, fmt::format("select RESC_ID where RESC_NAME = '{}'", _name)).size() > 0;
+        }
+        catch (const irods::exception& e) {
+            // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+            THROW(e.code(), e.what());
```
alanking, line 99: Do we need to do this? Looks like it is just re-constructing the exception and re-throwing without adding anything.

Also seen on 155 and 196.

unit_tests/src/test_resource_administration.cpp
```diff
@@ -16,199 +16,191 @@ TEST_CASE("resource administration")
...
 
-    auto conn_pool = irods::make_connection_pool();
-    auto conn = conn_pool->get_connection();
+    irods::experimental::client_connection conn;
 
     irods::at_scope_exit remove_resources{[&conn, &repl_info, &ufs_info] {
-        adm::client::remove_resource(conn, repl_info.resource_name);
-        adm::client::remove_resource(conn, ufs_info.resource_name);
+        try {
+            CHECK_NOTHROW(adm::client::remove_resource(conn, repl_info.resource_name));
```
alanking, line 36: I was under the impression that CHECK_NOTHROW and friends handled the exceptions or lack thereof which would render the try-catch unnecessary. Is that not the case?

I guess it doesn't matter since this is cleanup code, so it doesn't hurt to be safe.

Same thing for 41

unit_tests/src/unit_test_utils.hpp
```diff
@@ -78,7 +78,9 @@ namespace unit_test_utils
         ufs_info.host_name = host_name;
         ufs_info.vault_path = vault_path;
 
-        return adm::client::add_resource(_comm, ufs_info).value() == 0;
+        adm::client::add_resource(_comm, ufs_info);
+
+        return true; // TODO
```
alanking: Is this TODO referring to the fact that this function now throws? Or was that always the case?

unit_tests/src/test_zone_report.cpp
```diff
@@ -60,6 +59,6 @@ TEST_CASE("json zone report")
                 id_str = resource["id"].get<std::string>();
             }
         }
-        REQUIRE(info->id() == id_str);
+        REQUIRE(info.id() == id_str);
```
alanking: I am confused by this change since all the other `adm::client::resource_info`s are treated as `std::optional`, but this one is not. Can you explain?

lib/administration/resource/src/resource_administration.cpp
```diff
@@ -87,51 +84,48 @@ namespace irods::experimental::administration::NAMESPACE_IMPL
...
+                "RESC_LOC, RESC_VAULT_PATH, RESC_STATUS, "
+                "RESC_CONTEXT, RESC_COMMENT, RESC_INFO, "
+                "RESC_FREE_SPACE, RESC_FREE_SPACE_TIME, "
+                "RESC_PARENT, RESC_CREATE_TIME, RESC_MODIFY_TIME "
+                "where RESC_NAME = '{}'",
+                _name);
 
             for (auto&& row : query_builder{}.build(_comm, gql)) {
+                class resource_info info;
+                info.name_ = std::string{_name};
```
alanking, line 126: Why is this inside the loop now? I guess there will ever only be one result for this query, so this should be fine. Just curious

alanking (commented):
Looks good overall.

lib/administration/resource/src/resource_administration.cpp
korydraughn: No. I'll adjust it.

unit_tests/src/test_resource_administration.cpp
korydraughn: I wasn't sure and I didn't want to run the risk of the other call not executing.

I'll verify and tweak.

unit_tests/src/unit_test_utils.hpp
korydraughn: I think that was removed in the second commit.

lib/administration/resource/src/resource_administration.cpp
korydraughn: The previous impl returned a resource_info object even when the resource didn't exist. This makes the logic clearer because you only get the resource_info object if the query found a match.

This also makes it clearer to see when std::nullopt is returned.

unit_tests/src/test_zone_report.cpp
korydraughn: I don't see that in the final result.

I'm guessing you're looking at each commit individually?

unit_tests/src/unit_test_utils.hpp
alanking: Yep... I forgot to hit the next button. :)

unit_tests/src/test_zone_report.cpp
alanking: Indeed

lib/administration/resource/src/resource_administration.cpp
alanking: Makes sense.

lib/administration/resource/src/resource_administration.cpp
korydraughn: done

unit_tests/src/test_resource_administration.cpp
korydraughn: This comment from the Catch2 docs indicates that the `CHECK_NOTHROW` macro works as we suspect it.
```
The CHECK family are equivalent but execution continues in the same test case even if the assertion fails. This is useful if you have a series of essentially orthogonal assertions and it is useful to see all the results rather than stopping at the first failure.
```
I think we're covered.

korydraughn:
Ignoring single clang-format report (it produces a poorly formatted result).

unit_tests/src/test_resource_administration.cpp
korydraughn: I've verified through manual testing that the try-catch blocks around the `CHECK_NOTHROW` macros aren't necessary.

The code has been updated.

lib/administration/resource/src/resource_administration.cpp
```diff
@@ -191,10 +183,6 @@ namespace irods::experimental::administration::NAMESPACE_IMPL
 
             return std::nullopt;
         }
-        catch (const irods::exception& e) {
-            // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
-            THROW(e.code(), e.what());
-        }
```
alanking: whups

alanking (approved):
Add #s when ready

lib/administration/resource/src/resource_administration.cpp
korydraughn: Should I ignore this comment lol?

alanking: Please do. I felt like this change maybe belongs in the previous commit, but probably not worth the effort to address it.

korydraughn: Nice catch. I'll move that change into the previous one.

korydraughn: Fixed it.

korydraughn:
Added #'s

Issue #6693, opened 2022-11-04
Re-enable `test_auth`


alanking:
- [x] main
- [x] 4-3-stable
- [x] testing environment (configure pam)
---
`test_auth` is currently being skipped due to yet-unexplained failures. Please investigate.

korydraughn:
Reopening as test is failing with SSL_HANDSHAKE_ERROR when running iinit.

alanking:
The issue is that the test was actually passing on Ubuntu 18.04, but not 20.04.

The OpenSSL version installed for Ubuntu 18.04 in the testing environment is:
```
$ openssl version
OpenSSL 1.1.1  11 Sep 2018
```
The OpenSSL version installed for Ubuntu 20.04 in the testing environment is:
```
$ openssl version
OpenSSL 1.1.1f  31 Mar 2020
```
Apparently, something changed between these two versions such that a numbits value of 1024 for the `genrsa` command used to generate the certificate chain file is no longer sufficient for performing the SSL handshake between the client and the server.

If the value of numbits used to generate the certificate chain file is bumped to 2048, the test passes on both platforms.

I have confirmed that the SSL issues do not seem to manifest on Centos 7, which uses an even older version of OpenSSL. However, the test seems to fail for a different reason -- the PAM password is not being accepted. Investigating.

alanking:
Awaiting merge of irods/irods_testing_environment#132 before closing.

trel:
https://github.com/irods/irods_testing_environment/pull/132 is now merged.

alanking:
Yes, confirmed that this works on all of our platforms. Closing.

PR #6704, opened 2022-11-07
[#6693] Fix test_auth (main)

alanking:
Addresses #6693

Unskipped test, confirmed that it is passing

Also, typo

### DIFF
--- a/scripts/irods/test/test_auth.py
+++ b/scripts/irods/test/test_auth.py
@@ -24,7 +24,6 @@ from . import session
 
 
 # Requires existence of OS account 'irodsauthuser' with password ';=iamnotasecret'
-@unittest.skip('This test passes when run manually, but fails with automation.')
 class Test_Auth(resource_suite.ResourceBase, unittest.TestCase):
     plugin_name = IrodsConfig().default_rule_engine_plugin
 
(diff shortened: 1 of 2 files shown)


scripts/irods/test/test_auth.py
```diff
@@ -99,85 +99,93 @@ def test_authentication_PAM_without_negotiation(self):
...
+                    server_config_update = {
+                        'authentication' : {
+                            'pam_password' : {
+                                'password_length': 20,
+                                'no_extend': False,
+                                'password_min_time': 121,
+                                'password_max_time': 1209600,
+                                }
                             }
                         }
```
korydraughn, line 149: Curly braces are off.

scripts/irods/test/test_auth.py
```diff
@@ -99,85 +99,93 @@ def test_authentication_PAM_without_negotiation(self):
...
 
-                    # the test
-                    print(f'running iinit for PAM user [{self.auth_session.username}] [{self.auth_session.password}]')
-                    self.auth_session.assert_icommand('iinit', 'STDOUT_SINGLELINE',
-                                                      input=f'{self.auth_session.password}\n')
-                    self.auth_session.assert_icommand("icd")
-                    self.auth_session.assert_icommand("ils -L", 'STDOUT_SINGLELINE', "home")
+                        print('attempting to run an icommand')
+                        lib.execute_command(['ils'])
+                        print('Success!!')
```
korydraughn, line 173: Can you explain why this is necessary?

Just curious as to why `lib.execute_command` is used. It the `ils` fails, would that be caught or mean anything regarding the test?

korydraughn (commented):
Everything looks fine. Just one question so far.

scripts/irods/test/test_auth.py
alanking: Will fix

scripts/irods/test/test_auth.py
alanking: Oh, this is leftover debugging code. I will remove this.

For posterity, `lib.execute_command` is used instead of a session because the client environment for the service account is being modified in this test. I wanted to confirm that it wasn't getting screwed up before attempting to run something as some other account. It's not really any more valid than using one of the sessions, but removed a layer in my debugging journey.

korydraughn: Understood.

alanking:
Any other thoughts on this?

korydraughn (approved):
Squash and if the issue numbers are correct, add the pounds.

alanking:
Done. Mergin

PR #6700, opened 2022-11-05
[#6639] Refactored client_api_allowlist interface (main)

korydraughn:
Running core tests.

Issue #6639, opened 2022-10-13
Refactor client_api_allowlist interface to match style of replica_access_table, etc


korydraughn:
- [x] main
- [x] 4-3-stable
---
## Refactor
While the client api allowlist's functionality will remain the same, we should change how it is held to match other modern iRODS singleton interfaces.

This is more about interface consistency than anything.

PR #6707, opened 2022-11-07
[#6639] Refactored client_api_allowlist interface (4-3-stable)

korydraughn:
Cherry-picked #6700

### DIFF
--- a/server/core/src/client_api_allowlist.cpp
+++ b/server/core/src/client_api_allowlist.cpp
@@ -8,193 +8,191 @@
 #include "irods/rodsErrorTable.h"
 
 #include <exception>
-
-namespace logger = irods::experimental::log;
+#include <vector>
 
 namespace
 {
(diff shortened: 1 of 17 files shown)


server/core/src/client_api_allowlist.cpp
```diff
@@ -8,193 +8,191 @@
...
+            EXEC_RULE_EXPRESSION_AN,
+
+            SERVER_REPORT_AN,
+            ZONE_REPORT_AN,
+            CLIENT_HINTS_AN
+        };
+        // clang-format on
+    } // init
+
+    auto enforce(const RsComm& _comm) noexcept -> bool
```
korydraughn, line 187: This function is marked as `noexcept`, but can throw.

Either wrap the body in a try-catch block or remove the `noexcept` qualifier.

alanking (approved):

PR #6713, opened 2022-11-08
Modify `rstrcpy` to improve error logs (main)

MartinFlores751:
Update `rstrcpy` logging to produce shorter, informative messages.

### DIFF
--- a/lib/core/src/stringOpr.cpp
+++ b/lib/core/src/stringOpr.cpp
@@ -3,11 +3,15 @@
 #include "irods/rodsErrorTable.h"
 #include "irods/irods_stacktrace.hpp"
 #include "irods/rodsLog.h"
+#include "irods/irods_logger.hpp"
+#include "irods/rcGlobalExtern.h"
 
 #include <boost/regex.hpp>
 
 #include <cstring>
 #include <string>
+
+using log_server = irods::experimental::log::server;
 
 char *rmemmove( void *dest, const void *src, int strLen, int maxLen ) {
 
@@ -45,28 +49,35 @@ char *rmemcpy( void *dest, const void *src, int strLen, int maxLen ) {
     }
 }
 
-char *rstrcpy( char *dest, const char *src, int maxLen ) {
+char* rstrcpy(char* dest, const char* src, int maxLen)
+{
     // snprintf with logging on fail
+    if (dest == nullptr || src == nullptr) {
+        return nullptr;
+    }
 
-    if ( dest == NULL || src == NULL ) {
-        return NULL;
+    int status{std::snprintf(dest, maxLen, "%s", src)};
+    int err{errno};
+    if (status >= 0) {
+        if (status < maxLen) {
+            return dest;
+        }
+        if (::ProcessType != CLIENT_PT) {
+            log_server::error(
+                "rstrcpy not enough space in destination buffer, source length:[{}], destination length:[{}]; "
+                "resulting state of destination [{}], truncated from source [{}]",
+                status,
+                maxLen,
+                dest,
+                src);
+        }
+        return nullptr;
     }
-    int status = snprintf( dest, maxLen, "%s", src );
 
-    if ( status >= 0 && status < maxLen ) {
-        return dest;
+    if (::ProcessType != CLIENT_PT) {
+        log_server::error("rstrcpy encountered an error: [{}]", std::strerror(err));
     }
-    else if ( status >= 0 ) {
-        rodsLog( LOG_ERROR,
-                 "rstrcpy not enough space in dest, slen:%d, maxLen:%d, stacktrace:\n%s",
-                 status, maxLen, irods::stacktrace().dump().c_str());
-        rodsLog( LOG_DEBUG, "rstrcpy arguments dest [%s] src [%s]", dest, src );
-        return NULL;
-    }
-    else {
-        rodsLog( LOG_ERROR, "rstrcpy encountered an encoding error." );
-        return NULL;
-    }
+    return nullptr;
 }
 
 int is_empty_string(const char* _s, size_t _max_size)


lib/core/src/stringOpr.cpp
```diff
@@ -58,7 +58,7 @@ char *rstrcpy( char *dest, const char *src, int maxLen ) {
     }
     else if ( status >= 0 ) {
         rodsLog( LOG_ERROR, "rstrcpy not enough space in dest, slen:%d, maxLen:%d", status, maxLen );
-        rodsLog( LOG_DEBUG, "rstrcpy arguments dest [%s] src [%s]", dest, src );
+        rodsLog( LOG_ERROR, "rstrcpy arguments dest [%s] src [%s]", dest, src );
```
trel: should we just... show src?  we really don't care about dest if it didn't work, right?

MartinFlores751: Might be nice to have if some existing code uses `dest` without checking for errors in copying.

trel: fair enough.   any reason not to combine the two lines into a single error message?

MartinFlores751: Nope! Can do though.

lib/core/src/stringOpr.cpp
```diff
@@ -45,28 +45,27 @@ char *rmemcpy( void *dest, const void *src, int strLen, int maxLen ) {
...
-        return NULL;
+    int status{std::snprintf(dest, maxLen, "%s", src)};
+
+    if (status >= 0) {
+        if (status < maxLen) {
+            return dest;
+        }
+        rodsLog(LOG_ERROR,
+            "rstrcpy not enough space in dest, slen:%d, maxLen:%d; arguments dest [%s] src [%s]",
+            status, maxLen, dest, src);
```
korydraughn, line 63: If possible, let's switch to the new logger.
```cpp
using log_server = irods::experimental::log::server;
...
log_server::error(<your-args-here>);
```

Also, consider replacing `dest`, `src`, `sLen` and `maxLen`. Those words are terrible for these types of log messages. We can be as descriptive as we want here.

Think on it.

lib/core/src/stringOpr.cpp
```diff
@@ -45,28 +45,27 @@ char *rmemcpy( void *dest, const void *src, int strLen, int maxLen ) {
...
+        return nullptr;
     }
+
+    rodsLog(LOG_ERROR, "rstrcpy encountered an encoding error.");
```
korydraughn, line 67: Same here as far as the new logger.

korydraughn:
Have you verified the behavior of this function for clients?
Have you run this against the unit tests and core tests?

lib/core/src/stringOpr.cpp
```diff
@@ -45,28 +48,28 @@ char *rmemcpy( void *dest, const void *src, int strLen, int maxLen ) {
...
-        rodsLog( LOG_ERROR, "rstrcpy encountered an encoding error." );
-        return NULL;
+    int status{std::snprintf(dest, maxLen, "%s", src)};
+
+    if (status >= 0) {
+        if (status < maxLen) {
+            return dest;
+        }
+        log_server::error(
+            "rstrcpy not enough space in destination buffer, source length:{}, destination length:{}; "
```
alanking, line 65: Very minor: Could wrap the first two format parameters with square brackets (i.e. `[{}]`) to be consistent with the next line.

lib/core/src/stringOpr.cpp
```diff
@@ -45,28 +48,28 @@ char *rmemcpy( void *dest, const void *src, int strLen, int maxLen ) {
...
+            "resulting state of destination [{}], truncated from source [{}]",
+            status, maxLen, dest, src);
+        return nullptr;
     }
+
+    log_server::error("rstrcpy encountered an encoding error.");
```
alanking, line 71: Is this always the case? Maybe we could show the `errno` set by the system in this case (https://pubs.opengroup.org/onlinepubs/9699919799/functions/fprintf.html#tag_16_159_04).

I'm also okay with leaving this the way that it is because I don't like this function anyway.

alanking (commented):
Looking at this work again in its entirety, I think having 1 commit with both issue numbers should be fine. Just my two cents.

MartinFlores751:
Core and unit tests passed. An additional manual test for no client-side logging using `iinit` with a large `HOME` environment variable was successful.

korydraughn:
Very nice. Let's squash it and run `git clang-format`.

lib/core/src/stringOpr.cpp
```diff
@@ -45,28 +49,35 @@ char *rmemcpy( void *dest, const void *src, int strLen, int maxLen ) {
...
+            return dest;
+        }
+        if (::ProcessType != CLIENT_PT) {
+            log_server::error(
+                "rstrcpy not enough space in destination buffer, source length:[{}], destination length:[{}]; "
+                "resulting state of destination [{}], truncated from source [{}]",
+                status,
+                maxLen,
+                dest,
+                src);
```
trel, line 72: i think these are out of order

trel: oh, maybe not - just ... poorly named(?)... bah.

korydraughn:
Please mark any remaining comments as resolved and squash everything into a single commit.

Leave off the pound so we can verify the commits look good.

trel:
consider splitting the commits if you haven't already... agreed, it's not obvious looking at the diff.

MartinFlores751:
> consider splitting the commits if you haven't already... agreed, it's not obvious looking at the diff.

It's a bit difficult to do that. Both of the referenced issues are effectively solved [in this line](https://github.com/irods/irods/pull/6713/files#diff-39a355ce3d90a45756f8e2b46812a4f09e073c97fb6c5c6ce38e8639f7aaa116R66-R72). We can split the touch ups from the two issues, though.

trel (approved):
Gotcha - that's fine - just wanted to make sure we thought through it.

awaiting another vote or two...

korydraughn (approved):
Add the pounds and we'll get this merged.

PR #6728, opened 2022-11-11
Client API Allowlist / Delay Server Migration / gethostname (main)

korydraughn:


server/main_server/src/rodsServer.cpp
```diff
@@ -718,13 +718,26 @@ namespace
                 }
             }
 
-            char hostname[HOST_NAME_MAX]{};
+            char hostname[HOST_NAME_MAX + 1]{};
```
trel, line 721: this should be part of the next commit?

korydraughn: Well, this commit becomes broken if moved into the next commit.

This change sparked the next commit.

trel: fair enough, so leave as is... or reshuffle the commits if that helps.   all good.

korydraughn: After looking at the commit again, this line can be moved into the other commit.

Swizzling things around.

korydraughn: Done.

trel:
Nice - yes, looks good to me.

alanking (approved):
Seems good to me. Perhaps some explanatory words in the commit message could be helpful for future viewing without needing to reference the linked issues. Up to you

server/main_server/src/rodsServer.cpp
```diff
@@ -718,13 +718,26 @@ namespace
...
+                log_server::trace("Local server's hostname resolved via host_resolution configuration = [{}]", *hn);
+
+                if (hn->size() >= sizeof(hostname)) {
+                    log_server::error(
+                        "Local server's resolved hostname exceeds [{}] bytes (i.e. HOST_NAME_MAX).", HOST_NAME_MAX);
+                    return;
+                }
+
+                std::strncpy(hostname, hn->c_str(), sizeof(hostname));
+            }
```
korydraughn, line 739: This logic is incorrect.

korydraughn: Adjusted the implementation to resolve the local, leader, and successor host names against the host_resolution server_config.json property.

Verified at that the server will use the preferred hostnames for migration if defined.

korydraughn:
All unit tests and core tests pass.

As stated earlier, I've verified that the changes to the delay server migration logic give admins a way to handle aliases of the host name.

This PR does not add support for /etc/hosts. We need to discuss whether that is important moving forward.

korydraughn:
The migration logic uses `gethostname()` to determine if migration should happen. If the value it returns does not match the value of `$ hostname` and the `host_resolution` property is not configured, then migration will never happen.

As of this PR, the `host_resolution` property is the only way admins can support multiple names.

The question is, should the migration logic query the machine for ALL names?

As for testing, we do not have the facilities in place to test these types of changes.

trel:
Oh, I see... if they're providing the alias via the /etc/hosts... we might not see it.

Hm... that seems like potential additional work/consideration - and not part of this PR for now.   Do we walk/consider the /etc/hosts alias list anywhere else in our codebase?

korydraughn:
That's correct.

No, we don't walk the /etc/hosts file. We normally call `gethostname()` and use the host_resolution config.

I agree that work can be done in a different PR.

trel (approved):
seems good - a vote for # from me...

korydraughn:
Fixed a log message just now too.

Added the #'s.

PR #6734, opened 2022-11-21
[5916,6674] Improve `rstrcpy` logging (4-3-stable)

MartinFlores751:
Update log level to show rstrcpy arguments, stop rstrcpy printing stacktrace

### DIFF
--- a/lib/core/src/stringOpr.cpp
+++ b/lib/core/src/stringOpr.cpp
@@ -3,11 +3,15 @@
 #include "irods/rodsErrorTable.h"
 #include "irods/irods_stacktrace.hpp"
 #include "irods/rodsLog.h"
+#include "irods/irods_logger.hpp"
+#include "irods/rcGlobalExtern.h"
 
 #include <boost/regex.hpp>
 
 #include <cstring>
 #include <string>
+
+using log_server = irods::experimental::log::server;
 
 char *rmemmove( void *dest, const void *src, int strLen, int maxLen ) {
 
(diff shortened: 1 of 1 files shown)


korydraughn (approved):
Looks good. Does it compile and do what you want?

MartinFlores751:
Will check and get back to you on that.

On a side note, I just noticed that 4-2-stable doesn't have the modern logging from what I see. Might need to use `rodsLog(...)` for that one?

korydraughn:
> On a side note, I just noticed that 4-2-stable doesn't have the modern logging from what I see. Might need to use rodsLog(...) for that one?

Correct.

MartinFlores751:
All but two tests passed for me, being *test_icd* and *test_iadmin*. Looking at the logs, I believe they may have been introduced by the PR irods/irods_client_icommands#331.

The failing `icd` test expects the following string: `No such directory (collection):` but the commit changed the string to `No such collection:`

On *test_iadmin* I'm seeing `NO_COLUMN_NAME_FOUND` twice and `CAT_INSUFFICIENT_PRIVILEGE_LEVEL`, but I haven't looked into the exact source of this problem yet. I don't believe that `rstrcpy(...)` caused this, though…

korydraughn:
The `icd` failure sounds like a _fix the test_ kind of thing. If you're confident that is what it is, include a commit that fixes that test. You can use issue number 858 for things like that.

Please investigate the `iadmin` test failure and report what you find here. We can determine if that deserves a new issue based on what you find.

alanking:
The `test_iadmin` failure is likely from the same PR.

An equivalent PR was opened for 4-3-stable here for the iCommands: https://github.com/irods/irods_client_icommands/pull/332

And the tests were adjusted in these two commits in the main repo:
https://github.com/irods/irods/pull/6637 (main)
https://github.com/irods/irods/pull/6654 (4-3-stable)

Please confirm that the built iCommands are on the right commit

MartinFlores751:
Got it. My upstream was only tracking main for some reason? Going to run the full suite again, just in case.

MartinFlores751:
All tests passed!

korydraughn:
Add the pounds!

PR #6735, opened 2022-11-21
gethostname and client_api_whitelist (4-2-stable)

korydraughn:
These changes are trivial and don't require a full run of the test suites. I

I think compiling and executing an icommand to verify things work is acceptable. On top of that, I'll run the unit tests to make sure things are in order.


korydraughn:
I ran the unit tests twice and all of them passed.

If there aren't any objections, I'll add the pounds.

trel:
seems good to me.

alanking (approved):

PR #6736, opened 2022-11-22
[6716] Put base64 methods in irods namespace

JustinKyleJames:


lib/core/include/irods/base64.h
```diff
@@ -11,17 +11,12 @@
...
-#ifdef __cplusplus
-}
-#endif
+namespace irods
+{
+    int base64_encode( const unsigned char *in,  unsigned long inlen,
+                       unsigned char *out, unsigned long *outlen );
+    int base64_decode( const unsigned char *in,  unsigned long inlen,
+                       unsigned char *out, unsigned long *outlen );
+} // namespace irods
```
alanking, line 20: Since we are removing the C linkage, should we rename this file to base64.hpp to further indicate that C clients should not be using it?

alanking (commented):
A lot of the diff seems to be a clang-format run, so maybe run that once more to make sure it's happy.

Had another question/concern in the header file as well.

Also-also, more words in the commit message could be helpful. I know the issue will be linked, but, y'know. ;)

lib/core/include/irods/base64.h
```diff
@@ -11,17 +11,12 @@
 
 #include "irods/rodsDef.h"
```
korydraughn, line 12: This doesn't look like it is needed.

Try removing it and seeing if things still compile. If it fails to compile, put this `#include` in the cpp file.

lib/core/src/base64.cpp
```diff
@@ -4,135 +4,139 @@
...
-        unsigned a = in[0];
-        unsigned b = ( i + 1 < inlen ) ? in[1] : 0;
+    /**
+       base64 Encode a buffer (NUL terminated)
+       @param in      The input buffer to encode
+       @param inlen   The length of the input buffer
+       @param out     [out] The destination of the base64 encoded data
+       @param outlen  [in/out] The max size and resulting size
+       @return OK if successful
+    */
```
korydraughn, line 48: Let's move this to the header file.

The contents needs to be fixed to match how we document things now. See https://github.com/irods/irods/blob/c3c62ad4cfe8064bcc1475839cc79f362d0e1caa/lib/administration/user/include/irods/user_administration.hpp for examples.

lib/core/src/base64.cpp
```diff
@@ -4,135 +4,139 @@
...
-            /* prevent g < 0 which would potentially allow an overflow later */
-            if ( --g < 0 ) {
+    /**
+       base64 decode a block of memory
+       @param in       The base64 data to decode
+       @param inlen    The length of the base64 data
+       @param out      [out] The destination of the binary decoded data
+       @param outlen   [in/out] The max size and resulting size of the decoded data
+       @return OK if successful
+    */
```
korydraughn, line 94: Let's move this to the header file.

lib/core/src/base64.cpp
```diff
@@ -4,135 +4,139 @@
 
 #include "irods/rodsClient.h"
 #include "irods/base64.h"
+
 #define OK 0
```
korydraughn, line 8: Consider removing this and replacing all usage with `0`.

lib/core/src/base64.cpp
```diff
@@ -4,135 +4,139 @@
 
 #include "irods/rodsClient.h"
 #include "irods/base64.h"
```
korydraughn, line 6: Please swap the order of these includes and delete all lines above them.

korydraughn (commented):
The code changes are good.

My comments are mostly about modernizing the structure of the file. For these smaller files, we should go ahead and improve them when we touch them. That may not be possible for all instances, but these base64 files seem perfectly safe to update.

I can't leave comments on some of the things, so I'll mention them here.
- Delete everything before `#ifndef BASE64_H__`
- Change `BASE64_H__` to `IRODS_BASE64_H`
- Insert `/// \file` before the `#include`s in the header file

You can do that in a separate commit and attach issue number 858 to it.

lib/core/include/irods/base64.h
korydraughn: Given that this change breaks anyone originally relying on these functions, it is okay to change the file extension.

After saying that, we need to think about this a bit more.

Given that the dev pkg exposes almost everything used in iRODS, we can't break people's code. We need to provide a solution that let's the user transition gracefully or we need to define a list of things people are allowed to use.

Below is a potential plan for the transitional approach.
- Leave the old functions in base64.h and mark them deprecated
- Create a new file with the .hpp extension and copy @JustinKyleJames's new version of base64.h into it
- Add a `#warning` to the base64.h file that let's devs know the header is deprecated and that they shouldn't use it anymore. The warning could point the dev to the cpp file or maybe we don't ship the hpp version at all

JustinKyleJames: Are clients using this directly?  I have my doubts whether they are.

The Globus client only used it indirectly via the clientLogin() call.

JustinKyleJames:
This passed all testing.  I will run again once I make the changes mentioned above.

lib/core/include/irods/base64.h
JustinKyleJames: done

lib/core/src/base64.cpp
JustinKyleJames: done

lib/core/include/irods/base64.h
JustinKyleJames: done

lib/core/src/base64.cpp
JustinKyleJames: done

lib/core/src/base64.cpp
JustinKyleJames: done

lib/core/src/base64.cpp
JustinKyleJames: done

lib/core/src/base64.cpp
JustinKyleJames: done

lib/core/src/base64.cpp
JustinKyleJames: done

korydraughn (commented):
Looks good to me.

The only remaining question is whether we should instruct people to use the base64.hpp header (as mentioned in the warning message).

I think it is fine overall. The functions are in the irods namespace now and their implementations are correct. This means these functions are only available to C++ developers.

Making base64.hpp private can be handled in a separate PR.

lib/core/include/irods/base64.h
alanking: I still see the `#include`...

lib/core/src/base64.cpp
```diff
@@ -1,138 +1,120 @@
...
-    if ( y != 0 ) {
-        return BASE64_INVALID_PACKET;
-    }
-    *outlen = z;
-    return OK;
+} // namespace irods
+
+// For backwards compatibility we are keeping but deprecating
+// the methods that are not in the irods namespace.
+int base64_encode(const unsigned char* in, unsigned long inlen, unsigned char* out, unsigned long* outlen)
```
alanking, line 112: Should we mark the declarations in the header file with `__attribute__(deprecated)` as well?

lib/core/include/irods/base64.h
JustinKyleJames: The requested changes have been implemented.

lib/core/include/irods/base64.h
```diff
@@ -1,14 +1,8 @@
-/*** Copyright (c), The Regents of the University of California            ***
- *** For more information please refer to files in the COPYRIGHT directory ***/
-
-/* base64.h - header file for base64.c
- */
-
-
-
 #ifndef BASE64_H__
 #define BASE64_H__
```
alanking, line 2: I believe @korydraughn mentioned renaming these header guards to `IRODS_BASE64_H`

alanking (commented):
Please ignore my comments if they're addressed in a to-be-pushed commit :)

lib/core/include/irods/base64.h
JustinKyleJames: He mentioned it for the HPP one.  I was intending to leave he old deprecated one alone although I did remove the first few lines.

JustinKyleJames: I suppose since I have already touched it I might as well rename the header guard.

JustinKyleJames: done

lib/core/src/base64.cpp
JustinKyleJames: I did this.  Please take a look to make sure it is done correctly.  I have never used that.

JustinKyleJames:
I forgot to add/commit base64.hpp.  It has been added now.

lib/core/include/irods/base64.h
```diff
@@ -1,27 +1,26 @@
...
-                   unsigned char *out, unsigned long *outlen );
-int base64_decode( const unsigned char *in,  unsigned long inlen,
-                   unsigned char *out, unsigned long *outlen );
+// clang-format off
+
+__attribute__((deprecated("base64_encode is deprecated. Use irods::base64_encode instead.")))
+int base64_encode(const unsigned char* in, unsigned long inlen, unsigned char* out, unsigned long* outlen);
+
+__attribute__((deprecated("base64_decode is deprecated. Use irods::base64_decode instead.")))
+int base64_decode(const unsigned char* in, unsigned long inlen, unsigned char* out, unsigned long* outlen);
```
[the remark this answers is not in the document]
JustinKyleJames: From my testing it seems to work fine using these in a test program.  You just get the warning on compile.  After removing the warning on the header itself, iRODS compiles fine as it isn't actually calling these methods.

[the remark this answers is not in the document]
SwooshyCueb: As we discovered previously, there's overlap with declarations from another library. I think it would be preferable to deprecate the header itself rather than the declarations, but I don't feel super strongly about it.

lib/core/src/packStruct.cpp
```diff
@@ -2028,8 +2028,9 @@ namespace
         if ( packTypeTable[typeInx].number == PACK_BIN_TYPE ) {
             /* bin type. need to decode */
             unsigned long outLen = len;
-            if ( int status = base64_decode( static_cast<const unsigned char *>(inPtr), inLen,
-                        static_cast<unsigned char *>(outPtr), &outLen ) ) {
+            if (int status = irods::base64_decode(
+                    static_cast<const unsigned char*>(inPtr), inLen, static_cast<unsigned char*>(outPtr), &outLen))
+            {
```
alanking, line 2033: Well, I quite hate this. :p

alanking (dismissed):
Other than clang-format thing, seems good to me, assuming everything is building and tests pass.

lib/core/src/packStruct.cpp
korydraughn: Yeah. It's not pretty. PackStruct code isn't pretty either lol.

I'm fine with letting this ride though.

lib/core/include/irods/base64.hpp
```diff
@@ -0,0 +1,32 @@
...
+    /// \param[out]     The destination of the base64 encoded data
+    /// \param[in,out]  The max size and resulting size
+    ///
+    /// \return 0 if successful
+    ///
+    /// \since 4.3.1
+    int base64_encode(const unsigned char* in, unsigned long inlen, unsigned char* out, unsigned long* outlen);
+
+    /// Base64 decode a block of memory
+    /// \param[in]      The base64 data to decode
```
korydraughn, line 21: Let's add a newline between these lines so that doxygen doesn't get tripped up by it and so that it matches the documentation on the function above.

alanking:
I think everything has been addressed here? Please resolve comments if true and we shall await one more approval.

korydraughn (approved):
Looks good. Please squash it and leave off the pounds.

JustinKyleJames:
Squashed everything into one pull request.

korydraughn (dismissed):
There's one last comment that needs to be marked as resolved if it has been handled.

lib/core/include/irods/base64.h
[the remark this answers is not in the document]
JustinKyleJames: The header has been deprecated and the declarations left alone.

JustinKyleJames:
> There's one last comment that needs to be marked as resolved if it has been handled.

Done

korydraughn:
Let's add the pound and get this merged.

SwooshyCueb (approved):

JustinKyleJames:
Pound added to commit.

alanking (approved):

Issue #3564, opened 2017-05-30
After iexit full with PAM, icommands ask for Native irods password.


RobvanSchip:
- [x] main
- [x] 4-3-stable
---
Version 4.1.8

Using PAM authentication (in the .irods/irods_environment.json) , when I do a "iexit full" and then a icommand for example:

[rob@elitebook ~]$ ils
Enter your current iRODS password:

It asks for my iRODS password instead of my PAM password. 

What I think it should do, is check if I use Native or PAM and act accordingly:

So in my case ask: "Enter your current PAM password:"

trel:
Hi @RobvanSchip - this is an artifact of the authentication mechanism being handled primarily by `iinit`.  I think `clientLogin()` can be adjusted to handle this case for all the other iCommands.  We'll look into it.   Thanks.

alanking:
Related to efforts in #5479

alanking:
I think this has been resolved in 4.3.0:
```bash
$ ils
/tempZone/home/bobby:
$ iexit full
$ ils
Enter your current PAM password:
/tempZone/home/bobby:
```
However, I see the following in the server log after the `ils` completes which prompts for the password:
```
$ tail -n2 /var/log/irods/irods.log | jq '.log_message'
"[-]\t/irods_source/server/core/src/rodsAgent.cpp:588:int runIrodsAgentFactory(sockaddr_un) :  status [SSL_SHUTDOWN_ERROR]  errno [] -- message [failed to call 'agent stop']\n\t[-]\t/irods_source/lib/core/src/sockComm.cpp:160:irods::error sockAgentStop(irods::network_object_ptr) :  status [SSL_SHUTDOWN_ERROR]  errno [] -- message [failed to call 'agent stop']\n\t\t[-]\t/irods_source/plugins/network/src/ssl.cpp:850:irods::error ssl_agent_stop(irods::plugin_context &) :  status [SSL_SHUTDOWN_ERROR]  errno [] -- message [error completing shutdown of SSL connection]\n\n"
"Agent process [4444] exited with status [1]"
```
This does not happen with native authentication while SSL is enabled, so we will need to investigate why that is happening.

alanking:
Note: The message only appears in the catalog service provider's log when the client is connected to a catalog service consumer using `pam_password` authentication and SSL is enabled. When the client is connected directly to the catalog service provider, things work fine.

alanking:
[...]

PR #6755, opened 2022-12-02
[#3564] pam_password: CSP redirect only enables SSL if necessary (main)

alanking:
Addresses lingering issue found here (i.e. not the main issue, which has already been fixed): https://github.com/irods/irods/issues/3564#issuecomment-1256617949

Confirmed that the issue is resolved only at the bench because the issue manifests as log noise, as far as I can tell. If anybody can think of a way in which this might present a more serious problem which we can capture in a test, I will write an automated test.

### DIFF
--- a/plugins/auth/src/pam_password.cpp
+++ b/plugins/auth/src/pam_password.cpp
@@ -271,11 +271,27 @@ namespace irods
 
                 log_auth::trace("redirecting call to CSP");
 
-                if (const auto ec = sslStart(host->conn); ec) {
-                    THROW(ec, "could not establish SSL connection");
-                }
+                // Need to enable SSL here if it is not already being used because the PAM password
+                // is forwarded to the provider in the clear.
+                // clang-format off
+                const bool using_ssl = (0 == std::strncmp(
+                    irods::CS_NEG_USE_SSL.c_str(),
+                    host->conn->negotiation_results, // NOLINT(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+                    MAX_NAME_LEN));
+                // clang-format on
 
-                const auto end_ssl = irods::at_scope_exit{[host] { sslEnd(host->conn); }};
+                const auto end_ssl_if_we_enabled_it = irods::at_scope_exit{[host, using_ssl] {
+                    if (!using_ssl) {
+                        sslEnd(host->conn);
+                    }
+                }};
+
+                if (!using_ssl) {
+                    if (const int ec = sslStart(host->conn); ec) {
+                        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+                        THROW(ec, "failed to enable SSL in server-to-server communication");
+                    }
+                }
 
                 return irods_auth::request(*host->conn, req);
             }


plugins/auth/src/pam_password.cpp
```diff
@@ -271,11 +271,26 @@ namespace irods
...
-                if (const auto ec = sslStart(host->conn); ec) {
-                    THROW(ec, "could not establish SSL connection");
-                }
+                // Need to enable SSL here if it is not already being used because the PAM password
+                // is forwarded to the provider in the clear.
+                // clang-format off
+                const bool using_ssl = 0 == std::strncmp(
+                        irods::CS_NEG_USE_SSL.c_str(),
+                        host->conn->negotiation_results, // NOLINT(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+                        MAX_NAME_LEN);
```
korydraughn, line 280: Indentation?

korydraughn (commented):
Looks good.

plugins/auth/src/pam_password.cpp
korydraughn: I was referring to the indentation appearing to use more than 4 spaces.
```cpp
const bool using_ssl = 0 == std::strncmp(
￼        irods::CS_NEG_USE_SSL.c_str(),
￼        host->conn->negotiation_results, // NOLINT(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
￼        MAX_NAME_LEN);
```
VS
```cpp
const bool using_ssl = 0 == std::strncmp(
￼   irods::CS_NEG_USE_SSL.c_str(),
￼   host->conn->negotiation_results, // NOLINT(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
￼   MAX_NAME_LEN);
```
The `NOLINTBEGIN`/`END` thing is not available in Clang 13's clang-tidy. I've been using those knowing that we'll be upgrading the compiler some time in the future.

alanking: Ah, okay. Thanks. Will fix this up.

korydraughn (approved):
If all tests pass, add the pounds and merge.

trel:
agreed, remember to squash first

alanking:
Well, the only test which would exercise this code is skipped when run in topology from the resource server or when SSL is enabled, so there's that. It made me think that maybe we should consider making a mode for the test suite which uses pam_password. The rest of the test suite outside of `test_auth` uses native authentication and would never hit this.

Anyway, I confirmed at the bench that the fix addresses the linked issue and the test suite is passing.

korydraughn:
> It made me think that maybe we should consider making a mode for the test suite which uses pam_password.

Yes. We need to exercise as much of the code as possible. Try thinking about what a pam_password environment would look like. Perhaps it can be captured in an issue if not already.

> Anyway, I confirmed at the bench that the fix addresses the linked issue and the test suite is passing.

Good. Please merge when ready.

trel (approved):

alanking:
I created https://github.com/irods/irods/issues/6785 for the authentication method specification when running tests. Thanks!

PR #6757, opened 2022-12-02
[6726] Make `server_properties::map()` thread-safe and read-only

MartinFlores751:
Make `server_properties::map()` thread-safe and read-only, allowing for safe access to internal JSON object.

### DIFF
--- a/lib/core/include/irods/irods_server_properties.hpp
+++ b/lib/core/include/irods/irods_server_properties.hpp
@@ -148,13 +148,48 @@ namespace irods
 
         void remove( const std::string& _key );
 
-        /// \brief Get a reference to the underlying json object. Writing
-        /// in particular should be done with caution as there is no guarantee
-        /// that this will be safe.
-        /// In threaded/asynchronous contexts, this function bypasses safety guarantees.
-        nlohmann::json& map() noexcept
+    private:
+        class [[nodiscard("Do not discard lock!")]] readonly_config_handle
         {
-            return config_props_;
+          public:
+            readonly_config_handle(const nlohmann::json& json, std::shared_mutex& mutex)
+                : server_config{json}
+                , lock{mutex}
+            {
+            }
+
+            readonly_config_handle(const readonly_config_handle&) = delete;
+            auto operator=(const readonly_config_handle&)->readonly_config_handle& = delete;
+
+            [[nodiscard]] auto get_json() const& noexcept->const nlohmann::json&
+            {
+                return server_config;
+            }
+
+          private:
+            const nlohmann::json& server_config;
+            std::shared_lock<std::shared_mutex> lock;
+        };
+
+    public:
+        /// \brief Acquire a read-only handle containing the server configuration.
+        ///
+        /// This function acquires a read-only handle that ensures, for the lifetime of the handle, the server
+        /// configuration cannot be modified by other threads of execution.
+        ///
+        /// Example of usage is as follows:
+        /// \code{.cpp}
+        /// const auto config_handle{irods::server_properties::server_properties::instance().map()};
+        /// const auto& config{config_handle.get_json()};
+        /// \endcode
+        ///
+        /// \warning Do not store the handle in a `const auto&` variable. This will cause the lifetime of the handle
+        ///          to extend past what is normally expected, and result in code containing unexpected behavior.
+        ///
+        /// \since 4.3.1
+        readonly_config_handle map() const noexcept
+        {
+            return readonly_config_handle{config_props_, property_mutex_};
         }
 
     private:
--- a/plugins/microservices/src/get_server_property.cpp
+++ b/plugins/microservices/src/get_server_property.cpp
@@ -32,7 +32,8 @@ namespace
         try {
             const std::string name = parseMspForStr(_in_name);
             const nlohmann::json* json_value = nullptr;
-            const auto& config = irods::server_properties::server_properties::instance().map();
+            const auto config_handle{irods::server_properties::server_properties::instance().map()};
+            const auto& config{config_handle.get_json()};
 
             try {
                 if (name.find("/") != std::string::npos) {
--- a/unit_tests/src/test_server_properties.cpp
+++ b/unit_tests/src/test_server_properties.cpp
@@ -7,6 +7,8 @@
 
 #include <fmt/format.h>
 
+#include <atomic>
+#include <string>
 #include <thread>
 
 TEST_CASE("server_properties", "[exceptions]")
@@ -53,6 +55,45 @@ TEST_CASE("server_properties is threadsafe", "[thread-safety]")
         if (irods::server_property_exists(key)) {
             REQUIRE(irods::get_server_property<int>(key) == i);
             i++;
+        }
+    }
+}
+
+TEST_CASE("server_properties map function is threadsafe", "[thread-safety]")
+{
+    constexpr auto limit{1000};
+
+    // Populate object
+    for (int i{}; i < limit; i++) {
+        irods::set_server_property(std::to_string(i), i);
+    }
+
+    std::atomic_bool do_busy_loop{true};
+
+    // Attempt to contend access with write lock
+    std::thread persistent_write([&do_busy_loop]() {
+        // Actively wait for chance to run
+        while (do_busy_loop) {
+        }
+        for (int i{}; i < limit; i++) {
+            irods::set_server_property(std::to_string(i), -1);
+        }
+    });
+    irods::at_scope_exit exit_action([&persistent_write]() { persistent_write.join(); });
+
+    // Acquire read handle
+    const auto config_handle{irods::server_properties::server_properties::instance().map()};
+
+    // Allow write attempts from 'persistent_write'
+    do_busy_loop = false;
+
+    const auto& config{config_handle.get_json()};
+
+    // If lock works, object should be as originally populated
+    for (int i{}; i < limit; i++) {
+        auto key{std::to_string(i)};
+        if (auto res{config.find(key)}; res != std::end(config)) {
+            REQUIRE(*res == i);
         }
     }
 }
--- a/server/core/src/catalog.cpp
+++ b/server/core/src/catalog.cpp
@@ -71,7 +71,9 @@ namespace irods::experimental::catalog {
             capture_database_info(config, db_instance_name, db_username, db_password);
         }
         else {
-            capture_database_info(server_properties::instance().map(), db_instance_name, db_username, db_password);
+            const auto config_handle{server_properties::instance().map()};
+            const auto& config{config_handle.get_json()};
+            capture_database_info(config, db_instance_name, db_username, db_password);
         }
 
         try {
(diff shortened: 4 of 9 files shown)


unit_tests/src/test_server_properties.cpp
```diff
@@ -56,3 +56,27 @@ TEST_CASE("server_properties is threadsafe", "[thread-safety]")
...
+    const auto limit{1000};
+
+    // Start a thread that will generally get in the way a lot by writing
+    // to the irods::server_properties object
+    std::thread persistent_write([]() {
+        for (auto i = 0; i < limit; i++) {
+            irods::set_server_property(std::to_string(i), i);
+        }
+    });
+    irods::at_scope_exit([&persistent_write]() { persistent_write.join(); });
```
korydraughn, line 71: Your `at_scope_exit` needs a name.

`persistent_write` joins before the logic below is ever executed.

unit_tests/src/test_server_properties.cpp
```diff
@@ -56,3 +56,27 @@ TEST_CASE("server_properties is threadsafe", "[thread-safety]")
...
+TEST_CASE("server_properties map function is threadsafe", "[thread-safety]")
+{
+    const auto limit{1000};
+
+    // Start a thread that will generally get in the way a lot by writing
+    // to the irods::server_properties object
+    std::thread persistent_write([]() {
+        for (auto i = 0; i < limit; i++) {
+            irods::set_server_property(std::to_string(i), i);
+        }
```
korydraughn, line 69: Consider making the thread sleep for a few milliseconds so that there is overlap.

korydraughn (changes requested):
Good job.

My understanding of C++ references has been pushed because of this issue. While reviewing this, I learned that the following is legal.
```cpp
struct thing
{
    int& x;
};

int main()
{
    int i = 5;
    const thing j{i};
    ++j.x;
    std::cout << i << '\n'; // Produces 6!
}
```
It's very easy to assume that because `j` is const, its member, `x`, must also be const. Yet, `i` can be modified through member `x`.

This is very surprising to me.

---

After asking about this on stackoverflow, the reason becomes clear.

The **binding** of the reference is const, which is always the case. The thing being bound is not const.

References are just fancy pointers. This is equivalent to `int *const x;`.

unit_tests/src/test_server_properties.cpp
MartinFlores751: Got around this by mimicking `jthread` cooperative cancelling instead.

lib/core/include/irods/irods_server_properties.hpp
```diff
@@ -148,13 +148,33 @@ namespace irods
...
+            std::shared_lock<std::shared_mutex> lock;
+        };
+
+public:
         /// \brief Get a reference to the underlying json object. Writing
         /// in particular should be done with caution as there is no guarantee
         /// that this will be safe.
         /// In threaded/asynchronous contexts, this function bypasses safety guarantees.
-        nlohmann::json& map() noexcept
+        std::pair<const nlohmann::json&, read_lock> map() const noexcept
```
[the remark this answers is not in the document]
MartinFlores751: This would likely still allow for the lock to be stolen, so I don't believe that I took a proper look at this.

lib/core/include/irods/irods_server_properties.hpp
```diff
@@ -148,13 +148,32 @@ namespace irods
 
         void remove( const std::string& _key );
 
-        /// \brief Get a reference to the underlying json object. Writing
-        /// in particular should be done with caution as there is no guarantee
-        /// that this will be safe.
-        /// In threaded/asynchronous contexts, this function bypasses safety guarantees.
-        nlohmann::json& map() noexcept
+    private:
+        class [[nodiscard("Do not discard lock!")]] read_lock
```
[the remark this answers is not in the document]
korydraughn: I believe a private class is good. We don't want anyone to be able to construct or refer to that type and forcing people to use `auto` seems fine to me.

I agree that the name needs to be changed to something that better expresses the intent.

unit_tests/src/test_server_properties.cpp
```diff
@@ -56,3 +58,31 @@ TEST_CASE("server_properties is threadsafe", "[thread-safety]")
...
+    });
+    irods::at_scope_exit exit_action([&persistent_write, &stop_flag]() { stop_flag = true; persistent_write.join(); });
+
+    auto i{0};
+    while (i < limit) {
+        auto key = std::to_string(i);
+        const auto lock{irods::server_properties::server_properties::instance().map()};
+        const auto& config{lock.get_json()};
+        if (auto res{config.find(key)}; res != std::end(config)) {
+            REQUIRE(*res == i);
```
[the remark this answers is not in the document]
korydraughn: I think you're right.

lib/core/include/irods/irods_server_properties.hpp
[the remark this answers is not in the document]
MartinFlores751: Went with `config_handle` for the class name.

unit_tests/src/test_server_properties.cpp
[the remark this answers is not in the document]
MartinFlores751: Updated test should be better!

plugins/microservices/src/get_server_property.cpp
```diff
@@ -32,7 +32,8 @@ namespace
         try {
             const std::string name = parseMspForStr(_in_name);
             const nlohmann::json* json_value = nullptr;
-            const auto& config = irods::server_properties::server_properties::instance().map();
+            const auto lock{irods::server_properties::server_properties::instance().map()};
```
korydraughn, line 35: Consider renaming variable (e.g. `cfg_handle`).

`lock` does not convey what it is very well.

unit_tests/src/test_server_properties.cpp
```diff
@@ -56,3 +58,42 @@ TEST_CASE("server_properties is threadsafe", "[thread-safety]")
...
+    do_busy_loop = false;
+
+    const auto& config{lock.get_json()};
+
+    // If lock works, object should be as originally populated
+    for (int i{}; i < limit; i++) {
+        auto key{std::to_string(i)};
+        if (auto res{config.find(key)}; res != std::end(config)) {
+            REQUIRE(*res == i);
+            i++;
```
korydraughn, line 96: Is this needed?

lib/core/include/irods/irods_server_properties.hpp
```diff
@@ -148,13 +148,32 @@ namespace irods
...
+                return server_config;
+            }
+
+          private:
+            const nlohmann::json& server_config;
+            std::shared_lock<std::shared_mutex> lock;
+        };
+
+    public:
+        /// \brief Acquire a config_handle object containing a const reference to the underlying json object.
```
korydraughn, line 173: Add words explaining what `config_handle` is for. You can include example code if you want. There's a few functions in the codebase that provide example code. See irods_logger.hpp for an example.

Missing `\since` property.

unit_tests/src/test_server_properties.cpp
```diff
@@ -1,3 +1,4 @@
+#include <atomic>
```
korydraughn, line 1: Move this header so that it is grouped with `<string>` and `<thread>`.

korydraughn (commented):
The test looks much better.

lib/core/include/irods/irods_server_properties.hpp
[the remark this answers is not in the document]
korydraughn: `config_handle` is very generic. I think we need to add a word that describes what the type is for.

Perhaps something like `readonly_config_handle`?

lib/core/include/irods/irods_server_properties.hpp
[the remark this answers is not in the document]
MartinFlores751: Update, messed with `const std::shared_lock<std::shared_mutex>` in compiler explorer, and it works! Still, if we wish to use a pair we need to be sure _not_ to use `const auto& [data, lock]`, and be sure that we don't drop the lock either (e.g., `const auto& data = server_properties::map().first`)

Something to think about.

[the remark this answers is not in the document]
korydraughn: Use the design that makes it difficult to misuse the interface.

lib/core/include/irods/irods_server_properties.hpp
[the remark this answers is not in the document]
MartinFlores751: Done!

lib/core/include/irods/irods_server_properties.hpp
```diff
@@ -170,10 +170,21 @@ namespace irods
         };
 
     public:
-        /// \brief Acquire a config_handle object containing a const reference to the underlying json object.
-        config_handle map() const noexcept
+        /// \brief Acquire a readonly_config_handle object containing a const reference to the underlying json object.
+        ///
+        /// This function is used to acquire a read-only handler, to then access the underlying nlohmann::json object
+        /// containing the server's properties.
+        ///
```
korydraughn, line 177: Better, but it still needs a few words explaining why the handle is necessary. For example, you could say something like:
```
The readonly_config_handle guarantees that, for the lifetime of the handle, no other
thread can modify the object returned by config_handle::get_json().
```

alanking:
Please resolve any open conversation if they are indeed resolved. Looks good to squash to me. Any other thoughts?

trel:
a little clang-format and squash em up.

korydraughn:
We need to ignore clang-format for situations involving access specifiers (i.e. `public`, `private`) because it screws up the formatting of structs/classes.

It's fine to format other parts of the PR though.

unit_tests/src/test_server_properties.cpp
```diff
@@ -56,3 +58,42 @@ TEST_CASE("server_properties is threadsafe", "[thread-safety]")
...
+        while (do_busy_loop) {
+        }
+        for (int i{}; i < limit; i++) {
+            irods::set_server_property(std::to_string(i), -1);
+        }
+    });
+    irods::at_scope_exit exit_action([&persistent_write]() { persistent_write.join(); });
+
+    // Aquire read lock
+    const auto lock{irods::server_properties::server_properties::instance().map()};
```
korydraughn, line 85: lock -> config_handle

unit_tests/src/test_server_properties.cpp
```diff
@@ -56,3 +58,42 @@ TEST_CASE("server_properties is threadsafe", "[thread-safety]")
...
+
+    // Allow write attempts from 'persistent_write'
+    do_busy_loop = false;
+
+    const auto& config{lock.get_json()};
```
korydraughn, line 90: lock -> config_handle

korydraughn (commented):
Just two small tweaks and I think we can get this in.

Are the tests happy?

MartinFlores751:
Ran the unit and core tests before the latest commit, and all tests passed.

korydraughn:
Good. Squash commits and leave off the pounds.

alanking (approved):
Let's add #s

alanking (approved):

Issue #6370, opened 2022-04-29
Consider exposing timeout options for controlling delay server migration communication


korydraughn:
- [x] main
- [x] 4-3-stable
- [x] irods_docs (main only)
---
Depending on the network, admins may need to adjust the timeout values so that delay server migration operates correctly.
https://github.com/irods/irods/blob/db858d950a9efe6345f78ccb55e1f1d2f905bbf2/server/main_server/src/rodsServer.cpp#L576-L577

alanking:
Hmm... now that I've already opened all the PRs, I have some questions... :)

There are 4 levels of granularity for the socket timeout configuration that seem consistent to me:
1. Use `server_control_plane_timeout_milliseconds` for send and receive of all internal control plane message type
2. Use one value for the main control plane configuration and one value for each additional internal control plane message type (of which there is currently one(?), delay server migration)
3. Use separate values for send and receive for the main control plane configuration which will also serve as configuration for all internal control plane message types
4. Use separate values for send and receive for the main control plane configuration as well as separate values for each additional internal control plane message type

The implementation that I put up in #6762 is none of these options. It is one configuration value for send and receive for the main control plane, and separate values for send and receive for just delay server migration communications. This feels disjointed and confusing to me.

I think options 1 and 2 are my favorites, but I also don't claim to understand what system administrators may need to configure. For reference, here is where we are using the single configuration value for the control plane to set the send and receive timeouts:
https://github.com/irods/irods/blob/2f602874739f3d0d15551127cddbac181fb88e27/server/main_server/src/main_server_control_plane.cpp#L77
https://github.com/irods/irods/blob/2f602874739f3d0d15551127cddbac181fb88e27/server/main_server/src/main_server_control_plane.cpp#L91-L92

Thoughts?

trel:
Do we have a real/good/known use case for send and receive values needing to be different from one another?  If not, that would limit us to 1 or 2.

Since whatever we're doing (now and in the future) in a single iRODS zone is going over the same network as any thing else we're doing in that zone (as far as I know)... there is maybe no reason to have option 2?   aka it's good enough for the main server, it should be good enough for any others on the same network?   Then we are at option 1.

I can imagine different applications on a machine having per-app timeout settings... but this level of within-same-app granularity might be overkill...

korydraughn:
It's not clear which approach is best, so I think we start with 1 (the simplest solution).

If that isn't good enough, we implement 2 and so on.

alanking:
[...]

Issue #6342, opened 2022-04-19
Expose CRON settings in advanced settings


korydraughn:
- [x] main
- [x] 4-3-stable
- [x] irods_docs (main only)
---
## Feature
- dns cache
- hostname cache
- stacktrace file processor
- delay server migration

PR #6762, opened 2022-12-07
Expose some new configuration values (CRON tasks, socket timeouts) (main)

alanking:
Addresses #6342 
Addresses #6370 

Core tests passed on ubuntu:20.04.

Couple of notes:
1. Open to changing any and all names and default values for the configurations as well as their affiliated variables in the code.
2. I thought about making `get_advanced_setting` a template function based on the type of the configuration value, but this seemed a little overboard considering everything is just `int`s (for now) and the function is only available in the one file so it doesn't need to be so general. Any thoughts on this would be appreciated.
3. I'm happy to hold on merging this to yield to #6673 due to impending conflicts.
4. Ignoring the clang-tidy issue because it's just getting confused by the generated header file for the control plane commands.

### DIFF
--- a/packaging/server_config.json.template
+++ b/packaging/server_config.json.template
@@ -2,21 +2,25 @@
     "schema_name": "server_config",
     "schema_version": "v4",
     "advanced_settings": {
+        "agent_factory_watcher_sleep_time_in_seconds": 5,
         "default_number_of_transfer_threads": 4,
         "default_temporary_password_lifetime_in_seconds": 120,
         "delay_rule_executors": [],
         "delay_server_sleep_time_in_seconds" : 30,
         "dns_cache": {
             "shared_memory_size_in_bytes": 5000000,
-            "eviction_age_in_seconds": 3600
+            "eviction_age_in_seconds": 3600,
+            "cache_clearer_sleep_time_in_seconds": 600
         },
         "hostname_cache": {
             "shared_memory_size_in_bytes": 2500000,
-            "eviction_age_in_seconds": 3600
+            "eviction_age_in_seconds": 3600,
+            "cache_clearer_sleep_time_in_seconds": 600
         },
         "maximum_size_for_single_buffer_in_megabytes": 32,
         "maximum_size_of_delay_queue_in_bytes": 0,
         "maximum_temporary_password_lifetime_in_seconds": 1000,
+        "migrate_delay_server_sleep_time_in_seconds": 5,
         "number_of_concurrent_delay_rule_executors": 4,
         "stacktrace_file_processor_sleep_time_in_seconds": 10,
         "transfer_buffer_size_for_parallel_transfer_in_megabytes": 4,
--- a/server/main_server/src/rodsServer.cpp
+++ b/server/main_server/src/rodsServer.cpp
@@ -631,8 +631,18 @@ namespace
 
             zmq::context_t zmq_ctx;
             zmq::socket_t zmq_socket{zmq_ctx, zmq::socket_type::req};
-            zmq_socket.set(zmq::sockopt::sndtimeo, 500);
-            zmq_socket.set(zmq::sockopt::rcvtimeo, 5000);
+            const int server_control_plane_timeout_in_milliseconds = [] {
+                try {
+                    const auto config_handle = irods::server_properties::instance().map();
+                    return config_handle.get_json().at(irods::KW_CFG_SERVER_CONTROL_PLANE_TIMEOUT).get<int>();
+                }
+                catch (...) {
+                    constexpr int default_server_control_plane_timeout_in_milliseconds = 10000;
+                    return default_server_control_plane_timeout_in_milliseconds;
+                }
+            }();
+            zmq_socket.set(zmq::sockopt::sndtimeo, server_control_plane_timeout_in_milliseconds);
+            zmq_socket.set(zmq::sockopt::rcvtimeo, server_control_plane_timeout_in_milliseconds);
             zmq_socket.set(zmq::sockopt::linger, 0);
             zmq_socket.connect(zmq_server_target);
             log_server::trace("Connection established.");
--- a/server/main_server/src/rodsServer.cpp
+++ b/server/main_server/src/rodsServer.cpp
@@ -376,26 +376,51 @@ namespace
         fs::create_directories(irods::get_irods_stacktrace_directory(), ec);
     } // create_stacktrace_directory
 
-    int get_stacktrace_file_processor_sleep_time()
+    auto get_advanced_setting(const char* _keyword, int _default, bool _negative_values_allowed = false) -> int
     {
         try {
             const auto config_handle{irods::server_properties::instance().map()};
             const auto& config{config_handle.get_json()};
-            const auto adv_settings{config.at(irods::KW_CFG_ADVANCED_SETTINGS)};
+            const auto& adv_settings{config.at(irods::KW_CFG_ADVANCED_SETTINGS)};
 
-            const auto iter = adv_settings
-                .find(irods::KW_CFG_STACKTRACE_FILE_PROCESSOR_SLEEP_TIME_IN_SECONDS);
+            const auto iter = adv_settings.find(_keyword);
 
             if (iter != std::end(adv_settings)) {
-                if (const auto seconds = iter->get<int>(); seconds > 0) {
-                    return seconds;
+                const auto val = iter->get<int>();
+                if (_negative_values_allowed || val > 0) {
+                    return val;
                 }
             }
         }
         catch (...) {}
 
-        return 10; // The default sleep time.
-    } // get_stacktrace_file_processor_sleep_time
+        return _default;
+    } // get_advanced_setting
+
+    auto get_cache_clearer_sleep_time(const char* _keyword, int _default) -> int
+    {
+        try {
+            const auto config_handle = irods::server_properties::instance().map();
+            const auto& adv_settings = config_handle.get_json().at(irods::KW_CFG_ADVANCED_SETTINGS);
+
+            const auto cache_type_iter = adv_settings.find(_keyword);
+
+            if (cache_type_iter != std::end(adv_settings)) {
+                const auto cache_clearer_iter =
+                    cache_type_iter->find(irods::KW_CFG_CACHE_CLEARER_SLEEP_TIME_IN_SECONDS);
+
+                if (cache_clearer_iter != std::end(*cache_type_iter)) {
+                    if (const auto val = cache_clearer_iter->get<int>(); val > 0) {
+                        return val;
+                    }
+                }
+            }
+        }
+        catch (...) {
+        }
+
+        return _default;
+    } // get_cache_clearer_sleep_time
 
     void log_stacktrace_files()
     {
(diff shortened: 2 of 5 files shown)


alanking:
Oh, also, I did *not* add any automated tests for this mainly because they would be timing based by definition and kind of hard to detect anyway. I can certainly do this if it seems helpful.

I confirmed at the bench that the configurations have the intended effects (except for the socket timeouts).

korydraughn:
I haven't looked over this yet, but some immediate things come to mind that we need to do:
- Verify upgrades from 4.3.0 to 4.3.1 add the new properties to server_config.json if missing
- Update documentation for server_config.json in irods/irods_docs repo
- Explain the ZMQ timeout options for send/recv

alanking:
Right, upgrades. I'll take a look at that.

ZMQ timeout options will be documented along with the other options in irods_docs (see checkboxes in linked issues). Is that what you meant by "explain" or is there something more?

korydraughn:
Yes. That is what I meant.

packaging/server_config.json.template
```diff
@@ -10,13 +11,18 @@
...
             "shared_memory_size_in_bytes": 2500000,
             "eviction_age_in_seconds": 3600
         },
+        "hostname_cache_clearer_sleep_time_in_seconds": 600,
         "maximum_size_for_single_buffer_in_megabytes": 32,
         "maximum_size_of_delay_queue_in_bytes": 0,
         "maximum_temporary_password_lifetime_in_seconds": 1000,
+        "migrate_delay_server_sleep_time_in_seconds": 5,
+        "migrate_delay_server_socket_receive_timeout_in_milliseconds": 5000,
+        "migrate_delay_server_socket_send_timeout_in_milliseconds": 500,
```
alanking, line 25: This made me wonder if we should include a "delay_server" section in the configuration since we now have the following configurations:
 - delay_rule_executors
 - delay_server_sleep_time_in_seconds
 - maximum_size_of_delay_queue_in_bytes
 - migrate_delay_server_sleep_time_in_seconds
 - migrate_delay_server_socket_receive_timeout_in_milliseconds
 - migrate_delay_server_socket_send_timeout_in_milliseconds
 - number_of_concurrent_delay_rule_executors

alanking: I suppose some of these are configurations for the delay server itself and some are configurations for machinery around the delay server, i.e. delay server migration... Anyways, something to consider.

packaging/server_config.json.template
```diff
@@ -10,13 +11,18 @@
             "shared_memory_size_in_bytes": 5000000,
             "eviction_age_in_seconds": 3600
         },
+        "dns_cache_clearer_sleep_time_in_seconds": 600,
```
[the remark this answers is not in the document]
trel: i like a flat hierarchy for these settings i think.   less complicated.  less logic to parse.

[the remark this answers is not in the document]
korydraughn: Could that lead to more confusion given we have a section dedicated to dns_cache?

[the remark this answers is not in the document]
trel: yes, probably.   bah.

packaging/server_config.json.template
korydraughn: That's a possibility. The main challenge is upgrades and expectations of administrators.

There's also the discussion of a flat configuration vs a configuration with nesting.

alanking: Very true. Keeping the flat structure for these options in particular seems fine/preferred, so I'll leave it how it is.

alanking:
The configurations are moving around a little bit, so I'm going to put this in draft. Feel free to review anyway

alanking:
Regarding upgrades...

Is my understanding correct that injecting the new values would require a schema version bump? 

Every new setting that I've added is considered optional, and some of the settings are embedded in structures that are themselves optional (`dns_cache` and `hostname_cache`). So we may need to discuss the rules around this should we desire the values to be injected.

trel:
new things don't require a version bump - as optional/new items will still be valid according to our current schema ( we aren't leveraging additionalProperties... https://json-schema.org/understanding-json-schema/reference/object.html#additionalproperties )

alanking:
Sorry, by "injecting" new values, I meant... adding the new values to an existing server_config.json on upgrade.

So it sounds like our upgrade logic provides a mechanism for doing this without bumping the schema version?

trel:
Yes, I believe we have done this since 4.2.0 on the 4.2.x series - and this should be the same, except perhaps the mechanisms have moved a little.  Defer to @korydraughn.

korydraughn:
This is the latest bit of code that was written to add missing properties. The `update_base` function could be pulled out and used when an upgrade is detected.
https://github.com/irods/irods/blob/2f602874739f3d0d15551127cddbac181fb88e27/scripts/irods/upgrade_configuration.py#L186-L199

I think adding properties without a schema version bump requires a little thought. Basically we need a mechanism similar to how schema version upgrades are handled. Or maybe the version last step in the process is a call to add missing properties. That may work.

packaging/server_config.json.template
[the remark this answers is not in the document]
alanking: I think grouping this into the `dns_cache` and `hostname_cache` sections makes the most sense because the JSON objects already exist and it reduces the number of keywords we need to add. Plus, as we discussed, it would logically group the concepts together so that it is known that they are related.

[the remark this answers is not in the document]
korydraughn: Go for it.

alanking:
We have decided that the new configuration values will not be added on server upgrade because they are not required. See https://github.com/irods/irods/issues/7143.

Squashed and rebased. I think this is ready for review now.

Need to run tests again.

packaging/server_config.json.template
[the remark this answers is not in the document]
alanking: Done

trel:
looks good upon initial inspection.

server/main_server/src/rodsServer.cpp
```diff
@@ -606,8 +631,20 @@ namespace
 
             zmq::context_t zmq_ctx;
             zmq::socket_t zmq_socket{zmq_ctx, zmq::socket_type::req};
-            zmq_socket.set(zmq::sockopt::sndtimeo, 500);
-            zmq_socket.set(zmq::sockopt::rcvtimeo, 5000);
+            const int timeout = [] {
+                try {
+                    return irods::server_properties::instance()
+                        .map()
```
korydraughn, line 637: Same here.

server/main_server/src/rodsServer.cpp
```diff
@@ -376,26 +376,51 @@ namespace
         fs::create_directories(irods::get_irods_stacktrace_directory(), ec);
     } // create_stacktrace_directory
 
-    int get_stacktrace_file_processor_sleep_time()
+    auto get_advanced_setting(const char* _keyword, int _default, bool _negative_values_allowed = false)
+        -> decltype(_default)
```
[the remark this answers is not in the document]
alanking: I think at the time I was playing around with the types and didn't feel like changing it in 3 places. I'll just make it an int.

server/main_server/src/rodsServer.cpp
```diff
@@ -376,26 +376,51 @@ namespace
...
 
-        return 10; // The default sleep time.
-    } // get_stacktrace_file_processor_sleep_time
+        return _default;
+    } // get_advanced_setting
+
+    auto get_cache_clearer_sleep_time(const char* _keyword, int _default) -> decltype(_default)
+    {
+        try {
+            const auto adv_settings = irods::server_properties::instance().map().get_json().at(irods::KW_CFG_ADVANCED_SETTINGS);
```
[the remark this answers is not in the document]
alanking: Makes sense. Will do

server/main_server/src/rodsServer.cpp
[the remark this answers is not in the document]
alanking: Done

server/main_server/src/rodsServer.cpp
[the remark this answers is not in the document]
alanking: Done

server/main_server/src/rodsServer.cpp
alanking: Done

alanking:
Force-push due to rebase. New changes are in separate commits.

Clang-tidy has two complaints which I'm willing to address, but neither are new to my changes. Let me know if these are important to you all:
1. "Cognitive complexity" of `main`
2. Magic numbers for the config values

---

Unit tests passed. Running core tests now.

korydraughn:
Neither of those clang-tidy checks matter much in regard to your changes IMO.

Will reiew the latest commits.

server/main_server/src/rodsServer.cpp
```diff
@@ -631,8 +631,20 @@ namespace
 
             zmq::context_t zmq_ctx;
             zmq::socket_t zmq_socket{zmq_ctx, zmq::socket_type::req};
-            zmq_socket.set(zmq::sockopt::sndtimeo, 500);
-            zmq_socket.set(zmq::sockopt::rcvtimeo, 5000);
+            const int timeout = [] {
```
korydraughn, line 634: Consider changing the variable name so that it's clear we're dealing with milliseconds.

Perhaps `timeout_ms`?

Or maybe we introduce a new config keyword as an alias. This new keyword would include `_IN_MILLISECONDS` or something to help readers understand the `10000`.

server/main_server/src/rodsServer.cpp
```diff
@@ -398,10 +397,11 @@ namespace
         return _default;
     } // get_advanced_setting
 
-    auto get_cache_clearer_sleep_time(const char* _keyword, int _default) -> decltype(_default)
+    auto get_cache_clearer_sleep_time(const char* _keyword, int _default) -> int
     {
         try {
-            const auto adv_settings = irods::server_properties::instance().map().get_json().at(irods::KW_CFG_ADVANCED_SETTINGS);
+            const auto config_handle = irods::server_properties::instance().map();
+            const auto adv_settings = config_handle.get_json().at(irods::KW_CFG_ADVANCED_SETTINGS);
```
korydraughn, line 404: `const auto&`?

I just realized that L384/385 also needs to be a reference.

korydraughn (commented):
Just a couple things and then I think we'll be in a good spot.

server/main_server/src/rodsServer.cpp
alanking: Something like this?
```diff
--- a/server/main_server/src/rodsServer.cpp
+++ b/server/main_server/src/rodsServer.cpp
@@ -631,17 +631,18 @@ namespace
 
             zmq::context_t zmq_ctx;
             zmq::socket_t zmq_socket{zmq_ctx, zmq::socket_type::req};
-            const int timeout = [] {
+            const int server_control_plane_timeout_in_milliseconds = [] {
                 try {
                     const auto config_handle = irods::server_properties::instance().map();
                     return config_handle.get_json().at(irods::KW_CFG_SERVER_CONTROL_PLANE_TIMEOUT).get<int>();
                 }
                 catch (...) {
-                    return 10000;
+                    constexpr int default_server_control_plane_timeout_in_milliseconds = 10000;
+                    return default_server_control_plane_timeout_in_milliseconds;
                 }
             }();
-            zmq_socket.set(zmq::sockopt::sndtimeo, timeout);
-            zmq_socket.set(zmq::sockopt::rcvtimeo, timeout);
+            zmq_socket.set(zmq::sockopt::sndtimeo, server_control_plane_timeout_in_milliseconds);
+            zmq_socket.set(zmq::sockopt::rcvtimeo, server_control_plane_timeout_in_milliseconds);
             zmq_socket.set(zmq::sockopt::linger, 0);
             zmq_socket.connect(zmq_server_target);
             log_server::trace("Connection established.");
```

korydraughn: Very clear. No one has to guess what is happening.

I think we have a file that defines global const[expr] default values. Perhaps the constexpr default can go there. Then it's available for other locations.

Sadly, I can't remember which file that is. Feel free to look for it, but if not, don't worry about it. What you've presented is certainly good enough.

alanking: Hmm... if you can give an example for a default value that you have in mind, I may be able to locate it. So far, all default values I've found seem to be like this one - that is, local to where they are being used - or scattered throughout various header files which may or may not be related to the respective values. :)

server/main_server/src/rodsServer.cpp
alanking: Yep, done

server/main_server/src/rodsServer.cpp
korydraughn: The thing I had in mind turned out to be specific to the delay server.

The delay server's hpp file contains constexpr variables. The function where I remember one of those being used was in a MSI that was recently removed.

All of that to say, I think you can move forward with the diff you presented.

alanking: Okay. I say we leave it as this diff, then. Resolving

alanking:
Unit and core tests passed

korydraughn (approved):
Please squash, pound, and merge when ready.

PR #6765, opened 2022-12-08
[2279] Remove unneeded sleep calls, leave imports for other functions (main)

MartinFlores751:
The `time.sleep(...)` calls are safe to remove, however, the `import time`s are still used by other functions, so imports will remain.

Comments on removing `import time` are removed to avoid confusion in the future.

### DIFF
--- a/scripts/irods/test/test_auth.py
+++ b/scripts/irods/test/test_auth.py
@@ -162,9 +162,7 @@ class Test_Auth(resource_suite.ResourceBase, unittest.TestCase):
                     }
 
                     with temporary_core_file() as core:
-                        time.sleep(1)  # remove once file hash fix is committed #2279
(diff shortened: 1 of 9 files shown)


korydraughn:
So, #2279 is still open.

Have you verified the relationship between that issue and the comments attached to the removed calls are no longer an issue?

Have you run this through the test suite?

MartinFlores751:
Ran the test suite and all is well.

The sleep lines removed seemed to be related to the issue (or at least had no effect on results), given that the original issue's workaround was to `sleep()` writes/copies to guarantee that the file change is visible.

trel (dismissed):
code changes look safe, and if the test suite passes, i think we're good to go.

korydraughn (approved):
Let's # it up

alanking (approved):

Issue #6198, opened 2022-02-22
Add support for admin mode to `rx_atomic_apply_acl_operations()`


korydraughn:
- [x] main
- [x] 4-3-stable
- [x] 4-2-stable
---
## Enhancement
This would allow admins to bypass permissions and modify ACLs on other users's collections and data objects.

PR #6779, opened 2022-12-15
[#6198] Added admin_mode option to rx_atomic_apply_acl_operations() (4-2-stable)

korydraughn:
Unit tests for this API plugin passed.

Nothing in the server uses this API plugin so it's safe to merge based on unit test results alone.

### DIFF
--- a/plugins/api/src/atomic_apply_acl_operations.cpp
+++ b/plugins/api/src/atomic_apply_acl_operations.cpp
@@ -496,9 +496,27 @@ namespace
 
         rodsLog(LOG_DEBUG, "Checking if user has permission to modify permissions ...");
 
-        if (!ic::user_has_permission_to_modify_acls(*_comm, db_conn, db_instance_name, object_id)) {
-            rodsLog(LOG_ERROR, "User not allowed to modify ACLs [logical_path=%s, object_id=%d]", logical_path.data(), object_id);
-            *_output = irods::to_bytes_buffer(make_error_object(json{}, 0, "User not allowed to modify ACLs").dump());
+        if (const auto iter = input.find("admin_mode"); iter != std::end(input) && iter->get<bool>()) {
+            if (!irods::is_privileged_client(*_comm)) {
+                const auto msg = fmt::format("User [{}#{}] not allowed to modify ACLs [logical_path={}, object_id={}]. "
+                                             "admin_mode cannot be enabled because user is not an administrator.",
+                                             _comm->clientUser.userName,
+                                             _comm->clientUser.rodsZone,
+                                             logical_path,
+                                             object_id);
+                rodsLog(LOG_ERROR, msg.c_str());
+                *_output = irods::to_bytes_buffer(make_error_object(json{}, 0, msg).dump());
+                return CAT_INSUFFICIENT_PRIVILEGE_LEVEL;
+            }
+        }
+        else if (!ic::user_has_permission_to_modify_acls(*_comm, db_conn, db_instance_name, object_id)) {
+            const auto msg = fmt::format("User [{}#{}] not allowed to modify ACLs [logical_path={}, object_id={}]",
+                                         _comm->clientUser.userName,
+                                         _comm->clientUser.rodsZone,
+                                         logical_path.c_str(),
+                                         object_id);
+            rodsLog(LOG_ERROR, msg.c_str());
+            *_output = irods::to_bytes_buffer(make_error_object(json{}, 0, msg.c_str()).dump());
             return CAT_NO_ACCESS_PERMISSION;
         }
 
--- a/unit_tests/src/test_atomic_apply_acl_operations.cpp
+++ b/unit_tests/src/test_atomic_apply_acl_operations.cpp
@@ -1,21 +1,27 @@
 #include "catch.hpp"
 
+#include "rcConnect.h"
 #include "rodsClient.h"
 #include "atomic_apply_acl_operations.h"
 
-#include "connection_pool.hpp"
+#include "client_connection.hpp"
 #include "filesystem.hpp"
+#include "rodsErrorTable.h"
 #include "user_administration.hpp"
 #include "irods_at_scope_exit.hpp"
+#include "transport/default_transport.hpp"
+#include "dstream.hpp"
 
 #include "json.hpp"
 
 #include <cstdlib>
 #include <string>
 #include <iostream>
+#include <utility>
 
 namespace fs = irods::experimental::filesystem;
 namespace ua = irods::experimental::administration;
+namespace io = irods::experimental::io;
 
 using json = nlohmann::json;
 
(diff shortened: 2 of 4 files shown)


plugins/api/src/atomic_apply_acl_operations.cpp
```diff
@@ -496,8 +496,15 @@ namespace
 
         rodsLog(LOG_DEBUG, "Checking if user has permission to modify permissions ...");
 
-        if (!ic::user_has_permission_to_modify_acls(*_comm, db_conn, db_instance_name, object_id)) {
-            rodsLog(LOG_ERROR, "User not allowed to modify ACLs [logical_path=%s, object_id=%d]", logical_path.data(), object_id);
+        if (const auto iter = input.find("admin_mode"); iter != std::end(input) && iter->get<bool>()) {
+            if (!irods::is_privileged_client(*_comm)) {
+                rodsLog(LOG_ERROR, "User is not an administrator.");
```
alanking, line 501: I see that this is the same message as atomic_apply_metadata, but I figured I would just throw this out there :p

Would this benefit from a little more context in the log message? Just thinking about seeing this message appear in the logs and wondering what it could mean.

Maybe something like...

admin_mode cannot be enabled in rs_atomic_apply_acl_operations because the user is not an administrator.

unit_tests/src/test_atomic_apply_acl_operations.cpp
```diff
@@ -174,6 +175,143 @@ TEST_CASE("atomic_apply_acl_operations")
...
+    //
+
+    using namespace std::string_literals;
+
+    irods::experimental::client_connection conn;
+
+    // Create a test rodsuser.
+    ua::user test_user{"unit_test_test_user"};
+    irods::at_scope_exit remove_user{[&conn, &test_user] { ua::client::remove_user(conn, test_user); }};
+    CHECK(ua::client::add_user(conn, test_user).value() == 0);
```
alanking, line 191: Should this be `REQUIRE` since it's setting up the real test?

unit_tests/src/test_atomic_apply_acl_operations.cpp
```diff
@@ -174,6 +175,143 @@ TEST_CASE("atomic_apply_acl_operations")
...
+
+    // Create a test rodsgroup.
+    ua::group test_group{"unit_test_test_group"};
+    irods::at_scope_exit remove_group{[&conn, &test_group] { ua::client::remove_group(conn, test_group); }};
+    CHECK(ua::client::add_group(conn, test_group).value() == 0);
```
alanking, line 196: Same. Should this be `REQUIRE` since it's part of setting up the real test?

alanking (commented):
Couple of minor things

unit_tests/src/test_atomic_apply_acl_operations.cpp
korydraughn: It can be. Will change.

unit_tests/src/test_atomic_apply_acl_operations.cpp
korydraughn: It can be. Will change.

plugins/api/src/atomic_apply_acl_operations.cpp
korydraughn: Yes. Will add more words.

korydraughn: Done.

unit_tests/src/test_atomic_apply_acl_operations.cpp
korydraughn: Done.

unit_tests/src/test_atomic_apply_acl_operations.cpp
korydraughn: Done.

korydraughn:
All unit tests passed.

alanking (dismissed):
I had gotten confused at first because I noticed all the interfaces were different from what I had in my local clone but then I saw that this is 4-2-stable. I think this is good to go

korydraughn:
I added one more unit test that makes sure non-admin users can use the API plugin as long as they have appropriate permissions on the target object.

That will at least allow us to detect regressions down the road.

All unit tests pass.

plugins/api/src/atomic_apply_acl_operations.cpp
```diff
@@ -496,8 +496,19 @@ namespace
 
         rodsLog(LOG_DEBUG, "Checking if user has permission to modify permissions ...");
 
-        if (!ic::user_has_permission_to_modify_acls(*_comm, db_conn, db_instance_name, object_id)) {
-            rodsLog(LOG_ERROR, "User not allowed to modify ACLs [logical_path=%s, object_id=%d]", logical_path.data(), object_id);
+        if (const auto iter = input.find("admin_mode"); iter != std::end(input) && iter->get<bool>()) {
+            if (!irods::is_privileged_client(*_comm)) {
+                const auto msg = fmt::format("User not allowed to modify ACLs [logical_path={}, object_id={}]. "
```
trel, line 501: consider...

"User not" -> "User [{}] not"

and fill it with full user#zone

plugins/api/src/atomic_apply_acl_operations.cpp
```diff
@@ -496,8 +496,19 @@ namespace
...
+                                             "admin_mode cannot be enabled because user is not an administrator.",
+                                             logical_path,
+                                             object_id);
+                rodsLog(LOG_ERROR, msg.c_str());
+                *_output = irods::to_bytes_buffer(make_error_object(json{}, 0, msg).dump());
+                return CAT_INSUFFICIENT_PRIVILEGE_LEVEL;
+            }
+        }
+        else if (!ic::user_has_permission_to_modify_acls(*_comm, db_conn, db_instance_name, object_id)) {
+            rodsLog(LOG_ERROR, "User not allowed to modify ACLs [logical_path=%s, object_id=%d]", logical_path.c_str(), object_id);
```
trel, line 511: same

plugins/api/src/atomic_apply_acl_operations.cpp
korydraughn: Yeah, I couldn't decide on whether that should be included because of the log message being referenced in the review comment below.

Will change. It's better to provide more information in general.

korydraughn: Done

plugins/api/src/atomic_apply_acl_operations.cpp
korydraughn: Done

korydraughn:
All unit tests pass.

Any final thoughts before I add the pound?

trel (dismissed):

korydraughn:
Added #

alanking (approved):

PR #6787, opened 2022-12-19
[5411] Test for `iinit`

MartinFlores751:
This is a test to go along with the corresponding iinit change in icommands.

### DIFF
--- a/lib/core/src/obf.cpp
+++ b/lib/core/src/obf.cpp
@@ -78,6 +78,8 @@ int cipherBlockChaining = 0;
    HASH_TYPE_MD5 and HASH_TYPE_SHA1 */
 static int defaultHashType = HASH_TYPE_MD5;
 
+constexpr auto max_path_length{MAX_NAME_LEN + 10};
+
 /*
   What can be a main routine for some simple tests.
  */
@@ -117,40 +119,30 @@ obfSetDebug( int opt ) {
 
 int
 obfiGetFilename( char *fileName ) {
-    char *envVar = NULL;
+    char* envVar{getRodsEnvAuthFileName()};
 
-    envVar = getRodsEnvAuthFileName();
-    if ( envVar != NULL && *envVar != '\0' ) {
-        strcpy( fileName, envVar );
+    if (envVar != nullptr && *envVar != '\0') {
+        std::strncpy(fileName, envVar, max_path_length);
         return 0;
     }
 
-#ifdef windows_platform
-    if ( ProcessType != CLIENT_PT ) {
-        char *tmpstr1;
-        int t;
-        tmpstr1 = iRODSNtGetServerConfigPath();
-        sprintf( fileName, "%s\\irodsA.txt", tmpstr1 );
-        return 0;
-    }
-
-    /* for clients */
-    envVar = iRODSNt_gethome();
-#else
-    envVar =  getenv( "HOME" );
-#endif
-    if ( envVar == NULL ) {
+    envVar = std::getenv("HOME");
+    if (envVar == nullptr) {
         return ENVIRONMENT_VAR_HOME_NOT_DEFINED;
     }
-    strncpy( fileName, envVar, MAX_NAME_LEN );
-    strncat( fileName, "/", MAX_NAME_LEN );
-    strncat( fileName, AUTH_FILENAME_DEFAULT, MAX_NAME_LEN );
 
-#ifdef windows_platform
-    if ( envVar != NULL ) {
-        free( envVar );
+    const auto env_length_with_null_term{std::strlen(envVar) + 1};
+    constexpr auto auth_file_length{std::char_traits<char>::length(AUTH_FILENAME_DEFAULT)};
+    constexpr auto path_separator_length{1};
+
+    const auto required_size{env_length_with_null_term + auth_file_length + path_separator_length};
+    if (required_size > max_path_length) {
+        return USER_PATH_EXCEEDS_MAX;
     }
-#endif
+
+    std::strncpy(fileName, envVar, env_length_with_null_term);
+    std::strncat(fileName, "/", path_separator_length);
+    std::strncat(fileName, AUTH_FILENAME_DEFAULT, auth_file_length);
 
     return 0;
 }
@@ -159,7 +151,7 @@ int
 obfGetPw( char *pw ) {
     char myPw[MAX_PASSWORD_LEN + 10];
     char myPwD[MAX_PASSWORD_LEN + 10];
-    char fileName[MAX_NAME_LEN + 10];
+    char fileName[max_path_length];
     char *cp;
 
     int i;
@@ -216,7 +208,7 @@ obfGetPw( char *pw ) {
 */
 int
 obfRmPw( int opt ) {
-    char fileName[MAX_NAME_LEN + 10];
+    char fileName[max_path_length];
     char inbuf[MAX_NAME_LEN + 10];
 
     if ( int status = obfiGetFilename( fileName ) ) {
@@ -286,7 +278,7 @@ obfiSetTimeFromFile( int fd ) {
 int
 obfSavePw( int promptOpt, int fileOpt, int printOpt, const char *pwArg ) {
     using namespace boost::filesystem;
-    char fileName[MAX_NAME_LEN + 10];
+    char fileName[max_path_length];
     char inbuf[MAX_PASSWORD_LEN + 100];
     char myPw[MAX_PASSWORD_LEN + 10];
     int i = 0, fd = 0, envVal = 0;
@@ -383,7 +375,7 @@ obfSavePw( int promptOpt, int fileOpt, int printOpt, const char *pwArg ) {
 
 /* various options for temporary auth files/pws */
 int obfTempOps( int tmpOpt ) {
-    char fileName[MAX_NAME_LEN + 10];
+    char fileName[max_path_length];
     char pw[MAX_PASSWORD_LEN + 10];
     int i;
     if ( tmpOpt == 1 ) { /* store pw flagged as temporary */
--- a/scripts/irods/test/test_misc.py
+++ b/scripts/irods/test/test_misc.py
@@ -1,6 +1,7 @@
 import os
 import sys
 from datetime import datetime, timedelta
+import tempfile
 
 if sys.version_info < (2, 7):
     import unittest2 as unittest
@@ -331,3 +332,39 @@ class Test_Misc(session.make_sessions_mixin([('otherrods', 'rods')], []), unitte
     @unittest.skipIf(test.settings.RUN_IN_TOPOLOGY, "skip for topology testing")
     def test_scoped_client_identity_updates_zone_of_RsComm__issue_6268(self):
         self.admin.assert_icommand(['irule', '-r', 'irods_rule_engine_plugin-irods_rule_language-instance', 'msi_test_scoped_client_identity', 'null', 'ruleExecOut'])
+
+    def test_iinit_does_not_crash_from_long_home_directory__issue_5411(self):
+        with tempfile.TemporaryDirectory() as temp_dir_name:
+            # Create new home directory for test
+            temp_home = os.path.join(temp_dir_name, 'crazy', 'long', 'temp', 'home', 'directory',
+                                     'like', 'this', 'is', 'longer', 'than', 'the', 'solos', 'in', 'chameleon')
+            os.makedirs(temp_home)
+
+            # Setup environment
+            temp_env = os.environ.copy()
+            temp_env['HOME'] = temp_home
+
+            # Setup input for execution
+            user_input = os.linesep.join(
+                ['localhost', '1247', 'rods', 'tempZone', 'rods']) + os.linesep
+            command = 'iinit'
+
+            # Execute iinit
+            # Avoid using *_icommand(...), as they set the auth file location to a separate, shorter directory!
+            stdout, stderr, rc = lib.execute_command_permissive(command, input=user_input, env=temp_env)
+            lib.log_command_result(command, stdout, stderr, rc)
+
+            # Ensure program terminated gracefully
+            self.assertEqual(rc, 0)
+
+            # Check all input prompted
+            self.assertIn('Enter the host name (DNS) of the server to connect to:', stdout)
+            self.assertIn('Enter the port number:', stdout)
+            self.assertIn('Enter your irods user name:', stdout)
+            self.assertIn('Enter your irods zone:', stdout)
+            self.assertIn('Enter your current iRODS password:', stdout)
+
+            # Check expected error exists
+            error_string = ' ERROR: environment_properties::capture: missing environment file. should be at [{}/.irods/irods_environment.json]\n'.format(
+                temp_home)
+            self.assertEqual(stderr, error_string)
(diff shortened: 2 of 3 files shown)


scripts/irods/test/test_misc.py
```diff
@@ -317,3 +319,29 @@ def test_server_correctly_cleans_up_proc_files__issue_6231(self):
...
+                                     'like', 'this', 'is', 'longer', 'than', 'the', 'solos', 'in', 'chameleon')
+            os.makedirs(temp_home)
+
+            # Setup environment
+            temp_env = os.environ.copy()
+            temp_env['HOME'] = temp_home
+
+            # Execute iinit
+            user_input = os.linesep.join(
+                ['localhost', '1247', 'rods', 'tempZone', 'rods']) + os.linesep
```
korydraughn, line 335: That's pretty neat.

scripts/irods/test/test_misc.py
```diff
@@ -317,3 +319,29 @@ def test_server_correctly_cleans_up_proc_files__issue_6231(self):
...
+            res = subprocess.run('iinit', input=user_input.encode(
+            ), stdout=subprocess.PIPE, stderr=subprocess.PIPE, env=temp_env)
```
korydraughn, line 337: Let's make this one line. I'm guessing your editor's formatter caused the split at `.encode(`?

Why `subprocess.run()` instead of one of the `lib.execute*` functions?

scripts/irods/test/test_misc.py
```diff
@@ -317,3 +319,29 @@ def test_server_correctly_cleans_up_proc_files__issue_6231(self):
...
+
+            self.assertEqual(res.returncode, 0)
+
+            # Implicitly check we ran through our input
+            self.assertIn(b'Enter your current iRODS password:\n', res.stdout)
```
korydraughn, line 342: Consider checking for all prompts?

L334-L335 includes input for multiple prompts.

scripts/irods/test/test_misc.py
```diff
@@ -317,3 +319,29 @@ def test_server_correctly_cleans_up_proc_files__issue_6231(self):
...
+            temp_env = os.environ.copy()
+            temp_env['HOME'] = temp_home
+
+            # Execute iinit
+            user_input = os.linesep.join(
+                ['localhost', '1247', 'rods', 'tempZone', 'rods']) + os.linesep
+            res = subprocess.run('iinit', input=user_input.encode(
+            ), stdout=subprocess.PIPE, stderr=subprocess.PIPE, env=temp_env)
+
+            self.assertEqual(res.returncode, 0)
```
korydraughn, line 339: Should 0 be returned if the size of the home collection is too large?

scripts/irods/test/test_misc.py
```diff
@@ -317,3 +319,29 @@ def test_server_correctly_cleans_up_proc_files__issue_6231(self):
...
+
+            # Implicitly check we ran through our input
+            self.assertIn(b'Enter your current iRODS password:\n', res.stdout)
+
+            # Check that only expected error exists
```
korydraughn, line 344: "only expected" -> "only the expected"?

korydraughn (commented):
Nice. I didn't know about `os.linesep.join`.

scripts/irods/test/test_misc.py
MartinFlores751: Yep. When I wrote that code originally, it was in one line. I formatted the code before committing, so what you see is what the formatter chose.

From my understanding, lib.execute* functions end up calling `execute_command_nonblocking(...)` which masks the `env` parameter that is needed for this test. So, I went with the `run()` function instead.

scripts/irods/test/test_misc.py
MartinFlores751: I think we would get an error in `obf.cpp` if we put in a very long home directory of ~1000+ chars.

In this case, the string was no more than 200 chars, very likely less, so we expect `iinit` to work fine.

scripts/irods/test/test_misc.py
korydraughn: Are there any downsides to using `.run()`?

scripts/irods/test/test_misc.py
korydraughn: Did the test fail with 200 characters before?
What happens if the path exceeds 1k characters?

scripts/irods/test/test_misc.py
MartinFlores751: In this case, I don't believe so. In general, it's a bit less flexible than options such as `Popen`.

The [Python docs](https://docs.python.org/3.6/library/subprocess.html) recommend preferring the `.run()` function compared to other `subprocess` options:
> The recommended approach to invoking subprocesses is to use the `run()` function for all use cases it can handle. For more advanced use cases, the underlying `Popen` interface can be used directly.

scripts/irods/test/test_misc.py
MartinFlores751: Originally, the program failed on an input of 79 chars.

I'll throw some longer inputs at it to see what happens. 👍

scripts/irods/test/test_misc.py
MartinFlores751: “Check expected error exists”?

korydraughn: Go for it.

scripts/irods/test/test_misc.py
MartinFlores751: Running with a HOME greater than 1098 characters, including `/`s, leads to a return of 1 from main.
The following error is also printed: `ERROR: Save Password failure status = -900000 FILE_OPEN_ERR`.

korydraughn: That error message doesn't look very helpful.

Do we know what caused the failure? If so, we should be more explicit if possible.

alanking: My guess would be `obfSavePw` failing to write the `.irodsA` file? But that would be strange since I would have expected `iinit` to fail before reaching that point. Maybe the directory is created successfully, but adding `.irodsA` to the end to save the obfuscated password fails because only then is the path too long.

scripts/irods/test/test_misc.py
```diff
@@ -317,3 +319,32 @@ def test_server_correctly_cleans_up_proc_files__issue_6231(self):
         # Show that the proc directory will eventually become empty.
         lib.delayAssert(lambda: len(os.listdir(paths.proc_directory())) == 0)
 
+    def test_iinit_does_not_crash_from_long_home_directory__issue_5411(self):
```
MartinFlores751, line 322: Test currently does not provide enough information upon encountering an error.

Log errors like `assert_icommand` does, and if possible, don't quit upon the first failed assertion, so more information can be gathered.

MartinFlores751 (commented):
Some things to work on.

lib/core/src/obf.cpp
```diff
@@ -117,40 +119,33 @@ obfSetDebug( int opt ) {
...
-    strncat( fileName, "/", MAX_NAME_LEN );
-    strncat( fileName, AUTH_FILENAME_DEFAULT, MAX_NAME_LEN );
 
-#ifdef windows_platform
-    if ( envVar != NULL ) {
-        free( envVar );
+    using cstr = std::char_traits<char>;
+
+    const auto env_length_with_null_term{std::strlen(envVar) + 1};
+    constexpr auto auth_file_length{cstr::length(AUTH_FILENAME_DEFAULT)};
```
alanking, line 138: Is this preferred over `std::strlen`?

MartinFlores751: I think it's worth considering, at least for some cases. `std::char_traits<char>::length(...)` is constexpr, unlike `std::strlen(...)`, so we could do some more work at compile time.

scripts/irods/test/test_misc.py
MartinFlores751: Updated so when the path to `.irodsA` is too long, `USER_INPUT_PATH_ERR` is returned. I'm curious if there's a more appropriate error status available for this case?

korydraughn: What would you suggest the name be for a new error code?
What would the generic name be (so we can use it in other situations)?

lib/core/src/obf.cpp
```diff
@@ -117,40 +119,33 @@ obfSetDebug( int opt ) {
 
 int
 obfiGetFilename( char *fileName ) {
-    char *envVar = NULL;
+    char* envVar{nullptr};
```
korydraughn, line 122: This can be combined with L124.

lib/core/src/obf.cpp
```diff
@@ -117,40 +119,33 @@ obfSetDebug( int opt ) {
...
-    if ( envVar != NULL ) {
-        free( envVar );
+    using cstr = std::char_traits<char>;
+
+    const auto env_length_with_null_term{std::strlen(envVar) + 1};
+    constexpr auto auth_file_length{cstr::length(AUTH_FILENAME_DEFAULT)};
+    constexpr auto path_separator_length{1};
+
+    const auto required_size{env_length_with_null_term + auth_file_length + path_separator_length};
+    if (required_size > max_path_length) {
```
korydraughn, line 142: Is `>` correct?
Does `max_path_length` consider the null byte?

lib/core/src/obf.cpp
```diff
@@ -78,6 +78,8 @@ int cipherBlockChaining = 0;
    HASH_TYPE_MD5 and HASH_TYPE_SHA1 */
 static int defaultHashType = HASH_TYPE_MD5;
 
+constexpr auto max_path_length{MAX_NAME_LEN + 10};
```
korydraughn, line 81: Just curious ...

Any ideas as to why 10 is added?
Can 10 be replaced with a different value?

lib/core/src/obf.cpp
MartinFlores751: `max_path_length` should be the size of the entire buffer, so any size greater should not be able to fit into the existing buffer.

The additional null term for `required_size` is included in `env_length_with_null_term`.

lib/core/src/obf.cpp
MartinFlores751: Not sure why there is a +10… I just know that it shows up a lot. It's part of the reason I decided to give it a name here, at the very least.

trel: i really don't want this to be the answer... but '@tempZone' happens to be 10...

trel: well, okay, not 10, but ... yeah, i don't know.

korydraughn: If we don't know, that's fine. It has an appropriate name now which is better than nothing.

Carry on.

lib/core/src/obf.cpp
korydraughn: Seems fine.

I'd recommend changing the name of the type alias from `cstr` to something that indicates it is a traits type (e.g. `char_traits`) or using `std::char_traits` in L138 directly.

scripts/irods/test/test_misc.py
```diff
@@ -317,3 +319,37 @@ def test_server_correctly_cleans_up_proc_files__issue_6231(self):
...
+            temp_env['HOME'] = temp_home
+
+            # Setup input for execution
+            user_input = os.linesep.join(
+                ['localhost', '1247', 'rods', 'tempZone', 'rods']) + os.linesep
+            command = 'iinit'
+
+            # Execute iinit
+            res = subprocess.run(command, input=user_input.encode(), stdout=subprocess.PIPE, stderr=subprocess.PIPE, env=temp_env)
+            lib.log_command_result(command, res.stdout.decode(), res.stderr.decode(), res.returncode)
```
alanking, line 340: I think this warrants an explanatory comment in the code so we understand why the "usual" utilities are not used here.

scripts/irods/test/test_misc.py
alanking: I found `USER_PATH_EXCEEDS_MAX`, if that seems better.

alanking (commented):
Closer...

scripts/irods/test/test_misc.py
MartinFlores751: Just going to change `ERROR: Save Password failure` to `ERROR: Failed to save password!` in the other PR. Maybe the `!` will help separate the `status` a bit better, too.

lib/core/src/obf.cpp
```diff
@@ -117,40 +119,30 @@ obfSetDebug( int opt ) {
...
-
-    envVar = getRodsEnvAuthFileName();
-    if ( envVar != NULL && *envVar != '\0' ) {
-        strcpy( fileName, envVar );
-        return 0;
-    }
+    char* envVar{getRodsEnvAuthFileName()};
 
-#ifdef windows_platform
-    if ( ProcessType != CLIENT_PT ) {
```
korydraughn, line 129: Was this line removed on purpose? Is it needed?

MartinFlores751: It was intentional, thought we wanted to remove the windows code?
It shouldn't be needed to my knowledge, but maybe there's a reason otherwise?

alanking (commented):
Looks pretty good to me. How's everybody feel about squashing?

lib/core/src/obf.cpp
korydraughn: I see now. The PP else case doesn't check the `ProcessType`.

Carry on.

lib/core/src/obf.cpp
```diff
@@ -117,40 +119,30 @@ obfSetDebug( int opt ) {
...
-        char *tmpstr1;
-        int t;
-        tmpstr1 = iRODSNtGetServerConfigPath();
-        sprintf( fileName, "%s\\irodsA.txt", tmpstr1 );
+    if (envVar != nullptr && *envVar != '\0') {
+        std::strcpy(fileName, envVar);
```
korydraughn, line 125: Can this line result in a buffer overflow?

MartinFlores751: Seems like it. `std::strlen(...)` in the if statement?

korydraughn: I think that would improve it.

The alternative is to use something like `is_non_empty_string`. That function expects a string and a max length. It calls `strnlen` under the covers.

korydraughn: Or `is_empty_string` instead.

MartinFlores751: Looked in more detail, `getRodsEnvAuthFileName()` should return a buffer with a max size of 256. `filename` has a max size of 1098, so we should be fine as is. However, I'll change this to a `std::strncpy(...)`, just in case.

alanking (commented):
Please squash and flesh out commit message

trel:
squashed

alanking (approved):
Looks good. Let's add #s

alanking:
Gah. Sorry, I didn't realize #s were added and have merged things since, causing conflicts. @MartinFlores751 - Please rebase and resolve conflicts.

I don't get a notification (most of the time?) when people add #s to commits in PRs, so please add a comment on the PR or ping me next time and I'll be able to respond in a more timely fashion. :) Again, sorry about that.

MartinFlores751:
@alanking should be good now!

alanking (approved):
Thanks

PR #6790, opened 2022-12-20
Enabled full access to Filesystem lib and dstream lib (main)

korydraughn:
This PR adjusts include guards to allow the filesystem headers and dstream headers to be included twice (once for client-side and once for server-side).

If there's a better way to do this, please let me know.

I still need to come up with a way to test this. I'm open to ideas.

At the moment, I'm thinking I can create a small client that invokes a test API plugin. The test API plugin would invoke various client-side and server-side APIs within the server.

If only we had a server-side testing framework.


lib/filesystem/README.md
```diff
@@ -5,21 +5,7 @@ It can be used on both the client and server side.
...
-
 ## Using the Library
-To use the library, simply include the filesystem header (e.g. `<irods/filesystem.hpp>`) and link against the appropriate shared library. Client-side software should link against `irods_client.so`. Server-side software should link against `irods_server.so`.
-
-The library defaults to providing the client-side API. To enable the server-side API, define the following macro.
-- IRODS_FILESYSTEM_ENABLE_SERVER_SIDE_API
-
-## TODOs
-- Finish designing and implementing unit tests.
+For usage examples, see the unit tests directory of their repository.
```
alanking, line 11: Is "their" referring to the C++17 filesystem library?

lib/core/include/irods/transport/default_transport.hpp
```diff
@@ -1,5 +1,16 @@
-#ifndef IRODS_IO_DEFAULT_TRANSPORT_HPP
-#define IRODS_IO_DEFAULT_TRANSPORT_HPP
+#undef IRODS_IO_DEFAULT_TRANSPORT_HPP_INCLUDE_HEADER
```
alanking, line 1: Are `_INCLUDE_HEADER` suffixes meant to be how we want to do things moving forward, or is this a special case?

lib/core/include/irods/transport/default_transport.hpp
```diff
@@ -1,5 +1,16 @@
...
+#  if !defined(IRODS_IO_DEFAULT_TRANSPORT_HPP_FOR_SERVER)
+#    define IRODS_IO_DEFAULT_TRANSPORT_HPP_FOR_SERVER
+#    define IRODS_IO_DEFAULT_TRANSPORT_HPP_INCLUDE_HEADER
+#  endif
+#elif !defined(IRODS_IO_DEFAULT_TRANSPORT_HPP_FOR_CLIENT)
+#  define IRODS_IO_DEFAULT_TRANSPORT_HPP_FOR_CLIENT
+#  define IRODS_IO_DEFAULT_TRANSPORT_HPP_INCLUDE_HEADER
+#endif
+
+#ifdef IRODS_IO_DEFAULT_TRANSPORT_HPP_INCLUDE_HEADER
```
alanking, line 13: So, in order to include both would be like this...?
```c++
#include <irods/transport/default_transport.hpp> // client-side

#undef IRODS_IO_DEFAULT_TRANSPORT_HPP_INCLUDE_HEADER

#define IRODS_IO_TRANSPORT_ENABLE_SERVER_SIDE_API
#include <irods/transport/default_transport.hpp> // server-side
```
But then... everything gets `#undef`'d on L15-21... and redefined. So, I guess that wouldn't work.

How do you do it? :p

alanking: D'oh, it gets `undef`'d in this file. So I guess you would need `IRODS_IO_DEFAULT_TRANSPORT_HPP_FOR_SERVER` in there somewhere.

lib/filesystem/README.md
korydraughn: That is supposed to say `this` instead of `their`.

Will fix.

lib/core/include/irods/transport/default_transport.hpp
korydraughn: This PR is a special case.

Because it uses macros such as `rxComm`, I had to figure out a way to allow the same headers to be included multiple times while maintaining the same behavior as the include guards.

The preprocessor logic mirrors the behavior of normal include guards while also allowing the contents of the headers to change based on whether the special `_ENABLE_SERVER_API` macro is defined or not.

lib/core/include/irods/transport/default_transport.hpp
korydraughn: For the default transport:
```cpp
#include <irods/transport/default_transport.hpp>

#define IRODS_IO_TRANSPORT_ENABLE_SERVER_SIDE_API
#include <irods/transport/default_transport.hpp>
```
For filesystem:
```cpp
#include <irods/filesystem.hpp>

#define IRODS_FILESYSTEM_ENABLE_SERVER_SIDE_API
#include <irods/filesystem.hpp>
```

korydraughn:
All unit tests pass.

Running core tests.

korydraughn:
I think a simpler approach to testing things like this PR is through microservices.

The major advantages to a microservice being they can be triggered by the python tests and the client already exists (i.e. `irule`).

They would require a new build option so that we don't ship them as part of a release package. The build option would default to OFF.

lib/filesystem/README.md
korydraughn: Done.

korydraughn:
I've implemented a very simple API for writing server-side C/C++ tests. It provides the following:
- Macros for signaling the start of and end of a test
- A single macro for making assertions
- All assertions regardless of the outcome are written to syslog
- The first assertion that fails will end the test
- The python tests which invoke the microservices succeed/fail based on the outcome of the microservice
- The microservices are not compiled/packaged by default. Developers must opt into compiling them

We now need to agree on naming and the development environment needs to be updated to support the new build options.

Preparing to run unit tests and core tests.

korydraughn:

Awaiting approvals.

trel (dismissed):

korydraughn:
#'d

alanking (approved):

PR #6792, opened 2022-12-20
L1 descriptor fixes (main)

korydraughn:
Taken from #6673.

Will run tests soon.

### DIFF
--- a/server/core/src/objDesc.cpp
+++ b/server/core/src/objDesc.cpp
@@ -1,6 +1,7 @@
+#include "irods/objDesc.hpp"
+
 #include "irods/rcMisc.h"
 #include "irods/rodsDef.h"
-#include "irods/objDesc.hpp"
 #include "irods/dataObjOpr.hpp"
 #include "irods/rodsDef.h"
 #include "irods/rsGlobalExtern.hpp"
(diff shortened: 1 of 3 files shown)


korydraughn:
All unit tests passed.

Running core tests ...

alanking (commented):
Seems good

How is this interacting with #6673? Should we include the changes from #6793 here, too?

korydraughn:
I'm going to handle the deprecation in a separate PR since I've already tested this.

I expect there will be conflicts with the ASan PR, but those will be simple to resolve.

alanking (dismissed):
In that case, looks good to me. Shall we await other requested reviewers?

korydraughn:
Let's give it an hour. If no one shows up, I'll add the pounds.

server/core/src/objDesc.cpp
```diff
@@ -28,10 +29,47 @@
...
+    _l1d.dataObjInfo = nullptr;
+    _l1d.otherDataObjInfo = nullptr;
+    _l1d.replDataObjInfo = nullptr;
+    _l1d.remoteZoneHost = nullptr;
+
+    _l1d.pluginData.clear();
+    _l1d.replica_token.clear();
+
+    // NOLINTBEGIN(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+    std::memset(_l1d.chksum, 0, sizeof(l1desc::chksum));
```
MartinFlores751, line 68: Might be worth considering `std::fill` vs `std::memset`.

MartinFlores751 (dismissed):
Everything looks great! Just one small thing to think about is all I have to comment.

server/core/src/objDesc.cpp
korydraughn: I could use `std::fill`, but `std::memset` seems perfectly fine for this case given `l1desc::chksum` is a plain ole C array.

Is there an argument against using `std::memset`?

korydraughn: After some discussion, I'm going to move forward with `std::memset`.

There are online articles that hint at the possibility that `std::fill` can produce the same assembly as `std::memset`, but not always.

We should investigate in which cases it is better to use one function over the other.

korydraughn:
#'d

alanking (approved):

Issue #5913, opened 2021-10-13
write ticket for collection does not allow upload of new data objects


d-w-moore:
- [x] main
- [x] 4-3-stable
- [x] 4-2-stable
---
### iRODS Version, OS and Version

iRODS 4.2.10, Ubuntu 18

### What did you try to do?

As user alice, create a new object with `iput`, with a write ticket to the parent collection enabled.

### Expected behavior
Assuming that write tickets are intended to provide a dropbox-style upload functionality, then - giving user `alice` a write ticket to a collection - it is expected then that she would be able to use that ticket to create and write to (in effect, upload) a new data object in that collection.

### Observed behavior (including steps to reproduce, if applicable)
as rods (rodsadmin):
```
irods:~$ imkdir /tempZone/home/rods/test_write
irods:~$ iticket create write /tempZone/home/rods/test_write
ticket:6pBAw34U5g6HUFx
```

as alice (rodsuser):
```
~$ echo 'from: alice' > ALICE_abc.dat
~$ iput -t 6pBAw34U5g6HUFx ALICE_abc.dat /tempZone/home/rods/test_write/abc.dat
remote addresses: 127.0.0.1 ERROR: putUtil: put error for /tempZone/home/rods/test_write/abc.dat, status = -808000 status = -808000  CAT_NO_ROWS_FOUND
```

This is produced in the rodsLog:
```
Oct 13 13:48:31 pid:11541 NOTICE: session ticket setup, value: 6pBAw34U5g6HUFx
Oct 13 13:48:31 pid:11541 remote addresses: 127.0.0.1 ERROR: failed to determine collection name for object id [10096]
Oct 13 13:48:31 pid:11541 remote addresses: 127.0.0.1 ERROR: [-]        /irods/server/api/src/rsRegDataObj.cpp:85:int _rsRegDataObj(rsComm_t *, dataObjInfo_t *) :  status [CAT_NO_ROWS_FOUND]  errno [] -- message [_rsRegDataObj - Failed to register data object "/tempZone/home/rods/test_write/abc.dat" - CAT_NO_ROWS_FOUND ]

Oct 13 13:48:31 pid:11541 remote addresses: 127.0.0.1 ERROR: filePathReg: svrRegDataObj for /tempZone/home/rods/test_write/abc.dat failed, status = -808000
[...]

PR #6796, opened 2022-12-23
[#5913] Restored ability to create data objects using write tickets (main)

korydraughn:
Running unit tests and core tests.

### DIFF
--- a/plugins/database/src/mid_level_routines.cpp
+++ b/plugins/database/src/mid_level_routines.cpp
@@ -1380,10 +1380,17 @@ int checkObjIdByTicket( const char *dataId, const char *accessLevel,
 
     char original_collection_name[MAX_NAME_LEN];
     std::vector<std::string> bindVars;
-    bindVars.push_back( dataId );
+    // "dataId" will match the ID of a collection or data object.
+    // See the prepared statement to understand why "dataId" is passed twice.
+    bindVars.emplace_back(dataId);
+    bindVars.emplace_back(dataId);
     int cml_error = cmlGetStringValueFromSql(
-                 "select coll_name from R_COLL_MAIN where coll_id in (select coll_id from R_DATA_MAIN where data_id=?)",
-                 original_collection_name, MAX_NAME_LEN, bindVars, icss );
+        "select coll_name from R_COLL_MAIN "
+        "where coll_id = ? or coll_id in (select coll_id from R_DATA_MAIN where data_id = ?)",
+        original_collection_name, // NOLINT(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+        MAX_NAME_LEN,
+        bindVars,
+        icss);
     if(0 != cml_error) {
         rodsLog(
             LOG_ERROR,
--- a/scripts/irods/test/test_iticket.py
+++ b/scripts/irods/test/test_iticket.py
@@ -728,3 +728,41 @@ class Test_Iticket(SessionsMixin, unittest.TestCase):
         self.user.assert_icommand(['iticket', '-M', 'mod', ticket_string, 'remove', 'group', 'public'], 'STDERR', ['CAT_INSUFFICIENT_PRIVILEGE_LEVEL'])
         self.user.assert_icommand(['iticket', 'ls', ticket_string], 'STDOUT', ['No group restrictions'])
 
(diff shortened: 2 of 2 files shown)


korydraughn:
All unit tests passed.

Running core tests.

scripts/irods/test/test_iticket.py
```diff
@@ -728,3 +728,36 @@ def test_iticket_does_not_allow_rodsusers_to_use_admin_option(self):
...
+            self.assertEqual(out.strip(), self.admin.session_collection + ':')
+            self.assertEqual(len(err), 0)
+
+            # Show that the rodsuser does not have any permissions on the session collection.
+            out, err, ec = self.admin.run_icommand(['ils', '-A'])
+            self.assertEqual(ec, 0);
+            self.assertNotIn(f'{self.user.username}#{self.user.zone_name}', out)
+            self.assertEqual(len(err), 0)
+
+            # Create a new data object as
```
trel, line 751: as ... rodsuser?

plugins/database/src/mid_level_routines.cpp
```diff
@@ -1381,9 +1381,14 @@ int checkObjIdByTicket( const char *dataId, const char *accessLevel,
     char original_collection_name[MAX_NAME_LEN];
     std::vector<std::string> bindVars;
     bindVars.push_back( dataId );
+    bindVars.push_back(dataId);
```
trel, line 1384: spacing is different.    and we want to push `dataId` twice?   I guess once as `coll_id`...

scripts/irods/test/test_iticket.py
korydraughn: Will fix.

plugins/database/src/mid_level_routines.cpp
korydraughn: > spacing is different.

Yeah, I kept changing my mind on that. I figured it was better to show the actual fix rather than make everything line up, but the change is tiny so I think cleaning up both lines is fine too.

Will update.

> and we want to push `dataId` twice? I guess once as `coll_id`...

Correct. The query is attempting to find a collection name. The problem was that it was only considering collections that contained data objects. The fix was to also consider empty collections.

korydraughn:
All core tests passed.

Resolving PR comments.

scripts/irods/test/test_iticket.py
korydraughn: Done

plugins/database/src/mid_level_routines.cpp
korydraughn: Done

korydraughn:
I adjusted the test so that it does a single buffer put and a parallel transfer.

korydraughn:
I've verified that the test passes at the bench.

Any last comments before I add the pound?

plugins/database/src/mid_level_routines.cpp
```diff
@@ -1380,10 +1380,15 @@ int checkObjIdByTicket( const char *dataId, const char *accessLevel,
 
     char original_collection_name[MAX_NAME_LEN];
     std::vector<std::string> bindVars;
-    bindVars.push_back( dataId );
+    bindVars.emplace_back(dataId);
+    bindVars.emplace_back(dataId);
```
trel, line 1384: please add a little comment explaining the dataId is intentionally/correctly used as a collId - so it doesn't look like a mistake or require additional scrutiny in the future.

trel (dismissed):
one thing, and then the #s.

plugins/database/src/mid_level_routines.cpp
korydraughn: I can rename the variable to `object_id` assuming that is what you're getting at.

trel: no, a comment is plenty fine, no need to change any names/logic.

korydraughn: Updated. Please check the comment to make sure it is acceptable.

trel: sold.  squash and # it.

korydraughn: Done.

korydraughn:
#'d

alanking (approved):

Issue #6101, opened 2022-01-12
iquest `not like` operator not working for DATA_RESC_HIER


tedgin:
- [x] main
- [x] 4-3-stable
- [x] 4-2-stable
---
I'm using iCommands version 4.2.8.

The `not` portion of the `not like` operator appears to be ignored by `iquest`, meaning that the condition _A not like B_ and _A like B_ yield the same results. Here's an example.

```
ipc_admin@prod ~? iquest '%s' "select DATA_RESC_HIER where DATA_ID = '726482176' and DATA_RESC_HIER not like 'tacc%'"
taccCorralRes;tacc

ipc_admin@prod /iplant? iquest '%s' "select DATA_RESC_HIER where DATA_ID = '726482176' and DATA_RESC_HIER like 'tacc%'"
taccCorralRes;tacc
```

alanking:
It seems that the `banana` operator also functions like `like` and `not like`:
```
$ iquest "select DATA_RESC_HIER where DATA_NAME = 'foo' and DATA_RESC_HIER banana 'demo%'"
DATA_RESC_HIER = demoResc
------------------------------------------------------------
```
The logic for maintaining the facade for querying on `DATA_RESC_HIER` in GenQuery does not consider anything in the condition string other than the presence or absence of "=". In its absence, anything (including empty string) is treated like "like". This includes "!=", which is treated just like "=".

This may fall under the #6168 umbrella (i.e. use a real parser), but it may be worthwhile to fix this for 4.2.x users.

trel:
Yes, I think we should tighten up that condition check for our known 4.2.x tokens.

Open question whether work on #6168 will/does handle this situation already (@korydraughn)...

korydraughn:
Yes. The syntax for the GenQuery rewrite is already protected from things like that due to the flex/bison implementation.

At this time, `DATA_RESC_HIER` is not implemented. That is one of the final items to handle.

trel:
Nice.    So yes, please just make 4.2.x a bit safer/smarter and this can be closed as 4-2-stable only.

alanking:
We'll just include this in each branch for consistency until the new GenQuery parser is ready.

tedgin:
For posterity, this bug also broke iquest queries that use `||` with `DATA_RESC_HIER`.

Here's an example:
```console
rods@dev ~? iquest "select RESC_NAME where DATA_RESC_HIER like 'ingestRes;%'"
RESC_NAME = ingest
[...]

PR #6802, opened 2022-12-23
[#6101] Fix GenQuery operators for DATA_RESC_HIER (4-3-stable)

alanking:
Addresses #6101 
Cherry-picked from #6751 

Running tests

### DIFF
--- a/server/api/src/rsGenQuery.cpp
+++ b/server/api/src/rsGenQuery.cpp
@@ -12,15 +12,18 @@
 #include "irods/rodsGenQueryNames.h"
 
 #include <boost/format.hpp>
-#include <boost/regex.hpp>
 #include <boost/tokenizer.hpp>
 #include <boost/algorithm/string/predicate.hpp>
 
 #include <cstring>
+#include <regex>
+#include <span>
 #include <string>
 #include <string_view>
 
 namespace {
+    using log_api = irods::experimental::log::api;
+
     std::string
     get_column_name(int j) {
         const int n = sizeof(columnNames)/sizeof(columnNames[0]);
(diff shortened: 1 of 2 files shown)


server/api/src/rsGenQuery.cpp
```diff
@@ -405,39 +424,38 @@ irods::error strip_resc_hier_name_from_query_inp( genQueryInp_t* _inp, int& _pos
...
-            if(!ret.ok()) {
-                irods::log(PASS(ret));
+    try {
+        // iterate over tmp and replace resource group with resource name
+        for (int i = 0; i < tmpV.len; ++i) {
+            const int inx = inxs[i];
+            const char* val = vals[i];
+            if (inx == COL_D_RESC_HIER) {
+                const auto new_cond = get_resc_id_cond_for_hier_cond(val);
+                addInxVal(&_inp->sqlCondInp, COL_D_RESC_ID, new_cond.empty() ? "='0'" : new_cond.c_str());
```
korydraughn, line 447: Can `new_cond.c_str()` result in a dangling pointer when `new_cond` is destructed?

server/api/src/rsGenQuery.cpp
```diff
@@ -650,7 +668,7 @@ rsGenQuery( rsComm_t *rsComm, genQueryInp_t *genQueryInp,
         // the zone name must be skipped when getZoneHintForGenQuery() derives it.
         if (getValByKey(&genQueryInp->condInput, ZONE_KW)) {
             if (!is_zone_name_valid(zone_hint_str)) {
-                rodsLog(LOG_ERROR, "%s:%d :: Unknown zone name [%s].", __func__, __LINE__, zone_hint_str.data());
+                log_api::error("[{}:{}] - Unknown zone name [{}].", __func__, __LINE__, zone_hint_str.c_str());
```
korydraughn, line 671: You can remove the `.c_str()` since `zone_hint_str` is a `std::string`.

korydraughn (commented):
Got that first use of `std::span`!

server/api/src/rsGenQuery.cpp
alanking: `addInxVal` makes a copy of the passed in string for `condInput` via `strdup`, so I don't think that is a concern here:
https://github.com/irods/irods/blob/ec93f905d05776e67870ff80a92aba4227b926bd/lib/core/src/rcMisc.cpp#L866

server/api/src/rsGenQuery.cpp
alanking: Indeed. Should I port this back to main as well?

korydraughn: I think so primarily for cherry-picking reasons. Keeping them in sync makes dealing with multiple branches easier.

alanking: These changes are already merged in main. This *is* the cherry-pick. That being said, I don't mind porting this change over to main.

korydraughn: Oh, if this is identical to what's in main, then leave it as is. No need to change anything.

alanking:
Tests passed

korydraughn (approved):

korydraughn (approved):

trel (approved):

Issue #6794, opened 2022-12-22
std::error_code object's message( ) should not say "Unknown error" for known codes.


d-w-moore:
- [x] main
- [x] 4-3-stable
- [x] 4-2-stable
---
> The name for those error codes is known so we should never see `Unknown error`. It should be simple enough to fix.

_Originally posted by @korydraughn in https://github.com/irods/irods/issues/6627#issuecomment-1362928341_

d-w-moore:
This occurred in the pursuit of #6627, but it can be reproduced by placing these lines at the beginning of main( ) in itree.cpp ( in the icommands repo ):
```
       auto x = fs::detail::make_error_code(-310000);
       std::cout<< x.message()<<"\n"; 
       return 0;

```
The result is:
```

$ ./itree
Unknown error -310000
```
even though the error is known:
```
$ ierror -310000
irods error: -310000 USER_FILE_DOES_NOT_EXIST 
```

PR #6804, opened 2023-01-01
[#6794] std::error_code::message() reports error code name (main)

korydraughn:
Supersedes #6803


lib/core/include/irods/library_features.h
```diff
@@ -23,5 +23,10 @@
 /// \since 4.3.1
 #define IRODS_HAS_API_ENDPOINT_SWITCH_USER    202211L
 
-#endif // IRODS_LIBRARY_FEATURES_H
+/// Defined if the development library supports the C++ system_error library.
+///
+/// \since 4.3.1
+#define IRODS_HAS_LIBRARY_SYSTEM_ERROR        202301L
+    // TODO I don't like this name. We implement only a subset of the lib for our purposes.
```
korydraughn, line 30: See TODO

lib/core/include/irods/replica.hpp
```diff
@@ -24,6 +24,7 @@
 #endif
 
 #include "irods/filesystem.hpp"
+#include "irods/filesystem/detail.hpp" // FIXME This shouldn't be used outside of the fsys lib.
```
korydraughn, line 27: See FIXME comment

unit_tests/src/test_system_error.cpp
```diff
@@ -0,0 +1,41 @@
...
+    const auto errc = irods::experimental::make_error_code(ec);
+    // TODO
+    // Consider the case where a client or plugin wants to check for a specific error code or errno value.
+    // Should .value() return the same value it was constructed with?
+    // Or, should the interface contain more member functions that separate the iRODS error code and the errno value?
+    //     (Seems that separating them makes since. But, at that point, why would anyone ever use .value()?)
+    CHECK(errc.value() == ec);
+    CHECK(errc.message() == "USER_FILE_DOES_NOT_EXIST [errno=120]");
+    CHECK(errc.default_error_condition().category().name() == "iRODS"sv);
+}
```
korydraughn, line 31: See TODO comments

korydraughn: Introduced utility functions for isolating irods/errno codes.

lib/core/include/irods/library_features.h
korydraughn: This is actually fine.

Will remove the TODO comment.

unit_tests/src/test_system_error.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+TEST_CASE("irods_category with positive value with embedded errno value")
+{
+    using namespace std::string_view_literals;
+    constexpr auto ec = -USER_FILE_DOES_NOT_EXIST + 120; // 120 = EINTR
+    const auto errc = irods::experimental::make_error_code(ec);
+    CHECK(errc.value() == ec);
+    CHECK(irods::experimental::get_irods_error_code(errc) == -USER_FILE_DOES_NOT_EXIST);
+    CHECK(irods::experimental::get_errno(errc) == 120);
```
korydraughn, line 55: How do we feel about the `get_` prefix?

Does `extract_` seem better? I feel it expresses intent better at the cost of more characters.

I originally planned to use `strip_`, but that inverts the meaning and takes more effort to understand. For example, `strip_errno(errc)` means _give me the iRODS error code without the errno_.

lib/core/include/irods/system_error.hpp
```diff
@@ -0,0 +1,80 @@
...
+    ///
+    /// Obtains a reference to the static error category object for iRODS errors. The object
+    /// is required to override the virtual function \p std::error_category::name() to return
+    /// a pointer to the string \p "iRODS". It is used to identify error conditions that
+    /// correspond to the iRODS error codes defined in rodsErrorTable.h.
+    ///
+    /// \return A reference to the static object derived from \p std::error_category.
+    ///
+    /// \since 4.2.12
+    [[nodiscard]] auto irods_category() noexcept -> const error_category&;
```
alanking, line 56: Can we put the word "error" somewhere in the name to indicate what this is for? Right now, it will just show up as `irods::experimental::irods_category` which is vague to me.

unit_tests/src/test_system_error.cpp
trel: `extract_` seems the least confusing and the most specific.

korydraughn: `extract_` it is.

lib/core/include/irods/system_error.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+    /// Creates a \p std::error_code object holding the value \p _ec.
+    ///
+    /// \param[in] _ec The integer to create the \p std::error_code object for.
+    ///
+    /// \return A \p std::error_code object.
+    ///
+    /// \since 4.2.6
```
alanking, line 64: 4.2.12?

unit_tests/src/test_system_error.cpp
alanking: I personally like `get_` more because `extract_` implies a modification to the object (by definition, "extract" means "to remove or take out"). What intent does `extract_` express that `get_` is not expressing? I'm not strongly opinionated on this, just want to understand the appeal :)

trel: 'implies a modification'... oh, does it?   if so, then `get_` is more correct / fine with me.

alanking: Well, it implies a modification to me. There may be some standard usage of "extract" of which I am currently unaware

korydraughn: I didn't know that was the meaning of extract. I'll definitely remember that for next time.

`get` can mean anything, but does work. I was trying to express that the function will return an embedded value without modification.

If there is a word that can convey that, that would be nice.

lib/core/include/irods/system_error.hpp
korydraughn: This is mirroring the scene used by the C++ std lib (e.g. `generic_category`, `system_category`, `future_category`, `iostream_category`). No one should be reaching for this in 99% of the cases.

Developers should prefer `make_error_code` instead of constructing `std::error_code` objects themselves. That function will call `irods_category` for you.

Does that make sense?

lib/core/include/irods/system_error.hpp
korydraughn: This function was part of the filesystem lib. It was always available, but usage was not advertised because it was seen as an implementation detail at the time.

With that said, do you feel it is more correct to still change it?

lib/core/include/irods/system_error.hpp
alanking: Oooh, I did not know that. Okay, then I'm fine with this. I don't like it, but that's just how it is for these categories!

lib/core/include/irods/system_error.hpp
alanking: I view this more as a removal and adding a new function (even though it is functionally identical), but I only pointed it out because I noticed that it was different. I'm fine with it either way

unit_tests/src/test_system_error.cpp
trel: `get` seems great.

lib/core/include/irods/system_error.hpp
korydraughn: After some discussion, we've landed on 4.2.12

unit_tests/src/test_system_error.cpp
korydraughn: Moving forward with `get`.

korydraughn: Done.

lib/core/include/irods/system_error.hpp
korydraughn: Done.

lib/core/include/irods/replica.hpp
korydraughn: The question now is how do I address this? Hmm.

Seems my only option is to introduce another file and move the functions into it. Maybe `irods/filesystem/path_utilities.hpp`? Or maybe the detail header becomes `path_utilities.hpp`.

korydraughn: The only problem with that is if someone modifies the behavior in the future and it results in a change in behavior for the fsys lib.

Now, if the replica lib is an extension of the fsys lib, then maybe we don't have to change anything.

Thoughts?

korydraughn: The tests for these libraries aren't exhaustive as far as exceptions are concerned. How would we detect a behavior change before release?

The other thing is that having a detail namespace or header for a lib is signal that those types/functions/etc are private to the implementation and shouldn't be used. I feel we have to separate them to future proof ourselves.

I really don't want to duplicate work, but maybe that is okay here because the functions we're talking about aren't complex.

alanking: I feel that a shared space for these types of functions would still be good. Why are these functions in the detail namespace? I don't think I understand the concern with exposing them in something like `path_utilities.hpp`.

korydraughn: After some discussion, the plan is to move the functions into a new header file, `filesystem/path_utilities.hpp`.

korydraughn: Done.

korydraughn:
Running unit tests.

korydraughn:
The unit tests passed, but I'm going to rerun things due to recent changes.

korydraughn:
All unit tests pass.

Running core tests.

korydraughn:
Fixed core test failure and all core tests passed.

lib/filesystem/include/irods/filesystem/path_utilities.hpp
```diff
@@ -0,0 +1,46 @@
...
+        return path::preferred_separator == _c;
+    }
+
+    /// Throws a filesystem_error if the length of path \p _p exceeds the maximum length
+    /// allowed for a logical path.
+    ///
+    /// 4.2.12
+    inline auto throw_if_path_length_exceeds_limit(const path& _p) -> void
+    {
+        if (std::strlen(_p.c_str()) > MAX_NAME_LEN) {
```
korydraughn, line 29: Codacy is correct. Not only that, but the operator should be `>=` because `MAX_NAME_LEN` represents the buffer size including a null terminating byte, but `std::strlen` does not.

korydraughn: Replaced `std::strlen` with `strnlen` to account for the fixed size.

korydraughn:
Running tests one last time.

korydraughn:

Will add pounds once approved.

korydraughn:
Added directives for some of the clang-format and clang-tidy messages.

Waiting for approvals.

trel:
looks good by inspection

trel:
i notice 4.2.12 in this PR.   but it's for main.   in case that's news to you.

korydraughn:
> i notice 4.2.12 in this PR. but it's for main. in case that's news to you.

That is correct/expected. This code change will be part of 4.2.12 as well.

lib/filesystem/include/irods/filesystem/path_utilities.hpp
```diff
@@ -0,0 +1,46 @@
...
+#include "irods/rodsErrorTable.h"
+#include "irods/system_error.hpp"
+
+#include <cstring>
+
+namespace irods::experimental::filesystem
+{
+    /// Returns whether the path element \p _c represents the preferred path separator.
+    ///
+    /// 4.2.12
```
alanking, line 17: Should this be `\since 4.2.12`, or does Doxygen pick this up?

korydraughn: That's a typo. Will fix

korydraughn: Done

korydraughn:
Any last comments before I add the #'s?

trel:
please to #

alanking (dismissed):

trel:
clang format is not happy

korydraughn:
I'm ignoring remaining clang-format messages because they lead to improper formatting.
I'm ignoring the clang-tidy message because it is not necessary for that file.

korydraughn:
Added #'s

alanking (approved):

Issue #6482, opened 2022-07-20
Delayed rules do not honor remote user ownership


alanking:
- [x] main
- [x] 4-3-stable
- [x] 4-2-stable
---
### iRODS Version, OS and Version
4.2.11

### What did you try to do?

- Created 2 zones: zoneA and zoneB
- Created a rodsuser in zoneB named alice (alice#zoneB) and a rodsuser in zoneA named alice#zoneB
- Implemented the following rule in core.re in zoneA:
```
pep_api_data_obj_put_post(*INSTANCE_NAME, *COMM, *DATAOBJINP, *BUFFER, *PORTAL_OPR_OUT)
{
    delay("<PLUSET>31s</PLUSET>") {
        writeLine('serverLog', 'pep_api_data_obj_put_post... but delayed');
    }
    writeLine('serverLog', 'pep_api_data_obj_put_post... enqueued delayed rule');
}
```
As alice#zoneB in zoneB, I ran the following:
```
$ iput localfile /zoneA/home/public/foo
```

The put to zoneA completes successfully and of course the PEP in zoneA caused a delayed rule to be enqueued in zoneA:
```
$ iqstat -al
Pending rule-executions
id: 10028
name:   
        writeLine('serverLog', 'pep_api_data_obj_put_post... but delayed');
    
rei_file_path: EMPTY_REI_PATH
user_name: alice
address: 
time: 1658327271 : 2022-07-20.14:27:51
frequency: 
priority: 5
estimated_exe_time: 
notification_addr: 
last_exe_time: 
exec_status: 
```

### Expected behavior
See my writeLine call in the rodsLog and the delayed rule is removed from the queue.

### Observed behavior (including steps to reproduce, if applicable)
In the logs I see this:
```
Jul 20 14:27:20 pid:557 NOTICE: writeLine: inString = pep_api_data_obj_put_post... enqueued delayed rule
Jul 20 14:28:08 pid:607 NOTICE: writeLine: inString = pep_api_data_obj_put_post... but delayed
Jul 20 14:28:08 pid:607 remote addresses: 192.168.160.3, 192.168.160.4 ERROR: rsRuleExecDel: rcRuleExecDel failed, status = -350000
Jul 20 14:28:38 pid:613 NOTICE: writeLine: inString = pep_api_data_obj_put_post... but delayed
[...]

PR #6805, opened 2023-01-01
[#6482] Deleting a delay rule now verifies zone of the connected client (4-2-stable)

korydraughn:
This has not been tested and is still a work in progress.

### DIFF
--- a/plugins/api/include/api_plugin_number_data.h
+++ b/plugins/api/include/api_plugin_number_data.h
@@ -8,3 +8,8 @@ API_PLUGIN_NUMBER(ATOMIC_APPLY_ACL_OPERATIONS_APN,              20005)
 API_PLUGIN_NUMBER(DATA_OBJECT_FINALIZE_APN,                     20006)
 API_PLUGIN_NUMBER(TOUCH_APN,                                    20007)
 API_PLUGIN_NUMBER(REGISTER_PHYSICAL_PATH_APN,                   20008)
+//
+// API plugin numbers 20009-20012 are mapped to API plugins only
+// available in the 4.3 series.
+//
+API_PLUGIN_NUMBER(GET_DELAY_RULE_INFO_APN,                      20013)
--- a/plugins/api/src/get_delay_rule_info.cpp
+++ b/plugins/api/src/get_delay_rule_info.cpp
@@ -1,0 +1,195 @@
+#include "apiHandler.hpp"
+#include "api_plugin_number.h"
+#include "client_api_whitelist.hpp"
+#include "rcConnect.h"
+#include "rodsErrorTable.h"
+#include "rodsPackInstruct.h"
+
+#include <functional>
+
+#ifdef RODS_SERVER
+
+//
+// Server-side Implementation
+//
+
+#include "get_delay_rule_info.h"
+
+#include "catalog.hpp"
+#include "catalog_utilities.hpp"
+#include "icatHighLevelRoutines.hpp"
+#include "irods_re_serialization.hpp"
+#include "irods_server_api_call.hpp"
+#include "rodsLog.h"
+#include "server_utilities.hpp"
+
+#include <json.hpp>
+#include <fmt/format.h>
+
+#include <string>
+
+namespace
+{
+    //
+    // Function Prototypes
+    //
+
+    auto call_get_delay_rule_info(irods::api_entry*, RsComm*, const char*, BytesBuf**) -> int;
+
+    auto rs_get_delay_rule_info_impl(RsComm*, const char*, BytesBuf**) -> int;
+
+    auto is_valid_input(const char*) -> bool;
+
+    //
+    // Function Implementations
+    //
+
+    auto call_get_delay_rule_info(irods::api_entry* _api, RsComm* _comm, const char* _rule_id, BytesBuf** _output) -> int
+    {
+        return _api->call_handler<const char*, BytesBuf**>(_comm, _rule_id, _output);
+    } // call_get_delay_rule_info
+
+    auto is_valid_input(const char* _rule_id, BytesBuf** _output) -> bool
+    {
+        if (!_rule_id || !_output) {
+            rodsLog(LOG_ERROR, "Invalid input: _rule_id or _output is null");
+            return false;
+        }
+
+        // 64-bit integers can use up to 20 digits.
+        // We add 1 to the value to accomodate for the null byte.
+        // This size does not consider negative, hexidecimal, or octal values.
+        constexpr std::size_t max_number_of_digits = 20 + 1;
+
+        if (!is_non_empty_string(_rule_id, max_number_of_digits)) {
+            rodsLog(LOG_ERROR, "Invalid input: _rule_id is empty or cannot be represented by a 64-bit integer.");
+            return false;
+        }
+
+        // Try to parse the string into a valid signed 64-bit integer.
+        try {
+            if (const auto i = std::stoll(_rule_id); i <= 0) {
+                rodsLog(LOG_ERROR, "Invalid input: _rule_id [%d] must be greater than zero.", i);
+                return false;
+            }
+        }
+        catch (...) {
+            rodsLog(LOG_ERROR, "Invalid input: could not parse _rule_id [%s] into a signed integer.", _rule_id);
+            return false;
+        }
+
+        return true;
+    } // is_valid_input
+
+    auto rs_get_delay_rule_info_impl(RsComm* _comm, const char* _rule_id, BytesBuf** _output) -> int
+    {
+        if (!is_valid_input(_rule_id, _output)) {
+            return SYS_INVALID_INPUT_PARAM;
+        }
+
+        rodsLog(LOG_DEBUG, "%s: _rule_id = [%s]", __func__, _rule_id);
+
+        try {
+            namespace ic = irods::experimental::catalog;
+
+            if (!ic::connected_to_catalog_provider(*_comm)) {
+                irods::log(LOG_DEBUG8, "Redirecting request to catalog service provider ...");
+                auto* host_info = ic::redirect_to_catalog_provider(*_comm);
+                return rc_get_delay_rule_info(host_info->conn, _rule_id, _output);
+            }
+
+            ic::throw_if_catalog_provider_service_role_is_invalid();
+        }
+        catch (const irods::exception& e) {
+            rodsLog(LOG_ERROR, e.what());
+            return e.code();
+        }
+
+        std::vector<std::string> row;
+
+        if (const auto ec = chl_get_delay_rule_info(*_comm, _rule_id, &row); ec != 0) {
+            rodsLog(LOG_ERROR, "Could not get delay rule information [rule id=[%s]]", _rule_id);
+            return ec;
+        }
+
+        try {
+            using json = nlohmann::json;
+
+            const auto json_string = json{
+                {"rule_id", _rule_id},
+                {"rule_name", row.at(0)},
+                {"rei_file_path", row.at(1)},
+                {"user_name", row.at(2)},
+                {"exe_address", row.at(3)},
+                {"exe_time", row.at(4)},
+                {"exe_frequency", row.at(5)},
+                {"priority", row.at(6)},
+                {"last_exe_time", row.at(7)},
+                {"exe_status", row.at(8)},
+                {"estimated_exe_time", row.at(9)},
+                {"notification_addr", row.at(10)},
+                {"exe_context", row.at(11)}
+            }.dump();
+
+            *_output = irods::to_bytes_buffer(json_string);
+        }
+        catch (const std::exception& e) {
+            rodsLog(LOG_ERROR, "An error occurred while serializing the JSON object: %s", e.what());
+            return SYS_LIBRARY_ERROR;
+        }
+        catch (...) {
+            rodsLog(LOG_ERROR, "An unknown error occurred while serializing the JSON object");
+            return SYS_UNKNOWN_ERROR;
+        }
+
+        return 0;
+    } // rs_get_delay_rule_info_impl
+
+    using operation = std::function<int(RsComm*, const char*, BytesBuf**)>;
+    const operation op = rs_get_delay_rule_info_impl;
+    #define CALL_GET_DELAY_RULE_INFO call_get_delay_rule_info
+} // anonymous namespace
+
+#else // RODS_SERVER
+
+//
+// Client-side Implementation
+//
+
+namespace
+{
+    using operation = std::function<int(RsComm*, const char*, BytesBuf**)>;
+    const operation op{};
+    #define CALL_GET_DELAY_RULE_INFO nullptr
+} // anonymous namespace
+
+#endif // RODS_SERVER
+
+// The plugin factory function must always be defined.
+extern "C"
+auto plugin_factory(const std::string& _instance_name,
+                    const std::string& _context) -> irods::api_entry*
+{
+    // clang-format off
+    irods::apidef_t def{GET_DELAY_RULE_INFO_APN,         // API number
+                        RODS_API_VERSION,                // API version
+                        LOCAL_PRIV_USER_AUTH,            // Client auth
+                        LOCAL_PRIV_USER_AUTH,            // Proxy auth
+                        "STR_PI", 0,                     // In PI / bs flag
+                        "BinBytesBuf_PI", 0,             // Out PI / bs flag
+                        op,                              // Operation
+                        "api_get_delay_rule_info",       // Operation name
+                        nullptr,                         // Clear function
+                        (funcPtr) CALL_GET_DELAY_RULE_INFO};
+    // clang-format on
+
+    auto* api = new irods::api_entry{def};
+
+    api->in_pack_key = "STR_PI";
+    api->in_pack_value = STR_PI;
+
+    api->out_pack_key = "BinBytesBuf_PI";
+    api->out_pack_value = BytesBuf_PI;
+
+    return api;
+} // plugin_factory
--- a/plugins/database/src/db_plugin.cpp
+++ b/plugins/database/src/db_plugin.cpp
@@ -1,3 +1,5 @@
+#include "irods_at_scope_exit.hpp"
+#include "irods_re_structs.hpp"
 #include "rodsDef.h"
 #include "authenticate.h"
 #include "rodsQuota.h"
@@ -35,12 +37,15 @@
 #include "rodsErrorTable.h"
 #include "irods_lexical_cast.hpp"
 #include "catalog_utilities.hpp"
+#include "catalog.hpp"
+#include "json_deserialization.hpp"
 
 #include <boost/date_time.hpp>
 #include <boost/regex.hpp>
 #include <boost/lexical_cast.hpp>
 #include <fmt/format.h>
 #include <fmt/chrono.h>
+#include <nanodbc/nanodbc.h>
 
 #include <cstring>
 #include <sstream>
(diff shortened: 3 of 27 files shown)


server/api/src/rsRuleExecDel.cpp
```diff
@@ -4,72 +4,72 @@
...
+        if (!irods::is_privileged_client(*rsComm)) {
+            // TODO Add the following federation tests:
+            // 1. Local user cannot delete remote user's delay rule.
+            // 2. Remote user cannot delete local user's delay rule.
+            // 3. Remote user can delete their own delay rule.
+
+            // TODO Allow remote users to delete their own delay rules.
+            if (rsComm->clientUser.authInfo.authFlag != LOCAL_USER_AUTH) {
+                return USER_ACCESS_DENIED;
+            }
```
korydraughn, line 32: See TODO comments

server/api/src/rsRuleExecDel.cpp
```diff
@@ -4,72 +4,72 @@
...
 
-        clearGenQueryInp( &genQueryInp );
+            // TODO How do we verify if the client invoking this API is who they say they are?
+            // A1. Assign the user's ID to the delay rule instead of their username.
+            //     - Instead, add a new column to the table.
+            // A2. Lookup the exe_context info, inflate it, and compare the username and zone to the RsComm.
+            // A3. Store the zone in r_rule_exec and add zone to the RuleExecSubmitInput struct.
```
korydraughn, line 38: See TODO comments

server/icat/include/icatHighLevelRoutines.hpp
```diff
@@ -273,4 +273,18 @@ auto chl_check_permission_to_modify_data_object(RsComm& _comm, const rodsLong_t
...
+/// \param[in]     _rule_id  The ID of the delay rule.
+/// \param[in,out] _username The address of the char pointer that will hold the user's name.
+/// \param[in,out] _zone     The address of the char pointer that will hold the user's zone.
+///
+/// \returns An error code representing whether the operation was successful.
+/// \retval  0 On success.
+/// \retval <0 Otherwise.
+///
+/// \since 4.2.12
+auto chl_get_delay_rule_owner(RsComm& _comm, const char* _rule_id, char** _username, char** _zone) -> int;
```
korydraughn, line 288: Finish documentation

korydraughn (commented):
Still needs tests.

server/api/src/rsRuleExecDel.cpp
korydraughn: Went with (A2).

Remove comments.

korydraughn: done

trel:
i think the new name is clear enough.

the only case i can think that could be confusing/misleading is if the creator is not always the owner... in a proxy situation?  not sure that's a real concern... b/c if rods is proxying for alice, then 'alice' is the one who adds the delay rule, and therefore owns it.   so, i think it's okay?

korydraughn:
That was one of my concerns as well. I don't think there are cases where the user's name changes though so we are probably okay.

Moving forward with the rename.

trel:
agreed - rules cannot be edited once they're on the queue.   only executed and/or deleted.

server/api/src/rsRuleExecDel.cpp
```diff
@@ -4,72 +4,60 @@
...
-
-        const auto status = rsGenQuery( rsComm, &genQueryInp, genQueryOut );
-
-        clearGenQueryInp( &genQueryInp );
+        // Verify the user is the creator of the delay rule if they aren't an administrator.
+        if (!irods::is_privileged_client(*rsComm)) {
+            // TODO Add the following federation tests:
+            // 1. Local user cannot delete remote user's delay rule.
+            // 2. Remote user cannot delete local user's delay rule.
+            // 3. Remote user can delete their own delay rule.
```
korydraughn, line 27: Remove once tests are implemented.

server/core/src/rsApiHandler.cpp
```diff
@@ -394,11 +394,17 @@ int chkApiPermission(rsComm_t* rsComm, int apiInx)
 
     const int clientUserAuth = api_entry->clientUserAuth & 0xfff; // Remove XMSG_SVR_* flags
     if (clientUserAuth > rsComm->clientUser.authInfo.authFlag) {
+        rodsLog(LOG_NOTICE, "api_entry->clientUserAuth            = [%d]", api_entry->clientUserAuth);
+        rodsLog(LOG_NOTICE, "api_entry->clientUserAuth & 0xfff    = [%d]", (api_entry->clientUserAuth & 0xfff));
+        rodsLog(LOG_NOTICE, "rsComm->clientUser.authInfo.authFlag = [%d]", rsComm->clientUser.authInfo.authFlag);
```
korydraughn, line 399: Remove or convert to debug messages.

server/core/src/rsApiHandler.cpp
```diff
@@ -394,11 +394,17 @@ int chkApiPermission(rsComm_t* rsComm, int apiInx)
...
         return SYS_NO_API_PRIV;
     }
 
     const int proxyUserAuth = api_entry->proxyUserAuth & 0xfff;
     if (proxyUserAuth > rsComm->proxyUser.authInfo.authFlag) {
+        rodsLog(LOG_NOTICE, "api_entry->proxyUserAuth            = [%d]", api_entry->proxyUserAuth);
+        rodsLog(LOG_NOTICE, "api_entry->proxyUserAuth & 0xfff    = [%d]", (api_entry->proxyUserAuth & 0xfff));
+        rodsLog(LOG_NOTICE, "rsComm->proxyUser.authInfo.authFlag = [%d]", rsComm->proxyUser.authInfo.authFlag);
```
korydraughn, line 407: Remove or convert to debug messages.

korydraughn:
I've manually verified that this PR supports the following:
- Deleting delay rules as a local user
- Deleting delay rules as a remote user from the remote zone
- Deleting delay rules as a remote user from the local zone

That said, I don't think we have facilities in place to automate the remote cases. That is something we'll need address separately. It will most likely require direct support in the testing environment too.

Any thoughts?

server/core/src/rsApiHandler.cpp
korydraughn: Removed.

server/api/src/rsRuleExecDel.cpp
korydraughn: Removed based on https://github.com/irods/irods/pull/6805#issuecomment-1376144067

plugins/database/src/db_plugin.cpp
```diff
@@ -15757,6 +15723,45 @@ auto db_update_ticket_write_byte_count_op(
...
+    }
+
+    if (!_rule_id || !_username || !_zone) {
+        return ERROR(SYS_INVALID_INPUT_PARAM, "Invalid input: rule id, username, or zone is null");
+    }
+
+    auto [db_instance, db_conn] = irods::experimental::catalog::new_database_connection();
+
+    nanodbc::statement stmt{db_conn};
+    nanodbc::prepare(stmt, "select exe_context from R_RULE_EXEC where rule_exec_id = ?");
```
alanking, line 15743: Could this be done with a GenQuery? Or is there a reason we would want it in a database operation?

alanking (commented):
Here's one test idea which may or may not work with the existing test infrastructure.

The federation tests are run on otherZone. A session can be made which connects to tempZone as a tempZone user. Using the remote directive, a delayed rule can be scheduled on otherZone. Then we can switch back to a session connected to otherZone and try to delete the rule and watch it fail.

plugins/database/src/db_plugin.cpp
trel: as i understand it, genquery can't get the context - could be too big to serialize...

korydraughn: That's correct. It's more so has to do with the db plugin's implementation. The db plugin isn't designed to handle columns holding large amounts of data.

server/api/src/rsRuleExecDel.cpp
```diff
@@ -4,72 +4,55 @@
...
+        // Verify the user is the creator of the delay rule if they aren't an administrator.
+        if (!irods::is_privileged_client(*rsComm)) {
+            char* username{};
+            char* zone{};
+            irods::at_scope_exit free_gen_query_output{[&username, &zone] {
+                std::free(username);
+                std::free(zone);
+            }};
+
+            const int ec = chl_get_creator_of_delay_rule(*rsComm, ruleExecDelInp->ruleExecId, &username, &zone);
```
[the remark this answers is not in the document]
korydraughn: Turns out that we check the service role twice. By the time we've reached this function, we're already on the provider.

The service role logic you referenced should be removed. I'll do that now.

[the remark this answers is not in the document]
korydraughn: I've rewritten the whole API endpoint lol.

Everything should be better now.

server/api/src/rsRuleExecDel.cpp
```diff
@@ -4,72 +4,55 @@
...
 #include "irods_at_scope_exit.hpp"
 #include "irods_configuration_keywords.hpp"
+#include "irods_rs_comm_query.hpp"
 #include "miscServerFunct.hpp"
 #include "objMetaOpr.hpp"
 #include "rcMisc.h"
+#include "rodsErrorTable.h"
 #include "rsGenQuery.hpp"
 #include "ruleExecSubmit.h"
+#include "icatHighLevelRoutines.hpp"
```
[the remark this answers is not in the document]
korydraughn: Removed it.

korydraughn:
I realized that the delay server still relies on nanodbc directly. Given the latest changes, I should go ahead and extend the db plugin function to fetch the full row from the database. Then both deletion of delay rules and the delay server can leverage that.

server/api/src/rsRuleExecDel.cpp
```diff
@@ -1,142 +1,97 @@
...
         }
 
-        if (irods::CFG_SERVICE_ROLE_CONSUMER == svc_role) {
-            rodsLog(LOG_ERROR, "_rsRuleExecDel: chlDelRuleExec must be invoked on the catalog provider host");
-            return SYS_NO_RCAT_SERVER_ERR;
+        // Delete the delay rule from the catalog.
+        const auto ec = chlDelRuleExec(rsComm, ruleExecDelInp->ruleExecId);
+
+        if (ec < 0) {
+            rodsLog(LOG_ERROR, "{}: chlDelRuleExec for %s error, status = %d",
```
alanking, line 64: Is `{}` supposed to be `%s` here?

alanking (commented):
I like the more general database operation.

I recently did something similar to your refactor of the API except in `rsModColl` (in another branch), so, glad to see that was the right path :)

server/api/src/rsRuleExecDel.cpp
korydraughn: Fixed it.

korydraughn:
All unit tests and core tests passed for a8f2642 (contains no **new** tests).

plugins/api/include/api_plugin_number_data.h
```diff
@@ -8,3 +8,4 @@ API_PLUGIN_NUMBER(ATOMIC_APPLY_ACL_OPERATIONS_APN,              20005)
 API_PLUGIN_NUMBER(DATA_OBJECT_FINALIZE_APN,                     20006)
 API_PLUGIN_NUMBER(TOUCH_APN,                                    20007)
 API_PLUGIN_NUMBER(REGISTER_PHYSICAL_PATH_APN,                   20008)
+API_PLUGIN_NUMBER(GET_DELAY_RULE_INFO_APN,                      20009)
```
alanking, line 11: Heads up that `GET_GRID_CONFIGURATION_VALUE_APN` has value 20009 on main/4-3-stable.

An interesting situation that I'm not sure we have encountered as of yet...

plugins/api/src/get_delay_rule_info.cpp
```diff
@@ -0,0 +1,165 @@
+#include "api_plugin_number.h"
+#include "rcConnect.h"
+#include "rodsErrorTable.h"
+#include "rodsPackInstruct.h"
+#include "client_api_whitelist.hpp"
+#include "apiHandler.hpp"
```
alanking, line 6: Alphabetize?

plugins/api/src/get_delay_rule_info.cpp
```diff
@@ -0,0 +1,165 @@
...
+        }
+        catch (const irods::exception& e) {
+            rodsLog(LOG_ERROR, e.what());
+            return e.code();
+        }
+
+        std::vector<std::string> row;
+
+        if (const auto ec = chl_get_delay_rule_info(*_comm, _rule_id, &row); ec != 0) {
+            rodsLog(LOG_ERROR, "Failed to get delay rule info [rule id=[{}]]", _rule_id);
```
alanking, line 76: `{}` should be `%s`, I think?

plugins/database/src/db_plugin.cpp
```diff
@@ -15757,6 +15723,45 @@ auto db_update_ticket_write_byte_count_op(
...
+                           " exe_frequency, priority, last_exe_time, exe_status, estimated_exe_time,"
+                           " notification_addr, exe_context "
+                           "from R_RULE_EXEC where rule_exec_id = ?");
+
+    stmt.bind(0, _rule_id);
+
+    auto row = nanodbc::execute(stmt);
+
+    if (!row.next()) {
+        rodsLog(LOG_ERROR, "Could not find a delay rule with id [{}].");
```
alanking, line 15752: `{}` should be `%s`, and need to pass `_rule_id` as an arg

plugins/database/src/db_plugin.cpp
```diff
@@ -15757,6 +15723,45 @@ auto db_update_ticket_write_byte_count_op(
...
+        return ERROR(CAT_NO_ROWS_FOUND, "Could not find a delay rule with id [{}].");
```
alanking, line 15753: Missing argument and `fmt::format` for the `{}`

plugins/api/include/api_plugin_number_data.h
korydraughn: Most definitely.

trel: looks like 20013 is in your future...

https://github.com/irods/irods/blob/7a8ea89af09a3ab222f8009825dda0b207328735/plugins/api/include/irods/plugins/api/api_plugin_number_data.h#L15-L16

plugins/database/src/db_plugin.cpp
korydraughn: Done.

plugins/database/src/db_plugin.cpp
korydraughn: Done.

plugins/api/src/get_delay_rule_info.cpp
korydraughn: Done.

plugins/api/include/api_plugin_number_data.h
korydraughn: Done.

korydraughn: I think we need a comment explaining the gap in API plugin numbers. I just don't know where I want to put that. Hmm ...

plugins/api/src/get_delay_rule_info.cpp
korydraughn: Done

plugins/api/include/api_plugin_number_data.h
trel: just add it right there and say 4.3.0 already has 9-12.

korydraughn: I considered that, but hesitated because that file is changed based on the definition of the `API_PLUGIN_NUMBER` macro.

I'll thinking on it.

unit_tests/src/test_get_delay_rule_info.cpp
```diff
@@ -0,0 +1,108 @@
...
+        CHECK(info.count("exe_context"));
+    }
+
+    SECTION("invalid input")
+    {
+        BytesBuf* bbuf{};
+        CHECK(rc_get_delay_rule_info(conn_ptr,     nullptr,   &bbuf) == SYS_INVALID_INPUT_PARAM);
+        CHECK(rc_get_delay_rule_info(conn_ptr,     "00000", nullptr) == SYS_INVALID_INPUT_PARAM);
+        CHECK(rc_get_delay_rule_info(conn_ptr,        "-1",   &bbuf) ==       CAT_NO_ROWS_FOUND);
+        CHECK(rc_get_delay_rule_info(conn_ptr, "bad_input",   &bbuf) ==       SYS_LIBRARY_ERROR);
```
alanking, line 72: Is an empty string (`""`) considered valid input? What happens?

korydraughn:
Requires https://github.com/irods/irods_testing_environment/pull/147

unit_tests/src/test_get_delay_rule_info.cpp
korydraughn: Done.

plugins/api/include/api_plugin_number_data.h
korydraughn: Added comment explaining the gap.

korydraughn:
All federations tests passed.

korydraughn:

Testing with a concurrency level of 8!
```bash
-----
All tests passed! :)
time elapsed: [4752.0710]seconds ([   1]hours [19.2012]minutes)
```

korydraughn:
Made a few changes to the new api plugin.

All unit tests pass.

Rerunning core tests.

korydraughn:
All core tests passed.

Rerunning federation tests.

korydraughn:
All federation tests passed.

If there are no objections, I'm going to add the pounds.

trel:
let's see the issue number in the third commit first...

scripts/irods/test/test_federation.py
```diff
@@ -1584,3 +1601,97 @@ def test_peps_for_parallel_mode_transfers__issue_5017(self):
...
+    def test_local_zone_user_is_not_allowed_to_delete_delay_rules_created_by_remote_user_with_the_same_username__issue_6482(self):
+        try:
+            with temporary_core_file() as core_re:
+                # Users in a federated environment are not allowed to create delay rules in the
+                # remote zone directly. Remote users can only create delay rules indirectly
+                # (e.g. through a PEP in the federated zone).
+                core_re.add_rule(dedent('''
+                    pep_api_touch_pre(*a, *b, *c)
+                    {
+                        delay("<INST_NAME>irods_rule_engine_plugin-irods_rule_language-instance</INST_NAME>") {
```
alanking, line 1673: Should we put a `PLUSET` of some very large number so that the rule doesn't execute and leave the queue? There is a highly unlikely but theoretically possible (I think?) scenario in which the remote delay server executes the delay rule before we can fetch it from the queue on 1686 which could lead to a false negative.

scripts/irods/test/test_federation.py
```diff
@@ -1123,6 +1124,22 @@ def test_federation_support_for_replica_open_close_and_get_file_descriptor_info(
...
             user.run_icommand(['irm', '-f', parameters['remote_data_object']])
 
+    def test_itouch__issue_XXXX(self):
+        user = self.user_sessions[0]
+        parameters = self.config.copy()
+
+        parameters['filename'] = 'itouch_test_file.txt'
+        parameters['user_name'] = user.username
+        parameters['remote_home_collection'] = '/{remote_zone}/home/{user_name}#{local_zone}'.format(**parameters)
+        parameters['remote_data_object'] = '{remote_home_collection}/{filename}'.format(**parameters)
```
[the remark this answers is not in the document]
korydraughn: I agree. I don't like this style, but I'm going to leave it for now.

scripts/irods/test/test_federation.py
trel: yes, put it farther into the future please.

scripts/irods/test/test_federation.py
```diff
@@ -1123,6 +1124,22 @@ def test_federation_support_for_replica_open_close_and_get_file_descriptor_info(
         finally:
             user.run_icommand(['irm', '-f', parameters['remote_data_object']])
 
+    def test_itouch__issue_XXXX(self):
```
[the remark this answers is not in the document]
korydraughn: Done

scripts/irods/test/test_federation.py
korydraughn: Done

korydraughn:
The latest modifications have been tested.

Shall I add the pounds?

alanking (approved):
I think we can #, yep

korydraughn:
Added #'s

alanking (approved):

Issue #6791, opened 2022-12-20
itree --pattern and --ignore should mimic "tree" command more closely


d-w-moore:
- [x] main
- [x] 4-3-stable
- [x] 4-2-stable
---
## Bug Report
irods 4.2.11

### What did you try to do?
use `--pattern` (aka `-P`) and `--ignore` (`-I`) to control data objects listed in the manner of the Linux `tree` command.

### Expected behavior
Mimic tree behavior.

### Observed behavior (including steps to reproduce, if applicable)
```
$ mkdir -p tmp/coll ; touch tmp/coll/{x,y,z}{1,2}; iput -r tmp
$ tree tmp -P "X*"
tmp
└── coll
    ├── x1
    └── x2

$ itree tmp -P "x*"
tmp
  coll
Found 1 collections and 0 data objects
$ $ itree tmp -P'x.*'
tmp
  coll
    x1
    x2
```
Evidently itree uses regexes instead of file glob patterns.
So we might do one or two things along the lines of these fixes/enhancements:
   1. add a --regex-type option declaring what kind of patterns we want to use (filename-glob, regex, extended-regex)
   1. multiple search patterns could be encoded as '|'-separated expressions in an extended regex.(see the mention of  allowing "more than one argument"  at the end of the problem description for #6627)

d-w-moore:
There may be heavy overlap here with #6628 - there's confusion there, as well , centered around regexes vs globbing.

alanking:
@d-w-moore - Please close if complete. Thanks!

Issue #6806, opened 2023-01-02
allow itree options specifying multiple match/ignore patterns


d-w-moore:
- [x] main
- [x] 4-3-stable
- [x] 4-2-stable
---
## Feature
Since`itree` already includes regex matching, and [this pull request](https://github.com/irods/irods_client_icommands/pull/340) currently plans to continue that option with the command-line switches  `--pattern-regex` and `--ignore-regex`, and since the extended regex syntax includes an OR-selector via "ptn1|ptn2|...", it seems that one way to accommodate this capability (requested at the end of [this comment](https://github.com/irods/irods/issues/6627#issue-1402401130)) might be by the addition of another switch(`--regex-extended-syntax`?)
This may be preferable to allowing multiple --ignore* and --pattern* invocations,as this would run counter to usage in the `tree` command whose example we are ostensibly following.

alanking:
@d-w-moore - Please close if complete. Thanks!

PR #6807, opened 2023-01-03
[#6791] test both itree options (regex vs glob) (4-2-stable)

d-w-moore:


trel:
please add a test or two around using the pattern and ignore options together in different combinations.

separately, what happens if both regex and glob are used at the same time?  maybe those tests should be part of a different issue, not sure...

scripts/irods/test/test_itree.py
```diff
@@ -46,6 +46,23 @@ def test_depth(self):
...
+        # Using forward and reverse combination of -P and -I options, catch objects matching /bar[0-9]/
+        self.admin.assert_icommand(["itree", "-P", "ba*", "-I", "*ar?"], "STDOUT_SINGLELINE", "Found 0 collections and 10 data objects")
+        self.admin.assert_icommand(["itree", "-I", "*ar?", "-P", "ba*"], "STDOUT_SINGLELINE", "Found 0 collections and 10 data objects")
+
+        # Using regular expressions, catch objects matching /bat[0-9]/
+        self.admin.assert_icommand(["itree", "--ignore-regex", ".*ar.", "--pattern-regex", "ba.*"], "STDOUT_SINGLELINE", "Found 0 collections and 10 data objects")
+
+        # Using regular expressions with extended syntax, catch objects matching /(bar|bat)[0-2]/x
+        self.admin.assert_icommand(["itree",  "--regex-extended", "--ignore-regex", ".*[3-9]", "--pattern-regex", "ba(r|t)[[:digit:]]"],
+                                    "STDOUT_SINGLELINE", "Found 0 collections and 6 data objects")
```
korydraughn, line 64: Do we have tests that verify the output contains only the matched objects?

If not, then these assertions seem like good cases to do that. You can capture the value returned by `assert_icommand` and do that. That way if anything changes in the future, we'll catch it.

d-w-moore: It's my observation that none of the other tests for itree do that, thus far.  Should an issue be made for that?

d-w-moore:
> separately, what happens if both regex and glob are used at the same time? maybe those tests should be part of a different issue, not sure...

regex and glob can be used together, eg --pattern and --ignore-regex would make sense.   --pattern and --pattern-regex would not make sense, however, and possibly should be caught as an error. At present, pretty sure the later option would override the earlier one.

scripts/irods/test/test_itree.py
trel: separate or along with this set - either is fine.

trel:
> together would not make sense

yes, please catch anything that should result in `402000 USER_INCOMPATIBLE_PARAMS`

scripts/irods/test/test_itree.py
```diff
@@ -46,6 +46,23 @@ def test_depth(self):
         for index,item in enumerate(directory_names):
             self.admin.assert_icommand(["itree", "-L", str(1 + index)], 'STDOUT_SINGLELINE', ["  " * (1 + index) + item])
 
+    def test_pattern_and_ignore_options_together__6791_6806(self):
+        patterns = ['foo', 'bar', 'bat']
+        for i in range(10):
+            for p in patterns:
+                self.admin.assert_icommand(['itouch', p + str(i)])
+
+        # Using forward and reverse combination of -P and -I options, catch objects matching /bar[0-9]/
```
trel, line 55: matches `/bat[0-9]/`, rather than `bar`, right?

scripts/irods/test/test_itree.py
trel, line 64: can use `STDOUT_MULTILINE` to list the expected results a bit more specifically.

scripts/irods/test/test_itree.py
```diff
@@ -46,6 +46,23 @@ def test_depth(self):
...
+        self.admin.assert_icommand(["itree", "-P", "ba*", "-I", "*ar?"], "STDOUT_SINGLELINE", "Found 0 collections and 10 data objects")
+        self.admin.assert_icommand(["itree", "-I", "*ar?", "-P", "ba*"], "STDOUT_SINGLELINE", "Found 0 collections and 10 data objects")
+
+        # Using regular expressions, catch objects matching /bat[0-9]/
+        self.admin.assert_icommand(["itree", "--ignore-regex", ".*ar.", "--pattern-regex", "ba.*"], "STDOUT_SINGLELINE", "Found 0 collections and 10 data objects")
```
trel, line 60: can use `STDOUT_MULTILINE` to list the expected results a bit more specifically.

scripts/irods/test/test_itree.py
```diff
@@ -46,6 +46,23 @@ def test_depth(self):
...
 
+    def test_pattern_and_ignore_options_together__6791_6806(self):
+        patterns = ['foo', 'bar', 'bat']
+        for i in range(10):
+            for p in patterns:
+                self.admin.assert_icommand(['itouch', p + str(i)])
+
+        # Using forward and reverse combination of -P and -I options, catch objects matching /bar[0-9]/
+        self.admin.assert_icommand(["itree", "-P", "ba*", "-I", "*ar?"], "STDOUT_SINGLELINE", "Found 0 collections and 10 data objects")
+        self.admin.assert_icommand(["itree", "-I", "*ar?", "-P", "ba*"], "STDOUT_SINGLELINE", "Found 0 collections and 10 data objects")
```
trel, line 57: can use `STDOUT_MULTILINE` to list the expected results a bit more specifically.

scripts/irods/test/test_itree.py
alanking: The `-I` option has a value of `*ar?`, which would suggest that `bar` would match and `bat` would not,?

scripts/irods/test/test_itree.py
```diff
@@ -46,17 +46,41 @@ def test_depth(self):
...
             self.admin.assert_icommand(["itree", "-L", str(1 + index)], 'STDOUT_SINGLELINE', ["  " * (1 + index) + item])
 
+    def test_pattern_and_ignore_options_together__6791_6806(self):
+        patterns = ['foo', 'bar', 'bat']
+        for i in range(10):
+            for p in patterns:
+                self.admin.assert_icommand(['itouch', p + str(i)])
+        to_catch = [r'\b{}\b'.format('bat'+str(i)) for i in range(10)]
+
+        # Using forward and reverse combination of -P and -I options, catch 'baz*' objects as result
```
trel, line 56: catch 'bat*'

alanking:
Seems to be one unresolved comment left here and this is probably good to go (of course, waiting on irods/irods_client_icommands#340 as well)

alanking:
Looks like the comments have been resolved. Please squash to taste and leave off #s for now and we will have one more look.

alanking (commented):
Looks good to me. Is this also related to #6806? Consider linking that one in your commit message as well, if so.

Please squash to taste, leave off #s, and we'll take one last look.

d-w-moore:
> Looks good to me. Is this also related to #6806? Consider linking that one in your commit message as well, if so.
> 
> Please squash to taste, leave off #s, and we'll take one last look.

Related yes, in that instead of allowing multiple search patterns via repeating -I or -P, we are instead giving the user the ability to search using extended regex  syntax (`-x`), which includes the or (`|`) operator.

scripts/irods/test/test_itree.py
```diff
@@ -46,16 +46,45 @@ def test_depth(self):
...
+        # Using forward and reverse combination of -P and -I options, catch 'bat*' objects as result
+        self.admin.assert_icommand(["itree", "-P", "ba*", "-I", "*ar?"], "STDOUT_MULTILINE", to_catch, use_regex = True)
+        self.admin.assert_icommand(["itree", "-I", "*ar?", "-P", "ba*"], "STDOUT_MULTILINE", to_catch, use_regex = True)
+
+        # Using regular expressions, catch objects matching /bat[0-9]/
+        self.admin.assert_icommand(["itree", "--ignore-regex", ".*ar.", "--pattern-regex", "ba.*"], "STDOUT_MULTILINE", to_catch, use_regex = True)
+
+        # Using regular expressions with extended syntax, catch objects matching /(bar|bat)[0-2]/x
+        to_catch = [r'\b{}\b'.format(prefix+str(i)) for i in range(3) for prefix in ('bar','bat')]
+        self.admin.assert_icommand(["itree",  "--regex-extended", "--ignore-regex", ".*[3-9]", "--pattern-regex", "ba(r|t)[[:digit:]]"],
```
alanking, line 66: Codacy claims... trailing whitespace...

alanking:
Okay. I think that we should link both issues to the commit since both issue numbers are also being referenced in the name of the test.

scripts/irods/test/test_itree.py
d-w-moore: whitespace removed.

d-w-moore:
Ok - squashed with all the appropriate issue numbers applied.

trel:
looking very nice.

korydraughn (approved):
Looks good to me.

Add the pound when ready.

alanking (approved):

alanking (approved):

Issue #6820, opened 2023-01-04
Remove unnecessary preprocessor logic around `serverExit()` signal handler


korydraughn:
- [x] main
- [x] 4-3-stable
---
## Refactor
iRODS is designed to run on Linux systems, therefore, the preprocessor logic below is not needed.
https://github.com/irods/irods/blob/ec93f905d05776e67870ff80a92aba4227b926bd/server/main_server/src/rodsServer.cpp#L1516-L1533

korydraughn:
The preprocessor logic for OSX isn't needed and should be removed before closing this issue.
https://github.com/irods/irods/blob/36b22f0649932db8f4e331b619011c436df40746/server/main_server/src/rodsServer.cpp#L499-L507

PR #6823, opened 2023-01-04
[#6820] Removed preprocessor logic around signal handler (main)

korydraughn:


alanking (approved):

korydraughn:
Everything compiles, starts, and shuts down without any issues.

No need to run tests for this change since it removes dead code.

Added #

alanking (approved):

Issue #6842, opened 2023-01-12
Test failure: `test_federation.Test_ICommands.test_icp`


alanking:
- [x] main
- [x] 4-3-stable
- [x] 4-2-stable
---
Commit 99eab888f79f99aeb4560f76585b152a2165b36f causes copies across zones to produce an error message because the copy API is attempting to modify the collection mtime over the federation using the new client connection of the local zone rodsadmin. This causes `test_federation.Test_ICommands.test_icp` to fail.

If I revert that commit, the test passes again. So, we need to only use the privileged client connection when redirecting within a zone. A zone hint check similar to those found in `rsDataObjPut` and `rsDataObjOpen` should be used to determine which connection method to use:
https://github.com/irods/irods/blob/7a8ea89af09a3ab222f8009825dda0b207328735/server/api/src/rsDataObjPut.cpp#L685-L694

In addition to this, the copy API should not be updating the collection mtime at all. The open API already updates the collection mtime for the destination data object, except in the case of a cross-federation open. This means that the collection mtime is already being updated by the open API in the remote zone, so the open API call from the local zone does not need to reach out and update it, and neither does the copy API.

At the time of writing, the linked commit has not been merged into 4-2-stable, but the branch will require a similar fix, so I will put this in the 4.2.12 milestone.

trel:
nice find and proposed fix.

Issue #6527, opened 2022-08-11
Adjust clang-format and clang-tidy configuration


SwooshyCueb:
- [ ] main
- [ ] 4-3-stable
----
After going through the documentation for `clang-tidy` and `clang-format` and discussing it with @korydraughn, I put together [a document](https://gist.github.com/SwooshyCueb/bfea07c5176f74378f87821f1ba38392) going over some changes I think we need to make to our configurations. Opinions are welcome.

I will begin submitting PRs tomorrow/Monday to adjust our configurations.

korydraughn:
After running experiments based on the changes presented, here are my thoughts.

---

The comments that follow are specific to clang-format. I have not reviewed the clang-tidy changes, but they seem reasonable.

### Goals of the formatting style
- Maximum readability
- Formatting should help guide the reader's eyes
- Code should appear as if one person wrote it

### Response to changes

#### Tabs vs Spaces
- After seeing how the different options for `UseTab` work, the best setting seems to be `ForIndentation`
- `ForIndentation` resulted in the code always being formatted correctly regardless of the tab size. Things were aligned correctly using spaces
    - This option appears to align perfectly with Markus' opinion
- All tools I used (e.g. tig, github code viewer) produced the expected results
- I'm a spaces man, but I can also understand the case for developers that rely on visual aids

#### Access Specifiers (e.g. `public`, `protected`, and `private`)
- I'm not a fan of indenting these, but if the team likes it, I think one or two spaces for indentation is okay
- The indentation of the content for each access specifier should not be affected by this change
- Testing out the changes resulted in the contents of the access level being indented even more which is very weird to see
- We can use the `AccessModifierOffset` to control indentation (e.g. `-3` for one space indentation based on current config)
    - This may depend on the settings for indentation width, etc
- Things to think about:
    - Is code folding out-of-scope for the formatter?
        - I don't think this is something the formatter should be concerned with. Vim supports various different schemes around folding and even allows custom folding strategies. I imagine other editors provide similar controls
        - I feel this is a situation where IDEs and editors can be improved
        - I'm willing to be convinced on this topic though

#### Argument Packing
- I think setting `AllowAllArgumentsOnNextLine` to `true` is okay
- I'm not sure about changing `BinPackArguments` to `true`
    - For me, this depends on the context and I'm not sure if clang-format can be configured to consider the context
    - This may also indicate that the design of an interface needs some work (i.e. use a `struct` as the input type)
- I'm willing to be convinced

[...]

PR #6844, opened 2023-01-13
Fix federation test failure, avoid collection mtime update in copy, adjust clang-tidy (main)

alanking:


alanking:
Once again, I didn't mean to open this here... Converting to draft. Apologies.

alanking:
Addresses #6842
In service of #6527
Related to #6840 

After much struggling, I managed to make clang-tidy happy. The bulk of those changes have been kept in a separate commit to highlight the salient changes for fixing #6842. Also had to adjust the clang-tidy workflow such that warnings do not cause a failure.

Tests are running now.

server/api/src/rsDataObjCopy.cpp
```diff
@@ -152,15 +161,49 @@ namespace {
...
+        log_api::trace("[{}:{}] - closing [{}]", __func__, __LINE__, path);
+
+        openedDataObjInp_t close_inp{};
+        close_inp.l1descInx = _l1_index;
 
-        rodsLog(LOG_DEBUG8, "[%s:%d] - closing [%s]", __FUNCTION__, __LINE__, L1desc[_inx].dataObjInp->objPath);
+        // Declare this here because we need to return an error if the source index is invalid.
+        int ec = 0;
+
+        // The L1 descriptor index is const and confirmed to be in-bounds above. Therefore, ignore linter.
```
alanking, line 176: This isn't quite true anymore. The value is indeed `const`, but the check occurs outside of this function. Given that the function is local to this file, this is sufficient. I will update the comment.

server/api/src/rsDataObjCopy.cpp
```diff
@@ -152,15 +161,49 @@ namespace {
...
+        else {
+            // The L1 descriptor index is const and confirmed to be in-bounds above. Therefore, ignore linter.
+            //
+            // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-constant-array-index)
+            close_inp.bytesWritten = L1desc[source_l1_index].dataObjInfo->dataSize;
+        }
+
+        if (const int close_ec = rsDataObjClose(_comm, &close_inp); ec >= 0) {
+            ec = close_ec;
+        }
```
alanking, line 199: Wasn't sure whether the error code coming back from `rsDataObjClose` is more important than the source L1 descriptor being out of range.

Actually, looking at this again, we should probably set `L1desc[_l1_index].oprStatus` to `SYS_FILE_DESC_OUT_OF_RANGE` above so that close doesn't mark this replica as good.

korydraughn: Does anything care about `.oprStatus`?

alanking: `rsDataObjClose` uses `oprStatus` to determine whether to execute static post-PEPs and how to finalize data objects (i.e. if it's < 0, it can be marked stale). Seeing as `rsDataObjClose` is being used to close the two data objects in this API, that's why I feel it might be the right thing to do. Thoughts?

korydraughn: Sounds reasonable.

alanking: Done

alanking:
Encountered an issue due to the clang-tidy changes, but these have been resolved in the latest commit. Running tests one last time, barring other review comments.

alanking:
Core tests and federation tests all passed except `test_misc` because I forgot to rebuild my development environment to build the new test microservice plugins, so I'm doing that now. I do not anticipate any failures there.

server/api/src/rsDataObjCopy.cpp
```diff
@@ -195,13 +240,14 @@ namespace {
...
+        rodsServerHost_t* rodsServerHost = nullptr;
         int remoteFlag = connect_to_remote_zone( rsComm, dataObjCopyInp, &rodsServerHost );
-        if ( remoteFlag < 0 ) {
+        if (remoteFlag < 0) {
             return remoteFlag;
         }
         else if ( remoteFlag == REMOTE_HOST ) {
-            return _rcDataObjCopy(rodsServerHost->conn, dataObjCopyInp, transStat);
+            return !rodsServerHost ? NO_CATALOG_SERVICE_PROVIDER
+                                   : _rcDataObjCopy(rodsServerHost->conn, dataObjCopyInp, transStat);
```
korydraughn, line 250: Is it possible for `rodsServerHost` to be null if we're in this branch?

server/api/src/rsModColl.cpp
```diff
@@ -73,8 +74,24 @@ namespace
...
+    }
 
     try {
+        // If the zone name cannot be derived from the collection name in the input, the path cannot be valid.
+        const auto zone_name = fs::zone_name(fs::path{static_cast<char*>(modCollInp->collName)});
+        if (!zone_name) {
+            return SYS_INVALID_FILE_PATH;
+        }
+
+        if (!isLocalZone(zone_name->data())) {
```
korydraughn, line 90: Let's use `c_str()` instead of `data()`.

korydraughn (commented):
Looks good. Once the following comments are resolved, we can go ahead and get this merged (assuming tests pass).

server/api/src/rsDataObjCopy.cpp
alanking: If `getAndConnRemoteZone` is able to return `REMOTE_HOST` and not set the passed in `rodsServerHost_t**`, then yes. This was pointed out by clang-tidy.

server/api/src/rsModColl.cpp
alanking: Done

server/api/src/rsDataObjCopy.cpp
korydraughn: Carry on.

alanking: After a brief dive, it seems like it may not be possible to get here. If the pointer assigned to the `rodsServerHost` in `getRcatHost` is `nullptr`, a `nullptr` dereference would occur on the very next line:
https://github.com/irods/irods/blob/7a8ea89af09a3ab222f8009825dda0b207328735/server/core/src/rodsConnect.cpp#L98-L105
In any other scenario, the function will return a value that is not `REMOTE_HOST`.

Thoughts?

korydraughn: I feel we can assume that the pointer will be valid if `REMOTE_HOST` is returned.

I'm also fine with leaving the code as is. The compiler can probably optimize the nullptr check out.

Up to you.

alanking: Okay, here's the path clang-tidy is taking:
```
/home/runner/work/irods/irods/server/api/src/rsDataObjCopy.cpp:49:13: note: Assuming 'status' is >= 0
        if (status < 0 || !srcIcatServerHost) {
            ^~~~~~~~~~
/home/runner/work/irods/irods/server/api/src/rsDataObjCopy.cpp:49:13: note: Left side of '||' is false
/home/runner/work/irods/irods/server/api/src/rsDataObjCopy.cpp:49:27: note: Assuming 'srcIcatServerHost' is null
        if (status < 0 || !srcIcatServerHost) {
                          ^~~~~~~~~~~~~~~~~~
/home/runner/work/irods/irods/server/api/src/rsDataObjCopy.cpp:49:9: note: Taking true branch
        if (status < 0 || !srcIcatServerHost) {
        ^
/home/runner/work/irods/irods/server/api/src/rsDataObjCopy.cpp:53:13: note: Returning without writing to '*rodsServerHost'
            return status;
            ^
/home/runner/work/irods/irods/server/api/src/rsDataObjCopy.cpp:244:26: note: Returning from 'connect_to_remote_zone'
        int remoteFlag = connect_to_remote_zone( rsComm, dataObjCopyInp, &rodsServerHost );
                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/runner/work/irods/irods/server/api/src/rsDataObjCopy.cpp:245:13: note: 'remoteFlag' is >= 0
        if (remoteFlag < 0) {
            ^~~~~~~~~~
/home/runner/work/irods/irods/server/api/src/rsDataObjCopy.cpp:245:9: note: Taking false branch
        if (remoteFlag < 0) {
        ^
/home/runner/work/irods/irods/server/api/src/rsDataObjCopy.cpp:248:19: note: Assuming 'remoteFlag' is equal to REMOTE_HOST
        else if ( remoteFlag == REMOTE_HOST ) {
                  ^~~~~~~~~~~~~~~~~~~~~~~~~
```
The only way "'srcIcatServerHost' is null" can be true is if `getRcatHost` (called on line 47) returns a value < 0 (see code cited here, again):
https://github.com/irods/irods/blob/7a8ea89af09a3ab222f8009825dda0b207328735/server/core/src/rodsConnect.cpp#L93-L105
If this were not the case, a `nullptr` dereference would occur on line 100 or line 104 in the rodsConnect.cpp code. That is a problem on its own, sure, but it's not going to affect the code in this change. I will revert the change and leave a comment explaining the lack of check, just in case.

korydraughn: Cool. I fine with keeping/removing the check. Whichever you feel is good.

alanking: Done

alanking:
Builds are failing due to #6852 so I am ignoring those. See https://github.com/irods/irods/pull/6844#discussion_r1081727761 for why clang-tidy is upset

server/api/src/rsDataObjCopy.cpp
```diff
@@ -195,12 +240,13 @@ namespace {
...
 
-        rodsServerHost_t *rodsServerHost;
+        rodsServerHost_t* rodsServerHost = nullptr;
         int remoteFlag = connect_to_remote_zone( rsComm, dataObjCopyInp, &rodsServerHost );
-        if ( remoteFlag < 0 ) {
+        if (remoteFlag < 0) {
             return remoteFlag;
         }
         else if ( remoteFlag == REMOTE_HOST ) {
+            // It is not possible reach this case with rodsServerHost being nullptr, so no check is needed.
```
korydraughn, line 249: "possible reach" -> "possible to reach"

korydraughn (commented):
I think we can go ahead and squash this (without the pounds) given the latest changes do not change the behavior.

server/api/src/rsDataObjCopy.cpp
alanking: D'oh

lib/core/include/irods/rodsErrorTable.h
```diff
@@ -240,6 +240,7 @@ NEW_ERROR(NOT_A_DATA_OBJECT,                           -171000)
 NEW_ERROR(JSON_VALIDATION_ERROR,                       -172000)
 NEW_ERROR(DATABASE_TYPE_NOT_SUPPORTED,                 -173000)
 NEW_ERROR(RESOLUTION_ERROR,                            -174000)
+NEW_ERROR(NO_CATALOG_SERVICE_PROVIDER,                 -175000)
```
alanking, line 243: Also, this can be removed. I will include that with the #s

alanking:
All tests are passing

lib/core/include/irods/rodsErrorTable.h
korydraughn: Are we keeping this?

korydraughn: Nevermind. I misread your comment.

korydraughn (dismissed):
Add the pounds and merge when ready.

lib/core/include/irods/rodsErrorTable.h
alanking: Done.

korydraughn (approved):

alanking:
Whoops, okay, added #s lol

korydraughn (approved):

korydraughn:
I kept seeing my approval get dismissed lol.

Anyway, merge when ready.

PR #6853, opened 2023-01-17
[4063] Ensure AVUs set properly (4-2-stable)

MartinFlores751:
Separate collections with similar AVUs should be able to be set their AVUs without errors.

### DIFF
--- a/plugins/database/src/db_plugin.cpp
+++ b/plugins/database/src/db_plugin.cpp
@@ -155,6 +155,18 @@ validateAndParseUserName( const char *fullUserNameIn, char *userName, char *user
     return 0;
 } // validateAndParseUserName
 
+[[nodiscard]] auto is_valid_avu(const char *_attribute, const char *_value, const char *_unit) noexcept -> bool
+{
+    if (_attribute == nullptr || _value == nullptr || _unit == nullptr) {
+        return false;
+    }
+    if (*_attribute == '\0' || *_value == '\0') {
+        return false;
+    }
+
+    return true;
+}
+
 // =-=-=-=-=-=-=-
 // helper fcn to handle cast to pg object
 irods::error make_db_ptr(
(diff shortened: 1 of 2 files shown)


scripts/irods/test/test_imeta_set.py
```diff
@@ -355,6 +355,26 @@ def test_imeta_handles_trailing_slashes_on_collections__issue_4559(self):
...
+        for collection in collections:
+            self.admin.assert_icommand(['imkdir', collection])
+
+        # Set the same AVUs for the collections
+        for collection in collections:
+            self.admin.assert_icommand(['imeta', 'set', '-C', collection, 'is_cool', 'yes'])
+
+        # Clear Value of second collection
+        out, err, ec = self.admin.run_icommand(['imeta', 'set', '-C', collections[1], 'is_cool', "''"])
+
```
trel, line 373: Please also assert that the AVUs on both collections are what you want them to be.

plugins/database/src/db_plugin.cpp
```diff
@@ -155,6 +155,17 @@ validateAndParseUserName( const char *fullUserNameIn, char *userName, char *user
     return 0;
 } // validateAndParseUserName
 
+[[nodiscard]] auto is_valid_avu(const char *_attribute, const char *_value,
+                                const char *_unit) noexcept -> bool {
+    if (_attribute == nullptr || _value == nullptr || _unit == nullptr)
+        return false;
+
+    if (*_attribute == '\0' || *_value == '\0')
+        return false;
```
alanking, line 164: Maybe some curly braces?

plugins/database/src/db_plugin.cpp
```diff
@@ -155,6 +155,17 @@ validateAndParseUserName( const char *fullUserNameIn, char *userName, char *user
     return 0;
 } // validateAndParseUserName
 
+[[nodiscard]] auto is_valid_avu(const char *_attribute, const char *_value,
+                                const char *_unit) noexcept -> bool {
```
alanking, line 159:
```suggestion
[[nodiscard]] auto is_valid_avu(const char* _attribute, const char* _value, const char* _unit) noexcept -> bool
{
```

alanking (commented):
4-2-stable doesn't have a .clang-format file, so it may have led you astray here.

alanking (commented):
This looks good to me. Let's go ahead and squash but leave off the #s

MartinFlores751:
@alanking Take a peek!

alanking (approved):
Looks good to me. Please add # after 1 more approval (just in case)

korydraughn (approved):
Please add the pound.

PR #6858, opened 2023-01-18
[#6482] Deleting a delay rule now verifies zone of the connected client (main)

korydraughn:
Ported from #6805

Running unit tests.

Will run federation tests and core tests later.


korydraughn:
All unit tests passed.

Running federation tests.

korydraughn:
Federation tests that appear to be unrelated to my changes are failing.

Perhaps related to #6844?

alanking:
Indeed

korydraughn:
Rebased.

All unit tests passed.

Running federation tests.

korydraughn:
All federation tests passed.

Running core tests.

scripts/irods/test/session.py
```diff
@@ -124,6 +124,10 @@ def zone_name(self):
     def username(self):
         return self._environment_file_contents['irods_user_name']
 
+    @property
+    def qualified_username(self):
+        return self.username + '#' + self.zone_name
```
korydraughn, line 129: This breaks two tests in test_auth.py. The tests modify environment properties which cause `self.zone_name` to throw an exception.

My proposed fix is to check if `irods_zone_name` is defined in  `self.environment_file_contents` first. If it is defined, use it. Else, return the username with a suffix of `#?`. _(I've already verified that this fixes the tests)_

Other ideas:
- return just the username
- `#<unknown>`
- `#<not-defined>`
- `#???`

Thoughts?

trel: not exactly sure what is breaking - but we always are operating in the context of a zone.

if one is not specified, shouldn't we always assume and use the local zonename?  of the server?

korydraughn: This is specific to the test framework. The icommands/server are working as expected.

For example, one of the tests deletes the zone property from the python dict to verify iinit can handle it gracefully. When iinit is executed, the test framework attempts to log a line containing the command to execute. At this point, it falls over because it can't produce the qualified username. It can't produce the qualified username because the zone property was removed from the python dict.

trel: so we should log what would be used... yes?    so i think that's a vote for `just the username` or `#<not-defined>`.

since it's just the test framework, `#<not-defined>` seems good?

korydraughn: I agree. I went with `#<undefined>`.

It is just an indicator that the zone was not there.

korydraughn:
Forgot to mention that all core tests passed before the rebase.

korydraughn:
Will rerun the tests to be sure everything is good.

korydraughn:
All unit, core, and federation tests passed.

alanking (approved):
Some nitpicks, which may or may not have already been merged into 4-2-stable, so they can be ignored

server/delay_server/src/irodsDelayServer.cpp
```diff
@@ -248,25 +239,33 @@ namespace
...
-        rstrcpy(rule_exec_submit_inp.exeAddress, exec_info[3].c_str(), NAME_LEN);
-        rstrcpy(rule_exec_submit_inp.exeTime, exec_info[4].c_str(), TIME_LEN);
-        rstrcpy(rule_exec_submit_inp.exeFrequency, exec_info[5].c_str(), NAME_LEN);
-        rstrcpy(rule_exec_submit_inp.priority, exec_info[6].c_str(), NAME_LEN);
-        rstrcpy(rule_exec_submit_inp.lastExecTime, exec_info[7].c_str(), NAME_LEN);
-        rstrcpy(rule_exec_submit_inp.exeStatus, exec_info[8].c_str(), NAME_LEN);
-        rstrcpy(rule_exec_submit_inp.estimateExeTime, exec_info[9].c_str(), NAME_LEN);
-        rstrcpy(rule_exec_submit_inp.notificationAddr, exec_info[10].c_str(), NAME_LEN);
+        // clang-format off
+        const auto to_cstr = []<typename T>(T && _s) constexpr noexcept -> char*
```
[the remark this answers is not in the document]
korydraughn: Oh no. Clang-format tried to sneak that past me.

Will change.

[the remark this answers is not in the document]
korydraughn: Done.

server/icat/include/irods/icatHighLevelRoutines.hpp
```diff
@@ -282,4 +282,19 @@ auto chl_check_permission_to_modify_data_object(RsComm& _comm, const rodsLong_t
 /// \since 4.2.9
 auto chl_update_ticket_write_byte_count(RsComm& _comm, const rodsLong_t _data_id, const rodsLong_t _bytes_written) -> int;
 
+/// \brief High-level wrapper for fetching the all information about a delay rule.
```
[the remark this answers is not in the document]
korydraughn: Done.

PR #6877, opened 2023-01-26
[6743] nre.reLib1.cpp:extractVarNames: do not write past the end of varNames (main)

SwooshyCueb:
Addresses 6743 for main

### DIFF
--- a/plugins/rule_engines/irods_rule_language/src/nre.reLib1.cpp
+++ b/plugins/rule_engines/irods_rule_language/src/nre.reLib1.cpp
@@ -274,7 +274,7 @@ int extractVarNames( char **varNames, const char *outBuf ) {
     const char *p = outBuf;
     const char *psrc = p;
 
-    for ( ;; ) {
+    while (n < MAX_PARAMS_LEN) {
         if ( *psrc == '%' ) {
             varNames[n++] = strndup( p, psrc - p );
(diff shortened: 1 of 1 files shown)


alanking (commented):
Please apply clang-format changes and I think this is good to go.

Just asking... probably not worth a test, right?

korydraughn:
Because we're touching the rule language, let's run the change through the unit tests and core tests.

You can never be too sure what will happen once we start touching things down there.

alanking:
Right, sorry, I meant to say a *new* test. Yes, we should run this through the test suite(s) before merging. :)

SwooshyCueb:
I'm not currently set up to run the test suite. Will wait for results from someone else

korydraughn:
Let's go ahead and get the testing environment set up so you can run the tests.

You can use the test VMs if you don't want to run the tests locally. I don't think anyone is using them at the moment.

SwooshyCueb:
Yes, I'll need to use the VMs until the new workstation comes in. I'll poke you about it tomorrow.

SwooshyCueb:
Ran core and unit tests, all passed.

korydraughn (approved):
Please add the pound.

alanking (approved):

SwooshyCueb:
\#'d

Issue #6882, opened 2023-01-30
Add more details to `key_value_proxy` error messages


alanking:
- [x] main
- [x] 4-3-stable
- [x] 4-2-stable

## Feature

Seeing "key not found" when a key is not found in the `key_value_proxy` is not particularly helpful if the missing key is not shown in the error message. Let's add more information to the error message to help with troubleshooting.

Issue #6880, opened 2021-06-08
State of replica after itouch to S3 is 'X'


JustinKyleJames:
* [x]  main
* [x] 4-3-stable
* [x]  4-2-stable
-----

The state of the replica after itouch to S3 is 'X' and the data object remains locked.

The object is created successfully in S3 and there are no S3 plugin type errors generated.  (Also zero length puts work.)  The following error is in the logs.

ERROR: [finalize_replica_opened_for_create_or_write:75] - no entry found in replica state table for [/tempZone/home/rods/zerofile]

The replica status is good after itouch to a unixfilesystem resource so this has something to do with the interplay between irods and the S3 plugin.

Issue #6479, opened 2022-07-19
itouch non existent file creates undeletable object


darthtux:
- [x] main
- [x] 4-3-stable
- [x] 4-2-stable
---
I am running irods 4.3.0 with icommand 4.3.0
I tried running itouch on a file in irods that does not exist.  Doing so results in a file I have been unable to delete

```
[root@elite ~]# itouch test 
Level 0: SYS_NO_GOOD_REPLICA: No good replicas found.
```
```
-bash-4.2$ ils -L
/msi_test/projects/aso:
  emunsell          0 s3resc            0 2022-07-19.10:08 ? test
        generic    /msi_irods_tier2_poc/msi_test/projects/aso/test
```
As an administrator tried irm
```
-bash-4.2$ irm test
remote addresses: 10.31.6.36 ERROR: rmUtil: rm error for /msi_test/projects/aso/test, status = -1803000 status = -1803000 HIERARCHY_ERROR
```
this was logged on the server
```
Jul 19 10:45:15 elite irodsServer[24375]: {"log_category":"legacy","log_facility":"local0","log_level":"info","log_message":"[rsDataObjUnlink_impl:334] - [HIERARCHY_ERROR: no valid resource found for data object\n\n]","request_api_name":"DATA_OBJ_UNLINK_AN","request_api_number":615,"request_api_version":"d","request_client_user":"rods","request_host":"10.31.6.36","request_proxy_user":"rods","request_release_version":"rods4.3.0","server_host":"elite.msi.umn.edu","server_pid":24375,"server_timestamp":"2022-07-19T15:45:15.985Z","server_type":"agent"}
```
still as administrator irm -f
```
-bash-4.2$ irm -f test
remote addresses: 10.31.6.36 ERROR: rmUtil: rm error for /msi_test/projects/aso/test, status = -406000 status = -406000 LOCKED_DATA_OBJECT_ACCESS
```
this was logged on the server
```
[...]

Issue #6378, opened 2022-05-06
Data object stuck in locked/intermediate status when 'agent stop' network plugin operation fails


alanking:
- [x] main
- [x] 4-3-stable
- [x] 4-2-stable
---
On 4.2.11...

See https://groups.google.com/g/irod-chat/c/Aynp3qHPj4Q/m/c1wvVZR0CwAJ

When the 'agent stop' network plugin operation fails on agent teardown, the `cleanup()` step fails to occur (and some memory leaks as well), as shown here:
https://github.com/irods/irods/blob/649e2bef613a5c32bccc7cad2077887e968b1ba2/server/core/src/rodsAgent.cpp#L582-L593

We should be `free`ing the memory and calling `cleanup()` no matter what to ensure that no memory leaks and data objects do not get stuck in the locked/intermediate status.

PR #6884, opened 2023-01-30
Ensure that data objects cannot get stuck in locked status (main)

alanking:
Addresses #6154
Addresses #6378
Addresses #6479
Addresses #6880
Addresses #6882

Companion PR: https://github.com/irods/irods_resource_plugin_s3/pull/2094

The issues being addressed here are difficult to reproduce in the automated tests. I have some machinery for this in a separate branch which we can consider merging at another time. Tests for these things are in the companion PR.

Will run core tests soon, hence the draft PR.

### DIFF
--- a/plugins/api/src/replica_close.cpp
+++ b/plugins/api/src/replica_close.cpp
@@ -57,7 +57,6 @@
 namespace
 {
     // clang-format off
-    namespace ix  = irods::experimental;
     namespace fs  = irods::experimental::filesystem;
     namespace ill = irods::logical_locking;
     namespace ir  = irods::experimental::replica;
@@ -87,7 +86,8 @@ namespace
     auto unlock_and_publish_replica(rsComm_t& _comm,
                                     const ir::replica_proxy_t& _replica,
                                     const l1desc& _l1desc,
-                                    const bool _send_notifications) -> int;
+                                    const bool _send_notifications,
+                                    const int _unlock_statuses = ill::restore_status) -> int;
 
     auto update_replica_size_and_status(rsComm_t& _comm,
                                         const l1desc_t& _l1desc,
@@ -100,7 +100,10 @@ namespace
     auto update_replica_status(rsComm_t& _comm,
                                const l1desc_t& _l1desc,
                                const repl_status_t _new_status,
-                               const bool _send_notifications) -> int;
+                               const bool _send_notifications,
+                               const int _unlock_statuses = ill::restore_status) -> int;
+
+    auto update_replica_status_on_error(rsComm_t& _comm, const l1desc_t& _l1desc) -> int;
 
     auto free_l1_descriptor(int _l1desc_index) -> int;
 
@@ -159,18 +162,17 @@ namespace
         return rsFileClose(&_comm, &input);
     }
 
-    auto unlock_and_publish_replica(
-        rsComm_t& _comm,
-        const ir::replica_proxy_t& _replica,
-        const l1desc& _l1desc,
-        const bool _send_notifications) -> int
+    auto unlock_and_publish_replica(rsComm_t& _comm,
+                                    const ir::replica_proxy_t& _replica,
+                                    const l1desc& _l1desc,
+                                    const bool _send_notifications,
+                                    const int _unlock_statuses) -> int
     {
-        // The sibling replicas are always marked stale because replica_close considers itself to be the new
-        // truth when it is given permission to finalize the replica.
-        constexpr auto unlock_statuses = STALE_REPLICA;
-
         // Unlock the data object but do not publish to catalog because we may want to trigger file_modified.
-        if (const int ec = ill::unlock(_replica.data_id(), _replica.replica_number(), _replica.replica_status(), unlock_statuses); ec < 0) {
+        if (const int ec =
+                ill::unlock(_replica.data_id(), _replica.replica_number(), _replica.replica_status(), _unlock_statuses);
+            ec < 0)
+        {
             irods::log(LOG_ERROR, fmt::format(
                 "[{}:{}] - failed to unlock object [{}]:[{}]",
                 __FUNCTION__, __LINE__, _replica.logical_path(), ec));
@@ -245,7 +247,7 @@ namespace
         // Update the RST with the changes made above to the replica.
         rst::update(repl.data_id(), repl);
 
-        return unlock_and_publish_replica(_comm, repl, _l1desc, _send_notifications);
+        return unlock_and_publish_replica(_comm, repl, _l1desc, _send_notifications, STALE_REPLICA);
     } // update_replica_size_and_status
 
     auto update_replica_size(rsComm_t& _comm, const l1desc_t& _l1desc, const bool _send_notifications) -> int
@@ -275,22 +277,31 @@ namespace
         // Update the RST with the changes made above to the replica.
         rst::update(repl.data_id(), repl);
 
-        return unlock_and_publish_replica(_comm, repl, _l1desc, _send_notifications);
+        return unlock_and_publish_replica(_comm, repl, _l1desc, _send_notifications, STALE_REPLICA);
     } // update_replica_size
 
     auto update_replica_status(rsComm_t& _comm,
                                const l1desc_t& _l1desc,
                                const repl_status_t _new_status,
-                               const bool _send_notifications) -> int
+                               const bool _send_notifications,
+                               const int _unlock_statuses) -> int
     {
-        const auto repl = ir::make_replica_proxy(*_l1desc.dataObjInfo);
+        auto repl = ir::make_replica_proxy(*_l1desc.dataObjInfo);
+
+        repl.replica_status(_new_status);
 
         // Set target replica status in RST as logical locking will not set
         // the status of the target replica.
-        rst::update(repl.data_id(), repl.replica_number(), json{{"data_is_dirty", std::to_string(_new_status)}});
+        rst::update(
+            repl.data_id(), repl.replica_number(), json{{"data_is_dirty", std::to_string(repl.replica_status())}});
 
-        return unlock_and_publish_replica(_comm, repl, _l1desc, _send_notifications);
+        return unlock_and_publish_replica(_comm, repl, _l1desc, _send_notifications, _unlock_statuses);
     } // update_replica_status
+
+    auto update_replica_status_on_error(rsComm_t& _comm, const l1desc_t& _l1desc) -> int
+    {
+        return update_replica_status(_comm, _l1desc, STALE_REPLICA, false);
+    } // update_replica_status_on_error
 
     auto free_l1_descriptor(int _l1desc_index) -> int
     {
@@ -380,14 +391,14 @@ namespace
                 if (update_size && update_status) {
                     if (const auto ec = update_replica_size_and_status(*_comm, l1desc, send_notifications); ec != 0) {
                         log::api::error("Failed to update the replica size and status in the catalog [error_code={}].", ec);
-                        update_replica_status(*_comm, l1desc, STALE_REPLICA, false);
+                        update_replica_status_on_error(*_comm, l1desc);
                         return ec;
                     }
                 }
                 else if (update_size) {
                     if (const auto ec = update_replica_size(*_comm, l1desc, send_notifications); ec != 0) {
                         log::api::error("Failed to update the replica size in the catalog [error_code={}].", ec);
-                        update_replica_status(*_comm, l1desc, STALE_REPLICA, false);
+                        update_replica_status_on_error(*_comm, l1desc);
                         return ec;
                     }
                 }
@@ -410,9 +421,11 @@ namespace
                         new_status = ill::get_original_replica_status(l1desc.dataObjInfo->dataId, l1desc.dataObjInfo->replNum);
                     }
 
-                    if (const auto ec = update_replica_status(*_comm, l1desc, new_status, send_notifications); ec != 0) {
+                    if (const auto ec =
+                            update_replica_status(*_comm, l1desc, new_status, send_notifications, STALE_REPLICA);
+                        ec != 0) {
                         log::api::error("Failed to update the replica status in the catalog [error_code={}].", ec);
-                        update_replica_status(*_comm, l1desc, STALE_REPLICA, false);
+                        update_replica_status_on_error(*_comm, l1desc);
                         return ec;
                     }
                 }
@@ -444,30 +457,30 @@ namespace
             const auto ec = free_l1_descriptor(l1desc_index);
 
             if (ec != 0) {
-                update_replica_status(*_comm, l1desc, STALE_REPLICA, false);
+                update_replica_status_on_error(*_comm, l1desc);
             }
 
             return ec;
         }
         catch (const json::type_error& e) {
             log::api::error("Failed to extract property from JSON object [error_code={}]", SYS_INTERNAL_ERR);
-            update_replica_status(*_comm, l1desc, STALE_REPLICA, false);
+            update_replica_status_on_error(*_comm, l1desc);
             return SYS_INTERNAL_ERR;
         }
         catch (const irods::exception& e) {
             log::api::error("{} [error_code={}]", e.what(), e.code());
-            update_replica_status(*_comm, l1desc, STALE_REPLICA, false);
+            update_replica_status_on_error(*_comm, l1desc);
             return e.code();
         }
         catch (const fs::filesystem_error& e) {
             log::api::error("{} [error_code={}]", e.what(), e.code().value());
-            update_replica_status(*_comm, l1desc, STALE_REPLICA, false);
+            update_replica_status_on_error(*_comm, l1desc);
             return e.code().value();
         }
         catch (const std::exception& e) {
             log::api::error("An unexpected error occurred while closing the replica. {} [error_code={}]",
                             e.what(), SYS_INTERNAL_ERR);
-            update_replica_status(*_comm, l1desc, STALE_REPLICA, false);
+            update_replica_status_on_error(*_comm, l1desc);
             return SYS_INTERNAL_ERR;
         }
     } // rs_replica_close
--- a/plugins/api/src/replica_close.cpp
+++ b/plugins/api/src/replica_close.cpp
@@ -157,10 +157,17 @@ namespace
 
     auto close_physical_object(rsComm_t& _comm, int _l3desc_index) -> int
     {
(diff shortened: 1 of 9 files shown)


plugins/api/src/replica_close.cpp
```diff
@@ -369,6 +380,16 @@ namespace
 
             const auto is_write_operation = (O_RDONLY != (l1desc.dataObjInp->openFlags & O_ACCMODE));
 
+            // Close the underlying file object.
+            if (const auto ec = close_physical_object(*_comm, l1desc.l3descInx); ec != 0) {
+                log::api::error("Failed to close file object [error_code={}].", ec);
+                if (is_write_operation) {
+                    update_replica_status_on_error(*_comm, l1desc);
+                    irods::experimental::replica_access_table::erase_pid(l1desc.replica_token, getpid());
```
alanking, line 388: I'm not sure that I understand why the replica access table entry was being restored in the old logic. Failing to remove the replica access table entry causes overwrites in S3 to fail because the PID of the new agent is not found in the entry. If the failure occurs in an iCommand, there's no way you will be able to open this object again because the connection closes and the agent goes away.

Would it be appropriate to have the agent teardown logic also call `erase_pid` just in case?

korydraughn:
Is this ready for review? If not, please request reviewers when ready.

alanking:
Not yet. Still working through a couple of things and will do.

alanking:
All tests passed. Open question in above self-review still stands.

clang-format is complaining about "cognitive complexity" of 3 functions that were touched by these changes. The changes made here did not push these functions over their respective thresholds because they were already well over in each case:
```
error: function 'rs_replica_close' has cognitive complexity of 51 (threshold 25) 
error: function 'repl_object' has cognitive complexity of 48 (threshold 25) 
error: function 'runIrodsAgentFactory' has cognitive complexity of 65 (threshold 25)
```

So, I'm ignoring clang-format.

plugins/api/src/replica_close.cpp
alanking: Okay, I see. In that case, I think only removing the PID on success makes sense. Once the client disconnects or the agent process is reaped on termination, the PID will be removed from the replica access table if it's not done so already in this function (L453). That way, the connected client can try as many times as needed on failure to close the replica and the PID will be guaranteed to be removed.

> Worst case scenario is the agent factory has to do it (which it does - you'll need to verify what happens if an unknown PID is removed). Does the replica access table throw an exception or ignore it?

I added an assertion to unit_tests/src/test_replica_access_table.cpp which calls `erase_pid` with a PID that does not exist in the replica access table and it did not throw:
```c++
// Erase a PID which does not exist in the replica access table.
REQUIRE_NOTHROW(rat::erase_pid(7000));
```
Is that what you meant by "unknown"?

korydraughn: > Is that what you meant by "unknown"?

Yes.

> Okay, I see. In that case, I think only removing the PID on success makes sense.

That sounds reasonable. What happens if the file on disk is closed and then we fail on a catalog update? Will `close_physical_object` succeed on a closed file or are we stuck?

How does this compare to `rsDataObjClose`? Does anything need to change there?
https://github.com/irods/irods/blob/03e7e0515ab3e367c969ddc783b0e0d3dd995924/server/api/src/rsDataObjClose.cpp#L732-L762

alanking: After some discussion, we have determined that just erasing the PID on success in `replica_close` is probably okay since the tests are passing and there are no anticipated issues with the approach.

If the file on disk is closed and we fail on a catalog update, we will attempt to unlock the data object in all cases (see `update_replica_status_on_error`), in which case the open replica will be stale with the other replica statuses being restored to what they were before they were locked. If this fails, the object is stuck in a locked status because we cannot reach the catalog. In that case, there's nothing we can do. If we want to allow the user to try again, we will need to take a long, hard look at `replica_close` and `rsDataObjClose` (and, by extension, their "open" counterparts). After discussion, we think that's beyond the scope of this change, but should happen at some point.

The logic shown in `rsDataObjClose` there is not going to restore the replica access table entry at all the right times (I think). I've created an issue to address this: https://github.com/irods/irods/issues/6890 I initially thought this code might be dead, but it's not *quite* dead. Again, this is beyond the scope of this change.

I will mark this as resolved and we can raise a new thread if similar concerns arise later.

scripts/irods/test/test_resource_types.py
```diff
@@ -2808,6 +2810,106 @@ def test_archive_only_replica_does_not_result_in_SYS_REPLICA_DOES_NOT_EXIST__iss
...
+            self.assertEqual(str(1), lib.get_replica_status(self.user0, os.path.basename(logical_path), 0))
+            self.assertEqual(str(0), lib.get_replica_status(self.user0, os.path.basename(logical_path), 1))
+
+            # debugging
+            self.user0.assert_icommand(['ils', '-L', os.path.dirname(logical_path)], 'STDOUT', filename)
+
+            # Try to clean it up, even though the archive replica physical path does not exist.
+            self.user0.assert_icommand(['irm', '-f', logical_path])
+            self.assertFalse(lib.replica_exists_on_resource(self.user0, logical_path, self.cache_resc))
+            self.assertFalse(lib.replica_exists_on_resource(self.user0, logical_path, self.archive_resc))
```
korydraughn, line 2906: For my own understanding ...

Does the replica on the archive resource still on disk?

Based on the test, my guess is yes because the service account user only has read permission on the archive vault. Which also means we are determining that it is okay to remove the replica entry from the catalog?

korydraughn (commented):
Just one question.

scripts/irods/test/test_resource_types.py
```diff
@@ -2808,6 +2810,106 @@ def test_archive_only_replica_does_not_result_in_SYS_REPLICA_DOES_NOT_EXIST__iss
...
+            self.assertTrue(lib.replica_exists_on_resource(self.user0, logical_path, self.archive_resc))
+            self.assertEqual(str(1), lib.get_replica_status(self.user0, os.path.basename(logical_path), 0))
+            self.assertEqual(str(1), lib.get_replica_status(self.user0, os.path.basename(logical_path), 1))
+
+        finally:
+            # Make sure the archive vault is put back to a workable permissions set.
+            os.chmod(archive_vault_path, 0o750)
+
+
+    def test_synctoarch_failure_returns_an_error_and_can_be_removed__issue_6886(self):
```
alanking, line 2866: This needs to be skipped in topology testing

scripts/irods/test/test_resource_types.py
```diff
@@ -2808,6 +2810,106 @@ def test_archive_only_replica_does_not_result_in_SYS_REPLICA_DOES_NOT_EXIST__iss
             self.admin.run_icommand(['iadmin', 'rmresc', cache_resc])
             self.admin.run_icommand(['iadmin', 'rmresc', comp_resc])
 
+
+    def test_synctoarch_failure_returns_an_error_and_can_be_overwritten__issue_6886(self):
```
alanking, line 2814: This needs to be skipped in topology testing

scripts/irods/test/test_resource_types.py
alanking: The replica exists in the catalog, but the physical data represented by the physical path for the replica does not actually exist in the vault for the archive resource. We could perform a check after the failed `iput` above, but I felt like it didn't really matter for this test. As you pointed out, the service account user has read permission and was able to determine that the data indeed does not exist, which is why it is okay to remove the replica entry from the catalog.

This was originally written to be an "equivalent" to the tests found in https://github.com/irods/irods_resource_plugin_s3/pull/2094 for the invalid S3 secret keys, but it turns out that it's not equivalent because the service account user can actually still see what is and is not there.

That being said, do you think it would be worth adding another test where the service account user does not even have read permission for the vault of the archive resource? That would be more equivalent to the invalid S3 keys scenario.

trel: I think I would like that - and we can leave big fat comments to this effect - and the equivalences that we understand/intend.

korydraughn: I forgot that the test `irm -f`'s the data object before changing the permissions on the vault. I'm back now. As you've pointed out, there's no replica on the archive resource therefore it is safe to remove the catalog entry.

> That being said, do you think it would be worth adding another test where the service account user does not even have read permission for the vault of the archive resource? That would be more equivalent to the invalid S3 keys scenario.

I think adding a test to verify we return a helpful error code would be good. That increases coverage and confidence. The other part of that is it's not on the iRODS system to fix that. The admin has to make sure their deployment can read/write the vaults it owns.

You agree?

alanking: Makes sense to me. I'll add the test

alanking: The error code returned when the permissions are set to `0o000` is the same as the one returned in the current test: `UNIX_FILE_OPEN_ERR`. Does this seem okay?

korydraughn: I think so. When the permissions are bad, we now know what error will be reported.

We can also move forward knowing that it will be harder to introduce a regression around bad permissions and vaults.

alanking: Well, to be clear, that is the error produced by sync-to-archive. If you try to `iput` directly to a resource with a vault on which the service account has no permissions, you get `UNIX_FILE_GET_FS_FREESPACE_ERR`.

By the way, it turns out that the minimum permissions required for `iput` and `iget` (only iCommands I tried) is `0o100`. The service account just needs execute permission in the vault directory.

scripts/irods/test/test_resource_types.py
alanking: Done

scripts/irods/test/test_resource_types.py
alanking: Done

alanking:
Only calling `erase_pid` on success now as previously discussed. 

Also making sure that we only update the replica status when `update_status` has been specified. I wonder if we should do that in other places throughout this function... Anyway, probably don't need to mark the replica as stale if freeing the L1 descriptor failed - doesn't really make sense.

Sorry about all the extra changes, things just keep... appearing.

scripts/irods/test/test_resource_types.py
korydraughn: I see. It's fine if the error code test is specific to sync-to-archive. At the very least, we've increased coverage and we have a test we can reference in the future.

> By the way, it turns out that the minimum permissions required for iput and iget (only iCommands I tried) is 0o100. The service account just needs execute permission in the vault directory.

Makes sense.

I think you've resolved the main issue. The tests will protect us from regressions.

alanking: Good enough for me :) Thanks!

korydraughn:
> Also making sure that we only update the replica status when update_status has been specified. I wonder if we should do that in other places throughout this function...

So we didn't cover all the cases in previous versions of this API plugin?

> Anyway, probably don't need to mark the replica as stale if freeing the L1 descriptor failed - doesn't really make sense.

Agreed.

alanking:
> So we didn't cover all the cases in previous versions of this API plugin?

Yes, it seems this way. For instance, if only `update_size` is passed in (note: NOT `update_status`), we still mark the object as stale if something goes wrong, as seen here:
https://github.com/irods/irods/blob/03e7e0515ab3e367c969ddc783b0e0d3dd995924/plugins/api/src/replica_close.cpp#L387-L393
The same is true for `compute_checksum`. I'm not really sure what the right answer is right now.

korydraughn:
Got it. That makes sense. I think you made the right move. Make it work as advertised.

korydraughn:

plugins/api/src/replica_close.cpp
```diff
@@ -380,23 +380,24 @@ namespace
 
             const auto is_write_operation = (O_RDONLY != (l1desc.dataObjInp->openFlags & O_ACCMODE));
 
+            const auto update_size = !json_input.contains("update_size") || json_input.at("update_size").get<bool>();
+            const auto update_status =
+                !json_input.contains("update_status") || json_input.at("update_status").get<bool>();
```
alanking, line 385: Move this outside of the try-catch block so that the catch blocks can determine whether to update the status on error

korydraughn:
I'll wait for your final tweak and round of testing before giving approval.

plugins/api/src/replica_close.cpp
alanking: Done

alanking:
Yep, pushed the tweak. Running tests now... Will re-run S3 plugin tests for irods/irods_resource_plugin_s3#2094 as well

plugins/api/src/replica_close.cpp
```diff
@@ -425,49 +450,41 @@ namespace
...
-
-            const auto ec = free_l1_descriptor(l1desc_index);
-
-            if (ec != 0) {
-                update_replica_status(*_comm, l1desc, STALE_REPLICA, false);
+            // Remove the agent's PID from the replica access table for this replica. The replica access table entry is
+            // only erased when everythng else is successful.
+            if (is_write_operation) {
+                irods::experimental::replica_access_table::erase_pid(l1desc.replica_token, getpid());
             }
```
korydraughn, line 457: Looks like this can be merged into the block above.

plugins/api/src/replica_close.cpp
```diff
@@ -367,27 +380,37 @@ namespace
...
+                json_input.contains("compute_checksum") && json_input.at("compute_checksum").get<bool>();
+
+            // Close the underlying file object.
+            if (const auto ec = close_physical_object(*_comm, l1desc.l3descInx); ec != 0) {
+                log::api::error("Failed to close file object [error_code={}].", ec);
+                if (is_write_operation && update_status) {
+                    update_replica_status_on_error(*_comm, l1desc);
+                }
+                return ec;
+            }
```
korydraughn, line 394: Say this succeeds, but then one of the catalog updates that follow fails.

If `replica_close` is invoked again to perform the catalog updates, will `close_physical_object` be a hard stop due to the file on disk being closed?

_I feel I may have asked this question before?_

alanking: I think it depends on what part of the process failed (e.g. Is the L3 descriptor still valid?) and the underlying storage technology. I haven't tried it yet because I would have to get my operation failure testing tool set up again. Do you want me to try and find out?

In any case, as we discussed, the user has explicitly chosen not to update the replica status and is knowingly attempting an advanced maneuver. Should this occur, I feel like all we can do here is bail with an error and, if applicable, try to unlock the data object.

plugins/api/src/replica_close.cpp
alanking: Indeed

plugins/api/src/replica_close.cpp
alanking: Oh, sorry, I misread the first sentence.

In that case, the L3 descriptor would no longer refer to the same replica (or at least the same open call), so it would probably result in an error. The caller would be responsible for finalizing the data object at that point, in my mind, and they probably should not call `replica_close` again. It would be the same effect as calling `replica_close` and requesting that nothing in the catalog gets updated, right?

In the same way, calling `rsDataObjClose` on failure is (and as near as I can tell always has been) inadvisable because it tries really hard to make sure everything is cleaned up, even in the face of failure (L1 descriptor, locked objects, etc.).

korydraughn: > I think it depends on what part of the process failed (e.g. Is the L3 descriptor still valid?) and the underlying storage technology.

I agree. However, I think we have to determine if the L3 descriptor is open before calling `close_physical_object`. We can do that by setting it to 0 or -1 following success. If `replica_close` is called again due to a previous failure, `close_physical_object` will be skipped and the catalog updates will be allowed to execute.

> Do you want me to try and find out?

Yes please. Otherwise, I feel users only way out of this case is `iadmin modrepl`.

Thoughts?

alanking:
S3 plugin tests are now failing as a result of the latest changes. Objects are stuck in the locked status because I imagine something has been leaning on the behavior of modifying the replica status even when replica status updates have been explicitly disabled. Looking into it tomorrow

plugins/api/src/replica_close.cpp
alanking: So... you're saying if `close_physical_object` succeeds, we should set `l1desc.l3descInx` to -1 or something so that we know that this L1 descriptor is valid but the object has already been closed?

korydraughn: Correct.

> In that case, the L3 descriptor would no longer refer to the same replica (or at least the same open call), so it would probably result in an error. The caller would be responsible for finalizing the data object at that point, in my mind, and they probably should not call replica_close again. It would be the same effect as calling replica_close and requesting that nothing in the catalog gets updated, right?

I keep forgetting that we have that API. Seems there are two approaches.

I feel checking the L3 descriptor is very simple/approachable for devs using `replica_close`, so my preference is towards that as the solution, but I'm open to your idea as well. We need to think about the pros/cons for these solutions.

alanking:
So, I discovered a couple of things.

1. `rsDataObjPut` has been leaning on updating the status on failure to close the physical object. I have added a call to unlock the object in this case in that API and tests are passing in S3 again. Still messing with this to make sure it's what we want, and to make sure that other APIs are behaving as well.
2. Discovered that not freeing the L1 descriptor after finalizing can lead to some hairy results since agent teardown will attempt to close and finalize (read: stale) any open L1 descriptors. This probably warrants some investigation, but it's pretty tightly coupled with the first thing. Just wanted to mention it.

plugins/api/src/replica_close.cpp
korydraughn: Sounds good.

alanking:
All tests and S3 plugin tests are passing now.

plugins/api/src/replica_close.cpp
alanking: I think this has been addressed. To avoid getting too deep into changing `replica_close`, I am going to mark this resolved. As always, we can continue the discussion in a new thread if it seems appropriate. Thanks!

plugins/api/src/replica_close.cpp
alanking: Done

korydraughn (commented):
If you're happy with it, squash to taste and we'll look over the commits one last time.

alanking:
Squashed

korydraughn (approved):
Add the pounds and merge when ready.

alanking:
#'d and mergin

Issue #6887, opened 2023-01-31
igroupadmin mkuser doesn't properly set new user's password


d-w-moore:
- [x] main
- [x] 4-3-stable
- [x] 4-2-stable
----

### iRODS Version, OS and Version
4.2.11, Ubuntu 18

### What did you try to do?
Create a user with pre-defined password using `igroupadmin mkuser NEW_USER  INITIAL_PASSWORD`

### Expected behavior
New user should be able to log in.

### Observed behavior (including steps to reproduce, if applicable)

We can't even `iinit` with the chosen password, let alone log in. (In the sample session, we're the admin user , but the results for a groupadmin have been tested and  are the same if they were to attempt the `mkuser` subcommand.)
```
irods:~$ igroupadmin  mkuser alibaba opensesame
irods:~$ iadmin lu
rods#tempZone
alibaba#tempZone
irods:~$ logout
ubuntu:~$ rm -fr ~/.irods
ubuntu:~$ iinit
 ERROR: environment_properties::capture: missing environment file. should be at [/home/daniel/.irods/irods_environment.json]
One or more fields in your iRODS environment file (irods_environment.json) are
missing; please enter them.
Enter the host name (DNS) of the server to connect to: localhost
Enter the port number: 1247
Enter your irods user name: alibaba
Enter your irods zone: tempZone
Those values will be added to your environment file (for use by
other iCommands) if the login succeeds.

Enter your current iRODS password:
[-]	/repos/irods/plugins/auth/native/libnative.cpp:342:irods::error native_auth_client_response(irods::plugin_context &, rcComm_t *) :  status [CAT_INVALID_USER]  errno [] -- message [Call to rcAuthResponseFailed.]

 failed with error -827000 CAT_INVALID_USER 

```

d-w-moore:
Note it's possible that the password setting/encoding is not the problem here.  If a user is created with a known password, but the password entry is botched during iinit, the error message is actually what you'd expect from an incorrect password:
```
[-]	/repos/irods/plugins/auth/native/libnative.cpp:342:irods::error native_auth_client_response(irods::plugin_context &, rcComm_t *) :  status [CAT_INVALID_AUTHENTICATION]  errno [] -- message [Call to rcAuthResponseFailed.]
```

alanking:
Please checkbox and close if finished. Thanks!

PR #6889, opened 2023-02-02
[#6887] allow igroupadmin mkuser to initialize the new user's password (4-2-stable)

d-w-moore:


trel (commented):
looks good on inspection.    need a test or five.

server/api/src/rsUserAdmin.cpp
```diff
@@ -140,12 +148,28 @@ _rsUserAdmin( rsComm_t *rsComm, userAdminInp_t *userAdminInp ) {
...
     if ( strcmp( userAdminInp->arg0, "mkuser" ) == 0 ) {
-        return irods::create_user(*rsComm,
+        auto result = irods::create_user(*rsComm,
                                   userAdminInp->arg1 ? userAdminInp->arg1 : "",
                                   "rodsuser", "",
                                   userAdminInp->arg3 ? userAdminInp->arg3 : "");
+
+        if (result != 0) { return result; }
+
+        if (userAdminInp->arg2 && userAdminInp->arg2[0]) {
```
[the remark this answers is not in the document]
d-w-moore: done

server/api/src/rsUserAdmin.cpp
```diff
@@ -140,12 +141,30 @@ _rsUserAdmin( rsComm_t *rsComm, userAdminInp_t *userAdminInp ) {
...
+        auto result = irods::create_user(*rsComm,
                                   userAdminInp->arg1 ? userAdminInp->arg1 : "",
                                   "rodsuser", "",
                                   userAdminInp->arg3 ? userAdminInp->arg3 : "");
+
+        if (result != 0) { return result; }
+
+        // Initialize new user's password with the scrambled password string, but only if given
+        // and non-zero length.
+        if (userAdminInp->arg2 && userAdminInp->arg2[0]) {
```
korydraughn, line 155: Is the second condition safe?
Is `arg2` expected to be null-terminated?
What happens if `arg2` is not null-terminated?

d-w-moore: By convention, the char* arg[0-9] named members of the `userAdminInp_t` are - I'm fairly sure - either null pointers or point at null-terminated character buffers because of management by the  `packStruct` and `unpackStruct` routines.

korydraughn: Do you mean to check the first byte of `arg2`'s buffer though? `arg2` has a type of `char*`.

d-w-moore: Yeah, it'd probably be clearer in that regard, for any char* C, to say `*C` rather than `C[0]`, although they are equivalent.  On the other hand it probably scans into the human eye better this way

trel: due to this amount of discussion, please leave a comment, regardless of the decision here.

korydraughn: In my opinion, using `strnlen` or `is_empty_string` or `is_non_empty_string` express the intent better and guard against buffer overflows, etc.

I recommend considering using one of them instead.

What happens if a string consisting of only whitespace is passed? (e.g. `"   "`)
What happens if the password is too large for the database?

d-w-moore: Is MAX_PASSWORD_LEN the  appropriate constant to send to `strnlen` ?

d-w-moore: I think it'd be too much right not to check the form of the string (eg whether it contains pure whitespace , always has the right prefix, etc, as I just don't know the range of possible values.  But I can definitely check its length is below a certain value and leave a comment to help resolve confusion.

d-w-moore: Ha, learned a new function today.  Didn't know about `strnlen`

korydraughn: I was just wondering what it does when given a string consisting of only whitespace.

> Is MAX_PASSWORD_LEN the appropriate constant to send to strnlen

This looks interesting.
https://github.com/irods/irods/blob/03e7e0515ab3e367c969ddc783b0e0d3dd995924/plugins/database/src/db_plugin.cpp#L7430-L7432

Just so you know, `is_empty_string` and `is_non_empty_string` are simple wrappers around `strnlen`. Based on the code referenced above, `MAX_PASSWORD_LEN - 8` appears to be the answer.

d-w-moore: ok ... interesting. And I'll check to see what db_plugin does with whitespace values for the scrambled password.

korydraughn: Sounds good

korydraughn:
@d-w-moore Please ping someone on the team when this is ready for review. Thanks!

server/api/src/rsUserAdmin.cpp
d-w-moore: @korydraughn In the current state of this PR (and the matching one in icommands), the command 
```
igroupadmin mkuser NEWUSER PW
```
allows a PW value of a single character, like "a" for example, to be used. It also permits any positive number of spaces (chr(32)) to be entered as a password, down to 1 space (but presumably not more than the `MAX_PASSWORD_LEN`), and iinit will allow initialization using the same.  So if we reject the whitespace here, I guess we should file an issue regarding  `iinit` as well.... Do you agree?

korydraughn: If the handling of passwords is consistent across `iinit`, `iadmin moduser password`, and `ipasswd`, then let it ride.

Create an issue stating we need to consider applying some requirements to passwords. We don't have to define what those requirements are, but providing some simple suggestions could be helpful. For example:
- Minimum length of _N_
- No more than two consecutive characters at a time
- Can't match your username

scripts/irods/test/test_igroupadmin.py
```diff
@@ -72,6 +71,63 @@ def tearDownClass(self):
...
+                  'bob':40, 'eve': 3 }
+
+        # TODO -- eventually delete , but using for test r.n.
+        #obj_name = '{}/{}'.format(self.groupadmin.home_collection, 'test__6887__object')
+        #self.groupadmin.assert_icommand(['ichmod', 'write', self.groupadmin.username, obj_name], 'EMPTY' )
+
+        for username,pwlen in USERS.items():
+            try:
+                passwd = (username[0]+'pass_6887_the_quick_brown_fox_jumps_over_the_lazy_dog')[:pwlen]
+                self.groupadmin.assert_icommand(['igroupadmin', 'mkuser', username, passwd], 'EMPTY' )
```
korydraughn, line 85: Just an FYI, specifying `'EMPTY'` is redundant when there's no output.

So, you can remove that argument if you want. Up to you.

scripts/irods/test/test_igroupadmin.py
```diff
@@ -72,6 +71,63 @@ def tearDownClass(self):
             self.admin.__exit__()
             admin_session.assert_icommand(['iadmin', 'rmuser', self.admin.username])
 
+    def test_create_user_with_password_in_igroupadmin__6887(self):
+        USERS = { 'alice':42,
```
korydraughn, line 75: Python convention questions ...

Is the use of capital letters here a convention in Python?
Why not use lowercase letters? _(I'm guessing it's supposed to provide signal of being treated as readonly?)_

scripts/irods/test/test_igroupadmin.py
```diff
@@ -72,6 +71,63 @@ def tearDownClass(self):
...
+                  'bob':40, 'eve': 3 }
```
korydraughn, line 76: Please format the dict so that it is easier to read.

scripts/irods/test/test_igroupadmin.py
```diff
@@ -72,6 +71,63 @@ def tearDownClass(self):
...
+        # TODO -- eventually delete , but using for test r.n.
+        #obj_name = '{}/{}'.format(self.groupadmin.home_collection, 'test__6887__object')
+        #self.groupadmin.assert_icommand(['ichmod', 'write', self.groupadmin.username, obj_name], 'EMPTY' )
+
+        for username,pwlen in USERS.items():
+            try:
+                passwd = (username[0]+'pass_6887_the_quick_brown_fox_jumps_over_the_lazy_dog')[:pwlen]
+                self.groupadmin.assert_icommand(['igroupadmin', 'mkuser', username, passwd], 'EMPTY' )
+                with session.make_session_for_existing_user(username, passwd, lib.get_hostname(), self.admin.zone_name):
+                    pass
```
korydraughn, line 87: Add a comment explaining why these lines are needed.

scripts/irods/test/test_igroupadmin.py
```diff
@@ -72,6 +71,63 @@ def tearDownClass(self):
...
+            finally:
+                self.admin.run_icommand(['iadmin','rmuser',username])
+
+    def test_remove_grom_group_scenarios__6888(self):
```
korydraughn, line 91: `grom` -> `from`

`__6888` -> `__issue_6888`

Helps establish iRODS conventions around naming of tests. The prefix also helps with removing confusion in case other numbers appear in the test name.

scripts/irods/test/test_igroupadmin.py
```diff
@@ -72,6 +71,63 @@ def tearDownClass(self):
             self.admin.__exit__()
             admin_session.assert_icommand(['iadmin', 'rmuser', self.admin.username])
 
+    def test_create_user_with_password_in_igroupadmin__6887(self):
```
korydraughn, line 74: __6887 -> __issue_6887

scripts/irods/test/test_igroupadmin.py
```diff
@@ -72,6 +71,63 @@ def tearDownClass(self):
...
+
+            # Test another user (a rodsuser) can be created and added to the group by this.
+            self.groupadmin.assert_icommand( ['igroupadmin', 'mkuser', self.rodsuser_name], 'EMPTY' )
+            self.groupadmin.assert_icommand( ['igroupadmin', 'atg', self.group, self.rodsuser_name], 'EMPTY' )
+
+            user_with_zone = lambda user_name,zn='': '{}#{}'.format(user_name, zn if zn else self.groupadmin.zone_name)
+
+            # Test that the groupadmin itself, and the new rodsuser, are members of the group.
+            self.groupadmin.assert_icommand(['igroupadmin', 'lg', self.group, ], 'STDOUT_MULTILINE', [user_with_zone(self.rodsuser_name),
+                                                                                                       user_with_zone(self.groupadmin.username)])
```
korydraughn, line 110: Vertical alignment for `user_with_zone()` calls is off by one. I'm not sure if that was intentional.

scripts/irods/test/test_igroupadmin.py
```diff
@@ -72,6 +71,63 @@ def tearDownClass(self):
...
+                passwd = (username[0]+'pass_6887_the_quick_brown_fox_jumps_over_the_lazy_dog')[:pwlen]
+                self.groupadmin.assert_icommand(['igroupadmin', 'mkuser', username, passwd], 'EMPTY' )
+                with session.make_session_for_existing_user(username, passwd, lib.get_hostname(), self.admin.zone_name):
+                    pass
+            finally:
+                self.admin.run_icommand(['iadmin','rmuser',username])
+
+    def test_remove_grom_group_scenarios__6888(self):
+        """Thoroughly test atg, rfg (remove from group) and transfer of responsiblity to another group administrator"""
+        self.rodsuser_name = 'alice'
```
korydraughn, line 93: Does this need to be attached to the class instance / self?

scripts/irods/test/test_igroupadmin.py
```diff
@@ -72,6 +71,63 @@ def tearDownClass(self):
...
+        try:
+            # Use the groupadmin account to create group and set itself up as administrator.
+            self.groupadmin.assert_icommand( ['igroupadmin', 'mkgroup', self.group ], 'EMPTY' )
```
korydraughn, line 96: Does the group variable need to be attached to the class instance / self rather than being a normal local variable?

Also, I don't see where it is defined.

scripts/irods/test/test_igroupadmin.py
```diff
@@ -72,6 +71,63 @@ def tearDownClass(self):
...
+            self.assertNotIn(user_with_zone(self.groupadmin.username), out)
+
+            # Test that the new groupadmin can do some necessary things.
+            self.groupadmin2.assert_icommand(['igroupadmin', 'atg', self.group, self.rodsuser_name], 'EMPTY', desired_rc = 0)
+            self.groupadmin2.assert_icommand(['igroupadmin', 'rfg', self.group, self.rodsuser_name], 'EMPTY', desired_rc = 0)
+        finally:
+            self.admin.run_icommand(['iadmin', 'rmgroup', self.group])
+            self.groupadmin2.__exit__()
+            self.admin.assert_icommand(['iadmin', 'rmuser', self.groupadmin2.username])
+            self.admin.run_icommand(['iadmin', 'rmgroup', self.rodsuser_name])
```
korydraughn, line 130: Do we need to remove the `groupadmin2` from the group before removing the actual group?

server/api/src/rsUserAdmin.cpp
```diff
@@ -1,10 +1,12 @@
+#include "authenticate.h"
 #include "administration_utilities.hpp"
 #include "userAdmin.h"
 #include "rsUserAdmin.hpp"
 #include "icatHighLevelRoutines.hpp"
 #include "miscServerFunct.hpp"
 #include "irods_configuration_keywords.hpp"
 
+
```
korydraughn, line 9: Remove empty line.

korydraughn (commented):
The tests are looking real good. The comments explain what's happening perfectly.

server/api/src/rsUserAdmin.cpp
trel: zoey did a pretty good job of defining the rules with iadmin, i think we should start there and just make other things match.   of course, any new information can sway that.

korydraughn: Are you claiming that should be handled by this PR or referenced in the new issue?

trel: either way - just know that this was already mostly put in a box with some rules.

scripts/irods/test/test_igroupadmin.py
d-w-moore: It doesn't have the strength it does in e.g. Ruby, where a warning is invoked if you change it, but yes.  I think the convention across languages was to have constants announced as all-caps, esp in scripting languages where the "const" is less enforcible.

scripts/irods/test/test_igroupadmin.py
d-w-moore: Yeah, sorry. temporarily needed only. They'll be gone soon. I was exploring the `make_session_for_existing_user` thing as I hadn't used it before.

scripts/irods/test/test_igroupadmin.py
d-w-moore: Duly noted.

scripts/irods/test/test_igroupadmin.py
d-w-moore: Right, I knew that ( I think :) ).

scripts/irods/test/test_igroupadmin.py
korydraughn: Got it.

server/api/src/rsUserAdmin.cpp
d-w-moore: Rules w.r.t. to password format were probably not among those, however.  Length limitations yes, but no restrictions as yet requiring passwords to conform to a level of security.  Is it a thing we want to enforce within the server and/or clients?  Each organization will have their own requirements.

trel: ah, i see your point - yes, i think i was conflating work done on usernames.   of course, you are correct - we must be less restrictive than any possible deployment in a particular organization.  oh, and we require an empty password for anonymous, so, carry on!

scripts/irods/test/test_igroupadmin.py
d-w-moore: done!

scripts/irods/test/test_igroupadmin.py
d-w-moore: Or maybe, needed.  Just to ensure the password works for the login. Helps to issue an icommand as well.  So now that is in final form, I think.

scripts/irods/test/test_igroupadmin.py
d-w-moore: for efficiency, I suppose.  It's assigned in `setupClass`, which I had nothing to do with....

scripts/irods/test/test_igroupadmin.py
d-w-moore: The groupadmin2 account needs to be removed, yes, along with the session_collection, and that's done in the __exit__ call ... but I think it's immaterial the order in which it's done.

server/api/src/rsUserAdmin.cpp
d-w-moore: done!

scripts/irods/test/test_igroupadmin.py
d-w-moore: And done.

scripts/irods/test/test_igroupadmin.py
d-w-moore: No, actually. Ok, taken care of....

scripts/irods/test/test_igroupadmin.py
d-w-moore: Rectified

scripts/irods/test/test_igroupadmin.py
korydraughn: All good. I somehow missed the initialization of `self.group`.

Please mark as resolved.

scripts/irods/test/test_igroupadmin.py
korydraughn: You're saying `__exit__()` handles the session collection, not the removal of the groupadmin user from the group, correct?

> but I think it's immaterial the order in which it's done.

The order of what exactly?

d-w-moore: Sorry, I was thinking you were referring to a different line, so I had the wrong concept in my head...  I think you are right, I don't need to remove the groupadmin2 account; I can just delete the group and the DB handles it.  I'll change the test.

korydraughn: Well, you need to remove the user from the group before doing anything else.

d-w-moore: Do you? I think I just did it without needing to....
```
irods@precision-5530:~/scripts$ iadmin mkgroup lab
irods@precision-5530:~/scripts$ iadmin atg lab dan
irods@precision-5530:~/scripts$ iadmin rmgroup lab
irods@precision-5530:~/scripts$ 
```

d-w-moore: Confirmed, I can delete L129 above with the redundant groupadmin2 removal, and the test still passes.

d-w-moore: No, nevermind that. error on  my part , that doesn't remove the user from the group, it removes the user. Maybe we can get on Zoom and I can get the sense of what your original question was, because now I'm kind of confused.

scripts/irods/test/test_igroupadmin.py
```diff
@@ -72,6 +71,61 @@ def tearDownClass(self):
...
+            self.groupadmin.assert_icommand(['igroupadmin', 'rfg', self.group, self.groupadmin.username])   # remove self
+
+            # Test that groupadmin2 is now a member of the group, and that groupadmin and the rodsuser have successfully been removed.
+            rc, out, _ = self.groupadmin2.assert_icommand(['igroupadmin', 'lg', self.group], 'STDOUT_MULTILINE', user_with_zone(self.groupadmin2.username))
+            self.assertEqual(rc, 0)
+            self.assertNotIn(user_with_zone(rodsuser_name), out)
+            self.assertNotIn(user_with_zone(self.groupadmin.username), out)
+
+            # Test that the new groupadmin can do some necessary things.
+            self.groupadmin2.assert_icommand(['igroupadmin', 'atg', self.group, rodsuser_name], desired_rc = 0)
```
alanking, line 122: Should we assert that `rodsuser_name` is a member of the group after this `atg` call?

scripts/irods/test/test_igroupadmin.py
```diff
@@ -72,6 +71,61 @@ def tearDownClass(self):
             self.admin.__exit__()
             admin_session.assert_icommand(['iadmin', 'rmuser', self.admin.username])
 
+    def test_create_user_with_password_in_igroupadmin__issue_6887(self):
+        USERS = { 'alice': 42,
+                  'bob':   40,
+                  'eve':   3   }
```
alanking, line 77: Is there something special about these password lengths?

d-w-moore: They range from the supposed minimum to maximum, but ... it may not be necessary here to test that range .  If for example a minimum length is something we want to enforce (though I think not) it should be another issue, yes?

scripts/irods/test/test_igroupadmin.py
d-w-moore: perhaps

scripts/irods/test/test_igroupadmin.py
d-w-moore: So maybe just test that the long one works, and be done with it, maybe. @alanking - Let me know what you think.

korydraughn: I'm going to recommend that the test focus on just the one issue and you create a new/separate test that handles invalid password lengths.

That way the test names can be clear about what they are covering.

d-w-moore: ok.  Think I'd need guidance on how to write that issue if I were to do it, though... 
  - It's not always clear for me, when reading the code, whether a length field is a reference to a hashed password length (which in PAM's case I know can be longer than 1024 chars) or the raw one entered by the user on (for example) the passwd, ipasswd or iinit command line.  I'd require deeper knowledge to start writing about the problem. (If there is one ;) ).
  - Do we need a placeholder issue, or are there specific concerns where we think iRODS should be determining / limiting what is a "good" password length.    https://github.com/irods/irods/pull/6889#discussion_r1100205814

korydraughn: Let's create a placeholder issue. It can say something like "Add tests for invalid passwords". It doesn't need to go into great detail. It is simply an indicator that we need to investigate the space and improve coverage.

We can revisit it later and this PR gets to move forward.

alanking: With #6911, I think we can consider this addressed. Thanks!

alanking (commented):
Seems good to me. Anything else need to be done here?

trel:
let's squash/cleanup, no #

d-w-moore:
sounds good, doing a partial squash initially to graft onto 6887.m, then we'll do the whole kaboodle

alanking:
Please ping when squashing is done

d-w-moore:
> Please ping when squashing is done

It's done

alanking (approved):
Looks good to me. You may wish to consider cherry-picking in the other direction (main -> 4-2-stable vs 4-2-stable -> main) since the formatting will be different in the other branch(es), but that's not really a blocker in my opinion. Let's wait for a second approval before adding #s

d-w-moore:
> Looks good to me. You may wish to consider cherry-picking in the other direction (main -> 4-2-stable vs 4-2-stable -> main) since the formatting will be different in the other branch(es), but that's not really a blocker in my opinion. Let's wait for a second approval before adding #s

I guess that makes sense for the formatting, as you say.  But I do find the conflict resolution a bit more confusing when cherry-picking "backward in time", as it were.

alanking:
Indeed. A matter of personal workflow preference :)

scripts/irods/test/test_igroupadmin.py
```diff
@@ -72,6 +71,72 @@ def tearDownClass(self):
...
+        # until resolution https://github.com/irods/irods/issues/6913
+        password_length = 41
+        try:
+            # Create a user and assign a password of the given length.
+            passwd = (username[0]+'pass_6887_the_quick_brown_fox_jumps_over_the_lazy_dog')[:password_length]
+            self.groupadmin.assert_icommand(['igroupadmin', 'mkuser', username, passwd])
+
+            # Test that we can log in as the user, with the assigned password.
+            with session.make_session_for_existing_user(username, passwd, lib.get_hostname(), self.admin.zone_name) as user_session:
+                user_session.assert_icommand('icd')
```
trel, line 86: does this ... authenticate?  I think it just removes a local file...

d-w-moore:
Corresponds to https://github.com/irods/irods_client_icommands/pull/353

scripts/irods/test/test_igroupadmin.py
d-w-moore: The `icd` won't work properly unless the session is authenticated, right?  If I remember right, I did verify it manually by stepping through the test.

alanking: Hmm... I had this question before, but then I saw that `icd` does in fact authenticate: https://github.com/irods/irods_client_icommands/blob/0e65ee613b8be1533b667b85f9e8320574201d51/src/icd.cpp#L72

...and then didn't think about it again. However, I looked again just now and... https://github.com/irods/irods_client_icommands/blob/0e65ee613b8be1533b667b85f9e8320574201d51/src/icd.cpp#L44-L53

So... `icd` doesn't even connect to the server when you don't pass it anything. So, we should probably fix that up. `ils` or something should suffice. Good catch.

trel: mmm, no?    not sure.

4.2.11...
```
$ iexit full
$ ils
Enter your current iRODS password: ^C
$ icd
$ echo $?
0
```

trel: a deep knowledge of the code.

alanking: On the other hand, `make_session_for_existing_user` calls `iinit`, `imkdir <session_collection>` *and* `icd <session_collection>`, so `make_session_for_existing_user` would have failed if it couldn't authenticate. Hmm...

d-w-moore: Ok, will change to `ils`

alanking: For completeness...
https://github.com/irods/irods/blob/de233d4cf5a6ba41aee885000940ffadd7cff45a/scripts/irods/test/session.py#L102-L107
One of these things should be raising an `Exception` if authentication is failing, so I think this is probably fine. The `icd` line is unrelated, so you could replace it with an `ils` or just `pass`.

d-w-moore: I like the idea of forcing a useful, if trivial, a bit of "useful" network activity to be done, so I'll stick with the `ils` maybe...

alanking (commented):
Seems good to me. My only suggestion at this point would be to squash that last commit into the other one with that issue number, but it's not really important. Awaiting second approval

trel:
agreed on the squash to two comment.

alanking (approved):
Very good. Please # it

alanking (commented):
Okay, seems fine. Please squash that last commit into "respect password field in igroupadmin mkuser USER PW"

d-w-moore:
squashed.

d-w-moore:
# now, yes ?

alanking (approved):
Yeps

d-w-moore:
done

PR #6899, opened 2023-02-09
[6502] mark replica stale if staging to cache fails (4-2-stable)

JustinKyleJames:
Changed the compound behavior as follows:

If the policy is set to prefer the cache and the cache votes .25 (irc::vote::low) or less, get the vote from the archive.  If its vote is higher, repave the cache with the archive and return this.

If the cache votes > .25 then return the file from the cache.

Some situations where the vote is <= .25:

1. The replica does not exist in the cache.  (vote=0.0)
2. The replica exists but it stale.  (vote=.25)
3. The user requested a specific replica.  (vote=.25)

I have a test in the S3 plugin which covers this scenario with the archive file being in Glacier.

### DIFF
--- a/packaging/core.re.template
+++ b/packaging/core.re.template
@@ -701,7 +701,7 @@ acPostProcForDataCopyReceived(*leaf_resource) {}
 
 # =-=-=-=-=-=-=-
 # policy controlling when a dataObject is staged to cache from archive in a compound coordinating resource
-#  - the default is to stage when cache is not present ("when_necessary")
+#  - the default is to stage when cache is stale or not present ("when_necessary")
 # =-=-=-=-=-=-=-
 # pep_resource_resolve_hierarchy_pre(*INSTANCE, *CONTEXT, *OUT, *OPERATION, *HOST, *PARSER, *VOTE){*OUT="compound_resource_cache_refresh_policy=when_necessary";}  # default
 # pep_resource_resolve_hierarchy_pre(*INSTANCE, *CONTEXT, *OUT, *OPERATION, *HOST, *PARSER, *VOTE){*OUT="compound_resource_cache_refresh_policy=always";}
--- a/plugins/resources/compound/libcompound.cpp
+++ b/plugins/resources/compound/libcompound.cpp
@@ -57,6 +57,35 @@ const std::string AUTO_REPL_POLICY_ENABLED( "on" );
 
 namespace
 {
+    auto throw_error_if_archive_replica_requested(const irods::file_object_ptr& obj,
+                                                  const std::string_view archive_resc_name) -> void
+    {
+        // specifying replica by number for a replica in an archive resource is not allowed as it creates
+        // troublesome edge cases (see issue 6926)
+        if (obj->repl_requested() > -1) {
+            const auto& target_replica = obj->get_replica(obj->repl_requested());
+            if (!target_replica) {
+                // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+                THROW(
+                    SYS_REPLICA_DOES_NOT_EXIST,
+                    fmt::format(
+                        "Requested replica [{}] of [{}] does not exist.", obj->repl_requested(), obj->logical_path()));
+            }
+
+            auto target_resource_name = target_replica->get().resc_name();
+
+            if (target_resource_name == archive_resc_name) {
+                THROW(
+                    DIRECT_ARCHIVE_ACCESS,
+                    fmt::format("Direct access to an archive resource is not allowed.  replica number {} of object {} "
+                                "belongs to archive resource {}.",
+                                obj->repl_requested(),
+                                obj->logical_path(),
+                                target_resource_name));
+            }
+        }
+    }
+
     auto get_archive_replica_number(
         const irods::file_object_ptr _obj,
         const std::string_view _archive_resource_name) -> int
@@ -1591,9 +1620,34 @@ irods::error open_for_prefer_archive_policy(
         rodsLog(LOG_NOTICE, "[%s] - operation not found in property map; using open.", __FUNCTION__);
     }
 
+    irods::file_object_ptr f_ptr = boost::dynamic_pointer_cast<irods::file_object>(_ctx.fco());
+
+    // =-=-=-=-=-=-=-
+    // get the archive name
+    std::string archive_name;
+    ret = _ctx.prop_map().get<std::string>(ARCHIVE_CONTEXT_TYPE, archive_name);
+    if (!ret.ok()) {
+        return PASS(ret);
+    }
+
+    // =-=-=-=-=-=-=-
+    // get the cache name
+    std::string cache_name;
+    ret = _ctx.prop_map().get<std::string>(CACHE_CONTEXT_TYPE, cache_name);
+    if (!ret.ok()) {
+        return PASS(ret);
+    }
+
+    try {
+        throw_error_if_archive_replica_requested(f_ptr, archive_name);
+    }
+    catch (const irods::exception& e) {
+        irods::log(e);
+        return irods::error(e);
+    }
+
     // =-=-=-=-=-=-=-
     // repave the repl requested temporarily
-    irods::file_object_ptr f_ptr = boost::dynamic_pointer_cast< irods::file_object >( _ctx.fco() );
     int repl_requested = f_ptr->repl_requested();
     const irods::at_scope_exit restore_repl_requested{[&] { f_ptr->repl_requested(repl_requested); }};
 
@@ -1601,8 +1655,6 @@ irods::error open_for_prefer_archive_policy(
     irods::hierarchy_parser arch_check_parser = _out_parser;
 
     try {
-        std::string archive_name;
-        _ctx.prop_map().get<std::string>(ARCHIVE_CONTEXT_TYPE, archive_name);
         f_ptr->repl_requested(get_archive_replica_number(f_ptr, archive_name));
 
         // =-=-=-=-=-=-=-
@@ -1681,14 +1733,6 @@ irods::error open_for_prefer_archive_policy(
     // get this resc name
     std::string current_name;
     ret = _ctx.prop_map().get<std::string>( irods::RESOURCE_NAME, current_name );
-    if ( !ret.ok() ) {
-        return PASS( ret );
-    }
-
-    // =-=-=-=-=-=-=-
-    // get the cache name
-    std::string cache_name;
-    ret = _ctx.prop_map().get< std::string >( CACHE_CONTEXT_TYPE, cache_name );
     if ( !ret.ok() ) {
         return PASS( ret );
     }
@@ -1792,6 +1836,14 @@ irods::error open_for_prefer_cache_policy(
             cache_resc_name );
     if ( !ret.ok() ) {
         return PASS( ret );
+    }
+
+    try {
+        throw_error_if_archive_replica_requested(obj, archive_resc_name);
+    }
+    catch (const irods::exception& e) {
+        irods::log(e);
+        return irods::error(e);
     }
 
     std::string operation{irods::OPEN_OPERATION};
--- a/plugins/resources/compound/libcompound.cpp
+++ b/plugins/resources/compound/libcompound.cpp
@@ -22,6 +22,7 @@
 #include "irods_lexical_cast.hpp"
 #include "irods_random.hpp"
 #include "irods_at_scope_exit.hpp"
+#include "voting.hpp"
 
 #include <boost/lexical_cast.hpp>
 #include <boost/function.hpp>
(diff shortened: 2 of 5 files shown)


plugins/resources/compound/libcompound.cpp
```diff
@@ -1791,18 +1813,13 @@ irods::error open_for_prefer_cache_policy(
...
 
     // =-=-=-=-=-=-=-
-    // if the vote is 0 then the cache doesn't have it so it will need be staged
-    if ( 0.0 == cache_check_vote ) {
-        // =-=-=-=-=-=-=-
-        // repave the repl requested temporarily
-        irods::file_object_ptr f_ptr = boost::dynamic_pointer_cast< irods::file_object >( _ctx.fco() );
-        int repl_requested = f_ptr->repl_requested();
-        f_ptr->repl_requested( -1 );
+    // If the vote is irv::vote:low or less, either the replica does not exist in cache,
```
trel, line 1816: :low -> ::low

trel (commented):
i think this all makes sense - found one comment typo

trel:
do we need to change the core.re comment too?  i thought i heard it say 'if present'...

plugins/resources/compound/libcompound.cpp
```diff
@@ -533,10 +534,15 @@ namespace {
 
     int close_replica(
         irods::plugin_context& _ctx,
-        const int l1descInx) {
+        const int l1descInx,
+        const int error_code_set_to_oprStatus_on_dest_close = 0) {
```
alanking, line 538: This input also impacts the call to close for the source replica, so I think the bit about `on_dest_close` is a little misleading. I do appreciate how descriptive this is, so maybe something to indicate that this is for the operation status without the mention of source or destination? Also could consider splitting the function in two: one for each replica. Just my two cents

alanking (commented):
Overall looks good and makes sense

plugins/resources/compound/libcompound.cpp
JustinKyleJames: done

plugins/resources/compound/libcompound.cpp
JustinKyleJames: I am only currently setting the oprStatus on the destination close as the source isn't being altered and I don't think the status of the source should be altered.

plugins/resources/compound/libcompound.cpp
```diff
@@ -1817,9 +1834,20 @@ irods::error open_for_prefer_cache_policy(
...
-            *_out_vote = 0.0;
+    // =-=-=-=-=-=-=-
+    // If the archive voted higher it will need to be staged to cache
+    if ( arch_check_vote > cache_check_vote ) {
+
+        // =-=-=-=-=-=-=-
+        // repave the repl requested temporarily
+        irods::file_object_ptr f_ptr = boost::dynamic_pointer_cast< irods::file_object >( _ctx.fco() );
+        int repl_requested = f_ptr->repl_requested();
+        f_ptr->repl_requested( -1 );
```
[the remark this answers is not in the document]
JustinKyleJames: I updated the code to repave repl_requested before calling resolve resource hierarchy on the archive.

I verified that before this change asking for a stale cache replica tried to stage from archive.  Now if you are selecting the cache it will simply return the cache replica even when stale.

plugins/resources/compound/libcompound.cpp
korydraughn: I vote to pull the `oprStatus` change out of `close_replica` if possible. Pushing the handling of that value into `close_replica` doesn't appear to offer much.

scripts/irods/test/test_resource_types.py
```diff
@@ -2067,6 +2067,42 @@ def test_auto_repl_on__2941(self):
...
+        self.admin.assert_icommand("iget -f " + filename + " " +  get_filename)
+
+        # assert on the good replica file size
+        assert(os.path.getsize(get_filename)) == 10
+
+        # cleanup
+        self.admin.assert_icommand("irm -f " + filename)
+        os.remove(filename)
+        os.remove(stale_replica_filename)
+        os.remove(get_filename)
```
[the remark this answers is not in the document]
JustinKyleJames: done

scripts/irods/test/test_resource_types.py
```diff
@@ -2067,6 +2067,42 @@ def test_auto_repl_on__2941(self):
...
+        # swap the status of these two so archive is good and cache is stale
+        self.admin.assert_icommand("iadmin modrepl logical_path " + logical_path + " replica_number 0 DATA_REPL_STATUS 0")
+        self.admin.assert_icommand("iadmin modrepl logical_path " + logical_path + " replica_number 1 DATA_REPL_STATUS 1")
+
+        # get the stale replica
+        get_filename = "getfile.txt"
+        self.admin.assert_icommand("iget -n 0 " + filename + " " +  get_filename)
+
+        # assert on the stale replica file size
+        assert(os.path.getsize(get_filename)) == 20
```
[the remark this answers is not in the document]
JustinKyleJames: done

plugins/resources/compound/libcompound.cpp
JustinKyleJames: I am going to move it outside of `close_replica` but I am not sure that is a great design either.  We are introducing code that needs to be inserted every time `close_replica` is called.

JustinKyleJames: A couple of points.

1.  If I move oprStatus outside of `close_replica` I don't think it would make any sense to leave anything other than the actual call to `rsDataObjClose` inside that method.  At that point it wouldn't make a lot of sense to even have `close_replica`.

2.  I was completely wrong with my first comment on this.  For some reason I had it in my head that `close_replica` was closing both source and destination rather than it being called twice - once for each.  I am not sure why I was thinking that but I agree that my variable naming makes no sense in light of the fact that there are two calls to `close_replica`.

JustinKyleJames: I updated the variable name to `operation_status` as described above.  I did not use a leading underscore to be consistent with existing code.

alanking (commented):
Seems good to me. Please squash to taste and leave off the #s for now.

alanking:
Oh, and feel free to fix that formatting. :p

JustinKyleJames:
I pushed the fix for 6395 in with this pull request.

The test case relies on the unique behavior you get when you fail to sync-to-archive vs stage-to-cache.  That is the only way I know to verify that the stage-to-cache was not run.

The behavior is different in 4.3 due to the change for issue 6886.

JustinKyleJames:
I removed the test case I wrote for 6935 because the test case for 6886 should suffice now that that code has been merged.

alanking (commented):
Seems good to me. Has this been run through the full test gauntlet?

packaging/core.re.template
```diff
@@ -701,7 +701,7 @@ acPostProcForDataCopyReceived(*leaf_resource) {}
 
 # =-=-=-=-=-=-=-
 # policy controlling when a dataObject is staged to cache from archive in a compound coordinating resource
-#  - the default is to stage when cache is not present ("when_necessary")
+#  - the default is to stage when cache is stale or not present ("when_necessary")
```
alanking, line 704: @korydraughn / @trel - Are we okay changing the definition of `when_necessary` for 4.2? I know we had a discussion about this but can't remember where we landed. Just want to make sure this is acknowledged before this gets approved.

korydraughn: I don't remember having a discussion about changing the behavior of this option.

@JustinKyleJames What lead to this change? Is it required for this PR?

trel: i believe we did decide that 'when_necessary' should kick in if EITHER the cache is missing (current, most likely behavior) OR the cache is stale (which we just hadn't tested/considered before - it's not clear how it would even happen in practice).

Nobody wants a stale cache, so let's clean up a weird edge case while we're here.

korydraughn: I agree. I can't think of a downside to this.

alanking: Okay, just wanted to make sure we were okay with the change for 4.2. I think it is an overall Good Thing.

plugins/resources/compound/libcompound.cpp
```diff
@@ -56,6 +57,37 @@ const std::string AUTO_REPL_POLICY_ENABLED( "on" );
 
 namespace
 {
+    auto throw_error_if_archive_replica_requested(
+            const irods::file_object_ptr& obj,
+            const std::string_view archive_resc_name) -> void
```
alanking, line 62: clang-format?

plugins/resources/compound/libcompound.cpp
```diff
@@ -1853,7 +1853,7 @@ irods::error open_for_prefer_cache_policy(
 
     // =-=-=-=-=-=-=-
     // If the archive voted higher it will need to be staged to cache
-    if ( arch_check_vote > cache_check_vote ) {
+    if ( irods::WRITE_OPERATION != *_opr && arch_check_vote > cache_check_vote ) {
```
alanking: I think we should add a test for the unnecessary data movement. Here's a scenario that clearly demonstrates it:
```bash
$ ils -l foo
  rods              3 comp0;archive0          240 2023-02-27.18:24 & foo
  rods              4 other          240 2023-02-27.18:22 & foo
$ echo "this other stuff" | istream -R other write foo
$ ils -l foo
  rods              3 comp0;archive0          240 2023-02-27.18:24 X foo
  rods              4 other           17 2023-02-27.18:24 & foo
  rods              5 comp0;cache0          240 2023-02-27.18:24 X foo
```
With your changes, we would expect nothing to appear in the cache, as I understand it.

It's a little harder to show that the unnecessary movement isn't happening when the compound resource is actually targeted for the write, though.

Thoughts?

alanking (commented):
I left a comment about adding a test and some formatting stuff. Otherwise everything looks good overall.

plugins/resources/compound/libcompound.cpp
JustinKyleJames: I thought I didn't need to because I was copying from main which had run clang-format but I realize after the fact that I didn't copy some of it but manually edited it.  It is done now.

plugins/resources/compound/libcompound.cpp
JustinKyleJames: Test case added and pushed to 4-2-stable.  Please take a look and if it is good I will roll it into main.

korydraughn: I agree. We should definitely capture that in a test.

scripts/irods/test/test_resource_types.py
```diff
@@ -2109,6 +2109,36 @@ def test_get_stale_cache_replica__issue_6502(self):
             lib.remove_file_if_exists(stale_replica_filename)
             lib.remove_file_if_exists(get_filename)
 
+    def test_no_stage_to_cache_on_write__isue_6935(self):
```
alanking, line 2112:
```suggestion
    def test_no_stage_to_cache_on_write__issue_6935(self):
```

scripts/irods/test/test_resource_types.py
```diff
@@ -2109,6 +2109,36 @@ def test_get_stale_cache_replica__issue_6502(self):
...
+            self.admin.assert_icommand(['istream', 'write', '-R', 'demoResc', filename], input='testing issue 6935')
+
+            # repl the file to origResc (unixfilesystem)
+            self.admin.assert_icommand(['irepl', '-R', 'origResc', filename])
+
+            # trim the cache replica
+            self.admin.assert_icommand(['itrim', '-n0', filename], 'STDOUT', ['files trimmed = 1'])
+
+            # write a new file to origResc (unixfilesystem)
+            self.admin.assert_icommand(['istream', 'write', '-R', 'demoResc', filename], input='testing issue 6935 with more bytes')
```
alanking, line 2129: Is this supposed to be targeting `origResc`?

scripts/irods/test/test_resource_types.py
JustinKyleJames: fixed

scripts/irods/test/test_resource_types.py
JustinKyleJames: No, it was not.  I also created a new resource in case you also just don't want origResc used.

scripts/irods/test/test_resource_types.py
```diff
@@ -2069,6 +2070,174 @@ def test_auto_repl_on__2941(self):
...
+            # trim the cache replica
+            self.admin.assert_icommand(['itrim', '-n0', filename], 'STDOUT', ['files trimmed = 1'])
+
+            # write a new file to TestResc (unixfilesystem)
+            self.admin.assert_icommand(['istream', 'write', '-R', 'TestResc', filename], input='testing issue 6935 with more bytes')
+
+            # assert that the replica does not exist on cache_resc
+            self.assertFalse(lib.replica_exists_on_resource(self.admin, logical_path, self.cache_resc))
+
+            # assert that the replica exists on ufsresource and self.archive_resc
```
korydraughn, line 2134: ufsresource -> TestResc?

alanking (commented):
Everything seems to check out. Please squash to taste and leave off #s and confirm that tests are passing

scripts/irods/test/test_resource_types.py
JustinKyleJames: fixed

alanking:
Okay, if there's nothing else, let's go ahead and squash this.

scripts/irods/test/test_resource_types.py
JustinKyleJames: updated

JustinKyleJames:
> Okay, if there's nothing else, let's go ahead and squash this.

It has been split/rearanged/squashed.  I verified that the branch was identical before and after.

alanking (commented):
Code changes look good to me. One thing, and my approval is won :)

Should this commit message...
> Reject request with repl number to cache

be...
> Reject request with repl number to archive

korydraughn:
Has this PR been tested?

JustinKyleJames:
> Code changes look good to me. One thing, and my approval is won :)
> 
> Should this commit message...
> 
> > Reject request with repl number to cache
> 
> be...
> 
> > Reject request with repl number to archive

Yes it should.  Fixed.

JustinKyleJames:
I pushed changes. All of the original reviewed changes have been split/rearranged/squashed as necessary.  The last two commits are new.

alanking (approved):
Looks good to me. If tests are passing, I say we add #s.

korydraughn (commented):
Code looks good.

Are the core tests and unit tests passing?

JustinKyleJames:
> Code looks good.
> 
> Are the core tests and unit tests passing?

Yes, both are passing.

PR #6910, opened 2023-02-14
Issue 6764

### LINKED ISSUE #6764: Error for password too long only occurs for passwords between 43 and 45 characters.
- [x] main
- [x] 4-3-stable
- [x] 4-2-stable
---
## Bug Report

Error for password too long only occurs for passwords between 43 and 45 characters.

### iRODS Version, OS and Version

irods version 4.3.0-1
icommands version 4.3.0-1
ubuntu 20.04(on both client and provider host), 

### What did you try to do?

I random;y mashed my keyboard to get a long password, and it wasn't accepted by irods, with a weird error message. After tinkering I found out it was too long. However, the message PASSWORD_EXCEEDS_MAX_SIZE is not given that often.

### Expected behavior
iadmin moduser testuser password 1234567890123456789012345678901234567890123456
remote addresses: 10.4.203.69 ERROR: rcGeneralAdmin failed with error -903000 PASSWORD_EXCEEDS_MAX_SIZE

### Observed behavior (including steps to reproduce, if applicable)

iadmin moduser testuser password 1234567890123456789012345678901234567890123456
remote addresses: X.X.X.X ERROR: rcGeneralAdmin failed with error -850000 CAT_PASSWORD_ENCODING_ERROR
Level 0: Error with password encoding.  This can be caused by not connecting directly to the ICAT host, not using password authentication (using GSI or Kerberos instead), or entering your password incorrectly (if prompted).

JustinKyleJames:
This is to fix #6764.


alanking:
I'm going to adjust the PR message so that it doesn't close the issue when we merge this (see https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)

plugins/database/src/db_plugin.cpp
```diff
@@ -7839,7 +7839,8 @@ irods::error db_mod_user_op(
         int i;
         char userIdStr[MAX_NAME_LEN];
         i = decodePw( _ctx.comm(), _new_value, decoded );
-        if (strlen(decoded) > MAX_PASSWORD_LEN - 8) {
+        if (i == CAT_PASSWORD_ENCODING_ERROR || strlen(decoded) > MAX_PASSWORD_LEN - 8) {
```
alanking, line 7842: Is there ever a case where `CAT_PASSWORD_ENCODING_ERROR` is returned but it's not because the password is too long? What if the password is too short (<3 characters)?

JustinKyleJames: We discussed this in standup the other day.  It is hard to say due to all the stuff that is going on in decodePw.  That possibility is why we decided to not update decodePw but instead catch the error here.

scripts/irods/test/test_iadmin.py
```diff
@@ -2226,6 +2226,28 @@ def test_moduser_password(self):
...
+        pw4 = '1234567890123456789012345678901234567890123'
+        self.admin.assert_icommand(['iadmin', 'moduser', self.username, 'password', pw4], 'STDERR', 'PASSWORD_EXCEEDS_MAX_SIZE')
+
+        # Change password to 45 bytes (should fail)
+        pw4 = '123456789012345678901234567890123456789012345'
+        self.admin.assert_icommand(['iadmin', 'moduser', self.username, 'password', pw4], 'STDERR', 'PASSWORD_EXCEEDS_MAX_SIZE')
+
+        # Change password to 46 bytes (should fail)
+        pw4 = '1234567890123456789012345678901234567890123456'
+        self.admin.assert_icommand(['iadmin', 'moduser', self.username, 'password', pw4], 'STDERR', 'PASSWORD_EXCEEDS_MAX_SIZE')
```
korydraughn, line 2250: Consider adding an assertion following this that proves the failed calls to modify the user's password don't result in breaking the user's ability to execute commands. The assertion can do an `ils` or something else entirely.

JustinKyleJames: I added some code to verify that a session can be created with the last "good" password and not with the "bad" passwords.

JustinKyleJames: I added an "ils" command after creating the session for the last good password.

alanking:
For some reason the PR is still linked to the issue. We'll need to figure out how to make that go away.

Please resolve the open conversation if you feel that it is addressed or does not need to be addressed. Then I think we can go ahead and squash this to taste (leave off the #s for now).

trel:
I don't think we're going to be able to 'remove' the connection - if the owner/admin of the repo AND the creator of the issue neither have buttons... I think we just have to watch it when it autocloses, and then reopen/reclose manually.

plugins/database/src/db_plugin.cpp
alanking: @JustinKyleJames - I still think a comment would be helpful here, but if we're not going to do that, please resolve this conversation. Thanks!

korydraughn: Agreed.

@JustinKyleJames If you've acquired knowledge about this, let's write it down for the next person. This just helps us move faster.

JustinKyleJames: I added a comment for this.  I said that the CAT_PASSWORD_ENCODING_ERROR occurs when the password is not of the correct length.  There may be other cases where a CAT_PASSWORD_ENCODING_ERROR isn't due to that but I am not able to determine from the code if that is the case.

korydraughn (commented):
Changes look good.

Has this PR been run through the test suite?

JustinKyleJames:
> Changes look good.
> 
> Has this PR been run through the test suite?

I honestly don't remember as it has been a while.  I'll run it just to make sure.

korydraughn:
Sounds good.

JustinKyleJames:
> Changes look good.
> 
> Has this PR been run through the test suite?

I ran through all unit and CI tests.  They passed.

Before doing that I rebased onto 4-2-stable.  I also squashed and forced push.

JustinKyleJames:
Pounded.

JustinKyleJames:
pounded

korydraughn (approved):
Excellent.

d-w-moore:
I found that this commit breaks a [PRC test](https://github.com/irods/python-irodsclient/blob/57dbfc17aeb6627278106fc31bc61ed9ef680244/irods/test/user_group_test.py#L110) where the test relies on  CAT_PASSWORD_ENCODING_ERROR to be thrown on "alice's" (eve's?) attempt to detect the use of an invalid old-password value to change their password. On tip of 4-2-stable, we now get `PASSWORD_EXCEEDS_MAX_SIZE` thrown instead, even though the old pw value is much less than the maximum length ie. `(MAX_PASSWORD_LEN - 8)`

If the old-password value is merely incorrect, I think we should be flagging it as an incorrect/invalid user parameter or such, and not as "too long".

JustinKyleJames:
> merely incorrect, I think we should be flagging it as an incorrect/invalid user parameter or such, rather than being "too long"( which in this case, it certainly wasn't).

This is not entirely surprising.  I stated on the original issue that I can't guarantee that that encoding error was caused by the password being too long.  I wasn't sure because of the password obfuscating logic being in the middle.

d-w-moore:
If it helps, I do know that `__eq__` gets called from the most-derived class, so if  you have `class PathObj(str)`, then `"a"==PathObj("a")`  and `PathObj("a")=="a"` both evaluate to `True`, and both call `PathObj.__eq__` ... but you probably already knew that : )

Issue #6896, opened 2023-02-06
`iget -n` does not always return specified replica


alanking:
- [x] main
- [x] 4-3-stable
- [x] 4-2-stable
---
### iRODS Version, OS and Version
4.2.9+

### What did you try to do?
Tried to do something like the steps described here:
https://github.com/irods/irods/issues/6895#issue-1572767541
```bash
$ ilsresc -l pt0 | grep context
context: write=1;read=1
parent context: 
$ ilsresc -l pt1 | grep context
context: write=0.1;read=0.1
parent context: 
$ ilsresc
repl:replication
├── pt0:passthru
│   └── ufs0:unixfilesystem
└── pt1:passthru
    └── ufs1:unixfilesystem
$ ils -l
/tempZone/home/rods:
  rods              0 repl;pt0;ufs0          239 2023-02-06.21:38 X foo
  rods              1 repl;pt1;ufs1          284 2023-02-06.21:36 & foo
```

### Expected behavior
`iget -n1` should return replica 1 no matter how low it votes. If the vote for this replica is 0, an error should have been returned. 
```bash
$ iget -n1 foo - | wc -c
284
```

### Observed behavior (including steps to reproduce, if applicable)
The winner of the hierarchy resolution was the replica with the highest vote, which was *not* the specified replica number. Notice that the length is that of the stale replica on the hierarchy with a read weight of 1, not the good replica that we specified on the hierarchy with a read weight of 0.1:
```bash
$ iget -n1 foo - | wc -c
239
```
This is a regression.

PR #6923, opened 2023-02-20
[#6896] Requested replica always wins the vote (main)

alanking:
Addresses #6896

Core tests are now running

### DIFF
--- a/plugins/resources/src/compound.cpp
+++ b/plugins/resources/src/compound.cpp
@@ -1767,6 +1767,21 @@ irods::error open_for_prefer_archive_policy(
         return SUCCESS();
     }
 
+    try {
+        // Set the vote for the replica in the cache resource to be the same as the vote for the replica in the archive
+        // because the replica in the cache is going to be overwritten by the replica in the archive anyway. The vote
+        // must be set for the cache replica because the user may have requested the replica in the cache specifically
+        // and the requested replica must vote higher than 0 or else the operation will fail with an error. If the cache
+        // does not have a replica, there is no replica number to refer to and an error will occur if the user tries to
+        // request a non-existent replica.
+        if (const auto r = f_ptr->get_replica(resc_mgr.get_hier_to_root_for_resc(cache_name)); r) {
+            r->get().vote(arch_check_vote);
+        }
+    }
+    catch (const irods::exception& e) {
+        return {e};
+    }
+
     irods::data_object_ptr d_ptr = boost::dynamic_pointer_cast<irods::data_object>(f_ptr);
     add_key_val(d_ptr, NO_CHK_COPY_LEN_KW, "prefer_archive_policy");
 
--- a/plugins/resources/src/passthru.cpp
+++ b/plugins/resources/src/passthru.cpp
@@ -68,22 +68,32 @@
 const std::string WRITE_WEIGHT_KW( "write" );
 const std::string READ_WEIGHT_KW( "read" );
 
-namespace {
-
-auto apply_weight_to_object_votes(
-    irods::plugin_context& ctx,
-    const double weight) -> void
+namespace
 {
-    irods::file_object_ptr file_obj = boost::dynamic_pointer_cast<irods::file_object>(ctx.fco());
-    for (auto& r : file_obj->replicas()) {
-        rodsLog(LOG_DEBUG,
-            "[%s:%d] - applying weight [%f] to vote [%f] for [%s]",
-            __FUNCTION__, __LINE__,
-            weight, r.vote(),
-            r.resc_hier().c_str());
-        r.vote(r.vote() * weight);
-    }
-} // apply_weight_to_object_votes
+    using log_resource = irods::experimental::log::resource;
+
+    auto apply_weight_to_object_votes(irods::plugin_context& _ctx,
+                                      const std::string& _this_resource_name,
+                                      double _weight) -> void
+    {
+        irods::file_object_ptr file_obj = boost::dynamic_pointer_cast<irods::file_object>(_ctx.fco());
+        for (auto& r : file_obj->replicas()) {
+            // Only apply vote weight to replicas whose hierarchies contain this resource.
+            if (!irods::hierarchy_parser{r.resc_hier()}.contains(_this_resource_name)) {
+                continue;
+            }
+
+            log_resource::debug("[{}]: applying weight [{}] to replica [{}] of [{}] which voted [{}]",
+                                _this_resource_name,
+                                _weight,
+                                r.repl_num(),
+                                file_obj->logical_path(),
+                                r.vote());
+
+            // NOLINTNEXTLINE(bugprone-narrowing-conversions,cppcoreguidelines-narrowing-conversions)
+            r.vote(r.vote() * _weight);
+        }
+    } // apply_weight_to_object_votes
 
 } // anonymous namespace
 
@@ -754,7 +764,14 @@ irods::error passthru_file_resolve_hierarchy(
         return ERROR(SYS_INVALID_INPUT_PARAM, "Invalid input parameter.");
     }
 
-    _out_parser->add_child(irods::get_resource_name(_ctx));
+    std::string this_resource_name;
+    try {
+        this_resource_name = irods::get_resource_name(_ctx);
+    }
+    catch (const irods::exception& e) {
+        return {e};
+    }
+    _out_parser->add_child(this_resource_name);
 
     irods::resource_ptr resc;
     ret = passthru_get_first_child_resc( _ctx.prop_map(), resc );
@@ -774,42 +791,40 @@ irods::error passthru_file_resolve_hierarchy(
                                  _curr_host,
                                  _out_parser,
                                  _out_vote );
- 
+
     double orig_vote = *_out_vote;
-    if ( irods::OPEN_OPERATION == ( *_opr ) ) {
+
+    if (irods::OPEN_OPERATION == (*_opr)) {
         double read_weight = 1.0;
-        irods::error ret = capture_weight(_ctx, READ_WEIGHT_KW, read_weight);
-        if ( !ret.ok() ) {
-            irods::log( PASS( ret ) );
+        if (const auto ret = capture_weight(_ctx, READ_WEIGHT_KW, read_weight); !ret.ok()) {
+            irods::log(PASS(ret));
         }
         else {
-            ( *_out_vote ) *= read_weight;
-            apply_weight_to_object_votes(_ctx, read_weight);
+            // NOLINTNEXTLINE(bugprone-narrowing-conversions,cppcoreguidelines-narrowing-conversions)
+            (*_out_vote) *= read_weight;
+            apply_weight_to_object_votes(_ctx, this_resource_name, read_weight);
         }
-
     }
-    else if ( ( irods::CREATE_OPERATION == ( *_opr ) ||
-                irods::WRITE_OPERATION == ( *_opr ) ) ) {
+    else if ((irods::CREATE_OPERATION == (*_opr) || irods::WRITE_OPERATION == (*_opr))) {
         double write_weight = 1.0;
-        irods::error ret = capture_weight(_ctx, WRITE_WEIGHT_KW, write_weight);
-        if ( !ret.ok() ) {
-            irods::log( PASS( ret ) );
+        if (const auto ret = capture_weight(_ctx, WRITE_WEIGHT_KW, write_weight); !ret.ok()) {
+            irods::log(PASS(ret));
         }
         else {
-            ( *_out_vote ) *= write_weight;
-            apply_weight_to_object_votes(_ctx, write_weight);
+            // NOLINTNEXTLINE(bugprone-narrowing-conversions,cppcoreguidelines-narrowing-conversions)
+            (*_out_vote) *= write_weight;
+            apply_weight_to_object_votes(_ctx, this_resource_name, write_weight);
         }
     }
 
-    rodsLog(
-        LOG_DEBUG,
-        "passthru_file_resolve_hierarchy - [%s] : %f - %f",
-        _opr->c_str(),
-        orig_vote,
-        *_out_vote );
+    log_resource::debug("[{}] - resource [{}], operation [{}], original vote [{}], weighted vote [{}]",
+                        __func__,
+                        this_resource_name,
+                        *_opr,
+                        orig_vote,
+                        *_out_vote);
 
     return final_ret;
-
 } // passthru_file_resolve_hierarchy
 
 // =-=-=-=-=-=-=-
--- a/scripts/irods/lib.py
+++ b/scripts/irods/lib.py
@@ -684,6 +684,15 @@ def replica_exists_on_resource(session, logical_path, resource_name):
 
     return 'CAT_NO_ROWS_FOUND' not in out
 
+
+def get_replica_mtime(session, logical_path, replica_number):
+    return session.run_icommand(['iquest', '%s',
+        "select DATA_MODIFY_TIME where COLL_NAME = '{}' and DATA_NAME = '{}' and DATA_REPL_NUM = '{}'"
+        .format(os.path.dirname(logical_path),
+                os.path.basename(logical_path),
+                str(replica_number))])[0].strip()
+
+
 def iterfy(iterable):
     """Will return an iterable, even if input is a single item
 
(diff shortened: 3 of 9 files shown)


alanking (commented):
`test_ichksum` and `irods_logical_locking` (unit test) were failing due to unexpected error messages. As it turns out, the solution offered by isocpp.org does not seem to work for our compiler(s)/platform(s) ("mileage may vary"). I tried this out on Compiler Explorer, and things were working as expected, but then I saw some tests failing because votes were not coming out as 0 anymore in instances where they should have.

I'm going to remove this in favor of another function - `vote_is_zero` - which does the safe comparison with `vote::zero` like we do in some (but not all) places now. Should we choose to implement said comparison with something like this function in the future, now it is hidden behind the one interface.

Confirmed that failing tests are now passing, but doing another full suite run now to be sure.

server/core/include/irods/voting.hpp
```diff
@@ -22,6 +22,19 @@ float calculate(
     std::string_view curr_host,
     const irods::hierarchy_parser& parser);
 
+/// \brief Check to see whether \p _vote is equal to 0.
+///
+/// This function exists because comparing floating point values is fraught with peril.
+/// See: https://isocpp.org/wiki/faq/newbie#floating-point-arith
```
alanking, line 28: Remove this line

alanking: Done

alanking:
Unit and core tests passed

trel:
Nice - happy to see it squashed - this is very nice work.

alanking:
Went ahead and did the squash. Sorry if that messed with any reviews!

server/api/src/rsDataObjOpen.cpp
```diff
@@ -763,10 +763,15 @@ namespace
 
         if (hierarchy_for_open.empty()) {
             try {
+                // NOLINTNEXTLINE(hicpp-signed-bitwise)
+                const auto open_for_create = 0 != (_inp.openFlags & O_CREAT);
+                const auto open_for_write = 0 != getWriteFlag(_inp.openFlags);
+                const auto operation = open_for_create  ? irods::CREATE_OPERATION
+                                       : open_for_write ? irods::WRITE_OPERATION
+                                                        : irods::OPEN_OPERATION;
```
korydraughn, line 771: Consider reformatting or refactoring to use if-statements.

Chained ternary operators can hurt readability.

server/core/src/voting.cpp
```diff
@@ -247,4 +248,10 @@ float calculate(
     return vote;
 } // calculate
 
+auto vote_is_zero(float _vote) -> bool
+{
+    constexpr double epsilon = 0.00000001;
+    return _vote - epsilon <= vote::zero;
+} // vote_is_zero
```
korydraughn, line 255: So this is how we've always checked for zero, it's just that you've wrapped the logic so that it is in one location, correct?

scripts/irods/test/test_targeting_specific_replica_number.py
```diff
@@ -0,0 +1,356 @@
...
+            str(self.file_size), lib.get_replica_size(self.user, os.path.basename(self.logical_path), 0))
+        self.assertEqual(
+            str(self.file_size), lib.get_replica_size(self.user, os.path.basename(self.logical_path), 1))
+        self.assertEqual(str(1), lib.get_replica_status(self.user, os.path.basename(self.logical_path), 0))
+        self.assertEqual(str(1), lib.get_replica_status(self.user, os.path.basename(self.logical_path), 1))
+        self.user.assert_icommand(['ils', '-L', self.logical_path], 'STDOUT') # debugging
+
+
+    def tearDown(self):
+        self.user.assert_icommand(['irm', '-f', self.logical_path])
```
korydraughn, line 66: run_icommand?

scripts/irods/test/test_targeting_specific_replica_number.py
```diff
@@ -0,0 +1,356 @@
...
+    import unittest2 as unittest
+else:
+    import unittest
+
+from . import session
+from .. import lib
+from .. import test
+from ..test.command import assert_command
+
+class test_targeting_specific_replica_number__issue_6896(
```
korydraughn, line 15: Are we favoring lowercase instead of uppercase for class names now?
Is there some motivating factor behind this?

scripts/irods/test/test_targeting_specific_replica_number.py
```diff
@@ -0,0 +1,356 @@
...
+        self.assertEqual(str(0), lib.get_replica_status(self.user, os.path.basename(self.logical_path), 1))
+
+        self.user.assert_icommand(['ils', '-L', self.logical_path], 'STDOUT') # debugging
+
+        # set read weight to 0.1 (very low) on passthru with replica 1
+        self.admin.assert_icommand(['iadmin', 'modresc', self.pt2, 'context', 'read=0.1'])
+
+        # execute iget -n 1, assert that replica 1 is returned even though it will vote extremely low.
+        self.user.assert_icommand(['iget', '-n1', self.logical_path, local_iget_file])
+        assert_command(['diff', self.local_file, local_iget_file])
```
korydraughn, line 126: What is `assert_command`? Is that part of the unittest framework or do we provide that?

scripts/irods/test/test_targeting_specific_replica_number.py
```diff
@@ -0,0 +1,356 @@
...
+
+    def test_istream_n_returns_error_when_requested_replica_votes_0(self):
+        content = 'test_istream_n_returns_error_when_requested_replica_votes_0'
+        stream_error_msg = 'Error: Cannot open data object.'
+
+        try:
+            # Set write weight to 0.0 on passthru with replica 1
+            self.admin.assert_icommand(['iadmin', 'modresc', self.pt2, 'context', 'write=0.0'])
+
+            # Stream to the existing replica and confirm that it fails to update the mtime.
```
korydraughn, line 306: mtime?

plugins/resources/src/passthru.cpp
```diff
@@ -68,22 +68,32 @@
...
-            r.resc_hier().c_str());
-        r.vote(r.vote() * weight);
-    }
-} // apply_weight_to_object_votes
+    using log_resource = irods::experimental::log::resource;
+
+    auto apply_weight_to_object_votes(irods::plugin_context& _ctx, double _weight) -> void
+    {
+        std::string this_resource_name;
+        _ctx.prop_map().get<std::string>(irods::RESOURCE_NAME, this_resource_name);
```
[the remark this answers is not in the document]
alanking: It returns an `irods::error` object. I'll add a check here for it.

server/core/src/voting.cpp
alanking: Yes and no.

This is the way we checked for zero in several places in `irods_resource_redirect.cpp`. In other places, the vote value was compared directly against `0.0f` (or, more recently, `vote::zero`). You can see the specific instances elsewhere in the diff.

This function provides a thin abstraction so that the implementation can be swapped out later if we so choose. I think comparing against zero should be consistent, in any case.

scripts/irods/test/test_targeting_specific_replica_number.py
alanking: As near as I can tell, the only reason we use uppercase letters in these classes is because you can't name the class the same thing as the file in the `unittest` framework.

I personally hate the mixed-case naming convention, which is why I opted for this scheme. I have done this several times with already-merged test classes, so I figured it was okay.

I am willing to change this to whatever everybody finds most appropriate.

scripts/irods/test/test_targeting_specific_replica_number.py
alanking: We provide it: https://github.com/irods/irods/blob/ee3a689971435d81e12aef6500d27ff33a9c2a01/scripts/irods/test/command.py#L76-L77

scripts/irods/test/test_targeting_specific_replica_number.py
korydraughn: It is fine. I only brought it up since I have seen different schemes used in the codebase. Likely an indicator that we need more convention, but certainly not an urgent matter right now.

Please carry on.

server/core/src/voting.cpp
alanking: All of these comparisons have already been replaced by calls to `vote_is_zero` in this PR. I can put them back and make a new issue to address this at a later time if that seems better. Or are you referring to all floating point comparisons in the codebase?

scripts/irods/test/test_targeting_specific_replica_number.py
alanking: I chose `assert_command` because `diff` is not an iCommand, so `self.<some_user>.assert_icommand` seemed unfitting (although I have no reason to believe it wouldn't do the same thing). There is no member of `IrodsSession` which serves an equivalent role, as far as I can tell.

server/core/src/voting.cpp
korydraughn: No need to revert anything.

I was under the impression that there were other places outside of your PR that dealt with voting, but compared floats incorrectly. It sounds like you've already addressed all known locations though.

alanking: Oh, I understand now. Yeah, I didn't really look outside of `irods_resource_redirect.cpp`. Looks like some of the resource plugins are making direct comparisons, so I made an issue to address those places: https://github.com/irods/irods/issues/6932

korydraughn: Cool cool

plugins/resources/src/passthru.cpp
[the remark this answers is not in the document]
alanking: Well, I went the other direction and am now using a throwing function. Please let me know if returning an `irods::error` seems more appropriate.

server/api/src/rsDataObjOpen.cpp
```diff
@@ -763,10 +763,15 @@ namespace
 
         if (hierarchy_for_open.empty()) {
             try {
+                // NOLINTNEXTLINE(hicpp-signed-bitwise)
+                const auto open_for_create = 0 != (_inp.openFlags & O_CREAT);
+                const auto open_for_write = 0 != getWriteFlag(_inp.openFlags);
```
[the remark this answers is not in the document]
alanking: Done

server/api/src/rsDataObjOpen.cpp
alanking: Done

scripts/irods/test/test_targeting_specific_replica_number.py
alanking: Done

scripts/irods/test/test_targeting_specific_replica_number.py
alanking: Done

plugins/resources/src/passthru.cpp
```diff
@@ -784,32 +789,36 @@ irods::error passthru_file_resolve_hierarchy(
...
+            }
+            else {
+                // NOLINTNEXTLINE(bugprone-narrowing-conversions,cppcoreguidelines-narrowing-conversions)
+                (*_out_vote) *= write_weight;
+                apply_weight_to_object_votes(_ctx, write_weight);
+            }
         }
     }
+    catch (const irods::exception& e) {
+        log_resource::error("Error occurred while applying weight to vote: [{}]", e.client_display_what());
```
alanking, line 820: I would print the name of the resource here, but getting the name of the resource may be the very thing which causes the exception!

plugins/resources/src/passthru.cpp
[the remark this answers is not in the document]
korydraughn: Will take a look.

plugins/resources/src/passthru.cpp
korydraughn: Perhaps you can capture it earlier in the function?

plugins/resources/src/passthru.cpp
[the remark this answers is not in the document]
korydraughn: Looks good to me.

korydraughn (commented):
Changes look good. Anything else left to do for this?

plugins/resources/src/passthru.cpp
alanking: Hmm... that's true. I'll play around with that.

alanking:
Other than the pending comment, I need to run through tests one more time, I think.

alanking:
Well, ran into some issues regarding the "prefer archive" policy configuration for the compound resource, resulting in a couple of test failures. I think we arrived at a couple of good options for solving that, and I have confirmed that the tests which were failing are now passing with those changes.

However, when I tried running the full test suite, I observed some other very unrelated tests failing (e.g. `test_imeta_help`) and, after observing failures in the tests for known-working commits, have concluded that my workstation is having issues. I'll follow up once I get the tests to pass.

alanking:
Tests are now passing. I rebuilt the base images and things healed up. Latest commit shows the vote-settting logic.

plugins/resources/src/compound.cpp
```diff
@@ -1626,6 +1626,26 @@ irods::error open_for_prefer_archive_policy(
...
 
+    std::string cache_name;
+    if (const auto err = _ctx.prop_map().get<std::string>(CACHE_CONTEXT_TYPE, cache_name); !err.ok()) {
+        return PASS(err);
+    }
+
+    try {
+        // Set the vote for the replica in the cache resource to be the same as the vote for the replica in the archive
+        // because the replica in the cache is going to be overwritten by the replica in the archive anyway. The vote
+        // must be set for the cache replica because the user may have requested the replica in the cache replica
```
trel, line 1637: "replica in the cache replica" -> "replica in the cache"

plugins/resources/src/compound.cpp
```diff
@@ -1626,6 +1626,26 @@ irods::error open_for_prefer_archive_policy(
...
+        // specifically and the requested replica must vote higher than 0 or else the operation will fail with an error.
+        // If the cache does not have a replica, there is no replica number to refer to and an error will occur if the
+        // user tries to request a non-existent replica.
+        if (const auto r = f_ptr->get_replica(resc_mgr.get_hier_to_root_for_resc(cache_name)); r) {
```
korydraughn, line 1641: Is `r` a pointer?

If it is, consider appending an asterisk to `auto` so that it is clear to the reader what `r` is.

korydraughn: No it's fine. I figured it was either a pointer or something that offered pointer-like semantics.

Carry on.

plugins/resources/src/compound.cpp
alanking: Done

plugins/resources/src/compound.cpp
alanking: Okey dokey. Resolving!

korydraughn (commented):
If the tests are passing, please squash without the pound.

alanking:
Squashed. This includes a rebase, so I will run tests again soon.

Plus, I promised @JustinKyleJames I would hold off on merging this until #6933 is merged since there may be conflicts :)

alanking:
Rebased, running tests again

alanking:
Unit tests and core tests passed

scripts/irods/lib.py
```diff
@@ -751,3 +760,12 @@ def make_arbitrary_file(f_name, f_size, buffer_size=32*1024*1024):
...
             bytes_written += to_write
+
+def get_resource_parent(session, resource_name):
+    parent_id = session.run_icommand(['iquest', '%s',
+        "select RESC_PARENT where RESC_NAME = '{}'"
+        .format(resource_name)])[0].strip()
+
+    return session.run_icommand(['iquest', '%s',
+        "select RESC_NAME where RESC_ID = '{}'"
+        .format(parent_id)])[0].strip()
```
korydraughn, line 771: This could return a tuple containing the resc id and name. If not, then consider adding `_name` to the function name.

Feel free to ignore this.

alanking: I think I'd rather split it in two, in that case: `get_resource_parent_name` and `get_resource_parent_id`. Hold please.

alanking: Well, before I make the change, just know that this has already been merged in 4-2-stable as-is. Are we okay with the split?

korydraughn: Let's leave it as is.

Consider creating an issue to revisit this in main and 4-3-stable.

alanking: Yep, makes sense. Created https://github.com/irods/irods/issues/7029.

korydraughn (approved):
If you're happy with it, please pound and merge when ready.

Issue #6900, opened 2023-02-10
Removing collections without forcing throws `specificQuery status = -853000` in logs


mstfdkmn:
- [x] main (code change only)
- [x] 4-3-stable (code change only)
- [x] 4-2-stable (docs only)
---
### iRODS Version, OS and Version
4.2.11 centos7, 4.3.0 almalinux8

### What did you try to do?
I tried to remove a collection (no matter that collection has any content) without using `f` option by `irm r` and PRC. Meaning I tried to move it to my trash.

### Expected behavior
I expected that collection to be removed without any noise.

### Observed behavior (including steps to reproduce, if applicable)
I observed that collection is removed (moved to the trash collection) with info level logs outputs concerning with `specificQuery`.

client:
```
imkdir test
irm -r test
```
logs of 4.3.0:
```
irods stdout | {"log_category":"legacy","log_facility":"local0","log_level":"info","log_message":"_rsSpecificQuery: specificQuery status = -853000","request_api_name":"RM_COLL_AN","request_api_number":679,"request_api_version":"d","request_client_user":"rods","request_host":"10.0.2.100","request_proxy_user":"rods","request_release_version":"rods4.3.0","server_host":"dev-u0137480-instance1.irods.t.icts.kuleuven.be","server_pid":84121,"server_timestamp":"2023-02-10T09:57:23.821Z","server_type":"agent"}
irods stdout | {"log_category":"legacy","log_facility":"local0","log_level":"info","log_message":"rsSpecificQuery: rcSpecificQuery failed, status = -853000","request_api_name":"RM_COLL_AN","request_api_number":679,"request_api_version":"d","request_client_user":"rods","request_host":"10.0.2.100","request_proxy_user":"rods","request_release_version":"rods4.3.0","server_host":"dev-u0137480-instance1.irods.t.icts.kuleuven.be","server_pid":84121,"server_timestamp":"2023-02-10T09:57:23.821Z","server_type":"agent"}
[...]

PR #6934, opened 2023-02-22
[#6900] Restored specific query, DataObjInCollReCur, for MySQL 8 (main)

korydraughn:
This PR has not been tested.

Notice the bump to the database schema version and the importing of the unittest python module.

### DIFF
--- a/scripts/irods/database_upgrade.py
+++ b/scripts/irods/database_upgrade.py
@@ -117,6 +117,19 @@ def run_update(irods_config, cursor):
         database_connect.execute_sql_statement(cursor, "update R_TOKN_MAIN set token_name = 'read_object' where token_name = 'read object';")
         database_connect.execute_sql_statement(cursor, "update R_TOKN_MAIN set token_name = 'read_metadata' where token_name = 'read metadata';")
 
+    elif new_schema_version == 10:
+        if irods_config.catalog_database_type == 'mysql':
+            # Assume the database is MySQL 8 or later (support for the SQL WITH-clause was not added until MySQL 8).
+            if len(database_connect.execute_sql_statement(cursor, "select alias from R_SPECIFIC_QUERY where alias = 'DataObjInCollReCur';").fetchall()) == 0:
+                sql = ("insert into R_SPECIFIC_QUERY (alias, sqlStr, create_ts) "
+                       "values ('DataObjInCollReCur', "
+                                "'with COLL as (select coll_id, coll_name from R_COLL_MAIN where R_COLL_MAIN.coll_name = ? or R_COLL_MAIN.coll_name like ?)"
+                                " select distinct d.data_id, (select coll_name from COLL where COLL.coll_id = d.coll_id) coll_name, d.data_name, d.data_repl_num, d.resc_name, d.data_path, d.resc_id"
+                                " from R_DATA_MAIN d"
+                                " where d.coll_id = any(select coll_id from COLL) order by coll_name, d.data_name, d.data_repl_num', "
+                                "'1580297960');")
+                database_connect.execute_sql_statement(cursor, sql)
+
     else:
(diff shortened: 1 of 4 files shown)


scripts/irods/database_upgrade.py
```diff
@@ -117,6 +117,19 @@ def run_update(irods_config, cursor):
         database_connect.execute_sql_statement(cursor, "update R_TOKN_MAIN set token_name = 'read_object' where token_name = 'read object';")
         database_connect.execute_sql_statement(cursor, "update R_TOKN_MAIN set token_name = 'read_metadata' where token_name = 'read metadata';")
 
+    elif new_schema_version == 10:
+        if irods_config.catalog_database_type == 'mysql':
+            # Assume the database is MySQL 8 or later (support for the SQL WITH-clause was not added until MySQL 8).
+            if len(database_connect.execute_sql_statement(cursor, "select alias from R_SPECIFIC_QUERY where alias = 'DataObjInCollReCur';").fetchall()) == 0:
```
trel, line 123: there is also the syntax of `INSERT … ON DUPLICATE KEY UPDATE`

as an option.

this is fine as is - easily readable - two queries instead of one, but that's not a problem.

korydraughn: I thought about searching for alternative syntax that makes this a one liner, but decided against it because it didn't seem worth it. And I doubt most people on the team would understand the syntax without having to google it (including myself).

Willing to change what I have already if others feel it's worth it.

trel: Not worth it to me.

scripts/irods/test/test_specific_queries.py
```diff
@@ -0,0 +1,24 @@
...
+    def setUp(self):
+        super(Test_Specific_Queries, self).setUp()
+        self.admin = self.admin_sessions[0]
+
+    def tearDown(self):
+        super(Test_Specific_Queries, self).tearDown()
+
+    def test_DataObjInCollReCur__issue_6900(self):
+        collection = os.path.join(self.admin.session_collection, 'test.d')
+        self.admin.assert_icommand(['imkdir', 'test.d'])
```
alanking, line 18: Would it be worth running this as a `rodsuser` in addition to/instead of `rodsadmin`? We've run into some issues in the past by only testing for `rodsadmin`s. Totally fine if not

alanking (commented):
Looks good to me overall - just had one thought. I guess we just need to make sure to capture the DB schema bump in the release notes for 4.3.1, right?

Also, just noting for myself...

Companion PR: https://github.com/irods/irods_docs/pull/191

scripts/irods/test/test_specific_queries.py
```diff
@@ -0,0 +1,24 @@
+import os
+import sys
```
alanking, line 2: Ooh, Codacy says this import is unused. :p

korydraughn:
> I guess we just need to make sure to capture the DB schema bump in the release notes for 4.3.1, right?

Yes. We can include that as part of a checklist or comment in the 4.3.1 release activities issue.

scripts/irods/test/test_specific_queries.py
korydraughn: I don't think it is necessary, but I do think that will improve the test and provide some future proofing for us.

Will change it to a rodsuser.

korydraughn: Done.

scripts/irods/test/test_specific_queries.py
korydraughn: Done.

korydraughn:
All unit tests and core tests pass.

Tested against Ub20 / MySQL 8.

korydraughn:
Any last comments before I add the pound?

alanking (approved):
\# it

korydraughn:
Added the pound.

Issue #6827, opened 2023-01-05
Only the first item in `catalog_provider_hosts` list is considered in negotiation key determination


alanking:
- [x] main
- [x] 4-3-stable
- [x] 4-2-stable
---
## Feature

Not sure whether this is a bug or a feature request...

The "catalog_provider_hosts" key for the JSON structure in the `federation` list in `server_config.json` is supposed to be a list of catalog service provider hostnames which can service federation requests. Only the first entry of the list is checked:
https://github.com/irods/irods/blob/36b22f0649932db8f4e331b619011c436df40746/lib/core/src/irods_client_negotiation.cpp#L51

This means that only the first hostname considered in this list. This makes setups which use multiple catalog provider hosts difficult to maintain. Like #6826, this can be worked around by adding a separate entry for each host, but the logic should iterate over the list to see if the given hostname is there rather than only checking the first entry.

PR #6950, opened 2023-03-07
[#6827] Search all catalog_provider_hosts in federation stanza (main)

alanking:
Addresses #6827

Tests are running now

### DIFF
--- a/lib/core/src/irods_client_negotiation.cpp
+++ b/lib/core/src/irods_client_negotiation.cpp
@@ -20,6 +20,7 @@
 
 // =-=-=-=-=-=-=-
 // stl includes
+#include <algorithm>
 #include <cstdlib>
 #include <map>
 #include <regex>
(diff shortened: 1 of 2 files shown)


alanking:
Core tests, unit tests, and federation tests are passing.

scripts/irods/test/test_federation.py
```diff
@@ -1128,6 +1128,40 @@ def test_itouch__issue_6849(self):
...
+        # Dump to a string to repave the existing server_config.json.
+        new_server_config = json.dumps(svr_cfg, sort_keys=True, indent=4, separators=(',', ': '))
+        with lib.file_backed_up(server_config_filename):
+            # Repave the existing server_config.json. It will be restored when we leave this 'with' block.
+            with open(server_config_filename, 'w') as f:
+                f.write(new_server_config)
+
+            # Ensure that we can still list the contents in the remote zone because the server negotiation logic looked
+            # at all the entries in the catalog_provider_hosts (not just the first one).
+            user.assert_icommand(['ils', '-l', remote_home_collection], 'STDOUT')
```
trel, line 1162: is this sufficient?   no particular output to check?

alanking: I confirmed that without my change, this triggers a `SERVER_NEGOTIATION_ERROR` (shameless green!). I feel that it is sufficient because we are just trying to test the server negotiation for connections made over federation boundaries.

If I had my druthers, we would have topologies on both sides of the federation and put an object to a resource which would force a redirection to the catalog consumer on the remote zone. Our testing environment and python tests do not support this at this time.

trel: gotcha - please put the `SERVER_NEGOTIATION_ERROR` part in the comment as what would happen on a non-success.

alanking: Done

lib/core/src/irods_client_negotiation.cpp
```diff
@@ -43,37 +44,31 @@ namespace irods
     /// @brief given a property map and the target host name decide between a federated key and a local key
     const std::string determine_negotiation_key( const std::string& _host_name )
```
korydraughn, line 45: I know this isn't part of the main work, but let's remove the `const` from the return type. It forces unnecessary copies.

korydraughn (commented):
Looks good.

Do we have tests (python, c++, etc) that trigger the `SERVER_NEGOTIATION_ERROR`?

lib/core/src/irods_client_negotiation.cpp
alanking: Done

alanking:
> Do we have tests (python, c++, etc) that trigger the `SERVER_NEGOTIATION_ERROR`?

Not presently. I'll think about what that might look like.

Initial thought is that we simply modify the server configuration to remove the actual hostname of the catalog provider host from the list in the federation list entry and try running an `ils`.

korydraughn:
Yeah. I think having a test that shows you get a negotiation error if any of the federation config options is incorrect is a good thing.

Willing to discuss more if others feel differently.

alanking:
I have added a section to the existing test during which `ils` is run when the legitimate provider host is not in the list in order to ensure that `SERVER_NEGOTIATION_ERROR` is returned. I can add other cases but that might take a while (more tables!). Let me know what you think

trel (approved):

korydraughn:
I think handling the other cases is a good think, but not of the highest priority. Let's capture it in an issue and flesh it out in 4.3.

alanking:
I've created https://github.com/irods/irods/issues/6960 for expanding these tests.

korydraughn (approved):

Issue #6956, opened 2023-03-10
`test_configuration_reload` fails in topology


alanking:
- [x] main
- [x] 4-3-stable
---
This test only exists in 4.3.x. Fails for both provider and consumer.

alanking:
After fixing `test_configuration_reload`, the other tests passed. Changing issue title and description...

trel:
!

PR #6957, opened 2023-03-10
Fix test_reload_command__issue_6529 (main)

alanking:
Addresses #6956

There are two commits in this pull request. They are alternative solutions, so only one should be accepted. Please read the commit messages for explanations.

### DIFF
--- a/scripts/irods/test/test_configuration_reload.py
+++ b/scripts/irods/test/test_configuration_reload.py
@@ -2,6 +2,7 @@ import unittest
 
 import json
 import os
+import socket
 
 from . import session
 from ..configuration import IrodsConfig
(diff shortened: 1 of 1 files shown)


trel:
I think I prefer the second one - continue using the `--all` flag, but only assert against the local server changes.

I think the commit message should be tweaked a little...

```diff
- return a result with the configuration is reloaded. This test only cares
- about the changes made the configuration on the local server.
+ return a set of changes. This test only cares about the changes
+ made to the configuration on the local server.
```

korydraughn:
I vote for the second one as well.

korydraughn:
Have you considered skipping this test when running in a topology?
Should we consider adding a brand new test that is designed for running in a topology?

alanking:
> Have you considered skipping this test when running in a topology?

I don't quite understand why we would. Fixing it was easy enough so that it now runs in both contexts.

> Should we consider adding a brand new test that is designed for running in a topology?

I don't mind writing a new test, but I would need somebody to tell me what the test should do... :)

The only thing I can think of to improve on this would be to edit server configurations on remote machines by editing the file and then running `irods-grid reload`. We do not have a mechanism for doing this in our test framework that I'm aware of at this time. If anything, the existing test would be sufficient in that it could assert that each server returns at least 1 result when `irods-grid reload --all` is run because we know the reCacheSalt will always be captured by loading the configuration.

Thoughts?

korydraughn:
Those questions are about trying to uncover holes in the logic (if any).

And based on your response, I think we're okay here. No need to change anything.

alanking:
Sounds good. I have force-pushed so that the chosen solution is the remaining commit and I have adjusted the commit message to match the diff from @trel's comment.

trel (approved):

Issue #3513, opened 2017-04-11
Clients can still write when quota is exceeded


kellerb:
- [x] main
- [x] 4-3-stable
- [x] 4-2-stable
-----

https://github.com/irods/python-irodsclient/issues/68

trel:
Is this still true?

korydraughn:
I've verified that quotas are enforced.

I'll write some tests for this before closing it.

korydraughn:
Closing as resolved.

See conversation starting at https://github.com/irods/irods/pull/6962#issue-1622286517 for additional info.

PR #6958, opened 2023-03-12
[#3513] Added use of istream to quota test case (main)

korydraughn:
`istream` attempts to write bytes to a data object using `rcDataObjWrite`.

Given that the PRC uses the stream operations, this update to the quota test should be enough to make sure we don't introduce any regressions around quotas and writing bytes.


alanking (approved):

korydraughn:
#'d

PR #6962, opened 2023-03-13
[#3513] Added use of istream to quota test case (4-2-stable)

korydraughn:
Thinking about this more, I wonder if the call to `rxDataObjOpen` is failing rather than the call to `rxDataObjWrite`.

I'm not sure we can do more for the issue without jumping through hoops because the quota isn't updated as the bytes are written.

To verify if the write operation respects the quota, we'd need to open the data object and then wait for the quota to be exceeded before executing the write operation. Doing that is difficult to do from within the python tests. Doing that from C++ is also difficult due to the lack of tools which allow modifying the server's configuration.

At the very least, we know that this test change increases coverage around quota enforcement.


alanking (approved):
Good point. I can't think of any way to do this in our existing test framework either. I agree that this test at least ensures that quota enforcement in the open-write-close path is being covered, so... good enough for now.

korydraughn:
I agree. Adding the pound.

alanking (approved):

PR #6964, opened 2023-03-14
[#6827] Search all catalog_provider_hosts in federation stanza (4-2-stable)

alanking:
Addresses https://github.com/irods/irods/issues/6827

Cherry-picked from https://github.com/irods/irods/pull/6950

Some adaptations had to be made for this cherry-pick due to how the server configuration interface changed between 4-2-stable and main/4-3-stable.

Core tests, unit tests, and federation tests passed

### DIFF
--- a/lib/core/src/irods_client_negotiation.cpp
+++ b/lib/core/src/irods_client_negotiation.cpp
@@ -22,6 +22,7 @@
 
 // =-=-=-=-=-=-=-
 // stl includes
+#include <algorithm>
 #include <cstdlib>
 #include <map>
 #include <regex>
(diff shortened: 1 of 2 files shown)


lib/core/src/irods_client_negotiation.cpp
```diff
@@ -43,39 +44,46 @@ namespace irods
 
     /// =-=-=-=-=-=-=-
     /// @brief given a property map and the target host name decide between a federated key and a local key
-    const std::string& determine_negotiation_key( const std::string& _host_name )
+    auto determine_negotiation_key(const std::string& _host_name) -> std::string
     {
-        // search the federation map for the host name
         try {
-            for ( const auto& el : irods::get_server_property<const std::vector<boost::any>>(irods::CFG_FEDERATION_KW) ) {
+            for (const auto& el : irods::get_server_property<const std::vector<boost::any>>(irods::CFG_FEDERATION_KW)) {
```
korydraughn, line 50: Should the vector be a reference?

korydraughn (commented):
Everything looks good. Just one comment.

lib/core/src/irods_client_negotiation.cpp
alanking: I was wondering this myself. I didn't change it because this was the original type, but I'll give it a try and see how it goes.

korydraughn: The good thing about this change is that the objects are marked as const so the compiler would've told us if we violated the constness of something (assuming we aren't doing anything weird in there lol).

korydraughn:
If you're happy with it, squash and add the pounds.

I don't think we need to rerun the core tests given that everything is const and that function doesn't modify the data structures in any way.

Feel free to rerun the core tests to be sure if you want though. It's never a bad idea to be sure.

korydraughn (approved):

alanking:
Yah, will do. Doesn't take too long these days

alanking:
Core tests and unit tests passed, adding #s and mergin

Issue #6912, opened 2023-02-15
need test for use of || and DATA_RESC_HIER


trel:
- [ ] main
- [ ] 4-3-stable
- [ ] 4-2-stable
-----

### iRODS Version, OS and Version

4.2.8

### What did you try to do?

iquest with `||` and `DATA_RESC_HIER`

See comment from @tedgin - https://github.com/irods/irods/issues/6101#issuecomment-1430679798

### Expected behavior

Query should return the correct result

### Observed behavior (including steps to reproduce, if applicable)

```
rods@dev ~? iquest "select RESC_NAME where DATA_RESC_HIER like 'ingestRes;%'"
RESC_NAME = ingest
------------------------------------------------------------
rods@dev ~? iquest "select RESC_NAME where DATA_RESC_HIER = 'ingestRes' || like 'ingestRes;%'"
CAT_NO_ROWS_FOUND: Nothing was found matching your query
```

This may already be mitigated by the work done for https://github.com/irods/irods/issues/6101.

The work here is to add a new test and confirm it passes before 4.2.12 is released.

alanking:
Just confirming that the changes for #6101 did not address this, so we will need to fix this in addition to adding a test.

alanking:
Also look into the `uppercase` option for `iquest` not working with `DATA_RESC_HIER`:
```bash
$ iquest uppercase "SELECT SUM(DATA_SIZE) WHERE DATA_RESC_HIER LIKE 'NOREPLROOT%'"
 ERROR: iquest Error: queryAndShowStrCond failed status = -806000 CAT_SQL_ERR
$ iquest "SELECT SUM(DATA_SIZE) WHERE DATA_RESC_HIER LIKE 'noReplRoot%'"
DATA_SIZE = 20
------------------------------------------------------------
```

PR #6967, opened 2023-03-15
[#6912] GenQuery: Add support for || with DATA_RESC_HIER (4-2-stable)

alanking:
Addresses #6912 

Cherry-picked/Adapted from #6955 

The adaptation was exclusively in the tests:
 - Replaced f-strings with `str.format`
 - Removed `unittest.subTest` because it's not available in python2's version of `unittest`

Core tests and unit tests passed


korydraughn (approved):
Let's get it merged.

Issue #6136, opened 2022-01-26
`iexit full` can remove `.irodsA` file for service account


alanking:
- [x] main (iexit)
- [x] 4-3-stable (iexit)
- [x] 4-2-stable (iexit)
- [x] main (irods_docs)
- [x] 4-2-stable (irods_docs)
---
It is possible to run `iexit full` as the iRODS service account. This removes the `.irodsA` file. The environment of the service account is the means by which the server communicates with itself, authenticated as the "main" `rodsadmin` iRODS user using the `.irodsA` file.

If `iinit` is run again before restarting the server, the service account's iRODS user can authenticate again and things return to normal. If the server is restarted after running `iexit full`, the server will stand up, but new connections cannot be established with it.

Should an administrator get into this situation, `iinit` can be run as usual and the password entered to generate the `.irodsA` file again after waiting for the connection to fail. The server can then be restarted and things will return to normal.

So, two things need to happen:
1. This situation should be documented in the Troubleshooting section of the documentation.
2. `iexit` should detect this situation and prevent it from happening.

alanking:
@SwooshyCueb - Are you planning to update the documentation for this issue? If not, I can take that piece. Up to you.

SwooshyCueb:
@alanking I'll let you handle that, thanks.

Also, I just noticed I forgot to put a newline at the end of the `fprintf` in the catch block of `iexit` so let's untick those boxes until I get PRs in to fix that.

SwooshyCueb:
We can tick the boxes and close this issue now

korydraughn:
One of the `iexit` tests is failing. I have the fix in my JSON PR. Let's leave this open until the JSON microservice PR is merged into 4-2-stable.

SwooshyCueb:
roger dodger

korydraughn:
Then again, I'm going to add a new test that shows that a warning is produced when `iexit full` is run.

korydraughn:
This can now be closed. Go team!

PR #6969, opened 2023-03-15
[#6136] add irods::is_service_account function (main)

SwooshyCueb:
Prerequisite for irods/irods_client_icommands#379
In service of #6136

Will backport to stable branches after reviews done.


lib/core/include/irods/irods_service_account.hpp
```diff
@@ -0,0 +1,12 @@
+#ifndef __IRODS_SERVICE_ACCOUNT_HPP__
+#define __IRODS_SERVICE_ACCOUNT_HPP__
+
+namespace irods
+{
+    /// Attempts to determine whether the caller is running as the irods service unix account.
+    ///
+    /// \since 4.3.1
```
alanking, line 8: A long-standing question I've always had...

Do we put 4.2.12 here, or is 4.3.1 right for main/4-3-stable and 4.2.12 is just for 4-2-stable?

lib/core/src/irods_service_account.cpp
```diff
@@ -0,0 +1,132 @@
...
+#include <unistd.h>
+
+#include <boost/filesystem.hpp>
+#include <boost/algorithm/string.hpp>
+#include <fmt/format.h>
+
+namespace irods
+{
+    const std::string SERVICE_ACCOUNT_CONFIG_FILE("service_account.config");
+    const std::string SERTICE_ACCOUNT_KW("IRODS_SERVICE_ACCOUNT_NAME");
```
alanking, line 26:
```suggestion
    const std::string SERVICE_ACCOUNT_KW("IRODS_SERVICE_ACCOUNT_NAME");
```

lib/core/src/irods_service_account.cpp
```diff
@@ -0,0 +1,132 @@
...
+                break;
+            case EINTR:
+                THROW(SYS_CAUGHT_SIGNAL - pwd_e, "caught signal");
+                break;
+            default:
+                THROW(SYS_UNKNOWN_ERROR - pwd_e, "unknown error when calling getpwnam_r");
+                break;
+        }
+        if (svc_acct_pwdp == nullptr) {
+            irods::log(LOG_WARNING, fmt::format("%s - could not find user with name %s", service_account_username));
```
alanking, line 127: `%s`'s should be `{}`, also missing a parameter (`__PRETTY_FUNCTION__`?)

lib/core/include/irods/irods_service_account.hpp
```diff
@@ -0,0 +1,12 @@
+#ifndef __IRODS_SERVICE_ACCOUNT_HPP__
+#define __IRODS_SERVICE_ACCOUNT_HPP__
+
+namespace irods
+{
+    /// Attempts to determine whether the caller is running as the irods service unix account.
+    ///
```
alanking, line 7: Please note that this function `\throws irods::exception`

alanking (commented):
Nice. Would it be worth making an issue for the service_account file parsing?

lib/core/include/irods/irods_service_account.hpp
trel: since it wasn't in 4.3.0... putting 4.2.12 is a tad misleading.

lib/core/src/irods_service_account.cpp
SwooshyCueb: nice catch

lib/core/src/irods_service_account.cpp
```diff
@@ -0,0 +1,132 @@
...
+        if (_pwd_init_len >= PWD_BUF_LEN_MINIMUM) {
+            pwd_buf_len = static_cast<size_t>(_pwd_init_len);
+        }
+        int pwd_e = 0;
+        while (true) {
+            // this sure is a lot of fuss to allocate a buffer for data we're not going to use anyway...
+            const std::unique_ptr<char[]> pwd_buf( // NOLINT(cppcoreguidelines-avoid-c-arrays, modernize-avoid-c-arrays)
+                new char[pwd_buf_len]);
+            pwd_e =
+                getpwnam_r(service_account_username.c_str(), &svc_acct_pwd, pwd_buf.get(), pwd_buf_len, &svc_acct_pwdp);
```
trel, line 99: this formatting looks off to me...

lib/core/include/irods/irods_service_account.hpp
SwooshyCueb: There's no reason we can't put both. From doxygen docs:
> The paragraph that follows \since does not have any special internal structure.

korydraughn: I normally put the earliest release where it is available.

lib/core/include/irods/irods_service_account.hpp
```diff
@@ -0,0 +1,12 @@
+#ifndef __IRODS_SERVICE_ACCOUNT_HPP__
+#define __IRODS_SERVICE_ACCOUNT_HPP__
```
korydraughn, line 2: Let's remove the leading and trailing underscores.

lib/core/include/irods/irods_service_account.hpp
```diff
@@ -0,0 +1,12 @@
...
+
+namespace irods
+{
+    /// Attempts to determine whether the caller is running as the irods service unix account.
+    ///
+    /// \since 4.3.1
+    auto is_service_account() -> bool;
+} //namespace irods
+
+#endif
```
korydraughn, line 12: Missing trailing comment: `// IRODS_SERVICE_ACCOUNT_HPP`

lib/core/src/irods_service_account.cpp
korydraughn, line 26: These can be moved into an anonymous namespace so that the compiler marks them as hidden symbols in the binary.

lib/core/src/irods_service_account.cpp
```diff
@@ -0,0 +1,132 @@
...
+#include <boost/filesystem.hpp>
+#include <boost/algorithm/string.hpp>
+#include <fmt/format.h>
+
+namespace irods
+{
+    const std::string SERVICE_ACCOUNT_CONFIG_FILE("service_account.config");
+    const std::string SERTICE_ACCOUNT_KW("IRODS_SERVICE_ACCOUNT_NAME");
+
+    const size_t PWD_BUF_LEN_DEFAULT = 16384; // default size of buffer for strings in struct passwd
```
korydraughn, line 28: `size_t` -> `std::size_t`

I think I saw some other cases below as well.

lib/core/src/irods_service_account.cpp
```diff
@@ -0,0 +1,132 @@
...
+    const int PWD_BUF_LEN_MINIMUM = 128; // minimum size of buffer for strings in struct passwd
+
+    // FIXME: Parsing of service_account.config isn't super robust.
+    auto is_service_account() -> bool
+    {
+        std::string service_account_username;
```
korydraughn, line 34: This isn't used until much later in the function. Let's move it closer to where it is needed.

lib/core/src/irods_service_account.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+        std::ifstream svc_acct_cfg_strm{svc_acct_cfg_fn};
+        if (!svc_acct_cfg_strm.is_open()) {
+            // if we can't open the config file, it's likely due to a permission issue
+            // meaning it is extremely unlikely we are the service account
+            return false;
+        }
+
+        kvp_map_t service_acct_cfg;
+        uint_fast64_t cfg_line_no = 0;
```
korydraughn, line 52: Why use `uint_fast64_t` instead of `std::uint64_t`?

lib/core/src/irods_service_account.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+            // For now, let's just re-use our kvp-parsing functionality
+            // FIXME: does not handle spaces around "=", quotes, inline comments, or escape-newline
+            ret = parse_escaped_kvp_string(cfg_line, service_acct_cfg);
+            if (!ret.ok()) {
+                irods::log(LOG_WARNING,
+                           fmt::format("{} - {} parse failure on line {}",
+                                       __PRETTY_FUNCTION__,
+                                       SERVICE_ACCOUNT_CONFIG_FILE,
+                                       cfg_line_no));
```
korydraughn, line 73: Why not use the new logger?

lib/core/src/irods_service_account.cpp
```diff
@@ -0,0 +1,132 @@
...
+            }
+        }
+
+        const auto& svc_acct_itr = service_acct_cfg.find(SERTICE_ACCOUNT_KW);
```
korydraughn, line 77: `const auto&` -> `const auto`

korydraughn, line 77: `SERTICE_ACCOUNT_KW` -> `SERVICE_ACCOUNT_KW`

lib/core/src/irods_service_account.cpp
```diff
@@ -0,0 +1,132 @@
...
+        const auto _pwd_init_len = sysconf(_SC_GETPW_R_SIZE_MAX);
+        size_t pwd_buf_len = PWD_BUF_LEN_DEFAULT; // a reasonable buffer size in case sysconf gives us something awful
+        if (_pwd_init_len >= PWD_BUF_LEN_MINIMUM) {
+            pwd_buf_len = static_cast<size_t>(_pwd_init_len);
+        }
+        int pwd_e = 0;
+        while (true) {
+            // this sure is a lot of fuss to allocate a buffer for data we're not going to use anyway...
+            const std::unique_ptr<char[]> pwd_buf( // NOLINT(cppcoreguidelines-avoid-c-arrays, modernize-avoid-c-arrays)
+                new char[pwd_buf_len]);
```
korydraughn, line 97: Does the following work?
```cpp
std::array<char, pwd_buf_len> pwd_buf;
pwd_e = getpwnam_r(..., pwd_buf.data());
```

lib/core/src/irods_service_account.cpp
korydraughn, line 127: Why not use the new logger?

lib/core/src/irods_service_account.cpp
```diff
@@ -0,0 +1,132 @@
...
+#include <cstdint>
+#include <cstdlib>
+#include <fstream>
+#include <memory>
+
+#include <pwd.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+#include <boost/filesystem.hpp>
```
korydraughn, line 19: Is this used?

korydraughn (commented):
Looking really good.

lib/core/src/irods_service_account.cpp
SwooshyCueb: Does the new logger actually print to stdout/stderr? I thought it was only for the server

lib/core/include/irods/irods_service_account.hpp
korydraughn: Putting both works well considering it isn't in 4.3.0.

lib/core/src/irods_service_account.cpp
SwooshyCueb: Unfortunately, no, already tried that

lib/core/src/irods_service_account.cpp
SwooshyCueb: Nope, good catch

lib/core/src/irods_service_account.cpp
SwooshyCueb: It just seemed like the thing to do. I can change it

lib/core/src/irods_service_account.cpp
korydraughn: This feels like a clang-format penalty break issue.

While the current formatting of that line isn't my ideal, I don't feel we should fight the formatter too much as there will always be things we don't like about it.

I'm fine with how it is now, but I'm also fine with you formatting that line differently and wrapping it in `clang-format off/on` comments if that improves readability.

SwooshyCueb: This is a `getpwnam_r` call we're talking about; it's pretty unreadable by default IMO. I'll leave it as-is.

lib/core/src/irods_service_account.cpp
korydraughn: Yeah, let's go with a more common type. If switching to std::uint64_t costs us something, we can address it later.

lib/core/src/irods_service_account.cpp
korydraughn: It doesn't. Is that important for this function?

SwooshyCueb:
> Would it be worth making an issue for the service_account file parsing?

That's the plan. I'm also going to open an issue for the `iexit` arguments.

lib/core/src/irods_service_account.cpp
korydraughn: I'm guessing it's because the buffer size isn't known at compile-time.

Carry on.

lib/core/src/irods_service_account.cpp
SwooshyCueb: I think it's reasonable for `iexit` to be able to let users know when we have trouble parsing `service_account.config`, or when the user specified in `service_account.config` doesn't seem to exist on their system. These aren't fatal errors, so the function needs some way to be able to indicate these problems to the user before moving on.

korydraughn: Gotcha. It wasn't clear to me if this work was for the client or server or both.

Is this a function you imagine being usable from the server-side as well?

korydraughn: Something to consider as well is that `irods::log()` and `rodsLog()` will be going away in main and 4-3-stable eventually.

Not a blocker at the moment though.

SwooshyCueb: > Is this a function you imagine being usable from the server-side as well?

Sure. It's a client-oriented function, but I wrote it in such a way that it could be called from the server if needed.

> Something to consider as well is that `irods::log()` and `rodsLog()` will be going away in main and 4-3-stable eventually.

I figured as much. Before that happens, we should have functionality for client-friendly logging though, right? Probably goes without saying, but I think it would be a bad idea to rip these out without something to replace them with.

lib/core/src/irods_service_account.cpp
SwooshyCueb: Marking this one as resolved since discussion is happening in the other one

lib/core/src/irods_service_account.cpp
korydraughn: I don't think that is something we should provide. Of course, willing to discuss more and be persuaded otherwise as that time approaches.

> Probably goes without saying, but I think it would be a bad idea to rip these out without something to replace them with.

Absolutely. We're going to provide as much cushion for the users (that makes sense) with whatever is decided.

SwooshyCueb: Definitely a topic for discussion. I'll mark this as resolved.

SwooshyCueb:
\#'d.

alanking (approved):

Issue #6931, opened 2023-02-22
Putting test file while setting up catalog service consumers fails


alanking:
### iRODS Version, OS and Version
main/4-3-stable

### What did you try to do?
Set up a catalog service consumer with `setup_irods.py`

### Expected behavior
Success

### Observed behavior (including steps to reproduce, if applicable)
The following error message occurs when running `irodsTestPutGet` at the end:
```python
Error encountered running setup_irods:
Traceback (most recent call last):
  File "/var/lib/irods/scripts/setup_irods.py", line 515, in main
    setup_server(irods_config,
  File "/var/lib/irods/scripts/setup_irods.py", line 148, in setup_server
    test_put(irods_config)
  File "/var/lib/irods/scripts/setup_irods.py", line 180, in test_put
    raise IrodsError('Post-install test failed. Please check your configuration.')
irods.exceptions.IrodsError: Post-install test failed. Please check your configuration.
```
The log in the catalog service provider has the following message:
```json
{
  "log_category": "agent_factory",
  "log_level": "error",
  "log_message": "Agent process [1653] terminated by signal [6].",
  "server_host": "1c71f51165e8",
  "server_pid": 1628,
  "server_timestamp": "2023-02-22T16:29:02.309Z",
  "server_type": "agent_factory",
  "server_zone": "tempZone"
}
```
An object is left in the intermediate status because the servicing agent terminated without a chance to finalize:
```bash
$ ils -lr
/tempZone/home/rods:
  rods              0 ddb1b959ff62Resource            0 2023-02-22.16:29 ? post_install_test.1677083341733464.txt
```
I have not tried starting up the catalog service consumer and putting other files to see what might happen, but I believe this would be possible.

This issue does not exist in 4-2-stable.

PR #6973, opened 2023-03-15
[#6931] Use clearDataObjInfo for reg data obj output struct cleanup (main)

alanking:
Addresses #6931 

Leaving in draft for now because I haven't fully investigated the ramifications of this change. I know that it fixes setting up catalog consumers and I've even been able to run topology tests (somewhat) successfully.


korydraughn:
The change looks correct to me. The API definitions are supposed to clear the input and output structures only. Not free them.

alanking:
I'm going to mark this ready for review. Setting up catalog consumers will be required for testing certain changes, so if there's no major concerns, I think we can go ahead and get this in. I can't think of any adverse effects from this change.

korydraughn (approved):
Ship it.

Issue #6974, opened 2023-03-16
control_plane CMakeLists.txt


ll4strw:
- [x] main
- [x] 4-3-stable
---
Hi,
I am wondering what is the rationale behind the choice of overwriting `LD_LIBRARY_PATH` in the following `CMakeLists.txt`. 

https://github.com/irods/irods/blob/90e42c04246d57a9182fe67d9675dd89487706eb/server/control_plane/CMakeLists.txt#L11

When compiling from source I get all sorts of problems because `avrogencpp` does not find the libraries on which it depends. Would not it be better perhaps to modify an existing `LD_LIBRARY_PATH` instead of overwriting it? The following works for me

```
COMMAND "${CMAKE_COMMAND}" -E env LD_LIBRARY_PATH="${IRODS_EXTERNALS_FULLPATH_BOOST}/lib:${IRODS_EXTERNALS_FULLPATH_CLANG_RUNTIME}/lib:${IRODS_EXTERNALS_FULLPATH_AVRO}/lib:$ENV{LD_LIBRARY_PATH}" "${IRODS_EXTERNALS_FULLPATH_AVRO}/bin/avrogencpp" -n irods -o "${CMAKE_CURRENT_BINARY_DIR}/include/irods/server_control_plane_command.hpp" -i "${CMAKE_IRODS_SOURCE_DIR}/schemas/messaging/v1/server_control_plane_command.json"
```
Shall I open a PR?

korydraughn:
This is the first time I've heard of this being an issue.

Please provide more information about how you're compiling iRODS.
Are you using packages from packages.irods.org or are you compiling everything from source?
What OS are you compiling on/for?

korydraughn:
When you say you're building **everything** from source, does that include the externals packages?

korydraughn:
Can you provide the EasyBuild script you're using for the externals packages and the iRODS server?

How does EasyBuild work? For instance, when it comes to building the externals packages, are you copying the instructions from the README into an EasyBuild script and running that or are you hand crafting something special?

korydraughn:
I understand. Is that one of the real scripts you're attempting to use?

If so, that isn't going to work because iRODS has very strict requirements on the tools and versions used. For example, iRODS does not compile using GCC. We are working on that, but it's going to be a while before that is available. See https://github.com/irods/externals/blob/main/versions.json. See the README of the externals repo for additional information.

[...]

PR #6978, opened 2023-03-19
See issue #6974

ll4strw:


trel:
do we need to set a default value?

we do that for most/all of the other cmake variables that we introduce/use.

korydraughn:
I thought all of those options for externals packages can be overwritten.

The option used to be something like `-DIRODS_EXTERNALS_FULLPATH_CLANG=/path/to/your/clang/compiler` or `-DIRODS_EXTERNALS_FULLPATH_FMT=/path/to/your/fmt`.

@ll4strw Have you tried using these variables to change which libraries are used?

I think we only need to adjust the `LD_LIBRARY_PATH` line?

@SwooshyCueb Any thoughts on this PR?

korydraughn:
> My bad. I was mistaken. Overwriting the variables -DIRODS_EXTERNALS_FULLPATH_* works fine.

Excellent.

> Changing only LD_LIBRARY_PATH should be enough. By the way what variables can be set to build only the `irods-runtime` and `irods-dev`?

Normally, you can get a list of targets by using tab completion on the call to `make` or `ninja`.

My guess is you can do something like the following:
```bash
make -j irods-dev irods-runtime
```

@SwooshyCueb Do you know of a better way to do this?

korydraughn:
@ll4strw Please update the PR to only modify the `LD_LIBRARY_PATH`.

Please adjust the commit so that it resembles the following (leave off the pound until the PR is approved).
```
[6974] Fixed LD_LIBRARY_PATH for avrogencpp.
```

Issue #6976, opened 2023-03-16
Topology tests are failing on all branches


alanking:
- [x] 4-2-stable
---
The following tests fail when running topology tests on the catalog provider:
- [x] test_native_rule_engine_plugin
- [x] test_rulebase

The following tests fail when running topology tests on the catalog consumer:
- [x] test_all_rules
- [x] test_iadmin
- [x] test_native_rule_engine_plugin
- [x] test_rulebase

alanking:
See https://github.com/irods/irods/issues/7000 for main/4-3-stable fix.

PR #6995, opened 2023-03-23
[#6976] Fix tests for running in topology (4-2-stable)

alanking:
Addresses #6976 

Running tests:
- [x] Unit tests
- [x] Core tests
- [x] Federation tests
- [x] Topology on provider tests (2h 25m(!))
- [x] Topology on consumer tests (3h 10m(!))

NB: This change is not based on the tip of 4-2-stable at time of writing. My understanding is that tests are failing there, so this change will not be fixing those failing tests (unless I'm instructed otherwise).


korydraughn (approved):
Please add the pound when ready.

PR #7004, opened 2023-03-29
Microservices for JSON data / per-process key-value store API (4-2-stable)

korydraughn:
Ported from #6981.

### DIFF
--- a/plugins/microservices/CMakeLists.txt
+++ b/plugins/microservices/CMakeLists.txt
@@ -7,6 +7,7 @@ add_subdirectory(msi_get_agent_pid)
 add_subdirectory(msi_touch)
 add_subdirectory(msi_get_open_data_obj_l1desc_index)
 add_subdirectory(msi_get_file_descriptor_info)
+add_subdirectory(msi_json)
 
 if (IRODS_MICROSERVICE_TEST_PLUGINS_BUILD OR IRODS_ENABLE_ALL_TESTS)
   add_subdirectory(msi_tests)
--- a/plugins/microservices/msi_json/CMakeLists.txt
+++ b/plugins/microservices/msi_json/CMakeLists.txt
@@ -1,0 +1,57 @@
+set(
+  IRODS_MICROSERVICE_PLUGINS
+  json_compare
+  json_contains
+  json_dump
+  json_free
+  json_handles
+  json_names
+  json_parse
+  json_size
+  json_type
+  json_value
+)
+
+foreach (plugin IN LISTS IRODS_MICROSERVICE_PLUGINS)
+  set(plugin_name "msi_${plugin}")
+  
+  add_library(
+    ${plugin_name}
+    MODULE
+    ${CMAKE_CURRENT_SOURCE_DIR}/${plugin}.cpp)
+  
+  target_compile_definitions(
+    ${plugin_name}
+    PRIVATE
+    ENABLE_RE
+    ${IRODS_COMPILE_DEFINITIONS}
+    IRODS_ENABLE_SYSLOG)
+  
+  target_include_directories(
+    ${plugin_name}
+    PRIVATE
+    ${CMAKE_BINARY_DIR}/lib/core/include
+    ${CMAKE_SOURCE_DIR}/lib/core/include
+    ${CMAKE_SOURCE_DIR}/lib/api/include
+    ${CMAKE_SOURCE_DIR}/server/drivers/include
+    ${CMAKE_SOURCE_DIR}/server/api/include
+    ${CMAKE_SOURCE_DIR}/server/core/include
+    ${CMAKE_SOURCE_DIR}/server/icat/include
+    ${CMAKE_SOURCE_DIR}/server/re/include
+    ${IRODS_EXTERNALS_FULLPATH_BOOST}/include
+    ${IRODS_EXTERNALS_FULLPATH_FMT}/include)
+  
+  target_link_libraries(
+    ${plugin_name}
+    PRIVATE
+    irods_server
+    irods_common
+    ${IRODS_EXTERNALS_FULLPATH_BOOST}/lib/libboost_filesystem.so
+    ${IRODS_EXTERNALS_FULLPATH_BOOST}/lib/libboost_system.so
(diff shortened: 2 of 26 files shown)


alanking:
It would be very helpful if you could highlight sections which were changed in porting this across standards. Would help with reviewing more quickly so I don't have to comb through all the changes again :)

plugins/microservices/CMakeLists.txt
korydraughn, line 1: Different from main branch.

Doesn't exist in other branches.

plugins/microservices/msi_json/CMakeLists.txt
korydraughn, line 1: Different from main branch.

Doesn't exist in other branches.

plugins/microservices/msi_json/json_common.hpp
korydraughn, line 1: This file does not include the truncation detection function found in main. Hence, none of the microservices for 4-2-stable report when truncation occurs.

plugins/microservices/msi_json/json_parse.cpp
```diff
@@ -0,0 +1,148 @@
...
+                }
+
+                std::vector<char> buf;
+                buf.reserve(buf_size);
+
+                std::transform(int_array->value, int_array->value + buf_size, std::back_inserter(buf), [](auto _v) {
+                    return static_cast<char>(_v);
+                });
+
+                handle = irods::process_stash::insert(json::parse(std::begin(buf), std::end(buf)));
```
korydraughn, line 76: Here's a significant change from main.

4-2-stable uses a version of the JSON lib that requires that all elements of a container you wish to parse have a size of 1 byte and be contiguous. Because of that, we are required to **copy** the range if it represents an array containing integers.

Ultimately, integer arrays consume double the amount of memory when this microservice is invoked.

I'm going to add a note to the documentation about that.

plugins/microservices/msi_json/json_value.cpp
```diff
@@ -0,0 +1,141 @@
...
+/// \brief Fetches a value within a JSON structure.
+///
+/// \param[in]     _json_handle  The handle of a JSON structure obtained via msi_json_parse().
+/// \param[in]     _json_pointer \parblock
+/// A JSON pointer to the element of interest. If the element cannot be represented in the iRODS
+/// Rule Language, its type will be written to \p _value instead.
+///
+/// The following type strings can be returned:
+/// - array
+/// - binary
```
korydraughn, line 120: Remove this line.

The JSON lib (v3.7.3) does not support binary data.

scripts/irods/test/test_all_rules.py
```diff
@@ -1651,3 +1650,309 @@ def test_chksum_true_forceChksum_1_verifyChksum_1__issue_5927(self):
...
+        rep_name = 'irods_rule_engine_plugin-irods_rule_language-instance'
+        expected_output = ['result_1=[1]', 'result_2=[0]']
+        self.user.assert_icommand(['irule', '-r', rep_name, rule_text, 'null', 'ruleExecOut'], 'STDOUT', expected_output)
+
+    @unittest.skipUnless(plugin_name == 'irods_rule_engine_plugin-irods_rule_language', 'Only applicable to the NREP')
+    def test_msi_json_dump__issue_6968(self):
+        json_string = json.dumps({
+            'fname': 'john',
+            'lname': 'doe'
+        }, separators=(',', ':'), sort_keys=True)
```
korydraughn, line 1761: Had to add `sort_keys=True` here.

This wasn't necessary for the Python 3 version. I'll double check though.

scripts/irods/test/test_all_rules.py
```diff
@@ -6,6 +6,7 @@
 import getpass
 import tempfile
 import json
+import textwrap
```
korydraughn, line 9: Didn't need this for 4.3. Perhaps it was already imported.

plugins/microservices/msi_json/json_value.cpp
trel: maybe even better to say that binary is explicitly not supported, for that reason.  removes any doubt in the reader.

korydraughn:
All unit tests pass.

Running core tests.

plugins/microservices/msi_json/json_parse.cpp
alanking: Makes sense.

scripts/irods/test/test_all_rules.py
alanking: That is strange. It seems like if you define them in alphabetical order that's just how they would appear...

Or is the problem more with the expected output? Hmm...

alanking (commented):
Thanks for doing that. Everything looks good to me, pending the tweaks you mentioned

scripts/irods/test/test_all_rules.py
trel: the json spec explicitly says it's not sorted... as a format.

alanking: So the test in main/4-3-stable has a 50% chance of failing since `sort_keys` is `False` (default)?

trel: Python3.7+ remembers sort order by default.

korydraughn: Marking as resolved.

scripts/irods/test/test_all_rules.py
korydraughn: Nothing more to do here.

plugins/microservices/msi_json/json_value.cpp
korydraughn: Done.

plugins/microservices/msi_json/json_parse.cpp
korydraughn: Done.

korydraughn:
Added a new `iexit` test.

Investigating `irepl` test failure. Could be timing-based.

alanking:
The `irepl` test is known to be failing. It is being resolved by #7003.

korydraughn:
Squashed everything.

alanking (approved):
\# em.

PR #7005, opened 2023-03-29
[6675] db_[mod|reg]_zone_op: validate well-formed connection strings (main)

SwooshyCueb:
WIP


plugins/database/src/db_plugin.cpp
```diff
@@ -5089,6 +5090,30 @@ irods::error db_reg_zone_op(
...
+    const std::size_t addr_len = std::strlen(_zone_conn_info);
+    if (addr_len == 0) {
+        rodsHostAddr_t addr{};
+        // _zone_conn_info is const, so copy it
+        std::unique_ptr<char[]> addr_buf( // NOLINT(cppcoreguidelines-avoid-c-arrays, modernize-avoid-c-arrays)
+            new char[addr_len + 1]);
+        std::strncpy(addr_buf.get(), _zone_conn_info, addr_len + 1);
+        status = parseHostAddrStr(addr_buf.get(), &addr);
+        if (status < 0) {
+            log_db::info("chlRegZone failed to validate connection info {}", status);
```
korydraughn, line 5104: `{}` -> `[{}]`

That helps us detect empty strings by looking at the log file.

plugins/database/src/db_plugin.cpp
```diff
@@ -5089,6 +5090,30 @@ irods::error db_reg_zone_op(
...
+            return ERROR(status, "failed to validate connection info");
+        }
+        if (std::strlen(addr.hostAddr) == 0) {
+            log_db::info("chlRegZone failed to validate connection info due to empty hostname {}", status);
```
korydraughn, line 5108: `{}` -> `[{}]`

Perhaps we should log the connection string instead of `status`?
If `status` is negative, then the block above will catch it. What does it mean here?

_Same for the other locations._

plugins/database/src/db_plugin.cpp
```diff
@@ -5091,6 +5092,43 @@ irods::error db_reg_zone_op(
...
     }
 
+    // =-=-=-=-=-=-=-
+    // validate the connection string is well formed
+    const std::size_t addr_len = std::strlen(_zone_conn_info);
+    if (addr_len > 0) {
+        rodsHostAddr_t addr{};
+        // _zone_conn_info is const, so copy it
+        auto addr_buf = std::make_unique<char[]>(addr_len + 1);
+        std::strncpy(addr_buf.get(), _zone_conn_info, addr_len + 1);
```
korydraughn, line 5102: You can capture the `addr_len + 1` in a variable to guard against these lines becoming out of sync.

Alternatively, you can use a `std::vector` and avoid the size variable altogether.
```cpp
std::vector<char> addr_buf(addr_len + 1);
std::strncpy(addr_buf.data(), _zone_conn_info, addr_buf.size());
```

plugins/database/src/db_plugin.cpp
```diff
@@ -5225,7 +5263,45 @@ irods::error db_mod_zone_op(
...
+                             "chlRegZone failed to validate connection info due to invalid or unspecified port");
+                return ERROR(CAT_INVALID_ARGUMENT,
+                             "failed to validate connection info due to invalid or unspecified port");
+            }
+        }
+        else {
+            log_db::info("connection info empty or not specified");
+            addRErrorMsg(&_ctx.comm()->rError, CAT_INVALID_ARGUMENT, "connection info empty or not specified");
+            return ERROR(CAT_INVALID_ARGUMENT, "connection info empty or not specified");
+        }
```
korydraughn, line 5303: Is this block identical to the previous one?

If so, consider moving it (or parts of it) into a function so the logic is shareable across the two locations.

scripts/irods/test/test_iadmin.py
```diff
@@ -2539,23 +2539,32 @@ def test_mkgroup_with_remote_zone(self):
...
-        self.admin.assert_icommand_fail(['iadmin', 'mkzone', 'badzone7', 'remote', 'hostname:asdf1234'],
-                                        'STDERR', 'CAT_INVALID_ARGUMENT')
-        self.admin.assert_icommand_fail(['iadmin', 'mkzone', 'badzone8', 'remote', 'hostname:1234asdf'],
-                                        'STDERR', 'CAT_INVALID_ARGUMENT')
-        self.admin.assert_icommand_fail(['iadmin', 'mkzone', 'badzone9', 'remote', 'hostname:-1234'],
-                                        'STDERR', 'CAT_INVALID_ARGUMENT')
+    @classmethod
+    def setUpClass(cls):
+        """Set up the test class."""
+        cls.admin = session.mkuser_and_return_session('rodsadmin', 'otherrods', 'rods', lib.get_hostname())
```
korydraughn, line 2545: Is there a reason you chose to do this instead of something like the following:
https://github.com/irods/irods/blob/25d8f67df6c123ee6f0b3fe4695e4042b52cdb59/scripts/irods/test/test_ichmod.py#L14

SwooshyCueb: Yes. The mixin uses instance initializers instead of class initializers, which isn't needed for this test case.

SwooshyCueb:
Running the test suite one more time, but I think this PR is ready for review

scripts/irods/test/test_iadmin.py
```diff
@@ -2535,3 +2535,159 @@ def test_mkgroup_with_remote_zone(self):
...
+    @classmethod
+    def tearDownClass(cls):
+        """Tear down the test class."""
+        with session.make_session_for_existing_admin() as admin_session:
+            cls.admin.__exit__()
+            admin_session.assert_icommand(['iadmin', 'rmuser', cls.admin.username])
+
+    def setUp(self):
+        self.admin.assert_icommand(['iadmin', 'mkzone', 'goodzone_localhost', 'remote', 'localhost:1247'], 'EMPTY')
+        self.admin.assert_icommand(['iadmin', 'mkzone', 'goodzone_blank', 'remote', 'localhost:1247'], 'EMPTY')
```
alanking, line 2638: Should the 5th parameter be an empty string instead of 'localhost:1247'?

scripts/irods/test/test_iadmin.py
```diff
@@ -2535,3 +2535,159 @@ def test_mkgroup_with_remote_zone(self):
...
+    @classmethod
+    def tearDownClass(cls):
+        """Tear down the test class."""
+        with session.make_session_for_existing_admin() as admin_session:
+            cls.admin.__exit__()
+            admin_session.assert_icommand(['iadmin', 'rmuser', cls.admin.username])
+
+    def test_good_localhost(self):
+        try:
+            self.admin.assert_icommand(['iadmin', 'mkzone', 'goodzone', 'remote', 'localhost:1247'], 'EMPTY')
```
alanking, line 2556: Should we confirm the existence (or non-existence) of the zones as a result of the success (or failure) of `mkzone` for these tests?

scripts/irods/test/test_iadmin.py
```diff
@@ -2535,3 +2535,159 @@ def test_mkgroup_with_remote_zone(self):
...
+    def setUp(self):
+        self.admin.assert_icommand(['iadmin', 'mkzone', 'goodzone_localhost', 'remote', 'localhost:1247'], 'EMPTY')
+        self.admin.assert_icommand(['iadmin', 'mkzone', 'goodzone_blank', 'remote', 'localhost:1247'], 'EMPTY')
+
+    def tearDown(self):
+        self.admin.run_icommand(['iadmin', 'rmzone', 'goodzone_localhost'])
+        self.admin.run_icommand(['iadmin', 'rmzone', 'goodzone_blank'])
+
+    def test_good_localhost(self):
+        self.admin.assert_icommand(['iadmin', 'modzone', 'goodzone_blank', 'conn', 'localhost:1247'], 'EMPTY')
```
alanking, line 2645: Should we confirm that the zone was modified successfully (or not) based on the success (or failure) of `modzone`?

plugins/database/src/db_plugin.cpp
```diff
@@ -694,6 +695,43 @@ irods::error validate_resource_name( std::string _resc_name ) {
...
+        if (host_len == 0 || (host_len == 1 && addr.hostAddr[0] == ':')) {
+            log_db::info("failed to validate zone connection info due to empty hostname [{}]", _zone_conn_info);
+            addRErrorMsg(&_ctx.comm()->rError,
+                         CAT_HOSTNAME_INVALID,
+                         "failed to validate zone connection info due to empty hostname");
+            return ERROR(CAT_HOSTNAME_INVALID, "failed to validate zone connection info due to empty hostname");
+        }
+        if ((addr.portNum > 65535) || (addr.portNum <= 0)) {
+            log_db::info(
+                "failed to validate zone connection info due to invalid or unspecified port [{}]", _zone_conn_info);
```
alanking, line 724: Is there a reason the `_zone_conn_info` isn't included in the rError message or the `irods::error` message? Same question for the other cases, of course.

alanking (commented):
Overall looks good. Big fan of the tests. :) Spotted a few things

scripts/irods/test/test_iadmin.py
SwooshyCueb: Nice catch

plugins/database/src/db_plugin.cpp
SwooshyCueb: I vaguely remember the convention being to put this sort of thing in logs, but not in `rError` or `irods::error` messages. If that's not actually the case, I can change it.

scripts/irods/test/test_iadmin.py
SwooshyCueb: Good idea

scripts/irods/test/test_iadmin.py
SwooshyCueb: Good idea

plugins/database/src/db_plugin.cpp
trel: Seems a good default position to include the thing that didn't work... unless it's a password or other credential/secret I think...

alanking (commented):
Looks good to me. Consider addressing clang-format issue. Tests are passing?

SwooshyCueb:
Two Python tests are failing. It seems unrelated but I'm still looking into it.

SwooshyCueb:
After re-running the failing tests, I'm relatively certain they are unrelated to my changes. I'll open an issue for them later today.
If there are no objections, I will squash and \#

alanking (approved):
Looks good to me. Let's wait for one more approval and then I think we're good to go

SwooshyCueb:
Squashed

korydraughn (approved):
Commit messages look good.

If you're happy with it, please add the pounds.

alanking (approved):

Issue #6999, opened 2023-03-27
Add library that gives iRODS processes a space to write arbitrary data to


korydraughn:
- [x] main
- [x] main (unit tests)
- [x] 4-3-stable
- [x] 4-3-stable (unit tests)
- [x] 4-2-stable
- [x] 4-2-stable (unit tests)
---
## Feature
Giving iRODS processes a place to write objects to simplifies certain designs. For example GenQuery2 microservices could have been implemented as actual microservices instead of a rule engine plugin.

korydraughn:
Forgot to add unit tests.

PR #7006, opened 2023-03-29
[#6999] Added unit test for irods::process_stash library (4-2-stable)

korydraughn:
Passed in the testing environment.

### DIFF
--- a/unit_tests/src/test_process_stash.cpp
+++ b/unit_tests/src/test_process_stash.cpp
@@ -1,0 +1,68 @@
+#include "catch.hpp"
+
+#include "process_stash.hpp"
+
+#include <boost/any.hpp>
+
+#include <algorithm>
+#include <iterator>
+#include <string>
+
+TEST_CASE("process_stash basic operations")
+{
+    // A POD type which helps to demonstrate the stash's ability to store heterogeneous data.
+    struct pod
+    {
+        int x = 0;
+        int y = 0;
+    };
+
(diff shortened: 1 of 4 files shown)


unit_tests/src/test_process_stash.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include <iterator>
+#include <string>
+
+TEST_CASE("process_stash basic operations")
+{
+    // A POD type used to demonstrate the stash's ability to store heterogeneous data.
+    struct pod
+    {
+        int x = 0;
+        int y = 0;
```
trel, line 17: by heterogeneous... you mean... what?

there are both ints?    can you put different types here at the same time?

alanking: I think it's referring to the stash, not the struct.

trel: ohhh, found a parser :)

okay, so maybe a slight rephrasing of line 13 is in order...

korydraughn: Will do. I felt that line would need a tweak.

korydraughn: Updated.

korydraughn:
Any remaining thoughts before I add the pound?

alanking (approved):

korydraughn:
#'d

Issue #6997, opened 2023-03-24
Compound resource stuck with intermediate replica in cache 


alanking:
- [x] main
- [x] main (S3 test)
- [x] 4-3-stable
- [x] 4-2-stable
- [x] 4-2-stable (S3 test)
---
### iRODS Version, OS and Version
4.2.11

### What did you try to do?
Let `foo` be a data object which has only one replica, and it is in the archive of a compound resource hierarchy, but not in cache. A user with read permission tries to read `foo` (e.g. `iget`).
```bash
$ ils -Al foo
  rods              1 Comp;Archive          283 2023-03-24.14:23 & foo
        ACL - alice#tempZone:read object   rods#tempZone:own   
```

### Expected behavior
The user should get the data:
```bash
$ iget foo -
<some daterz and no errors>
$ ils -Al foo
  rods              1 Comp;Archive          283 2023-03-24.14:43 & foo
        ACL - alice#tempZone:read object   rods#tempZone:own   
  rods              2 Comp;Cache          283 2023-03-24.14:44 & foo
        ACL - alice#tempZone:read object   rods#tempZone:own   
```

### Observed behavior (including steps to reproduce, if applicable)
The user does not get the data, and a replica appears in the cache resource which is stuck in the intermediate status (note: the replica in the archive is still marked good):
```bash
$ iget foo -
remote addresses: 127.0.0.1 ERROR: getUtil: get error for - status = -405000 INTERMEDIATE_REPLICA_ACCESS
$ ils -Al foo
  rods              1 Comp;Archive          283 2023-03-24.14:23 & foo
        ACL - alice#tempZone:read object   rods#tempZone:own
  rods              2 Comp;Cache          283 2023-03-24.14:32 ? foo
        ACL - alice#tempZone:read object   rods#tempZone:own
```
And in the logs...
```
Mar 24 14:55:36 pid:952 NOTICE: rsGenQuery: rcGenQuery failed, status = -818000
Mar 24 14:55:36 pid:952 NOTICE: getDataObjInfo: rsGenQuery error, status = -818000
[...]

PR #7014, opened 2023-03-31
[#6997] Fix stage-to-cache with read-only permission (4-2-stable)

alanking:
In service of #6997 

Core tests and topology tests pass (still need to run unit tests)

### DIFF
--- a/plugins/microservices/msi_tests/libmsi_test_scoped_client_identity.cpp
+++ b/plugins/microservices/msi_tests/libmsi_test_scoped_client_identity.cpp
@@ -13,7 +13,7 @@
 
 namespace
 {
-    auto test_basic_usage() -> int
+    auto test_basic_usage(RuleExecInfo& _rei) -> int
     {
         IRODS_MSI_TEST_BEGIN("scoped_client_identity - basic usage")
 
@@ -45,7 +45,7 @@ namespace
         IRODS_MSI_TEST_END
     } // test_basic_usage
 
-    auto test_updates_zone() -> int
+    auto test_updates_zone(RuleExecInfo& _rei) -> int
     {
         IRODS_MSI_TEST_BEGIN("#6268: scoped_client_identity updates the zone")
 
@@ -93,7 +93,7 @@ namespace
         IRODS_MSI_TEST_END
     } // test_updates_zone
 
-    auto test_invalid_input() -> int
+    auto test_invalid_input(RuleExecInfo& _rei) -> int
     {
         IRODS_MSI_TEST_BEGIN("scoped_client_identity throws on invalid input")
 
@@ -123,9 +123,9 @@ namespace
 
     auto msi_impl(RuleExecInfo* _rei) -> int
     {
-        IRODS_MSI_TEST_CASE(test_basic_usage)
-        IRODS_MSI_TEST_CASE(test_updates_zone)
-        IRODS_MSI_TEST_CASE(test_invalid_input)
+        IRODS_MSI_TEST_CASE(test_basic_usage, *_rei)
+        IRODS_MSI_TEST_CASE(test_updates_zone, *_rei)
+        IRODS_MSI_TEST_CASE(test_invalid_input, *_rei)
 
         return 0;
     } // msi_impl
--- a/plugins/microservices/msi_tests/libmsi_test_scoped_permission.cpp
+++ b/plugins/microservices/msi_tests/libmsi_test_scoped_permission.cpp
@@ -1,0 +1,272 @@
+/// \file
+
+#include "dstream.hpp"
+#include "getRodsEnv.h"
+#include "irods_exception.hpp"
+#include "irods_ms_plugin.hpp"
+#include "irods_re_structs.hpp"
+#include "msParam.h"
+#include "msi_assertion_macros.hpp"
+#include "scoped_permission.hpp"
+
+#define IRODS_IO_TRANSPORT_ENABLE_SERVER_SIDE_API
+#include "transport/default_transport.hpp"
+
+#define IRODS_FILESYSTEM_ENABLE_SERVER_SIDE_API
+#include "filesystem.hpp"
+
+namespace
+{
+    namespace fs = irods::experimental::filesystem;
+    namespace io = irods::experimental::io;
+    using sp = irods::experimental::scoped_permission;
+
+    auto collection_path(const RsComm& comm) -> fs::path
+    {
+        fs::path p = "/";
+        p /= comm.myEnv.rodsZone; // NOLINT(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+        p /= "home";
+        p /= comm.clientUser.userName; // NOLINT(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+
+        return p;
+    } // collection_path
+
+    auto make_test_object(RsComm& comm) -> fs::path
+    {
+        const auto path = collection_path(comm) / "test_scoped_permission.txt";
+
+        const std::string_view contents = "testdata";
+
+        // Create a new data object using the server-side API.
+        {
+            io::server::native_transport tp{comm};
+            io::odstream{tp, path} << contents;
+        }
+
+        return path;
+    } // make_test_object
+
+    auto cleanup(RsComm& comm, const fs::path& path) -> void
+    {
+        if (fs::server::exists(comm, path)) {
+            // Ironically, we must use scoped permissions to ensure the ability to remove the object.
+            const auto temporary_ownership = sp{comm, path, fs::perms::own};
+            fs::server::remove(comm, path, fs::remove_options::no_trash);
+        }
+
+        // The filesystem API pushes a message onto the rError stack here because scoped_permission elevates
+        // permissions on a data object, the data object is removed, and scoped_permission cannot restore
+        // permissions on the data object because it no longer exists.
+        irods::pop_error_message(comm.rError);
+    } // cleanup
+
+    auto test_invalid_path(RuleExecInfo& _rei) -> int
+    {
+        IRODS_MSI_TEST_BEGIN("test scoped_permission with invalid path")
+
+        auto* comm = _rei.rsComm;
+
+        const auto nonexistent_path = collection_path(*comm) / "not_real.bin";
+        IRODS_MSI_ASSERT(!fs::server::exists(*comm, nonexistent_path))
+        IRODS_MSI_THROWS((sp{*comm, nonexistent_path, fs::perms::null}), fs::filesystem_error)
+
+        // Something deep inside the database plugin adds a message to the rError stack when attempting to modify
+        // permissions on a nonexistent data object. scoped_permission is not responsible for managing the error
+        // stack, so, pop it off here so the test calling this code doesn't need to worry about it.
+        irods::pop_error_message(comm->rError);
+
+        IRODS_MSI_TEST_END
+    } // test_invalid_path
+
+    auto test_empty_zone_name_in_rs_comm(RuleExecInfo& _rei) -> int
+    {
+        IRODS_MSI_TEST_BEGIN("test scoped_permission with empty zone name")
+
+        auto* comm = _rei.rsComm;
+
+        const auto path = make_test_object(*comm);
+
+        IRODS_MSI_ASSERT(fs::server::is_data_object(*comm, path))
+        IRODS_MSI_ASSERT(fs::perms::own == fs::server::status(*comm, path).permissions().at(0).prms)
+
+        auto remove_data_object = irods::at_scope_exit{[comm, &path] { cleanup(*comm, path); }};
+
+        // Clear the zone name and ensure that scoped_permission succeeds, defaulting to the local zone.
+        {
+            auto other_comm = *comm;
+            std::memset(other_comm.clientUser.rodsZone, 0, sizeof(other_comm.clientUser.rodsZone));
+
+            const auto permission_with_empty_zone_name = sp{other_comm, path, fs::perms::read};
+
+            IRODS_MSI_ASSERT(fs::perms::read == fs::server::status(*comm, path).permissions().at(0).prms)
+        }
+
+        // Own permission is restored.
+        IRODS_MSI_ASSERT(fs::perms::own == fs::server::status(*comm, path).permissions().at(0).prms)
+
+        IRODS_MSI_TEST_END
+    } // test_empty_zone_name_in_rs_comm
+
+    auto test_invalid_username_and_zone(RuleExecInfo& _rei) -> int
+    {
+        IRODS_MSI_TEST_BEGIN("test scoped_permission with invalid username and zone name")
+
+        auto* comm = _rei.rsComm;
+
+        const auto path = make_test_object(*comm);
+
+        IRODS_MSI_ASSERT(fs::server::is_data_object(*comm, path))
+        IRODS_MSI_ASSERT(fs::perms::own == fs::server::status(*comm, path).permissions().at(0).prms)
+
+        auto remove_data_object = irods::at_scope_exit{[comm, &path] { cleanup(*comm, path); }};
+
+        // Keep a copy of the client user information in the RsComm so we can restore it.
+        auto other_comm = *comm;
+
+        // Clear the username and ensure that scoped_permission fails.
+        std::memset(other_comm.clientUser.userName, 0, sizeof(other_comm.clientUser.userName));
+        IRODS_MSI_THROWS((sp{other_comm, path, fs::perms::read}), fs::filesystem_error)
+        IRODS_MSI_ASSERT(fs::perms::own == fs::server::status(*comm, path).permissions().at(0).prms)
+
+        // Make the username something that is not a user so that the qualified username is no longer valid.
+        std::strncpy(other_comm.clientUser.userName, comm->clientUser.rodsZone, sizeof(comm->clientUser.userName));
+        IRODS_MSI_THROWS((sp{other_comm, path, fs::perms::read}), fs::filesystem_error)
+        IRODS_MSI_ASSERT(fs::perms::own == fs::server::status(*comm, path).permissions().at(0).prms)
+
+        // Restore the original username.
+        std::strncpy(other_comm.clientUser.userName, comm->clientUser.userName, sizeof(comm->clientUser.userName));
+
+        // Make the zone name something that is not a zone so that the qualified username is no longer valid.
+        std::strncpy(other_comm.clientUser.rodsZone, comm->clientUser.userName, sizeof(comm->clientUser.rodsZone));
+        IRODS_MSI_THROWS((sp{other_comm, path, fs::perms::read}), fs::filesystem_error)
+        IRODS_MSI_ASSERT(fs::perms::own == fs::server::status(*comm, path).permissions().at(0).prms)
+
+        IRODS_MSI_TEST_END
+    } // test_invalid_username_and_zone
+
+    auto test_basic_usage(RuleExecInfo& _rei) -> int
+    {
+        IRODS_MSI_TEST_BEGIN("test scoped_permission basic usage")
+
+        auto* comm = _rei.rsComm;
+
+        const auto path = make_test_object(*comm);
+
+        IRODS_MSI_ASSERT(fs::server::is_data_object(*comm, path))
+        IRODS_MSI_ASSERT(fs::perms::own == fs::server::status(*comm, path).permissions().at(0).prms)
+
+        auto remove_data_object = irods::at_scope_exit{[comm, &path] { cleanup(*comm, path); }};
+
+        // Change permissions to write.
+        {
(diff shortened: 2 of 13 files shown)


server/api/src/rsModColl.cpp
```diff
@@ -106,8 +111,11 @@ auto rsModColl(rsComm_t* rsComm, collInp_t* modCollInp) -> int
         // server-to-server connection. Many operations in iRODS use this API to update the collection mtime after a
         // data object modification or creation and requires privileged status to modify the collection in the event
         // that the user does not have permissions on the parent collection. A client connection is made here using the
-        // service account rodsadmin credentials to modify the collection and then is immediately disconnected.
-        if (irods::is_privileged_client(*rsComm)) {
+        // service account rodsadmin credentials to modify the collection and then is immediately disconnected. This is
+        // only done for connections with elevated privileges that are not rodsadmins.
+        if (irods::is_privileged_client(*rsComm) &&
+            ua::user_type::rodsadmin != *ua::server::type(*rsComm, ua::user{rsComm->clientUser.userName}))
```
korydraughn, line 117: Does this cover remote users?

scripts/irods/test/resource_suite.py
```diff
@@ -1229,4 +1229,61 @@ def test_itrim_returns_on_negative_status__ticket_3531(self):
...
+            self.assertEqual(str(1), lib.get_replica_status(readonly_user, os.path.basename(logical_path), 1))
+            self.assertEqual(str(1), lib.get_replica_status(readonly_user, os.path.basename(logical_path), 2))
+
+            # Ensure that the user has the same permissions on the data object as before getting it.
+            assert_permissions_on_data_object_for_user(readonly_user.username, logical_path, 1050)
+
+        finally:
+            self.admin.assert_icommand(['ils', '-Al', logical_path], 'STDOUT') # Debugging
+
+            # Make sure that the data object can be removed by marking both replicas stale before removing.
```
korydraughn, line 1283: Is marking the replicas stale meant to guard against intermediate replicas and make the test idempotent?

server/api/src/rsModAccessControl.cpp
```diff
@@ -1,68 +1,95 @@
...
-                svc_role.c_str() );
-            status = SYS_SERVICE_ROLE_NOT_SUPPORTED;
+        // Some privileged clients may be affected by temporary elevated privileges. The redirect to the catalog service
+        // provider here will cause the temporary privileged status granted to this connection to be lost in the
+        // server-to-server connection. The compound resource temporarily elevates permissions for users with read-only
+        // access to data objects which need to be staged to cache. A client connection is made here using the service
+        // account rodsadmin credentials to modify the permissions and then is immediately disconnected. This is only
+        // done for connections with elevated privileges that are not rodsadmins.
+        if (irods::is_privileged_client(*rsComm) &&
+            ua::user_type::rodsadmin != *ua::server::type(*rsComm, ua::user{rsComm->clientUser.userName}))
```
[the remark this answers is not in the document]
alanking: Not sure... I guess we can make sure by passing `rsComm->clientUser.rodsZone`, right?

I will also add a test - similar to the one I've already written - to the federation test suite to make sure things work as expected.

server/api/src/rsModColl.cpp
alanking: See comment in `rsModAccessControl`.

I'll add a test similar to `test_remove_data_object_in_collection_with_read_permissions__issue_6428` to the federation test suite.

server/api/src/rsModAccessControl.cpp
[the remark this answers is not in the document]
korydraughn: Sure.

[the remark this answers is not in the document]
alanking: So, it turns out that the existing solution does not fix the issue for remote users (observed a cache replica stuck in the intermediate status).

This is because `scoped_permissions` also does not consider zone name when setting the permissions on the data object. I have adjusted the `scoped_permissions` class to hold a qualified username which is passed to the `irods::experimental::filesystem::permissions` call, which parses and interprets qualified usernames.

We should also make sure the user is correct by passing the zone name here, here but that did not resolve the issue (the resolution is noted in the above paragraph).

alanking:
Had to force-push due to rebase. New changes exist in new commits. Still need to run tests.

server/api/src/rsModAccessControl.cpp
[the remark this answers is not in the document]
alanking: Done

server/api/src/rsModColl.cpp
alanking: Done

scripts/irods/test/test_federation.py
```diff
@@ -1777,3 +1777,134 @@ def test_local_zone_user_is_not_allowed_to_delete_delay_rules_created_by_remote_
 
         finally:
             self.local_admin.run_icommand(['iqdel', '-a'])
+
+
+class test_compound_resource_operations(SessionsMixin, unittest.TestCase):
```
alanking, line 1827: I'm not too attached to this name or even the idea of this being a separate class. I did it this way because of all the setup required for this to work.

alanking:
That is to say, I need to run the full test suite. All the new tests appropriately failed and then passed after their associated fixes.

scripts/irods/test/test_federation.py
```diff
@@ -1222,6 +1222,51 @@ def test_catalog_provider_hosts_other_than_the_first_are_considered__issue_6827(
...
+
+                new_mtime = get_collection_mtime(owner, collection_path)
+
+                self.assertNotEqual(original_mtime, new_mtime, msg='collection mtime was not updated')
+
+            finally:
+                owner.assert_icommand(['ils', '-Al', collection_path], 'STDOUT') # Debugging
+
+                user.assert_icommand(['irmtrash'])
+                owner.assert_icommand(['irmtrash'])
```
alanking, line 1265: Remove this...

alanking: Done

server/core/include/scoped_permission.hpp
```diff
@@ -78,6 +81,8 @@ namespace irods::experimental
         RsComm& comm_;
         const filesystem::path path_;
         filesystem::perms old_perm_;
+        /// \var Automatically constructed username composed of the username and zone name, joined by a pound sign.
```
korydraughn, line 84: I don't think we need to document this. It is pretty self explanatory IMO.

scripts/irods/test/test_federation.py
```diff
@@ -1777,3 +1819,133 @@ def test_local_zone_user_is_not_allowed_to_delete_delay_rules_created_by_remote_
...
+        super(test_compound_resource_operations, self).setUp()
+
+        # load federation settings in dictionary (all lower case)
+        self.config = {}
+        for key, val in test.settings.FEDERATION.__dict__.items():
+            if not key.startswith('__'):
+                self.config[key.lower()] = val
+        self.config['local_zone'] = self.user_sessions[0].zone_name
+        if test.settings.FEDERATION.REMOTE_IRODS_VERSION < (4, 0, 0):
+            test.settings.FEDERATION.REMOTE_VAULT = '/home/irods/irods-legacy/iRODS/Vault'
```
korydraughn, line 1835: We don't have a way of testing this AFAIK. Do you feel this is necessary?

scripts/irods/test/test_resource_types.py
```diff
@@ -402,6 +402,11 @@ def test_local_iput_with_force_and_destination_resource__ticket_1706(self):
         os.remove(doublefile)
 
 
+    @unittest.skip('This test only applies to compound hierarchies.')
+    def test_iget_data_object_as_user_with_read_only_access_and_replica_only_in_archive__issue_6697(self):
+        pass
```
korydraughn, line 407: Are you planning on implementing these or do they exist for some other reason?

Just curious.

server/core/include/scoped_permission.hpp
```diff
@@ -33,6 +35,7 @@ namespace irods::experimental
             : comm_{_comm}
             , path_{_path}
             , old_perm_{filesystem::perms::null}
+            , qualified_username_{fmt::format("{}#{}", comm_.clientUser.userName, comm_.clientUser.rodsZone)}
```
korydraughn, line 38: I was trying to think if there was any potential issues with this, but I don't think there are.

I think we have to assume that within the server, the RsComm ALWAYS contains a proper username and zone for the client.

You agree?

scripts/irods/test/test_federation.py
korydraughn: It's fine with the name as long as it gives the reader a decent idea about what to expect.

If the current name is not clear, please use more words.

scripts/irods/test/test_federation.py
alanking: This was copied from `test_federation.Test_ICommands`. I will just remove it.

scripts/irods/test/test_resource_types.py
alanking: I put the test in `resource_suite` and that means that every test class in `test_resource_types` will run the test unless it is overridden and skipped in the appropriate test classes. So, we could either copy the test for all the test classes that we want to run this test, or we can do it this way.

I'm fine with either way.

korydraughn: Oh, this is foundational code.

Derived classes can choose to override the implementation if they need to.

scripts/irods/test/test_federation.py
alanking: Okay, then I will just leave it. Thanks

server/api/src/rsModAccessControl.cpp
```diff
@@ -1,68 +1,95 @@
...
-            status = SYS_SERVICE_ROLE_NOT_SUPPORTED;
+        // Some privileged clients may be affected by temporary elevated privileges. The redirect to the catalog service
+        // provider here will cause the temporary privileged status granted to this connection to be lost in the
+        // server-to-server connection. The compound resource temporarily elevates permissions for users with read-only
+        // access to data objects which need to be staged to cache. A client connection is made here using the service
+        // account rodsadmin credentials to modify the permissions and then is immediately disconnected. This is only
+        // done for connections with elevated privileges that are not rodsadmins.
+        const auto client_user_type =
+            *ua::server::type(*rsComm, ua::user{rsComm->clientUser.userName, rsComm->clientUser.rodsZone});
+        if (irods::is_privileged_client(*rsComm) && ua::user_type::rodsadmin != client_user_type) {
```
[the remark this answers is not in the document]
alanking: Done

server/api/src/rsModColl.cpp
```diff
@@ -106,8 +111,11 @@ auto rsModColl(rsComm_t* rsComm, collInp_t* modCollInp) -> int
         // server-to-server connection. Many operations in iRODS use this API to update the collection mtime after a
         // data object modification or creation and requires privileged status to modify the collection in the event
         // that the user does not have permissions on the parent collection. A client connection is made here using the
-        // service account rodsadmin credentials to modify the collection and then is immediately disconnected.
-        if (irods::is_privileged_client(*rsComm)) {
+        // service account rodsadmin credentials to modify the collection and then is immediately disconnected. This is
+        // only done for connections with elevated privileges that are not rodsadmins.
+        const auto client_user_type =
+            *ua::server::type(*rsComm, ua::user{rsComm->clientUser.userName, rsComm->clientUser.rodsZone});
+        if (irods::is_privileged_client(*rsComm) && ua::user_type::rodsadmin != client_user_type) {
```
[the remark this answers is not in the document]
alanking: Done

server/core/include/scoped_permission.hpp
alanking: Done

server/core/include/scoped_permission.hpp
korydraughn: No, I think this is fine.

I think we have the tools to test this in main and 4-3-stable now. We can implement MSIs that are called from the test_federation.py script.

See https://github.com/irods/irods/blob/95dac9a16094d33f4a534572e223440370045b76/plugins/microservices/src/test_scoped_client_identity.cpp.

alanking: Done.

Empty zone names are converted to the local zone for a qualified username every time. Unit test added to confirm behavior.

scripts/irods/test/test_resource_types.py
alanking: Correct. Resolving this since it is common practice in the resource test suite.

alanking:
Squashed

Unit, core, federation, and topology tests pass

plugins/microservices/msi_tests/libmsi_test_scoped_permission.cpp
```diff
@@ -0,0 +1,274 @@
...
+#define IRODS_FILESYSTEM_ENABLE_SERVER_SIDE_API
+#include "filesystem.hpp"
+
+namespace
+{
+    namespace fs = irods::experimental::filesystem;
+    namespace io = irods::experimental::io;
+    using sp = irods::experimental::scoped_permission;
+
+    auto collection_path(RsComm* comm) -> fs::path
```
korydraughn, line 24: Pass by reference?

scripts/irods/test/test_misc.py
```diff
@@ -216,6 +216,10 @@ def test_client_side_and_server_side_dstream_library_interfaces_can_be_used_in_s
     def test_scoped_client_identity_updates_zone_of_RsComm__issue_6268(self):
         self.admin.assert_icommand(['irule', '-r', 'irods_rule_engine_plugin-irods_rule_language-instance', 'msi_test_scoped_client_identity', 'null', 'ruleExecOut'])
 
+    @unittest.skipUnless(plugin_name == 'irods_rule_engine_plugin-irods_rule_language', "Not implemented for other REPs.")
+    def test_scoped_permission(self):
```
korydraughn, line 220: Issue number?

plugins/microservices/msi_tests/libmsi_test_scoped_permission.cpp
```diff
@@ -0,0 +1,274 @@
...
+        return path;
+    } // make_test_object
+
+    auto cleanup(RsComm* comm, const fs::path& path) -> void
+    {
+        if (fs::server::exists(*comm, path)) {
+            // Ironically, we must use scoped permissions to ensure the ability to remove the object.
+            const auto temporary_ownership = sp{*comm, path, fs::perms::own};
+            fs::server::remove(*comm, path, fs::remove_options::no_trash);
+        }
```
korydraughn, line 55: So this is the case where the perms cannot be restored due to the target object being removed.

plugins/microservices/msi_tests/libmsi_test_scoped_permission.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+        // The filesystem API pushes a message onto the rError stack here because scoped_permission elevates
+        // permissions on a data object, the data object is removed, and scoped_permission cannot restore
+        // permissions on the data object because it no longer exists.
+        irods::pop_error_message(comm->rError);
```
korydraughn, line 60: Should scoped_permission add a message to the rError object?

Perhaps it should only log an error message? Hmm.

plugins/microservices/msi_tests/libmsi_test_scoped_permission.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+        // Own permission is restored.
+        IRODS_MSI_ASSERT(fs::perms::own == fs::server::status(*comm, path).permissions().at(0).prms)
+
+        IRODS_MSI_TEST_END
+    } // test_basic_usage
+
+    auto msi_impl(RuleExecInfo* _rei) -> int
+    {
+        IRODS_MSI_TEST_CASE([](RuleExecInfo* _rei) { return !(_rei && _rei->rsComm); }, _rei)
```
korydraughn, line 251: What does this produce in the log file? Is it easy to recognize?

plugins/microservices/msi_tests/libmsi_test_scoped_client_identity.cpp
```diff
@@ -13,7 +13,7 @@
 
 namespace
 {
-    auto test_basic_usage() -> int
+    auto test_basic_usage(RuleExecInfo* _rei) -> int
```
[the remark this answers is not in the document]
alanking: Sure

plugins/microservices/msi_tests/libmsi_test_scoped_permission.cpp
alanking: Correct. Everything works as expected, you just get some errors on the way out the door. Something to consider.

plugins/microservices/msi_tests/libmsi_test_scoped_permission.cpp
alanking: scoped_permission does not add the message. Some API that the filesystem library calls adds the message.

scripts/irods/test/test_misc.py
korydraughn: Perhaps a new issue?

plugins/microservices/msi_tests/libmsi_test_scoped_permission.cpp
alanking: Oh, I guess I need to wrap this with something in order to see anything in the logs. Currently, nothing is being printed out because it's not using the other macros. Should I just get rid of it? It's a sanity check in a test, so, this assertion being false would end up crashing the agent later due to nullptr dereferences. But that would indicate an issue in the microservice plugin interface or the REP which called it.

plugins/microservices/msi_tests/libmsi_test_scoped_permission.cpp
korydraughn: Got it. That answers my question.

plugins/microservices/msi_tests/libmsi_test_scoped_client_identity.cpp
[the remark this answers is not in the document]
korydraughn: This goes for all locations where we take pointer.

I can't think of reason any of the tests would want to pass a nullptr. Is there motivation to support that case?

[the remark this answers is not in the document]
alanking: No reason other than it took more effort to make sure all the pointers were derefenced appropriately :p

plugins/microservices/msi_tests/libmsi_test_scoped_permission.cpp
korydraughn: I think it's fine to remove. If a nullptr is passed, then we have bigger problems. Also, I think it's safe to assume it is always valid.

Thoughts?

plugins/microservices/msi_tests/libmsi_test_scoped_client_identity.cpp
```diff
@@ -123,9 +123,11 @@ namespace
 
     auto msi_impl(RuleExecInfo* _rei) -> int
     {
-        IRODS_MSI_TEST_CASE(test_basic_usage)
-        IRODS_MSI_TEST_CASE(test_updates_zone)
-        IRODS_MSI_TEST_CASE(test_invalid_input)
+        IRODS_MSI_TEST_CASE([](RuleExecInfo* _rei) { return !_rei; }, _rei)
```
korydraughn, line 126: This one can be removed as well.

plugins/microservices/msi_tests/libmsi_test_scoped_permission.cpp
alanking: Yep, makes sense to me.

korydraughn (commented):
Aside from the comments, everything else looks good.

Will wait for remaining comments to be resolved before marking approved.

plugins/microservices/msi_tests/libmsi_test_scoped_permission.cpp
alanking: Done

trel: maybe make a note in the docs to make this assumption explicit

korydraughn: Let's capture that in an issue in irods_docs. That info is likely for the developer docs.

korydraughn:
Please squash and we'll look over it one last time.

plugins/microservices/msi_tests/libmsi_test_scoped_permission.cpp
alanking: Done. https://github.com/irods/irods_docs/issues/205

alanking:
Squashed.

korydraughn (approved):
Please add the pounds and merge.

Issue #6385, opened 2022-05-11
The pep_database_mod_data_obj_meta_* PEPs not called when new file uploaded 


tedgin:
- [x] main
- [x] 4-3-stable
- [x] 4-2-stable
---
### iRODS Version, OS and Version

iRODS 4.2.11, CentOS 7

### What did you try to do?

I use the PEP `pep_database_mod_data_obj_meta_post` enforce various policies. One of those policies is ensuring that every replica has a checksum,  In iRODS 4.2.8, when data  object was created for an uploaded file, this PEP was called after the replica was created on a storage resource. The `database_mod_data_obj_meta` operation was called to set the size of the new replica in the catalog. 

I tried to use iput to create a new data object having a checksum using this same rule logic in iRODS 4.2.11, but with a `writeLine('serverLog', 'database_mod_data_obj_meta');` added to the top of the rule attached to `pep_database_mod_data_obj_meta_post`.  

### Expected behavior

I expect a newly uploaded data object having a checksum. I also expect to see a message in the server log mentioning `database_mod_data_obj_meta`.  

### Observed behavior (including steps to reproduce, if applicable)

I use iput to upload a file, but the new data object doesn't have a checksum and there is no mention of `database_mod_data_obj_meta` in the server log.

```console
[centos@bms-irods ~]$ iput motd
[centos@bms-irods ~]$ ils -L motd
  centos            0 demoResc            5 2022-05-11.17:40 & motd
        generic    /var/lib/irods/Vault/home/centos/motd
```

In 4.2.11, the  `database_mod_data_obj_meta` operation isn't called when a file is uploaded, and a newly uploaded data object doesn't receive a checksum.

It looks like the database plugin is being bypassed. With SQL logging enabled, I can see an `INSERT` statement that creates an entry for the data object and it's zeroth replica with size set to zero, but I don't see an `UPDATE` statement that set's the replica's correct size after it has been written to the storage resource.  However, the catalog has the data object's correct size.

tedgin:
Which `pep_api_*` PEPs are triggered when a replica is created or updated? From what I can tell the following operations will likely create or update a replica. The ones preceded by an `!` will likely create or update a replica, while the ones preceded by a `?` will possibly create or update replica.  I'm going solely off of the names, so I may be wildly incorrect.

```
!  bulk_data_obj_put(*INSTANCE, *COMM, *BULKOPRINP, *BULKOPRINPBBUF)
[...]

PR #7033, opened 2023-04-10
[#6385] Add DB plugin op for data_object_finalize (main)

alanking:
Addresses #6385

Cherry-picked from #7012 

Supersedes #5665

Major difference from 4-2-stable is a hefty application of `NOLINTNEXTLINE` to placate clang-tidy and clang-format had some opinions which were addressed. There is one issue clang-tidy is still upset about that I am not going to address since it has to do with a static constant string in a long list of others like it which have the same issue.

Core tests are running

### DIFF
--- a/plugins/database/src/db_plugin.cpp
+++ b/plugins/database/src/db_plugin.cpp
@@ -39,25 +39,27 @@
 #include "irods/catalog_utilities.hpp"
 #include "irods/plugins/auth/pam_password.hpp"
 
-#include <boost/date_time.hpp>
-#include <boost/regex.hpp>
-#include <boost/lexical_cast.hpp>
-#include <fmt/format.h>
 #include <fmt/chrono.h>
+#include <fmt/format.h>
 #include <nanodbc/nanodbc.h>
+#include <nlohmann/json.hpp>
 
+#include <boost/date_time.hpp>
+#include <boost/lexical_cast.hpp>
+#include <boost/regex.hpp>
+
+#include <algorithm>
+#include <charconv>
+#include <chrono>
 #include <cstring>
+#include <iomanip>
+#include <iostream>
+#include <locale>
 #include <sstream>
 #include <string>
 #include <string_view>
-#include <iostream>
 #include <type_traits>
 #include <vector>
-#include <algorithm>
-#include <chrono>
-#include <iomanip>
-#include <locale>
-#include <charconv>
 
 // clang-format off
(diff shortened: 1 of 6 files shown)


korydraughn:
No need to backport formatting, etc to 4-2-stable.

alanking:
Unit tests and core tests passed

plugins/database/src/db_plugin.cpp
```diff
@@ -14920,6 +14922,165 @@ auto db_get_delay_rule_info_op(irods::plugin_context& _ctx, const char* _rule_id
...
+            // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-constant-array-index)
+            cllBindVars[cllBindVarCount++] = get_json_string_value(after, "data_mode");
+            // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-constant-array-index)
+            cllBindVars[cllBindVarCount++] = get_json_string_value(after, "r_comment");
+            // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-constant-array-index)
+            cllBindVars[cllBindVarCount++] = get_json_string_value(after, "create_ts");
+            // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-constant-array-index)
+            cllBindVars[cllBindVarCount++] = get_json_string_value(after, "modify_ts");
+            // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-constant-array-index)
+            cllBindVars[cllBindVarCount++] = get_json_string_value(after, "resc_id");
```
korydraughn, line 15030: Oh god. Consider wrapping these lines in `NOLINTBEGIN(...)` and `NOLINTEND(...)` so that it's not as loud.

ClangTidy won't understand it at this time, but it will once we decide to bump the compiler (in the future).

korydraughn (approved):
Feel free to ignore my comment.

If you're happy with it, please add the pounds and merge.

plugins/database/src/db_plugin.cpp
alanking: Okay. I wanted to use those but wasn't sure if using the future-things was acceptable. That being said, in order to avoid my inevitable screwing something up by changing it, I think I'm going to leave it as-is.

korydraughn: That's fine.

alanking: I promise it'll go away one day ;)

PR #7035, opened 2023-04-10
Issue 6367 main

JustinKyleJames:
Fix for 6467 on main branch.  I made one change to the test which is in its own commit.

All tests (CI and unit) passed.

### DIFF
--- a/lib/core/src/fsckUtil.cpp
+++ b/lib/core/src/fsckUtil.cpp
@@ -181,6 +181,9 @@ int chkObjConsistency(rcComm_t* conn,
             status = SYS_INTERNAL_ERR;
         }
     }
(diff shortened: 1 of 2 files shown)


korydraughn (approved):
Thanks for isolating the change for 4.3 to a single commit.

Noting here that the test was changed to use `-r` instead of `-C`. The option changed in 4.3.

Please squash to one commit and add the pound.

JustinKyleJames:
> Thanks for isolating the change for 4.3 to a single commit.
> 
> Noting here that the test was changed to use `-r` instead of `-C`. The option changed in 4.3.
> 
> Please squash to one commit and add the pound.

Done

alanking (approved):

Issue #7032, opened 2023-04-10
Add unit test for scoped_permission


alanking:
- [x] main
- [x] 4-3-stable
- [x] 4-2-stable
---
Add a unit test for scoped_permission which uses the MSI unit test framework.

Issue #6428, opened 2022-06-01
Inconsistent errors for `irm` of object with own permission in collection with read permission


alanking:
- [x] main
- [x] 4-3-stable
- [x] 4-2-stable
---
### iRODS Version, OS and Version
4.2.11/pre-4.3.0

### What did you try to do?
`irm` a data object as a `rodsuser` with own permission on the object but "read object" permission on the parent collection:
```bash
$ ils ../public/testdir -AL
/tempZone/home/public/testdir:
        ACL - alice#tempZone:read object   rods#tempZone:own
        Inheritance - Disabled
  rods              0 demoResc          283 2022-06-01.15:00 & foo
        generic    /var/lib/irods/Vault/home/public/testdir/foo
        ACL - alice#tempZone:own   rods#tempZone:own
$ irm ../public/testdir/foo
```

### Expected behavior
Consistent behavior when connected to any server in a given zone

### Observed behavior (including steps to reproduce, if applicable)
On the catalog service consumer:
```bash
$ irm ../public/testdir/foo
remote addresses: 127.0.0.1 ERROR: rmUtil: rm error for /tempZone/home/public/testdir/foo, status = -818000 status = -818000 CAT_NO_ACCESS_PERMISSION
$ ils ../public/testdir -AL
/tempZone/home/public/testdir:
        ACL - alice#tempZone:read object   rods#tempZone:own
        Inheritance - Disabled
$ ils ../../trash/home/alice/public/testdir -AL
/tempZone/trash/home/alice/public/testdir:
        ACL - alice#tempZone:own
        Inheritance - Disabled
  rods              0 demoResc          283 2022-06-01.15:11 & foo
        generic    /var/lib/irods/Vault/trash/home/alice/public/testdir/foo
        ACL - alice#tempZone:own   rods#tempZone:own
```
And in the log file:
```
Jun  1 15:11:33 pid:690 remote addresses: 127.0.0.1, 192.168.96.3 ERROR: cannot set mtime: Unknown error -818000
Jun  1 15:11:33 pid:690 remote addresses: 127.0.0.1, 192.168.96.3 ERROR: rsMvDataObjToTrash: rcDataObjRename error for /tempZone/trash/home/alice/public/testdir/foo, status = -818000
```

When connected to the catalog service provider...
```bash
$ ils -AL ../public/testdir
/tempZone/home/public/testdir:
        ACL - alice#tempZone:read object   rods#tempZone:own   
        Inheritance - Disabled
[...]

PR #7038, opened 2023-04-11
[#6997] Fix stage-to-cache with read-only permission (main)

alanking:
Addresses #6428 
Addresses #6997 
Addresses #7032

Cherry-picked from #7014 

Unit, core, and federation tests passed. I only ran `test_iget_data_object_as_user_with_read_only_access_and_replica_only_in_archive__issue_6697` in topology and it passes.

---

Cherry-picking this was a little tricky, so, keeping the diffs out in separate commits was difficult. As such, I went ahead and squashed in other changes I had to make because there were already diffs in the existing commits.

Almost all of the relevant diffs are in the MSI unit tests because the files all moved and the CMake project structure is different. So, most of those differences are actually in the CMake file (which should be easy to review). There are a couple of places I modified which had different formatting from 4-2-stable so those diffs will look slightly different. The `MSI_TEST_CASE` macro is one such instance.

The other relevant diff: I had to explicitly add `lib/administration/user/include` and `lib/administration/zone/include` in the CMake `target_include_directories` for `server/api` in order for `rsModAccessControl.cpp` and `rsModColl.cpp` to include `irods/user_administration.hpp`. Open to suggestions on other ways to address the issue, but I would request that they be punted to another issue for the sake of 4.2.12.

Finally, this is not a diff, but clang-format is going to be unhappy about `scoped_permission.hpp` because of the way we indent our access specifiers. clang-tidy has a number of suggestions which would introduce extensive changes in this branch and are pretty minor, IMO. I am ignoring these unless I'm instructed to follow the suggested changes.


korydraughn (approved):
Looks good. Please add the pounds and merge.

PR #7047, opened 2023-04-17
Client connection libraries allow deferring/changing the auth logic (main)

korydraughn:
This is in support of the C++ REST/HTTP API rewrite.

This is still a WIP, but it is mostly complete. The primary goal is to support proxy connections and allow client-side C++ implementations to choose how they authenticate. For example, the C++ REST/HTTP API rewrite needs to support login using usernames and passwords.

The unit tests pass. I still need to document things and figure out why `CAT_INVALID_AUTHENTICATION` is thrown from the lambdas instead of the the exception I defined.

Feel free to review or ask questions.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -28,6 +28,7 @@ Checks:
     hicpp-signed-bitwise,
     misc-*,
     modernize-*,
+    -modernize-use-trailing-return-type,
     performance-*,
     portability-std-allocator-const,
     portability-simd-intrinsics,
@@ -36,6 +37,7 @@ Checks:
 WarningsAsErrors:
     boost-*,
     bugprone-*,
+    -bugprone-easily-swappable-parameters,
     clang-diagnostic-*,
     clang-analyzer-*,
     cert-dcl*,
@@ -43,10 +45,12 @@ WarningsAsErrors:
     cert-env33-c,
     cert-err34-c,
     cert-err*-cpp,
+    -cert-err58-cpp,
     cert-flp30-c,
     cert-oop*,
     cppcoreguidelines-*,
-    -cppcoreguidelines-pro-bounds-array-to-pointer-decay
+    -cppcoreguidelines-avoid-magic-numbers,
+    -cppcoreguidelines-pro-bounds-array-to-pointer-decay,
     concurrency-thread-canceltype-asynchronous,
     google-build-namespaces,
     google-build-using-namespace,
@@ -60,11 +64,14 @@ WarningsAsErrors:
     hicpp-signed-bitwise,
     misc-*,
     modernize-*,
+    -modernize-use-trailing-return-type,
     performance-*,
     portability-std-allocator-const,
     readability-*,
     -readability-avoid-const-params-in-decls,
-    -readability-implicit-bool-conversion
+    -readability-function-cognitive-complexity,
+    -readability-implicit-bool-conversion,
+    -readability-named-parameter
 FormatStyle:     none
 CheckOptions:
   - key:             llvm-else-after-return.WarnOnConditionVariables
--- a/lib/core/include/irods/client_connection.hpp
+++ b/lib/core/include/irods/client_connection.hpp
@@ -3,6 +3,8 @@
 
 /// \file
 
+#include "irods/fully_qualified_username.hpp"
+
 #include <string_view>
 #include <memory>
 
@@ -14,13 +16,20 @@ struct RcComm;
 namespace irods::experimental
 {
     // clang-format off
-    //
+
     /// A tag type that indicates whether or not a client connection
     /// should allow the user to control when the connection and
     /// authentication occurs.
     ///
     /// \since 4.2.9
     inline const struct defer_connection {} defer_connection;
+
+    /// A tag type that indicates whether or not a client connection
+    /// should allow the user to control when the authentication occurs.
+    ///
+    /// \since 4.3.1
+    inline const struct defer_authentication {} defer_authentication;
+
     // clang-format on
 
     /// This move-only class provides a convenient way to connect to and
@@ -29,7 +38,7 @@ namespace irods::experimental
     /// \since 4.2.9
     class client_connection // NOLINT(cppcoreguidelines-special-member-functions)
     {
-    public:
+      public:
         /// Connects and authenticates using the information and credentials
         /// found in the user's irods_environment.json file.
         ///
@@ -37,6 +46,16 @@ namespace irods::experimental
         ///
         /// \since 4.2.9
         client_connection();
+
+        /// Connects using the information and credentials found in the user's
+        /// irods_environment.json file. No authentication is performed.
+        ///
+        /// \throws irods::exception If an error occurred.
+        ///
+        /// \param[in] _defer_tag Indicates that authentication must be skipped.
+        ///
+        /// \since 4.3.1
+        explicit client_connection(struct defer_authentication _defer_tag);
 
         /// Connects to the iRODS server identified by the passed arguments
         /// and authenticates the user using the information and credentials
@@ -47,34 +66,60 @@ namespace irods::experimental
         /// \param[in] _host     The host name of the iRODS server.
         /// \param[in] _port     The port to connect to.
         /// \param[in] _username The user to connect as.
-        /// \param[in] _zone     The zone managed by the iRODS server.
         ///
-        /// \since 4.2.9
-        client_connection(const std::string_view _host,
+        /// \since 4.3.1
+        client_connection(const std::string_view _host, const int _port, const fully_qualified_username& _username);
+
+        /// Connects to the iRODS server identified by the passed arguments
+        /// but does not perform any authentication.
+        ///
+        /// \throws irods::exception If an error occurred.
+        ///
+        /// \param[in] _defer_tag Indicates that authentication must be skipped.
+        /// \param[in] _host      The host name of the iRODS server.
+        /// \param[in] _port      The port to connect to.
+        /// \param[in] _username  The name of the user to connect as.
+        ///
+        /// \since 4.3.1
+        client_connection(struct defer_authentication _defer_tag,
+                          const std::string_view _host,
                           const int _port,
-                          const std::string_view _username,
-                          const std::string_view _zone);
+                          const fully_qualified_username& _username);
 
         /// Connects to the iRODS server identified by the passed arguments
         /// and authenticates the proxy user using the information and credentials
-        /// found in the user's irods_environment.json file
+        /// found in the user's irods_environment.json file.
         ///
         /// \throws irods::exception If an error occured.
         ///
-        /// \param[in] _host       The host name of the iRODS server.
-        /// \param[in] _port       The port to connect to.
-        /// \param[in] _proxy_user The name of the user to proxy
-        /// \param[in] _proxy_zone The name of the zone to proxy
-        /// \param[in] _username   The user to connect as.
-        /// \param[in] _zone       The zone managed by the iRODS server.
+        /// \param[in] _host           The host name of the iRODS server.
+        /// \param[in] _port           The port to connect to.
+        /// \param[in] _proxy_username The name of the user acting as the proxy.
+        /// \param[in] _username       The name of the user being proxied.
         ///
-        /// \since 4.3.0
+        /// \since 4.3.1
         client_connection(const std::string_view _host,
                           const int _port,
-                          const std::string_view _proxy_user,
-                          const std::string_view _proxy_zone,
-                          const std::string_view _username,
-                          const std::string_view _zone);
+                          const fully_qualified_username& _proxy_username,
+                          const fully_qualified_username& _username);
+
+        /// Connects to the iRODS server identified by the passed arguments
+        /// but does not perform any authentication.
+        ///
+        /// \throws irods::exception If an error occured.
+        ///
+        /// \param[in] _defer_tag      Indicates that authentication must be skipped.
+        /// \param[in] _host           The host name of the iRODS server.
+        /// \param[in] _port           The port to connect to.
+        /// \param[in] _proxy_username The name of the user acting as the proxy.
+        /// \param[in] _username       The name of the user being proxied.
+        ///
+        /// \since 4.3.1
+        client_connection(struct defer_authentication _defer_tag,
+                          const std::string_view _host,
+                          const int _port,
+                          const fully_qualified_username& _proxy_username,
+                          const fully_qualified_username& _username);
 
         /// Takes ownership of a raw iRODS connection and provides
         /// all of the safety guarantees found in other instantiations.
@@ -86,8 +131,10 @@ namespace irods::experimental
         /// Instantiations of this class via this constructor must call \ref connect
         /// to establish a connection and authenticate the user.
         ///
+        /// \param[in] _defer_tag Indicates that no connection will be made to the server.
+        ///
         /// \since 4.2.9
-        explicit client_connection(struct defer_connection);
+        explicit client_connection(struct defer_connection _defer_tag);
 
         client_connection(client_connection&& _other) = default;
         auto operator=(client_connection&& _other) -> client_connection& = default;
@@ -105,6 +152,16 @@ namespace irods::experimental
         /// \since 4.2.9
         auto connect() -> void;
 
+        /// Connects using the information and credentials found in the user's
+        /// irods_environment.json file. No authentication is performed.
+        ///
+        /// \throws irods::exception If an error occurred.
+        ///
+        /// \param[in] _defer_tag Indicates that authentication must be skipped.
+        ///
+        /// \since 4.3.1
+        auto connect(struct defer_authentication _defer_tag) -> void;
+
         /// Connects to the iRODS server identified by the passed arguments
         /// and authenticates the user using the information and credentials
         /// found in the user's irods_environment.json file.
@@ -113,27 +170,66 @@ namespace irods::experimental
         ///
         /// \param[in] _host     The host name of the iRODS server.
         /// \param[in] _port     The port to connect to.
-        /// \param[in] _username The user to connect as.
-        /// \param[in] _zone     The zone managed by the iRODS server.
+        /// \param[in] _username The name of the user to connect as.
         ///
-        /// \since 4.2.9
-        auto connect(const std::string_view _host,
+        /// \since 4.3.1
+        auto connect(const std::string_view _host, const int _port, const fully_qualified_username& _username) -> void;
+
+        /// Connects to the iRODS server identified by the passed arguments
+        /// but does not perform any authentication.
+        ///
+        /// \throws irods::exception If an error occurred.
+        ///
+        /// \param[in] _defer_tag Indicates that authentication must be skipped.
+        /// \param[in] _host      The host name of the iRODS server.
+        /// \param[in] _port      The port to connect to.
+        /// \param[in] _username  The name of the user to connect as.
+        ///
+        /// \since 4.3.1
+        auto connect(struct defer_authentication _defer_tag,
+                     const std::string_view _host,
                      const int _port,
-                     const std::string_view _username,
-                     const std::string_view _zone) -> void;
+                     const fully_qualified_username& _username) -> void;
+
+        /// Connects to the iRODS server identified by the passed arguments
+        /// but does not perform any authentication.
+        ///
+        /// \throws irods::exception If an error occurred.
+        ///
+        /// \param[in] _defer_tag      Indicates that authentication must be skipped.
+        /// \param[in] _host           The host name of the iRODS server.
+        /// \param[in] _port           The port to connect to.
+        /// \param[in] _proxy_username The name of the user acting as the proxy.
+        /// \param[in] _username       The name of the user to connect as.
+        ///
+        /// \since 4.3.1
+        auto connect(struct defer_authentication _defer_tag,
+                     const std::string_view _host,
+                     const int _port,
+                     const fully_qualified_username& _proxy_username,
+                     const fully_qualified_username& _username) -> void;
 
         /// Connects to the iRODS server with the specified proxy user and zone
-        /// authenticates the user using the information and credentials
+        /// and authenticates the proxy user using the information and credentials
         /// found in the user's irods_environment.json file.
         ///
         /// \throws irods::exception If an error occurred.
         ///
-        /// \param[in] _proxy_user The name of the user to proxy
-        /// \param[in] _proxy_zone The name of the zone to proxy
+        /// \param[in] _proxy_user The user acting as the proxy.
         ///
-        /// \since 4.3.0
-        auto connect(const std::string_view _proxy_user,
-                     const std::string_view _proxy_zone) -> void;
+        /// \since 4.3.1
+        auto connect(const fully_qualified_username& _proxy_user) -> void;
+
+        /// Connects to the iRODS server with the specified proxy user and zone
+        /// but does not perform any authentication.
+        ///
+        /// \throws irods::exception If an error occurred.
+        ///
+        /// \param[in] _defer_tag  Indicates that authentication must be skipped.
+        /// \param[in] _proxy_user The user acting as the proxy.
+        ///
+        /// \since 4.3.1
+        auto connect(struct defer_authentication _defer_tag, const fully_qualified_username& _proxy_user) -> void;
 
         /// Closes the underlying connection if active.
         ///
@@ -161,18 +257,21 @@ namespace irods::experimental
         /// \since 4.2.9
         explicit operator RcComm*() const noexcept;
 
-    private:
-        auto connect_and_login(const std::string_view _host,
-                               const int _port,
-                               const std::string_view _username,
-                               const std::string_view _zone) -> void;
+      private:
+        auto connect_and_login(const std::string& _host, const int _port, const fully_qualified_username& _username)
+            -> void;
 
-        auto connect_and_login(const std::string_view _host,
+        auto connect_and_login(const std::string& _host,
                                const int _port,
-                               const std::string_view _proxy_user,
-                               const std::string_view _proxy_zone,
-                               const std::string_view _username,
-                               const std::string_view _zone) -> void;
+                               const fully_qualified_username& _proxy_username,
+                               const fully_qualified_username& _username) -> void;
+
+        auto only_connect(const std::string& _host, const int _port, const fully_qualified_username& _username) -> void;
+
+        auto only_connect(const std::string& _host,
+                          const int _port,
+                          const fully_qualified_username& _proxy_username,
+                          const fully_qualified_username& _username) -> void;
 
         std::unique_ptr<RcComm, int (*)(RcComm*)> conn_;
     }; // class client_connection
--- a/lib/core/include/irods/connection_pool.hpp
+++ b/lib/core/include/irods/connection_pool.hpp
@@ -1,66 +1,197 @@
 #ifndef IRODS_CONNECTION_POOL_HPP
 #define IRODS_CONNECTION_POOL_HPP
 
+#include "irods/fully_qualified_username.hpp"
 #include "irods/rcConnect.h"
 
-#include <memory>
-#include <vector>
-#include <mutex>
 #include <atomic>
-#include <string>
 #include <functional>
+#include <memory>
+#include <mutex>
+#include <optional>
+#include <string>
+#include <string_view>
+#include <vector>
 
 namespace irods
 {
-    class connection_pool
+    /// Manages multiple iRODS connections for one or more users.
+    ///
+    /// Instances of this class are not copyable or moveable.
+    ///
+    /// \since 4.2.5
+    class connection_pool // NOLINT(cppcoreguidelines-special-member-functions)
     {
-    public:
-        // A wrapper around a connection in the pool.
-        // On destruction, the underlying connection is immediately returned
-        // to the pool.
-        class connection_proxy
+      public:
+        /// Manages a single connection within a connection_pool.
+        ///
+        /// On destruction, the underlying connection is immediately returned to the pool.
+        ///
+        /// Instances of this class are not copyable.
+        ///
+        /// \since 4.2.5
+        class connection_proxy // NOLINT(cppcoreguidelines-special-member-functions)
         {
-        public:
+          public:
             friend class connection_pool;
 
+            /// Constructs an empty connection_proxy.
+            ///
+            /// On success, the following assertions will be true:
+            /// - static_cast<RcComm*>(instance) will return a nullptr
+            /// - static_cast<RcComm&>(instance) will throw an irods::exception
+            ///
+            /// \since 4.2.9
             connection_proxy();
 
-            connection_proxy(connection_proxy&&);
-            connection_proxy& operator=(connection_proxy&&);
+            connection_proxy(connection_proxy&&) noexcept;
+            connection_proxy& operator=(connection_proxy&&) noexcept;
 
+            /// Destructs the connection_proxy and returns the underlying RcComm to the pool.
+            ///
+            /// \since 4.2.5
             ~connection_proxy();
 
-            operator bool() const noexcept;
+            /// Returns a boolean indicating whether the connection_proxy holds a valid RcComm.
+            ///
+            /// \retval true  If the underlying RcComm is valid.
+            /// \retval false Otherwise.
+            ///
+            /// \since 4.2.7
+            operator bool() const noexcept; // NOLINT(google-explicit-constructor)
 
-            operator rcComm_t&() const;
-            explicit operator rcComm_t*() const noexcept;
+            /// Returns a reference to the underlying RcComm.
+            ///
+            /// \throws irods::exception If the underlying RcComm is not valid (i.e. nullptr).
+            ///
+            /// \since 4.2.5
+            operator RcComm&() const; // NOLINT(google-explicit-constructor)
 
-            rcComm_t* release();
+            /// Returns a pointer to the underlying RcComm.
+            ///
+            /// If the underlying RcComm is not valid, a \p nullptr is returned.
+            ///
+            /// \since 4.2.7
+            explicit operator RcComm*() const noexcept;
 
-        private:
-            connection_proxy(connection_pool& _pool, rcComm_t& _conn, int _index) noexcept;
+            /// Transfers ownership of the underlying RcComm from the connection_proxy and
+            /// connection_pool to the caller.
+            ///
+            /// \since 4.2.7
+            RcComm* release();
+
+          private:
+            connection_proxy(connection_pool& _pool, RcComm& _conn, int _index) noexcept;
 
             static constexpr int uninitialized_index = -1;
 
             connection_pool* pool_;
-            rcComm_t* conn_;
+            RcComm* conn_;
             int index_;
-        };
+        }; // class connection_proxy
 
+        /// Constructs a connection_pool.
+        ///
+        /// Each connection in the pool is authenticated as \p _name (pound) _zone.
+        ///
+        /// \param[in] _size         The number of connections to create.
+        /// \param[in] _host         The hostname of the iRODS server.
+        /// \param[in] _port         The port to connect to.
+        /// \param[in] _name         The name of the user to connect as.
+        /// \param[in] _zone         The zone of the user to connect as.
+        /// \param[in] _refresh_time The number of seconds before a connection is refreshed.
+        ///
+        /// \throws irods::exception If an error occurs.
+        ///
+        /// \since 4.2.5
+        [[deprecated]] connection_pool(int _size,
+                                       const std::string& _host,
+                                       const int _port,
+                                       const std::string& _name,
+                                       const std::string& _zone,
+                                       const int _refresh_time);
+
+        /// Constructs a connection_pool.
+        ///
+        /// Each connection in the pool is authenticated as \p _username.
+        ///
+        /// \param[in] _size         The number of connections to create.
+        /// \param[in] _host         The hostname of the iRODS server.
+        /// \param[in] _port         The port to connect to.
+        /// \param[in] _username     The name of the user to connect as.
+        /// \param[in] _refresh_time The number of seconds before a connection is refreshed.
+        ///
+        /// \throws irods::exception If an error occurs.
+        ///
+        /// \since 4.3.1
         connection_pool(int _size,
                         const std::string& _host,
                         const int _port,
-                        const std::string& _username,
-                        const std::string& _zone,
+                        experimental::fully_qualified_username _username,
                         const int _refresh_time);
+
+        /// Constructs a connection_pool.
+        ///
+        /// Each connection in the pool is authenticated as \p _username.
+        ///
+        /// This constructor allows use of non-native authentication schemes (e.g. PAM, Kerberos, etc).
+        ///
+        /// \param[in] _size         The number of connections to create.
+        /// \param[in] _host         The hostname of the iRODS server.
+        /// \param[in] _port         The port to connect to.
+        /// \param[in] _username     The name of the user to connect as.
+        /// \param[in] _refresh_time The number of seconds before a connection is refreshed.
+        /// \param[in] _auth_func    The callback used for authenticating each connection.
+        ///
+        /// \throws irods::exception If an error occurs.
+        ///
+        /// \since 4.3.1
+        connection_pool(int _size,
+                        const std::string& _host,
+                        const int _port,
+                        experimental::fully_qualified_username _username,
+                        const int _refresh_time,
+                        std::function<void(RcComm&)> _auth_func);
+
+        /// Constructs a connection_pool.
+        ///
+        /// Each connection in the pool is authenticated as \p _proxy_username if provided.
+        /// Otherwise, the connections are authenticated as \p _username. API oeprations are
+        /// always executed as \p _username.
+        ///
+        /// This constructor allows use of non-native authentication schemes (e.g. PAM, Kerberos, etc).
+        ///
+        /// \param[in] _size           The number of connections to create.
+        /// \param[in] _host           The hostname of the iRODS server.
+        /// \param[in] _port           The port to connect to.
+        /// \param[in] _proxy_username The name of the user acting as the proxy.
+        /// \param[in] _username       The name of the user being proxied.
+        /// \param[in] _auth_func      The callback used for authenticating each connection.
+        /// \param[in] _refresh_time   The number of seconds before a connection is refreshed.
+        ///
+        /// \throws irods::exception If an error occurs.
+        ///
+        /// \since 4.3.1
+        connection_pool(int _size,
+                        std::string_view _host,
+                        const int _port,
+                        std::optional<experimental::fully_qualified_username> _proxy_username,
+                        experimental::fully_qualified_username _username,
+                        const int _refresh_time,
+                        std::function<void(RcComm&)> _auth_func);
 
         connection_pool(const connection_pool&) = delete;
         connection_pool& operator=(const connection_pool&) = delete;
 
+        /// Returns a connection from the pool.
+        ///
+        /// This function will block if all connections are in use.
+        ///
+        /// \since 4.2.5
         connection_proxy get_connection();
 
-    private:
-        using connection_pointer = std::unique_ptr<rcComm_t, int(*)(rcComm_t*)>;
+      private:
+        using connection_pointer = std::unique_ptr<RcComm, int (*)(RcComm*)>;
 
         struct connection_context
         {
@@ -70,13 +201,13 @@ namespace irods
             connection_pointer conn{nullptr, rcDisconnect};
             rErrMsg_t error{};
             std::time_t creation_time{};
-        };
+        }; // struct connection_context
 
         void create_connection(int _index,
-                               std::function<void()> _on_connect_error,
-                               std::function<void()> _on_login_error);
+                               const std::function<void()>& _on_connect_error,
+                               const std::function<void()>& _on_login_error);
 
-        rcComm_t* refresh_connection(int _index);
+        RcComm* refresh_connection(int _index);
 
         bool verify_connection(int _index);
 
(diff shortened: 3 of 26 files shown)


unit_tests/src/test_connection_pool.cpp
```diff
@@ -107,5 +113,129 @@ TEST_CASE("connection pool")
...
+        auto password = std::to_array("ppass");
+        const ia::user_password_property pwd_property{password.data()};
+        REQUIRE_NOTHROW(ia::client::modify_user(admin_conn, proxy_admin, pwd_property));
+
+        const auto auth_func = [](auto& _comm)
+        {
+            // This will cause the connection pool to throw an exception on construction.
+            if (clientLoginWithPassword(&_comm, "nope") != 0) {
+                THROW(SYS_SOCK_CONNECT_ERR, "Password authentication error");
+            }
```
korydraughn, line 226: For some reason, `SYS_SOCK_CONNECT_ERR` is not thrown. Instead, `CAT_INVALID_AUTHENTICATION` is thrown. I'm guessing this is due to the new authentication function related to the new authentication framework.

unit_tests/src/test_connection_pool.cpp
```diff
@@ -107,5 +113,129 @@ TEST_CASE("connection pool")
...
+
+    SECTION("connection pool throws exception on incorrect alternative authentication method")
+    {
+        REQUIRE_THROWS([&env] {
+            const auto auth_func = [](auto& _comm)
+            {
+                // This will cause the connection pool to throw an exception on construction.
+                if (clientLoginWithPassword(&_comm, "nope") != 0) {
+                    THROW(SYS_SOCK_CONNECT_ERR, "Password authentication error");
+                }
```
korydraughn, line 189: Same here.

unit_tests/src/test_connection_pool.cpp
```diff
@@ -107,5 +113,129 @@ TEST_CASE("connection pool")
...
+
+        const ia::user proxy_admin{"proxy_admin"};
+        REQUIRE_NOTHROW(ia::client::add_user(admin_conn, proxy_admin, ia::user_type::rodsadmin));
+
+        irods::at_scope_exit remove_proxy_admin{[&admin_conn, proxy_admin] {
+            ia::client::remove_user(admin_conn, proxy_admin);
+        }};
+
+        // The property is mutable so that we can use it for clientLoginWithPassword()
+        // later in the test.
```
korydraughn, line 151: Remove

unit_tests/src/test_connection_pool.cpp
```diff
@@ -107,5 +113,129 @@ TEST_CASE("connection pool")
...
-}
 
+    SECTION("connection pools allow the default authentication method to be overridden")
+    {
+        REQUIRE_NOTHROW([&env] {
+            const auto auth_func = [](auto& _comm)
+            {
+                if (clientLogin(&_comm) != 0) {
+                    THROW(SYS_SOCK_CONNECT_ERR, ".irodsA authentication error");
+                }
```
korydraughn, line 124: This is identical to what the connection pool does, but it demonstrates that it can be changed.

unit_tests/src/test_connection_pool.cpp
korydraughn: It appears that `CAT_INVALID_AUTHENTICATION` is being printed to the terminal. It's a result of the `clientLogin*` functions using `fprintf`.

They should not be printing messages to the terminal directly.

I'll write up an issue for this.

unit_tests/src/test_connection_pool.cpp
korydraughn: Done.

unit_tests/src/test_connection_pool.cpp
korydraughn: Done.

unit_tests/src/test_connection_pool.cpp
korydraughn: Adjusting the test proved that my assumptions about the `clientLogin*` functions and `fprintf`.

korydraughn: Created #7048.

Done.

lib/core/src/connection_pool.cpp
```diff
@@ -87,23 +88,61 @@ namespace irods
...
 
+        if (!proxy_username_.empty() && proxy_zone_.empty()) {
+            throw std::runtime_error{"invalid proxy zone"};
+        }
+
+        if (proxy_username_.empty() && !proxy_zone_.empty()) {
+            throw std::runtime_error{"invalid proxy username"};
+        }
+
+        const auto use_proxy_connection = (!proxy_username_.empty() && !proxy_zone_.empty());
```
korydraughn, line 137: Consider designing a better approach than relying on the emptiness of these strings.

korydraughn:
Don't forget to add the necessary macros to library_features.h.

alanking (commented):
All of this makes sense so far. Couple of miniscule things

lib/core/include/irods/client_connection.hpp
```diff
@@ -47,14 +62,31 @@ namespace irods::experimental
...
                           const int _port,
                           const std::string_view _username,
                           const std::string_view _zone);
 
+        /// Connects to the iRODS server identified by the passed arguments
+        /// but does not perform any authentication.
+        ///
+        /// \throws irods::exception If an error occurred.
+        ///
+        /// \param[in] _host     The host name of the iRODS server.
```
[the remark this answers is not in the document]
korydraughn: Hmm, I couldn't decide on whether I wanted to provide a parameter. I'll check how other projects handle tag types and provide my findings here.

lib/core/include/irods/client_connection.hpp
```diff
@@ -14,13 +14,20 @@ struct RcComm;
...
     /// should allow the user to control when the connection and
     /// authentication occurs.
     ///
     /// \since 4.2.9
     inline const struct defer_connection {} defer_connection;
+
+    /// A tag type that indicates whether or not a client connection
+    /// should allow the user to control when the authentication occurs.
+    ///
+    /// \since 4.2.12
```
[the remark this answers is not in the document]
korydraughn: You're right. Good catch.

Will update all references where it makes sense.

lib/core/include/irods/client_connection.hpp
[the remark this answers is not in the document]
korydraughn: cppreference.com doesn't give a name to the tag struct, but they also document things differently in general.

We need to come up with a good parameter name for the tag struct argument.

Ideas ...
- `_tag`
- `_flag`
- `_defer_tag`

Any other ideas?

lib/core/src/connection_pool.cpp
korydraughn: Probably better to make these `std::optional<std::string>`s instead of strings. It better expresses the intent and leaves room to treat an empty string as a valid argument.

May need to introduce a `user` type  (and a `proxy_user` type) to hold this info. However, we already have https://github.com/irods/irods/blob/1205fa3afd8de1e96e0e412184c7b30ae0b07c31/lib/administration/user/include/irods/user.hpp.

lib/core/include/irods/client_connection.hpp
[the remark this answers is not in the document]
trel: yeah, weird.

`_defer_struct` ?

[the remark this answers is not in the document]
korydraughn: I'm going to call it `_defer_tag`.

The name of this value doesn't matter that much. Developers will have to read the docs to see what it's for as it is aimed at special cases.

Ultimately, we just want something to refer to for documentation purposes.

_And we can always rename it later if it turns out to be terrible._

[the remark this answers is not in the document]
trel: perfect.

lib/core/src/connection_pool.cpp
korydraughn: Implemented `unique_user` class to enforce constraints for user identity info and connecting to a server.

lib/core/include/irods/client_connection.hpp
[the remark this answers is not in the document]
korydraughn: Done.

lib/core/include/irods/client_connection.hpp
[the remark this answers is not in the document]
korydraughn: Done.

lib/core/src/connection_pool.cpp
```diff
@@ -182,30 +223,31 @@ namespace irods
...
         }
 
         return ctx.conn.get();
-    }
+    } // refresh_connection
 
     connection_pool::connection_proxy connection_pool::get_connection()
     {
-        for (int i = 0;; i = ++i % conn_ctxs_.size()) {
+        // TODO Resolves irods/irods#5799
```
korydraughn, line 249: Add issue number to commit message.

korydraughn:
Add tests for ...
- unique_user.hpp
- connection_pool / unique_user / invalid args for constructor
- client_connection / unique_user / invalid args for constructor

Need to fix build errors from changes.
- delay server
- unit tests

lib/core/include/irods/connection_pool.hpp
```diff
@@ -86,14 +179,24 @@ namespace irods
...
+    /// Constructs a connection pool on the heap.
+    ///
+    /// The connection is constructed using client credentials defined in
+    /// $HOME/.irods/irods_environment.json.
+    ///
+    /// \param[in] _size The number of connections to create.
+    ///
+    /// \returns A shared pointer to a connection pool.
+    ///
+    /// \since ??? TODO
```
korydraughn, line 198: Handle TODO

lib/core/include/irods/unique_user.hpp
```diff
@@ -0,0 +1,90 @@
...
+namespace irods::experimental
+{
+    /// A class designed to hold a fully-qualified iRODS username.
+    ///
+    /// This class should be used when a fully-qualified iRODS username is required.
+    ///
+    /// Instances of this class are immutable outside of copy and move opertaions.
+    ///
+    /// \since 4.3.1
+    class unique_user
```
korydraughn, line 20: Is `fq_username` a better name?
How about `unique_username`?
Does `unique_user` convey things clearly?

`user` kind of implies `username` so I think all of these work.

Will think about this a little more.

unit_tests/src/test_client_connection.cpp
```diff
@@ -118,13 +131,141 @@ TEST_CASE("supports implicit and explicit casts to underlying type", "client_con
...
+        test_deferred_authentication(conn, "/");
+    }
+
+    SECTION("construct with user-provided arguments")
+    {
+        rodsEnv env;
+        _getRodsEnv(env);
+
+        ix::unique_user user{env.rodsUserName, env.rodsZone};
+
```
korydraughn, line 184: Remove empty line.

lib/core/include/irods/library_features.h
```diff
@@ -1,6 +1,10 @@
 #ifndef IRODS_LIBRARY_FEATURES_H
 #define IRODS_LIBRARY_FEATURES_H
 
+// clang-format off
+
+#ifndef IRODS_HIDE_FROM_CLANG_TIDY
```
korydraughn, line 6: Need to document this somewhere. Perhaps the developer docs.

lib/core/include/irods/connection_pool.hpp
```diff
@@ -2,65 +2,158 @@
...
-        class connection_proxy
+      public:
+        /// Manages a single connection within a connection pool.
+        ///
+        /// On destruction, the underlying connection is immediately returned to the pool.
+        ///
+        /// Instances of this class are NOT copyable.
+        ///
+        /// \since 4.2.5
+        class connection_proxy // NOLINT(cppcoreguidelines-special-member-functions)
```
korydraughn, line 33: Document member functions, etc.

lib/core/include/irods/client_connection.hpp
```diff
@@ -174,6 +355,18 @@ namespace irods::experimental
...
+                          const int _port,
+                          const std::string_view _username,
+                          const std::string_view _zone) -> void;
+
+        auto only_connect(const std::string_view _host,
+                          const int _port,
+                          const std::string_view _proxy_user,
+                          const std::string_view _proxy_zone,
+                          const std::string_view _username,
+                          const std::string_view _zone) -> void;
```
korydraughn, line 368: All `std::string_view`'s should probably be changed to `const std::string&` for safety purposes.

.github/workflows/linter-irods-clang-tidy.yml
```diff
@@ -75,7 +75,7 @@ jobs:
 
                   # Run clang-tidy on the changes.
                   git diff -U0 origin/$GITHUB_BASE_REF | \
-                      clang-tidy-diff.py -p1 -path build/compile_commands.json -quiet | \
+                      clang-tidy-diff.py -p1 -path build/compile_commands.json -extra-arg='-DIRODS_HIDE_FROM_CLANG_TIDY' -quiet | \
```
korydraughn, line 78: This needs to be applied to all projects.

.clang-tidy
```diff
@@ -60,11 +66,13 @@ WarningsAsErrors:
...
     misc-*,
     modernize-*,
+    -modernize-use-trailing-return-type,
     performance-*,
     portability-std-allocator-const,
     readability-*,
     -readability-avoid-const-params-in-decls,
-    -readability-implicit-bool-conversion
+    -readability-implicit-bool-conversion,
+    -readability-named-parameter
```
korydraughn, line 75: Create issues for these. Perhaps others in here too.

lib/core/include/irods/library_features.h
```diff
@@ -28,4 +32,11 @@
 /// \since 4.3.1
 #define IRODS_HAS_LIBRARY_SYSTEM_ERROR        202301L
 
+/// Defined if the C++ client connection libraries support proxy users.
+///
+/// \since 4.3.1
+#define IRODS_HAS_PROXY_USER_SUPPORT_FOR_CLIENT_CONNECTION_LIBRARIES    202306L
```
korydraughn, line 38: Perhaps this should be renamed to one of the following ...
```cpp
#define IRODS_HAS_ENHANCEMENT_PROXY_USERS_FOR_CONNECTION_LIBRARIES    202306L // A
#define IRODS_HAS_STRONG_TYPES_FOR_CPP_CONNECTION_LIBRARIES           202306L // B
```

Or something entirely different.

Currently, these macros are defined using the following convention ...
```cpp
#define IRODS_HAS_<type_of_thing_or_category>_<good_short_description>    YYYYMML
```

trel: A great naming discush.

.github/workflows/linter-irods-clang-tidy.yml
korydraughn: Also, create issue for this.

lib/core/include/irods/connection_pool.hpp
korydraughn: This function was added in 4.2.8.

.clang-tidy
```diff
@@ -28,14 +31,17 @@ Checks:
...
     portability-std-allocator-const,
     portability-simd-intrinsics,
     readability-*,
-    -readability-avoid-const-params-in-decls
+    -readability-avoid-const-params-in-decls,
+    -readability-function-cognitive-complexity
 WarningsAsErrors:
     boost-*,
     bugprone-*,
+    -bugprone-easily-swappable-parameters,
```
korydraughn, line 44: Consider removing this from this section as it is good to think about when designing new APIs.

.clang-tidy
korydraughn, line 75: Create issue for this.

.github/workflows/linter-irods-clang-tidy.yml
korydraughn: Then again, we can handle this for other projects as needed.

.clang-tidy
```diff
@@ -28,14 +31,17 @@ Checks:
     hicpp-signed-bitwise,
     misc-*,
     modernize-*,
+    -modernize-use-trailing-return-type,
     performance-*,
     portability-std-allocator-const,
     portability-simd-intrinsics,
     readability-*,
-    -readability-avoid-const-params-in-decls
+    -readability-avoid-const-params-in-decls,
```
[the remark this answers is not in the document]
korydraughn: Done.

.clang-tidy
korydraughn: Done.

korydraughn: Done.

lib/core/include/irods/connection_pool.hpp
korydraughn: Done.

unit_tests/src/test_client_connection.cpp
korydraughn: Done.

.clang-tidy
```diff
@@ -11,9 +11,12 @@ Checks:
     cert-env33-c,
     cert-err34-c,
     cert-err*-cpp,
+    -cert-err58-cpp,
```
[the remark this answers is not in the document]
korydraughn: Captured in commit message.

.clang-tidy
```diff
@@ -11,9 +11,12 @@ Checks:
...
     cert-flp30-c,
     cert-oop*,
     cppcoreguidelines-*,
+    -cppcoreguidelines-avoid-magic-numbers,
```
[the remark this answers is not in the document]
korydraughn: Captured in commit message.

.clang-tidy
```diff
@@ -11,9 +11,12 @@ Checks:
...
+    -cppcoreguidelines-pro-bounds-array-to-pointer-decay,
```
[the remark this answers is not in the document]
korydraughn: Captured in commit message.

.clang-tidy
```diff
@@ -28,14 +31,17 @@ Checks:
     hicpp-signed-bitwise,
     misc-*,
     modernize-*,
+    -modernize-use-trailing-return-type,
```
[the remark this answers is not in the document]
korydraughn: Captured in commit message.

.clang-tidy
```diff
@@ -28,14 +31,17 @@ Checks:
...
     performance-*,
     portability-std-allocator-const,
     portability-simd-intrinsics,
     readability-*,
-    -readability-avoid-const-params-in-decls
+    -readability-avoid-const-params-in-decls,
+    -readability-function-cognitive-complexity
```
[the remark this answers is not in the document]
korydraughn: Captured in commit message.

.clang-tidy
korydraughn: Removed.

lib/core/src/connection_pool.cpp
korydraughn: Just needs squashing and update to commit message.

lib/core/include/irods/library_features.h
korydraughn: After discussion, we've decided to go with the following.
```cpp
IRODS_HAS_FEATURE_<description>
```

lib/core/include/irods/unique_user.hpp
korydraughn: After discussion, we've decided to go with the following.
```cpp
fully_qualified_username fqun{"rods", "tempZone"};
```

.clang-tidy
```diff
@@ -11,9 +12,12 @@ Checks:
     cert-env33-c,
     cert-err34-c,
     cert-err*-cpp,
+    -cert-err58-cpp,
```
SwooshyCueb, line 15: I don't think we should disable this check entirely. I think it should be demoted to a warning instead. Thoughts?

.clang-tidy
```diff
@@ -11,9 +12,12 @@ Checks:
...
     cert-flp30-c,
     cert-oop*,
     cppcoreguidelines-*,
+    -cppcoreguidelines-avoid-magic-numbers,
```
SwooshyCueb, line 19: I don't think we should disable this check entirely. I think it should be demoted to a warning instead. Thoughts?

.clang-tidy
```diff
@@ -11,9 +12,12 @@ Checks:
...
+    -cppcoreguidelines-pro-bounds-array-to-pointer-decay,
```
SwooshyCueb, line 20: I am in favor of demoting this check to a warning at the very least, but I am unsure about disabling it entirely. Thoughts?

SwooshyCueb (changes requested):

.clang-tidy
korydraughn: Ohhh, I just realized that I disabled this entirely. I meant to only demote it.

Will update config.

.clang-tidy
korydraughn: Same here. I meant to demote it. Will update.

.clang-tidy
korydraughn: That's acceptable. Will demote to warning.

lib/core/include/irods/library_features.h
korydraughn: Removing this.

lib/core/include/irods/library_features.h

.clang-tidy
korydraughn: Done.

.clang-tidy
korydraughn: Done.

.clang-tidy
korydraughn: Done.

lib/core/include/irods/unique_user.hpp
korydraughn: Updated.

lib/core/include/irods/library_features.h

lib/core/include/irods/client_connection.hpp
korydraughn: Updated.

lib/core/include/irods/connection_pool.hpp
korydraughn: Done.

lib/core/src/connection_pool.cpp
```diff
@@ -36,74 +41,109 @@ namespace irods
...
-        , zone_{_zone}
+        , proxy_username_{std::move(_proxy_username)}
+        , username_{std::move(_username)}
         , refresh_time_(_refresh_time)
+        , auth_func_{std::move(_auth_func)}
         , conn_ctxs_(_size)
     {
         if (_size < 1) {
-            throw std::runtime_error{"invalid connection pool size"};
+            throw std::runtime_error{"Invalid connection pool size"}; // TODO These should be irods::exceptions.
```
korydraughn, line 137: Handle TODO

lib/core/include/irods/connection_pool.hpp
```diff
@@ -1,66 +1,198 @@
...
-        // A wrapper around a connection in the pool.
-        // On destruction, the underlying connection is immediately returned
-        // to the pool.
-        class connection_proxy
+      public:
+        /// Manages a single connection within a connection pool.
+        ///
+        /// On destruction, the underlying connection is immediately returned to the pool.
+        ///
+        /// Instances of this class are NOT copyable.
```
korydraughn, line 30: Or movable.

lib/core/include/irods/connection_pool.hpp
```diff
@@ -1,66 +1,198 @@
...
+                        const int _port,
+                        std::optional<experimental::fully_qualified_username> _proxy_username,
+                        experimental::fully_qualified_username _username,
+                        const int _refresh_time,
+                        std::function<void(RcComm&)> _auth_func);
+
         connection_pool(const connection_pool&) = delete;
         connection_pool& operator=(const connection_pool&) = delete;
 
+        /// Returns a connection from the pool.
```
korydraughn, line 189: Add a statement about this call blocking if all connections are in use.

lib/core/include/irods/connection_pool.hpp
```diff
@@ -86,14 +218,24 @@ namespace irods
...
         std::vector<connection_context> conn_ctxs_;
-    };
-
-    std::shared_ptr<connection_pool> make_connection_pool(int size = 1);
+    }; // class connection_pool
+
+    /// Constructs a connection pool on the heap.
+    ///
+    /// The connection pool is constructed using client credentials defined in
+    /// irods_environment.json.
```
korydraughn, line 231: "defined in ..." -> "found in the user's"

lib/core/include/irods/connection_pool.hpp
```diff
@@ -1,66 +1,198 @@
...
-        private:
-            connection_proxy(connection_pool& _pool, rcComm_t& _conn, int _index) noexcept;
+            /// Returns a boolean indicating whether the connection_proxy holds a valid RcComm.
+            ///
+            /// If a connection_proxy returns true, it can be used to execute API operations.
+            ///
+            /// \since 4.2.7
+            operator bool() const noexcept; // NOLINT(google-explicit-constructor)
+
+            /// Returns reference to the underlying RcComm.
```
korydraughn, line 62: "a reference"

lib/core/include/irods/connection_pool.hpp
```diff
@@ -1,66 +1,198 @@
...
-            operator rcComm_t&() const;
-            explicit operator rcComm_t*() const noexcept;
-
-            rcComm_t* release();
-
-        private:
-            connection_proxy(connection_pool& _pool, rcComm_t& _conn, int _index) noexcept;
+            /// Returns a boolean indicating whether the connection_proxy holds a valid RcComm.
+            ///
+            /// If a connection_proxy returns true, it can be used to execute API operations.
```
korydraughn, line 57: Rephrase or remove.

lib/core/include/irods/connection_pool.hpp
```diff
@@ -1,66 +1,198 @@
...
+            operator bool() const noexcept; // NOLINT(google-explicit-constructor)
+
+            /// Returns reference to the underlying RcComm.
+            ///
+            /// \throws irods::exception If the connection_proxy does not manage a valid RcComm.
+            ///
+            /// \since 4.2.5
+            operator RcComm&() const; // NOLINT(google-explicit-constructor)
+
+            /// Returns pointer to the underlying RcComm.
```
korydraughn, line 69: "a pointer"

lib/core/include/irods/connection_pool.hpp
```diff
@@ -1,66 +1,198 @@
...
+            ///
+            /// If the connection_proxy is valid, the pointer will be non-null. Otherwise,
+            /// a \p nullptr is returned.
```
korydraughn, line 72: Rephrase

lib/core/include/irods/connection_pool.hpp
korydraughn: Done.

lib/core/include/irods/connection_pool.hpp
korydraughn: Done.

lib/core/include/irods/connection_pool.hpp
korydraughn: Done.

lib/core/include/irods/fully_qualified_username.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+#include <string_view>
+
+namespace irods::experimental
+{
+    /// A class designed to hold a fully-qualified iRODS username.
+    ///
+    /// This class should be used when a fully-qualified iRODS username is required.
+    ///
+    /// Instances of this class are immutable outside of copy and move opertaions.
```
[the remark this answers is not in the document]
korydraughn: Done.

lib/core/include/irods/connection_pool.hpp
korydraughn: Done.

lib/core/include/irods/connection_pool.hpp
korydraughn: Done.

lib/core/include/irods/connection_pool.hpp
korydraughn: Done.

lib/core/include/irods/connection_pool.hpp
korydraughn: Done.

korydraughn:
This PR is officially open for review!

lib/core/src/client_connection.cpp
```diff
@@ -32,118 +60,225 @@ namespace irods::experimental
...
         return static_cast<bool>(conn_);
-    }
+    } // operator bool
 
     client_connection::operator RcComm&() const
     {
         if (!conn_) {
-            THROW(USER_SOCK_CONNECT_ERR, "invalid client_connection object");
+            // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+            THROW(USER_SOCK_CONNECT_ERR, "Invalid client_connection object");
```
korydraughn, line 208: Throw the same exception as connection_pool

lib/core/src/client_connection.cpp
```diff
@@ -32,118 +60,225 @@ namespace irods::experimental
...
-                              _zone.data(),
-                              NO_RECONN,
-                              &error));
+        only_connect(_host, _port, _username, _zone);
 
-        if (!conn_) {
-            THROW(USER_SOCK_CONNECT_ERR, "connect error");
+        if (clientLogin(conn_.get()) != 0) {
+            // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+            THROW(USER_SOCK_CONNECT_ERR, "Client login error");
```
korydraughn, line 228: Throw the same exception as connection_pool

lib/core/src/client_connection.cpp
```diff
@@ -32,118 +60,225 @@ namespace irods::experimental
...
+                                              const std::string& _proxy_zone,
+                                              const std::string& _username,
+                                              const std::string& _zone) -> void
+    {
+        only_connect(_host, _port, _proxy_user, _proxy_zone, _username, _zone);
 
         if (clientLogin(conn_.get()) != 0) {
-            THROW(USER_SOCK_CONNECT_ERR, "client login error");
+            // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+            THROW(USER_SOCK_CONNECT_ERR, "Client login error");
```
korydraughn, line 243: Throw the same exception as connection_pool

lib/core/include/irods/rodsErrorTable.h
```diff
@@ -243,6 +243,8 @@ NEW_ERROR(RESOLUTION_ERROR,                            -174000)
 NEW_ERROR(INVALID_HANDLE,                              -175000)
 NEW_ERROR(DOES_NOT_EXIST,                              -176000)
 NEW_ERROR(TYPE_NOT_SUPPORTED,                          -177000)
+NEW_ERROR(ILLEGAL_ACCESS_VIOLATION,                    -178000)
+NEW_ERROR(AUTHENTICATION_ERROR,                        -179000)
```
korydraughn, line 247: Notice the new error codes.

korydraughn:
All unit tests pass on ub20/postgres.

Running core tests now.

lib/core/include/irods/library_features.h
```diff
@@ -1,6 +1,8 @@
 #ifndef IRODS_LIBRARY_FEATURES_H
 #define IRODS_LIBRARY_FEATURES_H
 
+// NOLINTBEGIN
```
alanking, line 4: Is this intentional? Feels like there should be an end, or a comment explaining that this is intentional.

alanking (commented):
Looks good overall

lib/core/include/irods/client_connection.hpp
```diff
@@ -47,34 +66,98 @@ namespace irods::experimental
         /// \param[in] _host     The host name of the iRODS server.
         /// \param[in] _port     The port to connect to.
         /// \param[in] _username The user to connect as.
-        /// \param[in] _zone     The zone managed by the iRODS server.
+        /// \param[in] _zone     The zone of the user.
         ///
         /// \since 4.2.9
-        client_connection(const std::string_view _host,
+        [[deprecated]] client_connection(const std::string_view _host,
```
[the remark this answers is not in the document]
korydraughn: That's true. I only marked it deprecated because other projects we manage likely use these constructors, but as we already declared, anything within the experimental namespace can be torched at anytime.

Will think about this more.

lib/core/include/irods/fully_qualified_username.hpp
```diff
@@ -0,0 +1,92 @@
...
+    /// \since 4.3.1
+    class fully_qualified_username
+    {
+      public:
+        /// Constructs a fully_qualified_username.
+        ///
+        /// \param[in] _name The part of the username preceding the pound sign.
+        /// \param[in] _zone The part of the username following the pound sign.
+        ///
+        /// \throws irods::exception If an error occurs.
```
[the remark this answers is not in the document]
korydraughn: I've started moving away from mentioning that because it's difficult to maintain and keep true.

I personally think letting devs know that a particular type of exception can be throw is better. With that, they know they have to catch _N_ types of exceptions. From there, they can interrogate the exceptions for what caused it.

Thoughts?

lib/core/include/irods/fully_qualified_username.hpp
```diff
@@ -0,0 +1,92 @@
...
+        ///
+        /// \since 4.3.1
+        fully_qualified_username(const std::string_view _name, const std::string_view _zone)
```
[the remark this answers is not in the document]
korydraughn: If we want devs to have to include the type name for anonymous instances, then yes. It's really up to us whether we feel that is a good idea or not.

I don't feel it is necessary for this case, but I'm willing to be persuaded.

lib/core/include/irods/connection_pool.hpp
```diff
@@ -1,66 +1,197 @@
...
-                        const std::string& _username,
-                        const std::string& _zone,
+                        experimental::fully_qualified_username _username,
                         const int _refresh_time);
 
+        /// Constructs a connection_pool.
+        ///
+        /// Each connection in the pool is authenticated as \p _username.
+        ///
+        /// This constructor allows use of other authentication schemes (e.g. PAM).
```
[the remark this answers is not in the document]
korydraughn: Interesting. Do you feel the word, **other**, is too loose?

lib/core/src/connection_pool.cpp
```diff
@@ -117,51 +157,68 @@ namespace irods
...
         ctx.creation_time = std::time(nullptr);
-        ctx.conn.reset(rcConnect(host_.c_str(),
-                                 port_,
-                                 username_.c_str(),
-                                 zone_.c_str(),
-                                 NO_RECONN,
-                                 &ctx.error));
+
+        if (proxy_username_.has_value()) {
+            ctx.conn.reset(_rcConnect(host_.c_str(),
```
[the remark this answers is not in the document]
korydraughn: Correct.

unit_tests/src/test_connection_pool.cpp
```diff
@@ -107,5 +156,109 @@ TEST_CASE("connection pool")
...
+            };
+
+            irods::experimental::fully_qualified_username user{env.rodsUserName, env.rodsZone};
+            irods::connection_pool conn_pool{cp_size, env.rodsHost, env.rodsPort, user, cp_refresh_time, auth_func};
+        }());
+    }
+
+    SECTION("connection pools support proxy connections")
```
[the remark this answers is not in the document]
korydraughn: Yes. The connections in the pool are proxied connections.

Perhaps this should be changed to something like ...
```
connection pools support proxied connections
```

Thoughts?

lib/core/include/irods/library_features.h
korydraughn: Yes. This is intentional.

Clang-Tidy needs to ignore this file entirely because it is meant to be used for C and C++. We can't use anything other than macros unless we're going to require devs to link a shared library.

The `// NOLINTEND` is necessary, but I don't mind adding it if that helps.

korydraughn:
All core tests passed.

Will address the comments. None of the comments require rerunning the tests so we're good there.

lib/core/src/client_connection.cpp
korydraughn: Done.

lib/core/src/client_connection.cpp
korydraughn: Done

lib/core/src/client_connection.cpp
korydraughn: Done

lib/core/include/irods/connection_pool.hpp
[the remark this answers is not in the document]
korydraughn: I think changing this to use `non-native` is fine.

Will update.

lib/core/include/irods/fully_qualified_username.hpp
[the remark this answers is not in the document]
alanking: I think that's fine.

Just thought it could be helpful to mention that `_name` and `_zone` are required to be non-empty, whether that's in this part or in the input parameter descriptions. No need to do anything here.

lib/core/include/irods/fully_qualified_username.hpp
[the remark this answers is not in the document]
alanking: Yeah, I don't think it's really that important. We can always reference the docs/signatures for functions that use this to see what the type is supposed to be if it's not clear.

lib/core/include/irods/fully_qualified_username.hpp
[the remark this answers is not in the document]
korydraughn: Ah. That makes sense.

Will update the parameter descriptions.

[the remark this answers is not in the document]
korydraughn: Hmm, then again the name of the type says it all. This is hard lol.

Will think about this more... again :-)

lib/core/include/irods/connection_pool.hpp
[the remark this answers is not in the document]
alanking: I felt that "other" was a little ambiguous. I found myself asking "Other than _what_?", eventually reasoning that it must be referring to native authentication as the default.

unit_tests/src/test_connection_pool.cpp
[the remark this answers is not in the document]
alanking: Yes, I think that's better.

lib/core/include/irods/client_connection.hpp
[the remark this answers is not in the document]
korydraughn: I think I'm going to remove these constructors.

Nice catch on the `experimental` namespace. It's working!

lib/core/include/irods/library_features.h
alanking: Okay, I figured as much. Just wasn't sure if it was a mistake that there was no end. I think it would be helpful to add the `NOLINTEND` or an explanatory comment to the effect of what you've said here. Not required, of course, since it works.

lib/core/include/irods/fully_qualified_username.hpp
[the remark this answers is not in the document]
korydraughn: Will update the documentation to include more words. It's just never a bad idea to be explicit. In this particular case, it's obvious what to write.

lib/core/include/irods/connection_pool.hpp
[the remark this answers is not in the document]
korydraughn: Updated.

lib/core/include/irods/library_features.h
korydraughn: Added words above clang-tidy directive.

korydraughn:
Reran the unit tests because I changed the internals of client_connection.cpp.

All unit tests pass, so I think this PR is ready for merging if everyone agrees the changes are good.

lib/core/include/irods/rodsErrorTable.h
korydraughn: @trel @alanking Any thoughts on these names?

alanking: I realize the direction we are headed is more "generic" error codes, so, I don't really have anything better to suggest for these. They seem fine.

`ILLEGAL_ACCESS_VIOLATION` seemed to be used mainly when attempting to dereference a nullptr for the type-casting functions, right? I was tempted to suggest something like `NULLPTR_DEREFERENCE` but it doesn't communicate any more information than what you have - that is, it is a subset of the existing name you have now.

`AUTHENTICATION_ERROR` is okay as long as `clientLogin` or whatever is returning a little more information in the `rError` stack or something (but... maybe not *too much* information...)

korydraughn: What info would you like to see returned in the rError stack on a failed authentication attempt?

Perhaps the following ...
- username#zone
- hostname
- zone

Or is there other info you feel should be included?

trel: don't we already have an equivalent to `AUTHENTICATION_ERROR`?   or maybe we would transition the older one to this one over time...

korydraughn: > `ILLEGAL_ACCESS_VIOLATION` is a weird ... phrase... is one of these better? `ILLEGAL_ACCESS` or `ILLEGAL_MEMORY_ACCESS` or `ACCESS_VIOLATION`

`ILLEGAL_ACCESS` or `ACCESS_VIOLATION` work.

I think I like `ILLEGAL_ACCESS`, but if we already use `VIOLATION` somewhere, then I think that wins.

I'll verify this and make adjustments.

---

As for the auth error code, we have `CAT_INVALID_AUTHENTICATION`, but that prefix bothers me because it is catalog-specific and that may not be true for every case. This new code is generic enough that it can be used across all auth error cases.

trel: these seem most useful.
- username#zone
- connected server hostname
- perhaps also server hostname of provider
- type of auth attempted?

alanking: I'm not saying that `AUTHENTICATION_ERROR` has to do any more work, but `clientLogin` may have messages and/or error codes from the auth plugins that may be masked by throwing this. Just want to make sure that the intended communications make it back to the user. Having more information probably isn't a bad thing

korydraughn: Okay. I'll capture this conversation in a new issue so this PR can be merged.

Making those changes will apply to both connection classes and should be trivial to implement.

korydraughn: We don't use `ILLEGAL` or `VIOLATION` in any error code names.

Either would work, but I think `ILLEGAL` works better in that it would be used as a prefix on the error code name. That would allow similar errors to be grouped. For example ...
- `ILLEGAL_ACCESS`
- `ILLEGAL_OPERATION`
- `ILLEGAL_MODIFICATION`

Do either of you have a preference on this?

alanking: `ILLEGAL_ACCESS_VIOLATION` seems better to me. "Violation" is pretty commonly used for these types of situations in C/C++ land, and leaving it off could be conflated with "access" to, say, a data object, or something else in the iRODS system. Just my two cents

korydraughn: Good point. I stole those words from Java. I think they have an exception that uses those words. Or maybe that was nfs4j.

korydraughn: I think I also see @trel's point about that name.

@trel Correct me if I'm wrong, but are you getting at the use of **illegal** and **violation** being used in the same name?

I guess he's saying they mean the same thing so why have both.

korydraughn: Alternatively, how do we feel about `ILLEGAL_ACCESS_ERROR`?

That is equivalent to java's `IllegalAccessError` exception. See https://docs.oracle.com/javase/8/docs/api/java/lang/IllegalAccessError.html.

korydraughn: Well, I'm not a big fan of `_ERROR` since these are error codes.

trel: yeah, no `_ERROR`

trel: > We don't use `ILLEGAL` or `VIOLATION` in any error code names.
> 
> Either would work, but I think `ILLEGAL` works better in that it would be used as a prefix on the error code name. That would allow similar errors to be grouped. For example ...
> 
> * `ILLEGAL_ACCESS`
> * `ILLEGAL_OPERATION`
> * `ILLEGAL_MODIFICATION`
> 
> Do either of you have a preference on this?

illegal in this sense... means... invalid?  or not allowed due to permissions?

korydraughn: It is due to the rules of the class. You get this error if you attempt to call a function and the pre-condition for calling it is violated.

For example ...
```cpp
client_connection conn;                      // connect to irods.
auto other_conn = std::move(conn);           // transfer ownership of the RcComm to other_conn.
auto& conn_ref = static_cast<RcComm&>(conn); // BOOM! conn doesn't have a valid RcComm.
```

korydraughn: After some discussion, @alanking and I determined that we should drop the `ILLEGAL_ACCESS_VIOLATION` error code and use `SYS_LIBRARY_ERROR`.

The reason for this is that iRODS error codes are defined for users of the iRODS system, not developers. And, it could be misleading to return such error codes.

This also means we need to define rules around what constitutes a new error code.

trel: oh, well that's nice.

korydraughn: Done.

korydraughn:
Running unit tests one more time before adding pounds.

korydraughn:
All unit tests passed.

Adding pounds.

korydraughn:
Added #'s

alanking (approved):

PR #7049, opened 2023-04-18
[6918] Correct msiAddKeyVal behavior (4-2-stable)

MartinFlores751:
`msiAddKeyValToMspStr()` now can handle empty strings/key-value pairs properly.

### DIFF
--- a/lib/core/src/msParam.cpp
+++ b/lib/core/src/msParam.cpp
@@ -1451,60 +1451,61 @@ int
 addKeyValToMspStr( msParam_t * keyStr, msParam_t * valStr,
                    msParam_t * msKeyValStr ) {
 
-    if ( ( keyStr == NULL && valStr == NULL ) || msKeyValStr == NULL ) {
+    if ((keyStr == nullptr && valStr == nullptr) || msKeyValStr == nullptr) {
         return SYS_INTERNAL_NULL_INPUT_ERR;
     }
 
-    if ( msKeyValStr->type == NULL ) {
-        fillStrInMsParam( msKeyValStr, NULL );
+    if (msKeyValStr->type == nullptr) {
+        fillStrInMsParam(msKeyValStr, nullptr);
     }
 
-    char *keyPtr = parseMspForStr( keyStr );
-    int keyLen;
-    if ( keyPtr == NULL || strcmp( keyPtr, MS_NULL_STR ) == 0 ) {
-        keyLen = 0;
-    }
-    else {
-        keyLen = strlen( keyPtr );
-    }
+    auto get_length{[](const char* str) -> std::size_t {
+        if (str == nullptr || std::strcmp(str, MS_NULL_STR) == 0) {
+            return 0;
+        }
+        return std::strlen(str);
+    }};
 
-    char *valPtr = parseMspForStr( valStr );
-    int valLen;
-    if ( valPtr == NULL || strcmp( valPtr, MS_NULL_STR ) == 0 ) {
-        valLen = 0;
-    }
-    else {
-        valLen = strlen( valPtr );
-    }
-    if ( valLen + keyLen <= 0 ) {
+    const char *keyPtr{parseMspForStr(keyStr)};
+    int keyLen{static_cast<int>(get_length(keyPtr))};
+
+    const char *valPtr{parseMspForStr(valStr)};
+    int valLen{static_cast<int>(get_length(valPtr))};
+
+    // Handle case where key and value are empty strings
+    if (valLen + keyLen <= 0) {
         return 0;
     }
 
-    char *oldKeyValPtr = parseMspForStr( msKeyValStr );
-    char *newKeyValPtr, *tmpPtr;
-    if ( oldKeyValPtr == NULL ) {
-        int newLen = valLen + keyLen + 10;
-        newKeyValPtr = ( char * )malloc( newLen );
-        *newKeyValPtr = '\0';
-        tmpPtr = newKeyValPtr;
+    char *oldKeyValPtr{parseMspForStr(msKeyValStr)};
+
+    const int old_len{oldKeyValPtr == nullptr ? 0 : static_cast<int>(std::strlen(oldKeyValPtr))};
+    const int new_len{old_len + valLen + keyLen + 10};
+
+    char *newKeyValPtr{static_cast<char*>(std::malloc(new_len * sizeof(char)))};
+    char *tmpPtr{newKeyValPtr};
+    if (old_len > 0) {
+        std::snprintf(newKeyValPtr, new_len, "%s%s", oldKeyValPtr, MS_INP_SEP_STR);
+
+        constexpr auto separator_length{std::char_traits<char>::length(MS_INP_SEP_STR)};
+        tmpPtr += old_len + separator_length;
     }
     else {
-        int oldLen = strlen( oldKeyValPtr );
-        int newLen = oldLen + valLen + keyLen + 10;
-        newKeyValPtr = ( char * )malloc( newLen );
-        snprintf( newKeyValPtr, newLen, "%s%s", oldKeyValPtr, MS_INP_SEP_STR );
-        tmpPtr = newKeyValPtr + oldLen + 4;
-        free( oldKeyValPtr );
+        *newKeyValPtr = '\0';
     }
+    std::free(oldKeyValPtr);
 
-    if ( keyLen > 0 ) {
-        snprintf( tmpPtr, keyLen + 2, "%s=", keyPtr );
-        tmpPtr += keyLen + 1;
+    constexpr auto null_term_length{1};
+
+    if (keyLen > 0) {
+        constexpr auto null_term_and_kv_separator_length{2};
+        std::snprintf(tmpPtr, keyLen + null_term_and_kv_separator_length, "%s=", keyPtr);
+        tmpPtr += keyLen + null_term_length;
     }
-    if ( valLen > 0 ) {
-        snprintf( tmpPtr, valLen + 2, "%s", valPtr );
+    if (valLen > 0) {
+        std::snprintf(tmpPtr, valLen + null_term_length, "%s", valPtr);
     }
-    msKeyValStr->inOutStruct = ( void * ) newKeyValPtr;
+    msKeyValStr->inOutStruct = newKeyValPtr;
 
     return 0;
 }
--- a/scripts/irods/test/test_all_rules.py
+++ b/scripts/irods/test/test_all_rules.py
@@ -664,6 +664,110 @@ OUTPUT ruleExecOut
         self.rods_session.assert_icommand("irule -F " + rule_file);
 
     @unittest.skipUnless(plugin_name == 'irods_rule_engine_plugin-irods_rule_language', 'only applicable for irods_rule_language REP')
+    def test_msiAddKeyValToMspStr_works_with_empty_string__issue_6918(self):
+        rule_file="test_msiAddKeyValToMspStr.r"
+        rule_string= '''
+msiTestAddKeyValToMspStr {{
+    *Flag = "";
+    msiAddKeyValToMspStr("keyOnly", "", *Flag);
+    writeLine("stdout", *Flag);
+
+    *Flag = "";
+    msiAddKeyValToMspStr("", "valueOnly", *Flag);
+    writeLine("stdout", *Flag);
+
+    *Flag = "";
+    msiAddKeyValToMspStr("key", "value", *Flag);
+    writeLine("stdout", *Flag);
+}}
+
+INPUT null
+OUTPUT ruleExecOut
+'''
+
+        with open(rule_file, 'w') as f:
+            f.write(rule_string)
+
+        rep_name = 'irods_rule_engine_plugin-irods_rule_language-instance'
+        self.admin.assert_icommand(['irule', '-r', rep_name, '-F', rule_file], 'STDOUT_MULTILINE', ['keyOnly=', 'valueOnly', 'key=value'])
+
+    @unittest.skipUnless(plugin_name == 'irods_rule_engine_plugin-irods_rule_language', 'only applicable for irods_rule_language REP')
+    def test_msiRmColl_removes_trash__issue_6918(self):
+        collection_to_free = os.path.join(self.admin.session_collection, 'freeMe')
+
+        # Generate trash
+        self.admin.assert_icommand(['itouch', collection_to_free])
+        self.admin.assert_icommand(['irm', collection_to_free])
+
+        rule_file="test_msiRmCollTrash.r"
+        rule_string= '''
+msiTestRmCollRemovesTrash {{
+    *Flag = "";
+    *Coll = "{trash_col}";
+    msiAddKeyValToMspStr("irodsAdminRmTrash", "", *Flag);
+    msiRmColl(*Coll, *Flag, *result);
+}}
+
+INPUT null
+OUTPUT ruleExecOut
+'''.format(trash_col=os.path.join(self.admin.session_collection_trash))
+
+        with open(rule_file, 'w') as f:
+            f.write(rule_string)
+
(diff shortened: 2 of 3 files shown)


lib/core/src/msParam.cpp
```diff
@@ -1451,60 +1451,61 @@ int
...
+    const char *valPtr{parseMspForStr(valStr)};
+    int valLen{static_cast<int>(get_length(valPtr))};
+
+    if (valLen + keyLen <= 0) {
         return 0;
     }
 
-    char *oldKeyValPtr = parseMspForStr( msKeyValStr );
+    char *oldKeyValPtr{parseMspForStr(msKeyValStr)};
     char *newKeyValPtr, *tmpPtr;
```
trel, line 1480: split these up?   initialize?

lib/core/src/msParam.cpp
```diff
@@ -1451,60 +1451,61 @@ int
...
-    if ( oldKeyValPtr == NULL ) {
-        int newLen = valLen + keyLen + 10;
-        newKeyValPtr = ( char * )malloc( newLen );
+    if (oldKeyValPtr == nullptr) {
+        const int newLen{valLen + keyLen + 10};
```
trel, line 1482: what is 10?

lib/core/src/msParam.cpp
```diff
@@ -1451,60 +1451,61 @@ int
...
-    }
-
-    if ( keyLen > 0 ) {
-        snprintf( tmpPtr, keyLen + 2, "%s=", keyPtr );
+        const int oldLen{static_cast<int>(std::strlen(oldKeyValPtr))};
+        const int newLen{oldLen + valLen + keyLen + 10};
+        newKeyValPtr = reinterpret_cast<char*>(std::malloc(newLen));
+	if (oldLen > 0) {
+	    std::snprintf(newKeyValPtr, newLen, "%s%s", oldKeyValPtr, MS_INP_SEP_STR);
+	    tmpPtr = newKeyValPtr + oldLen + 4;
```
trel, line 1493: what is 4?

lib/core/src/msParam.cpp
```diff
@@ -1451,60 +1451,61 @@ int
...
+	} else {
+  	    *newKeyValPtr = '\0';
+	    tmpPtr = newKeyValPtr;
+	}
+	std::free(oldKeyValPtr);
+    }
+
+    if (keyLen > 0) {
+        std::snprintf(tmpPtr, keyLen + 2, "%s=", keyPtr);
```
trel, line 1502: what is 2?

trel (commented):
so these tests failed before the code edit?  because of the ++++?

scripts/irods/test/test_all_rules.py
```diff
@@ -663,6 +663,82 @@ def test_msiServerMonPerf_default_3736(self):
 
         self.rods_session.assert_icommand("irule -F " + rule_file);
 
+    @unittest.skipUnless(plugin_name == 'irods_rule_engine_plugin-irods_rule_language', 'only applicable for irods_rule_language REP')
+    def test_msiAddKeyValToMspStr_works_with_empty_string_6918(self):
```
korydraughn, line 667: `_6918` -> `__issue_6918`

scripts/irods/test/test_all_rules.py
```diff
@@ -663,6 +663,82 @@ def test_msiServerMonPerf_default_3736(self):
...
+'''.format(collection=collection_to_free)
+
+        with open(rule_file, 'w') as f:
+            f.write(rule_string)
+
+        rep_name = 'irods_rule_engine_plugin-irods_rule_language-instance'
+        self.admin.assert_icommand(['irule', '-r', rep_name, '-F', rule_file])
+
+    @unittest.skipUnless(plugin_name == 'irods_rule_engine_plugin-irods_rule_language', 'only applicable for irods_rule_language REP')
+    def test_msiRmColl_removes_collection_via_flag_6918(self):
```
korydraughn, line 718: `_6918` -> `__issue_6918`

lib/core/src/msParam.cpp
```diff
@@ -1451,60 +1451,61 @@ int
...
-    }
-    else {
-        keyLen = strlen( keyPtr );
-    }
+    auto get_length{[](const char* str) -> size_t {
+        if (str == nullptr || std::strcmp(str, MS_NULL_STR) == 0) {
+	    return 0;
+	}
+	return std::strlen(str);
+    }};
```
korydraughn, line 1467: Formatting.

lib/core/src/msParam.cpp
```diff
@@ -1451,60 +1451,61 @@ int
...
 
-    char *keyPtr = parseMspForStr( keyStr );
-    int keyLen;
-    if ( keyPtr == NULL || strcmp( keyPtr, MS_NULL_STR ) == 0 ) {
-        keyLen = 0;
-    }
-    else {
-        keyLen = strlen( keyPtr );
-    }
+    auto get_length{[](const char* str) -> size_t {
```
korydraughn, line 1462: `size_t` -> `std::size_t`

lib/core/src/msParam.cpp
```diff
@@ -1451,60 +1451,61 @@ int
...
 
-    char *oldKeyValPtr = parseMspForStr( msKeyValStr );
+    char *oldKeyValPtr{parseMspForStr(msKeyValStr)};
     char *newKeyValPtr, *tmpPtr;
-    if ( oldKeyValPtr == NULL ) {
-        int newLen = valLen + keyLen + 10;
-        newKeyValPtr = ( char * )malloc( newLen );
+    if (oldKeyValPtr == nullptr) {
+        const int newLen{valLen + keyLen + 10};
+        newKeyValPtr = reinterpret_cast<char*>(std::malloc(newLen));
```
korydraughn, line 1483: `reinterpret_cast` -> `static_cast`

For correctness: `newLen` -> `sizeof(char) * newLen`

lib/core/src/msParam.cpp
```diff
@@ -1451,60 +1451,61 @@ int
...
-        newKeyValPtr = ( char * )malloc( newLen );
-        snprintf( newKeyValPtr, newLen, "%s%s", oldKeyValPtr, MS_INP_SEP_STR );
-        tmpPtr = newKeyValPtr + oldLen + 4;
-        free( oldKeyValPtr );
-    }
-
-    if ( keyLen > 0 ) {
-        snprintf( tmpPtr, keyLen + 2, "%s=", keyPtr );
+        const int oldLen{static_cast<int>(std::strlen(oldKeyValPtr))};
+        const int newLen{oldLen + valLen + keyLen + 10};
```
korydraughn, line 1489: What is 10?

lib/core/src/msParam.cpp
```diff
@@ -1451,60 +1451,61 @@ int
...
+        newKeyValPtr = reinterpret_cast<char*>(std::malloc(newLen));
```
korydraughn, line 1490: - static_cast
- sizeof * newLen

lib/core/src/msParam.cpp
```diff
@@ -1451,60 +1451,61 @@ int
...
+        std::snprintf(tmpPtr, keyLen + 2, "%s=", keyPtr);
         tmpPtr += keyLen + 1;
     }
-    if ( valLen > 0 ) {
-        snprintf( tmpPtr, valLen + 2, "%s", valPtr );
+    if (valLen > 0) {
+        std::snprintf(tmpPtr, valLen + 2, "%s", valPtr);
     }
-    msKeyValStr->inOutStruct = ( void * ) newKeyValPtr;
+    msKeyValStr->inOutStruct = reinterpret_cast<void *>(newKeyValPtr);
```
korydraughn, line 1508: You can remove the cast here.

C++ allows pointers to be assigned to void pointers implicitly.

scripts/irods/test/test_all_rules.py
```diff
@@ -663,6 +663,82 @@ def test_msiServerMonPerf_default_3736(self):
...
+        self.admin.assert_icommand(['itouch', collection_to_free])
+        self.admin.assert_icommand(['irm', collection_to_free])
+
+        rule_file="test_msiAddKeyValToMspStr.r"
+        rule_string= '''
+msiTestAddKeyValToMspStr {{
+    *Flag = "";
+    *Coll = "{trash_col}";
+    msiAddKeyValToMspStr("irodsAdminRmTrash", "", *Flag);
+    msiRmColl(*Coll, *Flag, *result);
```
korydraughn, line 680: How does L680 help prove `msiAddKeyValToMspStr` handles empty strings correctly?

scripts/irods/test/test_all_rules.py
```diff
@@ -663,6 +663,82 @@ def test_msiServerMonPerf_default_3736(self):
...
+'''.format(trash_col=os.path.join(self.admin.session_collection_trash))
+
+        with open(rule_file, 'w') as f:
+            f.write(rule_string)
+
+        rep_name = 'irods_rule_engine_plugin-irods_rule_language-instance'
+        self.admin.assert_icommand(['irule', '-r', rep_name, '-F', rule_file])
+
+    @unittest.skipUnless(plugin_name == 'irods_rule_engine_plugin-irods_rule_language', 'only applicable for irods_rule_language REP')
+    def test_msiRmColl_removes_collection_6918(self):
```
korydraughn, line 694: `_6918` -> `__issue_6918`

lib/core/src/msParam.cpp
```diff
@@ -1451,60 +1451,61 @@ int
...
-        valLen = strlen( valPtr );
-    }
-    if ( valLen + keyLen <= 0 ) {
+    const char *keyPtr{parseMspForStr(keyStr)};
+    int keyLen{static_cast<int>(get_length(keyPtr))};
+
+    const char *valPtr{parseMspForStr(valStr)};
+    int valLen{static_cast<int>(get_length(valPtr))};
+
+    if (valLen + keyLen <= 0) {
```
korydraughn, line 1475: Consider adding some words explaining the purpose of this.

lib/core/src/msParam.cpp
```diff
@@ -1451,60 +1451,61 @@ int
 addKeyValToMspStr( msParam_t * keyStr, msParam_t * valStr,
                    msParam_t * msKeyValStr ) {
 
-    if ( ( keyStr == NULL && valStr == NULL ) || msKeyValStr == NULL ) {
+    if ((keyStr == nullptr && valStr == nullptr) || msKeyValStr == nullptr) {
```
korydraughn, line 1454: What happens if `valStr` is valid, but `keyStr` is null?
Do we protect against that case later in the function?

scripts/irods/test/test_all_rules.py
```diff
@@ -663,6 +663,82 @@ def test_msiServerMonPerf_default_3736(self):
...
+        collection_to_free = os.path.join(self.admin.session_collection, 'freeMe')
+
+        self.admin.assert_icommand(['imkdir', collection_to_free])
+
+        rule_file="test_msiRmCollSimple.r"
+        rule_string= '''
+msiTestAddKeyValToMspStr {{
```
alanking, line 701: Let's update the name of the rule to reflect the test

alanking (commented):
I'll throw my twig onto the pile :)

lib/core/src/msParam.cpp
MartinFlores751: The `char *newKeyValPtr, *tmpPtr;` right? Can do!

lib/core/src/msParam.cpp
MartinFlores751: Not sure myself, I'll try looking to see if there's something that might give a hint.

lib/core/src/msParam.cpp
MartinFlores751: I don't see it, and I don't think 4.2 has clang format, so I can't cheat and use that. Are you saying the `if` should be split at the `||` or is it something else?

korydraughn: Perhaps your editor is hiding the issue.

Try looking at those lines in your browser.

korydraughn: The indentation is incorrect.

lib/core/src/msParam.cpp
trel: or just use size_of (or whatever the correct equivalent is)...

lib/core/src/msParam.cpp
trel: i see tabs, and we're not doing tabs in irods/irods i believe...

korydraughn: Correct.

scripts/irods/test/test_all_rules.py
MartinFlores751: Should be done!

scripts/irods/test/test_all_rules.py
MartinFlores751: Added test to check `msiAddKeyValToMspStr`, kept this one to test `msiRmColl`.

lib/core/src/msParam.cpp
MartinFlores751: Done! Didn't know about that implicit cast.

lib/core/src/msParam.cpp
MartinFlores751: Should be good, wouldn't be surprised if more formatting errors showed up though. Haven't properly set up my new editor yet.

lib/core/src/msParam.cpp
MartinFlores751: Name given! If too wordy, I'll change to something else.

lib/core/src/msParam.cpp
MartinFlores751: I believe both `parseMspForStr()` and the lambda `get_length()` handle the `nullptr` case. Whether we want a value but no key is a different discussion: https://github.com/irods/irods/pull/7049#discussion_r1171576383.

scripts/irods/test/test_all_rules.py
```diff
@@ -664,16 +664,44 @@ def test_msiServerMonPerf_default_3736(self):
...
+    def test_msiAddKeyValToMspStr_works_with_empty_string__issue_6918(self):
+        rule_file="test_msiAddKeyValToMspStr.r"
+        rule_string= '''
+msiTestAddKeyValToMspStr {{
+    *Flag = "";
+    msiAddKeyValToMspStr("keyOnly", "", *Flag);
+    writeLine("stdout", *Flag);
+
+    *Flag = "";
+    msiAddKeyValToMspStr("", "valueOnly", *Flag);
```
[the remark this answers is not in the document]
trel: pretty sure we don't want that to ever happen.  someone should get a big fat error if they try.

can't really think of a legitimate use case for this scenario...

consider returning an error - and see if the test suite falls over somewhere.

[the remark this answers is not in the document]
MartinFlores751: Should be removed, currently returns `USER_INPUT_STRING_ERR`. Let me know if you know something that works better for this case!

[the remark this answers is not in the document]
alanking: My only suggestion would be to try using the value-only MSI invocation in a test and ensure that the expected error comes back. I can't really find an error code that conveys things any better, so, seems fine to me.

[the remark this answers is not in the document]
alanking: After some discussion, we've decided that we should just leave the value-only behavior for now. Needs more investigation. Please resolve this once https://github.com/irods/irods/pull/7049/commits/f53942b90b95a1ea7df59f47fc01aad975894190 is reverted.

lib/core/src/msParam.cpp
MartinFlores751: Note: We're going to open a separate issue to investigate this magical number.

alanking (approved):

Issue #7050, opened 2023-04-18
SQL statement leak: too many concurrent statements


dworkin:
- [x] main
- [x] 4-3-stable
- [x] 4-2-stable
---
Missing call to cmlFreeStatement. I made a pull request for 4-2-stable, but the other branches are also affected.

dworkin:
We ran into the 'too many concurrent statements' issue, without obvious cause. So I'm auditing the relevant code.

dworkin:
That's all (for this audit).

dworkin:
Well, that was embarrassing.  I had assumed that `freeGenQueryOut` worked like `cmlFreeStatement` at the lower level, and it doesn't.  All of those "statement leaks" are mere memory leaks, and I have withdrawn my related pull requests.

However, there *is* a statement leak in `checkPermissionByObjType`. It can be triggered as follows:
- generate 260 new groups with `iadmin mkgroup`
- put the `rods` user in each group with `iadmin atg`
- create a dataObj `foo` which the `rods` user does not have access to
- execute the following code in an endless loop:
    `*access=0; msiCheckAccess("foo", "read object", *access);`

The cause is that `checkPermissionByObjType` does not take into account that there could be more than one page of SQL rows.

korydraughn:
We'll reproduce and get a fix in. Thanks for finding this.

dworkin:
The code below is similarly afflicted:

https://github.com/irods/irods/blob/5b4b6530ab86c9cb7fc71990b50d3ffc53458570/server/re/src/reIn2p3SysRule.cpp#L494

alanking:
I was able to reproduce the issue by writing a test as described by @dworkin in https://github.com/irods/irods/issues/7050#issuecomment-1516615931.

Here is the rule that I wrote:
```python
check_access_a_lot {
    *big_number = 51;
    *incr = 1;
    for (*i = 0; *i < *big_number; *i = *i + *incr) {
        *access = 0;
        msiCheckAccess("*logical_path", "read object", *access);
    }
}
INPUT *logical_path="/tempZone/home/rods/foo"
OUTPUT ruleExecOut
```
It turns out that you only need 51 iterations of the loop to get this result:
```bash
$ irule -r irods_rule_engine_plugin-irods_rule_language-instance -F rule.r 
remote addresses: 127.0.0.1 ERROR: rcExecMyRule error.  status = -1105000 INVALID_OBJECT_TYPE
[...]

PR #7051, opened 2023-04-18
[#7050] Fix statement leak (4-2-stable)

dworkin:


trel:
excellent - is there a scenario (or even better, a test) that can trap/exhibit this leak?

trel:
okay - thanks for the context.

alanking (commented):
Yes, agreed. The same statement free is made just a few lines below for all of the "success" codepaths, so the failure codepath ought to do so as well.

@korydraughn / @trel - Shall I run this through the test gauntlet and call that sufficient, or should we think a little harder about how to add a test for this?

trel:
definitely gauntlet - and bonus points for big thoughts or new tests.   but i don't know that they're necessary.

'by inspection' is a legitimate argument, just not as great as a test.

alanking:
I tested this alongside #7062 and all the tests passed.

PR #7053, opened 2023-04-19
[#7050] Fix more statement leaks (4-2-stable)

dworkin:
Same problems also exist in the other branches.


dworkin:
The second one is significant.  It is triggered with msiCheckAccess, on a data object the user has no access to, by a user with no group access.

trel:
your comment is very helpful - we will need to get a test included to assert the good behavior.

PR #7066, opened 2023-04-26
[#7050] Fix statement leak in checkPermissionByObjType (4-2-stable)

dworkin:
A fix for the statement leak in `checkPermissionByObjType`. This also solves the memory leak which I erroneously reported as a statement leak, before.

### DIFF
--- a/scripts/irods/test/test_all_rules.py
+++ b/scripts/irods/test/test_all_rules.py
@@ -1587,6 +1587,58 @@ OUTPUT ruleExecOut
                     lib.remove_resource(resource, self.admin)
 
 
+    @unittest.skipUnless(plugin_name == 'irods_rule_engine_plugin-irods_rule_language', 'Only implemented for NREP.')
+    def test_adding_user_to_more_than_SQL_MAX_ROWS_groups_and_try_msiCheckAccess__issue_7050(self):
+        rule_map = {
+            'irods_rule_engine_plugin-irods_rule_language': textwrap.dedent('''
+                *access = 0;
+                *iterations = 51;
+                for (*i = 0; *i < *iterations; *i = *i + 1) {{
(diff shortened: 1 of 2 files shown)


alanking (commented):
I think we are in agreement that this change is fine. I have also confirmed that it passes the core test suite. I am going to write a test as described by this comment to add to this before we merge it: https://github.com/irods/irods/issues/7050#issuecomment-1516615931

alanking (approved):
Added test and watched it fail (appropriately) without the changes in this PR, then added the changes in this PR, and it passed. Will await one more approval

scripts/irods/test/test_all_rules.py
```diff
@@ -1587,6 +1587,58 @@ def test_msiDataObjChksum_with_admin_keyword__issue_6118(self):
...
+            # error. Otherwise, this will have no output.
+            groupie_user.assert_icommand(['irule', '-r', rep_instance, rule_text, 'null', 'ruleExecOut'])
+
+        finally:
+            self.admin.assert_icommand(['ils', '-AL', os.path.dirname(logical_path)], 'STDOUT') # debugging
+
+            for i in range(group_count):
+                group_name = 'group_' + str(i + 1)
+                self.admin.assert_icommand(['iadmin', 'rfg', group_name, groupie_user.username])
+                self.admin.assert_icommand(['iadmin', 'rmgroup', group_name])
```
korydraughn, line 1639: Should these calls use `run_icommand` instead?

alanking: For this case, I don't think it necessarily matters, but I will change them.

alanking: Done

korydraughn (approved):
Squash, pound, and merge when ready.

PR #7082, opened 2023-04-27
Fix statement leaks (main)

alanking:
Addresses #7050 
Cherry-picked from #7051 and #7066 

Unit tests passed. Core tests are running

Let me know if I should address the clang-format issues and if those should be ported back to 4-2-stable. There are no differences between this and the 4-2-stable changes.

### DIFF
--- a/server/core/src/objMetaOpr.cpp
+++ b/server/core/src/objMetaOpr.cpp
@@ -654,8 +654,7 @@ checkPermissionByObjType( rsComm_t *rsComm, char *objName, char *objType, char *
         int status = rsGenQuery( rsComm, &genQueryInp, &genQueryOut );
 
         // note:  if rsGenQuery has an error, just continue to below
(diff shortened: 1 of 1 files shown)


korydraughn:
Yes, please run clang-format on the changes.

alanking:
Core tests passed except for the new test, which required the change in the latest commit ("read object" -> "read_object"). After that change, the test passes. So, with the latest commit, core tests all pass.

trel:
spectacular.

korydraughn (commented):
Please squash to taste and leave off pounds.

alanking:
Squashed. I'm leaving the clang-format commit separate in case we want to backport to 4-2-stable. Else, I'll just squash that in too

trel:
let's preserve the original commit as much as possible, then a separate commit to clang-format.

korydraughn (approved):
Sounds good. Please pound and merge when ready.

Issue #5411, opened 2021-02-15
buffer too short when waking iinit


mcast:
- [x] main
- [x] 4-3-stable
- [x] 4-2-stable
---
### iRODS Version, OS and Version
iRODS 4.2.7 on Ubuntu Bionic 18.04.5 LTS

### What did you try to do?
Run self-tests on the next layer up, with `$HOME` set to something accidentally longer than usual. Then `iinit` inside this faked homedir.

### Expected behavior
Tests usually pass just fine.

### Observed behavior (including steps to reproduce, if applicable)
Translated into a set of commands you should be able to repeat,

```
mca@casm3-head2:~$ mkdir -p /tmp/mca.name-long-enough-to-cause-problems-with-a-buffer-somewhere-oops/.irods
mca@casm3-head2:~$ cp -a ~/.irods/irods_environment-cgpro.json /tmp/mca.name-long-enough-to-cause-problems-with-a-buffer-somewhere-oops/.irods/
mca@casm3-head2:~$ sed -i -e "s,$HOME,/tmp/mca.name-long-enough-to-cause-problems-with-a-buffer-somewhere-oops," /tmp/mca.name-long-enough-to-cause-problems-with-a-buffer-somewhere-oops/.irods/irods_environment-cgpro.json
mca@casm3-head2:~$ find /tmp/mca.name* -ls
     1277      0 drwxr-x---   3 mca      cancer         60 Feb 15 13:07 /tmp/mca.name-long-enough-to-cause-problems-with-a-buffer-somewhere-oops
     1278      0 drwxr-x---   2 mca      cancer         60 Feb 15 13:07 /tmp/mca.name-long-enough-to-cause-problems-with-a-buffer-somewhere-oops/.irods
     1280      4 -rw-r-----   1 mca      cancer        937 Feb 15 13:07 /tmp/mca.name-long-enough-to-cause-problems-with-a-buffer-somewhere-oops/.irods/irods_environment-cgpro.json
mca@casm3-head2:~$ HOME=/tmp/mca.name-long-enough-to-cause-problems-with-a-buffer-somewhere-oops IRODS_ENVIRONMENT_FILE=/tmp/mca.name-long-enough-to-cause-problems-with-a-buffer-somewhere-oops/.irods/irods_environment-cgpro.json iinit
 ERROR: rstrcpy not enough space in dest, slen:72, maxLen:64, stacktrace:

Dumping stack trace
[...]

PR #7092, opened 2023-05-04
[#5411] Skip iinit test when SSL is enabled (4-2-stable)

alanking:
Addresses #5411 

Found while running tests in topology on consumer with SSL enabled


korydraughn (approved):
Pound and merge please.

PR #7095, opened 2023-05-04
[#2279] Skip fast updates test when SSL is enabled (4-2-stable)

alanking:
Addresses #2279

This change will only exist in 4-2-stable.


korydraughn (approved):
Ship it.

PR #7129, opened 2023-06-22
[7125] Update modification time for tickets (main)

Nishant2022:
Update the modification time for tickets in the catalog

### DIFF
--- a/plugins/database/src/db_plugin.cpp
+++ b/plugins/database/src/db_plugin.cpp
@@ -13796,15 +13796,16 @@ irods::error db_get_hierarchy_for_resc_op(
 namespace
 {
     // A support function for executing SQL operations.
-    irods::error execute_sql(const std::string_view _sql)
+    irods::error execute_sql(const char* _sql, const char* _function_name)
     {
-        auto ec = cmlExecuteNoAnswerSql(_sql.data(), &icss);
+        auto ec = cmlExecuteNoAnswerSql(_sql, &icss);
         if (CAT_SUCCESS_BUT_WITH_NO_INFO == ec) {
             return CODE(ec);
         }
 
         if (0 != ec) {
             log_db::error("SQL execution error [{}].", ec);
+            _rollback(_function_name);
             return ERROR(ec, "SQL execution error.");
         }
 
@@ -13816,6 +13817,7 @@ namespace
         return SUCCESS();
     } // execute_sql
 
+    // NOLINTNEXTLINE(readability-function-cognitive-complexity)
     irods::error execute_ticket_operation_as_admin(irods::plugin_context& _ctx,
                                                    const char* _op_name,
                                                    const char* _ticket_string,
@@ -13913,27 +13915,45 @@ namespace
         // Modify ticket operation.
         if (std::strcmp(_op_name, "mod") == 0) {
             if (std::strcmp(_arg3, "uses") == 0) {
-                cllBindVars[0] = _arg4;
-                cllBindVars[1] = ticket_id_string.c_str();
-                cllBindVarCount = 2;
+                int i{0};
+                char myTime[TIME_LEN]{}; // NOLINT(cppcoreguidelines-avoid-c-arrays, modernize-avoid-c-arrays)
+                getNowStr(myTime);
+                cllBindVars[i++] = _arg4; // NOLINT(cppcoreguidelines-pro-bounds-constant-array-index)
+                cllBindVars[i++] = myTime; // NOLINT(cppcoreguidelines-pro-bounds-constant-array-index)
+                // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-constant-array-index)
+                cllBindVars[i++] = ticket_id_string.c_str();
+                cllBindVarCount = i;
 
-                return execute_sql("update R_TICKET_MAIN set uses_limit = ? where ticket_id = ?");
+                return execute_sql(
+                    "update R_TICKET_MAIN set uses_limit = ?, modify_ts = ? where ticket_id = ?", __func__);
             } // uses
 
             if (std::strcmp(_arg3, "write-file") == 0) {
-                cllBindVars[0] = _arg4;
-                cllBindVars[1] = ticket_id_string.c_str();
-                cllBindVarCount = 2;
+                int i{};
+                char myTime[TIME_LEN]{}; // NOLINT(cppcoreguidelines-avoid-c-arrays, modernize-avoid-c-arrays)
+                getNowStr(myTime);
+                cllBindVars[i++] = _arg4; // NOLINT(cppcoreguidelines-pro-bounds-constant-array-index)
+                cllBindVars[i++] = myTime; // NOLINT(cppcoreguidelines-pro-bounds-constant-array-index)
+                // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-constant-array-index)
+                cllBindVars[i++] = ticket_id_string.c_str();
+                cllBindVarCount = i;
 
-                return execute_sql("update R_TICKET_MAIN set write_file_limit = ? where ticket_id = ?");
+                return execute_sql(
+                    "update R_TICKET_MAIN set write_file_limit = ?, modify_ts = ? where ticket_id = ?", __func__);
             } // write-file
 
             if (std::strcmp(_arg3, "write-bytes") == 0) {
-                cllBindVars[0] = _arg4;
-                cllBindVars[1] = ticket_id_string.c_str();
-                cllBindVarCount = 2;
+                int i{};
+                char myTime[TIME_LEN]{}; // NOLINT(cppcoreguidelines-avoid-c-arrays, modernize-avoid-c-arrays)
+                getNowStr(myTime);
+                cllBindVars[i++] = _arg4; // NOLINT(cppcoreguidelines-pro-bounds-constant-array-index)
+                cllBindVars[i++] = myTime; // NOLINT(cppcoreguidelines-pro-bounds-constant-array-index)
+                // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-constant-array-index)
+                cllBindVars[i++] = ticket_id_string.c_str();
+                cllBindVarCount = i;
 
-                return execute_sql("update R_TICKET_MAIN set write_byte_limit = ? where ticket_id = ?");
+                return execute_sql(
+                    "update R_TICKET_MAIN set write_byte_limit = ?, modify_ts = ? where ticket_id = ?", __func__);
             } // write-bytes
 
             if (std::strcmp(_arg3, "expire") == 0) {
@@ -13977,11 +13997,18 @@ namespace
                     }
                 }
 
-                cllBindVars[0] = ticket_expiration_string.c_str();
-                cllBindVars[1] = ticket_id_string.c_str();
-                cllBindVarCount = 2;
+                int i{};
+                char myTime[TIME_LEN]{}; // NOLINT(cppcoreguidelines-avoid-c-arrays, modernize-avoid-c-arrays)
+                getNowStr(myTime);
+                // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-constant-array-index)
+                cllBindVars[i++] = ticket_expiration_string.c_str();
+                cllBindVars[i++] = myTime; // NOLINT(cppcoreguidelines-pro-bounds-constant-array-index)
+                // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-constant-array-index)
+                cllBindVars[i++] = ticket_id_string.c_str();
+                cllBindVarCount = i;
 
-                return execute_sql("update R_TICKET_MAIN set ticket_expiry_ts = ? where ticket_id = ?");
+                return execute_sql(
+                    "update R_TICKET_MAIN set ticket_expiry_ts = ?, modify_ts = ? where ticket_id = ?", __func__);
             } // expire
 
             if (std::strcmp(_arg3, "add") == 0) {
@@ -13996,7 +14023,42 @@ namespace
                     cllBindVars[1] = hostIp;
                     cllBindVarCount = 2;
 
-                    return execute_sql("insert into R_TICKET_ALLOWED_HOSTS (ticket_id, host) values (? , ?)");
+                    ec = cmlExecuteNoAnswerSql(
+                        "insert into R_TICKET_ALLOWED_HOSTS (ticket_id, host) values (?, ?)", &icss);
+                    if (ec != 0) {
+                        auto err_msg = fmt::format(
+                            "Failed to add host [{}] to ticket [{}], status [{}]", hostIp, ticket_id_string, ec);
+                        log_sql::error(err_msg);
+                        _rollback(__func__);
+                        return ERROR(ec, std::move(err_msg));
+                    }
+
+                    int i{};
+                    char myTime[TIME_LEN]{}; // NOLINT(cppcoreguidelines-avoid-c-arrays, modernize-avoid-c-arrays)
+                    getNowStr(myTime);
+                    cllBindVars[i++] = myTime; // NOLINT(cppcoreguidelines-pro-bounds-constant-array-index)
+                    // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-constant-array-index)
+                    cllBindVars[i++] = ticket_id_string.c_str();
+                    cllBindVarCount = i;
+
+                    ec = cmlExecuteNoAnswerSql("update R_TICKET_MAIN set modify_ts = ? where ticket_id = ?", &icss);
+                    if (ec != 0) {
+                        auto err_msg = fmt::format(
+                            "Failed to update modify_ts for ticket [{}] while adding host [{}], status [{}]",
+                            ticket_id_string,
+                            hostIp,
+                            ec);
+                        log_sql::error(err_msg);
+                        _rollback(__func__);
+                        return ERROR(ec, std::move(err_msg));
+                    }
+
+                    ec = cmlExecuteNoAnswerSql("commit", &icss);
+                    if (ec < 0) {
+                        return ERROR(ec, "Failed to commit SQL updates.");
+                    }
+
+                    return SUCCESS();
                 } // host
 
                 if (std::strcmp(_arg4, "user") == 0) {
@@ -14011,7 +14073,42 @@ namespace
                     cllBindVars[1] = _arg5;
                     cllBindVarCount = 2;
 
-                    return execute_sql("insert into R_TICKET_ALLOWED_USERS (ticket_id, user_name) values (? , ?)");
+                    ec = cmlExecuteNoAnswerSql(
+                        "insert into R_TICKET_ALLOWED_USERS (ticket_id, user_name) values (?, ?)", &icss);
+                    if (ec != 0) {
+                        auto err_msg = fmt::format(
+                            "Failed to add user [{}] to ticket [{}], status [{}]", _arg5, ticket_id_string, ec);
+                        log_sql::error(err_msg);
+                        _rollback(__func__);
+                        return ERROR(ec, std::move(err_msg));
+                    }
+
+                    int i{};
+                    char myTime[TIME_LEN]{}; // NOLINT(cppcoreguidelines-avoid-c-arrays, modernize-avoid-c-arrays)
+                    getNowStr(myTime);
+                    cllBindVars[i++] = myTime; // NOLINT(cppcoreguidelines-pro-bounds-constant-array-index)
+                    // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-constant-array-index)
+                    cllBindVars[i++] = ticket_id_string.c_str();
+                    cllBindVarCount = i;
+
+                    ec = cmlExecuteNoAnswerSql("update R_TICKET_MAIN set modify_ts = ? where ticket_id = ?", &icss);
+                    if (ec != 0) {
+                        auto err_msg = fmt::format(
+                            "Failed to update modify_ts for ticket [{}] while adding user [{}], status [{}]",
+                            ticket_id_string,
+                            _arg5,
+                            ec);
+                        log_sql::error(err_msg);
+                        _rollback(__func__);
+                        return ERROR(ec, std::move(err_msg));
+                    }
+
+                    ec = cmlExecuteNoAnswerSql("commit", &icss);
+                    if (ec < 0) {
+                        return ERROR(ec, "Failed to commit SQL updates.");
+                    }
+
+                    return SUCCESS();
                 } // user
 
                 if (std::strcmp(_arg4, "group") == 0) {
@@ -14026,7 +14123,42 @@ namespace
                     cllBindVars[1] = _arg5;
                     cllBindVarCount = 2;
 
-                    return execute_sql("insert into R_TICKET_ALLOWED_GROUPS (ticket_id, group_name) values (? , ?)");
+                    ec = cmlExecuteNoAnswerSql(
+                        "insert into R_TICKET_ALLOWED_GROUPS (ticket_id, group_name) values (?, ?)", &icss);
+                    if (ec != 0) {
+                        auto err_msg = fmt::format(
+                            "Failed to add group [{}] to ticket [{}], status [{}]", _arg5, ticket_id_string, ec);
+                        log_sql::error(err_msg);
+                        _rollback(__func__);
+                        return ERROR(ec, std::move(err_msg));
+                    }
+
+                    int i{};
+                    char myTime[TIME_LEN]{}; // NOLINT(cppcoreguidelines-avoid-c-arrays, modernize-avoid-c-arrays)
+                    getNowStr(myTime);
+                    cllBindVars[i++] = myTime; // NOLINT(cppcoreguidelines-pro-bounds-constant-array-index)
+                    // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-constant-array-index)
+                    cllBindVars[i++] = ticket_id_string.c_str();
+                    cllBindVarCount = i;
+
+                    ec = cmlExecuteNoAnswerSql("update R_TICKET_MAIN set modify_ts = ? where ticket_id = ?", &icss);
+                    if (ec != 0) {
+                        auto err_msg = fmt::format(
+                            "Failed to update modify_ts for ticket [{}] while adding group [{}], status [{}]",
+                            ticket_id_string,
+                            _arg5,
+                            ec);
+                        log_sql::error(err_msg);
+                        _rollback(__func__);
+                        return ERROR(ec, std::move(err_msg));
+                    }
+
+                    ec = cmlExecuteNoAnswerSql("commit", &icss);
+                    if (ec < 0) {
+                        return ERROR(ec, "Failed to commit SQL updates.");
+                    }
+
+                    return SUCCESS();
                 } // group
             } // add
             else if (std::strcmp(_arg3, "remove") == 0) {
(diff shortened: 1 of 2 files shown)


plugins/database/src/db_plugin.cpp
```diff
@@ -14499,7 +14507,20 @@ irods::error db_mod_ticket_op(
                     log_sql::debug("chlModTicket SQL 11");
                 }
 
-                return execute_sql("insert into R_TICKET_ALLOWED_HOSTS (ticket_id, host) values (? , ?)");
+                irods::error ret = execute_sql("insert into R_TICKET_ALLOWED_HOSTS (ticket_id, host) values (? , ?)");
+                if (!ret.status()) {
+                    log_sql::error("TODO"); // TODO
```
trel, line 14512: TODO

plugins/database/src/db_plugin.cpp
```diff
@@ -14517,7 +14538,19 @@ irods::error db_mod_ticket_op(
                     log_sql::debug("chlModTicket SQL 12");
                 }
 
-                return execute_sql("insert into R_TICKET_ALLOWED_USERS (ticket_id, user_name) values (? , ?)");
+                irods::error ret = execute_sql("insert into R_TICKET_ALLOWED_USERS (ticket_id, user_name) values (? , ?)");
+                if (!ret.status()) {
+                    log_sql::error("TODO"); // TODO
```
trel, line 14543: TODO

plugins/database/src/db_plugin.cpp
```diff
@@ -14535,7 +14568,19 @@ irods::error db_mod_ticket_op(
                     log_sql::debug("chlModTicket SQL 13");
                 }
 
-                return execute_sql("insert into R_TICKET_ALLOWED_GROUPS (ticket_id, group_name) values (? , ?)");
+                irods::error ret = execute_sql("insert into R_TICKET_ALLOWED_GROUPS (ticket_id, group_name) values (? , ?)");
+                if (!ret.status()) {
+                    log_sql::error("TODO"); // TODO
```
trel, line 14573: TODO

plugins/database/src/db_plugin.cpp
```diff
@@ -14554,7 +14599,19 @@ irods::error db_mod_ticket_op(
                     log_sql::debug("chlModTicket SQL 14");
                 }
 
-                return execute_sql("delete from R_TICKET_ALLOWED_HOSTS where ticket_id=? and host=?");
+                irods::error ret = execute_sql("delete from R_TICKET_ALLOWED_HOSTS where ticket_id=? and host=?");
+                if (!ret.status()) {
+                    log_sql::error("TODO"); // TODO
```
trel, line 14604: TODO

plugins/database/src/db_plugin.cpp
```diff
@@ -14572,7 +14629,19 @@ irods::error db_mod_ticket_op(
                     log_sql::debug("chlModTicket SQL 15");
                 }
 
-                return execute_sql("delete from R_TICKET_ALLOWED_USERS where ticket_id=? and user_name=?");
+                irods::error ret = execute_sql("delete from R_TICKET_ALLOWED_USERS where ticket_id=? and user_name=?");
+                if (!ret.status()) {
+                    log_sql::error("TODO"); // TODO
```
trel, line 14634: TODO

plugins/database/src/db_plugin.cpp
```diff
@@ -14590,7 +14659,19 @@ irods::error db_mod_ticket_op(
                     log_sql::debug("chlModTicket SQL 16");
                 }
 
-                return execute_sql("delete from R_TICKET_ALLOWED_GROUPS where ticket_id=? and group_name=?");
+                irods::error ret = execute_sql("delete from R_TICKET_ALLOWED_GROUPS where ticket_id=? and group_name=?");
+                if (!ret.status()) {
+                    log_sql::error("TODO"); // TODO
```
trel, line 14664: TODO

trel (commented):
are these second execute_sql commands happening in the same db transaction?  honest question, not sure.

Nishant2022:
At the moment, whenever someone add or removes a host, user, or group from a ticket, there is only an insertion or removal of a row from the database. This means that the modify time stamp is never updated.

What I implemented would first try to insert or remove the row from the allowed tables, and if successful, would update the modify time stamp in the ticket main table.

Maybe that clears things up?

trel:
Yes - understood what's happening and why - and it is good.

I'm asking whether this is done 'atomically' from the perspective of the user and/or other concurrent agents who are looking at the database.    If these two updates are done in the same transaction, then that's ideal... but I'm not sure if that's happening in this file (might be too 'high' in the stack).

plugins/database/src/db_plugin.cpp
```diff
@@ -14499,7 +14507,20 @@ irods::error db_mod_ticket_op(
...
+                }
+
+                getNowStr(myTime);
+                i = 0;
+                cllBindVars[i++] = myTime;
+                cllBindVars[i++] = ticketIdStr;
+                cllBindVarCount = i;
+
+                return execute_sql("update R_TICKET_MAIN set modify_ts=? where ticket_id = ?");
+
```
korydraughn, line 14523: Remove empty line.

plugins/database/src/db_plugin.cpp
```diff
@@ -14499,7 +14507,20 @@ irods::error db_mod_ticket_op(
                     log_sql::debug("chlModTicket SQL 11");
                 }
 
-                return execute_sql("insert into R_TICKET_ALLOWED_HOSTS (ticket_id, host) values (? , ?)");
+                irods::error ret = execute_sql("insert into R_TICKET_ALLOWED_HOSTS (ticket_id, host) values (? , ?)");
+                if (!ret.status()) {
```
korydraughn, line 14511: I believe we normally use `if (!ret.ok())`, why `.status()`?

_This applies to the other locations too._

plugins/database/src/db_plugin.cpp
korydraughn, line 14512: You can get the iRODS error code by calling `ret.code()`.

Please include it in the log message.

_This applies to other locations too._

korydraughn (commented):
Looking good so far.

There's a function in this file that handles the **run-as-admin** case. I believe it is invoked earlier in this function if the conditions are met.

You'll need to check the implementation of that function to make sure the mtime is updated.

korydraughn:
> Yes - understood what's happening and why - and it is good.
> 
> I'm asking whether this is done 'atomically' from the perspective of the user and/or other concurrent agents who are looking at the database. If these two updates are done in the same transaction, then that's ideal... but I'm not sure if that's happening in this file (might be too 'high' in the stack).

Not all of these operations happen in a single transaction. See the following.
https://github.com/irods/irods/blob/1205fa3afd8de1e96e0e412184c7b30ae0b07c31/plugins/database/src/db_plugin.cpp#L13798-L13817

The operations where the mtime is updated separated have to do with adding/removing host, user, and/or group ticket constraints.

trel:
Okay, so we can't be guaranteed they'll be atomically handled.    They'll be real close, but could drift a smidge.

Not a blocker, this is more correct than before, definitely.

korydraughn:
Yes, worst case scenario is the mtime updates are interleaved in a way where the latest mtime is overwritten by an earlier mtime value.

Regardless, the mtime will be updated to reflect a later time than before.

plugins/database/src/db_plugin.cpp
Nishant2022: I had noticed in `irods_error.cpp:257` that there was a comment saying `irods::error::ok()` was deprecated in 4.0.3. If this is not the case I will change my code to use `irods::error::ok()` as I believe it is clearer.

korydraughn: I was not aware of that. We still use it afaik.

Let's stick with `.ok()` for now.

plugins/database/src/db_plugin.cpp
```diff
@@ -13912,28 +13912,40 @@ namespace
 
         // Modify ticket operation.
         if (std::strcmp(_op_name, "mod") == 0) {
+            int i;
+            char myTime[50];
```
trel, line 13916: always initialize

trel:
some clang-tidy and clang-format opinions to address as well.    but looking very good.

alanking:
Please mark as Ready for Review when ready for review. Thanks!

plugins/database/src/db_plugin.cpp
```diff
@@ -14012,19 +14020,22 @@ namespace
...
                     }
 
-                    getNowStr(myTime);
+                    getNowStr(myTime); // NOLINT(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
                     i = 0;
-                    cllBindVars[i++] = myTime;
+                    // NOLINTBEGIN(cppcoreguidelines-pro-bounds-constant-array-index)
+                    cllBindVars[i++] = myTime; // NOLINT(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
                     cllBindVars[i++] = ticket_id_string.c_str();
+                    // NOLINTBEGIN(cppcoreguidelines-pro-bounds-constant-array-index)
```
korydraughn: NOLINTEND?

plugins/database/src/db_plugin.cpp
```diff
@@ -14012,19 +14020,22 @@ namespace
...
-                    irods::error ret = execute_sql("insert into R_TICKET_ALLOWED_HOSTS (ticket_id, host) values (? , ?)");
+                    irods::error ret =
+                        execute_sql("insert into R_TICKET_ALLOWED_HOSTS (ticket_id, host) values (? , ?)");
                     if (!ret.ok()) {
                         log_sql::error("Failed to add host [{}] to ticket [{}], status [{}]",
-                                        hostIp,
-                                        ticket_id_string.c_str(),
-                                        ret.code());
+                                       hostIp,
+                                       ticket_id_string.c_str(),
```
korydraughn, line 14032: If `ticket_id_string` is a `std::string` or `std::string_view`, then you drop the use of `.c_str()`.

fmtlib is used inside of the logger, and so it understands how to handle those types.

_This applies to other locations too._

plugins/database/src/db_plugin.cpp
```diff
@@ -14042,19 +14053,22 @@ namespace
...
-                    irods::error ret = execute_sql("insert into R_TICKET_ALLOWED_USERS (ticket_id, user_name) values (? , ?)");
+                    irods::error ret =
+                        execute_sql("insert into R_TICKET_ALLOWED_USERS (ticket_id, user_name) values (? , ?)");
                     if (!ret.ok()) {
                         log_sql::error("Failed to add user [{}] to ticket [{}], status [{}]",
-                                        _arg5,
-                                        ticket_id_string.c_str(),
-                                        ret.code());
+                                       _arg5,
+                                       ticket_id_string.c_str(),
```
korydraughn, line 14065: Drop `.c_str()`?

plugins/database/src/db_plugin.cpp
```diff
@@ -14072,19 +14086,22 @@ namespace
...
-                    irods::error ret = execute_sql("insert into R_TICKET_ALLOWED_GROUPS (ticket_id, group_name) values (? , ?)");
+                    irods::error ret =
+                        execute_sql("insert into R_TICKET_ALLOWED_GROUPS (ticket_id, group_name) values (? , ?)");
                     if (!ret.ok()) {
                         log_sql::error("Failed to add group [{}] to ticket [{}], status [{}]",
-                                        _arg5,
-                                        ticket_id_string.c_str(),
-                                        ret.code());
+                                       _arg5,
+                                       ticket_id_string.c_str(),
```
korydraughn, line 14098: Drop `.c_str()`?

plugins/database/src/db_plugin.cpp
```diff
@@ -14102,19 +14119,22 @@ namespace
...
-                    irods::error ret = execute_sql("delete from R_TICKET_ALLOWED_HOSTS where ticket_id = ? and host = ?");
+                    irods::error ret =
+                        execute_sql("delete from R_TICKET_ALLOWED_HOSTS where ticket_id = ? and host = ?");
                     if (!ret.ok()) {
                         log_sql::error("Failed to remove host [{}] from ticket [{}], status [{}]",
-                                        hostIp,
-                                        ticket_id_string.c_str(),
-                                        ret.code());
+                                       hostIp,
+                                       ticket_id_string.c_str(),
```
korydraughn, line 14131: Drop `.c_str()`?

plugins/database/src/db_plugin.cpp
```diff
@@ -14162,19 +14185,22 @@ namespace
...
-                    irods::error ret = execute_sql("delete from R_TICKET_ALLOWED_GROUPS where ticket_id = ? and group_name = ?");
+                    irods::error ret =
+                        execute_sql("delete from R_TICKET_ALLOWED_GROUPS where ticket_id = ? and group_name = ?");
                     if (!ret.ok()) {
                         log_sql::error("Failed to remove group [{}] from ticket [{}], status [{}]",
-                                        _arg5,
-                                        ticket_id_string.c_str(),
-                                        ret.code());
+                                       _arg5,
+                                       ticket_id_string.c_str(),
```
korydraughn, line 14197: Drop `.c_str()`?

korydraughn (commented):
Looking good.

scripts/irods/test/test_iticket.py
```diff
@@ -766,3 +767,86 @@ def test_write_tickets_for_collections_allow_creation_of_data_objects__issue_591
                 self.admin.run_icommand(['ichmod', '-M', 'own', self.admin.username, fname])
 
             self.admin.run_icommand(['iticket', 'delete', ticket_name])
+
+    def test_tickets_update_catalog_modify_time_when_modified__issue_7125(self):
+
+        def get_modification_and_creation_time():
+            out, err, ec = self.admin.run_icommand(['iquest', '%s...%s', "select TICKET_MODIFY_TIME, TICKET_CREATE_TIME where TICKET_STRING = '{}'".format(ticket_string)])
```
korydraughn, line 774: Why is `...` necessary? Would a space not work?

If you have a valid ticket, can't we assume that both the mtime and ctime come back as non-empty strings?

scripts/irods/test/test_iticket.py
```diff
@@ -766,3 +767,86 @@ def test_write_tickets_for_collections_allow_creation_of_data_objects__issue_591
...
+            self.assertEqual(ec, 0)
+            self.assertEqual(len(err), 0)
+            modification_time = out.strip().split('...')[0]
+            creation_time = out.strip().split('...')[1]
```
korydraughn, line 778: You're splitting the result twice rather than once and then reusing the result.

scripts/irods/test/test_iticket.py
```diff
@@ -766,3 +767,86 @@ def test_write_tickets_for_collections_allow_creation_of_data_objects__issue_591
...
+
+            time.sleep(1) # necessary because epoch time is in seconds
```
korydraughn, line 780: Consider pulling this out into the main test logic or rename the utility function so it's clear what's happening.

At the moment, this utility function is serving two jobs. (1) Get the mtime/ctime. (2) Sleep so there's a gap between mtime/ctime checks.

This hides info from the next reader of this test.

scripts/irods/test/test_iticket.py
```diff
@@ -766,3 +767,86 @@ def test_write_tickets_for_collections_allow_creation_of_data_objects__issue_591
...
+            return int(modification_time), int(creation_time)
+
+        ticket_string = "issue_7125_ticket"
+        file_name = "issue_7125_file"
+        _, _, ec = self.admin.run_icommand(['itouch', file_name])
```
korydraughn, line 785: Why run_icommand instead of assert_icommand here?

scripts/irods/test/test_iticket.py
Nishant2022: I had copied from work Alan had done earlier. A space would work.

korydraughn: Ok. I was mostly just wondering. The use of `...` is fine.

It wasn't clear to me if there was some other issue that that resolves.

scripts/irods/test/test_iticket.py
```diff
@@ -766,3 +767,110 @@ def test_write_tickets_for_collections_allow_creation_of_data_objects__issue_591
                 self.admin.run_icommand(['ichmod', '-M', 'own', self.admin.username, fname])
 
             self.admin.run_icommand(['iticket', 'delete', ticket_name])
+
+    def test_tickets_update_catalog_modify_time_when_modified__issue_7125(self):
```
korydraughn, line 771: We need to add another test for non-admin users.

korydraughn:
@Nishant2022 When you feel this is ready, please click the **Ready for review** button. I think we get a notification when the PR's status changes to/from draft.

Nishant2022:
I believe this is ready. I have run the unit tests and they pass. However, I have not run the core tests yet.

korydraughn:
Will review and post comments.

korydraughn:
Don't forget to mark as ready for review.

scripts/irods/test/test_iticket.py
```diff
@@ -768,7 +768,7 @@ def test_write_tickets_for_collections_allow_creation_of_data_objects__issue_591
 
             self.admin.run_icommand(['iticket', 'delete', ticket_name])
 
-    def test_tickets_update_catalog_modify_time_when_modified__issue_7125(self):
+    def test_tickets_update_catalog_modify_time_when_modified_admin__issue_7125(self):
```
korydraughn, line 771: How do you feel about the following?
```
test_ticket_mtime_is_updated_when_ticket_properties_are_modified_by_admin__issue_7125
```

scripts/irods/test/test_iticket.py
```diff
@@ -874,3 +874,110 @@ def get_modification_and_creation_time():
         previous_ts = modify_ts
         modify_ts, _ = get_modification_and_creation_time()
         self.assertGreater(modify_ts, previous_ts)
+
+    def test_tickets_update_catalog_modify_time_when_modified_user__issue_7125(self):
```
korydraughn, line 878: How do you feel about the following?
```
test_ticket_mtime_is_updated_when_ticket_properties_are_modified_by_non_admins__issue_7125
```

scripts/irods/test/test_iticket.py
```diff
@@ -874,3 +874,110 @@ def get_modification_and_creation_time():
...
+
+            return int(modification_time), int(creation_time)
+
+        ticket_string = "issue_7125_ticket_user"
+        file_name = "issue_7125_file_user"
+        self.user.assert_icommand(['itouch', file_name])
+
+        self.user.assert_icommand(['iticket', 'create', 'write', file_name, ticket_string])
+
+        modify_ts, create_ts = get_modification_and_creation_time()
```
korydraughn, line 897: Does anything use `create_ts`?

scripts/irods/test/test_iticket.py
Nishant2022: This looks clearer that what I had. I will implement it.

scripts/irods/test/test_iticket.py
Nishant2022: This looks clearer that what I had. I will implement it.

scripts/irods/test/test_iticket.py
Nishant2022: `create_ts` is only used once. Initially I was going to compare the modify time to the create time for every modification but realized that it would be redundant after the first time the modification time was different.

korydraughn: After creating the ticket, are the mtime and ctime identical?

Nishant2022: Immediately after creating the ticket, the mtime and ctime will be identical. After a ticket is modified, the ctime and mtime should differ.

korydraughn: It seems you don't need the ctime for this test. You agree?

Nishant2022: Yeah that is true. I will update the test.

alanking: Thinking on this a little, it might be worth doing the comparison. It's only redundant if we assume that every operation properly updates the mtime and also *does not* update the create time. Thoughts?

alanking:
Oops, my review is a little stale. I started it before some other updates came through that I missed. Sorry if there's some redundancy

plugins/database/src/db_plugin.cpp
```diff
@@ -13912,28 +13913,49 @@ namespace
...
             if (std::strcmp(_arg3, "uses") == 0) {
-                cllBindVars[0] = _arg4;
-                cllBindVars[1] = ticket_id_string.c_str();
-                cllBindVarCount = 2;
+                getNowStr(myTime); // NOLINT(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+                i = 0;
+                // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-constant-array-index)
+                cllBindVars[i++] = _arg4;
+                // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-constant-array-index)
+                cllBindVars[i++] = myTime; // NOLINT(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
```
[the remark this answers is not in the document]
Nishant2022: Clang-format would complain because it would go over the line length check. Also correct me if I am wrong but in 851cfc98277ad5968674d7a05bcf262fac666fa6 `
 -cppcoreguidelines-pro-bounds-array-to-pointer-decay` was not demoted to a warning as it is still in the "Warnings As Errors" section.

plugins/database/src/db_plugin.cpp
```diff
@@ -13912,28 +13913,49 @@ namespace
...
+                // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-constant-array-index)
+                cllBindVars[i++] = ticket_id_string.c_str();
+                cllBindVarCount = i;
 
-                return execute_sql("update R_TICKET_MAIN set uses_limit = ? where ticket_id = ?");
+                return execute_sql("update R_TICKET_MAIN set uses_limit = ?, modify_ts = ? where ticket_id = ?");
```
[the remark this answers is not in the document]
Nishant2022: I am not sure. @korydraughn do you have an answer?

scripts/irods/test/test_iticket.py
Nishant2022: I think that makes sense. I can check that the creation time is equal to the original create time and not equal to the modification time.

plugins/database/src/db_plugin.cpp
```diff
@@ -13912,28 +13913,49 @@ namespace
...
         if (std::strcmp(_op_name, "mod") == 0) {
+            int i{0};
+            char myTime[TIME_LEN]{}; // NOLINT(cppcoreguidelines-avoid-c-arrays, modernize-avoid-c-arrays)
+
             if (std::strcmp(_arg3, "uses") == 0) {
-                cllBindVars[0] = _arg4;
-                cllBindVars[1] = ticket_id_string.c_str();
-                cllBindVarCount = 2;
+                getNowStr(myTime); // NOLINT(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+                i = 0;
```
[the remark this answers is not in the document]
Nishant2022: You're right. There are only a few places, when `i` has already been incremented when I have to set it back to 0. Otherwise it is already at 0.

plugins/database/src/db_plugin.cpp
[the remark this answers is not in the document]
korydraughn: That linter option was demoted.

The leading minus sign means **remove**.

plugins/database/src/db_plugin.cpp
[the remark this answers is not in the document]
korydraughn: It looks like we do need to call `_rollback()`, but let's confirm that first. It's possible the rollback is handled earlier in the stack _(I don't think it is though)_.

@Nishant2022 Review the logic of `db_mod_ticket_op()` in db_plugin.cpp and let us know what you find.

plugins/database/src/db_plugin.cpp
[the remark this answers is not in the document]
korydraughn: If `i` is only used within a particular scope, consider moving `int i{}` into each block. That way, you're guaranteed no two uses of `i` overlap.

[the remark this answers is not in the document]
Nishant2022: Should that also include the `myTime` string?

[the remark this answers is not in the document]
korydraughn: Sure. That's fine.

The closer things are together, the easier it is to reason about.

plugins/database/src/db_plugin.cpp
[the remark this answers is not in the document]
Nishant2022: There does not appear to be any rollback logic in `db_mod_ticket_op()`.

We only need to roll back in the instances when two SQL statements are required correct? Such as when we insert and then update the modify time. If the update modify time fails, then we roll back the changes made during the insert.

However, I do not believe we need to do it when there is only one SQL statement because if it fails, we return the error code and nothing changes, correct?

plugins/database/src/db_plugin.cpp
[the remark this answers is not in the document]
Nishant2022: I rebased to include the changes from https://github.com/irods/irods/commit/851cfc98277ad5968674d7a05bcf262fac666fa6 yet I still see `cppcoreguidelines-pro-bounds-array-to-pointer-decay` errors in clang-tidy.

plugins/database/src/db_plugin.cpp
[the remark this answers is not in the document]
korydraughn: I'd argue it is more about the auto-commit option on every database connection.

Disabling auto-commit is the default for database connections in iRODS AFAIK. For that reason, we need to make sure we keep the database happy by properly committing or rolling back at the appropriate time.

This also helps the next developer who encounters the code.

korydraughn (commented):
Much better.

korydraughn:
Do the unit/core tests pass?

Nishant2022:
I am running them now.

Nishant2022:
All unit tests pass.

All core tests apart from `test_control_plane` pass. This is document in https://github.com/irods/irods/issues/7172

scripts/irods/test/test_iticket.py
```diff
@@ -888,3 +889,259 @@ def test_ticket_create_with_relative_paths_non_admin__issue_7103(self):
...
+        def get_modification_and_creation_time():
+            out, err, ec = self.admin.run_icommand(['iquest', '%s...%s', "select TICKET_MODIFY_TIME, TICKET_CREATE_TIME where TICKET_STRING = '{}'".format(ticket_string)])
+            self.assertEqual(ec, 0)
+            self.assertEqual(len(err), 0)
+
+            out = out.strip().split('...')
+            modification_time = out[0]
+            creation_time = out[1]
+
+            return int(modification_time), int(creation_time)
```
alanking, line 1032: This appears to be identical between the two tests. Could be broken out into a separate function for both to use. Up to you.

alanking (approved):
Optional thing I noticed

scripts/irods/test/test_iticket.py
Nishant2022: Yeah I can do that.

scripts/irods/test/test_iticket.py
```diff
@@ -1134,14 +1110,25 @@ def get_modification_and_creation_time():
...
         time.sleep(1)
         self.user.assert_icommand(['iticket', 'mod', ticket_string, 'remove', 'group', 'public'])
         previous_ts = modify_ts
-        modify_ts, create_ts = get_modification_and_creation_time()
+        modify_ts, create_ts = get_modification_and_creation_time(self, ticket_string)
         self.assertEqual(original_create_ts, create_ts)
         self.assertGreater(modify_ts, create_ts)
         self.assertGreater(modify_ts, previous_ts)
+
+def get_modification_and_creation_time(session, ticket_string):
```
alanking, line 1125: I think that `self` or `cls` would be a little more clear here. This isn't actually an `IrodsSession` but in fact is the instance of the test class.

Also, the indentation indicates that this is outside of the test class (i.e. top level indentation). Is that on purpose?

Nishant2022: I changed the parameter name to cls.

As for the indentation. It is meant to be outside the test class. This is because if it is inside the test class, it will be treated like a test case and fails.

alanking (approved):
Looks good to me. Please squash to one and leave off #s. We shall await one more approval...

korydraughn (approved):
Please add the pound.

Nishant2022:
Added #

alanking (approved):

Issue #7048, opened 2023-04-18
`clientLogin` family of functions must not print messages to stdout / stderr


korydraughn:
- [x] main
- [x] 4-3-stable
---
## Refactor
The `clientLogin` family of functions use the following to print errors to the terminal. This isn't good for client applications running as servers (e.g. the C++ REST API). For client applications which produce structured log messages, this is bad because the messages produced by these functions can break log file parsers.
https://github.com/irods/irods/blob/25d8f67df6c123ee6f0b3fe4695e4042b52cdb59/lib/core/src/clientLogin.cpp#L134-L155
https://github.com/irods/irods/blob/25d8f67df6c123ee6f0b3fe4695e4042b52cdb59/lib/core/src/clientLogin.cpp#L515

These functions must never write to the terminal. Instead, they should simply return the error as an integer value.

Nishant2022:
Should they log instead?

korydraughn:
These may not be able to log anything without linking to the logging libraries and we really don't want that.

Let's simply return the error and not print anything to the terminal.

PR #7149, opened 2023-06-28
[7048, 7161] Use error stack for `clientLogin` (main)

Nishant2022:
Removed calls to printError in clientLogin.cpp
Moved printError function to rcMisc.cpp

Addresses https://github.com/irods/irods/issues/7048

### DIFF
--- a/lib/core/include/irods/rodsError.h
+++ b/lib/core/include/irods/rodsError.h
@@ -2,7 +2,10 @@
 #define RODS_ERROR_H__
 
 #ifdef __cplusplus
-#include <string>
+#  include <cstdio>
+#  include <string>
+#else
+#  include <stdio.h>
 #endif
 
 #define ERR_MSG_LEN             1024
@@ -33,6 +36,20 @@ typedef struct ErrorStack {
     struct ErrorMessage **errMsg;
 } rError_t;
 
+/// \brief Allocate memory for the ErrorStack struct if the dereferenced pointer is null.
+///
+/// Below is an example demonstrating the usage.
+/// \code{.cpp}
+/// struct ErrorStack* p = NULL;
+/// allocate_error_stack_if_necessary(&p);
+/// assert(p != NULL);
+/// \endcode
+///
+/// \param[in,out] _stack the ErrorStack to be allocated
+///
+/// \since 4.3.1
+void allocate_error_stack_if_necessary(struct ErrorStack** _stack);
+
 /// \brief Add an error msg to the ErrorStack struct up to MAX_ERROR_MESSAGES.
 ///
 /// \param[in/out] myError the ErrorStack struct for the error msg.
@@ -43,6 +60,25 @@ typedef struct ErrorStack {
 /// \retval 0 on success
 /// \retval USER__NULL_INPUT_ERR if myError is NULL
 int addRErrorMsg(struct ErrorStack *myError, const int status, const char *msg);
+
+/// \brief Add an error msg to the ErrorStack struct up to MAX_ERROR_MESSAGES. Allocates ErrorStack if necessary.
+///
+/// Below is an example demonstrating the usage.
+/// \code{.cpp}
+/// struct ErrorStack* p = NULL;
+/// allocate_if_necessary_and_add_rError_msg(&p, 0, "Error Message");
+/// assert(p != NULL);
+/// \endcode
+///
+/// \param[in,out] _stack the ErrorStack struct for the error msg. Will be allocated if necessary.
+/// \param[in] _status the input error status.
+/// \param[in] _msg the error msg string. This string will be copied to myError.
+///
+/// \returns error code
+/// \retval 0 on success
+///
+/// \since 4.3.1
+int allocate_if_necessary_and_add_rError_msg(struct ErrorStack** _stack, const int _status, const char* _msg);
 
 /// \brief Frees the ErrorStack and its contents
 ///
@@ -72,7 +108,7 @@ int replErrorStack(struct ErrorStack *srcRError, struct ErrorStack *destRError);
 /// \retval USER__NULL_INPUT_ERR if myError is NULL
 int freeRErrorContent(struct ErrorStack *myError);
 
-/// \brief Print the contents of the provided ErrorStack
+/// \brief Print the contents of the provided ErrorStack to stdout
 ///
 /// \parblock
 /// The output takes the following form:
@@ -91,6 +127,28 @@ int freeRErrorContent(struct ErrorStack *myError);
 /// \retval 0 always
 /// \retval USER__NULL_INPUT_ERR if rError is NULL
 int printErrorStack(struct ErrorStack *rError);
+
+/// \brief Print the contents of the provided ErrorStack to given file
+///
+/// \parblock
+/// The output takes the following form:
+///
+/// Level 0: <error message>
+/// Level 1: <error message>
+/// ...
+/// Level 99: <error message>
+///
+/// If the error status for a particular ErrorMessage is STDOUT_STATUS, "Level <int>: " is not printed.
+/// \endparblock
+///
+/// \param[in,out] rError the ErrorStack which is to have its contents printed
+/// \param[in,out] stream the stream where the ErrorStack will be printed to
+///
+/// \returns error code
+/// \retval 0 always
+/// \retval USER__NULL_INPUT_ERR if rError is NULL
+/// \retval FILE_WRITE_ERROR if std::fprintf fails
+int print_error_stack_to_file(const struct ErrorStack* rError, FILE* stream);
 
 #ifdef __cplusplus
 }
--- a/lib/core/src/clientLogin.cpp
+++ b/lib/core/src/clientLogin.cpp
@@ -55,7 +55,8 @@ namespace
 
             log_auth::info(msg);
 
-            printError(&_comm, ec, const_cast<char*>(msg.data()));
+            // NOLINTNEXTLINE(bugprone-narrowing-conversions,cppcoreguidelines-narrowing-conversions)
+            allocate_if_necessary_and_add_rError_msg(&_comm.rError, ec, msg.c_str());
 
             return ec;
         }
@@ -68,7 +69,7 @@ namespace
 
             log_auth::info(msg);
 
-            printError(&_comm, ec, const_cast<char*>(msg.data()));
+            allocate_if_necessary_and_add_rError_msg(&_comm.rError, ec, msg.c_str());
 
             return ec;
         }
@@ -81,7 +82,7 @@ namespace
 
             log_auth::info(msg);
 
-            printError(&_comm, ec, const_cast<char*>(msg.data()));
+            allocate_if_necessary_and_add_rError_msg(&_comm.rError, ec, msg.c_str());
 
             return ec;
         }
@@ -94,7 +95,7 @@ namespace
 
             log_auth::info(msg);
 
-            printError(&_comm, ec, const_cast<char*>(msg.data()));
+            allocate_if_necessary_and_add_rError_msg(&_comm.rError, ec, msg.c_str());
 
             return ec;
         }
@@ -130,29 +131,6 @@ void setSessionSignatureClientside( char* _sig ) {
         ( unsigned char )_sig[15] );
 
 } // setSessionSignatureClientside
-
-int printError( rcComm_t *Conn, int status, char *routineName ) {
-    rError_t *Err;
-    rErrMsg_t *ErrMsg;
-    int i, len;
-    if ( Conn ) {
-        if ( Conn->rError ) {
-            Err = Conn->rError;
-            len = Err->len;
-            for ( i = 0; i < len; i++ ) {
-                ErrMsg = Err->errMsg[i];
-                fprintf( stderr, "Level %d: %s\n", i, ErrMsg->msg );
-            }
-        }
-    }
-    char *mySubName = NULL;
-    const char *myName = rodsErrorName( status, &mySubName );
-    fprintf( stderr, "%s failed with error %d %s %s\n", routineName,
-             status, myName, mySubName );
-    free( mySubName );
-
-    return 0;
-}
 
 int clientLoginPam( rcComm_t* Conn,
                     char*     password,
@@ -199,7 +177,7 @@ int clientLoginPam( rcComm_t* Conn,
        communication socket. */
     status = sslStart( Conn );
     if ( status ) {
-        printError( Conn, status, "sslStart" );
+        allocate_if_necessary_and_add_rError_msg(&Conn->rError, status, "sslStart");
         return status;
     }
 
@@ -209,7 +187,7 @@ int clientLoginPam( rcComm_t* Conn,
     pamAuthReqInp.timeToLive = ttl;
     status = rcPamAuthRequest( Conn, &pamAuthReqInp, &pamAuthReqOut );
     if ( status ) {
-        printError( Conn, status, "rcPamAuthRequest" );
+        allocate_if_necessary_and_add_rError_msg(&Conn->rError, status, "rcPamAuthRequest");
         sslEnd( Conn );
         return status;
     }
@@ -247,7 +225,7 @@ int clientLoginTTL( rcComm_t *Conn, int ttl ) {
     if ( int status = rcGetLimitedPassword( Conn,
                 &getLimitedPasswordInp,
                 &getLimitedPasswordOut ) ) {
-        printError( Conn, status, "rcGetLimitedPassword" );
+        allocate_if_necessary_and_add_rError_msg(&Conn->rError, status, "rcGetLimitedPassword");
         memset( userPassword, 0, sizeof( userPassword ) );
         return status;
     }
@@ -299,7 +277,8 @@ int clientLogin(rcComm_t* _comm, const char* _context, const char* _scheme_overr
 
         log_auth::info(msg);
 
-        printError(_comm, ec, const_cast<char*>(msg.data()));
+        // NOLINTNEXTLINE(bugprone-narrowing-conversions,cppcoreguidelines-narrowing-conversions)
+        allocate_if_necessary_and_add_rError_msg(&_comm->rError, ec, msg.c_str());
 
         return ec;
     }
@@ -369,7 +348,8 @@ int clientLogin(rcComm_t* _comm, const char* _context, const char* _scheme_overr
 
             log_auth::info(msg);
 
-            printError(_comm, ec, const_cast<char*>(msg.data()));
+            // NOLINTNEXTLINE(bugprone-narrowing-conversions,cppcoreguidelines-narrowing-conversions)
+            allocate_if_necessary_and_add_rError_msg(&_comm->rError, ec, msg.c_str());
 
             return ec;
         }
@@ -379,7 +359,7 @@ int clientLogin(rcComm_t* _comm, const char* _context, const char* _scheme_overr
 
             log_auth::info(msg);
 
-            printError(_comm, ec, const_cast<char*>(msg.data()));
+            allocate_if_necessary_and_add_rError_msg(&_comm->rError, ec, msg.c_str());
 
             return ec;
         }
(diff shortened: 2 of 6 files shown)


lib/core/src/clientLogin.cpp
```diff
@@ -131,29 +123,6 @@ void setSessionSignatureClientside( char* _sig ) {
...
-    }
-    char *mySubName = NULL;
-    const char *myName = rodsErrorName( status, &mySubName );
-    fprintf( stderr, "%s failed with error %d %s %s\n", routineName,
-             status, myName, mySubName );
-    free( mySubName );
-
-    return 0;
-}
-
```
korydraughn, line 156: Let's leave the implementation of this function here.

People writing C/C++ clients may be using it and moving it from this location could break their code.

korydraughn (commented):
Couple things ...

All of the places where you removed the `printError()` call could cause tests to fail. Especially ones that are checking for these error messages in stdout.

This change may require converting all of these calls to use the `rError` stack.

For one, the iCommands use these clientLogin functions and they rely on this `printError()` function to print when auth fails.

Please confirm what happens when you fail authentication with iCommands built on top of your PR.

lib/core/src/clientLogin.cpp
trel: can it be a wrapper/reference to just call the new/generic one?

korydraughn: I understand the wrapper/reference idea, but I'm not aware of a new/generic implementation of this function.

Please elaborate.

Nishant2022: @korydraughn at the moment, `printError` is defined in in rcMisc.h but implemented in clientLogin.cpp. That doesn't feel right to me but I see your point.

korydraughn: Oh. Thanks for the clarification.

If that's the case, then what you've done is correct.

Nishant2022:
At the moment, I know that these changes break the `test_ipasswd` test.

Currently in main, providing a bad password produces an error like this:

```
irods@1075626ff3d2:~$ ipasswd BAD_PASS
Error occurred while authenticating user [rods] [CAT_INVALID_AUTHENTICATION: failed to perform request

] [ec=-826000] failed with error -826000 CAT_INVALID_AUTHENTICATION 
```

With my changes there is no output. The current `test_ipasswd` test checks for the output so it fails with my changes.

I am not sure what you mean by using the `rError` stack.

korydraughn:
The `rError` stack is a type that is available in the `RsComm` which is meant to pass messages back to the client.

See https://github.com/irods/irods/blob/main/lib/core/include/irods/rodsError.h.

You'd attach the error messages using `addRErrorMsg()`.

The iCommands print out the content of the `rError` stack at the end of each run.

Nishant2022:
I found that the error stack is never initialized. As a result, we cannot pass errors to icommands using the error stack.

korydraughn:
@Nishant2022 and I discussed the changes of this PR and the current plan is to postpone work on this until we learn more.

It's possible these modifications change the behavior of the iCommands and C/C++ applications enough that it could be too risky to continue this work.

lib/core/src/rodsError.cpp
```diff
@@ -4,6 +4,13 @@
 
 struct ErrorStack;
 
+void allocateRErrorMsgIfNecessary(ErrorStack** myError) {
+    if (!*myError) {
+        *myError = (rError_t*)malloc(sizeof(rError_t));
```
korydraughn, line 9: `static_cast<rError_t*>(std::malloc(sizeof(rError_t)))`

lib/core/src/rodsError.cpp
```diff
@@ -4,6 +4,13 @@
...
+        memset(*myError, 0, sizeof(rError_t));
```
korydraughn, line 10: `memset` -> `std::memset`

lib/core/include/irods/rodsError.h
```diff
@@ -33,6 +33,8 @@ typedef struct ErrorStack {
     struct ErrorMessage **errMsg;
 } rError_t;
 
+void allocateRErrorMsgIfNecessary(struct ErrorStack** myError);
```
korydraughn, line 36: camel-case -> snake-case

korydraughn (commented):
Nice.

I'm hoping the unit/core tests come back clean.

lib/core/src/clientLogin.cpp
korydraughn: @Nishant2022 I think this can be marked as resolved.

Nishant2022:
Unit tests pass but I highly doubt core tests will pass until I update all calls to `clientLogin` in the icommands to call `printErrorStack` on failure.

Nishant2022:
Should `printErrorStack` be printing to `stderr`? Currently it prints to `stdout` but I feel it would make sense if it printed to `stderr`.

lib/core/include/irods/rodsError.h
```diff
@@ -33,7 +33,10 @@ typedef struct ErrorStack {
     struct ErrorMessage **errMsg;
 } rError_t;
 
-void allocateRErrorMsgIfNecessary(struct ErrorStack** myError);
+/// \brief Allocate memory for the ErrorStack struct if it is uninitialized
+///
+/// \param[in/out] myError the ErrorStack to be initialized
```
korydraughn, line 38: I think `[in/out]` is supposed to be `[in,out]`.

lib/core/include/irods/rodsError.h
```diff
@@ -33,7 +33,10 @@ typedef struct ErrorStack {
...
+void initialize_error_stack_if_necessary(struct ErrorStack** myError);
```
korydraughn, line 39: Consider replacing the word initialize with allocate.

Initialize can mean several things, but allocate is very clear in its definition.

lib/core/include/irods/rodsError.h
```diff
@@ -33,7 +33,10 @@ typedef struct ErrorStack {
     struct ErrorMessage **errMsg;
 } rError_t;
 
-void allocateRErrorMsgIfNecessary(struct ErrorStack** myError);
+/// \brief Allocate memory for the ErrorStack struct if it is uninitialized
```
korydraughn, line 36: Consider rephrasing this.

Uninitialized can mean several things.

This is just one example that you could use.
```
Allocates an ErrorStack if the deferenced pointer is null.

More details ...
```

lib/core/include/irods/rodsError.h
korydraughn, line 38: It's difficult to express `T**` in words.

Consider including an example demonstrating how to use this function.

Perhaps something like:
```cpp
struct ErrorStack* p = NULL;
allocate_error_stack_if_necessary(&p);
assert(NULL != p);
```

Here's a link to how code examples can be included in documentation.
https://github.com/irods/irods/blob/8797c1c944175fe28024474b53cdef826672e74b/lib/core/include/irods/irods_logger.hpp#L565-L572

lib/core/include/irods/rodsError.h
korydraughn, line 39: Consider replacing `myError` with something else.

How about `_stack`?

This aligns with the function signature of `pop_error_message` at the bottom of this file.

Don't forget to update the documentation to match whatever you pick.

korydraughn (commented):
Very close!

korydraughn:
> Should `printErrorStack` be printing to `stderr`? Currently it prints to `stdout` but I feel it would make sense if it printed to `stderr`.

People may have written scripts that depend on the existing behavior. Let's leave the target output target as-is.

We'll make the new CLI use the appropriate output target.

alanking:
Please mark as Ready for Review when this is ready for review. Thanks!

Nishant2022:
> > Should `printErrorStack` be printing to `stderr`? Currently it prints to `stdout` but I feel it would make sense if it printed to `stderr`.
> 
> People may have written scripts that depend on the existing behavior. Let's leave the output target as-is.
> 
> We'll make the new CLI use the appropriate output target.

This would mean additional changes to `test_ipasswd` which currently checks for the error message in `stderr`. I don't mind changing it, just letting you know.

korydraughn:
Let's find out how many tests fail due to this PR.

That will help us make a decision.

Nishant2022:
After running the core tests with my changes in https://github.com/irods/irods_client_icommands/pull/398, `test_iphymv` and `test_ipasswd` fail because `printErrorStack` prints to `stdout`. I can modify those two tests to instead look for the errors in `stdout`.

korydraughn:
Are those the only failing tests?

SwooshyCueb:
How many tests in those test suites are failing? Just one from each?

Nishant2022:
There is one failure in `iphymv` and two in `ipasswd`. There is also a failure in `test_iadmin` which is solved when the error stack is printed to `stderr`. Maybe instead of changing the tests to look at `stdout` or change `printErrorStack` to print to `stderr` we instead make another print error function that prints to `stderr`?

korydraughn:
I think you're headed in the right direction. You can add a new function that supports changing which stream to write to. The old function can then call into that with `stderr` to maintain backwards compatibility.

Thoughts?

Nishant2022:
Kory and I talked. I will implement a new function that prints to a file and the old function will be a wrapper over it that prints to `stderr`.

trel:
I like it.

Nishant2022:
Including the changes made in https://github.com/irods/irods_client_icommands/pull/398, all unit tests pass and all core test_pass apart from `test_control_plane`.

scripts/irods/test/test_ipasswd.py
```diff
@@ -139,7 +139,7 @@ def test_ipasswd_rodsuser_should_fail(self):
         #
         expected_lines = [
                             'CAT_INVALID_AUTHENTICATION',
-                            '826000 CAT_INVALID_AUTHENTICATION'
+                            '826000'
```
korydraughn, line 142: Why does this need to be changed?
Did the output change?

scripts/irods/test/test_ipasswd.py
```diff
@@ -224,11 +224,6 @@ def test_ipasswd_rodsuser_should_fail(self):
         str1 = "CAT_INVALID_AUTHENTICATION: failed to perform request"
         self.assertIn(str1, stderr, "ipasswd: Expected stderr: \"...{0}...\", got: \"{1}\"".format(str1, stderr))
 
-        # and...
-
-        str1 = "failed with error -826000 CAT_INVALID_AUTHENTICATION"
-        self.assertIn(str1, stderr, "ipasswd: Expected stderr: \"...{0}...\", got: \"{1}\"".format(str1, stderr))
-
```
korydraughn, line 231: Has the output changed for ipasswd?

scripts/irods/test/test_ipasswd.py
```diff
@@ -321,7 +316,7 @@ def test_ipasswd_rodsadmin_should_fail(self):
         #
         expected_lines = [
                             'CAT_INVALID_AUTHENTICATION',
-                            '826000 CAT_INVALID_AUTHENTICATION'
+                            '826000'
```
korydraughn, line 319: Same question here.

scripts/irods/test/test_ipasswd.py
```diff
@@ -405,8 +400,3 @@ def test_ipasswd_rodsadmin_should_fail(self):
 
         str1 = "CAT_INVALID_AUTHENTICATION: failed to perform request"
         self.assertIn(str1, stderr, "ipasswd: Expected stderr: \"...{0}...\", got: \"{1}\"".format(str1, stderr))
-
-        # and...
-
-        str1 = "failed with error -826000 CAT_INVALID_AUTHENTICATION"
-        self.assertIn(str1, stderr, "ipasswd: Expected stderr: \"...{0}...\", got: \"{1}\"".format(str1, stderr))
```
korydraughn, line 412: Same question here.

korydraughn (commented):
We're really close now.

scripts/irods/test/test_ipasswd.py
Nishant2022: Yes this is a product of moving from `printError` to `printErrorStack` or the new `print_error_stack_to_file` functions.

The old output would look like this:
```
Error occurred while authenticating user [rods] [CAT_INVALID_AUTHENTICATION: failed to perform request

] [ec=-826000] failed with error -826000 CAT_INVALID_AUTHENTICATION
```

The new output looks like this:
```
Level 0: Error occurred while authenticating user [rods] [CAT_INVALID_AUTHENTICATION: failed to perform request

] [ec=-826000]
```

We lose the extra "failed with error -826000 CAT_INVALID_AUTHENTICATION" which In my opinion was redundant anyway.

scripts/irods/test/test_ipasswd.py
Nishant2022: See https://github.com/irods/irods/pull/7149#discussion_r1270750865

scripts/irods/test/test_ipasswd.py
Nishant2022: See https://github.com/irods/irods/pull/7149#discussion_r1270750865

scripts/irods/test/test_ipasswd.py
Nishant2022: See https://github.com/irods/irods/pull/7149#discussion_r1270750865

scripts/irods/test/test_ipasswd.py
korydraughn: Is it just the message that has changed?

lib/core/src/clientLogin.cpp
```diff
@@ -55,7 +55,9 @@ namespace
 
             log_auth::info(msg);
 
-            printError(&_comm, ec, const_cast<char*>(msg.data()));
+            allocate_error_stack_if_necessary(&_comm.rError);
+            // NOLINTNEXTLINE(bugprone-narrowing-conversions,cppcoreguidelines-narrowing-conversions)
+            addRErrorMsg(_comm.rError, ec, msg.c_str());
```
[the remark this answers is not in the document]
Nishant2022: I think a wrapper may be better as `addRErrorMsg` can continue to return `USER__NULL_INPUT_ERR` if it is provided a null value.

scripts/irods/test/test_ipasswd.py
Nishant2022: Yeah. The error message changed so the tests are invalid.

lib/core/src/clientLogin.cpp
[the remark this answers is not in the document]
alanking: Yep, sounds good to me. We can always make an issue for that too for the sake of time. Up to you

scripts/irods/test/test_ipasswd.py
korydraughn: Okay, thinking out loud ...

It's an unstructured error message for the most part. Therefore, users should not rely on the message to automate anything. They need to react to the error code returned from the program. The error code being the code returned from `main()`.

With that said, I think this is fine.

@trel @alanking Thoughts?

alanking: Does this affect anything outside of `ipasswd`? If not, then I think it's okay. If so, then it may be worth reconsidering/adjusting something to make the message the same.

I feel that `ipasswd` being solely affected is well-scoped enough to be manageable.

Nishant2022: It affects almost all icommands if a user is not logged in. 

Currently if a user is not logged in, they will receive the first message. That being said, `test_ipasswd` is the only test that looks for that message.

trel: it's still going to check for the full string, just alone, without the space in the literal string.

feels okay to me.   i'll be surprised to hear from anyone who has a script that is parsing/caring about return codes for logged-out attempts.

korydraughn: Ah, that's right. This is mostly matters for failed login attempts.

Everything else using `printErrorStack` will remain unchanged.

@Nishant2022 Is that correct?

Nishant2022: That's right. Everything that uses `printErrorStack` will remain unchanged. In this PR I am only adding the ability to pass the client login errors on failed logins through the error stack, and as a result, we need to print them to `stderr` from the icommands to maintain the same functionality. The same amount of information is being passed through, but it is in a slightly different format.

korydraughn: Excellent. Thanks.

lib/core/include/irods/rodsError.h
```diff
@@ -92,6 +106,28 @@ int freeRErrorContent(struct ErrorStack *myError);
...
+/// Level 0: <error message>
+/// Level 1: <error message>
+/// ...
+/// Level 99: <error message>
+///
+/// If the error status for a particular ErrorMessage is STDOUT_STATUS, "Level <int>: " is not printed.
+/// \endparblock
+///
+/// \param[in/out] stream the stream where the ErrorStack will be printed to
+/// \param[in/out] rError the ErrorStack which is to have its contents printed
```
korydraughn, line 123: `\param[in/out]` -> `\param[in,out]`

korydraughn:
Once the review comments are resolved, running test_ipasswd.py should be enough to confirm things are good.

If the test is happy, squash to taste and leave off the pounds.

Nishant2022:
Addressed the remaining review comments. Unit tests pass and `test_ipasswd` passes.

Squashed commits.

lib/core/include/irods/rodsError.h
```diff
@@ -92,6 +125,28 @@ int freeRErrorContent(struct ErrorStack *myError);
...
+/// \endparblock
+///
+/// \param[in,out] stream the stream where the ErrorStack will be printed to
+/// \param[in,out] rError the ErrorStack which is to have its contents printed
+///
+/// \returns error code
+/// \retval 0 always
+/// \retval USER__NULL_INPUT_ERR if rError is NULL
+/// \retval FILE_WRITE_ERROR if std::fprintf fails
+int print_error_stack_to_file(FILE* stream, const struct ErrorStack* rError);
```
korydraughn, line 148: Missing header include for `FILE`.

```cpp
#ifdef __cplusplus
#  include <cstdio>
#  include <string>
#else
#  include <stdio.h>
#endif
```

korydraughn, line 148: Consider swapping the order of the parameters.

The other functions in this file take `ErrorStack` as the first parameter.

lib/core/include/irods/rodsError.h
```diff
@@ -44,6 +58,25 @@ typedef struct ErrorStack {
...
+///
+/// \param[in,out] _stack the ErrorStack struct for the error msg. Will be allocated if necessary.
+/// \param[in] status the input error status.
+/// \param[in] msg the error msg string. This string will be copied to myError.
+///
+/// \returns error code
+/// \retval 0 on success
+///
+/// \since 4.3.1
+int allocate_if_necessary_and_add_rError_msg(struct ErrorStack** _stack, const int status, const char* msg);
```
korydraughn, line 78: Missing leading underscores for other parameters here and in other locations.

You can remove the leading underscore from `_stack` or add one to the other parameters.

korydraughn (commented):
Spotted a few more things.

Please squash to taste once they are resolved.

lib/core/include/irods/rodsError.h
Nishant2022: I put the `FILE` first to look like some c functions such as `fprintf`. However, there are other c functions like `fputs` where the data goes first followed by the `FILE` pointer. Which one do you think makes more sense?

korydraughn: Yeah, C isn't very consistent with ordering of parameters.

Let's swap them.

Nishant2022:
Addressed review comments and updated companion PR.

Unit tests pass and `test_ipasswd` passes.

korydraughn (approved):
Good job!

Please add the pounds.

alanking (approved):

korydraughn:
@Nishant2022 Please cherry-pick PRs to 4-3-stable branch and leave off pounds.

alanking:
@Nishant2022 - I can take care of it. Thanks!

Issue #7158, opened 2023-06-29
Provide a mechanism for plugins to tell the server what flags to pass to `dlopen()`


SwooshyCueb:
- [x] main
- [x] 4-3-stable
---
Related: irods/irods_rule_engine_plugin_python#136

Currently, we pass `RTLD_LAZY | RTLD_LOCAL` to `dlopen()` when loading plugins. We need to provide a mechanism that plugins can use to tell the server if different flags are required.

The current plan is to use `DT_FLAGS_1` for this purpose.

korydraughn:
Is there documentation you can link to this issue people interested in reading about `DT_FLAGS_1`?

SwooshyCueb:
Ehhh sorta?
So the `DT_FLAGS_1` dynamic table entry is an extension to ELF that is not in the LSB Specification or the System V ABI. My understanding is that Sun was the only vendor to ever use it, so [what little documentation there is](https://docs.oracle.com/cd/E19957-01/806-0641/chapter6-42444/index.html) doesn't actually reflect how it is used (or rather, not used) in Linux today. 
However, everything you need to parse a `DT_FLAGS_1` dynamic table entry is in `elf.h` (both glibc's and muslc's), and there are arguments that can be passed to `bfd` and `lld` (the most common linkers in use on Linux (and FreeBSD)) that can direct the linker to create a `DT_FLAGS_1` dynamic table entry and set certain flags. The one we're concerned with is `DT_1_GLOBAL`, which is supposed to tell the runtime linker to always load the library's symbols in the global namespace, but almost no implementation actually respects. The argument to pass to the build-time linker is `-z global` (or `-Wl,z,global` if calling via clang or gcc).

SwooshyCueb:
Moved the documentation task to irods/irods_docs#207. This issue and related issues can be closed now

PR #7159, opened 2023-06-29
[#7158] Determine RTLD flags from DT_FLAGS_1 for loading plugins

SwooshyCueb:
Addresses #7158 for main.
Prerequisite for irods/irods_rule_engine_plugin_python#157

### DIFF
--- a/lib/core/include/irods/irods_load_plugin.hpp
+++ b/lib/core/include/irods/irods_load_plugin.hpp
@@ -4,6 +4,7 @@
 // =-=-=-=-=-=-=-
 // My Includes
 #include "irods/irods_log.hpp"
+#include "irods/irods_logger.hpp"
 #include "irods/irods_plugin_name_generator.hpp"
 #include "irods/irods_configuration_keywords.hpp"
 #include "irods/getRodsEnv.h"
@@ -13,10 +14,15 @@
 
 // =-=-=-=-=-=-=-
 // STL Includes
+#include <array>
+#include <cstdint>
+#include <fstream>
 #include <string>
+#include <string_view>
 #include <sstream>
 #include <iostream>
 #include <algorithm>
+#include <vector>
 
 // =-=-=-=-=-=-=-
 // Boost Includes
@@ -26,6 +32,7 @@
 // =-=-=-=-=-=-=-
 // dlopen, etc
 #include <dlfcn.h>
+#include <elf.h>
 
 namespace irods
 {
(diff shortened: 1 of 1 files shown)


SwooshyCueb:
Still running tests, but they're far enough along that I think it's safe to mark this PR as ready for review.

SwooshyCueb:
I think we should document this functionality and how to use it, but I'm not sure where

lib/core/include/irods/irods_load_plugin.hpp
```diff
@@ -94,6 +101,274 @@ namespace irods
...
+                // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+                lib.read(reinterpret_cast<char*>(&elf_hdr_in), sizeof(elf_hdr_in));
+                e_shoff = elf_hdr_in.e_shoff;
+                e_shentsize = elf_hdr_in.e_shentsize;
+                e_shnum = elf_hdr_in.e_shnum;
+                e_shstrndx = elf_hdr_in.e_shstrndx;
+            }
+            else {
+                // clang-format off
+                logger::warn({
```
trel, line 178: confirming this is `warn` on purpose.

lib/core/include/irods/irods_load_plugin.hpp
```diff
@@ -94,6 +101,274 @@ namespace irods
...
+                });
+                // clang-format on
+                // NOLINTNEXTLINE(hicpp-signed-bitwise)
+                return rtld_lazy_or_now | rtld_local_or_global | rtld_nodelete;
+            }
+
+            // TODO: some platforms do not require a section table for libraries
+            if (e_shoff == 0) {
+                // clang-format off
+                logger::warn({
```
trel, line 191: same, warn vs error.

lib/core/include/irods/irods_load_plugin.hpp
SwooshyCueb: I used `warn` because the situation is less severe than than the cases we use `error` for above, but either one makes sense here to me.

lib/core/include/irods/irods_load_plugin.hpp
SwooshyCueb: `warn` is intentional here. The absence of a section table doesn't necessarily mean there's something wrong, just that we haven't taught the server how to find `DT_FLAGS_1` without one.

trel: great.

lib/core/include/irods/irods_load_plugin.hpp
trel: got it - up to you.

lib/core/include/irods/irods_load_plugin.hpp
```diff
@@ -94,6 +101,274 @@ namespace irods
 
     } // resolve_plugin_path
 
+    /**
+     * \brief attempt to parse DT_FLAGS_1 to identify flags to pass to dlopen()
+     *
+     * \since 4.3.1
+     **/
```
korydraughn, line 108: Let's stick with `///` since that appears to have become the convention.

Also, missing `\param[in] _plugin_path description ...`, `\throws`, and `\returns`.

lib/core/include/irods/irods_load_plugin.hpp
```diff
@@ -94,6 +101,274 @@ namespace irods
...
+        int rtld_local_or_global = RTLD_LOCAL;
+        int rtld_nodelete = 0;
+
+        try {
+            std::ifstream lib(_plugin_path, std::ios::binary);
+            if (!lib) {
+                // clang-format off
+                logger::error({
+                    {"plugin_path", _plugin_path},
+                    {"message", "Could not parse DT_FLAGS_1: failed to open plugin file"}
```
alanking, line 124: Our regular/default log messages use the key "log_message". Should we use the same key for consistency?

I realize that by now this may not be consistent everywhere, but figured I'd ask.

lib/core/include/irods/irods_load_plugin.hpp
```diff
@@ -94,6 +101,274 @@ namespace irods
 
     } // resolve_plugin_path
 
+    /**
+     * \brief attempt to parse DT_FLAGS_1 to identify flags to pass to dlopen()
+     *
+     * \since 4.3.1
+     **/
+    // NOLINTNEXTLINE(readability-function-cognitive-complexity) this is technically a parser
+    inline int get_plugin_rtld_flags(const std::string& _plugin_path) noexcept
```
korydraughn, line 110: Consider dropping the use of `noexcept`.

The logger can throw exceptions.

lib/core/include/irods/irods_load_plugin.hpp
```diff
@@ -94,6 +101,274 @@ namespace irods
...
+                    {"plugin_path", _plugin_path},
+                    {"message", "Could not parse DT_FLAGS_1: unknown or invalid ELFCLASS"},
+                    {"ELFCLASS", std::to_string(static_cast<std::uint32_t>(ei_class))}
+                });
+                // clang-format on
+                // NOLINTNEXTLINE(hicpp-signed-bitwise)
+                return rtld_lazy_or_now | rtld_local_or_global | rtld_nodelete;
+            }
+
+            // TODO: some platforms do not require a section table for libraries
```
korydraughn, line 188: If the section table isn't available, will that affect the PREP in regard to the DT_FLAGS_1 solution you've built?

lib/core/include/irods/irods_load_plugin.hpp
```diff
@@ -94,6 +101,274 @@ namespace irods
...
+                // clang-format on
+                // NOLINTNEXTLINE(hicpp-signed-bitwise)
+                return rtld_lazy_or_now | rtld_local_or_global | rtld_nodelete;
+            }
+
+            // seek to section table
+            lib.seekg(e_shoff);
+
+            // some data may be contained in the first entry
+            if ((e_shnum == 0) || (e_shstrndx == SHN_XINDEX)) {
```
korydraughn, line 204: For non-const variables, let's invert the operands on each side of `==` to guard against future modifications.
```c++
0 == e_shnum || ...
```

lib/core/include/irods/irods_load_plugin.hpp
```diff
@@ -94,6 +101,274 @@ namespace irods
...
+                    Elf32_Shdr s_hdr_in{};
+                    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+                    lib.read(reinterpret_cast<char*>(&s_hdr_in), sizeof(s_hdr_in));
+
+                    sh_type = s_hdr_in.sh_type;
+                    if (sh_type != SHT_DYNAMIC) {
+                        continue;
+                    }
+                    sh_name = s_hdr_in.sh_name;
+                    s_name = std::string_view(&e_strtbl_vec.data()[sh_name]);
```
korydraughn, line 279: So `&e_strtbl_vec[sh_name]` doesn't work?

Is the pointer returned null-terminated?

alanking (commented):
Seems fine overall in that nothing really jumped out at me. I might need to go read the ELF spec :) Just saw one thing which may or may not matter.

lib/core/include/irods/irods_load_plugin.hpp
SwooshyCueb: Good catch. I used `log_message` in PREP so I'll use that here too.

lib/core/include/irods/irods_load_plugin.hpp
```diff
@@ -94,6 +101,274 @@ namespace irods
...
+
+                        if (de_in.d_tag != DT_FLAGS_1) {
+                            continue;
+                        }
+
+                        // NOLINTNEXTLINE(cppcoreguidelines-pro-type-union-access)
+                        dt_flags_1 = de_in.d_un.d_val;
+                    }
+
+                    if ((dt_flags_1 & static_cast<std::uint64_t>(DF_1_NOW)) != 0U) {
```
korydraughn, line 339: Just curious. Why `!= 0U` instead of `== DF_1_NOW`?

Are you interested in a subset of the bits being set?

lib/core/include/irods/irods_load_plugin.hpp
```diff
@@ -94,6 +101,274 @@ namespace irods
...
+            // clang-format off
+            logger::error({
+                {"plugin_path", _plugin_path},
+                {"message", "Caught unknown exception while parsing DT_FLAGS_1"}
+            });
+            // clang-format on
+        }
+        // NOLINTNEXTLINE(hicpp-signed-bitwise)
+        return rtld_lazy_or_now | rtld_local_or_global | rtld_nodelete;
+    }
```
korydraughn, line 370: Let's add a trailing comment containing the name of this function here.

lib/core/include/irods/irods_load_plugin.hpp
korydraughn, line 110: Consider adding comments to the most important parts of this function. It's just nice not having to find documentation for all the scary data structures to understand what is happening here.

korydraughn (commented):
Very nice implementation.

Once approved, please capture the TODOs in issues and link to the relevant parts of the implementation.

lib/core/include/irods/irods_load_plugin.hpp
SwooshyCueb: Yes. Without a section table, we can't currently get to the dynamic entry with `DT_FLAGS_1`, so the plugin would be loaded with the default flags.

I don't see this coming up any time soon. We presently only support x86_64 on Linux, so there will always be a section table. Nonetheless, I'll be opening an issue covering the TODOs once everything else passes review.

lib/core/include/irods/irods_load_plugin.hpp
SwooshyCueb: > So `&e_strtbl_vec[sh_name]` doesn't work?

My understanding is that depends on the implementation of `std::vector`, but I could be mistaken.

> Is the pointer returned null-terminated?

Yes.

lib/core/include/irods/irods_load_plugin.hpp
SwooshyCueb: Yes, if any other flags are also set, `== DF_1_NOW` would return false, even if `DF_1_NOW` is set.

lib/core/include/irods/irods_load_plugin.hpp
SwooshyCueb: You're gonna have to unwrap that one for me. How does switching the operands guard against future modifications?

trel: just protects against fat-fingering the `==` to an assignment...

https://en.wikipedia.org/wiki/Yoda_conditions

SwooshyCueb: That makes sense now that I think about it. I'll make the changes.

lib/core/include/irods/irods_load_plugin.hpp
SwooshyCueb: To my understanding, the only potential source of exceptions from this function would be the logging in the exception handlers, so I'm not sure what to put for `\throws`.

korydraughn: The logger uses nlohmann::json, fmtlib, etc.

I think std::exception is good enough. At the very least, anyone calling this needs to catch std::exception to be safe.

If it helps, I've started leaving out **why** a particular exception may be thrown. I find that the message can quickly become incorrect over time and therefore it's safer to just let the dev know which exceptions can be thrown. They can interrogate the `.what()` for the reason.

lib/core/include/irods/irods_load_plugin.hpp
korydraughn: Ok. That works for me.

lib/core/include/irods/irods_load_plugin.hpp
korydraughn: Unless I'm misunderstanding, the bitwise-and with `DF_1_NOW` will only return the bits matching `DF_1_NOW`.

lib/core/include/irods/irods_load_plugin.hpp
korydraughn: Yes. Please mark as resolved.

lib/core/include/irods/irods_load_plugin.hpp
SwooshyCueb: oh, I misunderstood your question.

clang-tidy suggested `!= 0U`, so that's what I did. But `== DF_1_NOW` seems more correct. I'll make the changes.

korydraughn (commented):
I think we're in a good spot.

Please squash without the pounds and we'll do one last look over it before approving.

SwooshyCueb:
Squashed

korydraughn (approved):
Please add the pound.

alanking (approved):

SwooshyCueb:
Looks like there may be a problem with Centos 7, so I'm going to hold off until I can figure it out.

SwooshyCueb:
The problem was a test I added relying on a Python module that didn't exist yet. I've removed the test in question. https://github.com/irods/irods_rule_engine_plugin_python/pull/157#issuecomment-1623880239

\#'d

alanking (approved):

Issue #6297, opened 2022-04-04
Consider defining and adding constants for common database column sizes to Packing instructions


korydraughn:
See conversation starting here: https://github.com/irods/irods/pull/6272#discussion_r835787176

PR #7178, opened 2023-07-08
Exposed new options for rc_switch_user and added missing feature test macro (main)

korydraughn:
Required for the HTTP API.


lib/api/include/irods/switch_user.h
```diff
@@ -19,35 +21,33 @@ extern "C" {
...
+///     - It must be non-empty
+///     - It must not include the zone
+/// - \p zone: The part of the iRODS username following the pound sign. The following requirements must be satisfied:
+///     - The length must not exceed sizeof(UserInfo::rodsZone) - 1 bytes
+///     - It must be null-terminated
+///     - It must be non-empty
+/// - \p update_proxy_user: If set to 1, the proxy user is updated as well. This option will result in losing the
+///                         ability to invoke rc_switch_user if this option is set to 1 and the user to switch to is
+///                         not a \p rodsadmin. This option does not affect server-to-server connections.
+/// - \p close_svr_to_svr_connections: If set to 1, all connections created through redirection will be closed.
```
korydraughn, line 42: This needs to be moved to the data type instead of this function.

plugins/api/include/irods/plugins/api/switch_user_types.h
```diff
@@ -7,9 +7,11 @@ typedef struct SwitchUserInput // NOLINT(modernize-use-using)
     char username[64];
     char zone[64];
     // NOLINTEND(modernize-avoid-c-arrays)
+    int update_proxy_user;
+    int close_svr_to_svr_connections;
```
korydraughn, line 11: Document this struct.

lib/api/include/irods/switch_user.h
```diff
@@ -19,35 +21,33 @@ extern "C" {
...
-/// - The length must not exceed sizeof(UserInfo::rodsZone) - 1 bytes
-/// - It must be null-terminated
-/// - It must be non-empty
+/// On failure, the RcComm should be closed to avoid potential issues.
+///
+/// \param[in] _comm  A pointer to a RcComm.
+/// \param[in] _input \parblock A pointer to a SwitchUserInput.
+///
+/// SwitchUserInput member variables:
+/// - \p username: The part of the iRODS username preceding the pound sign. The following requirements must be satified:
```
trel, line 30: satified- > satisfied

lib/api/include/irods/switch_user.h
```diff
@@ -19,35 +21,33 @@ extern "C" {
...
+///     - The length must not exceed sizeof(UserInfo::userName) - 1 bytes
+///     - It must be null-terminated
+///     - It must be non-empty
+///     - It must not include the zone
```
trel, line 34: does this really mean 'it must not include a pound (#) sign'?

lib/api/include/irods/switch_user.h
```diff
@@ -19,35 +21,33 @@ extern "C" {
...
+/// - \p zone: The part of the iRODS username following the pound sign. The following requirements must be satisfied:
+///     - The length must not exceed sizeof(UserInfo::rodsZone) - 1 bytes
+///     - It must be null-terminated
+///     - It must be non-empty
+/// - \p update_proxy_user: If set to 1, the proxy user is updated as well. This option will result in losing the
+///                         ability to invoke rc_switch_user if this option is set to 1 and the user to switch to is
+///                         not a \p rodsadmin. This option does not affect server-to-server connections.
```
trel, line 41: 'does not affect' -> 'will be ignored by' or 'ignored on' or 'ignored for' ?

lib/core/include/irods/library_features.h
```diff
@@ -19,6 +19,11 @@
 ///
 /// \since 4.3.1
 
+/// Defined if the development library supports the C++ ticket administration library.
+///
+/// \since 4.3.1
+#define IRODS_HAS_LIBRARY_TICKET_ADMINISTRATION 202209L
```
trel, line 25: 202209L ?     not 202307?

plugins/api/src/switch_user.cpp
```diff
@@ -121,24 +121,36 @@ namespace
...
-        const auto switch_user_or_disconnect = [&client = _comm.clientUser](rodsServerHost& _host) {
+        const auto switch_user_or_disconnect = [&](rodsServerHost& _host) {
             if (!_host.conn) {
                 log_api::trace("No connection to remote host [{}].", _host.hostName->name);
                 return;
             }
 
-            // Disconnect if the remote server's version is older than 4.3.1.
+            // Disconnect if the client requested that all server-to-server connections be closed or
+            // if the remote server's version is older than 4.3.1.
```
trel, line 133: 'older than' -> 'less than' ?

or 'prior to'

or 'earlier than'... as i see 'later' just below in the next couple lines...

unit_tests/src/test_rc_switch_user.cpp
```diff
@@ -282,9 +305,122 @@ TEST_CASE("rc_switch_user cannot be invoked by non-admins")
...
+    const adm::user bob{"test_user_bob", env.rodsZone};
+    REQUIRE_NOTHROW(adm::client::add_user(conn, bob));
+    REQUIRE_NOTHROW(adm::client::modify_user(conn, bob, adm::user_password_property{"rods"}));
+
+    irods::at_scope_exit remove_test_user_bob{[&conn, &bob] { REQUIRE_NOTHROW(adm::client::remove_user(conn, bob)); }};
+
+    //
+    // Call rc_switch_user multiple times and toggle between two rodsusers.
+    // It is important that the users being switched to are rodsusers because doing this
+    // with rodsadmins may bypass important permission checks.
```
trel, line 399: consider groupadmins too - even if just in the comment... (as i'm sure they're the same as rodsusers for the purposes of this testing)

trel (commented):
very nice.

lib/core/include/irods/library_features.h
korydraughn: The value represents when the ticket admin lib was first introduced into the dev library.

plugins/api/src/switch_user.cpp
korydraughn: Good catch. This should be "earlier than".

Will change.

unit_tests/src/test_rc_switch_user.cpp
korydraughn: Will do. I can add a check for groupadmins too.

lib/api/include/irods/switch_user.h
korydraughn: That is correct.

lib/api/include/irods/switch_user.h
korydraughn: done

plugins/api/src/switch_user.cpp
korydraughn: done

unit_tests/src/test_rc_switch_user.cpp
korydraughn: done

korydraughn: Made one of the users a groupadmin.

Now the test toggles between a rodsuser and groupadmin.

lib/api/include/irods/switch_user.h
korydraughn: done

plugins/api/include/irods/plugins/api/switch_user_types.h
korydraughn: done

lib/api/include/irods/switch_user.h
korydraughn: Updated.

See switch_user_types.h for new words.

lib/api/include/irods/switch_user.h
korydraughn: Updated, but went with a different phrase entirely. Willing to change to one of your suggestions if you feel strongly about it or the new words aren't clear enough.

See switch_user_types.h for new words.

unit_tests/src/test_rc_switch_user.cpp
```diff
@@ -13,6 +13,7 @@
 #include "irods/user_administration.hpp"
 #include "irods/zone_administration.hpp"
 
+#include <array>
```
korydraughn, line 16: Remove. Not used.

korydraughn: done

lib/api/include/irods/switch_user.h
trel: No worries. Thanks.

plugins/api/include/irods/plugins/api/switch_user_types.h
```diff
@@ -1,15 +1,63 @@
...
+    // NOLINTBEGIN(cppcoreguidelines-avoid-magic-numbers, readability-magic-numbers)
+
+    /// The part of a fully-qualified iRODS username preceding the pound sign.
+    ///
+    /// The following requirements must be satisfied:
+    /// - The length must not exceed sizeof(UserInfo::userName) - 1 bytes
+    /// - It must be null-terminated
+    /// - It must be non-empty
+    /// - It must not include the pound sign
+    /// - It must not include the zone
```
alanking, line 19: Is this true? `tempZone_user` cannot exist in zone `tempZone`? Or did I read this wrong?

plugins/api/include/irods/plugins/api/switch_user_types.h
```diff
@@ -1,15 +1,63 @@
...
+
+    /// Influences how the server handles the proxy user.
+    ///
+    /// If set to 1, the proxy user is updated. This option will result in losing the
+    /// ability to invoke rc_switch_user if this option is set to 1 and the user to
+    /// switch to is not a \p rodsadmin. This option is not used when handling
+    /// server-to-server connections.
+    ///
+    /// \since 4.3.1
+    int update_proxy_user;
```
alanking, line 46: I feel like a word other than "update" may be better here... Does `switch_proxy_user` seem okay? If "update" is clear to everyone, I think it's fine. Just felt a little vague to me.

alanking (commented):
Looks good. Just had a couple of questions which may not need addressing

plugins/api/include/irods/plugins/api/switch_user_types.h
korydraughn: Your confusion indicates this wording isn't clear.

It is suppose to express that passing a string such as `"alice#tempZone"` is not allowed. Instead, `alice` and `tempZone` must be passed via the `username` and `zone` member variables.

plugins/api/include/irods/plugins/api/switch_user_types.h
korydraughn: I'm fine with either, but I prefer the one that is the clearest.

Given the name of the API uses `switch`, it seems more correct to use `switch` here as well.

Are there other words/phrases that are better?

korydraughn:
I also need to make this API plugin clear any tickets associated with the agent.

Will handle that soonish.

plugins/api/include/irods/plugins/api/switch_user_types.h
alanking: Ohhh, I see. I think this is fine. Not sure why I read it that way...

I don't have any suggestions for how to improve it.

korydraughn: An example for `rc_switch_user` would help here. Will add one.

unit_tests/src/test_process_stash.cpp
```diff
@@ -65,3 +65,38 @@ TEST_CASE("process_stash basic operations")
...
+    // Show that the size of the stash is equivalent to the number of handles.
+    auto handles = irods::process_stash::handles();
+    REQUIRE(handles.size() == 3);
+
+    // Show that the handles can be erased using a predicate.
+    irods::process_stash::erase_if([&h1, &h3](auto&& _k, auto&&) {
+        return _k == h1 || _k == h3;
+    });
+    CHECK_FALSE(irods::process_stash::find(h1));
+    CHECK_FALSE(irods::process_stash::find(h3));
```
alanking, line 98: Can we round this out? :)
```c++
    CHECK(irods::process_stash::find(h2));
```

alanking (commented):
Looks good

unit_tests/src/test_process_stash.cpp
korydraughn: Done.

plugins/api/include/irods/plugins/api/switch_user_types.h
korydraughn: Changed to `switch_proxy_user`.

plugins/api/src/switch_user.cpp
```diff
@@ -171,6 +209,15 @@ namespace
...
             }
 
+            // Verify there are no open L1 descriptors. Return an error if there are. This forces the
+            // client to clean up behind themselves and keeps this API plugin from becoming overly complicated.
+            for (const auto& l1d : std::span{L1desc}) {
+                if (FD_INUSE == l1d.inuseFlag) {
+                    log_api::error("Switching to another user is not allowed while there are open L1 descriptors.");
+                    return SYS_NOT_ALLOWED;
+                }
+            }
```
korydraughn, line 219: Consider making this an option.

Behavior of option:
- Return SYS_NOT_ALLOWED if there are open l1 descriptors
- Or, close open l1 descriptors

Proposed names for option:
- close_all_open_file_descriptors
- close_any_open_file_descriptors
- close_any_open_replicas (second/third choice)
- close_all_open_replicas (second/third choice)
- close_open_file_descriptors
- close_open_l1_descriptors
- **close_open_replicas** (winner?)
- torch_open_l1_descriptors
- l1_descriptor_behavior
- open_l1_descriptor_behavior
- on_open_l1_descriptor

Should all options be defined via a key-value pair/condInput instead? If yes, then the names of the properties can be ...
- switch_proxy_user => `KW_SWITCH_PROXY_USER`
  - Defaults to No.
- close_svr_to_svr_connections => `KW_CLOSE_SVR_TO_SVR_CONNECTIONS`
  - Defaults to Yes.
- close_open_replicas => `KW_CLOSE_OPEN_REPLICAS`
  - Defaults to No.

korydraughn: Done.

lib/api/include/irods/switch_user.h
```diff
@@ -17,37 +19,58 @@ extern "C" {
...
+///
+/// // (optional step)
+/// // Switch the identity of the proxy user to bob#tempZone too.
+/// // Without this, alice#tempZone would be acting as the proxy for bob#tempZone.
+/// input.switch_proxy_user = 1;
+///
+/// // (optional step)
+/// // Closing server-to-server connections isn't required, but is recommended to
+/// // cut down on the number of resources used by agents.
+/// input.close_svr_to_svr_connections = 1;
```
korydraughn, line 60: Update to use `options` key value pair / condInput.

plugins/api/include/irods/plugins/api/switch_user_types.h
```diff
@@ -1,15 +1,82 @@
...
     char zone[64];
-    // NOLINTEND(modernize-avoid-c-arrays)
+
+    // NOLINTEND(cppcoreguidelines-avoid-magic-numbers, readability-magic-numbers)
+    // NOLINTEND(cppcoreguidelines-avoid-c-arrays, modernize-avoid-c-arrays)
+
+    /// TODO
+    ///
+    /// \since 4.3.1
+    struct KeyValPair options;
```
korydraughn, line 44: `options` seems like a much better name than `condInput` to me. However, `condInput` is a common property name used across many APIs.

I think `options` does a better job at expressing the intent of the member variable.

Thoughts anyone?

plugins/api/include/irods/plugins/api/switch_user_types.h
```diff
@@ -1,15 +1,82 @@
...
+/// TODO
+/// Influences how the server handles the proxy user.
+///
+/// If set to 1, the proxy user is switched too. This option will result in
+/// losing the ability to invoke rc_switch_user if this option is set to 1 and
+/// the user to switch to is not a \p rodsadmin. This option is not used when
+/// handling server-to-server connections.
+///
+/// \since 4.3.1
+#define KW_SWITCH_PROXY_USER            "switch_proxy_user"
```
korydraughn, line 63: Should these keywords be moved to the rodsKeyWdDef.h header?

They are designed specifically for this API, but there's no reason they can't be used with others.

Thoughts anyone?

plugins/api/src/switch_user.cpp
```diff
@@ -159,7 +188,24 @@ namespace
...
+    {
+        log_api::trace("Clearing enabled ticket session.");
+
+        TicketAdminInput input{};
+        input.arg1 = "session";
+        input.arg2 = "";
+        input.arg3 = "";
+        //input.arg4 = "";
+        //input.arg5 = "";
+        //input.arg6 = "";
```
korydraughn, line 201: Remove

plugins/api/src/switch_user.cpp
```diff
@@ -171,6 +217,26 @@ namespace
                 return SYS_PROXYUSER_NO_PRIV;
             }
 
+            const irods::experimental::key_value_proxy opts{_input->options};
+
+            {
+                const auto iter = opts.find(KW_CLOSE_OPEN_REPLICAS);
+
+                if (std::end(opts) != iter && "1" == (*iter).value()) {
```
korydraughn, line 225: Just check for existence of the keyword and be done.

unit_tests/src/test_rc_switch_user.cpp
```diff
@@ -282,9 +309,216 @@ TEST_CASE("rc_switch_user cannot be invoked by non-admins")
...
+    // Create a second rodsuser.
+    const adm::user test_user_2{"rc_switch_user_test_user_2"};
+
+    REQUIRE_NOTHROW(adm::client::add_user(conn, test_user_2));
+
+    irods::at_scope_exit remove_user_2{
+        [&conn, &test_user_2] { CHECK_NOTHROW(adm::client::remove_user(conn, test_user_2)); }};
+
+    REQUIRE_NOTHROW(adm::client::modify_user(conn, test_user_2, password_prop));
+#endif
```
korydraughn, line 479: Remove once #7192 is resolved.

unit_tests/src/test_rc_switch_user.cpp
```diff
@@ -282,9 +309,216 @@ TEST_CASE("rc_switch_user cannot be invoked by non-admins")
...
```
korydraughn, line 513: Enable once #7192 is resolved.

unit_tests/src/test_rc_switch_user.cpp
```diff
@@ -282,9 +309,216 @@ TEST_CASE("rc_switch_user cannot be invoked by non-admins")
...
+        input.arg3 = "";
+
+        REQUIRE(rcTicketAdmin(static_cast<RcComm*>(test_user_conn), &input) == 0);
+        CHECK(fs::client::exists(test_user_conn, env.rodsHome));
+    }
+
+    // FIXME Does modifying the password of the second user fail because of the additional client connection
+    // updating the session signature?
+    //
+    // I bet it does.
```
korydraughn, line 502: Yes, this has been confirmed.

plugins/api/include/irods/plugins/api/switch_user_types.h
```diff
@@ -1,15 +1,82 @@
...
+/// Influences how the server handles server-to-server connections.
+///
+/// If set to 1, all connections created through redirection will be closed.
+/// If set to 0, rc_switch_user is invoked on all server-to-server connections.
+///
+/// If the server-to-server connection represents a connection to an iRODS
+/// server earlier than 4.3.1, the connection is closed.
+///
+/// \since 4.3.1
+#define KW_CLOSE_SVR_TO_SVR_CONNECTIONS "close_svr_to_svr_connections"
```
korydraughn, line 80: How about we invert the meaning of this and rename it to one of the following ...
- `KW_KEEP_SVR_TO_SVR_CONNECTIONS`
- `KW_DO_NOT_CLOSE_SVR_TO_SVR_CONNECTIONS`

Are there other phrases?

Doing this allows the API plugin to react based on the existence of the keyword rather than its value.

plugins/api/include/irods/plugins/api/switch_user_types.h
trel: it's the name of the structure as well, right?   holding the bag of keywords?

plugins/api/include/irods/plugins/api/switch_user_types.h
trel: i like the idea of allowing 'just existence'... 

- `KW_KEEP_SVR_TO_SVR_CONNECTIONS`
- `KW_RETAIN_SVR_TO_SVR_CONNECTIONS`

hm.

plugins/api/include/irods/plugins/api/switch_user_types.h
trel: defer

plugins/api/include/irods/plugins/api/switch_user_types.h
trel: since this does the same as all the other things, `condInput` feels normal, even if we'd never choose that today, anew.

korydraughn: Right. Usage would look like ...
```cpp
SwitchUserInput input{};
addKeyVal(&input.options, KEY_1, "");

// OR ...
irods::experimental::key_value_proxy kvp{input.options};
kvp[KEY_2] = "";
```

plugins/api/include/irods/plugins/api/switch_user_types.h
korydraughn: Either of those work and simplifies the API.

Thinking of going with `KEEP` only because it's shorter. Do you feel `RETAIN` is clearer in this case?

This keyword will cause the server to invoke `rc_switch_user` on all svr-to-svr connections if possible.

trel: `KEEP` is fine/great.

plugins/api/include/irods/plugins/api/switch_user_types.h
korydraughn: These keywords belong in rodsKeyWdDef.h.

The documentation for each keyword will become part of the `options` member variable's documentation.

lib/api/include/irods/switch_user.h
korydraughn: done

plugins/api/include/irods/plugins/api/switch_user_types.h
```diff
@@ -1,15 +1,82 @@
...
+    /// - It must not include the username
+    ///
+    /// \since 4.3.1
     char zone[64];
-    // NOLINTEND(modernize-avoid-c-arrays)
+
+    // NOLINTEND(cppcoreguidelines-avoid-magic-numbers, readability-magic-numbers)
+    // NOLINTEND(cppcoreguidelines-avoid-c-arrays, modernize-avoid-c-arrays)
+
+    /// TODO
```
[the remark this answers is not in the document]
korydraughn: done

plugins/api/include/irods/plugins/api/switch_user_types.h
```diff
@@ -1,15 +1,82 @@
...
 
 // NOLINTNEXTLINE(cppcoreguidelines-macro-usage)
-#define SwitchUserInp_PI "str username[64]; str zone[64];"
+#define SwitchUserInp_PI "str username[64]; str zone[64]; struct KeyValPair_PI;"
+
+//
+// Options
+//
+
+/// TODO
```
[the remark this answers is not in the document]
korydraughn: done

plugins/api/include/irods/plugins/api/switch_user_types.h
```diff
@@ -1,15 +1,82 @@
...
+///
+/// If set to 1, the proxy user is switched too. This option will result in
+/// losing the ability to invoke rc_switch_user if this option is set to 1 and
+/// the user to switch to is not a \p rodsadmin. This option is not used when
+/// handling server-to-server connections.
+///
+/// \since 4.3.1
+#define KW_SWITCH_PROXY_USER            "switch_proxy_user"
+
+/// TODO
```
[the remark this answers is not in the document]
korydraughn: done

plugins/api/include/irods/plugins/api/switch_user_types.h
```diff
@@ -1,15 +1,82 @@
...
+///
+/// \since 4.3.1
+#define KW_CLOSE_OPEN_REPLICAS          "close_open_replicas"
+
+/// TODO
```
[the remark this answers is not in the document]
korydraughn: done

plugins/api/include/irods/plugins/api/switch_user_types.h
korydraughn: done

plugins/api/src/switch_user.cpp
korydraughn: done

plugins/api/src/switch_user.cpp
korydraughn: done

plugins/api/include/irods/plugins/api/switch_user_types.h
korydraughn: done

lib/core/include/irods/process_stash.hpp
```diff
@@ -52,6 +53,16 @@ namespace irods::process_stash
...
 
+    /// Removes all entries satisfying the predicate.
+    ///
+    /// This function is thread-safe.
+    ///
+    /// \param[in] _pred The predicate to test each entry against. If \p _pred returns \p true,
+    ///                  the entry will be removed.
+    ///
+    /// \since 4.3.1
+    auto erase_if(const std::function<bool(const std::string&, const boost::any&)>& _pred) -> void;
```
korydraughn, line 64: Document parameters of predicate.

unit_tests/src/test_rc_switch_user.cpp
korydraughn: Removed.

unit_tests/src/test_rc_switch_user.cpp
korydraughn: Removed.

unit_tests/src/test_rc_switch_user.cpp
korydraughn: Restored and confirmed to work.

Solution: Modified the user administration library by replacing the use of `obfEncodeByKeyV2` with `obfEncodeByKey`. This change aligns with how `iadmin` works.

lib/administration/user/include/irods/user_administration.hpp
```diff
@@ -74,7 +74,14 @@ namespace irods::experimental::administration
     /// \since 4.3.1
     struct user_password_property
     {
-        std::string value;
+        std::string value;              ///< The new password.
+        std::optional<std::string> key; ///< The password of the user changing the password.
+
+        explicit user_password_property(std::string _value, std::optional<std::string> _key = std::nullopt)
```
korydraughn, line 80: Document

plugins/auth/src/native.cpp
```diff
@@ -85,6 +87,26 @@ namespace irods
...
+                          (unsigned char) md5_buf[6],
+                          (unsigned char) md5_buf[7],
+                          (unsigned char) md5_buf[8],
+                          (unsigned char) md5_buf[9],
+                          (unsigned char) md5_buf[10],
+                          (unsigned char) md5_buf[11],
+                          (unsigned char) md5_buf[12],
+                          (unsigned char) md5_buf[13],
+                          (unsigned char) md5_buf[14],
+                          (unsigned char) md5_buf[15]);
```
korydraughn, line 109: Put in a function.

plugins/auth_legacy/src/native.cpp
```diff
@@ -94,6 +95,27 @@ irods::error native_auth_establish_context(irods::plugin_context& _ctx)
...
+                  (unsigned char) md5_buf[6],
+                  (unsigned char) md5_buf[7],
+                  (unsigned char) md5_buf[8],
+                  (unsigned char) md5_buf[9],
+                  (unsigned char) md5_buf[10],
+                  (unsigned char) md5_buf[11],
+                  (unsigned char) md5_buf[12],
+                  (unsigned char) md5_buf[13],
+                  (unsigned char) md5_buf[14],
+                  (unsigned char) md5_buf[15]);
```
korydraughn, line 118: See https://github.com/irods/irods/pull/7178/files#r1271309378

plugins/auth/src/native.cpp
korydraughn: Setting the session signature also happens in `clientLoginWithPassword()`.

unit_tests/src/test_rc_switch_user.cpp
trel: Why does V2 exist?   Presumably it was added for a good reason?

korydraughn: I haven't investigated why it exists, but it is only used in iadmin, in one place. Nothing in the server uses it.

It seems to exist as a stronger version of the original, with no obvious insight into why it was introduced.

lib/administration/user/include/irods/user_administration.hpp
korydraughn: Unnecessary.

The constructor exists to enforce proper usage which is transparent to all readers of the type definition.

korydraughn:
Add server-side unit/behavior test which demonstrates the password modification logic in the user administration library works from within the server.

unit_tests/src/test_rc_switch_user.cpp
trel: https://github.com/irods/irods-legacy/commit/0cb144cbd4c9bc57b31f834ab38bc8eb73b72ab3

korydraughn: Nice. V2 presents a problem though. It's not clear to me if there's a session signature available on the server-side. I certainly couldn't find anything in the codebase.

Without that, native auth password management doesn't seem doable from within the server.

korydraughn:
> Add server-side unit/behavior test which demonstrates the password modification logic in the user administration library works from within the server.

Done.

lib/core/include/irods/process_stash.hpp
korydraughn: Done

plugins/microservices/src/test_user_administration.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+#include <exception>
+#include <functional>
+#include <string>
+
+namespace
+{
+    auto msi_impl(RuleExecInfo* _rei) -> int
+    {
+        IRODS_MSI_TEST_BEGIN("#NNNN: user administration library can change passwords inside the server")
```
korydraughn, line 25: Replace NNNN

unit_tests/src/test_user_administration.cpp
```diff
@@ -184,4 +185,41 @@ TEST_CASE("user group administration")
             adm::group{"unit_test_test_group_c"}
         });
     }
+
+    SECTION("#NNNN: password can be changed without obfGetPw")
```
korydraughn, line 189: Replace NNNN

plugins/auth_legacy/src/native.cpp
korydraughn: Done

plugins/auth/src/native.cpp
korydraughn: Done

plugins/microservices/src/test_user_administration.cpp
korydraughn: Use #7208

unit_tests/src/test_user_administration.cpp
korydraughn: Use #7208

lib/api/include/irods/check_auth_credentials.h
```diff
@@ -0,0 +1,137 @@
...
+    /// - It must be non-empty
+    ///
+    /// \since 4.3.1
+    char password[250];
+
+    // NOLINTEND(cppcoreguidelines-avoid-magic-numbers, readability-magic-numbers)
+    // NOLINTNEND(cppcoreguidelines-avoid-c-arrays, modernize-avoid-c-arrays)
+
+    /// TODO Probably not needed.
+    //char* auth_scheme;
```
korydraughn, line 54: Consider removing for now.

server/icat/include/irods/icatHighLevelRoutines.hpp
```diff
@@ -363,4 +363,21 @@ auto chl_get_delay_rule_info(RsComm& _comm, const char* _rule_id, std::vector<st
 /// \since 4.2.12
 auto chl_data_object_finalize(RsComm& _comm, const char* _json_input) -> int;
 
+/// \brief High-level wrapper for fetching all the information about a delay rule.
```
korydraughn, line 366: Incorrect description.

lib/api/include/irods/check_auth_credentials.h
```diff
@@ -0,0 +1,137 @@
...
+/// \file
+
+struct RcComm;
+
+/// The input data type used to invoke #rc_check_auth_credentials.
+///
+/// \since 4.3.1
+typedef struct CheckAuthCredentialsInput // NOLINT(modernize-use-using)
+{
+    // NOLINTBEGIN(cppcoreguidelines-avoid-c-arrays, modernize-avoid-c-arrays)
```
trel, line 13: maybe a note about how this is currently only for native

server/api/include/irods/rs_check_auth_credentials.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+/// \file
+
+#include "irods/check_auth_credentials.h"
+
+struct RsComm;
+
+/// Verifies the credentials passed match the credentials stored in the catalog.
+///
+/// This API endpoint only supports Native authentication and requires rodsadmin
```
trel, line 12: oh, there it is.

server/icat/include/irods/icatHighLevelRoutines.hpp
```diff
@@ -363,4 +363,21 @@ auto chl_get_delay_rule_info(RsComm& _comm, const char* _rule_id, std::vector<st
...
+/// \param[in] _zone     The part of a fully-qualified iRODS username following the pound sign.
+/// \param[in] _password The expected obfuscated password.
+///
+/// \return An integer.
+/// \retval 1                 If the obfuscated password matches what is stored in the catalog.
+/// \retval CAT_NO_ROWS_FOUND If no match was found in the catalog.
+/// \retval <0                If an error occurred.
+///
+/// \since 4.3.1
+auto chl_check_auth_credentials(RsComm& _comm, const char* _username, const char* _zone, const char* _password) -> int;
```
korydraughn, line 381: Consider taking an out parameter instead of using the return type.

While this works, the documentation here can be misleading because `CAT_NO_ROWS_FOUND` is a negative integer.

lib/api/include/irods/check_auth_credentials.h
korydraughn: The API endpoint's documentation captures that information.

lib/api/include/irods/check_auth_credentials.h
korydraughn: Removed

plugins/database/src/db_plugin.cpp
```diff
@@ -15599,6 +15599,57 @@ auto db_data_object_finalize_op(irods::plugin_context& _ctx, const char* _json_i
...
+            return ERROR(CAT_NO_ROWS_FOUND, std::move(msg));
+        }
+
+        return CODE(1);
+    }
+    catch (const std::exception& e) {
+        return ERROR(SYS_LIBRARY_ERROR, e.what());
+    }
+
+    return SUCCESS();
```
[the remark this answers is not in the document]
korydraughn: Removed

server/icat/include/irods/icatHighLevelRoutines.hpp
korydraughn: Fixed

server/icat/include/irods/icatHighLevelRoutines.hpp
korydraughn: Updated to take an additional parameter.

plugins/microservices/src/test_user_administration.cpp
korydraughn: Done

korydraughn:
I think this is finally in a good spot. Taking out of draft.

Will run clang-format once everyone agrees with the changes.

lib/api/include/irods/apiNumberData.h
alanking, line 1: What was the motivation for not adding this API as a plugin?

lib/api/include/irods/apiTable.hpp
```diff
@@ -1251,6 +1253,13 @@ static irods::apidef_t client_api_table_inp[] = {
         "api_exec_rule_expression", clearExecRuleExpressionInput, irods::clearOutStruct_noop,
         (funcPtr)CALL_EXECRULEEXPRESSIONINP
     },
+    {
+        CHECK_AUTH_CREDENTIALS_AN, RODS_API_VERSION, LOCAL_PRIV_USER_AUTH, LOCAL_PRIV_USER_AUTH,
+        "CheckAuthCredentialsInput_PI", 0,  "CheckAuthCredentialsOutput_PI", 0,
+        boost::any(std::function<int(rsComm_t*, checkAuthCredsInp_t*, checkAuthCredsOut_t**)>(RS_CHECK_AUTH_CREDENTIALS)),
+        "api_exec_rule_expression", irods::clearInStruct_noop, irods::clearOutStruct_noop,
```
alanking, line 1260:
```suggestion
        "api_check_auth_credentials", irods::clearInStruct_noop, irods::clearOutStruct_noop,
```

lib/api/include/irods/check_auth_credentials.h
```diff
@@ -0,0 +1,134 @@
...
+    /// - It must not include the pound sign
+    /// - It must not include the username
+    ///
+    /// \since 4.3.1
+    char zone[64];
+
+    /// The expected obfuscated password for the user of interest.
+    ///
+    /// The following requirements must be satisfied:
+    /// - The length must not exceed 250 - 1 bytes
```
alanking, line 43: I know we're ignoring magic numbers in this section, but why 250?

lib/api/include/irods/check_auth_credentials.h
```diff
@@ -0,0 +1,134 @@
...
+///
+/// \since 4.3.1
+typedef struct CheckAuthCredentialsOutput // NOLINT(modernize-use-using)
+{
+    /// An integer indicating whether the credentials for a particular user are correct
+    /// or not.
+    ///
+    /// If the credentials are correct, this member will be set to 1. Otherwise, it is
+    /// set to 0.
+    int would_succeed;
```
alanking, line 67: A couple of alternative name suggestions...
 - credentials_correct
 - auth_would_succeed

lib/api/include/irods/check_auth_credentials.h
```diff
@@ -0,0 +1,134 @@
...
+/// char* scrambled_password = // expected_password, but scrambled. See #obfEncodeByKey.
+/// strncpy(input.password, scrambled_password, sizeof(input.password) - 1);
+///
+/// struct CheckAuthCredentialsOutput* output = NULL;
+///
+/// // Verify if the given password for bob is correct.
+/// const int ec = rc_check_auth_credentials(comm, &input, &output);
+///
+/// if (ec < 0) {
+///     // Handle error.
```
alanking, line 113: Is it possible for `output` to be allocated but an error has occurred? It looks like it is not, but maybe worth mentioning somewhere?

lib/core/include/irods/rcConnect.h
```diff
@@ -35,6 +23,8 @@ struct thread_context;
 
 #define RECONN_TIMEOUT_TIME  600   /* re-connection timeout time in sec */
 
+// Forard declaration of thread context.
```
alanking, line 26:
```suggestion
// Forward declaration of thread context.
```

lib/core/include/irods/rcConnect.h
```diff
@@ -239,8 +231,10 @@ clientLoginPam( rcComm_t *conn, char *password, int ttl );
 int
 clientLoginTTL( rcComm_t *conn, int ttl );
 
-char *
-getSessionSignatureClientside();
+__attribute__((deprecated("Use session_signature member variable in RcComm instance.")))
```
alanking, line 234: Should we mention `set_session_signature_client_side` here?

lib/core/src/clientLogin.cpp
```diff
@@ -479,7 +504,7 @@ clientLoginWithPassword( rcComm_t *Conn, char* password ) {
     memset( md5Buf, 0, sizeof( md5Buf ) );
     strncpy( md5Buf, authReqOut->challenge, CHALLENGE_LEN );
     setSessionSignatureClientside( md5Buf );
-
+    set_session_signature_client_side(Conn, md5Buf);
```
alanking, line 507: Perhaps the documentation around this function would be enough, but maybe we should leave a comment to remind our future selves why we are calling both `setSessionSignatureClientside` and `set_session_signature_client_side`?

plugins/api/include/irods/plugins/api/switch_user_types.h
```diff
@@ -1,15 +1,81 @@
...
+    ///
+    ///     This option is never applied to server-to-server connections.
+    ///
+    /// - \p KW_CLOSE_OPEN_REPLICAS
+    ///
+    ///     Influences how #rc_switch_user handles open replicas.
+    ///
+    ///     If set, #rc_switch_user will attempt to close any open replicas before switching
+    ///     the identity associated with the connection. Otherwise, an error is returned if
+    ///     #rc_switch_user detects an open replica.
```
alanking, line 62: Perhaps worth mentioning that replicas opened for write will be marked stale as a result?

plugins/database/src/db_plugin.cpp
```diff
@@ -15599,6 +15599,60 @@ auto db_data_object_finalize_op(irods::plugin_context& _ctx, const char* _json_i
...
+        }
+        else {
+            *_would_succeed = 1;
+        }
+
+        return SUCCESS();
+    }
+    catch (const std::exception& e) {
+        return ERROR(SYS_LIBRARY_ERROR, e.what());
+    }
```
alanking, line 15653: Not that this would happen, but should we cover the `catch (...)` case as well?

plugins/database/src/db_plugin.cpp
```diff
@@ -15599,6 +15599,60 @@ auto db_data_object_finalize_op(irods::plugin_context& _ctx, const char* _json_i
...
+
+    if (!_username || !_zone || !_password || !_would_succeed) {
+        log_db::error("{}: Received one or more null pointers.", __func__);
+        return ERROR(SYS_INVALID_INPUT_PARAM, "Received one or more null pointers.");
+    }
+
+    *_would_succeed = -1; // Indicates the correctness of the creds is unknown.
+
+    std::array<char, MAX_PASSWORD_LEN + 20> decoded_password{};
```
alanking, line 15619: +20 for that hard-coded string?

scripts/irods/test/test_misc.py
```diff
@@ -337,6 +337,10 @@ def test_scoped_client_identity_updates_zone_of_RsComm__issue_6268(self):
     def test_scoped_permission__issue_7032(self):
         self.admin.assert_icommand(['irule', '-r', 'irods_rule_engine_plugin-irods_rule_language-instance', 'msi_test_scoped_permission', 'null', 'ruleExecOut'])
 
+    @unittest.skipUnless(plugin_name == 'irods_rule_engine_plugin-irods_rule_language', "Not implemented for other REPs.")
+    def test_user_administration__issue_NNNN(self):
```
alanking, line 341: Issue number?

server/api/include/irods/rs_check_auth_credentials.hpp
```diff
@@ -0,0 +1,63 @@
...
+/// char* scrambled_password = // expected_password, but scrambled. See #obfEncodeByKey.
+/// strncpy(input.password, scrambled_password, sizeof(input.password) - 1);
+///
+/// struct CheckAuthCredentialsOutput* output = NULL;
+///
+/// // Verify if the given password for bob is correct.
+/// const int ec = rs_check_auth_credentials(comm, &input, &output);
+///
+/// if (ec < 0) {
+///     // Handle error.
```
alanking, line 46: Same comment here about mentioning the fact that `output` will only be allocated when `ec >= 0`.

Should one of these Doxygen comments refer to the other, or just have 2 copies? Not sure what our process is for this.

server/icat/include/irods/icatHighLevelRoutines.hpp
```diff
@@ -363,4 +363,27 @@ auto chl_get_delay_rule_info(RsComm& _comm, const char* _rule_id, std::vector<st
...
+/// \brief High-level wrapper for verifying if the native authentication credentials for a specific
+///        user are correct.
+///
+/// Triggers policy associated with database operations.
+///
+/// \param[in] _comm          The communication object.
+/// \param[in] _username      The part of a fully-qualified iRODS username preceding the pound sign.
+/// \param[in] _zone          The part of a fully-qualified iRODS username following the pound sign.
+/// \param[in] _password      The expected obfuscated password.
+/// \param[in] _would_succeed \parblock The integer that will hold whether the credentials are correct.
```
alanking, line 375: Should this be `\param[out]`?

alanking (commented):
Looks good overall!

lib/api/include/irods/apiNumberData.h
korydraughn: I just wanted to see what it took to add it to the system directly.

I also feel this is slightly better from a performance standpoint because we don't have to load a shared object into memory.

There's also one less plugin to think about.

lib/api/include/irods/apiTable.hpp
korydraughn: Good catch.

lib/api/include/irods/check_auth_credentials.h
korydraughn: That is the max size defined in the database for passwords.

I couldn't think of any other value to use here because it wasn't clear what the max size for an obfuscated password is.

lib/api/include/irods/check_auth_credentials.h
korydraughn: That is the intent. If an error occurs, `output` should be null.

Certainly worth mentioning in the documentation for the API.

lib/core/include/irods/rcConnect.h
korydraughn: It's not clear to me why we'd do that. Can you elaborate?

plugins/database/src/db_plugin.cpp
korydraughn: I always flip flop on that. I don't think it is necessary, but you never know.

Will add.

plugins/database/src/db_plugin.cpp
korydraughn: I wish the reason for the `+ N` was obvious so we can give it a name, but at this point, IDK if it's even worth it.

I'm on the side of perhaps we just hide all of this so no one ever has to see it.

Thoughts?

_Am I misunderstanding?_

scripts/irods/test/test_misc.py
korydraughn: Nice catch.

So easy to forget things.

server/api/include/irods/rs_check_auth_credentials.hpp
korydraughn: No process established yet. We could link the two, but it hasn't become an issue for me to want to investigate that yet. Especially when new API endpoints tend to not change very much once the dust settles.

Thoughts?

server/icat/include/irods/icatHighLevelRoutines.hpp
korydraughn: Yes. Will update.

lib/administration/user/include/irods/user_administration.hpp
```diff
@@ -74,7 +74,19 @@ namespace irods::experimental::administration
...
+            , key{std::move(_key)}
+        {
+        }
+
+        // NOLINTBEGIN(misc-non-private-member-variables-in-classes)
+        std::string value; ///< The new password.
+
+        /// The password of the user requesting the change.
+        /// If not set, obfGetPw() is used. Use of obfGetPw() implies a .irodsA file exists.
+        std::optional<std::string> key;
```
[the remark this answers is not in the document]
korydraughn: `key` and `value` are somewhat related.

I chose `key` because I couldn't come up with anything else at the time. It is the caller's deobfuscated password. It exists for cases where calling `obfGetPw` is impossible and/or not preferred.

I didn't want to put what it actually was, but the comment implies what is happening for anyone familiar with the code.

Alternative naming welcome.

lib/api/include/irods/check_auth_credentials.h
korydraughn: Thinking about how all of these would read in code.
```cpp
checkAuthCredsOut_t* output;

if (1 == output->would_succeed)       { ... }
if (1 == output->credentials_correct) { ... }
if (1 == output->auth_would_succeed)  { ... }
```

I can see any of these working, but the suggestions are starting to feel superior from a readability standpoint.

`credentials_correct` stand out the most.

Will think about this one some more. Are there other things we need to consider (for the future) that may affect the naming?

lib/api/include/irods/apiNumberData.h
korydraughn: Thoughts?

trel: see #7214

alanking: Okay. Adding APIs as plugins in the server has always confused me, so I'm in favor of moving away from that, in addition to the potential performance gains. This should especially be considered for `data_object_finalize`, `register_physical_path`, etc.

lib/administration/user/include/irods/user_administration.hpp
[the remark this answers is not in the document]
alanking: Some ideas...
 - `requesting_user_password`
 - `credentials`
 - `caller_credentials`
 - `requesting_user_credentials`

I just feel like `key` forms a relationship to `value` which doesn't necessarily exist. It's not a big deal, it just confused me when I saw it.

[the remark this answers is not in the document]
korydraughn: Totally understand. Will change.

plugins/database/src/db_plugin.cpp
korydraughn: Yep, I think I misunderstood the question.

You're asking if the +20 is for the hard-coded string known by the server and the clients, correct?

alanking: Oh, yes, sorry. That was a real question because I wanted to make sure I understood what it meant. A comment or something would of course be helpful, but that was mainly for my own understanding.

korydraughn: I really don't know what the +20 is for.

If it helps, I pretty much copied what I saw the existing functions doing when handling passwords.

For now, we can create an issue to investigate the hard-coded integers and handle it later.

lib/api/include/irods/check_auth_credentials.h
korydraughn: Here's one more alternative.
```cpp
checkAuthCredsOut_t* output;

if (IRODS_CREDS_CORRECT   == output->result) { ... }
if (IRODS_CREDS_INCORRECT == output->result) { ... }
if (IRODS_CREDS_NOT_SET   == output->result) { ... }
```

Thoughts?

trel: I don't like shortened 'CREDS' - we can use all the letters, if we choose this option.

One vote for 'credentials_correct'.

korydraughn: Discush!

lib/api/include/irods/apiTable.hpp
korydraughn: Done

lib/api/include/irods/check_auth_credentials.h
korydraughn: Done

scripts/irods/test/test_misc.py
korydraughn: Done

server/icat/include/irods/icatHighLevelRoutines.hpp
korydraughn: Done

plugins/database/src/db_plugin.cpp
korydraughn: Done

lib/core/include/irods/rcConnect.h
korydraughn: Done

plugins/api/include/irods/plugins/api/switch_user_types.h
korydraughn: Done

lib/core/src/clientLogin.cpp
korydraughn: Done

lib/api/include/irods/check_auth_credentials.h
korydraughn: Done.

server/api/include/irods/rs_check_auth_credentials.hpp
korydraughn: There is one small difference between the two and that's the wording for the `RxComm` parameter. There's also the case of example code appearing in the documentation. Dealing with the parameter is simple, but dealing with the example will take a little thought.

We should investigate how to do that, but that can happen at any time.

Marking this resolved. Consider opening an issue to investigate that so we don't forget.

plugins/database/src/db_plugin.cpp
korydraughn: Created issue for this. See #7224.

lib/administration/user/include/irods/user_administration.hpp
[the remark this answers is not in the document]
korydraughn: Changed to `requester_password`.

I considered changing it to `requesting_user_password`, but that's a lot of words lol and `requester` implies `user` to an extent.

lib/api/include/irods/check_auth_credentials.h
alanking: Okay, mostly just curious. This can be resolved.

lib/administration/user/include/irods/user_administration.hpp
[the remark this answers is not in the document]
alanking: I like it

lib/core/include/irods/rcConnect.h
alanking: Oh, that says "get" not "set". Nevermind.

lib/api/include/irods/check_auth_credentials.h
```diff
@@ -0,0 +1,131 @@
...
+///
+/// It does NOT modify any state on either side of the connection.
+///
+/// \param[in] _comm    A pointer to a RcComm.
+/// \param[in] _input   A pointer to a CheckAuthCredentialsInput.
+/// \param[in] _correct \parblock A pointer to an integer pointer used to indicate
+/// whether the native authentication credentials are correct.
+///
+/// On success, the pointer will point to memory on the heap. The caller is
+/// responsible for freeing the memory.
```
alanking, line 75: What is the reason we aren't just using an `int*` here? I don't remember if this was discussed...

alanking (commented):
Everything looks good. I either didn't notice or didn't think about the `int**` before, so just had a question about that.

lib/api/include/irods/check_auth_credentials.h
korydraughn: Returning values from the server requires a pointer-to-pointer parameter.

I normally hide that detail, but I chose not to for this. Usage is identical to that of `rxDataObjChksum`.

lib/api/include/irods/check_auth_credentials.h
korydraughn: Resolving https://github.com/irods/irods/issues/6297 will help here. So there's a path forward.

lib/administration/user/include/irods/user_administration.hpp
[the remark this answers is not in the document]
korydraughn: Cool cool.

lib/api/include/irods/check_auth_credentials.h
korydraughn: Before these latest commits, I updated things to use an `int*`, but that didn't sit well with me because of PackStruct. While I did not run that code, I didn't have confidence things would work that way just based on what I knew about PackStruct.

Still, pointer-to-pointer parameters aren't anything unfamiliar to C devs.

korydraughn:
All unit tests pass.

Confirming the changes work for the HTTP API PR at https://github.com/irods/irods_client_http_api/pull/57 before running core tests.

lib/api/include/irods/check_auth_credentials.h
alanking: Ah, right. Sounds good.

korydraughn:
All unit tests pass.
All HTTP API python tests pass.

Running core tests.

alanking:
Go ahead and squash (no #s) when tests pass

korydraughn:
All core tests passed.

korydraughn:
Rebased and squashed.

korydraughn:
Working on clang-format and clang-tidy reports.

korydraughn:
Made some small tweaks due to clang-tidy report. See the commits which haven't been squashed yet.

Ignoring the majority of the clang-tidy report because some changes are due to having to support C headers. Some uses of `strcpy` are reported, but those calls are safe by design. Willing to go through with people if necessary.

With that said, all unit/core tests pass.

Will apply clang-format and squash if everyone agrees with the changes.

trel:
no red flags for me - leave to another for more eyeballs.

korydraughn:
Going to look through clang-tidy results one last time before wrapping this up.

alanking (commented):
Yep, looks good.

b840b1775e5f609dc7504a5efeb05cc4ac3c7f77 has "admininstration" in the commit title and message.

korydraughn:
There's nothing interesting happening in the new commits. They are focused on clang-tidy/format results.

I just ran all the unit tests and everything is in working order.

Squashing now.

korydraughn:
Squashed!

alanking (approved):
Please add #s.

korydraughn:
Added #'s

PR #7182, opened 2023-07-12
[7154] Change irods::file_object size to rodsLong_t (main)

Nishant2022:


Nishant2022:
All unit tests pass

Nishant2022:
Running them now

korydraughn (approved):
Great. Add the pound.

alanking (approved):
I took a look around and don't see many places where this is being used, let alone places where any resulting errors wouldn't be caught, so I think it is safe to merge.

Issue #7172, opened 2023-07-05
Investigate `test_control_plane` failing.


Nishant2022:
When running the core tests, `test_control_plane` fails, specifically the `test_shutdown_local_server` and `test_shutdown_local_server_using_hostname_alias_defined_in_server_config__issue_5700` functions. 

They fail because `lib.get_file_size_by_path(IrodsConfig().server_parent_log_path)` throws an error.

`test_control_plane` did pass in topology testing though. The two functions that are causing problems were not skipped and ran successfully.

SwooshyCueb:
I am currently running the python suite in a development environment Ubuntu 20.04 runner, and `test_control_plane` passed for me. The only test that was skipped was `test_empty_message` due to lack of `pyzmq`

korydraughn:
Agreed. I've also seen all of the reported tests pass.

Let's leave the issues around testing open until the non-testing issues are resolved.

Nishant2022:
Perhaps it's a resource limited test. I ran the tests on a RENCI VM with 4 cores and 16 GB of ram. I also believe there is a storage issue because the only tests that would fail were ones trying to access logs. However it may just be some quirk of the machines I used.

korydraughn:
At this point, we're confident this is likely due to the hardware. The test passes consistently on a large enough computer.

Closing.

PR #7184, opened 2023-07-12
[6866] Add bound checks for environment variables (main)

Nishant2022:
Use snprintf in capture_string_property and add size of string information

### DIFF
--- a/lib/core/src/getRodsEnv.cpp
+++ b/lib/core/src/getRodsEnv.cpp
@@ -24,6 +24,7 @@
   the env file, either creating it or appending to it.
 */
 
+#include "irods/rodsDef.h"
 #include "irods/rods.h"
 #include "irods/rodsErrorTable.h"
 #include "irods/getRodsEnv.h"
@@ -32,6 +33,8 @@
 #include "irods/irods_version.h"
 #include "irods/irods_environment_properties.hpp"
 #include "irods/irods_configuration_keywords.hpp"
+
+#include <cstring>
 
 #define BUF_LEN 100
 #define LARGE_BUF_LEN MAX_NAME_LEN+20
@@ -139,12 +142,19 @@ extern "C" {
         createRodsEnvDefaults( &rodsEnvArg );
     }
 
-    static
-    int capture_string_property(
-        const std::string&             _key,
-        char*                          _val ) {
+    static int capture_string_property(const std::string& _key, char* _val, std::size_t _val_size)
+    {
+        if (_val == nullptr || _val_size == 0) {
+            return SYS_GETENV_ERR;
+        }
         try {
-            sprintf(_val, "%s", irods::get_environment_property<const std::string>(_key).c_str());
+            auto property = irods::get_environment_property<const std::string>(_key);
+
+            // If truncation would occur because property.size() >= _val_size, return an error
+            if (property.size() >= _val_size) {
+                return SYS_GETENV_ERR;
+            }
+            std::strncpy(_val, property.c_str(), _val_size);
             return 0;
         } catch ( const irods::exception& e ) {
             if ( e.code() == KEY_NOT_FOUND ) {
@@ -156,6 +166,11 @@ extern "C" {
         }
 
     } // capture_string_property
+
+    int capture_string_property(const char* _key, char* _val, size_t _val_size)
+    {
+        return capture_string_property(std::string{_key}, _val, _val_size);
+    }
 
     static
     int capture_integer_property(
@@ -193,49 +208,41 @@ extern "C" {
             sizeof( _env->rodsAuthScheme ),
             "native" );
 
-        capture_string_property(
-            irods::KW_CFG_IRODS_SESSION_ENVIRONMENT_FILE,
-            configFileName );
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+        capture_string_property(irods::KW_CFG_IRODS_SESSION_ENVIRONMENT_FILE, configFileName, LONG_NAME_LEN);
 
-        capture_string_property(
-            irods::KW_CFG_IRODS_USER_NAME,
-            _env->rodsUserName );
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+        capture_string_property(irods::KW_CFG_IRODS_USER_NAME, _env->rodsUserName, NAME_LEN);
 
-        capture_string_property(
-            irods::KW_CFG_IRODS_HOST,
-            _env->rodsHost );
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+        capture_string_property(irods::KW_CFG_IRODS_HOST, _env->rodsHost, NAME_LEN);
 
-        capture_string_property(
-            irods::KW_CFG_IRODS_HOME,
-            _env->rodsHome );
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+        capture_string_property(irods::KW_CFG_IRODS_HOME, _env->rodsHome, MAX_NAME_LEN);
 
-        capture_string_property(
-            irods::KW_CFG_IRODS_CWD,
-            _env->rodsCwd );
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+        capture_string_property(irods::KW_CFG_IRODS_CWD, _env->rodsCwd, MAX_NAME_LEN);
 
-        capture_string_property(
-            irods::KW_CFG_IRODS_AUTHENTICATION_SCHEME,
-            _env->rodsAuthScheme );
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+        capture_string_property(irods::KW_CFG_IRODS_AUTHENTICATION_SCHEME, _env->rodsAuthScheme, NAME_LEN);
 
         capture_integer_property(
             irods::KW_CFG_IRODS_PORT,
             _env->rodsPort );
 
-        capture_string_property(
-            irods::KW_CFG_IRODS_DEFAULT_RESOURCE,
-            _env->rodsDefResource );
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+        capture_string_property(irods::KW_CFG_IRODS_DEFAULT_RESOURCE, _env->rodsDefResource, NAME_LEN);
 
-        capture_string_property(
-            irods::KW_CFG_IRODS_ZONE,
-            _env->rodsZone );
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+        capture_string_property(irods::KW_CFG_IRODS_ZONE, _env->rodsZone, NAME_LEN);
 
-        capture_string_property(
-            irods::KW_CFG_IRODS_CLIENT_SERVER_POLICY,
-            _env->rodsClientServerPolicy );
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+        capture_string_property(irods::KW_CFG_IRODS_CLIENT_SERVER_POLICY, _env->rodsClientServerPolicy, LONG_NAME_LEN);
 
-        capture_string_property(
-            irods::KW_CFG_IRODS_CLIENT_SERVER_NEGOTIATION,
-            _env->rodsClientServerNegotiation );
+        capture_string_property(irods::KW_CFG_IRODS_CLIENT_SERVER_NEGOTIATION,
+                                // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+                                _env->rodsClientServerNegotiation,
+                                LONG_NAME_LEN);
 
         capture_integer_property(
             irods::KW_CFG_IRODS_ENCRYPTION_KEY_SIZE,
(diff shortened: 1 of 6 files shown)


SwooshyCueb:
Is there a reason for using `snprintf` over `strncpy` or `strncpy_s`?

Nishant2022:
Both `strncpy` and `strncpy_s` do not guarantee that there will be a null terminator. If you provide a buffer size of `n` to `snprintf`, it will copy at most `n-1` characters and place the null terminator at the end.

lib/core/src/getRodsEnv.cpp
```diff
@@ -139,12 +140,11 @@ extern "C" {
...
-    static
-    int capture_string_property(
-        const std::string&             _key,
-        char*                          _val ) {
+    static int capture_string_property(const std::string& _key, char* _val, size_t _val_size)
+    {
         try {
-            sprintf(_val, "%s", irods::get_environment_property<const std::string>(_key).c_str());
+            // NOLINTNEXTLINE(cppcoreguidelines-pro-type-vararg)
+            snprintf(_val, _val_size, "%s", irods::get_environment_property<const std::string>(_key).c_str());
```
korydraughn, line 147: It's not only about avoiding buffer overflows. We need to make sure that the value of the environment variable is captured in full.

Consider what happens if the env var exceeds the size of `_val_size`.

korydraughn, line 147: Given that we're writing C++, please qualify C std functions with the `std::` namespace.

Make sure the correct header is included as well. I think it's `<cstring>`, but please verify.

SwooshyCueb: If my understanding is correct, and we have to check whether the destination buffer can contain the string plus null terminator either way, it still makes more sense to me to use `strncpy` or `strncpy_s` over `snprintf`. I'm willing to be proven wrong.

korydraughn: I also find the `strcpy` functions easier to use/understand, but I'm not opposed to `snprintf` assuming it's used correctly/safely.

Other than that, I agree we must check the results and make sure the stored value is what we read from the env.

Nishant2022: `_val` is a character array with a set size. I changed all calls to `capture_string_property` so that they pass in the size of that array which is defined in `getRodsEnv.h`. If a string is passed in and is larger than `_val_size` then it will also be larger than what `_val` can hold. In the past, this could lead to a buffer overflow but using `snprintf` prevents that. 

If you want to be able to capture the value of `_key` regardless of its size, then we would have to change how we store environment variables, opting for a string rather than a `char*`.

Nishant2022: I looked into the documentation for `strncpy_s`. If we were to provide a source string that is larger than the destination, it will set the destination to an empty string and return an error code.  You can set it to truncate like `snprintf` by passing `_TUNCATE` or a value one less than the size of the destination string.

If a null value is passed as an input, `strcpy_s` will return an error, while `snprintf` will copy "(null)" into the destination.

Which would we prefer?

korydraughn: `strncpy_s` sounds really good.

If the input string cannot be copied in full, that is an error. We must report it because we're now dealing with an invalid state.

I can't see a reason to ever allow the input string to be truncated here.

SwooshyCueb: > `strncpy_s` sounds really good. Is it part of the standard library or POSIX?

It's C11, according to cppreference.
https://en.cppreference.com/w/c/string/byte/strncpy

Nishant2022: After a bit of digging I found that `strncpy_s` is a microsoft specific function. We can manually check and use `strncpy` if that is preferable to `snprintf`.

Nishant2022: > > `strncpy_s` sounds really good. Is it part of the standard library or POSIX?
> 
> It's C11, according to cppreference. https://en.cppreference.com/w/c/string/byte/strncpy

According to the reference we need to define `__STDC_LIB_EXT1__`

alanking (commented):
Looks good to me. Any other thoughts, @SwooshyCueb or @korydraughn?

lib/core/src/getRodsEnv.cpp
```diff
@@ -139,12 +142,17 @@ extern "C" {
...
-        char*                          _val ) {
+    static int capture_string_property(const std::string& _key, char* _val, size_t _val_size)
+    {
         try {
-            sprintf(_val, "%s", irods::get_environment_property<const std::string>(_key).c_str());
+            auto property = irods::get_environment_property<const std::string>(_key);
+
+            // If _val is undefined, _val_size is zero, or truncation would
+            // occur because property.size() >= _val_size, return an error
+            if (!_val || !_val_size || property.size() >= _val_size) {
```
korydraughn, line 152: Consider doing the pointer and size checks for `_val` and `_val_size` before L148.

The check comparing the size of the property and `_val_size` can stay here.

korydraughn, line 152: For numerical comparisons, let's explicitly check for zero.
```cpp
if (!_val || 0 == _val_size) {
```

Readers unfamiliar with the codebase can easily misinterpret the `!_val_size` as a pointer check or boolean check.

lib/core/src/getRodsEnv.cpp
```diff
@@ -139,12 +142,17 @@ extern "C" {
         createRodsEnvDefaults( &rodsEnvArg );
     }
 
-    static
-    int capture_string_property(
-        const std::string&             _key,
-        char*                          _val ) {
+    static int capture_string_property(const std::string& _key, char* _val, size_t _val_size)
```
korydraughn, line 145: `size_t` -> `std::size_t`

Don't forget to include the appropriate header if not included already.

I think it is defined in `<cstdint>`. Check cppreference.com to be sure.

korydraughn (commented):
Let's add a unit test for `capture_string_property`.

lib/core/src/getRodsEnv.cpp
Nishant2022: `std::size_t` is also defined in `<cstring>`.

lib/core/src/getRodsEnv.cpp
```diff
@@ -139,12 +142,19 @@ extern "C" {
         createRodsEnvDefaults( &rodsEnvArg );
     }
 
-    static
-    int capture_string_property(
-        const std::string&             _key,
-        char*                          _val ) {
+    static int capture_string_property(const std::string& _key, char* _val, std::size_t _val_size)
```
SwooshyCueb, line 145: Since the function is used in a unit test and declared in a header file, it should probably not be defined as static, as that typically makes the function unavailable outside its translation unit.

SwooshyCueb:
LGTM! (Ignore the comment I made about `static`, I misread the diff)

SwooshyCueb (approved):

Nishant2022:
All unit tests pass.

All core tests pass, apart from `test_control_plane` which is documented in https://github.com/irods/irods/issues/7172 and `test_rulebase`, which seems to be related to the control plane issue as it fails when trying to collect logs using `lib.get_file_size_by_path`. 

I believe that this PR should still be able to move ahead, and the issue regarding `lib.get_file_size_by_path` should be investigated at a later time.

unit_tests/src/test_environment_variables.cpp
```diff
@@ -21,39 +22,54 @@ TEST_CASE("environment variables")
...
         REQUIRE(SYS_GETENV_ERR == capture_string_property(env_var_key.c_str(), mismatched_buffer, SMALL_BUFFER_SIZE));
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
         REQUIRE(strnlen(mismatched_buffer, SMALL_BUFFER_SIZE) == 0);
 
         // Assert it fails with zero buffer size:
         // If the given buffer size is zero, capture_string_property should
         // return a SYS_GET_ENV_ERR and the buffer should not be touched
         constexpr std::size_t ZERO_BUFFER_SIZE = 0;
+        // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays,modernize-avoid-c-arrays,-warnings-as-errors)
         char zero_buffer[SMALL_BUFFER_SIZE]{};
```
korydraughn: I don't think this is valid C++.

We need to investigate whether zero-length arrays are acceptable.

unit_tests/src/test_environment_variables.cpp
```diff
@@ -21,39 +22,54 @@ TEST_CASE("environment variables")
         // Assert the working case:
         // Given a large enough buffer, the environment variable will be coppied into the buffer
         constexpr std::size_t LARGE_BUFFER_SIZE = 100;
+        // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays,modernize-avoid-c-arrays,-warnings-as-errors)
         char large_buffer[LARGE_BUFFER_SIZE]{};
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
         REQUIRE(0 == capture_string_property(env_var_key.c_str(), large_buffer, LARGE_BUFFER_SIZE));
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
         REQUIRE(strnlen(large_buffer, LARGE_BUFFER_SIZE) == env_var_val.size());
```
korydraughn: Each one of these blocks could live in their own section.

Not required, but could improve the test by allowing users to pick which ones are run.

korydraughn:
> All unit tests pass.
> 
> All core tests pass, apart from `test_control_plane` which is documented in #7172 and `test_rulebase`, which seems to be related to the control plane issue as it fails when trying to collect logs using `lib.get_file_size_by_path`.
> 
> I believe that this PR should still be able to move ahead, and the issue regarding `lib.get_file_size_by_path` should be investigated at a later time.

Is the `test_rulebase` failure consistent or intermittent?

unit_tests/src/test_environment_variables.cpp
Nishant2022: I guess I wasn't really clear what I was doing there.

I was testing if you provided a size of zero, would the function throw an error?

the "`zero_buffer`" actually has a size of 10 as it is instantiated with the `SMALL_BUFFER_SIZE`.

korydraughn: I misread the code. I somehow thought the buffer was initialized with a size of zero.

I'll look over that one again and leave a comment on whether anything needs to be changed.

Nishant2022:
`test_rulebase` seems to be consistently failing for me across multiple branches including main. Perhaps it is an issue with my VM (storage, memory)?

korydraughn:
If you haven't, try freeing up some space and running that test by itself a few times. If it passes even once, then I think we blame your VM's resources (and the test for being weak against external factors).

unit_tests/src/test_environment_variables.cpp
```diff
@@ -0,0 +1,92 @@
+
```
korydraughn, line 1: Remove empty line.

unit_tests/src/test_environment_variables.cpp
```diff
@@ -0,0 +1,92 @@
...
+    {
+        // Assert the working case:
+        // Given a large enough buffer, the environment variable will be coppied into the buffer
+        // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays,modernize-avoid-c-arrays,-warnings-as-errors)
+        char large_buffer[LARGE_BUFFER_SIZE]{};
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+        REQUIRE(0 == capture_string_property(env_var_key.c_str(), large_buffer, LARGE_BUFFER_SIZE));
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+        REQUIRE(strnlen(large_buffer, LARGE_BUFFER_SIZE) == env_var_val.size());
+    }
```
korydraughn, line 35: Missing assertion that `large_buffer` is equal to `env_var_val`.

unit_tests/src/test_environment_variables.cpp
korydraughn: I think we're good here. Please mark as resolved.

unit_tests/src/test_environment_variables.cpp
```diff
@@ -0,0 +1,92 @@
...
+    {
+        // Assert it fails with small buffer:
+        // If the buffer is too small, capture_string_property should
+        // return a SYS_GET_ENV_ERR and the buffer should remain empty
+        // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays,modernize-avoid-c-arrays,-warnings-as-errors)
+        char small_buffer[SMALL_BUFFER_SIZE]{};
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+        REQUIRE(SMALL_BUFFER_SIZE < env_var_val.size());
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+        REQUIRE(SYS_GETENV_ERR == capture_string_property(env_var_key.c_str(), small_buffer, SMALL_BUFFER_SIZE));
```
alanking, line 47: I think the empty buffer assertion is missing?

unit_tests/src/test_environment_variables.cpp
```diff
@@ -0,0 +1,92 @@
...
+TEST_CASE("environment_variables")
+{
+    const std::string env_var_key{"ISSUE_6866_ENV_KEY"};
+    const std::string env_var_val{"ISSUE_6866_ENV_VAL"};
+
+    irods::set_environment_property(env_var_key, env_var_val);
+
+    // Buffer size constants
+    constexpr std::size_t ZERO_BUFFER_SIZE = 0;
+    constexpr std::size_t SMALL_BUFFER_SIZE = 10;
```
alanking, line 22: If this indicated that the buffer size is *too* small, that might be clearer (e.g. `TOOSMALL_BUFFER_SIZE`). Not required to change this since we are asserting that the size is smaller than the length of the value as part of the tests below - just a suggestion.

alanking (commented):
Nice

alanking (commented):
Looks good to me. Any other thoughts before squash, @korydraughn?

korydraughn (commented):
Code looks good.

Github is reporting conflicts with some of the files.

korydraughn:
There's a merge commit in there.

We need to remove that and resolve the conflicts a different way. Let us know if you want assistance on how to do that.

Nishant2022:
Yeah I accidentally did that. Got it sorted out now. I'm currently running the core tests and unit tests again.

Nishant2022:
All unit tests pass.

All core tests pass including `test_control_plane` and `test_rulebase`.

alanking (commented):
Let's squash it. Please omit #

Nishant2022:
Squashed commits

alanking (commented):
Please change commit message to say "std::strncpy" instead of "snprintf" and I think this is good for #.

Nishant2022:
Updated commit message and added #

alanking (approved):

korydraughn (approved):
Nice work.

PR #7187, opened 2023-07-14
[7154] Change irods::file_object size to rodsLong_t (4-3-stable)

Nishant2022:
Cherry picked from main

### DIFF
--- a/server/core/include/irods/irods_file_object.hpp
+++ b/server/core/include/irods/irods_file_object.hpp
@@ -5,6 +5,7 @@
 // irods includes
 #include "irods/fileCreate.h"
 #include "irods/rodsConnect.h"
+#include "irods/rodsType.h"
 
 // =-=-=-=-=-=-=-
 #include "irods/irods_data_object.hpp"
@@ -103,7 +104,7 @@ namespace irods {
             virtual int                            l1_desc_idx()     const {
                 return l1_desc_idx_;
             }
-            virtual size_t                         size()            const {
+            virtual rodsLong_t                         size()            const {
                 return size_;
             }
             virtual int                            repl_requested()  const {
@@ -134,7 +135,7 @@ namespace irods {
             virtual void comm( rsComm_t* _c )                   {
                 comm_            = _c;
             }
-            virtual void size( size_t _v )                       {
+            virtual void size( rodsLong_t _v )                       {
                 size_            = _v;
             }
             virtual void repl_requested( int _v )                {
@@ -167,7 +168,7 @@ namespace irods {
             std::string                    data_type_;       // data type as described in objInfo.h:32
             int                            file_descriptor_; // file descriptor, if the file is in flight
             int                            l1_desc_idx_;     // index into irods L1 file descriptor table
-            size_t                         size_;            // size of the file in bytes
+            rodsLong_t                     size_;            // size of the file in bytes
             int                            repl_requested_;  // requested replica number
             std::string                    in_pdmo_;         // hierarchy indicating the current operations are
             // occurring from within a pdmo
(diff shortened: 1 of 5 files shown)


korydraughn:
Let's add the unit test as a separate commit. That way it's easy to cherry-pick back to the main branch.

unit_tests/src/test_file_object.cpp
```diff
@@ -0,0 +1,16 @@
+#include <catch2/catch.hpp>
+
+#include "irods/irods_file_object.hpp"
+
+TEST_CASE("file_object")
+{
+    SECTION("test_file_object_size_can_be_negative__issue_7154")
```
korydraughn, line 7: The the C++ unit tests, we don't include underscores to separate words.

Compare your section title to other unit tests and try to make it align with how those are worded.

alanking:
The unit test you wrote is fine and good, but the unit test that I was proposing is less of a unit test in the traditional sense and more of a unit test in the iRODS sense. :) It's more like a C API test. Apologies...

We should try using `rcDataObjOpen` with a `DataObjInp::dataSize` of -1 to make sure that the server does not misbehave like it does when PRC passes a `dataSize` of -1.

unit_tests/src/test_file_object.cpp
```diff
@@ -2,9 +2,9 @@
 
 #include "irods/irods_file_object.hpp"
 
-TEST_CASE("file_object")
+TEST_CASE("file object")
 {
-    SECTION("test_file_object_size_can_be_negative__issue_7154")
+    SECTION("test file object size can be negative issue 7154")
```
korydraughn: My bad. I was referring to everything except the type.

`file object` -> `file_object`

unit_tests/src/test_file_object.cpp
```diff
@@ -0,0 +1,16 @@
...
+
+TEST_CASE("file_object")
+{
+    SECTION("test file object size can be negative issue 7154")
+    {
+        irods::file_object obj{};
+        REQUIRE(obj.size() == 0);
+
+        obj.size(-1);
+        REQUIRE(obj.size() == -1);
```
alanking, line 13: Perhaps we should modify this or add another assertion that `obj.size() < 0` to make sure that there's no funny business with casting...?

server/core/include/irods/irods_file_object.hpp
```diff
@@ -103,7 +104,7 @@ namespace irods {
             virtual int                            l1_desc_idx()     const {
                 return l1_desc_idx_;
             }
-            virtual size_t                         size()            const {
+            virtual rodsLong_t                         size()            const {
```
trel, line 107: want to keep these aligned over on the right?

server/core/include/irods/irods_file_object.hpp
```diff
@@ -134,7 +135,7 @@ namespace irods {
             virtual void comm( rsComm_t* _c )                   {
                 comm_            = _c;
             }
-            virtual void size( size_t _v )                       {
+            virtual void size( rodsLong_t _v )                       {
```
trel, line 138: same here

alanking (commented):
Seems good to me. I vote for a squash, and leave off #s for now.

korydraughn:
Let's squash it without the pounds.

korydraughn:
Given that the file_object header's formatting is not up to par, I think we revert the alignment change in the unit test commit.

That header is going to be formatted in whole at some point in the future.

alanking (approved):

korydraughn (approved):
Please cherry-pick the second commit back to main and then add the pounds to this PR.

Nice work.

Nishant2022:
Added pounds

Issue #7120, opened 2023-06-11
test_authentication_PAM_without_negotiation fails due to unexpected PAM_AUTH_PASSWORD_FAILED from iinit 


SwooshyCueb:
Probably related: #7102
I ran into a test failure with `test_authentication_PAM_without_negotiation` while working on PREP.
(This test fails regardless of whether or not PREP is installed, so it's not a PREP issue.)

The test is running in an Ubuntu 20.04 container, using postgres.

Here's some of the test output. I've put a more complete log [here](https://gist.github.com/SwooshyCueb/81d24be5855776bcbeaf077af24249e6).

```
 --- IrodsSession: icommand executed by [rods#tempZone] [iinit] --- 
Assert Command: iinit
Expecting STDOUT_SINGLELINE: ['']
  stdout:
    | WARNING: Error 25 disabling echo mode. Password will be displayed in plaintext.
    | Enter your current iRODS password:
  stderr:
    | 
Output found

 --- IrodsSession: icommand executed by [rods#tempZone] [iadmin mkuser irodsauthuser rodsuser] --- 
Assert Command: iadmin mkuser irodsauthuser rodsuser
Expecting EMPTY: ['']
  stdout:
    | 
  stderr:
    | 
Output found

 --- IrodsSession: icommand executed by [rods#tempZone] [iadmin moduser irodsauthuser password ;=iamnotasecret] --- 
Assert Command: iadmin moduser irodsauthuser password ;=iamnotasecret
Expecting EMPTY: ['']
  stdout:
    | 
  stderr:
    | 
Output found

 --- IrodsSession: icommand executed by [irodsauthuser#tempZone] [iinit] --- 
Assert Command: iinit
Expecting STDOUT_SINGLELINE: ['']
  stdout:
    | WARNING: Error 25 disabling echo mode. Password will be displayed in plaintext.
    | Enter your current iRODS password:
  stderr:
    | 
Output found

 --- IrodsSession: icommand executed by [irodsauthuser#tempZone] [imkdir /tempZone/home/irodsauthuser/2023-06-11Z15:50:09--irods-testing-u14zfb04] --- 
Assert Command: imkdir /tempZone/home/irodsauthuser/2023-06-11Z15:50:09--irods-testing-u14zfb04
Expecting EMPTY: ['']
  stdout:
    | 
  stderr:
    | 
Output found

[...]

Issue #7102, opened 2023-05-22
test_authentication_PAM_with_server_params fails due to unexpected PAM_AUTH_PASSWORD_FAILED from iinit


SwooshyCueb:
- [x] main
- [x] 4-3-stable

---

I ran into a test failure with `test_authentication_PAM_with_server_params` while working on PREP.
(This test fails regardless of whether or not PREP is configured, so it's not a PREP issue.)

The test is running in an Ubuntu 20.04 container, using postgres.

Here's some of the test output. I've put a more complete log [here](https://gist.github.com/SwooshyCueb/542668a2cdd88a0decc7dc3eca75739a).

```
--- IrodsSession: icommand executed by [rods#tempZone] [iinit] --- 
Assert Command: iinit
Expecting STDOUT_SINGLELINE: ['']
  stdout:
    | WARNING: Error 25 disabling echo mode. Password will be displayed in plaintext.
    | Enter your current iRODS password:
  stderr:
    | 
Output found

 --- IrodsSession: icommand executed by [rods#tempZone] [iadmin mkuser irodsauthuser rodsuser] --- 
Assert Command: iadmin mkuser irodsauthuser rodsuser
Expecting EMPTY: ['']
  stdout:
    | 
  stderr:
    | 
Output found

 --- IrodsSession: icommand executed by [rods#tempZone] [iadmin moduser irodsauthuser password ;=iamnotasecret] --- 
Assert Command: iadmin moduser irodsauthuser password ;=iamnotasecret
Expecting EMPTY: ['']
  stdout:
    | 
  stderr:
    | 
Output found

 --- IrodsSession: icommand executed by [irodsauthuser#tempZone] [iinit] --- 
Assert Command: iinit
Expecting STDOUT_SINGLELINE: ['']
  stdout:
    | WARNING: Error 25 disabling echo mode. Password will be displayed in plaintext.
    | Enter your current iRODS password:
  stderr:
    | 
Output found

 --- IrodsSession: icommand executed by [irodsauthuser#tempZone] [imkdir /tempZone/home/irodsauthuser/2023-05-22Z19:03:07--irods-testing-1byta41r] --- 
Assert Command: imkdir /tempZone/home/irodsauthuser/2023-05-22Z19:03:07--irods-testing-1byta41r
Expecting EMPTY: ['']
  stdout:
    | 
  stderr:
    | 
Output found

[...]

PR #7189, opened 2023-07-14
[#7102,#7120] Skip auth test when missing irodsauthuser (main)

alanking:
Addresses #7102 
Addresses #7120 

Confirmed by manually running test without the user existing and observing the appropriate skip. Confirmed that the affected tests run and pass when the irodsauthuser exists. There's not really a way to confirm that the password is correct when running the tests as the iRODS service account.


korydraughn (approved):
Please add the pounds.

alanking:
After thinking on this some more and discussing offline, we have determined that it would be better to fail the test here instead of skipping it. We should use `unittest.TestCase.fail`: https://docs.python.org/3/library/unittest.html?highlight=unittest%20fail#unittest.TestCase.fail

korydraughn (commented):
Looks good.

Please squash to taste and leave off pounds if you're happy with it.

alanking:
Squashed

korydraughn (approved):
Pound/Merge when ready.

PR #7209, opened 2023-07-25
Multiple simple improvements (main)

korydraughn:
Still need to write a test for the microservice.

### DIFF
--- a/server/re/CMakeLists.txt
+++ b/server/re/CMakeLists.txt
@@ -55,6 +55,8 @@ target_include_directories(
   PUBLIC
   "$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>"
   "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
+  "$<BUILD_INTERFACE:${CMAKE_IRODS_SOURCE_DIR}/lib/administration/user/include>"
+  "$<BUILD_INTERFACE:${CMAKE_IRODS_SOURCE_DIR}/lib/administration/zone/include>"
   "$<BUILD_INTERFACE:${CMAKE_IRODS_SOURCE_DIR}/plugins/api/include>"
   "$<BUILD_INTERFACE:${CMAKE_IRODS_SOURCE_DIR}/server/api/include>"
   "$<BUILD_INTERFACE:${CMAKE_IRODS_SOURCE_DIR}/server/core/include>"
--- a/server/re/include/irods/reAction.hpp
+++ b/server/re/include/irods/reAction.hpp
@@ -64,6 +64,7 @@ int msiCommit( ruleExecInfo_t *rei );
 int msiDeleteCollByAdmin( msParam_t *parColl, msParam_t *childName, ruleExecInfo_t *rei );
 int msiDeleteUser( ruleExecInfo_t *rei );
 int msiAddUserToGroup( msParam_t *msParam, ruleExecInfo_t *rei );
+int msiRemoveUserFromGroup(MsParam* _group, MsParam* _user, MsParam* _user_zone, RuleExecInfo* _rei);
 int msiSendMail( msParam_t *toAddr, msParam_t *subjectLine, msParam_t *body, ruleExecInfo_t *rei );
 int msiGetObjType( msParam_t *objNameP, msParam_t *objTypeP,
                    ruleExecInfo_t *rei );
@@ -125,6 +126,8 @@ int msiTakeThreeArgumentsAndDoNothing(msParam_t *arg1, msParam_t *arg2, msParam_
 
 namespace irods
 {
+    // clang-format off
+
     // =-=-=-=-=-=-=-
     // implementation of the microservice table class, which initializes the table during the ctor
     // inserting the statically compiled microservices during construction providing RAII behavior
@@ -149,6 +152,7 @@ namespace irods
         table_[ "msiDeleteCollByAdmin" ] = new irods::ms_table_entry( "msiDeleteCollByAdmin", 2, std::function<int(msParam_t*,msParam_t*,ruleExecInfo_t*)>( msiDeleteCollByAdmin ) );
         table_[ "msiDeleteUser" ] = new irods::ms_table_entry( "msiDeleteUser", 0, std::function<int(ruleExecInfo_t*)>( msiDeleteUser ) );
         table_[ "msiAddUserToGroup" ] = new irods::ms_table_entry( "msiAddUserToGroup", 1, std::function<int(msParam_t*,ruleExecInfo_t*)>( msiAddUserToGroup ) );
+        table_[ "msiRemoveUserFromGroup" ] = new irods::ms_table_entry("msiRemoveUserFromGroup", 3, std::function<int(MsParam*, MsParam*, MsParam*, RuleExecInfo*)>(msiRemoveUserFromGroup)); // NOLINT(cppcoreguidelines-owning-memory)
         table_[ "msiSetDefaultResc" ] = new irods::ms_table_entry( "msiSetDefaultResc", 2, std::function<int(msParam_t*,msParam_t*,ruleExecInfo_t*)>( msiSetDefaultResc ) );
         table_[ "msiSetDataObjPreferredResc" ] = new irods::ms_table_entry( "msiSetDataObjPreferredResc", 1, std::function<int(msParam_t*,ruleExecInfo_t*)>( msiSetDataObjPreferredResc ) );
         table_[ "msiSetDataObjAvoidResc" ] = new irods::ms_table_entry( "msiSetDataObjAvoidResc", 1, std::function<int(msParam_t*,ruleExecInfo_t*)>( msiSetDataObjAvoidResc ) );
@@ -248,6 +252,8 @@ namespace irods
         table_[ "msiDoSomething" ] = new irods::ms_table_entry( "msiDoSomething", 2, std::function<int(msParam_t*,msParam_t*,ruleExecInfo_t*)>( msiDoSomething ) );
         table_[ "msiTakeThreeArgumentsAndDoNothing" ] = new irods::ms_table_entry( "msiTakeThreeArgumentsAndDoNothing", 3, std::function<int(msParam_t*,msParam_t*,msParam_t*,ruleExecInfo_t*)>( msiTakeThreeArgumentsAndDoNothing ) );
     }; // ms_table::ms_table
+
+    // clang-format on
 
     // =-=-=-=-=-=-=-
     // implement the micros table dtor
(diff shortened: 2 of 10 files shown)


server/re/include/irods/reAction.hpp
```diff
@@ -64,6 +64,7 @@ int msiCommit( ruleExecInfo_t *rei );
 int msiDeleteCollByAdmin( msParam_t *parColl, msParam_t *childName, ruleExecInfo_t *rei );
 int msiDeleteUser( ruleExecInfo_t *rei );
 int msiAddUserToGroup( msParam_t *msParam, ruleExecInfo_t *rei );
+int msiRemoveUserFromGroup(MsParam* _group, MsParam* _user, MsParam* _user_zone, RuleExecInfo* _rei);
```
trel, line 67: i notice these parameters are not consistent with the ones around it... others are all `_t`...  can you explain why?

korydraughn: `MsParam` is the concrete type name while `msParam_t` is a type alias for `MsParam`.

They are identical to the implementation. `msParam_t` exists because of C.

trel: gotcha - so we *could* update all the others if we wanted to 'modernize'...   but not today.

korydraughn: That's correct.

Modernize is probably not the right word for this since `msParam_t` isn't outdated. I just chose to use that today. Could change my mind tomorrow for any number of reasons :-)

server/re/CMakeLists.txt
```diff
@@ -60,6 +60,8 @@ target_include_directories(
   "$<BUILD_INTERFACE:${CMAKE_IRODS_SOURCE_DIR}/server/core/include>"
   "$<BUILD_INTERFACE:${CMAKE_IRODS_SOURCE_DIR}/server/delay_server/include>"
   PRIVATE
+  "${CMAKE_IRODS_SOURCE_DIR}/lib/administration/user/include"
+  "${CMAKE_IRODS_SOURCE_DIR}/lib/administration/zone/include"
```
korydraughn, line 64: Need to think about this a little more.

Need the user administration library (which relies on a type from the zone administration) for a new microservice (i.e. `msiRemoveUserFromGroup`), hence why I added this.

I'm wonder if this should use `$<BUILD_INTERFACE:...>` and/or whether it should be in the public section.

Is there a better way to do this?

@SwooshyCueb Thoughts?

korydraughn:
Unit tests pass.

Running core tests now.

server/re/CMakeLists.txt
SwooshyCueb: Yeah you want `$<BUILD_INTERFACE:...>` and in the `PUBLIC` section.

korydraughn: Can you explain how you arrived at that conclusion?

korydraughn:
All core tests passed.

Will update the CMake and re-run the tests to be on the safe side.

korydraughn:
Running core tests with cmake updates.

korydraughn:
Core tests passed.

server/re/include/irods/reAction.hpp
korydraughn: Do you have a preference for one over the other?

trel: no, defer to you and team.   just caught my eye is all.

alanking (commented):
We need tests for the new microservice, yes?

server/re/src/icatAdminMS.cpp
```diff
@@ -341,6 +354,68 @@ msiAddUserToGroup( msParam_t *msParam, ruleExecInfo_t *rei ) {
     return i;
 }
 
+/// Removes a user from a group.
+///
+/// \param[in] _group     The name of the group to remove \p _user from.
+/// \param[in] _user      The part of a fully-qualified username preceding the pound sign.
```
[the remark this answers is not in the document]
korydraughn: I had a similar thought. The fact that you also mentioned it means it's probably a good idea to do it.

Will change.

korydraughn:
> We need tests for the new microservice, yes?

Yes. Will add tests for bad inputs.

alanking:
Oh, just saw the tests. Disregard last comment

korydraughn:
Well, we cover good inputs, but not bad ones. We definitely need to cover the negative space.

server/re/CMakeLists.txt
korydraughn: Created irods/irods_docs#208 to capture CMake knowledge in the developer docs.

server/re/include/irods/reAction.hpp
korydraughn: Leaving as is. This isn't urgent at the moment.

korydraughn:
Added python tests for bad inputs and remote users.

The new tests pass.
Re-ran the unit tests and they passed too.

Please review and let me know if anything looks suspicious.

alanking (commented):
Seems good to me. Squash as desired and leave off #s

trel:
didn't see a test that shows it works / doesn't break when the requested user to be removed is NOT part of that group.

korydraughn:
Ah, will add a test.

korydraughn:
Now the question becomes ... should the MSI return an error when attempting to remove the user from the group, but the user isn't a member of it?

At the moment, `CAT_SUCCESS_BUT_WITH_NO_INFO` is returned which **could** mean the user isn't a member, but no one can rely on that exactly.

Should anyone care about this case? I don't think there's much value in knowing whether the removal failed in this case. Why would someone want to know that? They can always query for that info.

Thoughts?

korydraughn:
New test passed at the bench which means all tests pass.

korydraughn:
Ubuntu returns an error if the user isn't a member of the group.
```bash
sudo addgroup testgroup
sudo deluser <username> testgroup
/usr/sbin/deluser: The user `<username>` is not a member of group `testgroup`.
```

However, does it make sense to return an error in a policy context?

Should we add an option that changes the MSI's behavior?

An additional option would move the decision to the caller. Thoughts?

trel:
unix behavior is a strong indicator.    i think we need a reason to diverge from that answer.

`CAT_SUCCESS_BUT_WITH_NO_INFO` is a poor return code... do we have something else that complains more correctly?

korydraughn:
Not sure we have an error that captures that case. Can add one though.

How about `USER_NOT_A_MEMBER_OF_GROUP`?

That code should likely come from the database plugin.

trel:
so it's historically always given the `CAT_SUCCESS_BUT_WITH_NO_INFO` error?

korydraughn:
That's my guess.

The MSI calls the user administration library, which calls rsGeneralAdmin. Everything's a thin wrapper around rsGeneralAdmin.

trel:
Then let’s just leave it for now.   It’s not broken, just a bit esoteric.

korydraughn:
Sure. I'll write up an issue to introduce a better error code for this case.

Can handle it in a later PR.

korydraughn:
Created new issue: #7216

korydraughn:
With that, I'm going to squash the changes.

korydraughn:
Squashed.

Checking the issue numbers for correctness.

SwooshyCueb:
There's a lot of `clang-tidy` whinging. I'd like to see the `bugprone-lambda-function-name` check failures in `icatAdminMS.cpp` either addressed or disabled with `NOLINTNEXTLINE`, but I don't feel that strongly about it if others feel otherwise

korydraughn:
Good catch. Will fix.

SwooshyCueb:
Looks like we still have a `bugprone-lambda-function-name` check failure
https://github.com/irods/irods/actions/runs/5683534551/job/15404295230?pr=7209#step:7:58

korydraughn:
I'm pretty sure that one is not real. The HTTP API does the same thing and results in the correct behavior. I'll verify with GCC to see if it results in different behavior. I'm hoping it handles the `fn = __func__` like Clang. If not, then I'll pull it out.

Will take care of all the workflow results.

korydraughn:
I've confirmed GCC handles the capture of the function name the same way as Clang. Address sanitizer doesn't report any issues either, so we're good.

Will add a `NOLINT` directive to the lambda.

alanking (commented):
Looks good to me

SwooshyCueb (approved):

korydraughn:
I made a change to one of the serialization routines where clang-tidy reported a performance issue.

Will run through the test suite again to confirm everything is still good.

korydraughn:
All unit tests passed.

Running core tests.

server/re/src/irods_re_serialization.cpp
```diff
@@ -928,24 +963,32 @@ namespace irods::re_serialization
...
-                        _out["value_" + j_str + "_" + i_str] = boost::lexical_cast<std::string>(l->sqlResult[i].value+j*l->sqlResult[i].len);
+                        // NOLINTNEXTLINE(boost-use-to-string, cppcoreguidelines-pro-bounds-constant-array-index)
+                        _out["attriInx_" + i_str] = boost::lexical_cast<std::string>(tmp->sqlResult[i].attriInx);
+                        // NOLINTNEXTLINE(boost-use-to-string, cppcoreguidelines-pro-bounds-constant-array-index)
+                        _out["len_" + i_str] = boost::lexical_cast<std::string>(tmp->sqlResult[i].len);
+
+                        // clang-format off
+                        _out[fmt::format("value_{}_{}", j_str, i_str)] =
+                            // NOLINTNEXTLINE(bugprone-implicit-widening-of-multiplication-result, cppcoreguidelines-pro-bounds-constant-array-index, cppcoreguidelines-pro-bounds-pointer-arithmetic)
+                            boost::lexical_cast<std::string>(tmp->sqlResult[i].value + j * tmp->sqlResult[i].len);
```
korydraughn, line 986: Capture the bugprone clang-tidy error in an issue.

korydraughn:
All core tests passed.

Squashing now.

korydraughn:
I changed the formatting for a line that was really long, but left the actual logic as it was.

Will add the pounds once approved.

korydraughn:
Just to be safe, I ran the unit tests and they all passed.

server/re/src/irods_re_serialization.cpp
alanking: Is this done?

alanking (approved):
Yep, add #s when ready.

server/re/src/irods_re_serialization.cpp
korydraughn: The line in question will change with this PR. I'm going to create the issue after this is merged.

alanking: Makes sense. So this will be resolved after merging, then.

korydraughn: Yep

korydraughn:
Added #'s

server/re/src/irods_re_serialization.cpp
korydraughn: Created #7222

Issue #3824, opened 2018-02-21
set socket options to avoid timeouts


HowardLander:
- [x] main
- [x] 4-3-stable
---

### iRODS Version, OS and Version
4.2.2 Ubuntu 7

### What did you try to do?
When trying to iput a 10GB file from a RENCI Ubuntu VM into a compound resource (using the new azure driver)on a Ubuntu VM inside of azure, the transfer failed because azure decided to close the connection on port 1247.  Oops... I fixed this on the client side by altering the kernel TCP parameters like so:

sudo echo 120 > /proc/sys/net/ipv4/tcp_keepalive_time
sudo echo 30 > /proc/sys/net/ipv4/tcp_keepalive_intvl

but arguably this should be handled by the iRODS client code using setsockopt to set the keep alive parameters (see http://tldp.org/HOWTO/html_single/TCP-Keepalive-HOWTO/) on the individual socket . This could be configurable, if desired.

trel:
"The TCP idle timeout on the Azure load balancer is 240 seconds by default."

trel:
related to https://github.com/irods/irods/issues/5218

stsnel:
It would also be helpful for us to have iRODS handle these settings on a socket level, mainly for the following use cases:
* Researchers who want to upload large files using the iCommands from shared infrastructure, such as a compute cluster, where sysadmins would prefer not to change kernel parameters for one application.
* Researchers who want to upload data using the iCommands on WSL are often not familiar with changing kernel parameters. It would be easier to have iRODS handle this automatically rather than talk users through these changes.

trel:
Agreed - do you know the setting we'd need to manipulate?

kript:
> Agreed - do you know the setting we'd need to manipulate?

There is code in the linked article above for a test program enabling and disabling, in case you hadn't seen it (I'd used it to investigate similar options for https://github.com/irods/irods/issues/3986).

Issue #2533, opened 2015-02-19
Investigate TCP_KEEPALIVE on the server


kellerb:
- [x] main
- [x] 4-3-stable
---
@michael-conway 

To address parallel transfer timeouts noticed at iPlant.

Possibly allow setting of the following in `server_config.json`:
- `tcp_keepalive_time`
- `tcp_keepalive_intvl`
- `tcp_keepalive_probes`

trel:
related to https://github.com/irods/irods/issues/5218

PaulVanSchayck:
Having spent a couple of days debugging idle iRODS connections being killed across firewalls, I would strongly argue in getting this implemented.

In our case the 1247 server between two resource servers was timed out when one side of a msiDataObjRepl command was performing the checksum on a file. The resulting error messages looked very confusing (errors during checksum) and it took a while to figure out it was really the connection timing out. 

Setting the tcp_keepalive flag on all iRODS server sockets (catalog provider and consumer) , and having its value configurable from `server_config.json` makes a lot of sense to me. This is quite a bit better than the server wide sysctl setting applying to all sockets. TBH, I see little reason not to do this and also set a relatively aggressive default 10 minute value on `tcp_keepalive_time`. iRODS is dependent on keeping these connections alive, and an automatic reconnect mechanism is much harder to implement (due to PEPs).

For future readers, I found it useful to look at `netstat -n --tcp --timers` to debug the situation. I didn't find a reliable `tcpdump` filter to only catch keepalive ACK packets.

trel:
Agree on all points, including the default 10min.   Thanks Paul.

PR #7225, opened 2023-07-28
[#2533,#3824] Add TCP keepalive option configs (4-3-stable)

alanking:
Addresses https://github.com/irods/irods/issues/2533, https://github.com/irods/irods/issues/3824

Cherry-picked from  #7215


korydraughn (approved):
Add the pounds.

Issue #7169, opened 2023-07-05
Creating a data object containing a single quote in the name using itouch results in SYS_NO_GOOD_REPLICA


korydraughn:
## Bug Report

We need to investigate why `SYS_NO_GOOD_REPLICA` is returned for this case. The data object appears to have been created successfully.
```
irods@cc9948e56928:~$ ils
/tempZone/home/rods:
irods@cc9948e56928:~$ itouch "just testin'"
Level 0: SYS_NO_GOOD_REPLICA: No good replicas found.


irods@cc9948e56928:~$ ils
/tempZone/home/rods:
  just testin'
irods@cc9948e56928:~$ ils -l
/tempZone/home/rods:
  rods              0 demoResc            0 2023-07-05.12:24 & just testin'
```

alanking:
Related: Related: https://github.com/irods/irods/issues/5727

FifthPotato:
Verified that this is due to the genQuery issue. Throws on line 384 of `plugins/api/src/touch.cpp`

https://github.com/irods/irods/blob/198873a0c2ac137ec6cbe18bff314bc34b8db311/plugins/api/src/touch.cpp#L373-L385

Above it, you see a genQuery. I tested as follows:
```
root@bde8b726cabd:/# iquest "select DATA_MODIFY_TIME, DATA_REPL_NUM where COLL_NAME = '/tempZone/home/rods' and DATA_NAME = 'testin'' and DATA_REPL_STATUS = '1'"             
CAT_NO_ROWS_FOUND: Nothing was found matching your query
```

```
root@bde8b726cabd:/# iquest "select DATA_MODIFY_TIME, DATA_REPL_NUM where COLL_NAME = '/tempZone/home/rods' and DATA_REPL_STATUS = '1' and DATA_NAME = 'testin''"             
DATA_MODIFY_TIME = 01694509324
DATA_REPL_NUM = 0
------------------------------------------------------------
```

Since swapping the where clauses caused the embedded quote to stop malfunctioning (exactly as the other embedded quote issues behaved) this is identical to the others.

korydraughn:
Great investigative work.

Bumping. This will be fixed once the GenQuery2 parser is available.

trel:
Swapping was demonstrated to 'work' in https://github.com/irods/irods/issues/5727.

But depending on this buggy behavior is not healthy.

Yes, GenQuery2 is the answer.

korydraughn:
Related #3902

[...]

Issue #7164, opened 2023-06-30
`ils -A` does not print ACLs for data object with single quotation mark in name


ll4strw:
- iRODS v4.3.0
- Rocky v8
- icommands v4.3.0

``iput test1 "just testin'"`` completes without errors and the file ``test1`` is successfully transferred on the server. However, ``irm -f "just testin'"`` returns 

```
remote addresses: x.x.x.x ERROR: rmUtil: rm error for /zone/home/username/just testin', status = -1205000 status = -1205000 RE_RUNTIME_ERROR
Level 0: no good replica found: Unknown error -166000
```

I would rather expect a bidirectional success or a bidirectional failure. How can I remove the test file using the icommands? Thanks.

---

## Update
The primary goal of this issue is to make it so that ACLs are printed correctly when handling a data object with a single quotation mark in its name.

See https://github.com/irods/irods/issues/7164#issuecomment-1621647178 for immediate details.

korydraughn:
This feels like a bug in 4.3.0.

I tried this using a custom build based on tip of 4-3-stable and it handled the single quote just fine.

Let me see if I can reproduce this.

trel:
Please share the output just after your initial `iput`:

```
ils -L "just testin'"
```

trel:
Interesting... since there is clearly a good replica (the `&` in your `ils` output)...

```
$ ierror 1205000
irods error: -1205000 RE_RUNTIME_ERROR 
$ ierror 166000
irods error: -166000 SYS_NO_GOOD_REPLICA
```

What type of resources are these?

```
ilsresc -l ilorentzNoReplicaResource  | grep type
ilsresc -l TrueNASNoReplicaResource  | grep type
```

trel:
Is there anything useful in the server log (`/var/log/irods/irods.log`)?

korydraughn:
This worked just fine for me on a fresh install of 4.3.0. Seems there's something else going on?
```bash
root@9f7378215fc8:/# su - irods                                            
irods@9f7378215fc8:~$ iput version.json "just testin'"                     
irods@9f7378215fc8:~$ ils -l                                               
/tempZone/home/rods:                                                       
  rods              0 demoResc          239 2023-06-30.12:57 & just testin'
irods@9f7378215fc8:~$ irm -f "just testin'"                                
irods@9f7378215fc8:~$ cat version.json                                     
{                                                                          
    "catalog_schema_version": 9,                                           
[...]

PR #7237, opened 2023-08-01
Added single quote escaping to general_query

FifthPotato:
Potentially scary PR that allows for embedded single quote stuff to be used in genQuery1.

Pros:
- Fixes the embedded single quote issue (embedded quotes parsed by general_query should now be in doubles, so the where clause WHERE DATA_NAME = 'test''in' -> WHERE DATA_NAME = 'test'in' in the SQL query)

Cons:
- Any client code using existing genQuery1 queries has potential points of failure and need to be rewritten (includes stuff like ils, iput, etc...)

e.g. of an issue with iput if no fixes are implemented and this change is merged in: any file with double embedded quotes will be treated as if they only had a single quote, and passing it with -f could clobber a file unexpectedly.

With genQuery2 on the horizon, this potentially breaking change is questionably worth merging. However, I will leave it here as a point of discussion or for posterity.

EDIT: Relevant issues appear to be #7169 #7164 (Verified this fix works for the latter, haven't checked the former)

### DIFF
--- a/plugins/database/src/general_query.cpp
+++ b/plugins/database/src/general_query.cpp
@@ -1525,6 +1525,9 @@ insertWhere( char *condition, int option ) {
     }
 
     for ( cp1 = cpFirstQuote + 1; cp1 < cpSecondQuote; cp1++ ) {
+        if ( *cp1 == '\'' && *(cp1 + 1) == '\'' ) {
+            cp1++;
+        }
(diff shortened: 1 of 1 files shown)


alanking:
Is there an issue number for this? If so, please mention here in a comment so we have some additional context. Thanks!

FifthPotato:
Added it to the original post.

Issue #7110, opened 2023-06-02
 iput -b -f creates extra replicas on random resource instead of overwriting existing replicas


tedgin:
- [x] main
- [x] 4-3-stable
- [x] `iput` help text (main)
- [x] `iput` help text (4-3-stable)
---
### iRODS Version, OS and Version
iRODS Version 4.2.12, CentOS 7
iCommands Version 4.2.8

### What did you try to do?

Using iput version 4.2.8, I try to bulk upload a folder on top of an existing collection. Before the upload, for each file in the folder, there existed a data object in the collection with the same name. The default resource is a `random` resource with two `passthru` child resources. Each child has a `unixfilesystem` child resource,

```console
4.2.8? ilsresc
primaryRes:random
├── primary_1:passthru
│   └── primaryStorage_1:unixfilesystem
└── primary_2:passthru
    └── primaryStorage_2:unixfilesystem
4.2.8? ls bulk
test1B  test2B
4.2.8? ils -L bulk  
/tempZone/home/user/bulk:
  user              0 primaryRes;primary_2;primaryStorage_2            1 2023-06-02.09:01 & test1B
        generic    /var/lib/irods/vaults/primary_2/home/user/bulk/test1B
  user              0 primaryRes;primary_2;primaryStorage_2            2 2023-06-02.09:01 & test2B
        generic    /var/lib/irods/vaults/primary_2/home/user/bulk/test2B
4.2.8? iput -bfr bulk
Running recursive pre-scan... pre-scan complete... transferring data...
```

### Expected behavior

I expected each replica of each data object being overwritten by its corresponding file.

### Observed behavior (including steps to reproduce, if applicable)

Each data object got a second replica on the opposite storage resource from the first replica with the first replica being write-locked.

```console
4.2.12? ils -L bulk
/tempZone/home/user/bulk:
  user              0 primaryRes;primary_2;primaryStorage_2            1 2023-06-02.09:01 ? test1B
        generic    /var/lib/irods/vaults/primary_2/home/user/bulk/test1B
  user              1 primaryRes;primary_1;primaryStorage_1            1 2023-06-02.09:02 & test1B
        generic    /var/lib/irods/vaults/primary_1/home/user/bulk/test1B
  user              0 primaryRes;primary_2;primaryStorage_2            2 2023-06-02.09:01 ? test2B
        generic    /var/lib/irods/vaults/primary_2/home/user/bulk/test2B
[...]

PR #7244, opened 2023-08-02
Overwrites for bulk puts (main)

alanking:
There are two and a half solutions presented here, as discussed in #7110. The "half" is whether to use `SYS_NOT_ALLOWED` or `USER_INCOMPATIBLE_PARAMS` in the case where we do not allow overwrites with bulk puts.

Please provide input on which solution seems more palatable.

### DIFF
--- a/scripts/irods/test/test_iput.py
+++ b/scripts/irods/test/test_iput.py
@@ -541,6 +541,79 @@ class Test_Iput(session.make_sessions_mixin(rodsadmins, rodsusers), unittest.Tes
 
             self.user.assert_icommand(['ils', '-l', collection_path], 'STDOUT', coll)
 
+    def test_recursive_bulk_overwrite_is_not_allowed__issue_7110(self):
+        filenames = ['file1', 'file2']
+        bonus_filename = 'file3'
+
+        dirname = 'bulkputme'
+        directory_path = os.path.join(self.user.local_session_dir, dirname)
+
+        collection_path = os.path.join(self.user.session_collection, dirname)
+        logical_paths = [os.path.join(collection_path, f) for f in filenames]
+        bonus_logical_path = os.path.join(collection_path, bonus_filename)
+
+        random_resource = 'rand'
+        ufs_resource_1 = 'ufs1'
+        ufs_resource_2 = 'ufs2'
+
+        try:
+            # Create the two files so that we can perform the bulk put.
+            os.mkdir(directory_path)
+            for f in filenames:
+                physical_path = os.path.join(directory_path, f)
+                lib.make_arbitrary_file(physical_path, 2)
+
+            # Set up random resource hierarchy such that only 1 replica will appear.
+            lib.create_random_resource(self.admin, random_resource)
+            lib.create_ufs_resource(self.admin, ufs_resource_1, hostname=test.settings.HOSTNAME_2)
+            lib.create_ufs_resource(self.admin, ufs_resource_2, hostname=test.settings.HOSTNAME_3)
+            lib.add_child_resource(self.admin, random_resource, ufs_resource_1)
+            lib.add_child_resource(self.admin, random_resource, ufs_resource_2)
+
+            # Bulk put the directory ensure that the objects appear as expected.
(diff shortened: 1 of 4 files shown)


trel:
`USER_INCOMPATIBLE_PARAMS` feels more specific.

I think that's a good thing here.

alanking:
So that's one vote for "disallow and use `USER_INCOMPATIBLE_PARAMS`", and I think I agree.

korydraughn:
I vote for `USER_INCOMPATIBLE_PARAMS` too.

scripts/irods/test/test_iput.py
```diff
@@ -541,6 +541,70 @@ def test_iput_with_apostrophe_logical_path__issue_5759(self):
...
+
+            for lp in logical_paths:
+                self.assertFalse(lib.replica_exists(self.user, lp, 1))
+                self.assertEqual(str(1), lib.get_replica_status(self.user, os.path.basename(lp), 0))
+
+        finally:
+            self.user.assert_icommand(['ils', '-lr', self.user.session_collection], 'STDOUT') # debugging
+
+            # Just in case, make sure any existing replicas are stale before removing. The issue results in replicas
+            # being stuck in the intermediate status.
```
korydraughn, line 592: Is this true? Can the replicas be stuck in the intermediate state via this test?

Or is this just about making sure the test is able to clean up behind itself?

korydraughn (commented):
This feels like something that should be discussed in PC/TWG.

scripts/irods/test/test_iput.py
alanking: I misspoke and said "intermediate" here when I meant to say "write-locked" but yes, the linked issue causes extra replicas to be created and the original replicas which were not overwritten are left in the write-locked status. This is here to ensure that the test cleans up after itself in the event of a regression.

korydraughn: Okay. I am understanding correctly then.

This is no longer an issue and the code in the finally block exists so that things are cleaned up appropriately in case of an error.

Thanks!

trel:
ready to squash?    can put the merge on hold until after PC/TWG discush.

korydraughn:
Not yet. The PC/TWG discussion is about which direction to take in regard to not allowing overwrites vs making each PUT do resource hierarchy resolution.

alanking:
A third option (an undesirable one, in my opinion) would be to try to match the behavior described in the help text for `iput` and attempt to overwrite the data objects which are on the selected resource hierarchy and return errors for the others (or, stop on the first error - unclear).

korydraughn:
Yes, we'll make sure to include all the known options for the meeting.

trel:
I bet there are zero.

scripts/irods/test/test_iput.py
```diff
@@ -541,6 +541,79 @@ def test_iput_with_apostrophe_logical_path__issue_5759(self):
...
+                self.assertFalse(lib.replica_exists(self.user, lp, 1))
+                self.assertEqual(str(1), lib.get_replica_status(self.user, os.path.basename(lp), 0))
+
+            # Add a bonus file to the directory so that the overwrite has something to do.
+            lib.make_arbitrary_file(os.path.join(directory_path, bonus_filename), 2)
+
+            # Bulk put the directory with force flag to overwrite and ensure that the action is disallowed.
+            self.user.assert_icommand(
+                ['iput', '-R', random_resource, '-brf', directory_path, self.user.session_collection],
+                'STDERR', 'USER_INCOMPATIBLE_PARAMS')
```
korydraughn, line 587: Consider including the error code with the error name to avoid false test results.

alanking: Will do

alanking: Done

korydraughn:
The best kind of fixes.

Anything left for this PR?

alanking:
This one is good to go, I think.

korydraughn (approved):
Squash to taste and add the pounds.

Issue #7098, opened 2023-05-08
PAM passwords cannot exceed `MAX_PASSWORD_LEN` (50) characters


alanking:
- [x] main
- [x] 4-3-stable
---
### iRODS Version, OS and Version
4.2.11 (can trace this back as far as 4.2.0, and still affects 4.3.0)

### What did you try to do?
Set a user's PAM password to a value with 50 characters or more.

### Expected behavior
I expect to be able to use PAM passwords up to `pam_password_length` in length.

### Observed behavior (including steps to reproduce, if applicable)

`PAM_AUTH_PASSWORD_FAILED` is returned when trying to authenticate with `iinit`. A 49-character password works fine.

I found that using `irodsPamAuthCheck` does work with the 50-character password:
```bash
$ echo 'irodsauthuser:00000000000000000000000000000000000000000000000000' | chpasswd
$ iinit
Enter your current PAM password:
[-]     /repos/irods/plugins/auth/pam/libpam.cpp:241:irods::error pam_auth_client_request(irods::plugin_context &, rcComm_t *) :  status [PAM_AUTH_PASSWORD_FAILED]  errno [] -- message [call to rcAuthRequest failed.]

 failed with error -993000 PAM_AUTH_PASSWORD_FAILED 
$ /usr/sbin/irodsPamAuthCheck irodsauthuser
00000000000000000000000000000000000000000000000000
Authenticated
```

This is because the PAM authentication plugin prompts the user for a password and stores the result from stdin in a buffer by copying the first `MAX_PASSWORD_LEN` bytes into a buffer of length `MAX_PASSWORD_LEN + 2`:
https://github.com/irods/irods/blob/bc6f9f1cdef6c4ec01ea14402428988892615321/plugins/auth/pam/libpam.cpp#L102

`MAX_PASSWORD_LEN` is 50. `pam_password_length` is not used by any code in the server or the authentication plugin. This setting should be used to dictate the length of the password sent to the server.

Originally reported here: https://groups.google.com/g/irod-chat/c/ltCVOloHWbg/m/PqohzBQHBgAJ

alanking:
Some notes for posterity, after some snooping around...

`pam_password_length` is not being used, as noted before. However, in `server_config`, `/plugin_configuration/authentication/pam_password/password_length` is being used in the database plugin in `db_update_pam_password_op`.

[...]

PR #7245, opened 2023-08-02
Remove password_length configuration (main)

alanking:
This presents two solutions to the pam_password length enforcement problem presented in #7098.

In one, the length restriction is increased from 50 (`MAX_PASSWORD_LEN`) to `PAM_MAX_RESP_SIZE` (on my system, this value is 512). This is the solution discussed in https://github.com/linux-pam/linux-pam/issues/118.

In the other, the length restriction is removed entirely on the client side and the entire password entered by the user is copied into the JSON input and passed to the server. I have tested a password of length 6000 and experienced no problems with this solution.

The question is... which of these is what we want, if either?

Bonus question: Should the `password_length` be signed or unsigned?

A test has been added to ensure that the configured `password_length` is being enforced by the server-side configuration. However, by this point, a password of very long length may have been passed over the wire, which may or may not have been the purpose of the limited password buffer in the first place.

No automated test was added for a PAM password of length greater than 50 because this would require a change to the testing infrastructure. I am willing to do this as well, but wasn't sure.

### DIFF
--- a/plugins/auth/src/pam_password.cpp
+++ b/plugins/auth/src/pam_password.cpp
@@ -43,15 +43,13 @@ namespace
         printf("Enter your current PAM password:");
         std::string password;
         getline(std::cin, password);
-        char new_password[MAX_PASSWORD_LEN + 2]{};
-        strncpy(new_password, password.c_str(), MAX_PASSWORD_LEN);
         printf("\n");
         tty.c_lflag = oldflag;
         if (tcsetattr(STDIN_FILENO, TCSANOW, &tty)) {
             printf("Error reinstating echo mode.\n");
         }
 
-        return new_password;
+        return password;
     } // get_password_from_client_stdin
 
 #ifdef RODS_SERVER
@@ -312,8 +310,8 @@ namespace irods
                 }
             }
 
-            const auto username = req.at("user_name").get_ref<const std::string&>();
-            const auto password = req.at(irods::AUTH_PASSWORD_KEY).get_ref<const std::string&>();
+            const auto& username = req.at("user_name").get_ref<const std::string&>();
+            const auto& password = req.at(irods::AUTH_PASSWORD_KEY).get_ref<const std::string&>();
 
             log_auth::trace("performing PAM auth check for [{}]", username);
 
@@ -321,14 +319,16 @@ namespace irods
 
             log_auth::trace("PAM auth check done", username);
 
-            char password_out[MAX_NAME_LEN]{};
-            char* pw_ptr = &password_out[0];
-            const int ec = chlUpdateIrodsPamPassword(&comm, const_cast<char*>(username.c_str()), ttl, nullptr, &pw_ptr);
+            // Plus 1 for null terminator.
+            std::array<char, MAX_PASSWORD_LEN + 1> password_out{};
+            char* password_ptr = password_out.data();
+            const int ec =
+                chlUpdateIrodsPamPassword(&comm, username.c_str(), ttl, nullptr, &password_ptr, password_out.size());
             if (ec < 0) {
                 THROW(ec, "failed updating iRODS pam password");
             }
 
-            resp["request_result"] = password_out;
+            resp["request_result"] = password_out.data();
 
             if (comm.auth_scheme) {
                 free(comm.auth_scheme);
--- a/plugins/database/src/db_plugin.cpp
+++ b/plugins/database/src/db_plugin.cpp
@@ -103,7 +103,6 @@ static rodsLong_t MAX_PASSWORDS = 40;
 // =-=-=-=-=-=-=-
 // local variables externed for config file setting in
 bool irods_pam_auth_no_extend = false;
-size_t irods_pam_password_len = 20;
 char irods_pam_password_min_time[ NAME_LEN ]     = { "121" };
 char irods_pam_password_max_time[ NAME_LEN ]     = { "1209600" };
 char irods_pam_password_default_time[ NAME_LEN ] = { "1209600" };
@@ -1993,11 +1992,6 @@ irods::error db_open_op(
             std::vector<std::string>{irods::KW_CFG_PLUGIN_TYPE_AUTHENTICATION,
                                      auth_scheme::pam_password,
                                      irods::KW_CFG_PAM_NO_EXTEND});
-
-        irods_pam_password_len = irods::get_server_property<const size_t>(
-            std::vector<std::string>{irods::KW_CFG_PLUGIN_TYPE_AUTHENTICATION,
-                                     auth_scheme::pam_password,
-                                     irods::KW_CFG_PAM_PASSWORD_LENGTH});
 
         snprintf(irods_pam_password_min_time, NAME_LEN, "%s",
                  irods::get_server_property<const std::string>(
@@ -7070,12 +7064,13 @@ irods::error db_make_limited_pw_op(
 
 // =-=-=-=-=-=-=-
 // authenticate user
-irods::error db_update_pam_password_op(
-    irods::plugin_context& _ctx,
-    const char*            _user_name,
-    int                    _ttl,
-    const char*            _test_time,
-    char**                 _irods_password ) {
+auto db_update_pam_password_op(irods::plugin_context& _ctx,
+                               const char* _user_name,
+                               int _ttl,
+                               const char* _test_time,
+                               char** _password_buffer,
+                               std::size_t _password_buffer_size) -> irods::error
+{
     // =-=-=-=-=-=-=-
     // check the context
     irods::error ret = _ctx.valid();
@@ -7083,14 +7078,19 @@ irods::error db_update_pam_password_op(
         return PASS( ret );
     }
 
-    // =-=-=-=-=-=-=-
-    // check the params
-    if (
-        !_user_name ||
-        !_irods_password ) {
+    if (!_user_name || !_password_buffer) {
+        return ERROR(CAT_INVALID_ARGUMENT, "null parameter");
+    }
+
+    // Plus 1 for null terminator.
+    std::array<char, MAX_PASSWORD_LEN + 1> random_password{};
+    if (random_password.size() > _password_buffer_size) {
         return ERROR(
-                   CAT_INVALID_ARGUMENT,
-                   "null parameter" );
+            SYS_INVALID_INPUT_PARAM,
+            fmt::format("{}: Buffer not large enough to hold password. Requires [{}] bytes, received [{}] bytes.",
+                        __func__,
+                        random_password.size(),
+                        _password_buffer_size));
     }
 
     // =-=-=-=-=-=-=-
@@ -7108,12 +7108,7 @@ irods::error db_update_pam_password_op(
 //        _ctx.prop_map().get< icatSessionStruct >( ICSS_PROP, icss );
 
     char myTime[50];
-    char rBuf[200];
-    size_t i, j;
-    char randomPw[50];
-    char randomPwEncoded[50];
     int status;
-    char passwordInIcat[MAX_PASSWORD_LEN + 2];
     char passwordModifyTime[50];
     char *cVal[3];
     int iVal[3];
@@ -7184,9 +7179,11 @@ irods::error db_update_pam_password_op(
         return ERROR( status, "delete failure" );
     }
 
-    if (logSQL != 0)
+    if (logSQL != 0) {
         log_sql::debug("chlUpdateIrodsPamPassword SQL 3");
-    cVal[0] = passwordInIcat;
+    }
+
+    cVal[0] = random_password.data();
     iVal[0] = MAX_PASSWORD_LEN;
     cVal[1] = passwordModifyTime;
     iVal[1] = sizeof( passwordModifyTime );
@@ -7212,7 +7209,8 @@ irods::error db_update_pam_password_op(
             cllBindVars[cllBindVarCount++] = myTime;
             cllBindVars[cllBindVarCount++] = expTime;
             cllBindVars[cllBindVarCount++] = selUserId;
-            cllBindVars[cllBindVarCount++] = passwordInIcat;
+            // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-constant-array-index)
+            cllBindVars[cllBindVarCount++] = random_password.data();
             status =  cmlExecuteNoAnswerSql( "update R_USER_PASSWORD set modify_ts=?, pass_expiry_ts=? where user_id = ? and rcat_password = ?",
                                              &icss );
             if ( status ) {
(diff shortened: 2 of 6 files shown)


ccacciari:
We have a use case where a token of length 1300 is used in iRODS as a PAM password. So we support a solution that allows to configure the max length server side.

scripts/irods/test/test_auth.py
```diff
@@ -745,3 +745,126 @@ def test_iinit_with_pam_password_and_ssl(self):
         # Make sure that the authentication succeeded.
         assert_command('ils', 'STDOUT', f'/{self.zone}/home/{self.pam_user}', env=self.env)
 
+
+@unittest.skipIf(test.settings.USE_SSL, 'These tests set up SSL so if SSL is already set up, this should be skipped.')
+class test_pam_password(resource_suite.ResourceBase, unittest.TestCase):
```
korydraughn, line 750: Do we need to inherit from `ResourceBase`?

scripts/irods/test/test_auth.py
```diff
@@ -745,3 +745,126 @@ def test_iinit_with_pam_password_and_ssl(self):
...
+                            irods_config.server_config['plugin_configuration']['authentication'] = {
+                                'pam_password': { 'password_length': length }
+                            }
+                            lib.update_json_file_from_dict(irods_config.server_config_path, irods_config.server_config)
+
+                            IrodsController().restart()
+
+                            # The actual test
+                            self.auth_session.assert_icommand('iinit', 'STDERR', 'PASSWORD_EXCEEDS_MAX_SIZE',
+                                                              input=f'{self.auth_session.password}\n')
```
korydraughn, line 847: Perhaps each iteration through the loop can be a subTest?

korydraughn (commented):
Looks good.

korydraughn:
I forgot to read your description for the PR.

Will do that and leave my thoughts on it before the end of the week.

scripts/irods/test/test_auth.py
alanking: Mmmm... probably not. :) I believe I was being especially lazy when I wrote this so this can probably be pared down a lot. I'll look into it.

korydraughn:
> The question is... which of these is what we want, if either?

In the first solution, `PAM_MAX_RESP_SIZE` becomes the new default, but admins are allowed to change it, correct?

I certainly think that is the correct way to proceed. Allowing passwords of any length feels too loose/unsafe.

Should the max password length be stored in the database rather than server_config.json? (i.e. `R_GRID_CONFIGURATION`)

@ccacciari @alanking @trel Thoughts?

> Bonus question: Should the password_length be signed or unsigned?

Let's stick with signed.

trel:
> Should the max password length be stored in the database rather than server_config.json? (i.e. `R_GRID_CONFIGURATION`)
 
password length/requirements feels like a zone-wide setting...

so database seems like a more appropriate location than a particular server's server_config

korydraughn:
I agree.

alanking:
> > The question is... which of these is what we want, if either?
> 
> In the first solution, `PAM_MAX_RESP_SIZE` becomes the new default, but admins are allowed to change it, correct?
> 
> I certainly think that is the correct way to proceed. Allowing passwords of any length feels too loose/unsafe.
> 
> Should the max password length be stored in the database rather than server_config.json? (i.e. `R_GRID_CONFIGURATION`)
> 
> @ccacciari @alanking @trel Thoughts?

The passwords are being sent from the client, so we would need to reach out to the server to get the configuration in order to adjust the size of the buffer being sent. I suppose that if this were a database configuration, this could become a simple GenQuery.

I'm wondering if we want to add a new database configuration at this stage rather than using the existing machinery. Granted, the existing machinery was not being used (which is why this work is necessary) but the pam_password plugin requires redirecting to the catalog provider, which is why a server_config solution seems okay. This can of course be changed in the future, and may be made easier by putting the configuration in the catalog (not to mention HA/all-provider setups). Just thought I'd mention.

korydraughn:
Discush!

alanking:
After some discussion, @korydraughn and I decided that not placing a limit on the password size on the client side should be okay given that the packing instructions for the authenticate API is just a a `BytesBuf`. The only concern would be exceeding the memory available on the server, but this is no worse than before.

Will await thoughts from @trel.

korydraughn:
To add to that, the server would return an error if the password length is exceeded.

We're also planning to move the password length setting into the database.

trel:
agreed - let's remove limits where we can - and document where the limits actually are and the reasons people may hit them (aka server side memory).

korydraughn:
Hmm, what happens if someone sends a ridiculously large password to the server?
What if they do that using multiple machines at the same time?

That could possibly bring a server down.

Perhaps we need to consider applying a reasonable limit on the password length on the client-side. Perhaps 4k?

alanking:
Well, this can be done today with any API which uses a `BytesBuf` (including the authenticate API, which we are talking about). Are we proposing a size limit enforced by PackStruct?

korydraughn:
Good point.

Thinking about this a little deeper, it's fine as is. Given that iRODS is a platform, admins shouldn't expose the iRODS server to untrusted parties without providing some isolation.

I think we're fine. No size limit should be enforced on PackStruct.

korydraughn:
Just so it's clear, that isolation being an app/service designed to enforce whatever constraints the admins deems necessary.

alanking:
Okay. We can of course discuss this more in the future.

alanking:
Still not ready for review, but feel free to have a look.

Since last time:
- Rebased
- Added new test file: test_pam_password.py
- Moved pam_password configurations to `R_GRID_CONFIGURATION`
- Removed existing usage of `password_length` in the database plugin

TODO:
- Pull existing configuration values from server_config (see TODO in database upgrade logic)
- Manually test various upgrades
- Run all the tests
- Maybe write more tests

scripts/irods/test/test_pam_password.py
```diff
@@ -0,0 +1,225 @@
...
+            ['iadmin', 'get_pam_password_config'], 'STDOUT', f'"{option_name}": "{option_value}"')
+
+        # Assert that the value has been set to an empty string.
+        self.admin.assert_icommand(['iadmin', 'set_pam_password_config', option_name])
+        self.admin.assert_icommand(
+            ['iadmin', 'get_pam_password_config'], 'STDOUT', f'"{option_name}": ""')
+
+    def test_really_long_option_value(self):
+        # The input buffer to set_grid_configuration_value API is only 2700 characters long.
+        really_long_option_value = 'this_is_27_characters_long_' * 100
```
trel, line 216: so sending exactly 2700 is the problem?   because 2700 plus the null terminator is 2701?

i think a slightly better comment will make this clear

plugins/auth/src/pam_password.cpp
```diff
@@ -120,6 +123,51 @@ namespace
...
+            }
+
+            if (static_cast<rodsLong_t>(_password.size()) > max_password_length) {
+                THROW(PASSWORD_EXCEEDS_MAX_SIZE,
+                      fmt::format("PAM password cannot exceed [{}] characters.", max_password_length));
+            }
+        }
+        catch (const std::invalid_argument& e) {
+            THROW(CONFIGURATION_ERROR,
+                  "Configured max_password_length for pam_password cannot be converted to an integer.");
```
trel, line 164: `max_password_length` ... or just `password_length` ?

plugins/auth/src/pam_password.cpp
```diff
@@ -120,6 +123,51 @@ namespace
...
+        }
+        catch (const std::out_of_range& e) {
+            THROW(CONFIGURATION_ERROR,
+                  "Configured max_password_length for pam_password not in supported range of values.");
```
trel, line 168: `max_password_length` ... or just `password_length` ?

trel (commented):
looking pretty good.       we planning on making that random string generator a bit less horrible?    or one thing at a time :P

plugins/auth/src/pam_password.cpp
alanking: Yes, need to update the actual config name.

scripts/irods/test/test_pam_password.py
alanking: Right, if you send 2700 characters, packstruct falls over; which is good, but in automated testing, the manner of falling over is interesting... can discuss. I'll update this comment.

scripts/irods/test/test_auth.py
alanking: Done

plugins/auth/src/pam_password.cpp
```diff
@@ -296,6 +344,10 @@ namespace irods
                 return irods_auth::request(*host->conn, req);
             }
 
+            const auto password = req.at(irods::AUTH_PASSWORD_KEY).get_ref<const std::string&>();
```
korydraughn, line 347: `const auto&`

plugins/auth/src/pam_password.cpp
```diff
@@ -313,7 +365,6 @@ namespace irods
             }
 
             const auto username = req.at("user_name").get_ref<const std::string&>();
```
korydraughn, line 367: `const auto&`

plugins/database/src/db_plugin.cpp
```diff
@@ -7225,8 +7219,14 @@ irods::error db_update_pam_password_op(
...
         } // if !irods_pam_auth_no_extend
-        icatDescramble( passwordInIcat );
-        strncpy( *_irods_password, passwordInIcat, irods_pam_password_len );
+
+        // passwordInIcat is the randomly generated password (see while loop below) in a scrambled form.
+        icatDescramble(passwordInIcat);
+
+        // _irods_password is a buffer of size MAX_NAME_LEN which is 1024 + 64. passwordInIcat is MAX_PASSWORD_LEN + 1.
+        // The unscrambled password at time of generation is 50 characters or less (see below).
+        std::strncpy(*_irods_password, passwordInIcat, sizeof(passwordInIcat));
```
korydraughn, line 7228: So the db plugin function doesn't know the size of `*_irods_password` exactly.

Seems like this can result in a buffer overflow if we're not careful.

alanking: Indeed. Do we want to address this? I'm guessing the answer is to add the buffer size as a parameter in the operation signature.

plugins/auth/src/pam_password.cpp
alanking: Good catch

plugins/auth/src/pam_password.cpp
alanking: Good catch

scripts/irods/test/test_pam_password.py
```diff
@@ -0,0 +1,225 @@
...
+            ['iadmin', 'get_pam_password_config'], 'STDOUT', json.dumps(expected_output, indent=4))
+
+    def test_empty_value(self):
+        option_name = 'password_length'
+        option_value = 500
+        self.admin.assert_icommand(['iadmin', 'set_pam_password_config', option_name, str(option_value)])
+
+        # Assert that the value has been updated to the expected new value.
+        expected_output = self.default_setup.copy()
+        expected_output['password_length'] = str(option_value)
```
korydraughn, line 205: You could replace the string literal with `option_name`?

Up to you.

scripts/irods/test/test_pam_password.py
```diff
@@ -0,0 +1,225 @@
...
+        self.admin.assert_icommand(
+            ['iadmin', 'get_pam_password_config'], 'STDOUT', f'"{option_name}": "{option_value}"')
+
+        # Assert that the value has been set to an empty string.
+        self.admin.assert_icommand(['iadmin', 'set_pam_password_config', option_name])
```
korydraughn, line 210: Oh, is NOT passing an empty string a feature of this subcommand?

scripts/irods/test/test_pam_password.py
```diff
@@ -0,0 +1,225 @@
...
+from ..core_file import temporary_core_file
+from ..test.command import assert_command
+
+
+@unittest.skipIf(test.settings.USE_SSL, 'These tests set up SSL so if SSL is already set up, this should be skipped.')
+class test_pam_password(unittest.TestCase):
+    plugin_name = IrodsConfig().default_rule_engine_plugin
+
+    def setUp(self):
+        cfg = lib.open_and_load_json(os.path.join(IrodsConfig().irods_directory, 'test', 'test_framework_configuration.json'))
```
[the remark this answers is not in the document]
alanking: Nope, it's been around for a while, from `test_auth.py` (not sure if it has been used elsewhere).

scripts/irods/test/test_pam_password.py
alanking: Yep, will do

scripts/irods/test/test_pam_password.py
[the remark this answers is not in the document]
korydraughn: Good to know. Thanks.

scripts/irods/test/test_pam_password.py
alanking: Not passing an empty string will actually set the value of the configuration option, like this:
```bash
iadmin set_pam_password_config password_max_length 500
```

Setting the value to nothing by passing an empty string will instruct the system to use the default value, like this:
```bash
iadmin set_pam_password_config password_max_length
```

Happy to try something else if it's more intuitive

korydraughn: It feels a little surprising to not have to provide at least `''` as the empty string.

I'd imagine `argc` in `main` would match 3 in that case. How does iadmin not crash?

alanking: The `cmdToken` array elements are initialized to empty strings here: https://github.com/irods/irods_client_icommands/blob/9d06556c3f0a9c2424bcd07d93889af02d949f2e/src/iadmin.cpp#L1694-L1698

alanking: Ehhh sorta. I think that implicitly this is supported in the sense that an arg can be both required and empty, but I think most subcommands require a non-empty string. Either way, I'm not particularly attached to this approach, so we can do something else here.

How should the admin indicate that the default value should be used? A special string? A negative value?

plugins/database/src/db_plugin.cpp
alanking: At the moment, I don't have any concerns, but I'll take a look into it since I'm touching everything in here anyway

korydraughn: Okay.

scripts/irods/test/test_pam_password.py
alanking: I think "default" seems like a good option. 0 is a valid value for one of these configurations and hyphen is confusing (to me). Thoughts, @trel?

korydraughn: `default` seems fine. There's certainly no room for confusion there.

Will await @trel's thoughts.

trel: feels like empty string is problematic / different from others - and causes more edge cases?

can we just require a value?  and if they want the default, then ... we can document the magic string(s) 'default' and/or 'reset'?     hmmm....

trel: really don't like hyphen or 0 or negative.

plugins/auth/src/pam_password.cpp

plugins/auth/src/pam_password.cpp

plugins/auth/src/pam_password.cpp

plugins/database/src/db_plugin.cpp

scripts/irods/test/test_pam_password.py
alanking: Done

scripts/irods/test/test_pam_password.py
alanking: Still need to look at this. I think we decided that "default" can be sent via `iadmin` which will cause the default value to be stored in `R_GRID_CONFIGURATION`.

alanking:
After some discussion, we have decided that `password_length` will be removed as a configurable option. The randomly generated password will use a hard-coded value of `MAX_PASSWORD_LEN` (50 characters). With the existing protocol, we cannot protect against large amounts of data being sent to the server, so we will just remove the upper limit on passwords sent to the `pam_password` auth plugin.

I pushed the changes that I have so far, but I will be trying the removal on another local branch, probably to be pushed here soon.

Still not ready for review.

scripts/irods/test/test_pam_password.py
alanking: Created https://github.com/irods/irods/issues/7274 for the `R_GRID_CONFIGURATION` changes. That will happen in a different PR.

alanking:
I opened #7274 for the `R_GRID_CONFIGURATION` move. This PR now addresses #7098 alone, which allows for longer PAM passwords and removes the configuration which set a max length on the randomly generated password.

Core tests are running.

korydraughn (approved):
If there's nothing else to do and you're happy with the changes, add the pound.

plugins/database/src/db_plugin.cpp
```diff
@@ -7225,45 +7223,38 @@ irods::error db_update_pam_password_op(
...
-            if ( c < '0' ) {
-                c += '0';
-            }
-            if ( ( c > 'a' && c < 'z' ) || ( c > 'A' && c < 'Z' ) ||
-                    ( c > '0' && c < '9' ) ) {
-                randomPw[j++] = c;
-            }
-        }
-        randomPw[j] = '\0';
+    std::array<char, MAX_PASSWORD_LEN> scrambled_random_password{};
```
alanking, line 7236: Need another + 1 here, I believe. Sorry

korydraughn: The +1s definitely make me nervous.

Good catch though.

alanking:
Unit tests passed. Gonna run through core tests one more time

alanking:
Core tests passed

korydraughn (approved):
Ship it.

Issue #7250, opened 2023-08-03
with ticket activated, data object OPEN locks up on other clients after rcGetHostFor(Put|Get)


d-w-moore:
- [x] main
- [x] 4-3-stable
---
### iRODS Version, OS and Version

iRODS, 4-3-stable (https://github.com/irods/irods/tree/bb7798a61226618e647d7d12700c2121c956e15a)
Ubuntu 18.04

### What did you try to do?

On attempting to implement client redirection to resource server in the python-irodsclient, the following actions caused a lockup:
  1) connection **_A_** activates ticket **_T_** ( allowing access to data object **_D_**) and performs rcGetHostForGet with DastaObjInp initialized for **_D_**
  2) connection **_B_** is made based on parameters of connection **_A_** (hostname could be different, eg based on rcGetHostForGet output, but need not be)
  3) connection **_B_** also activates ticket **_T_**
  4) DATA_OBJ_OPEN_AN is attempted on connection **_B_** (with read flags) to open data object **_D_**

### Expected behavior

Data Open should succeed.

### Observed behavior (including steps to reproduce, if applicable)

On an update to the ticket object row in the ICAT, we lock up during an `_rsGenQuery`.  This shows a DB handle is in conflict between different processes. Note the max uses count must be set on the ticket to reproduce the issue.

Debugging showed the locking up to happen in this call chain, as connection **_B_** is trying to open the data object:

```
rsDataObjOpen
get_data_object_info_for_open
resolve_resource_ hierarchy
file_object_factory
getDataObjInfoIncSpecColl
chl_gen_query_impl
checkCondInputAccess
cmlCheckTicketRestrictions
----------------------------------------
REPO_ROOT/plugins/database/src/mid_level_routines.cpp                                                                 │
  >│1573                status =  cmlExecuteNoAnswerSql(                                                                                                │
   │1574                              "update R_TICKET_MAIN set uses_count=? where ticket_id=?", icss );                                                │
   │1575                if ( status != 0 ) 
```

d-w-moore:
Watch this space for C++ and possibly Python examples for reproducing.

trel:
Debugging session revealed a lack of 'commit' to close out the transaction from connection 1, blocking connection 2.

d-w-moore:
Updated title and description since the issue manifested only when tickets were activated for both sessions. Also, it must be the same ticket, and not two different tickets offering access to the same data object.

d-w-moore:
```
On single-node iRODS system:

[...]

PR #7253, opened 2023-08-03
[#7250] Updating ticket stats now call commit or rollback (main)

korydraughn:
I verified that the unit tests correctly reproduces the issue of #7250.

Will run unit and core tests.

### DIFF
--- a/plugins/database/src/db_plugin.cpp
+++ b/plugins/database/src/db_plugin.cpp
@@ -2427,9 +2427,8 @@ irods::error db_mod_data_obj_meta_op(
                                                 dataSizeString,
                                                 objIdString, &icss );
             if ( status != 0 ) {
-                return ERROR(
-                           status,
-                           "cmlTicketUpdateWriteBytes failed" );
+                //_rollback("chlModDataObjMeta");
+                return ERROR(status, "cmlTicketUpdateWriteBytes failed");
             }
         }
 
@@ -2744,6 +2743,7 @@ irods::error db_reg_data_obj_op(
             errMsg << "no permission to update collection '" << logicalDirName << "'";
             addRErrorMsg( &_ctx.comm()->rError, 0, errMsg.str().c_str() );
         }
+        //_rollback("chlRegDataObj");
         return ERROR( iVal, "" );
     }
     snprintf( collIdNum, MAX_NAME_LEN, "%lld", iVal );
--- a/plugins/database/src/general_query.cpp
+++ b/plugins/database/src/general_query.cpp
@@ -25,14 +25,18 @@
  */
 #include "irods/rodsClient.h"
 #include "irods/icatHighLevelRoutines.hpp"
+#include "irods/irods_logger.hpp"
 #include "irods/private/mid_level.hpp"
 #include "irods/private/low_level.hpp"
 #include "irods/irods_virtual_path.hpp"
+#include "irods/stringOpr.h"
 
 #include <boost/algorithm/string.hpp>
 
 #include <string>
 #include <algorithm>
+
+using log_db = irods::experimental::log::database;
 
 extern int logSQLGenQuery;
 
@@ -2213,8 +2217,23 @@ checkCondInputAccess( genQueryInp_t genQueryInp, int statementNum,
                      genQueryInp.condInput.value[accessIx],
                      /*                  sessionTicket, accessControlHost, icss); */
                      sessionTicket, sessionClientAddr, icss );
-        prevStatus = status;
-        return status;
+
+        if (is_non_empty_string(sessionTicket, sizeof(sessionTicket)) == 1) {
+            if (status < 0) {
+                log_db::error("{}: cmlCheckDataObjId error [{}]. Rolling back database updates.", __func__, status);
+
+                if (const auto ec = cmlExecuteNoAnswerSql("rollback", icss); ec < 0) {
+                    log_db::error("{}: Database rollback error [{}].", __func__, ec);
+                }
+            }
+            else {
+                if (const auto ec = cmlExecuteNoAnswerSql("commit", icss); ec < 0) {
+                    log_db::error("{}: Database commit error [{}].", __func__, ec);
+                }
+            }
+        }
+
+        return prevStatus = status;
     }
 
     if ( collIx >= 0 ) {
--- a/plugins/database/src/mid_level_routines.cpp
+++ b/plugins/database/src/mid_level_routines.cpp
@@ -1545,6 +1545,7 @@ int checkObjIdByTicket( const char *dataId, const char *accessLevel,
                 if ( status != 0 ) {
                     return status;
                 }
+
 #ifndef ORA_ICAT
                 /* do a commit on disconnect if needed */
                 cllCheckPending( "", 2, icss->databaseType );
@@ -1575,6 +1576,7 @@ int checkObjIdByTicket( const char *dataId, const char *accessLevel,
             if ( status != 0 ) {
                 return status;
             }
+
 #ifndef ORA_ICAT
             /* do a commit on disconnect if needed*/
             cllCheckPending( "", 2, icss->databaseType );
@@ -1671,6 +1673,7 @@ cmlTicketUpdateWriteBytes( const char *ticketStr,
     if ( status != 0 ) {
         return status;
     }
+
 #ifndef ORA_ICAT
     /* do a commit on disconnect if needed */
     cllCheckPending( "", 2, icss->databaseType );
--- a/plugins/database/src/mid_level_routines.cpp
+++ b/plugins/database/src/mid_level_routines.cpp
@@ -23,6 +23,7 @@
 #include <string>
 
 #include <boost/filesystem.hpp>
+#include <fmt/format.h>
 
 /* Size of the R_OBJT_AUDIT comment field;must match table column definition */
 #define AUDIT_COMMENT_MAX_SIZE       1000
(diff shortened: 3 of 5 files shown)


unit_tests/src/test_rc_data_obj.cpp
```diff
@@ -1357,3 +1359,146 @@ TEST_CASE("#6600: server no longer changes the openmode from O_WRONLY to O_RDWR"
...
+    }
+
+    namespace ticket = irods::experimental::administration::ticket;
+
+    const auto enable_ticket = [](auto& _conn, const auto* _ticket_string) {
+        TicketAdminInput input{};
+        input.arg1 = const_cast<char*>("session");
+        input.arg2 = const_cast<char*>(_ticket_string);
+        input.arg3 = const_cast<char*>("");
+        REQUIRE(rcTicketAdmin(static_cast<RcComm*>(_conn), &input) == 0);
```
korydraughn, line 1398: Add missing header: ticketAdmin.h

plugins/database/src/mid_level_routines.cpp
```diff
@@ -1543,8 +1546,18 @@ int checkObjIdByTicket( const char *dataId, const char *accessLevel,
                               "update R_TICKET_MAIN set write_file_count=? where ticket_id=?",
                               icss );
                 if ( status != 0 ) {
+                    if (const auto ec = cmlExecuteNoAnswerSql("rollback", icss); ec < 0) {
+                        log_db::error("{}: Rollback error [{}].", __func__, ec);
+                    }
```
korydraughn, line 1551: Consider reading through the impl to make sure this is safe/necessary.

korydraughn:
All unit tests pass.

Running core tests.

korydraughn:
All core tests passed.

unit_tests/src/test_rc_data_obj.cpp
korydraughn: done

plugins/database/src/mid_level_routines.cpp
korydraughn: It appears we aren't supposed to handle commit/rollback in the functions I've modified.

To highlight what I mean, `cmlCheckDataObjOnly` is called in the following db_plugin.cpp functions:
- `db_add_avu_metadata_op`
- `db_del_avu_metadata_op`
- `db_copy_avu_metadata_op`
- `db_mod_ticket_op`
- `db_reg_replica_op`
- `checkAndGetObjectId` _(called by other functions in db_plugin.cpp)_
    - `db_check_and_get_object_id_op`
    - `db_set_avu_metadata_op`
    - `db_copy_avu_metadata_op`

The majority of those functions call `_rollback()` on error which means these changes are incorrect. I think the commit and rollback are supposed to be handled in the calling function.

Assuming what I said is true, the solution to #7250 becomes more difficult because we have to review and possibly modify every function which calls `cmlCheckDataObjOnly`.

Functions calling `cmlTicketUpdateWriteBytes` must also be reviewed. Those functions are:
- `db_update_ticket_write_byte_count_op`
- `db_mod_data_obj_meta_op`

Not good at all.

d-w-moore: Do we anticipate this being difficult, or just very labor intensive ?

korydraughn: I'm close. Tracking down why my test is failing.

korydraughn: Turns out I was creating a ticket with the wrong type in the test.

It's working now.

korydraughn: Now for what I've learned ...

The true cause of the bug was the GenQuery implementation.

When GenQuery is invoked with a ticket enabled on the connection, GenQuery will eventually result in a call to `cmlCheckDataObjId`. `cmlCheckDataObjId` **may** update the ticket stats in the catalog. The issue occurs because GenQuery doesn't commit or rollback the changes.

korydraughn:
@d-w-moore Please verify the changes in this PR resolve the PRC issue.

plugins/database/src/db_plugin.cpp
```diff
@@ -2427,9 +2427,8 @@ irods::error db_mod_data_obj_meta_op(
                                                 dataSizeString,
                                                 objIdString, &icss );
             if ( status != 0 ) {
-                return ERROR(
-                           status,
-                           "cmlTicketUpdateWriteBytes failed" );
+                //_rollback("chlModDataObjMeta");
```
korydraughn, line 2430: The fact that we don't rollback here may be a bug.

I'm going to capture this in an issue as it doesn't affect the issue this PR is attempting to resolve.

plugins/database/src/db_plugin.cpp
```diff
@@ -2744,6 +2743,7 @@ irods::error db_reg_data_obj_op(
             errMsg << "no permission to update collection '" << logicalDirName << "'";
             addRErrorMsg( &_ctx.comm()->rError, 0, errMsg.str().c_str() );
         }
+        //_rollback("chlRegDataObj");
```
korydraughn, line 2746: The fact that we don't rollback here may be a bug.

I'm going to capture this in an issue as it doesn't affect the issue this PR is attempting to resolve.

korydraughn:
All core tests pass.

unit_tests/src/test_rc_data_obj.cpp
```diff
@@ -1357,3 +1360,171 @@ TEST_CASE("#6600: server no longer changes the openmode from O_WRONLY to O_RDWR"
...
+            REQUIRE_NOTHROW(ticket_string =
+                                ticket::client::create_ticket(conn1, ticket::ticket_type::write, data_object.c_str()));
+        }
+
+        irods::at_scope_exit delete_ticket{
+            [&conn1, &ticket_string] { ticket::client::delete_ticket(conn1, ticket_string); }};
+
+        REQUIRE_FALSE(ticket_string.empty());
+
+        // Set a upper limit on the total number of bytes that can be written to the replica.
```
korydraughn, line 1474: "Set a" -> "Set an"

d-w-moore:
> > @d-w-moore Please verify the changes in this PR resolve the PRC issue.
> 
> Will do

Tested atop 95465c9d9e0047e5a96bfcbed388d211fc007d52, and PRC works fine with tickets and client redirect.

unit_tests/src/test_rc_data_obj.cpp
korydraughn: Done

korydraughn:
Squashed.

d-w-moore (approved):
So the actual problem where PRC and #7250 was concerned was in genQuery's  `checkCondInputAccess` funcion not doing a commit/rollback (depending on status value).?

Like the extra comments, we'll get this documentation beast tamed eventually : )

korydraughn:
> So the actual problem where PRC and #7250 was concerned was in genQuery's `checkCondInputAccess` funcion not doing a commit/rollback (depending on status value).?

That's correct.

plugins/database/src/general_query.cpp
```diff
@@ -2213,8 +2216,23 @@ checkCondInputAccess( genQueryInp_t genQueryInp, int statementNum,
                      genQueryInp.condInput.value[accessIx],
                      /*                  sessionTicket, accessControlHost, icss); */
                      sessionTicket, sessionClientAddr, icss );
-        prevStatus = status;
-        return status;
+
+        if (!std::string_view{sessionTicket, sizeof(sessionTicket) - 1}.empty()) {
```
korydraughn, line 2220: This is incorrect.

The constructor causes the call to `.empty()` to always return false even if the range is full of null bytes.

korydraughn: Done

korydraughn:
Core tests passed.

@d-w-moore Please verify the PRC is still happy.

korydraughn:
Squashed.

alanking (commented):
Yep, looks good. I just saw one thing which you may choose to ignore. I wrote the review earlier and forgot to post it :(

plugins/database/src/mid_level_routines.cpp
```diff
@@ -1434,54 +1438,81 @@ int checkObjIdByTicket( const char *dataId, const char *accessLevel,
...
+                                          cVal,
+                                          iVal,
+                                          5, // NOLINT(cppcoreguidelines-avoid-magic-numbers, readability-magic-numbers)
+                                          bindVars,
+                                          icss);
+            if (0 == status) {
+                break;
+            }
+
+            coll_path = coll_path.parent_path();
```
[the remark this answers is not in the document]
korydraughn: Done.

alanking (commented):
I think the comments commit either needs more details in its message or should be squashed into the other commit. Thoughts? Whatever seems most appropriate to folks.

If tests are passing, squash to taste and add #s.

korydraughn:
Will add more words.

korydraughn:
@alanking Please review update to commit message.

korydraughn:
#'d

alanking (approved):
Looks good. Please leave a comment for the new issue(s) to be opened.

plugins/database/src/db_plugin.cpp
korydraughn: Created #7260.

plugins/database/src/db_plugin.cpp
korydraughn: Created #7260.

PR #7261, opened 2023-08-14
[7134] msiCollRsync spurious stacktrace (main)

FifthPotato:
This is caused by rsSyncDataToData performing a checksum on a potentially nonexistent data object. If it doesn't exist, it will toss out a CAT_NO_ROWS_FOUND, which is caught and produces the stacktrace.

### DIFF
--- a/server/api/src/rsDataObjRsync.cpp
+++ b/server/api/src/rsDataObjRsync.cpp
@@ -12,6 +12,9 @@
 
 #include "irods/irods_resource_redirect.hpp"
 
+#define IRODS_FILESYSTEM_ENABLE_SERVER_SIDE_API
+#include "irods/filesystem.hpp"
+
 int
 rsDataObjRsync( rsComm_t *rsComm, dataObjInp_t *dataObjInp,
                 msParamArray_t **outParamArray ) {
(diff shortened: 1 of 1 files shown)


korydraughn:
Have you run this against the unit tests and core tests?

If not, please do.

trel:
does this introduce a time of check time of use window?   i guess we'd still get the CAT_NO_ROWS_FOUND in that case...

FifthPotato:
> Have you run this against the unit tests and core tests?
> 
> If not, please do.

A bit late, but yes, I ran this against both.

alanking:
Did you happen to look into the return code of `irule` before your changes to see if we can add an automated test for this?

FifthPotato:
> Did you happen to look into the return code of `irule` before your changes to see if we can add an automated test for this?

Just checked and nope, irule gives me `0` even when I rebuilt tip-of-main to test it.

server/api/src/rsDataObjRsync.cpp
```diff
@@ -254,23 +257,25 @@ rsRsyncDataToData( rsComm_t *rsComm, dataObjInp_t *dataObjInp ) {
...
-        rodsLog( LOG_ERROR,
-                 "rsRsyncDataToData: _rsDataObjChksum error for %s, status = %d",
-                 dataObjCopyInp.destDataObjInp.objPath, status );
-        free( srcChksumStr );
-        free( destChksumStr );
-        clearKeyVal( &dataObjCopyInp.srcDataObjInp.condInput );
-        clearKeyVal( &dataObjCopyInp.destDataObjInp.condInput );
-        return status;
+    if (irods::experimental::filesystem::server::is_data_object(*rsComm, destObjPath)) {
+        /* use rsDataObjChksum because the path could in in remote zone */
```
alanking, line 261: Please update this comment to... have less typo...

alanking (commented):
Okay, thanks for checking. For posterity, I will say aloud that we do not want to add any more tests which check the logs, so we will not add an automated test for this.

I just saw one last thing and then I think this will be ready for squashing.

alanking (approved):
Looks good to me. Please squash to one but leave off the #s for now and wait for one more approval

korydraughn (approved):
Looks good.

alanking:
Also, make sure to resolve review comments as they are addressed.

PR #7269, opened 2023-08-19
Address Sanitizer fixes (main)

korydraughn:

All leaks fixed by this PR were discovered by running the unit tests with address sanitizer enabled.

We need to do the same thing with the core tests at some point.

### DIFF
--- a/server/api/src/rsRmColl.cpp
+++ b/server/api/src/rsRmColl.cpp
@@ -1,8 +1,7 @@
-/*** Copyright (c), The Regents of the University of California            ***
- *** For more information please refer to files in the COPYRIGHT directory ***/
-/* rsRmColl.c
- */
+#include "irods/rsRmColl.hpp"
 
+#include "irods/irods_at_scope_exit.hpp"
+#include "irods/rcMisc.h"
 #include "irods/rmColl.h"
 #include "irods/objMetaOpr.hpp"
 #include "irods/specColl.hpp"
@@ -18,7 +17,6 @@
 #include "irods/dataObjRename.h"
 #include "irods/genQuery.h"
 #include "irods/miscServerFunct.hpp"
-#include "irods/rsRmColl.hpp"
 #include "irods/rsOpenCollection.hpp"
 #include "irods/rsCloseCollection.hpp"
 #include "irods/rsReadCollection.hpp"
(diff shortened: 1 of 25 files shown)


server/api/src/rsRmColl.cpp
```diff
@@ -746,14 +745,13 @@ rsMvCollToTrash( rsComm_t *rsComm, collInp_t *rmCollInp ) {
         }
     }
 
-    memset( &dataObjRenameInp, 0, sizeof( dataObjRenameInp ) );
+    dataObjCopyInp_t dataObjRenameInp{};
+    irods::at_scope_exit clear_dataObjCopyInp{[&dataObjRenameInp] { clearDataObjCopyInp(&dataObjRenameInp); }};
 
-    dataObjRenameInp.srcDataObjInp.oprType =
-        dataObjRenameInp.destDataObjInp.oprType = RENAME_COLL;
+    dataObjRenameInp.srcDataObjInp.oprType = dataObjRenameInp.destDataObjInp.oprType = RENAME_COLL;
```
trel, line 751: on two lines instead of one?

optimize readability...

korydraughn: Done

korydraughn:
Addressed a few more leaks in the icommands.

Rerunning unit tests and core tests.

korydraughn:
All unit tests passed.

Running core tests.

korydraughn:
All core tests passed.

server/core/src/irods_file_object.cpp
```diff
@@ -156,7 +156,6 @@ namespace irods {
         repl_requested_ = _rhs.repl_requested_;
         replicas_       = _rhs.replicas_;
         in_pdmo_        = _rhs.in_pdmo_;
-        replKeyVal( &_rhs.cond_input_, &cond_input_ );
```
alanking, line 159: Is this being handled by the base class `operator=`? If not, are there any known effects of not copying the `cond_input_` here?

alanking (commented):
Everything checks out. Had a question about one edit

server/core/src/irods_file_object.cpp
korydraughn: It's handled by the base class.

That line was causing the work to be done twice, which resulted in a leak.

alanking: Great, thanks

korydraughn:
Added 6634 as the issue number.

Any last comments before I add the pounds?

alanking (commented):
Any reason these can't be squashed to fewer commits? I see three instances where there are identical commit messages one after the other. I don't feel strongly about this, just wondering

korydraughn:
They can be squashed. I kept them separate to highlight individual fixes.

alanking (approved):
I leave it up to you or anybody else who wants to throw in an opinion, but I think this is fine. Add #s when ready, I say

korydraughn:
Ok. I'll squash them. The changes are tiny and keeping them separate doesn't benefit anyone.

trel:
Just easier to revert if kept separate.

Issue #7254, opened 2023-08-07
Give non-C/C++ clients a way to determine the library features provided by an iRODS server


korydraughn:
- [x] main
- [x] 4-3-stable
---
## Feature
With the release of iRODS 4.3.1 comes support for detecting library features at compile-time. This is great for applications written in C and C++, but what about other languages?

The iRODS server needs to provide a way for Java, Python, Go, etc to detect what's provided.

Perhaps this is achieved via a new API endpoint. For example:
```python
with iRODSSession(...) as sess:
    # Invokes rc_library_features() or something equivalent
    # and returns a JSON string/object or dict.
    features = sess.library_features()

    # features is a JSON object or dict in python.
    if features.get('IRODS_HAS_API_ENDPOINT_SWITCH_USER', 0) > 202307:
        # Do something interesting.
```

_As of right now, this should be limited to whatever's defined in **library_features.h**._

PR #7273, opened 2023-08-21
[#7254] Implemented new API endpoint: rc_get_library_features (main)

korydraughn:
This API endpoint does not modify any server state.

Running the unit test for this endpoint is enough to confirm whether it's good for merging.


trel:
commit message has "the the"

trel:
so... is this a new source of truth?

korydraughn:
No. It is supposed to match the contents of library_features.h.

trel:
so it can drift from library_features.h?    can't we define/respond based on what's defined over there?

korydraughn:
We'd have to build similar machinery to the irods error map stuff to achieve that. Which is certainly an option.

trel:
Yeah, not feeling good about a list that we know will continue to grow having to be kept in sync.

server/api/src/rs_get_library_features.cpp
```diff
@@ -0,0 +1,34 @@
...
+#define IRODS_FEATURE(F)        {#F, IRODS_FEATURE_VALUE(F)}, // NOLINT(cppcoreguidelines-macro-usage)
+
+    // clang-format off
+    *_features = strdup(nlohmann::json{
+        IRODS_FEATURE(IRODS_HAS_LIBRARY_TICKET_ADMINISTRATION)
+        IRODS_FEATURE(IRODS_HAS_LIBRARY_ZONE_ADMINISTRATION)
+        IRODS_FEATURE(IRODS_HAS_API_ENDPOINT_SWITCH_USER)
+        IRODS_FEATURE(IRODS_HAS_LIBRARY_SYSTEM_ERROR)
+        IRODS_FEATURE(IRODS_HAS_FEATURE_PROXY_USER_SUPPORT_FOR_CLIENT_CONNECTION_LIBRARIES)
+        IRODS_FEATURE(IRODS_HAS_API_ENDPOINT_CHECK_AUTH_CREDENTIALS)
```
korydraughn, line 29: I really wish this could be generated at compile-time.

Is it possible? Hmm ...

korydraughn:
> Yeah, not feeling good about a list that we know will continue to grow having to be kept in sync.

This may not be possible with just C++. We may need an alternative solution to achieve what you're asking.

korydraughn:
One way to handle this is to parse the library_features.h header file and produce something the API endpoint can consume.

That would give us an opportunity to embed a compile-time JSON string into the API implementation. Of course that means the build system has to launch the parser for us.

Perhaps this is an opportunity to learn the clang-tools for parsing the C++ language.

korydraughn:
Created issue #7276.

korydraughn:
The clang-tidy issues are harmless and cannot be fixed easily at this time.

Will add the pound if there are no other concerns/comments.

lib/api/include/irods/get_library_features.h
```diff
@@ -0,0 +1,28 @@
...
+struct RcComm;
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/// Returns a JSON string containing information about the connected server's library features.
+///
+/// \param[in]  _comm     A pointer to a RcComm.
+/// \param[out] _features A pointer to a pointer that will hold the results.
```
alanking, line 15: Worth describing the structure of the JSON string?

alanking (commented):
Sorry, didn't realize this was out of draft.

I think this is fine, but just wondering aloud whether this could be part of a new and improved `rcGetMiscSvrInfo`. Just an idea, not attached to it.

korydraughn:
> I think this is fine, but just wondering aloud whether this could be part of a new and improved `rcGetMiscSvrInfo`. Just an idea, not attached to it.

It could. My only concern is the JSON output of this new endpoint. It has the potential to grow very large over time.

Overall, the idea doesn't sound bad. Consider capturing it in a new issue and we can circle back to it after 4.3.1 is released.

lib/api/include/irods/get_library_features.h
korydraughn: Done.

alanking:
Okay, opened an issue here: https://github.com/irods/irods/issues/7278

korydraughn:
Cool. Any thoughts on the updated documentation?

alanking (approved):
Look good to me. Please squash and add # when happy with it

korydraughn:
Added #

alanking (approved):

PR #7294, opened 2023-08-29
Leader conflict draft

FifthPotato:
This appears to fail tests, though I'm baffled as to why. Putting this here to get more eyes on it, maybe?
The last run I had, these tests fail: 

EDIT: 
```
        test_resource_types.Test_Resource_Replication test_resource_types.Test_Resource_ReplicationToTwoCompoundResourcesWithPreferArchive test_resource_types.Test_Resource_WeightedPassthru test_rule_engine_plugin_framework test_rule_engine_plugin_passthrough test_rulebase test_specific_queries test_ssl test_symlink_operations test_targeting_specific_replica_number
```

Observed behaviors:
Commenting out line 845 does not appear to fix it; simply calling the function in the loop causes the issue
Commenting out lines 695 - 719 (and manually returning true) **does** appear to fix the issue
The function never returns false during testing.

### DIFF
--- a/server/delay_server/src/irodsDelayServer.cpp
+++ b/server/delay_server/src/irodsDelayServer.cpp
@@ -4,6 +4,7 @@
 #include "irods/connection_pool.hpp"
 #include "irods/fully_qualified_username.hpp"
 #include "irods/get_delay_rule_info.h"
+#include "irods/get_grid_configuration_value.h"
 #include "irods/initServer.hpp"
 #include "irods/irods_at_scope_exit.hpp"
 #include "irods/irods_client_api_table.hpp"
@@ -674,6 +675,47 @@ namespace
 
         return {qstr, job};
     } // make_delay_queue_query_processor
+
+    auto is_local_server_defined_as_delay_server_leader() -> bool
+    {
+        std::string hostname;
+        //NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays, modernize-avoid-c-arrays)
+        char hn[HOST_NAME_MAX + 1]{};
+
+        if (const auto err = gethostname(hn, sizeof(hn)); err != 0) {
+            logger::delay_server::error("{}: Failed to retrieve local server's hostname. Error {}", __func__, err);
+            return false;
+        }
+
+        if (const auto hn_resolved = resolve_hostname(hn, hostname_resolution_scheme::match_preferred); hn_resolved) {
+            hostname = *hn_resolved;
+        }
+        else {
+            hostname = hn;
+        }
+        irods::experimental::client_connection conn;
+        GridConfigurationInput input{};
+        std::strcpy(input.name_space, "delay_server");
+        std::strcpy(input.option_name, "leader");
+
+        GridConfigurationOutput* output{};
+        //NOLINTNEXTLINE(cppcoreguidelines-no-malloc, cppcoreguidelines-owning-memory)
+        irods::at_scope_exit free_output{[&output] { std::free(output); }};
+
+        if (const auto ec = rc_get_grid_configuration_value(static_cast<RcComm*>(conn), &input, &output); ec != 0) {
+            logger::delay_server::warn("{}: Failed to retrieve leader config option. Error {}", __func__, ec);
+            return false;
+        }
+
+        if (const auto hn_resolved =
+                resolve_hostname(output->option_value, hostname_resolution_scheme::match_preferred);
+            hn_resolved)
+        {
+            return hostname == *hn_resolved;
+        }
+        return hostname == output->option_value;
+    } //is_local_server_defined_as_delay_server_leader
+
 } // anonymous namespace
 
 int main(int argc, char** argv)
@@ -794,7 +836,10 @@ int main(int argc, char** argv)
         while (!delay_server_terminated) {
             try {
                 irods::server_properties::instance().capture();
-
+                if (!is_local_server_defined_as_delay_server_leader()) {
+                    logger::delay_server::warn("This server is not the leader. Terminating...");
+                    break;
+                }
                 auto delay_queue_processor = make_delay_queue_query_processor(thread_pool, queue);
 
                 logger::delay_server::trace("Gathering rules for execution ...");


server/delay_server/src/irodsDelayServer.cpp
```diff
@@ -676,6 +677,51 @@ namespace
...
+    {
+        GridConfigurationInput input{};
+        std::strcpy(input.name_space, "delay_server");
+        std::strcpy(input.option_name, "leader");
+
+        GridConfigurationOutput* output{};
+        irods::at_scope_exit free_output{[&output] { std::free(output); }};
+
+        if (const auto ec = rc_get_grid_configuration_value(static_cast<RcComm*>(conn), &input, &output); ec != 0) {
+            logger::delay_server::warn("is_current_server_leader: Failed to retriever leader config option.");
```
trel, line 706: retriever -> retrieve

server/delay_server/src/irodsDelayServer.cpp
```diff
@@ -676,6 +677,51 @@ namespace
...
+        result = output->option_value;
+    }
+
+    std::string leader;
+    if (const auto hn_resolved = resolve_hostname(result, hostname_resolution_scheme::match_preferred); hn_resolved) {
+        leader = *hn_resolved;
+    }
+    else {
+        leader = result;
+    }
```
trel, line 719: Do we need to resolve the leader value?   It's just coming from the database... so, that's a setting, not something that is coming from the OS that needs to have bumpers applied...?

Does this behave if we always set `leader = result`?

server/delay_server/src/irodsDelayServer.cpp
```diff
@@ -676,6 +677,51 @@ namespace
...
+        if (const auto ec = rc_get_grid_configuration_value(static_cast<RcComm*>(conn), &input, &output); ec != 0) {
+            logger::delay_server::warn("is_current_server_leader: Failed to retriever leader config option.");
+            return false;
+        }
+
+        result = output->option_value;
+    }
+
+    std::string leader;
+    if (const auto hn_resolved = resolve_hostname(result, hostname_resolution_scheme::match_preferred); hn_resolved) {
```
korydraughn, line 714: Should we check to see if `result` is empty?

server/delay_server/src/irodsDelayServer.cpp
```diff
@@ -676,6 +677,51 @@ namespace
     } // make_delay_queue_query_processor
 } // anonymous namespace
 
+bool is_current_server_leader(void) {
+    std::string hostname;
+    char hn[HOST_NAME_MAX + 1]{};
+
+    if (gethostname(hn, sizeof(hn)) != 0) {
+        logger::delay_server::warn("is_current_server_leader: Failed to retrieve local server's hostname.");
```
korydraughn, line 685: You can use `__func__` as an argument to avoid having to manually type the function name.
```cpp
logger::delay_server::warn("{}: Failed to retrieve local server's hostname.", __func__);
```

server/delay_server/src/irodsDelayServer.cpp
```diff
@@ -676,6 +677,51 @@ namespace
...
+        return false;
+    }
+
+    if (const auto hn_resolved = resolve_hostname(hn, hostname_resolution_scheme::match_preferred); hn_resolved) {
+        hostname = *hn_resolved;
+    }
+    else {
+        hostname = std::string{hn};
```
korydraughn, line 693: `hostname = hn`?

server/delay_server/src/irodsDelayServer.cpp
```diff
@@ -676,6 +677,51 @@ namespace
...
+    }
+    auto conn = get_new_connection(std::nullopt);
```
korydraughn, line 695: Consider constructing the client connection directly.

That makes the code clearer and avoids the unnecessary checks in `get_new_connection`. It also guards against `get_new_connection` changing in behavior and possibly breaking your code.
```cpp
irods::experimental::client_connection conn;
```

server/delay_server/src/irodsDelayServer.cpp
```diff
@@ -676,6 +677,51 @@ namespace
...
+    std::string result;
+    {
```
korydraughn, line 697: How about we drop the additional scope (i.e. remove curly braces) and move `std::string result;` to where it's used.

The additional scoping isn't really necessary here.

Perhaps we don't even need `result` ...

server/delay_server/src/irodsDelayServer.cpp
```diff
@@ -676,6 +677,51 @@ namespace
...
+    std::string leader;
+    if (const auto hn_resolved = resolve_hostname(result, hostname_resolution_scheme::match_preferred); hn_resolved) {
+        leader = *hn_resolved;
+    }
+    else {
+        leader = result;
+    }
+
+    return hostname == leader;
+}
```
korydraughn, line 722: As a convention, we normally add a trailing comment after the closing brace of a function. The comment contains the function's name.
```cpp
} // is_current_server_leader
```
This helps in situations where function implementations are large and lead to confusion around where you are located.

server/delay_server/src/irodsDelayServer.cpp
```diff
@@ -676,6 +677,51 @@ namespace
     } // make_delay_queue_query_processor
 } // anonymous namespace
 
+bool is_current_server_leader(void) {
```
korydraughn, line 680: `void` isn't necessary here.

korydraughn, line 680: Let's move this into the anonymous namespace just above.

korydraughn (commented):
Nothing unusual about the implementation.

Is there anything interesting about how you run the tests?
Have you tried running the tests that fail in isolation?

BTW, I'm running the tests you reported as failures against your PR one by one to see what happens.

korydraughn:
So far, `test_resource_types.Test_Resource_Random` passed against your PR.

Trying the other tests you reported now.

korydraughn:
> List of failed tests: test_resource_types.Test_Resource_Random test_resource_types.Test_Resource_Replication test_resource_types.Test_Resource_ReplicationToTwoCompound test_resource_types

Wait, tell me about that last item here? `test_resource_types`

I'm assuming there should be more letters attached to that.

korydraughn:
`test_resource_types.Test_Resource_Replication` and `test_resource_types.Test_Resource_ReplicationToTwoCompound` also passed.

server/delay_server/src/irodsDelayServer.cpp
korydraughn: The resolve_hostname function is using what used to be the hosts_config.json file to obtain a valid iRODS hostname.

I don't think we can remove that call because we can't guarantee the host resolution information is always available on every server.

FifthPotato: To add onto this, I mostly lifted this from rodsServer...if the rationale is the same as it is there, this should be fine?

server/delay_server/src/irodsDelayServer.cpp
FifthPotato: Before it is a check to see if `rc_get_grid_configuration_value` failed, and then after it is a check to see if `resolve_hostname` fails...I think that catches both cases where something could go awry; either `resolve_hostname` returns `std::nullopt` and it fails, or it was already broken before it...so I think it might be OK as is?

server/delay_server/src/irodsDelayServer.cpp
FifthPotato: Oops, knew about `__func__` but didn't know the logging was formatted. Fixing!

server/delay_server/src/irodsDelayServer.cpp
FifthPotato: Lifted from rodsServer as well, but I agree; should be fine for it to be freed end-of-function instead of this scope. Fixing!

server/delay_server/src/irodsDelayServer.cpp
FifthPotato: Adding!

FifthPotato:
> > List of failed tests: test_resource_types.Test_Resource_Random test_resource_types.Test_Resource_Replication test_resource_types.Test_Resource_ReplicationToTwoCompound test_resource_types
> 
> Wait, tell me about that last item here? `test_resource_types`
> 
> I'm assuming there should be more letters attached to that.

My bad, got messed up by wordwrap and me being bad at pasting. Updated in original post!

alanking:
Please link an issue number either in the commit message (omit the #) or in a comment (don't need to omit the #).

FifthPotato:
I re-tested with only 4 executors and it appears to have solved many of the test failures; 3 tests still fail though, but they also fail on tip-of-main for me. Is that "good enough" or should I aim for an all-tests-passed message?

List of failed tests:
        test_auth test_ibun test_icd

korydraughn:
Try rerunning the failed tests again if you haven't already.

test_auth.py has a real bug. Ignore that one for now. @alanking and I are looking into that one.

korydraughn:
I just searched those files for "delay" and "log" and nothing related to your changes appeared.

Please take out of draft, squash commits to taste, and leave off the pounds.

Don't forget to resolve comments if they have been addressed.

korydraughn:
Oh, please address clang-format/tidy github action reports.

server/delay_server/src/irodsDelayServer.cpp
```diff
@@ -674,6 +675,46 @@ namespace
...
+        irods::experimental::client_connection conn;
+        GridConfigurationInput input{};
+        std::strcpy(input.name_space, "delay_server");
+        std::strcpy(input.option_name, "leader");
+
+        GridConfigurationOutput* output{};
+        irods::at_scope_exit free_output{[&output] { std::free(output); }};
+
+        if (const auto ec = rc_get_grid_configuration_value(static_cast<RcComm*>(conn), &input, &output); ec != 0) {
+            logger::delay_server::warn("{}: Failed to retrieve leader config option.", __func__);
```
alanking, line 703: Consider including the error code in the message as well.

server/delay_server/src/irodsDelayServer.cpp
```diff
@@ -674,6 +675,46 @@ namespace
 
         return {qstr, job};
     } // make_delay_queue_query_processor
+
+    bool is_current_server_leader() {
+        std::string hostname;
+        char hn[HOST_NAME_MAX + 1]{};
+
+        if (gethostname(hn, sizeof(hn)) != 0) {
+            logger::delay_server::warn("{}: Failed to retrieve local server's hostname.", __func__);
```
alanking, line 684: Should we include `errno` in the message?

server/delay_server/src/irodsDelayServer.cpp
```diff
@@ -674,6 +675,46 @@ namespace
...
+
+        std::string leader;
+        if (const auto hn_resolved = resolve_hostname(output->option_value, hostname_resolution_scheme::match_preferred); hn_resolved) {
+            leader = *hn_resolved;
+        }
+        else {
+            leader = output->option_value;
+        }
+
+        return hostname == leader;
```
alanking, line 715: Maybe could avoid the extra string here like this:
```suggestion
        if (const auto hn_resolved = resolve_hostname(output->option_value, hostname_resolution_scheme::match_preferred); hn_resolved) {
            return hostname == *hn_resolved;
        }
        
        return hostname == output->option_value;
```
But, not as "self-documenting". Just a thought.

alanking (commented):
I know it's still in draft but here's a review anyways :) Looks good overall

server/delay_server/src/irodsDelayServer.cpp
korydraughn: Given the changes to this PR since this conversation happened, this can be marked as resolved.

server/delay_server/src/irodsDelayServer.cpp
korydraughn: Agreed. Marking as resolved.

server/delay_server/src/irodsDelayServer.cpp
```diff
@@ -674,6 +675,47 @@ namespace
...
     } // make_delay_queue_query_processor
+
+    auto is_current_server_leader() -> bool
+    {
+        std::string hostname;
+        //NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays, modernize-avoid-c-arrays)
+        char hn[HOST_NAME_MAX + 1]{};
+
+        if (const auto err = gethostname(hn, sizeof(hn)); err != 0) {
+            logger::delay_server::warn("{}: Failed to retrieve local server's hostname. Error {}", __func__, err);
```
korydraughn, line 686: Getting an error from `gethostname` seems bad, perhaps the log message should be considered an error rather than a warning.

Thoughts?

korydraughn (commented):
Other than my one comment, the code looks good.

server/delay_server/src/irodsDelayServer.cpp
trel: Elsewhere in our code, I see it both as an LOG_ERROR, and a LOG_NOTICE.

I'm leaning ERROR...

server/delay_server/src/irodsDelayServer.cpp
```diff
@@ -674,6 +675,47 @@ namespace
 
         return {qstr, job};
     } // make_delay_queue_query_processor
+
+    auto is_current_server_leader() -> bool
```
trel, line 679: 'current' server... perhaps... 'local' server instead?

`is_local_server_leader()`
or
with one more word...
`is_local_server_the_leader()`

server/delay_server/src/irodsDelayServer.cpp
```diff
@@ -794,7 +836,10 @@ int main(int argc, char** argv)
         while (!delay_server_terminated) {
             try {
                 irods::server_properties::instance().capture();
-
+                if (!is_current_server_leader()) {
+                    logger::delay_server::warn("This process is not the leader. Terminating...");
```
trel, line 840: process -> server ?

trel:
this is pretty surgical and good.    i like it.

server/delay_server/src/irodsDelayServer.cpp
trel: or `is_set_as_delay_server_leader`

alanking: I find the implication introduced by removing "current" or "local" from the name a tad confusing. I like `is_local_server_defined_as_delay_server_leader` or something like that. Just my two cents

korydraughn: That sounds fine.

Let's see what @FifthPotato thinks.

trel: yeps `is_local_server_defined_as_delay_server_leader` has all the wordses.

FifthPotato: Guess I'll add all the wordses!

trel (approved):
i'm happy with this.

server/delay_server/src/irodsDelayServer.cpp
```diff
@@ -18,6 +18,7 @@
 #include "irods/irods_server_state.hpp"
 #include "irods/json_deserialization.hpp"
 #include "irods/json_serialization.hpp"
+#include "irods/get_grid_configuration_value.h"
```
korydraughn, line 21: Please alphabetize.

korydraughn (approved):
Once the header includes are sorted, take out of draft and add the pound.

Issue #7296, opened 2023-08-29
Process stash 'erase_if' function should indicate number of elements removed


MartinFlores751:
- [x] main
- [x] 4-3-stable
---
## Feature

https://github.com/irods/irods/blob/21253de25be97fdc5ca6c458fab0759c0ee9cfc2/lib/core/src/process_stash.cpp#L59-L66

Currently, `process_stash::erase_if(...)` does not give any indication of elements being removed from the stash.

Consider returning the number of elements removed, which is returned by `std::erase_if`.

korydraughn:
Don't forget to update the documentation once this change is implemented.

trel:
Yes, this seems very nice and expected.

MartinFlores751:
https://github.com/irods/irods/blob/21253de25be97fdc5ca6c458fab0759c0ee9cfc2/lib/core/src/process_stash.cpp#L53-L57

While working on the code, I noticed that we could update `process_stash::erase(...)` to return a boolean, indicating if an element was removed or not.

After a brief exchange, @korydraughn agreed we should update this function in this issue.

alanking:
@MartinFlores751 - Please close if complete. Thanks

PR #7301, opened 2023-09-01
[#7296] Have process stash `erase_if` and `erase` return indication of change

MartinFlores751:
- `process_stash::erase_if(..)` now returns the number of elements removed.
- `process_stash::erase(...)` now returns a boolean, indicating if an element was removed.


MartinFlores751:
Ran unit tests, `irods_parallel_transfer_engine` failed. Not sure if related?

korydraughn:
It's not related. Ignore that failure.

korydraughn (commented):
Changes look fine. Please squash to taste and leave off the pounds.

unit_tests/src/test_process_stash.cpp
```diff
@@ -53,15 +53,22 @@ TEST_CASE("process_stash basic operations")
     CHECK(boost::any_cast<pod&>(*p).y == a_pod.y);
 
     // Show that the handles can be used to erase data in the stash.
-    irods::process_stash::erase(h1);
+    auto was_removed{irods::process_stash::erase(h1)};
+    CHECK(was_removed);
```
alanking, line 57: Could the `erase` call be put directly into `CHECK`?

alanking (commented):
One "improvement" if you're into that sorta thing. Otherwise looks good

unit_tests/src/test_process_stash.cpp
korydraughn: I had the same thought :-)

MartinFlores751: Can remove the `was_removed`. 👍

alanking (approved):
Add #s after one more approval

unit_tests/src/test_process_stash.cpp
```diff
@@ -91,7 +94,8 @@ TEST_CASE("process_stash::erase_if")
     REQUIRE(handles.size() == 3);
 
     // Show that the handles can be erased using a predicate.
-    irods::process_stash::erase_if([&h1, &h3](auto&& _k, auto&&) { return _k == h1 || _k == h3; });
+    auto num_deleted{irods::process_stash::erase_if([&h1, &h3](auto&& _k, auto&&) { return _k == h1 || _k == h3; })};
+    CHECK(num_deleted == 2);
```
korydraughn, line 98: Same thing here. Let's go ahead and collapse these lines into one too.

lib/core/include/irods/process_stash.hpp
```diff
@@ -70,8 +73,10 @@ namespace irods::process_stash
     /// If \p _pred returns \p true, the entry is removed.
     /// \endparblock
     ///
+    /// \returns The number of elements removed.
+    ///
     /// \since 4.3.1
-    auto erase_if(const std::function<bool(const std::string&, const boost::any&)>& _pred) -> void;
+    auto erase_if(const std::function<bool(const std::string&, const boost::any&)>& _pred) -> std::unordered_map<std::string, boost::any>::size_type;
```
korydraughn, line 79: Let's just return `std::size_t`. This exposes an implementation detail which doesn't need to be known by users of the lib.

Consider what happens if we switch the underlying container.

korydraughn (commented):
Looks good. Squash to taste and leave off pounds.

korydraughn (approved):
Add the pound when ready!

alanking (approved):

PR #7305, opened 2023-09-05
More connection pool updates / New Database column

korydraughn:


plugins/database/src/db_plugin.cpp
```diff
@@ -3936,19 +3947,20 @@ irods::error db_reg_resc_op(
...
+    cllBindVarCount = 13;
 
     if ( logSQL != 0 ) {
         log_sql::debug("chlRegResc SQL 4");
     }
-    status =  cmlExecuteNoAnswerSql(
-                  "insert into R_RESC_MAIN (resc_id, resc_name, zone_name, resc_type_name, resc_class_name, resc_net, resc_def_path, create_ts, modify_ts, resc_children, resc_context, resc_parent) values (?,?,?,?,?,?,?,?,?,?,?,?)",
-                  &icss );
+    status = cmlExecuteNoAnswerSql(
+        "insert into R_RESC_MAIN (resc_id, resc_name, zone_name, resc_type_name, resc_class_name, resc_net, resc_def_path, create_ts, modify_ts, modify_ts_millis, resc_children, resc_context, resc_parent) values (??,?,?,?,?,?,?,?,?,?,?,?)",
```
trel: i think the new column will be at the end?

scripts/irods/database_upgrade.py
```diff
@@ -152,6 +152,13 @@ def run_update(irods_config, cursor):
                         ['mysql', '='.join(['--defaults-file', f.name]), irods_config.database_config['db_name']],
                         stdin=sql_file)
 
+        # Add a new column that will be responsible for holding the fractional part of the modify_ts
+        # for a resource. The value stored will be in milliseconds.
+        if irods_config.catalog_database_type == 'oracle':
+            database_connect.execute_sql_statement(cursor, "alter table R_RESC_MAIN add (modify_ts_millis varchar(3) default '0');")
+        else:
+            database_connect.execute_sql_statement(cursor, "alter table R_RESC_MAIN add column modify_ts_millis varchar(3) default '0';")
```
trel, line 160: these are not default '000'

plugins/database/src/db_plugin.cpp
korydraughn: Order doesn't matter. The inputs simply need to be passed in the correct order.

trel: Oh?  I always thought inserts needed to be ordered.    Okay.

Ah, internet says if you name the variables, the values just have to match the named order.    If you don't name the variables, and only pass the values in, then the table order is what is followed.  Got it.

scripts/irods/database_upgrade.py
korydraughn: Done.

lib/core/src/connection_pool.cpp
```diff
@@ -169,23 +169,32 @@ namespace irods
...
             std::uint64_t latest_mtime = 0;
+            std::uint16_t latest_mtime_millis = 0;
+            std::int32_t resc_count = 0;
 
             // Capture the latest time any resource was modified.
             // This will be used to track when a connection should be refreshed.
             // This helps with long-running agents.
-            for (auto&& row : irods::query{conn_ctxs_[0].conn.get(), "select max(RESC_MODIFY_TIME)"}) {
+
+            constexpr const char* gql = "select max(RESC_MODIFY_TIME), max(RESC_MODIFY_TIME_MILLIS), count(RESC_ID)";
```
trel, line 179: what is the thinking here?   this will get the largest millis in the database... is that sufficient/good/desired?

`max(RESC_MODIFY_TIME_MILLIS)`

trel: i guess we're only detecting if anything moved... so... max is fine...

plugins/database/src/general_query_setup.cpp
```diff
@@ -225,6 +226,7 @@ icatGeneralQuerySetup() {
     sColumn( COL_R_RESC_STATUS, "R_RESC_MAIN", "resc_status" );
     sColumn( COL_R_CREATE_TIME, "R_RESC_MAIN", "create_ts" );
     sColumn( COL_R_MODIFY_TIME, "R_RESC_MAIN", "modify_ts " );
+    sColumn( COL_R_MODIFY_TIME_MILLIS, "R_RESC_MAIN", "modify_ts_millis" );
```
trel, line 229: does this appear in `iadmin attrs` automatically as well?

lib/core/src/connection_pool.cpp
korydraughn: Correct. We can't concat anything using GenQuery and I'm trying to keep the number of operations down to a minimum.

I have another idea that may not require the need to track the number of resources at all.

Will let you know how it goes.

trel: gotcha.

plugins/database/src/db_plugin.cpp
```diff
@@ -126,6 +126,17 @@ static const auto intermediate_replica_status_str = std::to_string(INTERMEDIATE_
...
 
+auto get_time() -> std::pair<std::string, std::string>
+{
+    using std::chrono::duration_cast;
+
+    const auto now = std::chrono::system_clock::now();
+    const auto secs = duration_cast<std::chrono::seconds>(now.time_since_epoch());
+    const auto millis = duration_cast<std::chrono::milliseconds>(now.time_since_epoch()) - secs;
+
+    return {fmt::format("{:011}", secs.count()), fmt::format("{:03}", millis.count())};
```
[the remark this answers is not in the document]
korydraughn: > is `:011` left-zero padded?

Yes.

> is that what we want? maybe a comment here to explain what and why.

Yes. That is how we've represented time in the database for as long as I've been here.

Will add a comment.

plugins/database/src/general_query_setup.cpp
korydraughn: The GenQuery name does.

trel: perfect - thanks.

plugins/database/src/db_plugin.cpp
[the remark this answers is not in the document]
trel: got it - yes - thanks.

korydraughn:
The current solution in this PR doesn't solve the issue I'm seeing in the HTTP API.

I think we have to store the seconds and milliseconds in the same column. Having the milliseconds be a separate column means we can't use SQL aggregate functions to properly detect changes across resources. The primary reason why the presented solution fails is because the milliseconds column wraps.

Alternative ideas welcome.

korydraughn:
Maybe a specific query can help here. Hmm ...

korydraughn:
And here's the specific query which may be the missing piece.
```sql
select max(modify_ts || '.' || modify_ts_millis) from R_RESC_MAIN;
```

korydraughn:
Here's the more portable version.
```sql
select max(concat(modify_ts, concat('.', modify_ts_millis))) mtime from R_RESC_MAIN;
```

trel:
Specific query OR special endpoint in the server... but...

Can't we just wrap the millis check inside the seconds check?   And then we'd detect any wrapping correctly?

korydraughn:
> Can't we just wrap the millis check inside the seconds check? And then we'd detect any wrapping correctly?

I don't understand what you mean. Please elaborate.

korydraughn:
The specific query solved the issue and even allows me to delete code. Nice!

korydraughn:
After some discussion, there are a couple ways to move forward.
- Add a specific query to the 4.3.1 server
- Add a new API endpoint that does what the specific query does

Aternative idea provided by @trel.
- Investigate sorting and retrieving the top _N_ results

trel:
As I understand it... GenQuery1 cannot do `LIMIT` or `ORDER(TOKEN1, TOKEN2)`, so I think we we're back to either...

 - specific query in the server
 - new api, with query baked in

or

 - download 'all' resource timestamps (and then sort and count them locally)

the last one is easiest with fewest moving parts, just might be a tad slower.   but the network back and forth will dwarf the 'bigger' data payload being transmitted.

i think i'm leaning towards `SELECT ORDER_DESC(RESC_MODIFY_TIME), ORDER_DESC(RESC_MODIFY_TIME_MILLIS)"` and be done with it.

korydraughn:
Another alternative is to make the new column large enough to hold the mtime (seconds) and milliseconds. Meaning the seconds would be duplicated in the database. That would allow GenQuery1 to do what the specific query does.

trel:
> make the new column large enough to hold the mtime (seconds) and milliseconds 

Could work... but introduces a possibility of inconsistency/drift (two sources of truth).

trel:
Just to make sure we're on the same page...

The vanilla GenQuery would return a sorted full listing of two columns.

The only processing would be to concatenate/process the first in the list (to get the full timestamp), and count the rows to have the 'total'.

Most Zones have less than 20 resources and the largest we know of has ~300.

korydraughn:
I think I'm getting a little lost in the details, but you're right.

If the query iterator returns the number of rows even when paging the resultset, then I think we're good.

Will give it a try and report back.

plugins/database/src/db_plugin.cpp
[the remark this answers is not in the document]
korydraughn: Updated. See documentation comments just above the function.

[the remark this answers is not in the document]
trel: perfect - thanks.

korydraughn:
The GenQuery only approach doesn't report the total number of rows matched. It returns the total rows in the page.

We have to run another query to get the resource count, which is probably okay.

There's an option provided by GenQuery1 which instructs it to return the total rows in the resultset. If we can expose that through the query iterator's interface, then we can avoid the need of an extra query.

I'm going to see if that works. Will report back soon.

korydraughn:
Modifying the query iterator implementation so we can pass GenQuery1 options to it solves the issue.

Taking out of draft.

unit_tests/src/test_resource_administration.cpp
```diff
@@ -41,6 +42,15 @@ TEST_CASE("resource administration")
...
     SECTION("add / remove child resource")
     {
+        // TODO Update test to check the mtime following each operation.
+#if 0
+        auto repl_resc_info = adm::client::resource_info(conn, repl_info.resource_name);
+        REQUIRE(repl_resc_info);
+
+        auto ufs_resc_info = adm::client::resource_info(conn, ufs_info.resource_name);
+        REQUIRE(ufs_resc_info);
+#endif
```
korydraughn, line 52: Still have to do this.

korydraughn: Done.

lib/core/include/irods/connection_pool.hpp
```diff
@@ -232,7 +232,8 @@ namespace irods
             connection_pointer conn{nullptr, rcDisconnect};
             rErrMsg_t error{};
             std::chrono::steady_clock::time_point creation_time;
-            std::uint64_t latest_resc_mtime{};
+            std::string latest_resc_mtime;
+            std::int32_t resc_count{};
```
korydraughn, line 236: std::int32_t -> std::int16_t

scripts/irods/database_upgrade.py
```diff
@@ -152,6 +152,13 @@ def run_update(irods_config, cursor):
                         ['mysql', '='.join(['--defaults-file', f.name]), irods_config.database_config['db_name']],
                         stdin=sql_file)
 
+        # Add a new column that will be responsible for holding the fractional part of the modify_ts
+        # for a resource. The value stored will be in milliseconds.
+        if irods_config.catalog_database_type == 'oracle':
+            database_connect.execute_sql_statement(cursor, "alter table R_RESC_MAIN add (modify_ts_millis varchar(3) default '000');")
```
trel, line 158: we confirm this parentheses notation is correct?   looks a little different than above oracle examples.

unit_tests/src/test_resource_administration.cpp
```diff
@@ -42,6 +48,8 @@ TEST_CASE("resource administration")
     SECTION("add / remove child resource")
     {
         {
+            auto deferred_mtime_changed_assertion = make_deferred_mtime_update_assertion(conn, ufs_info.resource_name);
```
korydraughn, line 51: `_update_` -> `_changed_`

scripts/irods/database_upgrade.py
korydraughn: That is based on their documentation, websites I found, and chatgpt.

But, I can't claim it works since I don't have a way to test it.

trel: fair enough.   we'll fix it if we hear from anyone.

lib/core/include/irods/connection_pool.hpp
korydraughn: Leaving as is since `irods::query::total_row_count()` returns an `int`.

unit_tests/src/test_resource_administration.cpp
korydraughn: Done.

korydraughn:
All unit tests pass.

Running core tests.

lib/core/src/connection_pool.cpp
```diff
@@ -301,27 +308,52 @@ namespace irods
...
 
-            // Check if the connection is still valid.
-            // This query will always succeed unless there's an issue in which case an exception will
-            // be thrown.
-            query{ctx.conn.get(), "select ZONE_NAME where ZONE_TYPE = 'local'"}; // NOLINT(bugprone-unused-raii)
+                // Only return false if it was determined that the connection needed to be synchronized
+                // with the catalog.
+                if (!in_sync) {
+                    return in_sync;
+                }
```
alanking, line 349: This function returns `true` in the implicit else case here (see L362), so we could just return `in_sync` here in any case.

unit_tests/src/test_connection_pool.cpp
```diff
@@ -379,6 +379,51 @@ TEST_CASE("connection pool")
             REQUIRE_NOTHROW(adm::client::modify_resource(conn, "demoResc", adm::resource_comments_property{""}));
         }
 
+        SECTION("refresh connection after adding/removing resource")
```
alanking, line 382: Would it be worth making a test in which a resource is added and a different, older resource is removed? This would make the resource count appear to be the same on the connection fetch but the latest modify_time would be different due to the resource being added. Just a suggestion

plugins/database/src/db_plugin.cpp
```diff
@@ -618,27 +637,28 @@ irods::error _childIsValid(
...
-    const std::string& _child_resc_id,
-    const std::string& _parent_resc_id,
-    const std::string& _parent_child_context ) {
+irods::error _updateChildParent(const std::string& _child_resc_id,
+                                const std::string& _parent_resc_id,
+                                const std::string& _parent_child_context)
+{
     irods::sql_logger logger( "_updateChildParent", logSQL );
 
+    const auto [secs, millis] = get_current_time();
```
alanking, line 646: Consider different names for these...

`secs` -> `current_time_sec` (or `*_s` or `*_seconds` or...)
`millis` -> `current_time_msec` (or `*_ms` or `*_millis` or `*_milliseconds` or...)

alanking (commented):
Seems good. Had a few suggestions

unit_tests/src/test_connection_pool.cpp
alanking: Sorry, to clarify, the timeline would look like this, as a minimal example:
1. Create resource A
2. Instantiate connection pool
3. Create resource B, remove resource A
4. Fetch a connection from the pool and assert that it has refreshed

plugins/database/src/db_plugin.cpp
```diff
@@ -8343,27 +8367,39 @@ irods::error db_mod_resc_data_paths_op(
...
+        cllBindVars[cllBindVarCount++] = millis.c_str();
         cllBindVars[cllBindVarCount++] = rescId;
         cllBindVars[cllBindVarCount++] = oldPath2;
-        status =  cmlExecuteNoAnswerSql(
-                      "update R_DATA_MAIN set data_path = replace (R_DATA_MAIN.data_path, ?, ?) where resc_id=? and data_path like ?",
-                      &icss );
+        status = cmlExecuteNoAnswerSql(
+            "update R_DATA_MAIN set data_path = replace (R_DATA_MAIN.data_path, ?, ?), modify_ts=?, modify_ts_millis=? "
+            "where resc_id=? and data_path like ?",
+            &icss);
```
korydraughn, line 8402: Revert these changes.

This code has nothing to do with modifying resource info.

lib/core/src/connection_pool.cpp
korydraughn: Thinking about this a little more, if we change this to always return here, that will increase the chance of introducing a bug if a developer wants to add new refresh options after the resource change detection logic.

Conditionally returning here allows new refresh options to be added without risk of breaking the implementation.

Thoughts?

trel: either seems fine - but leave comments in the code to tell the next developer the intent of the design.

unit_tests/src/test_connection_pool.cpp
trel: this seems a good thing to do.   and i also think the current code/plan will handle it just fine.  (also good).

korydraughn: I'm confused lol. What are we testing?

My guess is you're talking about proving the connections are refreshed even when the latest mtime hasn't changed, but the resource count has due to the removal of an old resource. Is that correct?

trel: no... the count will stay the same, but the latest time is updated/newer.

this is just a complicated version of just updating a single resource.

korydraughn: I see. We're talking about another client modifying the resources on the side.

Yes, definitely need tests for those cases.

alanking: Yeah, like @trel said, it's not different from the modify case in essence but could catch some future issue should the logic change.

plugins/database/src/db_plugin.cpp
korydraughn: Done.

plugins/database/src/db_plugin.cpp
korydraughn: Done.

unit_tests/src/test_connection_pool.cpp
korydraughn: I've added a new test to capture the behavior you mentioned.

Please review.

lib/core/src/connection_pool.cpp
korydraughn: I added a comment explaining the reason for leaving this in.

korydraughn:
All unit tests pass.

Running core tests.

korydraughn:
Other than clang-tidy, are there any other thoughts concerning the PR?

korydraughn:
All core tests pass.

korydraughn:
Squashed and ready for pounds.

I've addressed the one or two real clang-tidy issues. I've ignored all magic-number issues reported by clang-tidy.

unit_tests/src/test_connection_pool.cpp
```diff
@@ -379,6 +379,108 @@ TEST_CASE("connection pool")
...
+                REQUIRE_NOTHROW(adm::client::remove_resource(conn, resc_name));
+            }
+
+            // Show the connection in the pool has been updated again.
+            auto conn = conn_pool.get_connection();
+            REQUIRE(conn);
+            REQUIRE(static_cast<RcComm*>(conn) != conn_ptr);
+        }
+
+        SECTION("refresh connection after resource is modified by external source")
```
alanking, line 427: I guess it's a form of modification, but should this read like this?

```suggestion
        SECTION("refresh connection after resource is removed by external source")
```

alanking (approved):
Just one minor thing. Feel free to ignore or address. In either case, seems good to me

unit_tests/src/test_connection_pool.cpp
korydraughn: Will tweak to be more specific about what is happening.

alanking (approved):
Add #

korydraughn:
#'d

PR #7307, opened 2023-09-07
[7023] fix for catalog_service_roles, no longer a set (4-3-stable)

trel:
oops, missed this in my local diff.

added some comments.

PR #7308, opened 2023-09-07
[7023] fix for catalog_service_roles, no longer a set (main)

trel:
oops, missed this in my local diff.

added some comments.

same as https://github.com/irods/irods/pull/7307

### DIFF
--- a/scripts/setup_irods.py
+++ b/scripts/setup_irods.py
@@ -194,13 +194,20 @@ def check_hostname():
         raise IrodsError('The hostname (%s) must resolve to the local machine.' % (hostname))
 
 def determine_server_role(irods_config):
-    catalog_service_roles = set(['provider', 'consumer'])
+    # The list of catalog service roles supported by the server.
+    catalog_service_roles = ['provider', 'consumer']
+
+    # Get the service role previously configured if available. Otherwise, set to "provider".
     default_catalog_service_role = irods_config.server_config.get('catalog_service_role', 'provider')
+
+    # Get the new value from user via default_prompt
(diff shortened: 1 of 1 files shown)


scripts/setup_irods.py
```diff
@@ -194,13 +194,20 @@ def check_hostname():
         raise IrodsError('The hostname (%s) must resolve to the local machine.' % (hostname))
 
 def determine_server_role(irods_config):
-    catalog_service_roles = set(['provider', 'consumer'])
+    # enum
```
korydraughn, line 197: I think this needs more words. How about ...
```py
# The list of catalog service roles supported by the server.
```

scripts/setup_irods.py
```diff
@@ -194,13 +194,20 @@ def check_hostname():
...
+    catalog_service_roles = ['provider', 'consumer']
+
+    # get earlier setting, set to 'provider' if not set
```
korydraughn, line 200: How about ...
```py
# Get the service role previously configured if available. Otherwise, set to "provider".
```

korydraughn (commented):
Just a few comments :-)

Overall, the code is fine. Arrays work!

scripts/setup_irods.py
trel: certainly

scripts/setup_irods.py
trel: will do

trel:
updated

trel:
squashed and #'d

alanking (approved):

PR #7317, opened 2023-09-11
Updated API plugin configurations (main)

korydraughn:


alanking (commented):
Do these APIs work when using federation? Does `REMOTE_USER_AUTH` mean that a remote user can authenticate directly with a server in a remote zone and invoke these APIs?

korydraughn:
I haven't tested this yet, but each API plugin in this PR uses the same values that `rxDataObjPut` uses, so there shouldn't be any issues.

We'll know more once I run the tests.

korydraughn:
All unit tests pass.

Running federation tests.

korydraughn:
All federation tests passed.

Running core tests.

trel:
hooray!

korydraughn:
Yes. Will do that soon.

Do we agree with the changes? I'm going to check a few other APIs before we merge.

alanking:
I think I would like it more if we understood the change better. It works with or without the change, so, I would like to know what's changing.

korydraughn:
Let's discuss over zoom today.

korydraughn:
After some discussion, we've decided that these changes need unit tests to assert expected behavior.

unit_tests/src/test_rc_data_obj.cpp
```diff
@@ -1528,3 +1530,23 @@ TEST_CASE("#7250: updates to ticket stats do not hang")
...
+    CHECK(rc_register_physical_path(static_cast<RcComm*>(conn), &input, &json_error_string) == SYS_NO_API_PRIV);
+    CHECK(json_error_string == nullptr);
+}
+
+TEST_CASE("api plugin")
+{
+    load_client_api_plugins();
+    irods::experimental::client_connection conn{irods::experimental::defer_authentication};
+    CHECK(rc_touch(static_cast<RcComm*>(conn), "") == SYS_NO_API_PRIV);
+}
```
korydraughn, line 1552: How do we feel about these test cases living here?

NGL, I really didn't want to create two additional files to test these cases. If others feel that is incorrect, then I'll create the files.

Thoughts?

alanking: This file is becoming a bit of a dumping ground anyway, so it seems fine here to me. I would request that the test case names be a little more descriptive and/or a comment indicating the purpose, please :)

korydraughn: Agreed. Those names are placeholders until I figure out how to describe what they're about.

korydraughn: I still have to write an issue for this PR too.

korydraughn:
All unit tests pass.

Running core tests.

korydraughn:
All core tests passed.

Running federation tests.

korydraughn:
All federation tests passed.

alanking (approved):
Seems good to me. # when ready

korydraughn:
#'d

Issue #7283, opened 2023-08-25
Expose `set_grid_configuration` (and get) API through `iadmin` subcommand


alanking:
- [x] main
- [x] 4-3-stable
---
## Feature

General work underpinning the effort for https://github.com/irods/irods/issues/7274

`set_grid_configuration` should require a namespace, option name, and option value.

`get_grid_configuration` should require a namespace and option name and should only return the value if found. We will not support generically listing all configurations for a given namespace at this time.

Part of this effort should be to add a list of "protected" grid configurations. That is, grid configurations which should not be modifiable through this API.

The database schema_version should be protected because modifying this should only happen on server upgrade.

The entire delay_server namespace should be protected because the `set_delay_server` API/`iadmin` subcommand should be used for setting those configurations.

Issue #7274, opened 2023-08-22
Move pam_password configuration to `R_GRID_CONFIGURATION`


alanking:
- [x] main
- [x] 4-3-stable
- [x] irods_docs
---
## Feature

The following `pam_password` configurations are supported in server_config.json:
 - no_extend
 - password_max_time
 - password_min_time

These configurations should be moved out of server_config to new rows in `R_GRID_CONFIGURATION`. This allows for greater support of multiple providers (or HA setups) and will allow authenticating from other servers in a zone in the future.

Additionally, no_extend should be renamed to extend_password_lifetime to remove the negative language and prevent confusion. The setting should only affect the lifetime of the randomly generated password for PAM authentication. It's not meant to extend the lifetime by "less", as it does now:
https://github.com/irods/irods/blob/34cd4a21b88477772a309054bc9f710353d4260b/plugins/database/src/db_plugin.cpp#L2020-L2024

The configuration values should be pulled from server_config.json if available. If not, default values should be used as follows:
 - extend_password_lifetime: 1 - This will be interpreted as "true", which is logically equivalent to no_extend defaulting to "false".
 - password_max_time: 1209600 - Current default. Measured in seconds (14 days).
 - password_min_time: 121 - Current default. Measured in seconds (2 minutes and 1 second)

This issue is a follow-up to #7098.

alanking:
Discovered a major issue with the current approach... native and pam_password authentication cannot be distinguished in the auth check database operation because pam_password literally performs native authentication with the randomly generated password.

We had some discussion and agreed that the best/least destructive path forward would be to keep the options combined as they have been historically, but make the auth configuration more generic. Instead of namespacing `authentication::pam_password` and `authentication::native` configurations in `R_GRID_CONFIGURATION`, we will introduce a single namespace `authentication` which will configure for both plugins (and any other authentication plugins which so choose to use the setting).

In the future we may introduce plugin-specific configurations which could be used to override relevant settings in the generic namespace.

Issue #7266, opened 2023-08-16
Debian 12 support


SwooshyCueb:
- [x] irods_development_environment
- [x] irods_testing_environment
- [x] GitHub Actions (main)
- [x] GitHub Actions (4-3-stable)
---
We are planning to support Debian 12 for 4.3.1. I didn't see an issue for this already, so I'm opening one.

PR #7331, opened 2023-09-18
Move auth configs to `R_GRID_CONFIGURATION`, plus fixes (main)

alanking:
In service of #6597
Addresses #7098 
In service of #7266 
In service of #7283 
Addresses #7274 

Companion PR: https://github.com/irods/irods_client_icommands/pull/407

### DIFF
--- a/lib/core/include/irods/irods_random.hpp
+++ b/lib/core/include/irods/irods_random.hpp
@@ -1,8 +1,19 @@
 #ifndef IRODS_RANDOM_HPP
 #define IRODS_RANDOM_HPP
 
+#include <string>
+
 namespace irods {
     void getRandomBytes( void * buf, int bytes );
+
+    /// \brief Generate a string of size \p _length consisting of randomly generated alphanumeric characters.
+    ///
+    /// \param[in] _length Desired length of the generated string.
+    ///
+    /// \throws irods::exception If \p _length is negative.
+    ///
+    /// \since 4.3.1
+    auto generate_random_alphanumeric_string(std::int16_t _length) -> std::string;
 
     template <typename T>
     T
--- a/lib/core/src/irods_random.cpp
+++ b/lib/core/src/irods_random.cpp
@@ -1,7 +1,35 @@
-#include <boost/random.hpp>
-#include <openssl/rand.h>
 #include "irods/irods_random.hpp"
+
+#include "irods/irods_exception.hpp"
+#include "irods/rodsErrorTable.h"
+
+#include <boost/random.hpp>
+#include <fmt/format.h>
+#include <openssl/rand.h>
+
+#include <cctype>
 #include <ctime>
+#include <random>
+#include <string>
+
+namespace
+{
+    auto generate_random_alphanumeric_character() -> char
+    {
+        constexpr char min_char = '0';
+        constexpr char max_char = 'z';
+
+        static std::random_device rd;
+        static std::default_random_engine e{rd()};
+        static std::uniform_int_distribution<> d{min_char, max_char};
+
+        char c{};
+        while (!std::isalnum(static_cast<unsigned char>(c))) {
+            c = static_cast<char>(d(e));
+        }
+        return c;
+    } // generate_random_alphanumeric_character
+} // anonymous namespace
 
 void irods::getRandomBytes( void * buf, int bytes ) {
     if ( RAND_bytes( ( unsigned char * )buf, bytes ) != 1 ) {
@@ -13,3 +41,30 @@ void irods::getRandomBytes( void * buf, int bytes ) {
         }
     }
 }
+
+auto irods::generate_random_alphanumeric_string(std::int16_t _length) -> std::string
+{
+    if (_length < 0) {
+        THROW(SYS_INVALID_INPUT_PARAM, fmt::format("{}: _length must be non-negative.", __func__));
+    }
+
+    try {
+        std::string s;
+        s.reserve(_length);
+
+        for (std::int16_t i = 0; i < _length; ++i) {
+            s += generate_random_alphanumeric_character();
+        }
+
+        return s;
+    }
+    catch (const std::length_error& e) {
+        THROW(SYS_INVALID_INPUT_PARAM, fmt::format("{}: Invalid length. [{}]", __func__, e.what()));
+    }
+    catch (const std::exception& e) {
+        THROW(SYS_LIBRARY_ERROR, fmt::format("{}: Failed to generate string. [{}]", __func__, e.what()));
+    }
+    catch (...) {
+        THROW(SYS_UNKNOWN_ERROR, "Failed to generate string: [Unknown error]");
+    }
+} // generate_random_alphanumeric_string
--- a/plugins/api/src/set_grid_configuration_value.cpp
+++ b/plugins/api/src/set_grid_configuration_value.cpp
@@ -15,6 +15,8 @@
 // Server-side Implementation
 //
 
+// clang-format off
+
 #include "irods/set_grid_configuration_value.h"
 #include "irods/irods_server_api_call.hpp"
 #include "irods/icatHighLevelRoutines.hpp"
@@ -22,6 +24,10 @@
 #include "irods/irods_logger.hpp"
 
 #include <string>
+#include <string_view>
+#include <utility> // for std::pair
+
+// clang-format on
 
 namespace ic = irods::experimental::catalog;
 
@@ -37,9 +43,38 @@ namespace
 
     auto rs_set_grid_configuration_value(RsComm*, const GridConfigurationInput*) -> int;
 
+    auto is_grid_configuration_allowed_to_be_modified(const char* _namespace, const char* _option_name) -> bool;
+
     //
     // Function Implementations
     //
+
+    auto is_grid_configuration_allowed_to_be_modified(const char* _namespace, const char* _option_name) -> bool
+    {
+        constexpr auto protected_config_list = std::to_array<std::pair<std::string_view, std::string_view>>(
+            {{"database", "schema_version"}, {"delay_server", ""}});
+
+        // If none of the namespace/option name combinations in the protected list match the passed in namespace and
+        // option name, it is allowed to be modified and so the function should return true. If any instance returns
+        // true, that means a match was found in the list of protected configurations, so we should return false in this
+        // function because that means the grid configuration is not allowed to be modified.
+        return std::none_of(std::begin(protected_config_list),
+                            std::end(protected_config_list),
+                            [&_namespace, &_option_name](const auto& _pair) {
+                                const auto& [protected_namespace, protected_option_name] = _pair;
+
+                                // If this evaluates to true, this is not a protected option and is allowed to be
+                                // modified. Return false because this is not a match.
+                                if (protected_namespace != _namespace) {
+                                    return false;
+                                }
+
+                                // Empty string indicates that modifications for all options in the namespace are
+                                // forbidden. If this expression evaluates to true, this a protected option and is not
+                                // allowed to be modified.
+                                return protected_option_name.empty() || protected_option_name == _option_name;
+                            });
+    } // is_grid_configuration_allowed_to_be_modified
 
     auto call_set_grid_configuration_value(irods::api_entry* _api,
                                            RsComm* _comm,
@@ -63,6 +98,12 @@ namespace
         catch (const irods::exception& e) {
             log_api::error(e.what());
             return e.code();
+        }
+
+        if (!is_grid_configuration_allowed_to_be_modified(_input->name_space, _input->option_name)) {
+            constexpr auto ec = SYS_NOT_ALLOWED;
+            addRErrorMsg(&_comm->rError, ec, "Specified grid configuration is not allowed to be modified.");
+            return ec;
         }
 
         return chlSetGridConfigurationValue(_comm,
--- a/plugins/database/CMakeLists.txt
+++ b/plugins/database/CMakeLists.txt
@@ -77,7 +77,6 @@ foreach(plugin IN LISTS IRODS_DATABASE_PLUGINS)
     ${plugin_target}
     PRIVATE
     "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
-    "$<BUILD_INTERFACE:${CMAKE_IRODS_SOURCE_DIR}/plugins/auth/include>"
     "${IRODS_EXTERNALS_FULLPATH_BOOST}/include"
     "${IRODS_EXTERNALS_FULLPATH_FMT}/include"
     "${IRODS_EXTERNALS_FULLPATH_NANODBC}/include"
--- a/plugins/database/src/db_plugin.cpp
+++ b/plugins/database/src/db_plugin.cpp
@@ -101,13 +101,6 @@ static rodsLong_t MAX_PASSWORDS = 40;
 
 #define MAX_HOST_STR             2700
 
-// =-=-=-=-=-=-=-
-// local variables externed for config file setting in
-bool irods_pam_auth_no_extend = false;
-char irods_pam_password_min_time[ NAME_LEN ]     = { "121" };
-char irods_pam_password_max_time[ NAME_LEN ]     = { "1209600" };
-char irods_pam_password_default_time[ NAME_LEN ] = { "1209600" };
-
 size_t log_sql_flg = 0;
 icatSessionStruct icss; // JMC :: only for testing!!!
 extern int logSQL;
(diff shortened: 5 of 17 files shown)


plugins/database/src/db_plugin.cpp
```diff
@@ -123,6 +116,98 @@ const std::string ZONE_PROP( "irods_zone_property" );
...
+    constexpr std::size_t grid_configuration_size = 2700;
+    struct auth_config
+    {
+        bool initialized = false;
+
+        bool password_extend_lifetime = true;
+
+        static constexpr rodsLong_t default_password_max_time = 1209600;
+        rodsLong_t password_max_time = default_password_max_time;
+        std::array<char, grid_configuration_size + 1> password_max_time_str{};
```
alanking, line 131: This either needs to default to `default_password_max_time` as a string, or we don't want to keep this in the struct.

alanking:
A couple other things I wanted to mention...

All unit tests and core tests pass

I think there's some negligible performance left on the table with the function checking for whether a grid configuration is in the protected list. It could be a map rather than a flat array to potentially reduce the number of loop iterations (and may be more readable in the code). I don't think this command will be run frequently enough for it to matter.

lib/core/include/irods/irods_random.hpp
```diff
@@ -1,9 +1,20 @@
...
+#include <string>
+
 namespace irods {
     void getRandomBytes( void * buf, int bytes );
 
+    /// \brief Generate a string of size \p _length consisting of randomly generated alphanumeric characters.
+    ///
+    /// \param[in] _length Desired length of the generated string.
+    ///
+    /// \throws irods::exception If \p _length exceeds max_size or on allocation failure.
```
korydraughn, line 13: What is "max_size"?
Do we have a variable that defines it?

lib/core/include/irods/irods_random.hpp
```diff
@@ -1,9 +1,20 @@
...
+    ///
+    /// \since 4.3.1
+    auto generate_random_alphanumeric_string(std::size_t _length) -> std::string;
```
korydraughn, line 16: Should we constrain the length to a smaller data type?

size_t seems way too large IMO.

lib/core/src/irods_random.cpp
```diff
@@ -1,7 +1,34 @@
...
+    {
+        constexpr char min_char = '0';
+        constexpr char max_char = 'z';
+
+        static std::random_device rd;
+        static std::default_random_engine e{rd()};
+        static std::uniform_int_distribution<> d{min_char, max_char};
+
+        char c{};
+        while (!std::isalnum(c)) {
```
korydraughn, line 26: Missing cast to `unsigned char`.

See https://en.cppreference.com/w/cpp/string/byte/isalnum.

korydraughn, line 26: Needs more `<cctype>` header for `isalnum`.

lib/core/src/irods_random.cpp
```diff
@@ -1,7 +1,34 @@
...
+namespace
+{
+    auto generate_random_alphanumeric_character() -> char
+    {
+        constexpr char min_char = '0';
+        constexpr char max_char = 'z';
+
+        static std::random_device rd;
+        static std::default_random_engine e{rd()};
+        static std::uniform_int_distribution<> d{min_char, max_char};
```
korydraughn, line 23: I wonder if we should pass this into the function rather than make it static?

Passing it in gives us the ability to change the RNG, but I'm not sure if it's worth the effort.

plugins/api/src/set_grid_configuration_value.cpp
```diff
@@ -37,10 +43,40 @@ namespace
...
+
+        // If none of the namespace/option name combinations in the protected list match the passed in namespace and
+        // option name, it is allowed to be modified and so the function should return true. If any instance returns
+        // true, that means a match was found in the list of protected configurations, so we should return false in this
+        // function because that means the grid configuration is not allowed to be modified.
+        return std::none_of(std::begin(protected_config_list),
+                            std::end(protected_config_list),
+                            [&_namespace, &_option_name](const auto& _pair) {
+                                const auto& protected_namespace = _pair.first;
+                                const auto& protected_option_name = _pair.second;
```
korydraughn, line 65: Structured bindings?

Not required to change anything. Just highlighting it.

I always wonder if it's better to copy the string views rather than take a reference. We should measure the difference one day.

scripts/irods/database_upgrade.py
```diff
@@ -159,6 +159,30 @@ def run_update(irods_config, cursor):
...
+
+        password_config_dict = {
+            # Removing negative language from the option name results in the meaning being reversed.
+            'password_extend_lifetime': '0' if True == pam_password_config.get('no_extend', False) else '1',
+            'password_max_time': str(pam_password_config.get('password_min_time', 121)),
+            'password_min_time': str(pam_password_config.get('password_max_time', 1209600))
+        }
+
+        scheme_namespaces = ['authentication::pam_password', 'authentication::native']
+        statement_str = "insert into R_GRID_CONFIGURATION values ('{}','{}','{}');"
```
korydraughn, line 179: Consider defining the names of the columns that are being updated.
```sql
insert into R_GRID_CONFIGURATION (column_1, column_2, column_3) values ('{}', '{}', '{}');
```

unit_tests/src/test_generate_random_alphanumeric_string.cpp
```diff
@@ -0,0 +1,31 @@
...
+    CHECK(irods::generate_random_alphanumeric_string(0).empty());
+}
+
+TEST_CASE("length of 1 effectively calls generate_random_alphanumeric_character")
+{
+    constexpr std::size_t character_count = 1;
+    const auto s = irods::generate_random_alphanumeric_string(character_count);
+    REQUIRE(!s.empty());
+    CHECK(s.size() == character_count);
+    CHECK(std::all_of(std::begin(s), std::end(s), [](const auto& c) { return std::isalnum(c); }));
```
korydraughn, line 21: `c` needs to be cast before passing to `isalnum` otherwise, it's UB.

unit_tests/src/test_generate_random_alphanumeric_string.cpp
```diff
@@ -0,0 +1,31 @@
...
+}
+
+TEST_CASE("only alphanumeric characters appear in string of statistically significant length")
+{
+    constexpr std::size_t character_count = 1'000'000;
+    const auto s = irods::generate_random_alphanumeric_string(character_count);
+    REQUIRE(!s.empty());
+    CHECK(s.size() == character_count);
+    CHECK(std::all_of(std::begin(s), std::end(s), [](const auto& c) { return std::isalnum(c); }));
```
korydraughn, line 30: Same here.

plugins/database/CMakeLists.txt
```diff
@@ -64,8 +64,8 @@ foreach(plugin IN LISTS IRODS_DATABASE_PLUGINS)
     ${plugin_target}
     PRIVATE
     irods_common
-    irods_server
     irods_plugin_dependencies
+    irods_server
```
korydraughn, line 68: What motivated this change?

plugins/database/CMakeLists.txt
```diff
@@ -77,7 +77,6 @@ foreach(plugin IN LISTS IRODS_DATABASE_PLUGINS)
     ${plugin_target}
     PRIVATE
     "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
-    "$<BUILD_INTERFACE:${CMAKE_IRODS_SOURCE_DIR}/plugins/auth/include>"
```
korydraughn, line 80: What motivated this change?

korydraughn (commented):
Looking good so far.

I'll review the larger files soon.

lib/core/include/irods/irods_random.hpp
alanking: Meant to update this for context: https://en.cppreference.com/w/cpp/string/basic_string/max_size

lib/core/include/irods/irods_random.hpp
alanking: Open to suggestions on a type

lib/core/src/irods_random.cpp
alanking: Oh, good to know. Will do

alanking: Could have sworn I did that... Will do

lib/core/src/irods_random.cpp
alanking: I don't think it is worth the effort for 4.3.1. Willing to do it in the future, though. No interfaces would really be changing here since this is in the anonymous namespace. Thoughts?

plugins/database/CMakeLists.txt
alanking: Removed `#include "irods/plugins/auth/pam_password.hpp"` from `db_plugin.cpp` so we don't need this anymore (and it was a mistake to create/include that file to begin with).

plugins/database/CMakeLists.txt
alanking: I had made some changes in this stanza before but backed them out and didn't back this out. Eh, alphabetized. Can revert

plugins/api/src/set_grid_configuration_value.cpp
alanking: I feel like... I meant to use structured bindings, but never revisited this.

scripts/irods/database_upgrade.py
alanking: Sure

lib/core/include/irods/irods_random.hpp
alanking: Going with `std::int32_t` for now. Allowing for a signed type lets us play defense against negative values a little better and it is still sufficient to hold a decently large string.

lib/core/include/irods/irods_random.hpp
korydraughn: Ah I see. Seems tricky to word that without mentioning `std::string` as an implementation detail.

lib/core/include/irods/irods_random.hpp
korydraughn: Sorry on the late response. Maybe even `std::int16_t`. That gives people quite a bit of room for generating random strings.

Alternatively, you could make it a template and let the caller decide.

I'll leave the final decision up to you.

lib/core/src/irods_random.cpp
korydraughn: That sounds fine. Certainly not required for 4.3.1.

plugins/database/CMakeLists.txt
korydraughn: It's likely fine. I know there are situations where link order matters. I'm not so sure it matters here though.

I'd recommend to leave it as it was originally to avoid potential issues in case this does depend on link order.

Perhaps CMake frees us from that situation though.

lib/core/include/irods/irods_random.hpp
alanking: I think `std::int16_t` is fine. Like the RNG backend, this can be turned into an option later if we even continue to use this.

alanking: Or, this thing that I just discovered...
```c++
auto generate_random_alphanumeric_string(std::signed_integral auto _length) -> std::string;
```
Thoughts?

plugins/database/src/db_plugin.cpp
```diff
@@ -6520,12 +6572,14 @@ irods::error db_check_auth_op(
     getNowStr( myTime );
 
     /* Check for PAM_AUTH type passwords */
-    pamMaxTime = atoll( irods_pam_password_max_time );
-    pamMinTime = atoll( irods_pam_password_min_time );
 
-    if ( ( strncmp( goodPwExpiry, "9999", 4 ) != 0 ) &&
-            expireTime >=  pamMinTime &&
-            expireTime <= pamMaxTime ) {
+    if (const auto err = get_auth_config("authentication::pam_password", ac); !err.ok()) {
```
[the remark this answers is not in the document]
alanking: It will not. I think we need to decide on a direction here, because it feels like we might start moving in a different direction: https://github.com/irods/irods/issues/7318 I realize that issue has to do with config *files* but it is a configuration nonetheless.

lib/core/include/irods/irods_random.hpp
korydraughn: I can't see a reason to allow very large string sizes. As long as it can't be misused and exhaust memory, I'm fine with whatever choice you make.

Using concepts like you have in the sample is always cool to see. I don't how helpful it is here though. Perhaps the template idea is pointless since it would allow virtually any range of signed integers (i.e. someone can pass int64_t).

alanking: Sure, let's just stick with `std:int16_t` then

lib/core/include/irods/irods_random.hpp
alanking: I just mentioned the negative length. Given the smaller type (`std::int16_t`) it is not possible to exceed `max_size` for all the implementations I'm aware of, and allocation failures are not worth mentioning because I imagine other issues will appear first. Please start a new conversation for wording concerns

lib/core/include/irods/irods_random.hpp
alanking: Done

plugins/database/src/db_plugin.cpp
```diff
@@ -123,6 +117,98 @@ const std::string ZONE_PROP( "irods_zone_property" );
 
 static const auto intermediate_replica_status_str = std::to_string(INTERMEDIATE_REPLICA);
 
+namespace
+{
+    constexpr std::size_t grid_configuration_size = 2700;
+    struct auth_config
```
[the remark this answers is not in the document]
alanking: Done... Please open new conversation if these seem insufficient.

plugins/database/src/db_plugin.cpp
```diff
@@ -123,6 +117,98 @@ const std::string ZONE_PROP( "irods_zone_property" );
...
+
+            if (option_name == irods::KW_CFG_PAM_PASSWORD_MIN_TIME) {
+                std::strncpy(_out.password_min_time_str.data(),
+                             result.get<std::string>(1).c_str(),
+                             _out.password_min_time_str.size());
+                try {
+                    _out.password_min_time = std::stoll(_out.password_min_time_str.data());
+                }
+                catch (...) {
+                    log_db::error("Grid configuration value [{}] in namespace [{}] is invalid. Using default.",
```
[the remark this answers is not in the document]
alanking: Done

plugins/database/src/db_plugin.cpp
```diff
@@ -123,6 +117,98 @@ const std::string ZONE_PROP( "irods_zone_property" );
...
+
+            if (option_name == irods::KW_CFG_PAM_PASSWORD_MAX_TIME) {
+                std::strncpy(_out.password_max_time_str.data(),
+                             result.get<std::string>(1).c_str(),
+                             _out.password_max_time_str.size());
+                try {
+                    _out.password_max_time = std::stoll(_out.password_max_time_str.data());
+                }
+                catch (...) {
+                    log_db::error("Grid configuration value [{}] in namespace [{}] is invalid. Using default.",
```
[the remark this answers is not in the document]
alanking: Done

plugins/database/src/db_plugin.cpp
```diff
@@ -123,6 +117,98 @@ const std::string ZONE_PROP( "irods_zone_property" );
...
+                const auto option_value = result.get<std::string>(1);
+                if (option_value == "1") {
+                    _out.password_extend_lifetime = true;
+                }
+                else if (option_value == "0") {
+                    _out.password_extend_lifetime = false;
+                }
+                else {
+                    // Use default if it's neither, and print an annoying error message.
+                    log_db::error("Grid configuration value [{}] in namespace [{}] is invalid. Using default.",
```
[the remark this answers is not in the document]
alanking: Done

alanking:
GitHub appears to be having some problems updating hte PR, but you can see the changes in the branch, promise :)

Unit tests pass. Running core tests now.

plugins/api/src/set_grid_configuration_value.cpp
korydraughn: Not really. It's not urgent. Of course, you're free to look into it if you want.

I've just wondered which is better in general. The design of string_view implies copying is best, but there may be cases where passing by reference may be better.

Hmm, saying that out loud makes me think copying is always better for string_view since passing it by reference introduces one more level of indirection.

alanking: Well, would you look at that...

https://quick-bench.com/q/LhApqVtkVv5BthpOiF19FeMyXEI

trel: If i'm reading that link correctly... copy is slightly faster.... but then changing to `-O3` makes them equivalent.

korydraughn: Looks like they are nearly identical in performance.

plugins/database/src/db_plugin.cpp
```diff
@@ -123,6 +117,98 @@ const std::string ZONE_PROP( "irods_zone_property" );
...
+        std::array<char, grid_configuration_size + 1> password_min_time_str{};
+    };
+
+    auto get_auth_config(const char* _namespace, auth_config& _out) -> irods::error
+    {
+        // Initializing the auth configs incurs a query. We only need to do this once per agent lifetime, so a member
+        // variable has been provided to prevent extraneous initializations.
+        if (_out.initialized) {
+            return SUCCESS();
+        }
```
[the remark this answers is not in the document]
alanking: Decided that the auth configs will be updated every time. Resolving.

plugins/api/src/set_grid_configuration_value.cpp
alanking: I think this is okay as-is given the bench test results. Resolving.

plugins/database/src/db_plugin.cpp
```diff
@@ -7520,8 +7528,8 @@ irods::error db_mod_user_op(
     if ( strncmp( _option, "rmPamPw", 9 ) == 0 ) {
         auth_config ac{};
         if (const auto err = get_auth_config("authentication::pam_password", ac); !err.ok()) {
-            irods::log(err);
-            log_db::warn("Failed to get password configuration - using defaults.");
+            log_db::warn("Failed to get auth configuration. Using default values. [{}]", err.result());
+            ac = auth_config{};
```
alanking, line 7532: I'm not so sure about this now... Might be better to have `get_auth_config` handle this on a case-by-case basis and not return an error object.

korydraughn: I don't understand. Please elaborate.

Aside from that, should we be falling back to the default config or returning an error? The default config could be very different from what the admin wants.

alanking: Looking at the previous implementation, it seems that an error was indeed returned in the start operation if any of the configurations fails to be extracted for whatever reason. Will adjust this.

alanking:
Okay, gotta run tests again, but... seems this is in a good spot now, pending any further review comments.

lib/core/src/irods_random.cpp
```diff
@@ -1,7 +1,35 @@
...
 #include <ctime>
+#include <random>
+#include <string>
+
+namespace
+{
+    auto generate_random_alphanumeric_character() -> char
+    {
+        constexpr char min_char = '0';
+        constexpr char max_char = 'z';
```
trel, line 20: is this ascii?   if so... there are other things between 0 and z...

alanking: Yes, see `std::isalnum` check below.

korydraughn:
It just occurred to me, should we rename the subcommands to something like:
- `get_grid_option` or `get_grid_configuration_option`
- `set_grid_option` or `set_grid_configuration_option`

As I was looking through the tests, `get_grid_configuration` kept standing out to me. Seems that if we ever wanted to show the entire configuration, that name may be what we want.

I'm not trying to create unnecessary work, but I do want to raise the question to see what others thought.

lib/core/src/irods_random.cpp
trel: while loop.   sorry.   as you were.

trel:
generally in favor of more letters to reduce confusion / increase specificity.

alanking:
> It just occurred to me, should we rename the subcommands to something like:
> 
>     * `get_grid_option` or `get_grid_configuration_option`
> 
>     * `set_grid_option` or `set_grid_configuration_option`
> 
> 
> As I was looking through the tests, `get_grid_configuration` kept standing out to me. Seems that if we ever wanted to show the entire configuration, that name may be what we want.
> 
> I'm not trying to create unnecessary work, but I do want to raise the question to see what others thought.

Is the only reason for the change to reserve the name for a potential future `iadmin` subcommand which can get values for all options in a given namespace? If so, I think we should just support passing an empty string for the `OPTION_NAME` in the future. This would be fully backward compatible and could be wrapped into the same subcommand.

Just my two cents. I'm willing to update it, just let me know.

korydraughn:
That's a fine solution for `get_grid_configuration`.

How would that work for `set_grid_configuration`?

alanking:
I feel that `set_grid_configuration` should not support not providing an `OPTION_NAME`.

One downside, IMO, to having `set_grid_configuration_option` and `get_grid_configuration_option` for the purpose of protecting a theoretical future `get_grid_configuration` is that there likely never will be a `set_grid_configuration` because it doesn't make a lot of sense; but, it feels like there should be one for the "parity".

alanking:
By the way, the work is already done in my local branch, so I'm just trying to weigh whether this is what we want.

trel:
`get_grid_configuration_value` ?

It’s a lot to type… but, we can do fancy autocomplete bash stuff to help with that.

As for protecting the smaller namespace for later… I don’t feel too strongly.   The longer name parser better… to agree with the original observation.

korydraughn:
> I feel that `set_grid_configuration` should not support not providing an `OPTION_NAME`.

I agree. That is exactly the reason why I feel the name of the subcommand is slightly incorrect. If passing no arguments resulted in default values being set for all options across a namespace or even across all namespaces, that would fit the name better. However, I can't see allowing that behavior to be a good thing (i.e. executing the wrong command can corrupt their grid config).

> One downside, IMO, to having `set_grid_configuration_option` and `get_grid_configuration_option` for the purpose of protecting a theoretical future `get_grid_configuration` is that there likely never will be a `set_grid_configuration` because it doesn't make a lot of sense; but, it feels like there should be one for the "parity".

I get what you're saying. I don't think the current naming for the subcommands is bad. I wonder if it will be clear to users.

korydraughn:
`get_grid_configuration_value` is a lot to type. Perhaps we just handle this in the new CLI.

I don't want to stop this PR from getting in due to a thing that may never be real.

alanking:
Okay, let's leave the names as they are.

I'm going to run through core tests again and will squash if/when they pass. Still seeing the issue described in #6954 pretty frequently, so, might take me a few attempts.

alanking:
Unit and core tests passed

korydraughn (approved):
Looks like one commit message needs to be updated? That one being for 7274 and about static auth configs.

Everything else looks good.

Once the commit message is handled (if incorrect), please add pounds and merge when ready.

alanking:
Good catch. Done

trel:
so good.   nice work all around.

Issue #5937, opened 2021-10-20
Packages should generate their own filenames


SwooshyCueb:
- [x] 4-2-stable
- [x] master
-----
At present, irods generates the `IRODS_PACKAGE_FILE_NAME_SUFFIX` variable and stores it in the project configuration. This variable is pulled in by other projects to name their packages.
Since we are moving towards proper package versioning, this won't work any more, as changes to the package revision string would affect the package names of unrelated packages.
Instead, packages should generate their own filenames. This can be done by setting `CPACK_RPM_FILE_NAME` to `RPM-DEFAULT` and `CPACK_DEBIAN_FILE_NAME` to `DEB-DEFAULT`.

trel:
moving back to 4.2.11 - done before deadline.

PR #7332, opened 2023-09-18
[7312] Move package revision logic to IrodsCPackCommon.cmake and append EL version to RPM package revision (main)

SwooshyCueb:
Additionally, append EL version to RPM package revision


SwooshyCueb:
- Moving the package revision logic to `IrodsCPackCommon` means that plugin/client projects that rely on this functionality being present in the module will not be able to build against 4.3.0.
- Changing the package revision *might* make passing version numbers to `dnf`/`yum` a little more cumbersome.
- Any scripts relying on assumptions about flat package revisions might need adjustments.
- Any scripts relying on assumptions about package filenames might need adjustments.

Except for the first point, these also applied to the prior change we made to debian packages in #5937.

I'd say these changes are pretty safe.

korydraughn:
> * Moving the package revision logic to `IrodsCPackCommon` means that plugin/client projects that rely on this functionality being present in the module will not be able to build against 4.3.0.

That sounds fine. And given the memory leaks in 4.3.0, I can't see anyone choosing that over 4.3.1.

> * Changing the package revision _might_ make passing version numbers to `dnf`/`yum` a little more cumbersome.

How so? How would someone select the correct version using the changes in this PR?

> * Any scripts relying on assumptions about flat package revisions might need adjustments.
> * Any scripts relying on assumptions about package filenames might need adjustments.

How common is it for package names to change in this way?

alanking (commented):
Seems fine. Just had a couple of questions for my own understanding/future reference.

In order to install 4.3.1 explicitly via `apt` on Ubuntu 20.04, this is what that would look like:
```bash
apt-get install irods-server=4.3.1-0~focal
```

And if we ever bump the revision number for whatever reason, it would be:
```bash
apt-get install irods-server=4.3.1-1~focal
```

Can you give an example of the `dnf` or `yum` command required to explicitly target, for example, 4.3.1 on Rocky Linux 9?

SwooshyCueb:
> > * Changing the package revision _might_ make passing version numbers to `dnf`/`yum` a little more cumbersome.
> 
> How so? How would someone select the correct version using the changes in this PR?

> Can you give an example of the `dnf` or `yum` command required to explicitly target, for example, 4.3.1 on Rocky Linux 9?

If `dnf`/`yum` requires the package release strings when specifying a package version, this will make the command vary depending on the EL version, similar to how the `apt` command already varies depending on the distro codename.

To explicitly install the `irods-server` package built from irods 4.3.1 with the initial package revision on EL9, you'd run `dnf install irods-server-4.3.1-0.el9`. *the `.el9` part is new.

> > * Any scripts relying on assumptions about flat package revisions might need adjustments.
> > * Any scripts relying on assumptions about package filenames might need adjustments.
> 
> How common is it for package names to change in this way?

Package filenames almost always change when the package release string changes.

korydraughn (approved):
That makes sense.

Pound when ready.

alanking (approved):
Got it, thanks. Yes, please to #.

SwooshyCueb:
\#'d

Issue #6810, opened 2023-01-03
Investigate `to_string()`'s usage of `parseMspForStr()` in json_serialization.cpp


korydraughn:
- [x] main
- [ ] 4-3-stable
---
## Investigation
`parseMspForStr` will return a null pointer if MsParam, `_p`, is set to a string value of `"null"`. `parseMspForStr` also returns a null pointer on error. If `parseMspForStr` returns a null pointer, the developer will not know if it represents an error or `"null"`.
https://github.com/irods/irods/blob/ec93f905d05776e67870ff80a92aba4227b926bd/server/core/src/json_serialization.cpp#L14-L23

Below is the function's implementation:
https://github.com/irods/irods/blob/ec93f905d05776e67870ff80a92aba4227b926bd/lib/core/src/msParam.cpp#L957-L975

trel:
still one outstanding checkbox/cherry-pick...

alanking:
I'm not sure that the main checkbox should be checked either considering the new function is not being used anywhere?

korydraughn:
That's true.

PR #7341, opened 2023-09-21
[6810] New parseMspForStr() (main)

FifthPotato:
New function does C-style stuff, returning an integer code (nonzero for error) and return value as 2nd argument. Caller passes in a pointer-to-pointer-to-char to store return value. Removes ambiguity from return values.

Currently named parseMspForStr_new, which is a bad, awful, no-good name. Taking suggestions for new name.

Also, a question: how "C"-like does this need to be? Should I type-annotate it (and use nullptr over NULL) or does it need to be truly C-compatible (i.e. a C compiler wouldn't get mad at it)?


lib/core/src/msParam.cpp
```diff
@@ -1022,6 +1022,28 @@ parseMspForPosInt( msParam_t* inpParam ) {
     return myInt;
 }
 
+int parseMspForStr_new(msParam_t* inpParam, char** outParam)
```
korydraughn, line 1025: Here's some suggestions for naming (ordered). _You're not obligated to use them._
- `msp_to_string` _(This would be nice if all parsing functions had the `msp_` prefix)_
- `parse_msparam_as_string`
- `parse_msp_as_string`
- `parse_msp_for_string` _(yuck)_

The first one fits the namespacing scheme used in many C libs.

If there are better names, please list them.

korydraughn, line 1025: As a convention, prefix parameter names with leading underscores.

lib/core/src/msParam.cpp
```diff
@@ -1022,6 +1022,28 @@ parseMspForPosInt( msParam_t* inpParam ) {
...
+{
+    if (inpParam == NULL || inpParam->inOutStruct == NULL) {
```
korydraughn, line 1027: Within the body of the function we can use C++. That means we can also return `nullptr` from the function even in a C context.

Let's use `nullptr` instead of `NULL`.

_This applies to the other locations too._

lib/core/src/msParam.cpp
```diff
@@ -1022,6 +1022,28 @@ parseMspForPosInt( msParam_t* inpParam ) {
...
+        rodsLog(LOG_ERROR, "%s: inpParam type %s is not STR_MS_T", __func__, inpParam->type);
```
korydraughn, line 1028: Let's use the new logger (i.e. `rodsLog` is going to be removed eventually).

I'm pretty sure the new logger is used in other functions within this file. Look for `log_msi` within this file to find example usage.

_This applies to the other locations too._

lib/core/src/msParam.cpp
```diff
@@ -1022,6 +1022,28 @@ parseMspForPosInt( msParam_t* inpParam ) {
...
+        *outParam = NULL;
+        return 1;
+    }
+
+    if (strcmp(inpParam->type, STR_MS_T) != 0) {
+        rodsLog(LOG_ERROR, "%s: inpParam type %s is not STR_MS_T", __func__, inpParam->type);
+        return 1;
+    }
+
+    if (strcmp((char*) inpParam->inOutStruct, "null") == 0) {
```
korydraughn, line 1038: Prefer the C++ casts rather than the C casts.

_This applies to the other locations too._

lib/core/src/msParam.cpp
```diff
@@ -1022,6 +1022,28 @@ parseMspForPosInt( msParam_t* inpParam ) {
...
+        *outParam = NULL;
+        return 0;
+    }
+
+    *outParam = (char*) (inpParam->inOutStruct);
+    return 0;
+}
```
korydraughn, line 1045: As a convention, we add a trailing comment containing the name of the function. For example:
```c++
} // parseMspForStr_new
```

That helps with understanding where you are in the code. Especially when functions become very long and cannot be viewed in full within your editor.

korydraughn (commented):
Looking good so far.

Once we settle on the naming, etc, we're going to need a unit test.

It also needs documentation.

lib/core/src/msParam.cpp
trel: I like the first one, too - there would be a family of these (`msp_to_*`)

The second one also works...

Perhaps a combination?   `parse_msparam_to_string` or `msparam_to_string` ?

FifthPotato: 1 looks pretty nice. Feels nice and orderly to have a family of happy functions.

trel:
Seeing `msp` alone without context feels a little loose - too much acronym - needs more letters…

I’m leaning towards `msparam_to_string`…

korydraughn:
I get what you're saying, but most C APIs do that. For example, OpenGL uses `gl` as the prefix for many things.

If the things we are talking about were C++ APIs, I'd agree since `msparam` could be the namespace and devs could alias it to `msp` if they wanted to.

`msp` is already unofficially the abbreviation for `MsParam` in the codebase too.

If there's a chance for a naming collision if `msp_` is used, then we should think about it more, but I can't think of a possible collision at this time.

trel:
No collisions, just aesthetics.     Fine as is.

korydraughn:
Yes. We some tests around this.

You should be able to write unit tests for this. See the unit_tests directory for examples.

alanking (commented):
The function itself looks real good.

Besides tests, I guess the other thing we need is... to use it? Presumably by replacing some calls to `parseMspForStr` in the JSON serialization, per #6810?

korydraughn:
> Besides tests, I guess the other thing we need is... to use it? Presumably by replacing some calls to `parseMspForStr` in the JSON serialization, per #6810?

Tests, yes.

Actually using it, not in 4.3.1.

The plan is to begin using the new function starting with 4.3.2 or later. To goal being to avoid potential issues for 4.3.1.

alanking:
Oh. I didn't even look at the milestone. My bad

alanking:
So if we are trying to get the function into 4.3.1, should we make an issue and put it in the milestone? The existing issue can be used to track actually using the function per the earlier conversation. Or did I miss the point again? :)

korydraughn:
Sure. A new issue is good.

FifthPotato:
Added the server-side unit test through the plugins...I think I got it all? As always, taking suggestions on names if my random names are yucky.

plugins/microservices/src/test_msp_functions.cpp
```diff
@@ -0,0 +1,71 @@
...
+    {
+        IRODS_MSI_TEST_BEGIN("testing msp_to_ function: msp_to_string")
+        {
+            auto ret = msp_to_string(nullptr, nullptr);
+            IRODS_MSI_ASSERT(ret == SYS_NULL_INPUT);
+        }
+        {
+            char* good = nullptr;
+            auto ret = msp_to_string(nullptr, &good);
+            IRODS_MSI_ASSERT((ret == SYS_NULL_INPUT && good == nullptr));
```
korydraughn, line 20: If `ret` is `const`, this is fine. If it's not `const`, consider swapping the order of the operands. That protects us from `ret = SYS_NULL_INPUT`.

plugins/microservices/src/test_msp_functions.cpp
```diff
@@ -0,0 +1,71 @@
...
+        }
+        {
+            msParam_t in; //NOLINT(cppcoreguidelines-init-variables)
```
korydraughn, line 23: Is the missing initialization on purpose?

If so, please add a comment explaining why. Otherwise, let's initialize it (`msParam_t in{}`).

plugins/microservices/src/test_msp_functions.cpp
```diff
@@ -0,0 +1,71 @@
...
+            in.inOutStruct = nullptr;
+            char* good = nullptr;
+            auto ret = msp_to_string(&in, &good);
+            IRODS_MSI_ASSERT((ret == SYS_NULL_INPUT && good == nullptr));
+        }
+        {
+            msParam_t in; //NOLINT(cppcoreguidelines-init-variables)
+            in.inOutStruct = malloc(sizeof(int));
+            in.type = strdup(INT_MS_T);
```
korydraughn, line 32: Found some memory leaks.

plugins/microservices/src/test_msp_functions.cpp
```diff
@@ -0,0 +1,71 @@
...
+            char* good = nullptr;
+            auto ret = msp_to_string(&in, &good);
+            IRODS_MSI_ASSERT((ret == USER_PARAM_TYPE_ERR && good == nullptr));
+        }
+        {
+            msParam_t in; //NOLINT(cppcoreguidelines-init-variables)
+            char* nullstr = "null";
+            in.inOutStruct = nullstr;
+            in.type = strdup(STR_MS_T);
```
korydraughn, line 41: Memory leak.

plugins/microservices/src/test_msp_functions.cpp
```diff
@@ -0,0 +1,71 @@
...
+            char* out = nullptr;
+            auto ret = msp_to_string(&in, &out);
+            IRODS_MSI_ASSERT((ret == 0 && out == nullptr));
+        }
+        {
+            msParam_t in; //NOLINT(cppcoreguidelines-init-variables)
+            char* validstr = "S. tuberosum";
+            in.inOutStruct = validstr;
+            in.type = strdup(STR_MS_T);
```
korydraughn, line 50: Memory leak.

plugins/microservices/src/test_msp_functions.cpp
```diff
@@ -0,0 +1,71 @@
...
+            msParam_t in; //NOLINT(cppcoreguidelines-init-variables)
+            in.inOutStruct = malloc(sizeof(int));
+            in.type = strdup(INT_MS_T);
+            char* good = nullptr;
+            auto ret = msp_to_string(&in, &good);
+            IRODS_MSI_ASSERT((ret == USER_PARAM_TYPE_ERR && good == nullptr));
+        }
+        {
+            msParam_t in; //NOLINT(cppcoreguidelines-init-variables)
+            char* nullstr = "null";
```
korydraughn, line 39: Hitting that -fpermissive flag.

ISO C++ doesn't allow string literals to be pointed to by non-const char pointers.

`char*` -> `const char*`

plugins/microservices/src/test_msp_functions.cpp
```diff
@@ -0,0 +1,71 @@
...
+            in.inOutStruct = nullstr;
+            in.type = strdup(STR_MS_T);
+            char* out = nullptr;
+            auto ret = msp_to_string(&in, &out);
+            IRODS_MSI_ASSERT((ret == 0 && out == nullptr));
+        }
+        {
+            msParam_t in; //NOLINT(cppcoreguidelines-init-variables)
+            char* validstr = "S. tuberosum";
```
korydraughn, line 48: `char*` -> `const char*`

scripts/irods/test/test_misc.py
```diff
@@ -337,6 +337,11 @@ def test_scoped_client_identity_updates_zone_of_RsComm__issue_6268(self):
     def test_scoped_permission__issue_7032(self):
         self.admin.assert_icommand(['irule', '-r', 'irods_rule_engine_plugin-irods_rule_language-instance', 'msi_test_scoped_permission', 'null', 'ruleExecOut'])
 
+    @unittest.skipUnless(plugin_name == 'irods_rule_engine_plugin-irods_rule_language', "Not implemented for other REPs.")
+    def test_msp_functions(self):
+        self.admin.assert_icommand(['irule', '-r', 'irods_rule_engine_plugin-irods_rule_language-instance', 'msi_test_msp_functions', 'null', 'ruleExecOut'])
+
```
korydraughn, line 343: Remove extra empty line.

plugins/microservices/src/test_msp_functions.cpp
```diff
@@ -0,0 +1,71 @@
...
+#include "irods/irods_ms_plugin.hpp"
+#include "irods/msParam.h"
+
+#include "msi_assertion_macros.hpp"
+
+namespace
+{
+    auto msi_impl(RuleExecInfo* _rei) -> int
+    {
+        IRODS_MSI_TEST_BEGIN("testing msp_to_ function: msp_to_string")
```
korydraughn, line 12: Tell me about the name of the file and the name of the test?

Is the plan to add more `msp` functions to this file over time?

plugins/microservices/src/test_msp_functions.cpp
```diff
@@ -0,0 +1,71 @@
...
+        {
+            auto ret = msp_to_string(nullptr, nullptr);
```
korydraughn, line 14: `const auto ret`?

lib/core/include/irods/msParam.h
```diff
@@ -195,6 +195,7 @@ int
 parseMspForPhyPathReg( msParam_t *inpParam, keyValPair_t *condInput );
 int
 parseMspForPosInt( msParam_t *inpParam );
+int msp_to_string(msParam_t* _inpParam, char** _outParam);
```
korydraughn, line 198: Let's stick to snake case for function and parameter names.

_Don't forget to update the implementation to use the same casing. Keeps clang-tidy happy/quiet._

korydraughn (commented):
Looks like you figured out the structure. Good job.

Have you run the tests and seen them pass?
Did you know the test framework will write log messages to the log file so you can see what is happening?

plugins/microservices/src/test_msp_functions.cpp
FifthPotato: Yep, that was the original idea, though I'm not sure how watertight it is. Terrell mentioned having a family of these, potentially, so I thought I could consider doing that.
If that's a bad idea though, I'm not against just making this test for msp_to_string only!

lib/core/include/irods/msParam.h
```diff
@@ -195,6 +195,7 @@ int
 parseMspForPhyPathReg( msParam_t *inpParam, keyValPair_t *condInput );
 int
 parseMspForPosInt( msParam_t *inpParam );
+int msp_to_string(msParam_t* inp_param, char** out_param);
```
korydraughn, line 198: Missing leading underscores on parameters.

lib/core/src/msParam.cpp
```diff
@@ -1022,6 +1022,34 @@ parseMspForPosInt( msParam_t* inpParam ) {
     return myInt;
 }
 
+int msp_to_string(msParam_t* inp_param, char** out_param)
```
korydraughn, line 1025: Missing leading underscores on parameters.

Don't forget to update the log statements.

lib/core/src/msParam.cpp
```diff
@@ -1022,6 +1022,34 @@ parseMspForPosInt( msParam_t* inpParam ) {
...
+{
+    if (out_param == nullptr) {
+        log_msi::error("{}: out_param is null", __func__);
+        return SYS_NULL_INPUT;
+    }
+
+    if (inp_param == nullptr || inp_param->inOutStruct == nullptr) {
+        log_msi::error("{}: inp_param is null or _inpParam's inOutStruct is null", __func__);
+        *out_param = nullptr;
```
korydraughn, line 1034: Why set the out param if the caller gives you bad inputs?

_Same question for the other error cases._

lib/core/src/msParam.cpp
```diff
@@ -1022,6 +1022,34 @@ parseMspForPosInt( msParam_t* inpParam ) {
...
+        return SYS_NULL_INPUT;
+    }
+
+    if (strcmp(inp_param->type, STR_MS_T) != 0) {
```
korydraughn, line 1038: What happens if `->type` is null?

lib/core/src/msParam.cpp
```diff
@@ -1022,6 +1022,34 @@ parseMspForPosInt( msParam_t* inpParam ) {
...
+        log_msi::error("{}: inp_param type {} is not STR_MS_T", __func__, inp_param->type);
```
korydraughn, line 1039: Let's wrap the input argument's placeholder in square brackets. That way, if we receive an empty string, it's noticeable.

```
"{}: inp_param type [{}] is not STR_MS_T"
```

plugins/microservices/src/test_msp_functions.cpp
```diff
@@ -0,0 +1,94 @@
...
+        IRODS_MSI_TEST_BEGIN("testing msp_to_string: msp_to_string null cases")
+
+        char* good = nullptr;
+        msParam_t in{};
+
+        auto ret = msp_to_string(nullptr, nullptr);
+        IRODS_MSI_ASSERT(SYS_NULL_INPUT == ret);
+
+        ret = msp_to_string(nullptr, &good);
+        IRODS_MSI_ASSERT((SYS_NULL_INPUT == ret && nullptr == good));
```
korydraughn, line 21: Consider separating individual checks into multiple assertions. That'll make it easy to see which one causes a failure.

_Same thing for other assertions._

alanking (commented):
I think there could be a potential problem with the `free` statements not being reached if any of the tests fail. Not really a big deal as it's just a test, but we probably want to clean em up

lib/core/src/msParam.cpp
```diff
@@ -1022,6 +1022,34 @@ parseMspForPosInt( msParam_t* inpParam ) {
     return myInt;
 }
 
+int msp_to_string(msParam_t* inp_param, char** out_param)
+{
+    if (out_param == nullptr) {
```
korydraughn, line 1027: Consider swapping the order of the operands to avoid accidental writes.

_This applies to other conditionals too._

lib/core/src/msParam.cpp
FifthPotato: I think it was to somewhat align with the "old" behavior where the output is null under any error case. That being said, you're right in that it's probably better just to not touch it so that the return value is the one way to tell whether or not it errors...I'll change that!

plugins/microservices/src/test_msp_functions.cpp
```diff
@@ -0,0 +1,117 @@
+/// \file
+
```
korydraughn, line 2: This can be removed since this is related to doxygen.

korydraughn (commented):
After the comment is resolved, please verify it passes the tests.

Since the MSI is new and not used by anything yet, seeing the python test for this MSI pass is all that's needed. Running the full test suite isn't necessary.

Leave a comment so we know when testing is complete/passing.

FifthPotato:
Ran test_misc and it passes.

korydraughn (approved):
Pound it.

For future reference, don't forget to leave a comment letting us know when things have been squashed. GitHub doesn't notify us about force-pushes.

alanking (approved):

korydraughn:
Don't forget to cherry-pick to 4-3-stable (remove the #'s).

Issue #6954, opened 2023-03-09
Tests are failing with `RE_UNABLE_TO_READ_SESSION_VAR`


alanking:
- [ ] main
- [ ] 4-3-stable
---

### iRODS Version, OS and Version
main/4-3-stable (haven't seen this in 4-2-stable)

### What did you try to do?
Run the test suite

### Expected behavior
Consistently passing/failing tests

### Observed behavior (including steps to reproduce, if applicable)
Tests which should be passing are failing, but only sometimes. This is the common thread between all of the spurious failures:
```python
 --- IrodsSession: icommand executed by [rods#tempZone] [iadmin mkuser otherrods rodsadmin] --- 
Assert Command: iadmin mkuser otherrods rodsadmin
Expecting EMPTY: ['']
  stdout:
    | Level 0: DEBUG: error: unable to read session variable $rodsZoneProxy.
    | line 23, col 27, rule base core
    |   acCreateCollByAdmin("/"++$rodsZoneProxy++"/home", $otherUserName);
    |                            ^
    | 
    | ==========
    | error: unable to read session variable $otherUserName.
    | line 13, col 4, rule base core
    |  ON($otherUserName == "anonymous") {
    |     ^
    | 
    | 
  stderr:
    | remote addresses: 172.23.0.6 ERROR: rcGeneralAdmin failed with error -1213000 RE_UNABLE_TO_READ_SESSION_VAR 
Unexpected output on stderr

FAILED TESTING ASSERTION
```
Please investigate.

alanking:
Through some excavating of logs, it appears that the main server process is crashing. Here are the log messages...

```json
{
  "server_pid": 82168,
  "log_message": "Initializing server ..."                                                                                                                                                                                                                                      
}                                                                                                                                       
{
  "server_pid": 82168, 
  "log_message": "Setting up UNIX domain socket for agent factory ..."
}
{
  "server_pid": 82168,
  "log_message": "Forking agent factory ..."
}
{
  "server_pid": 82168,
  "log_message": "Connecting to agent factory [agent_factory_pid=82169] ..."
}
{
  "server_pid": 82169,
  "log_message": "Initializing agent factory ..."
}
{
  "server_pid": 82168,
  "log_message": null
}
{
  "server_pid": 82168,
  "log_message": null
}
{
  "server_pid": 82168,
  "log_message": "rodsServer Release version rods4.3.0 - API Version d is up"
}
{
  "server_pid": 82168,
[...]

PR #7347, opened 2023-09-27
Fixes and tests for auth configuration (main)

alanking:
In service of #7274 

New test passes and unit tests are passing. Running core tests.

### DIFF
--- a/plugins/database/src/db_plugin.cpp
+++ b/plugins/database/src/db_plugin.cpp
@@ -122,14 +122,16 @@ namespace
     // This structure holds authentication configuration values.
     struct auth_config
     {
-        // Holds the value for the irods::KW_CFG_PAM_PASSWORD_EXTEND_LIFETIME configuration.
-        bool password_extend_lifetime = true;
-
+        static constexpr bool default_password_extend_lifetime = true;
         static constexpr rodsLong_t default_password_max_time = 1209600;
+        static constexpr rodsLong_t default_password_min_time = 121;
+
+        // Holds the value for the irods::KW_CFG_PAM_PASSWORD_EXTEND_LIFETIME configuration.
+        bool password_extend_lifetime = default_password_extend_lifetime;
+
         // Holds the value for the irods::KW_CFG_PAM_PASSWORD_MAX_TIME configuration.
         rodsLong_t password_max_time = default_password_max_time;
 
-        static constexpr rodsLong_t default_password_min_time = 121;
         // Holds the value for the irods::KW_CFG_PAM_PASSWORD_MIN_TIME configuration.
         rodsLong_t password_min_time = default_password_min_time;
     };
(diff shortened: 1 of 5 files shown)


alanking:
Still running into #6954 but eventually getting over the hurdles.

korydraughn:
Hmm. I'll plan to look into that soon. Maybe the reason will be more obvious now.

alanking:
Core tests passed

korydraughn:
Excellent. Will review as soon as I can.

plugins/database/src/db_plugin.cpp
```diff
@@ -145,47 +147,64 @@ namespace
...
-                        return ERROR(CONFIGURATION_ERROR,
-                                     fmt::format("Invalid configuration for option [{}] in namespace [{}]: [{}]",
-                                                 option_name,
-                                                 _namespace,
-                                                 _out.password_min_time));
+                const auto option_value = result.get<std::string>(1);
+
+                if (option_name == irods::KW_CFG_PAM_PASSWORD_MIN_TIME ||
+                    option_name == irods::KW_CFG_PAM_PASSWORD_MAX_TIME) {
+                    const bool min = (option_name == irods::KW_CFG_PAM_PASSWORD_MIN_TIME);
```
korydraughn, line 154: Consider finding a way to do the following string comparison once.
```
option_name == irods::KW_CFG_PAM_PASSWORD_MIN_TIME
```

korydraughn, line 154: Consider renaming `min`.

Perhaps something like ...
- `is_min_option`
- `is_min_option_name`
- `is_min`
- `found_min_option`

I don't have strong feelings on this. Up to you.

scripts/irods/test/test_pam_password.py
```diff
@@ -0,0 +1,340 @@
...
+                self.auth_session.environment_file_contents.update(client_update)
+
+                with temporary_core_file() as core:
+                    core.add_rule(test_configurations.get_pep_for_ssl(self.plugin_name))
+
+                    IrodsController().start()
+
+                    # Make sure the settings applied correctly...
+                    self.admin.assert_icommand(
+                        ['iadmin', 'get_grid_configuration', self.configuration_namespace, _option_name], 'STDOUT')
```
korydraughn, line 134: If the settings were not applied correctly, I take it this assertion would fail, right?

Does it make sense to assert a specific value in the output?

scripts/irods/test/test_pam_password.py
```diff
@@ -0,0 +1,340 @@
...
+                with temporary_core_file() as core:
+                    core.add_rule(test_configurations.get_pep_for_ssl(self.plugin_name))
+
+                    IrodsController().start()
+
+                    # Make sure the settings applied correctly...
+                    self.admin.assert_icommand(
+                        ['iadmin', 'get_grid_configuration', self.configuration_namespace, max_time_option_name], 'STDOUT')
+
+                    for base_ttl_in_hours in [2, 336]:
```
korydraughn, line 203: Are 2 and 336 special?

korydraughn (commented):
Looks good.

plugins/database/src/db_plugin.cpp
alanking: Done, I think

scripts/irods/test/test_pam_password.py
alanking: That's right. In the event that the settings were not applied correctly, an error would occur most likely in the SSL negotiations, resulting in a non-zero return code.

That said,  I have updated these calls to look for the output for these configurations as returned by `get_grid_configuration`.

scripts/irods/test/test_pam_password.py
alanking: Yes and no :)

2 is the lowest possible value that we can test with (that is, 2 - 1 = 1, which is the smallest TTL that we can specify with `iinit`). 336 is 1209600 seconds, which is the default `max_password_time`, so it seemed like these were appropriate as boundary tests.

I added a comment explaining why these were chosen for the subTests.

plugins/database/src/db_plugin.cpp
alanking: I would also like to point out that this section is the part I like least of these changes, but I couldn't think of a better way to do it in this timeframe nor do I think it's worth putting much more effort into. Maybe that's just me :)

korydraughn: Yep, understood. It passes the tests and hopefully no one has to touch this chunk of code again for a long time.

scripts/irods/test/test_pam_password.py
korydraughn: Ah, that makes sense.

korydraughn (approved):
If you're happy with it, please squash, pound, and merge.

PR #7348, opened 2023-09-27
Fixed init bug in connection pool constructor (main)

korydraughn:
This is a trivial fix. Running tests aren't necessary.

### DIFF
--- a/lib/core/src/connection_pool.cpp
+++ b/lib/core/src/connection_pool.cpp
@@ -135,13 +135,14 @@ namespace irods
         , proxy_username_{std::move(_proxy_username)}
         , username_{std::move(_username)}
         , auth_func_{std::move(_auth_func)}
-        , conn_ctxs_(_size)
         , options_{_options}
     {
         if (_size < 1) {
             // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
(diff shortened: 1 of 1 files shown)


korydraughn:
That's pretty funny. The contents of the container isn't movable.

Let's see if I can get around that.

alanking (commented):
Seems... fine? Perhaps a description in the commit message or an issue describing the problem will shed some light for me.

korydraughn:
The issue is ...
- The constructor can take a negative integer value for the size
- The size is passed directly to the vector's constructor, which expects a positive integer so that it can allocate the requested number of connections
- The size is checked after the negative integer value has already been used

The vector's constructor will likely wrap the value which isn't what we want.

korydraughn:
All of that will be captured in the issue, etc.

alanking:
Makes sense to me. Thanks!

korydraughn:
Just noting here that the unit tests for the connection pool pass. I'll run the rest of the unit tests to be on the safe side.

alanking (commented):
Please squash and add issue number (no #) when ready

korydraughn:
Squashed.

korydraughn:
All unit tests pass.

Ready to pound.

alanking (approved):
Yep, please #

korydraughn:
#'d

PR #7356, opened 2023-09-29
Make auth config namespace more general, fix native auth config, and add more tests (main)

alanking:
Addresses #7274 (finally)

Unit tests pass and all the new auth tests pass. Running the rest of the core tests now.

PR #7362, opened 2023-10-02
Make auth config namespace more general, fix native auth config, and add more tests (4-3-stable)

alanking:
Addresses #7274

Cherry-picked from #7356


korydraughn (approved):
Pound it.

Issue #7000, opened 2023-03-28
Topology tests are failing in main/4-3-stable


alanking:
- [x] main
- [x] 4-3-stable
---
Separated from #6976 so the milestone can move forward.

Consider cherry-picking changes from https://github.com/irods/irods/pull/6995 in addition to any other changes which are required.

alanking:
We have confirmed that topology tests are passing for both provider and consumer.

PR #7363, opened 2023-10-02
[#7000] Fix tests for running in topology (main)

alanking:
Possibly addresses #7000

Cherry-picked from https://github.com/irods/irods/pull/6995

---

Haven't run topology tests all the way through (keeps failing due to #6954 at random places) but putting this up for transparent progress.

### DIFF
--- a/lib/api/include/irods/apiTable.hpp
+++ b/lib/api/include/irods/apiTable.hpp
@@ -1121,7 +1121,7 @@ static irods::apidef_t client_api_table_inp[] = {
         GET_RESC_QUOTA_AN, RODS_API_VERSION, REMOTE_USER_AUTH, REMOTE_USER_AUTH,
         "getRescQuotaInp_PI", 0, "rescQuota_PI", 0,
         boost::any(std::function<int(rsComm_t*,getRescQuotaInp_t*,rescQuota_t**)>(RS_GET_RESC_QUOTA)),
-        "api_get_resc_quota", irods::clearInStruct_noop, clearDataObjInfo,
+        "api_get_resc_quota", clearRescQuotaInp, clearRescQuota,
         (funcPtr)CALL_GETRESCQUOTAINP_RESCQUOTAOUT
     },
     {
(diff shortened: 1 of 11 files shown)


scripts/irods/test/test_native_rule_engine_plugin.py
```diff
@@ -92,6 +92,9 @@ def assert_presence_of_attr_value_avu(attr, value, assert_true=True):
             assert_presence_of_attr_value_avu(attr, value, assert_true)
 
     def helper_test_pep(self, rules_to_add, icommand, strings_to_check_for=['THIS IS AN OUT VARIABLE'], number_of_strings_to_look_for=1):
+        # Restart the server here because sometimes the log just stops receiving messages for some unknown reason.
```
SwooshyCueb, line 95: Is there an issue for this?

alanking: I don't think so. I'm not even sure that this affects 4.3 since the logging system is different from 4.2 where this is cherry-picked from. Maybe we should try without the change to see if it's needed...?

korydraughn: This may be resolved by the move to syslog, but let's confirm that first.

If we can't confirm it, then let's remove the call to restart and keep it moving. We can always open more issues.

alanking: Tests seem to pass without these, so, I guess we don't need them.

korydraughn:
I ran **tip of 4-3-stable** (c08f4cdf5c525d70db2e7dad87aeb3c9d3e2181a) through the topology tests for the provider and all tests passed except test_iadmin.py.

I took your change to test_iadmin.py (minus the skip annotation) and it passed so (topology + provider) is in good shape.

I noticed the commits have been squashed. Are the topology tests + consumer happy now?

alanking:
There are two tests still failing:
`test_resource_types.Test_Resource_Unixfilesystem.test_detached_mode_in_topology_put_with_resource_host_down__issue_4421`
`test_resource_types.Test_Resource_Unixfilesystem.test_detached_mode_in_topology_with_host_list__issue_4421`

alanking:
The remaining tests were failing due to a bug in the testing environment, which is being fixed here: https://github.com/irods/irods_testing_environment/pull/172

korydraughn (approved):

PR #7365, opened 2023-10-03
[#6412] Bump version number to 4.3.1 (4-3-stable)

korydraughn:

Issue #7361, opened 2023-09-30
Proposal: .9x versioning for main branch


SwooshyCueb:
This is one of two related proposals meant to address related problems.

At present, we increment the version number of the main branch to that of the next major release once it has diverged significantly from the current stable series branch. This means that the major release itself may not have a higher version number than a development build from the same branch from before the release. This poses a problem for packaging, as it may not result in a release taking priority over an older development build. In our stable series branches, we avoid this by incrementing version numbers just before release.

I am proposing that instead of incrementing to the next major release on significant divergence from stable, we follow the example set by Gnome and increment to a minor release with a value of .90 (ex: 4.3.90) and increment to the next major version just before release.

With this change to our versioning, downstream projects being developed side-by-side with iRODS will still have a version number to target, but we will not run into the version number precedent issue described above.

This change also comes with some additional benefits:
- If a downstream project that inherits iRODS' version number is built against the main branch rather than a release, it will be evident in the version number.
- When a user reports an issue with iRODS (or a downstream project that inherits iRODS' version number), we can tell from the version number whether they are using a development build from the main branch.

trel:
This makes sense to me and I don't see any real downsides.

We could implement this *today* on main, I think, without any issues?   Thought experiments welcome.

SwooshyCueb:
I think we should continue to track the version of the stable branch until main has diverged. AFAIK right now main and 4-3-stable are exactly or almost exactly the same in terms of content.

trel:
I think that's where I am as well.

I think we wait until at least 4.3.1 is out before we touch this though...

SwooshyCueb:
For 5.0, we can reclaim a dot, and do this for the stable branches as well, avoiding the need for even/odd versioning

SwooshyCueb:
If we decide to move forward with .9x versioning (and it seems like we will), we should write our own CMake package version file template that will match .9x versions as belonging to the next major release (i.e. 4.4(/5.0) will match for downstream projects looking for a minimum of 4.3.90).

trel:
Also want to note that the reclaiming of a dot is interesting/good for the .90 scenario... because the plugins don't need / can't use the extra dot (they're the end of the line, nobody is using them to build against).

korydraughn:
And development builds (i.e. those that have .90 in the package name) are not a concern for the upgrade process.

[...]

Issue #6412, opened 2022-05-21
release activities for 4.3.1


trel:
- [x] Server Version Bump
- [x] iCommands Version Bump
- [x] Plugin Version Bumps
- [x] Release Notes

korydraughn:
Don't forget to include the database schema version bump in the release notes.

See https://github.com/irods/irods/pull/6934

SwooshyCueb:
When bumping the version to 4.3.1, make sure to reset the package revisions back to `0`, for *every repo* we're releasing from.

korydraughn:
The icommands shouldn't need any changes since the version number is obtained from the dev package.

korydraughn:
I take that back, we do have to bump the version for `find_package`.

trel:
so... this shouldn't be set to `0`?

https://github.com/irods/irods_client_icommands/blob/4-3-stable/CMakeLists.txt#L8

korydraughn:
Yes, that too needs to be changed to 0.

Creating the PR for that now.

PR #7394, opened 2023-10-25
[#6412] Bump version number to 4.3.1. (main)

alanking:
In service of #6412

Cherry-picked from #7365 

I imagine this will be changed very soon due to #7361?


korydraughn (approved):
Pound it once @SwooshyCueb approves.

Doxyfile
```diff
@@ -38,7 +38,7 @@ PROJECT_NAME           = "iRODS"
 # could be handy for archiving the generated documentation or if some version
 # control system is used.
 
-PROJECT_NUMBER         = main
+PROJECT_NUMBER         = 4.3.1
```
SwooshyCueb, line 41: This should probably still say main? Not sure

trel: yes please leave as main

korydraughn: Ultimately this only matters when generating documentation.

I can see it going either way, but to keep things simple, let's set it to `main`.

Perhaps this file becomes a .in file and the project number is assigned at build time. Then we don't ever have to touch this file in relation to the version number again.

alanking: Good catch.

alanking: Created https://github.com/irods/irods/issues/7395 for the .in file idea. Resolving.

SwooshyCueb (approved):

PR #7411, opened 2023-11-03
[#6497] Require icp -f -R to target resource with replica (main)

alanking:
Addresses #6497 

New tests all work for federation, core, and topology (provider and consumer). Running the rest of the tests now

Issue #6497, opened 2022-07-25
`-R` option in `icp` should be a directive, not a preference


alanking:
- [x] main
- [x] 4-3-stable
---
### iRODS Version, OS and Version
4.3.0 (and probably back into much older versions)

### What did you try to do?
Overwrite a data object via `icp -f -R` to a resource which does not have a replica.

Let's say I have a data object `foo` on `demoResc` and a data object `foo0` on resource `ufs0`:
```
$ ils -l
/tempZone/home/rods:
  rods              0 demoResc          185 2022-07-25.17:53 & foo
  rods              0 ufs0          239 2022-07-25.17:53 & foo0
```

From the help text:
```
-R resource - specifies the resource to store to. This can also be specified
       in your environment or via a rule set up by the administrator.
```

### Expected behavior
I would expect the behavior to be consistent with `iput -f -R` to a resource which does not have a replica for an existing data object:
```
$ iput -f -R ufs1 version.json foo0
remote addresses: 192.168.240.3 ERROR: putUtil: put error for /tempZone/home/rods/foo0, status = -1803000 status = -1803000 HIERARCHY_ERROR
```

### Observed behavior (including steps to reproduce, if applicable)
The resource specified by `-R` is ignored and the existing replica on `ufs0` is overwritten:
```
$ icp -f -R ufs1 foo foo0 
$ ils -l
/tempZone/home/rods:
  rods              0 demoResc          185 2022-07-25.17:53 & foo
  rods              0 ufs0          185 2022-07-25.17:53 & foo0
```
If a replica exists on `ufs1`, the data object is appropriately overwritten by `icp -f -R`:
```
$ ils -l
/tempZone/home/rods:
  rods              0 demoResc          239 2022-07-25.17:57 & foo
  rods              0 ufs0          185 2022-07-25.17:53 & foo0
  rods              1 ufs1          185 2022-07-25.17:56 & foo0
$ icp -f -R ufs1 foo foo0 
$ ils -l
/tempZone/home/rods:
  rods              0 demoResc          239 2022-07-25.17:57 & foo
  rods              0 ufs0          185 2022-07-25.17:53 X foo0
  rods              1 ufs1          239 2022-07-25.17:57 & foo0
```

PR #7415, opened 2023-11-09
[#6497] Require icp -f -R to target resource with replica (4-3-stable)

alanking:
Addresses #6497

Cherry-picked from #7411


korydraughn (approved):
Pound it.

PR #7416, opened 2023-11-10
[7293] Changed mail microservice to use boost::process instead of system() (main)

FifthPotato:
Tested by hand to check functionality is as expected, but haven't run test suites yet.
A few questions:
I deleted the block with the solaris_platform stuff; is that okay? ...do we even support Solaris?
I also deleted the block for empty subjects, since in my testing, an empty subject still sent properly, but with no subject line.
Finally, not sure if checkStringForSystem() is the best name anymore for that function. I did a recursive grep and couldn't find anything else using it, but maybe we need to keep it for other peoples' clients? Should I write a new function with a better name and leave it as-is?

### DIFF
--- a/lib/core/src/stringOpr.cpp
+++ b/lib/core/src/stringOpr.cpp
@@ -316,6 +316,8 @@ checkStringForSystem( const char * inString ) {
     if ( inString == NULL ) {
         return 0;
     }
+    // Now that boost::process is being used, no need to check for shell injections
+    // Disallowing \r and \n is by spec
     if (boost::regex_match(inString, boost::regex("[^\r\n]*"))) {
         return 0;
     }
--- a/lib/core/src/stringOpr.cpp
+++ b/lib/core/src/stringOpr.cpp
@@ -316,7 +316,7 @@ checkStringForSystem( const char * inString ) {
     if ( inString == NULL ) {
         return 0;
     }
-    if ( boost::regex_match( inString, boost::regex( "[a-zA-Z0-9,./ ]*" ) ) ) {
+    if (boost::regex_match(inString, boost::regex("[^\r\n]*"))) {
         return 0;
     }
     return USER_INPUT_STRING_ERR;
--- a/server/re/src/mailMS.cpp
+++ b/server/re/src/mailMS.cpp
@@ -1,4 +1,4 @@
-// \file
+/// \file
 
 #include "irods/icatHighLevelRoutines.hpp"
 
@@ -10,6 +10,10 @@
 #include <sys/types.h>
 #include <pwd.h>
 
+namespace
+{
+    using log_re = irods::experimental::log::rule_engine;
+} // anonymous namespace
 /**
  * \fn msiSendMail(msParam_t* xtoAddr, msParam_t* xsubjectLine, msParam_t* xbody, ruleExecInfo_t *)
  *
@@ -20,7 +24,8 @@
  * \since pre-2.1
  *
  *
- * \note   This microservice sends e-mail using the mail command in the unix system. No attachments are supported. The sender of the e-mail is the unix user-id running the irodsServer.
+ * \note   This microservice sends e-mail using the mail command in the unix system. No attachments are supported. The
+ *sender of the e-mail is the unix user-id running the irodsServer.
  *
  * \usage See clients/icommands/test/rules/
  *
@@ -42,7 +47,7 @@
  * \pre none
  * \post none
  * \sa none
-**/
+ **/
 int msiSendMail( msParam_t* xtoAddr, msParam_t* xsubjectLine, msParam_t* xbody, ruleExecInfo_t* ) {
 
     const char * toAddr = ( char * ) xtoAddr->inOutStruct;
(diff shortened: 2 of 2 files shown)


lib/core/src/stringOpr.cpp
```diff
@@ -316,7 +316,7 @@ checkStringForSystem( const char * inString ) {
     if ( inString == NULL ) {
         return 0;
     }
-    if ( boost::regex_match( inString, boost::regex( "[a-zA-Z0-9,./ ]*" ) ) ) {
+    if (boost::regex_match(inString, boost::regex("[^\r\n]*"))) {
```
trel, line 319: let's get a comment to say in words what this regex is doing (and why)... to be kind for the next reader/debugger.

server/re/src/mailMS.cpp
```diff
@@ -87,20 +88,11 @@ int msiSendMail( msParam_t* xtoAddr, msParam_t* xsubjectLine, msParam_t* xbody,
         return SYS_MALLOC_ERR;
     }
 
-#ifdef solaris_platform
```
trel, line 90: removing solaris is fine, yes.

server/re/src/mailMS.cpp
```diff
@@ -87,20 +88,11 @@ int msiSendMail( msParam_t* xtoAddr, msParam_t* xsubjectLine, msParam_t* xbody,
...
-        sprintf( mailStr, "cat %s| mail -s '%s'  '%s'", fName, subjectLine, toAddr );
-    }
-    else {
-        sprintf( mailStr, "cat %s| mail  '%s'", fName, toAddr );
-    }
-#endif
     int ret = 0;
-    ret = system( mailStr );
+    namespace bp = boost::process;
+    ret = bp::system(bp::search_path("mail"), "-s", subjectLine, toAddr, bp::std_in < fName);
```
trel, line 93: can combine to `int ret = ...` ?

trel (commented):
and did we figure out how to test this?   throw some edge cases at it to confirm it behaves?  thinking about when we expand to other operating systems again, we want to have bumpers to know where we fall over / hit the corners.

lib/core/src/stringOpr.cpp
korydraughn: Now that we're using boost process, is this check still necessary?

server/re/src/mailMS.cpp
```diff
@@ -87,20 +88,11 @@ int msiSendMail( msParam_t* xtoAddr, msParam_t* xsubjectLine, msParam_t* xbody,
...
     if ( ret ) {
-        irods::log( ERROR( ret, "mailStr command returned non-zero status" ) );
+        irods::log(ERROR(ret, "boost::process::system command returned non-zero status"));
```
korydraughn, line 95: Boost process is an implementation detail and doesn't provide enough information to help the admin know where to start looking.

How do you feel about something like ... `msiSendMail error: <error_code>`?

That immediately tells the admin it's specific to this MSI.

korydraughn, line 95: Let's use the new logger instead of `irods::log`.

server/re/src/mailMS.cpp
```diff
@@ -1,10 +1,11 @@
-/// \file
+// \file
```
korydraughn, line 1: Need that third forward slash for doxygen. Without it, doxygen may not produce documentation for this file.

korydraughn (commented):
Good stuff.

lib/core/src/stringOpr.cpp
FifthPotato: Ah, so this check is actually for spec reasons: subject lines aren't allowed to include \r or \n by spec...thing is, in my tests, `mail` handles it just fine (by simply not adding a subject line)...but I thought it'd be good to still check.

trel: yep - leave it and a comment saying the 'why'.

korydraughn:
There should be information on which tests failed. How many tests are failing?

korydraughn:
Also, please mark any review comments you've handled as resolved.

korydraughn:
Good detective work.

I see a couple options for resolving this.
- (a) Declare a package dependency for installing mail
- (b) Detect whether the mail binary is installed and respond accordingly _(your current solution)_
- (c) We replace the use of the mail binary with libcurl's SMTP support

I think we have to do (a) and (b).

Thoughts?

FifthPotato:
> Good detective work.
> 
> I see a couple options for resolving this.
> 
>     * (a) Declare a package dependency for installing mail
> 
>     * (b) Detect whether the mail binary is installed and respond accordingly _(your current solution)_
> 
>     * (c) We replace the use of the mail binary with libcurl's SMTP support
> 
> 
> I think we have to do (a) and (b).
> 
> Thoughts?

Since we agreed that maybe adding more deps is bad, I'll see how painful it is to do (c) and maybe we can go from there.

alanking:
@FifthPotato - What's the status of this PR?

FifthPotato:
> @FifthPotato - What's the status of this PR?

Dormant? Currently focusing on some of the other tickets.

korydraughn:
This isn't high priority right now. Let's focus on other issues first. We'll come back to this later.

korydraughn:
`msiSendMail` is now deprecated.

This PR can be closed.

Issue #5096, opened 2020-09-03
pam_password_max_time does not allow pass_expiry_ts above 1209600


bh9:
- [x] main
- [x] 4-3-stable
---
## Bug Report
Probably related to #3742 #4198 and #4904 
### iRODS Version, OS and Version
4.2.7 Ubuntu 18.04

### What did you try to do?
Set `pam_password_max_time` in `/etc/irods/server_config.json` to allow for longer pam authentication (in this case to 5184000 for 60 days)

### Expected behavior
`iinit --ttl 1400` should give me an `pass_expiry_ts` of `5040000` since that is less than the maximum

### Observed behavior (including steps to reproduce, if applicable)
`iinit --ttl 1400` gives me a `pass_expiry_ts` of `1209600`

Of note: without the `pam_password_max_time` setting, `iinit --ttl 1400` gives 2 prompts: the first PAM prompt as expected, and then an iRODS password prompt afterwards. With the setting, this second prompt does not occur, so it does change something.

bh9:
I also tried the below config from #4904, but that didn't remove the 2nd prompt
```
'authentication' : { 
    'pam' : { 
        'password_max_time': 5184000
    } 
}
```
Combining the two options also gave the unexpected iRODS password prompt
```
...
  "pam_password_max_time": 5184000,
  "plugin_configuration": {
    "authentication": {
      "pam" : {
        "password_max_time": 5184000
      }
    },
...
```

darthtux:
This issue is of particular interest to us as well.  As bh9 said, no matter what we do we are unable to get an iRODS pass_expiry_ts to last longer then 2 weeks.  

@jasoncoposky This is the PAM issue I was talking about during our meeting yesterday.

jasoncoposky:
I believe the issue is [here](https://github.com/irods/irods/blob/master/plugins/database/src/db_plugin.cpp#L1917) where we first need to get the `plugins {}` object of the server configuration, then the `authentication{...}` section, then the `pam {...}` section.

bzizou:
Same problem here. We're using PAM and we are unable to do `iinit` with the `--ttl` option. It asks for the PAM password, and then the unexpected iRods password, and returns without any error, but the auth is not ok:

```
bzizou@f-dahu:~$ iinit --ttl 9000
Enter your current PAM password:
Enter your current iRODS password:
bzizou@f-dahu:~$ ils
[...]

Issue #3742, opened 2017-11-29
pam_password_max_time not working


kuntzagk:
- [x] main
- [x] 4-3-stable
---

I have two conflicting interpretations, what this pam_password_max_time should achieve. 

###  1st interpretation

After pam_password_max_time (hours/seconds ?) the user is asked again for his password

### 2nd interpretation

After pam_password_max_time the irods server transparently checkes the cached password again using PAM

### test

I put pam_password_min_time and pam_password_max_time into server_config.json (values 1&2), restarted the server.

As a user with "irods_authentication_scheme": "PAM", I did an iinit. 
This is logged in /var/log/secure
`Nov 28 09:35:39 irods_server PamAuthCheck: pam_warn(irods:auth): function=[pam_sm_authenticate] service=[irods] terminal=[<unknown>] user=[XXXX] ruser=[<unknown>] rhost=[<unknown>]`

After 24 hours I can still do an "ils" and do not see any additional PAM actions logged in /var/log/secure.

kuntzagk:
After digging around in the code (my C++ is a little bit rusty) I at least found this:

- in table R_USER_PASSWORD the pass_expiry_ts is set to 1209600 (the default value), so the setting in server_config.json seems to be ignored.

- according to  irods/lib/core/include/irods_configuration_keywords.hpp the configuration keyword is "password_max_time" , but setting this does not work either

trel:
Ah, good digging.

othmar-weber:
Any progress on this one?

jasoncoposky:
@othmar-weber, @kuntzagk 

Per the database plugin, it appears the original intent was to only honor those settings when the TTL was set during the ```iinit``` process.  If you set a ttl, do you get the expected behavior?  

[The code for reference](
https://github.com/irods/irods/blob/master/plugins/database/src/db_plugin.cpp#L7369)

We can discuss whether this makes sense and what the future behavior should be, once we confirm there is no bug with the intended behavior.

jasoncoposky:
@othmar-weber, @kuntzagk 

Any updates regarding the use of the TTL?

othmar-weber:
@jasoncoposky , @kuntzagk 
Not yet, sorry. Andreas will test that when he is back from vacation.

alanking:
I think that this is fixed now if we are going by the 1st interpretation in OP.

korydraughn:
Hmm, aren't both interpretations related?

[...]

PR #7417, opened 2023-11-11
[#3742][#5096] Add test for password_max_time > 1209600 (main)

alanking:
Adds test for #3742 and #5096

The tests both pass.

### DIFF
--- a/scripts/irods/test/test_native_authentication.py
+++ b/scripts/irods/test/test_native_authentication.py
@@ -93,8 +93,9 @@ class test_configurations(unittest.TestCase):
                     self.admin.assert_icommand(
                         ['iadmin', 'set_grid_configuration', self.configuration_namespace, max_time_option_name, option_value])
 
-                    # Note: The min/max check does not occur when no TTL parameter is passed, for some reason.
-                    # We must pass TTL explicitly for each test. Even though this is native authentication,
+                    # Note: The min/max check does not occur when no TTL parameter is passed. If no TTL is passed, the
+                    # password never expires (i.e. no TTL). Therefore, to test TTL lifetime configurations, we must pass
+                    # TTL explicitly for each test. Even though this is native authentication,
                     # PAM_AUTH_PASSWORD_INVALID_TTL is the returned error.
 
                     # This is lower than the minimum and higher than the maximum. The TTL is invalid.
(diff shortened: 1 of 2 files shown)


scripts/irods/test/test_native_authentication.py
```diff
@@ -354,3 +354,51 @@ def test_password_extend_lifetime_set_to_false_invalidates_other_authentications
...
+            # about that value.
+            base_ttl_in_hours = 336 + 2
+            base_ttl_in_seconds = base_ttl_in_hours * 3600
+
+            # Set password_max_time to the value for the test.
+            option_value = str(base_ttl_in_seconds)
+            self.admin.assert_icommand(
+                ['iadmin', 'set_grid_configuration', self.configuration_namespace, max_time_option_name, option_value])
+
+            # Note: The min/max check does not occur when no TTL parameter is passed, for some reason.
```
korydraughn, line 380: Seems this behavior is surprising to you.

Do you feel it should always be checked?

korydraughn (approved):
Tests look good to me.

Let's get one more approval before pounding.

scripts/irods/test/test_native_authentication.py
alanking: I guess this may be okay now that we've changed the behavior for min/max password time. Here's the behavior as I understand it...

 - For native authentication, if no TTL is passed, the password never expires. We don't need to check the TTL in this case because there is no TTL.
 - For pam_password authentication, if no TTL is passed, the configured `min_password_time` is used. We don't need to check the TTL in this case because it is by definition in the range.

I think the "for some reason" is from the test I copied and it was confusing in the past because PAM/pam_password would use a third configuration as the default TTL if none was specified, and this value could fall outside the range of the min/max password times.

korydraughn: Ok. Sounds like we can remove that part then. Your explanation seems to cover all cases.

You agree?

alanking: Agreed. Added explanations in place of the "for some reason"s.

korydraughn (approved):
Let's go ahead and get this in.

Squash/Pound it.

alanking:
#'d. Mergin.

Issue #7104, opened 2023-05-25
Create `rc_replica_truncate` POSIX-like `truncate(2)` API


korydraughn:
- [x] main
- [x] 4-3-stable
---
## Investigation / Refactor
Clients such as NFSRODS need to be able to truncate (resize) data objects. There are UNIX tools which rely on this behavior. To support this, clients must simulate the operation. This results in degraded performance.

The current implementation of `rxDataObjTruncate` is partially there. All that's needed is for it to honor logical locking.

Given that truncation is a destructive operation, the data object being targeted should only be modified if it has a replica status of _good_ or _stale._ The truncation operation must fail immediately if the data object has a replica status of _intermediate._

This issue is motivated by the following conversation.
- https://github.com/irods/irods_client_nfsrods/issues/178

alanking:
A set of requirements has formed... I will number them for easy reference in any further discussion.

1. Introduce 1 or 2 new API plugins to mimic `truncate` and `ftruncate`. These may or may not deprecate `rxDataObjTruncate`. We will try to follow `truncate` behavior as closely as possible.
2. Only one replica can be truncated at a time. This implies support for targeting a replica number and a root resource. If none is supplied, the default resource will be targeted. Hierarchy resolution will determine which replica will be targeted. I propose that it operates on a "write" operation. This differs from the existing `rxDataObjTruncate` which truncates all replicas to the specified size.
3. Open objects can be targeted for truncate and only the intermediate replica can be truncated.
4. `truncate` cannot create files, so these APIs will not create data objects or replicas.
5. We need to write extensive unit tests to make sure things are working as intended because this operation is destructive to data.
6. There may or may not be a new iCommand to go with the new API (`itrunc` or `itruncate`). This is definitely not a priority item.

alanking:
This may require implementing a new resource plugin operation for `ftruncate`. The unixfilesystem resource implementation of the truncate resource plugin operation literally just calls `truncate`. In order to support `ftruncate` we would have to...

a. modify the signature of the operation to accept a file descriptor
b. smuggle it in via a keyword
c. create a different operation for `ftruncate`

I think implementing a separate plugin operation (c) is the cleanest choice.

korydraughn:
I agree.

korydraughn:
Noting here this will affect any plugins which track changes in data objects (e.g. logical quotas).

alanking:
[...]

PR #7463, opened 2024-01-10
[#7104] Honor logical locking in rsDataObjTruncate (main)

alanking:
Addresses #7104...?

I'm opening the PR for eyeballs. The new unit test fails appropriately before the C++ changes and passes appropriately after the C++ changes. Still need to run all the other tests.

I have a few questions that we can discuss here or offline:
1. As has been discussed elsewhere, the expectation was to implement a new API endpoint. I figured it would be just as well to modify the existing endpoint. Do we want to introduce a new API over modifying the existing one? Why?
2. If we make a new API, should we deprecate the existing one?
3. Do we want to expand testing for truncate as part of this effort? Currently, there are no tests for this API.

### DIFF
--- a/server/api/src/rsDataObjTruncate.cpp
+++ b/server/api/src/rsDataObjTruncate.cpp
@@ -5,7 +5,9 @@
 #include "irods/fileTruncate.h"
 #include "irods/getRemoteZoneResc.h"
 #include "irods/icatDefines.h"
+#include "irods/irods_logger.hpp"
 #include "irods/irods_resource_backport.hpp"
+#include "irods/logical_locking.hpp"
 #include "irods/modDataObjMeta.h"
 #include "irods/objMetaOpr.hpp"
 #include "irods/rcGlobalExtern.h"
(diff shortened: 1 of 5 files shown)


korydraughn:
My first thought on this is ... should we make this an API plugin (that lives in its own repo)?

The reason for that thought is due to NFSRODS. It's next release is very close, but depends on whether we want to add support for truncate. I think it would be a nice feature to have.

Think on it and we can discuss tomorrow.

Will think about your questions in the meantime.

trel:
your 4. seems... reasonable?   any weird side effects we can predict?   it's just a replica, we do stuff to it, we close it... and then the other replicas may be informed/updated as well to match.

alanking:
When an object is locked, it means one of the replicas is intermediate with the assumption that the other replicas are at rest. The API currently truncates *all replicas*. Saying "yes" to 4 may prove tricky. There's currently no support for targeting a specific replica to truncate using this API (as far as I can tell). This definitely warrants a discussion about The Future.

korydraughn:
Shouldn't need the replica access token in 99% of cases. I can't think of a reason why someone would need that, but you never know.

(4) sounds fine. It kind of removes the need for logical locking in the truncate implementation. If clients are required to open a replica, then on success, the locking has already happened.

alanking:
After discussing things, I think I'm going to close this PR. I'll leave the branch around in my repo and will post more updates in #7104.

PR #7481, opened 2024-01-31
[7474] Changed ibun tar format from ustar to pax (4-3-stable)

JustinKyleJames:
All tests passed.

I verified the new test failed when the code changes were not included and passed with the code changes.

PR #7486, opened 2024-02-02
[7474] Changed ibun tar format from ustar to pax (main)

JustinKyleJames:
This is the exact same as the pull request for 4-3-stable.  All tests were re-run and passed.  I also verified the new test passed when the skip was removed and failed when the code change was not in place.


alanking (approved):
Yep, looks identical to changes in #7481. # it.

JustinKyleJames:
> Yep, looks identical to changes in #7481. # it.

Done

Issue #7476, opened 2024-01-26
Voting for write should prefer good replicas


alanking:
- [x] main
- [x] 4-3-stable
- [x] irods_docs
---
Currently, hierarchy resolution for the `irods::WRITE_OPERATION` prefers stale replicas:
https://github.com/irods/irods/blob/29f4802edf1066fe10d6654289838743aa2e0803/server/core/src/voting.cpp#L214-L217

This was done this way to get around the limitation in the replication API that does not allow replicating over top of a good replica (see https://docs.irods.org/4.3.1/system_overview/data_objects/#replicate) and the way that interacts with certain situations in a replication resource hierarchy. In order for write operation voting to prefer good replicas, the rules of replication need to change in order to accommodate.

This implies a change in the linked documentation, so I'm going to put the documentation tag on this.

Maybe closely linked to https://github.com/irods/irods/issues/6940 as well...

trel:
The main use case for this change is appending to an existing data object.  Appending or truncating a stale replica could lead to data loss.

alanking:
Appending to objects will be tested as part of #7475, I imagine. Closing.

PR #7490, opened 2024-02-08
Change voting for write operations + associated replication fallout (main)

alanking:
Addresses #7476
Workaround for #6954 

Sibling PR: irods/irods_docs#241

---

Unit tests and core tests pass except for `test_delay_queue.Test_Delay_Queue`. Will investigate that, but it does not seem related to these changes.

Important changes occurred in `rsDataObjRepl.cpp` and I tried to leave explanatory comments/log messages where I could. The rest is mostly changes to the tests.

I kept the self-targeting commits separated in case we change our minds about that.

### DIFF
--- a/plugins/resources/replication/src/irods_create_write_replicator.cpp
+++ b/plugins/resources/replication/src/irods_create_write_replicator.cpp
@@ -57,8 +57,6 @@ namespace irods {
             copyKeyVal( ( keyValPair_t* )&object.cond_input(), &dataObjInp.condInput );
             addKeyVal( &dataObjInp.condInput, RESC_HIER_STR_KW, child_.c_str() );
             addKeyVal( &dataObjInp.condInput, DEST_RESC_HIER_STR_KW, hierarchy_string.c_str() );
-            addKeyVal( &dataObjInp.condInput, RESC_NAME_KW, root_resource_.c_str() );
-            addKeyVal( &dataObjInp.condInput, DEST_RESC_NAME_KW, root_resource_.c_str() );
             addKeyVal( &dataObjInp.condInput, IN_PDMO_KW, sub_hier.c_str() );
 
             try {
--- a/scripts/irods/test/test_irepl.py
+++ b/scripts/irods/test/test_irepl.py
@@ -824,15 +824,15 @@ class test_irepl_repl_status(session.make_sessions_mixin([('otherrods', 'rods')]
             {'start':{'a':'-', 'b':'X', 'c':'-'}, 'end':{'a':'X', 'b':'X', 'c':'-'}, 'output':{'out':None, 'err':None, 'rc':None}},
             {'start':{'a':'-', 'b':'X', 'c':'&'}, 'end':{'a':'&', 'b':'X', 'c':'&'}, 'output':{'out':None, 'err':None, 'rc':None}},
             {'start':{'a':'-', 'b':'X', 'c':'X'}, 'end':{'a':'X', 'b':'X', 'c':'X'}, 'output':{'out':None, 'err':None, 'rc':None}},
-            {'start':{'a':'&', 'b':'-', 'c':'-'}, 'end':{'a':'&', 'b':'-', 'c':'-'}, 'output':{'out':None, 'err':'SYS_NOT_ALLOWED', 'rc':None}},
-            {'start':{'a':'&', 'b':'-', 'c':'&'}, 'end':{'a':'&', 'b':'-', 'c':'&'}, 'output':{'out':None, 'err':'SYS_NOT_ALLOWED', 'rc':None}},
-            {'start':{'a':'&', 'b':'-', 'c':'X'}, 'end':{'a':'&', 'b':'-', 'c':'X'}, 'output':{'out':None, 'err':'SYS_NOT_ALLOWED', 'rc':None}},
-            {'start':{'a':'&', 'b':'&', 'c':'-'}, 'end':{'a':'&', 'b':'&', 'c':'-'}, 'output':{'out':None, 'err':'SYS_NOT_ALLOWED', 'rc':None}},
-            {'start':{'a':'&', 'b':'&', 'c':'&'}, 'end':{'a':'&', 'b':'&', 'c':'&'}, 'output':{'out':None, 'err':'SYS_NOT_ALLOWED', 'rc':None}},
-            {'start':{'a':'&', 'b':'&', 'c':'X'}, 'end':{'a':'&', 'b':'&', 'c':'X'}, 'output':{'out':None, 'err':'SYS_NOT_ALLOWED', 'rc':None}},
-            {'start':{'a':'&', 'b':'X', 'c':'-'}, 'end':{'a':'&', 'b':'X', 'c':'-'}, 'output':{'out':None, 'err':'SYS_NOT_ALLOWED', 'rc':None}},
-            {'start':{'a':'&', 'b':'X', 'c':'&'}, 'end':{'a':'&', 'b':'X', 'c':'&'}, 'output':{'out':None, 'err':'SYS_NOT_ALLOWED', 'rc':None}},
-            {'start':{'a':'&', 'b':'X', 'c':'X'}, 'end':{'a':'&', 'b':'X', 'c':'X'}, 'output':{'out':None, 'err':'SYS_NOT_ALLOWED', 'rc':None}},
+            {'start':{'a':'&', 'b':'-', 'c':'-'}, 'end':{'a':'&', 'b':'-', 'c':'-'}, 'output':{'out':None, 'err':None, 'rc':None}},
+            {'start':{'a':'&', 'b':'-', 'c':'&'}, 'end':{'a':'&', 'b':'-', 'c':'&'}, 'output':{'out':None, 'err':None, 'rc':None}},
+            {'start':{'a':'&', 'b':'-', 'c':'X'}, 'end':{'a':'&', 'b':'-', 'c':'X'}, 'output':{'out':None, 'err':None, 'rc':None}},
+            {'start':{'a':'&', 'b':'&', 'c':'-'}, 'end':{'a':'&', 'b':'&', 'c':'-'}, 'output':{'out':None, 'err':None, 'rc':None}},
+            {'start':{'a':'&', 'b':'&', 'c':'&'}, 'end':{'a':'&', 'b':'&', 'c':'&'}, 'output':{'out':None, 'err':None, 'rc':None}},
+            {'start':{'a':'&', 'b':'&', 'c':'X'}, 'end':{'a':'&', 'b':'&', 'c':'X'}, 'output':{'out':None, 'err':None, 'rc':None}},
+            {'start':{'a':'&', 'b':'X', 'c':'-'}, 'end':{'a':'&', 'b':'X', 'c':'-'}, 'output':{'out':None, 'err':None, 'rc':None}},
+            {'start':{'a':'&', 'b':'X', 'c':'&'}, 'end':{'a':'&', 'b':'X', 'c':'&'}, 'output':{'out':None, 'err':None, 'rc':None}},
+            {'start':{'a':'&', 'b':'X', 'c':'X'}, 'end':{'a':'&', 'b':'X', 'c':'X'}, 'output':{'out':None, 'err':None, 'rc':None}},
             {'start':{'a':'X', 'b':'-', 'c':'-'}, 'end':{'a':'X', 'b':'-', 'c':'-'}, 'output':{'out':None, 'err':'SYS_NOT_ALLOWED', 'rc':None}},
             {'start':{'a':'X', 'b':'-', 'c':'&'}, 'end':{'a':'&', 'b':'-', 'c':'&'}, 'output':{'out':None, 'err':None, 'rc':None}},
             {'start':{'a':'X', 'b':'-', 'c':'X'}, 'end':{'a':'X', 'b':'-', 'c':'X'}, 'output':{'out':None, 'err':'SYS_NOT_ALLOWED', 'rc':None}},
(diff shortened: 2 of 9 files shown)


alanking:
The test_delay_queue failure occurs when I build the commit used for 4.3.1 as well... in debug mode. When I run 4.3.1 in release mode, everything works as expected. I've created an issue here to look at it later: #7491. Meanwhile, I'm adding a commit to skip that test.

scripts/run_tests.py
```diff
@@ -202,7 +203,9 @@ def startTest(self, test):
     if options.run_plugin_tests:
         test_identifiers.extend(get_plugin_tests())
 
-    IrodsController().restart(test_mode=True)
+    IrodsController().stop()
+    irods.lib.clear_shm()
```
korydraughn, line 207: Have you seen a noticeable difference with this?

server/api/src/rsDataObjRepl.cpp
```diff
@@ -690,9 +692,11 @@ namespace
         auto source_cond_input = irods::experimental::make_key_value_proxy(source_inp.condInput);
         auto source_obj = resolve_hierarchy_and_get_data_object_info(_comm, source_inp, irods::OPEN_OPERATION);
 
+        const auto* source_hierarchy = source_cond_input.at(RESC_HIER_STR_KW).value().data();
```
korydraughn, line 695: Is the type of `.value()` a std::string or std::string_view?

If it's a std::string, prefer `.c_str()` instead of `.data()`.

server/api/src/rsDataObjRepl.cpp
```diff
@@ -728,31 +734,100 @@ namespace
         auto destination_cond_input = irods::experimental::make_key_value_proxy(destination_inp.condInput);
         auto destination_obj = resolve_hierarchy_and_get_data_object_info(_comm, destination_inp, irods::CREATE_OPERATION);
 
+        const auto* destination_hierarchy = destination_cond_input.at(DEST_RESC_HIER_STR_KW).value().data();
```
korydraughn, line 737: Is the type of `.value()` a std::string or std::string_view?

If it's a std::string, prefer `.c_str()` instead of `.data()`.

server/api/src/rsDataObjRepl.cpp
```diff
@@ -728,31 +734,100 @@ namespace
...
+            if (GOOD_REPLICA != source_replica.replica_status()) {
+                const std::string msg =
+                    fmt::format("Selected source replica in hierarchy [{}] is not good and will overwrite an "
+                                "existing replica. Replication is not allowed.",
+                                source_replica.resc_hier());
+
+                log_api::debug("[{}:{}] - [{}]", __func__, __LINE__, msg);
+
+                if (log_errors) {
+                    addRErrorMsg(&_comm.rError, SYS_NOT_ALLOWED, msg.data());
```
korydraughn, line 776: Prefer `.c_str()`.

server/api/src/rsDataObjRepl.cpp
```diff
@@ -839,17 +914,65 @@ namespace
...
+                // If the destination replica is good, it does not need to be updated. Skip it.
+                //
+                // TODO Should this trigger fileModified like regular replications do?
+                log_api::debug("[{}:{}] - Selected destination replica in hierarchy [{}] is already good. Skipping.",
+                               __func__,
+                               __LINE__,
+                               destination_replica.resc_hier());
+
+                continue;
+            }
```
korydraughn, line 957: So this little block is running inside a loop which iterates over all replicas.

Do you see a situation where invoking fileModified seems correct?

korydraughn (commented):
I haven't reviewed the test changes yet.

Will look at those soonish. Other than that, things look correct.

scripts/run_tests.py
alanking: Yes. I get far fewer failures with this change. Of course, I've left it in a separate commit so that it can be easily removed/not merged. I just found it useful and figured I'd put it out there.

plugins/resources/replication/src/irods_create_write_replicator.cpp
```diff
@@ -57,8 +57,6 @@ namespace irods {
             copyKeyVal( ( keyValPair_t* )&object.cond_input(), &dataObjInp.condInput );
             addKeyVal( &dataObjInp.condInput, RESC_HIER_STR_KW, child_.c_str() );
             addKeyVal( &dataObjInp.condInput, DEST_RESC_HIER_STR_KW, hierarchy_string.c_str() );
-            addKeyVal( &dataObjInp.condInput, RESC_NAME_KW, root_resource_.c_str() );
-            addKeyVal( &dataObjInp.condInput, DEST_RESC_NAME_KW, root_resource_.c_str() );
```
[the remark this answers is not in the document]
alanking: Right, these keywords are only useful for hierarchy resolution and since we already have resolved hierarchies, they won't serve any purpose.

server/api/src/rsDataObjRepl.cpp
alanking: `value()` is a `std::string_view`

server/api/src/rsDataObjRepl.cpp
alanking: `value()` is a `std::string_view`

server/api/src/rsDataObjRepl.cpp
alanking: Will do

alanking: Done

server/api/src/rsDataObjRepl.cpp
korydraughn: If the changes pass all of the tests, then I think we're fine. Until someone brings us a case we don't cover, I say let it ride.

scripts/run_tests.py
korydraughn: My gut reaction is great. Now I'm wondering if this would hide other issues.

It's not clear to me whether we should keep this in or not. Will think about it more.

trel: consider that we are hoping to get into k8s/autotesting soonish.   agree, still don't know if that's a vote for or against leaving this in.

server/api/src/rsDataObjRepl.cpp
trel: principle of tiny boxes says let the replication trigger fileModified... if for no other reason that it's easier for US to reason about?

scripts/run_tests.py
alanking: I'm pretty sure the problem will persist in k8s, so I don't think that should affect the decision.

The only issue I can think of that might be hidden is #6954 itself. If the issue occurs, it would manifest itself in (hopefully) only one test failing and this change would allow the next test to be executed to start fresh.

trel: i think i agree - that's the only known artifact at this time of what we **think** the problem is.

server/api/src/rsDataObjRepl.cpp
korydraughn: That should be okay as long as we aren't doing any string manipulation on the string_view and the string is null-terminated.

The null-terminated requirement is covered due to the implementation. I think we're good on the string manipulation too. I don't remember seeing the strings change in any way.

You agree?

server/api/src/rsDataObjRepl.cpp
korydraughn: See https://github.com/irods/irods/pull/7490#discussion_r1484421460

server/api/src/rsDataObjRepl.cpp
alanking: Correct, we do not change the hierarchy string once it has been resolved in this function, at least not in the input structure. The string comes from the hierarchy resolution which I believe is using `std::string` (and/or `irods::hierarchy_parser`, which uses `std::string`), so the string should be null terminated.

We can always make a copy if that seems safer.

server/api/src/rsDataObjRepl.cpp
korydraughn: > Because this is going to update all the replicas anyway, I guess it doesn't need to invoke fileModified for replication. It may be needed for something like compound where you might expect sync-to-archive to trigger, though. Or some other future use case that leans on that operation.

Can you elaborate on the compound resource thought?

plugins/resources/replication/src/irods_create_write_replicator.cpp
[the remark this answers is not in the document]
korydraughn: Cool cool.

server/api/src/rsDataObjRepl.cpp
alanking: We talked about this offline and have decided that it is more correct NOT to trigger fileModified in this case. `irepl -a` is surreptitiously circumventing policy - much like `iphymv` - and so any fileModified behavior will need to be achieved by a direct replication to the hierarchy or by invoking the desired operation through some other means. I will leave a comment to this effect and I created this issue for implementing proper testing around this in the future: https://github.com/irods/irods/issues/7493

server/api/src/rsDataObjRepl.cpp
korydraughn: If the string_view is not being modified (e.g. .substr is called), then we're good.

Feel free to mark as resolved.

scripts/run_tests.py
alanking: I left a comment and moved `clear_shm` into `run_tests.py` since it's only meant to be used there (for now) anyway.

korydraughn: Cool cool.

scripts/irods/test/test_irepl.py
korydraughn, line 1: I take it the test cases which are now commented out were commented out because they are duplicates, correct?

korydraughn (commented):
Looks like we're very close. Other than the review comments I just left, what remains for this PR?

scripts/irods/test/test_irepl.py
alanking: Correct, they don't really add anything.

scripts/irods/test/test_irepl.py
```diff
@@ -965,9 +965,9 @@ def test_irepl_Sb_foo(self):
...
             {'start':{'a':'&', 'b':'-', 'c':'X'}, 'end':{'a':'&', 'b':'-', 'c':'X'}, 'output':{'out':None, 'err':'SYS_REPLICA_INACCESSIBLE', 'rc':None}},
-            {'start':{'a':'&', 'b':'&', 'c':'-'}, 'end':{'a':'&', 'b':'&', 'c':'-'}, 'output':{'out':None, 'err':'SYS_NOT_ALLOWED', 'rc':None}},
-            {'start':{'a':'&', 'b':'&', 'c':'&'}, 'end':{'a':'&', 'b':'&', 'c':'&'}, 'output':{'out':None, 'err':'SYS_NOT_ALLOWED', 'rc':None}},
-            {'start':{'a':'&', 'b':'&', 'c':'X'}, 'end':{'a':'&', 'b':'&', 'c':'X'}, 'output':{'out':None, 'err':'SYS_NOT_ALLOWED', 'rc':None}},
+            {'start':{'a':'&', 'b':'&', 'c':'-'}, 'end':{'a':'&', 'b':'&', 'c':'-'}, 'output':{'out':None, 'err':None, 'rc':None}},
+            {'start':{'a':'&', 'b':'&', 'c':'&'}, 'end':{'a':'&', 'b':'&', 'c':'&'}, 'output':{'out':None, 'err':None, 'rc':None}},
+            {'start':{'a':'&', 'b':'&', 'c':'X'}, 'end':{'a':'&', 'b':'&', 'c':'X'}, 'output':{'out':None, 'err':None, 'rc':None}}, # TODO Stale replica won't win the vote. Feels wrong.
```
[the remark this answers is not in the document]
alanking: I think... probably this is okay. If the user wants to update the stale replica, they should either specify it or use `irepl -a`. Otherwise, it's going to select the default resource or, if that is not provided, it's left up to the vote and the user should not have any expectations. Thoughts?

scripts/run_tests.py
```diff
@@ -137,6 +138,15 @@ def startTest(self, test):
         print('{0} ... '.format(test.id()), end='', file=self.stream)
         unittest.TestResult.startTest(self, test)
 
+def clear_irods_shared_memory_files():
+    for shm_location in irods.paths.possible_shm_locations():
+        try:
+            for f in os.listdir(shm_location):
+                if 'irods' in f.lower():
+                    os.unlink(f)
```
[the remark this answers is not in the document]
alanking: I don't see why not. Could be signal to the person running the test that something is/was afoot

scripts/irods/test/test_irepl.py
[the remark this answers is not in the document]
korydraughn: That sounds fair to me. The important thing is that the user/admin has a way to deal with it.

server/api/src/rsDataObjRepl.cpp
```diff
@@ -839,17 +914,73 @@ namespace
...
+
+                // fileModified is NOT triggered in this case because all of the replicas are being updated. If it were
+                // to be triggered, new replicas could be created or existing replicas updated which are not accounted
+                // for in the irods::file_object over whose replicas we are currently iterating here. This could result
+                // in a replica getting wastefully or even erroneously overwritten (including the source replica!).
+
+                // TODO #7493 - Currently, the only resource type which may be affected by not triggering fileModified
+                // here would be the compound resource. We need to add a test for this but cannot add a real test
+                // because the unixfilesystem resource is the only storage resource available for testing. It needs to
+                // be configured for acting as a true archive.
```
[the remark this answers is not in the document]
alanking: I'll take a look and see whether I can get this working with the metadata/PEP solution proposed in #7493. Will report back here.

scripts/irods/test/test_irepl.py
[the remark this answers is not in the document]
alanking: Right, if `-R c` had been specified, the stale replica would be targeted for update. I'll remove the TODO comment.

scripts/run_tests.py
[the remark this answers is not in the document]
korydraughn: Good. Please proceed with adding a message.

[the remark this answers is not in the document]
alanking: Done

scripts/irods/test/test_irepl.py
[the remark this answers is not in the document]
alanking: Done

server/api/src/rsDataObjRepl.cpp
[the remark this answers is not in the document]
alanking: I've documented a (mostly) working POC for the metadata/PEP solution. I don't think we need to worry about that right now. I'll remove the TODO comment and leave a comment in #7493 about the test.

korydraughn (commented):
All of the review comments are resolved. Has this passed the test suite?

alanking:
All the tests passed last time I ran them, but I'm going to squash, build, and run it again to be sure.

korydraughn:
Sounds good.

alanking:
Pushed up the squashed commits. All tests passed.

korydraughn (commented):
The body of commit `[ 7476] Update tests for new replication rules` appears to have a grammar error around `a non-deterministic tests`?

Other than that, looks good.

alanking:
Ah, I think I meant "a few non-deterministic tests". I'll update it.

korydraughn (approved):
Pound it.

Issue #7498, opened 2024-02-13
Trim/`itrim`: Deprecate `AGE_KW`/`--age` support


alanking:
- [x] main (API)
- [x] main (iCommands)
- [x] 4-3-stable (API)
- [x] 4-3-stable (iCommands)

---

The trim API should be used as a precision tool for removing unneeded/unwanted replicas. The `AGE_KW`, which allows the client to specify only trimming replicas which are older than the specified number of *minutes*, is a level of decision making for which the API should not be responsible. If such behavior is desired, it can be achieved through queries, rules, and/or custom clients.

Issue #7468, opened 2024-01-19
itrim trimming where it shouldn't


luijs:
- [x] main
- [x] 4-3-stable
---
### iRODS Version, OS and Version
irods 4.3.1, ubuntu 20.04
Also seen in 4.3.0/ubuntu 20.04

### What did you try to do?
Trim a file that only had stale copies

### Expected behavior
The file would not be trimmed.
According to the docs here row 8: https://docs.irods.org/4.3.1/system_overview/data_objects/#trim
`8 | X | X | X | X | Data object has fewer than the minimum number of good replica`

If you have two stale replicas of a file and you do an itrim it should not delete any replica.

### Observed behavior (including steps to reproduce, if applicable)
The file was trimmed.

In more detail:
```
iput version.json
irepl -S hot_1 -R tape_1 version.json
iadmin modrepl logical_path /zone/home/user/version.json replica_number 1 DATA_REPL_STATUS 0
iadmin modrepl logical_path /zone/home/user/version.json replica_number 0 DATA_REPL_STATUS 0
ils -rl
/zone/home/user/:
  irods             0 hot_1          240 2024-01-19.14:25 X version.json
  irods             1 tape_1          240 2024-01-19.14:25 X version.json
  C- /zone/home/user/
/zone/home/user/:
itrim -N 1 version.json -S hot_1
Total size trimmed = 0.000 MB. Number of files trimmed = 1.
ils -rl
/zone/home/user/:
  irods             1 tape_1          240 2024-01-19.14:25 X version.json

```
I did have some rules in effect that were doing something in `pep_api_data_obj_trim_pre` and 'pep_api_data_obj_unlink_pre'. However, that should impact this, and if I uncommented those rules I still had the same behaviour. Curious to see if this is only my machine or if this can be replicated.

Note: When I put the replica_status of the file I wanted to keep to 2 or 3 the file was not trimmed as expected

trel:
If you specify a replica (with -n or -S), then I think this is expected behavior.

That table you reference is what happens when there is no particular replica targeted.   We can make that more explicit in the documentation.

luijs:
Hmm, is that really the expected behaviour? And is it also the desired behaviour? 
Why would this be different for repl_status 0 and repl_status 2, and why would it be different when you specify the exact replica to trim or not? Shouldn't blocking trimming be independent from if you specify a replica or not?

luijs:
[...]

PR #4886, opened 2020-04-27
[#4860] Deprecated minimum copies parameter for trim operation (4-2-stable)

korydraughn:

PR #7500, opened 2024-02-13
[#7468] Trim: prevent removing the last replica (main)

alanking:
Addresses #7468 
Addresses #7498
Addresses #7502
Addresses #7515

(sorta) Cherry-pick of #4886

---

Still need to run tests.


alanking:
Wasn't quite ready to open this, sorry. Converting to draft for the time being

alanking:
Core tests are all passing now. Marking ready for review.

trel (commented):
7468 commit message body

"and the new replicas" -> "and then new replicas"

server/api/src/rsDataObjTrim.cpp
```diff
@@ -54,42 +56,76 @@ namespace
...
+            THROW(SYS_INTERNAL_ERR, fmt::format("{}: Error occurred getting current time.", __func__));
+        }
+
         try {
-            const auto age_in_minutes = std::atoi(age_kw);
-            if ( age_in_minutes > 0 ) {
-                const auto now = time(0);
-                const auto age_in_seconds = age_in_minutes * 60;
-                return now - age_in_seconds;
+            const auto age_in_minutes = std::stol(age_kw);
```
korydraughn, line 71: Should we use `std::stoll` instead of `std::stol`?

server/api/src/rsDataObjTrim.cpp
```diff
@@ -54,42 +56,76 @@ namespace
...
+
+            constexpr decltype(age_in_minutes) seconds_in_one_minute = 60;
+            const auto age_in_seconds = age_in_minutes * seconds_in_one_minute;
+            if (0 != age_in_minutes && age_in_seconds / age_in_minutes != seconds_in_one_minute) {
+                // If the provided time was so large that representing it in seconds caused the value to wrap, just
+                // return 0 because the replicas would have to be older than epoch time 0 in order to be old enough to
+                // be trimmed. iRODS itself has not existed that long, so none of the replicas are old enough to be
+                // trimmed based on the input.
+                return 0;
+            }
```
korydraughn, line 85: Is the comment still true if the replicas were registered and existed before iRODS?

server/api/src/rsDataObjTrim.cpp
```diff
@@ -116,38 +152,42 @@ namespace
...
+        }
+
+        const auto good_replica_count = std::count_if(replica_list.begin(), replica_list.end(), [](const auto& repl) {
+            return GOOD_REPLICA == repl.replica_status();
+        });
+        constexpr decltype(good_replica_count) minimum_good_replica_count = 1;
+
+        const auto* age_kw = getValByKey(&_inp.condInput, AGE_KW);
+        const auto expiration_time = get_time_of_expiration(age_kw);
+        const auto expired = [&expiration_time](const irods::physical_object& _replica) {
```
korydraughn, line 175: Copying the expiration time into the lambda is likely better since it's a readonly value.

The compiler may be able to optimize the code more since it wouldn't need to deal with the indirection introduced by the reference to `expiration_time`.

server/api/src/rsDataObjTrim.cpp
```diff
@@ -116,38 +152,42 @@ namespace
...
-        if (repl_num) {
+        if (const char* repl_num = getValByKey(&_inp.condInput, REPL_NUM_KW); repl_num) {
             try {
                 const auto num = std::stoi(repl_num);
 
-                const auto repl = std::find_if(list.begin(), list.end(),
-                    [&num](const auto& repl) {
-                        return num == repl.repl_num();
-                    });
+                const auto repl = std::find_if(replica_list.begin(), replica_list.end(), [&num](const auto& repl) {
```
korydraughn, line 186: Same here. `&num` -> `num`

server/api/src/rsDataObjTrim.cpp
```diff
@@ -193,39 +232,65 @@ namespace
         }
 
         const char* resc_name = getValByKey(&_inp.condInput, RESC_NAME_KW);
-        const auto matches_target_resource = [&resc_name](const irods::physical_object& obj) {
-            return resc_name && irods::hierarchy_parser{obj.resc_hier()}.first_resc() == resc_name;
+        const auto matches_target_resource = [&resc_name](const irods::physical_object& _replica) {
```
korydraughn, line 235: `&resc_name` -> `resc_name`?

scripts/core_tests_list.json
```diff
@@ -91,7 +91,8 @@
     "test_iticket",
     "test_itouch",
     "test_itree",
-    "test_itrim",
+    "test_itrim.Test_Itrim",
+    "test_itrim.test_itrim_magic_with_four_replicas__issue_7468",
```
korydraughn, line 95: Can "magic" be expanded to better express what the test is about?

korydraughn (commented):
So far so good.

I'm still working through the PR. Will share more review comments a little later.

scripts/irods/test/test_resource_types.py
```diff
@@ -1717,7 +1722,9 @@ def test_iput_with_purgec(self):
 
         try:
             self.admin.assert_icommand_fail("ils -L " + filename, 'STDOUT_SINGLELINE', filename)  # should not be listed
-            self.admin.assert_icommand("iput --purgec " + filename)  # put file
+            # put file
+            self.admin.assert_icommand(
+                "iput --purgec " + filename, 'STDOUT', 'Specifying a minimum number of replicas to keep is deprecated.')
```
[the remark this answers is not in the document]
alanking: Yes, `--purgec` uses the trim API and provides the `COPIES_KW` to be able to trim down to 1 replica.

server/api/src/rsDataObjTrim.cpp
alanking: Yeah, I couldn't decide on a type here. Your suggestion works.

server/api/src/rsDataObjTrim.cpp
alanking: Can Linux systems handle times from before epoch time 0? If they can, and the iRODS catalog can, then we need to handle this.

If that is the case, we need to decide if such handling is worth the effort given that this feature is deprecated as of this PR.

server/api/src/rsDataObjTrim.cpp
alanking: Will update

server/api/src/rsDataObjTrim.cpp
alanking: Will update

server/api/src/rsDataObjTrim.cpp
alanking: Will update

server/api/src/rsDataObjTrim.cpp
trel: Some chaos here...
    https://stackoverflow.com/a/31709796

> For a signed 32-bit integer, the values range from 1901-12-13 to 2038-01-19 03:14:07 UTC

and then also...

> Add to this lack of specificity the fact that [a couple dozen other epochs](https://en.wikipedia.org/wiki/Epoch_(computing)#Notable_epoch_dates_in_computing) have been used by various software systems, some extremely popular and common. Examples include January 1, 1601 for NTFS file system & COBOL, January 1, 1980 for various FAT file systems, January 1, 2001 for Apple Cocoa, and January 0, 1900 for Excel & Lotus 1-2-3 spreadsheets.

And this crazy take...

> I recommend against tracking date-time as a count-from-epoch. Instead use specific data types where available in your programming language or database.

server/api/src/rsDataObjTrim.cpp
alanking: Done

server/api/src/rsDataObjTrim.cpp
alanking: Done

server/api/src/rsDataObjTrim.cpp
alanking: Done

server/api/src/rsDataObjTrim.cpp
korydraughn: I forgot the age option was being deprecated. It's fine as is.

scripts/core_tests_list.json
korydraughn: Sounds good to me.

alanking: Done

server/api/src/rsDataObjTrim.cpp
```diff
@@ -54,42 +56,76 @@ namespace
         return info;
     }
 
-    int get_time_of_expiration(const char* age_kw) {
+    auto get_time_of_expiration(const char* age_kw) -> std::uint64_t
```
korydraughn, line 59: Let's rename `age_kw` so it doesn't confuse the reader into thinking it's expected to be a keyword.

scripts/irods/test/test_itrim.py
```diff
@@ -78,11 +82,85 @@ def test_itrim_num_copies_repl_num_conflict__issue_3346(self):
...
+
+    def test_itrim_option_age_is_deprecated__issue_7498(self):
+        resc1 = 'issue_7498_1'
+        resc2 = 'issue_7498_2'
+        filename = 'test_itrim_option_age_is_deprecated__issue_7498'
+        logical_path = os.path.join(self.admin.session_collection, filename)
+
+        try:
+            lib.create_ufs_resource(self.admin, resc1, test.settings.HOSTNAME_2)
+            lib.create_ufs_resource(self.admin, resc2, test.settings.HOSTNAME_3)
```
korydraughn, line 135: The use of `test.settings.HOSTNAME_N` makes sense for topology, but in a zone with one server, are these variables set to resolve to the same host?

I'm looking at the following and wondering how the test succeeds.
https://github.com/irods/irods/blob/e6df49f58759d6609650b82ce4a907d82ee10881/scripts/run_tests.py#L192-L201

Is the testing environment setting those variables?

scripts/irods/test/test_itrim.py
```diff
@@ -218,3 +296,326 @@ def test_itrim_displays_incorrect_count__issue_3531(self):
...
+            self.admin.assert_icommand(['itrim', '-N1', '-n1', logical_path], 'STDOUT', 'Number of files trimmed = 0.')
+            self.assertTrue(lib.replica_exists_on_resource(self.admin, logical_path, resc2))
+
+        finally:
+            self.admin.assert_icommand(['ils', '-l', logical_path], 'STDOUT', filename) # debugging
+            self.admin.run_icommand(['irm', '-f', logical_path])
+            lib.remove_resource(self.admin, resc1)
+            lib.remove_resource(self.admin, resc2)
+
+    def test_itrim_minimum_replicas_to_keep_invalid_values(self):
```
korydraughn, line 356: Missing issue number.

scripts/irods/test/test_itrim.py
```diff
@@ -218,3 +296,326 @@ def test_itrim_displays_incorrect_count__issue_3531(self):
...
+            self.assertTrue(lib.replica_exists_on_resource(self.admin, logical_path, resc2))
+
+        finally:
+            self.admin.assert_icommand(['ils', '-l', logical_path], 'STDOUT', filename) # debugging
+            self.admin.run_icommand(['irm', '-f', logical_path])
+            lib.remove_resource(self.admin, resc1)
+            lib.remove_resource(self.admin, resc2)
+
+
+class test_itrim_target_replica_selection_decision_making(unittest.TestCase):
```
korydraughn, line 415: Missing issue number on class name?

korydraughn (commented):
The tests look really good.

korydraughn:
I think this PR will be ready once the final review comments are resolved.

scripts/irods/test/test_itrim.py
alanking: This if statement is only reached when `topology_test` is active:
https://github.com/irods/irods/blob/e6df49f58759d6609650b82ce4a907d82ee10881/scripts/run_tests.py#L189-L201

scripts/irods/test/test_itrim.py
alanking: I don't know that this is tied to a particular issue... It's more just for making sure default `itrim` behavior is covered by tests. I guess I started thinking about it more in - and was inspired to write the test as a result of - #7468? Thoughts?

scripts/irods/test/test_itrim.py
korydraughn: Makes sense. That implies the `HOSTNAME_N` variables are initialized to the same value earlier in the code.

Is that correct?

server/api/src/rsDataObjTrim.cpp
alanking: Done

scripts/irods/test/test_itrim.py
alanking: Done

scripts/irods/test/test_itrim.py
alanking: Oops, forgot the rest of the comment...

https://github.com/irods/irods/blob/e6df49f58759d6609650b82ce4a907d82ee10881/scripts/irods/test/settings.py#L6

The values of these hostnames are all the same at time of import as shown above.

scripts/irods/test/test_itrim.py
korydraughn: I get what you're saying, but having that issue number gives us a point of reference for when we have to revisit the tests and figure out why they were introduced. _Git history also achieves this if we're committing things in a reasonable fashion._

If iRODS was in pre-1.0 status, the issue number would be less important to me. But given the history of the repo, I feel it's helpful to include the issue number.

Willing to be persuaded otherwise.

Have we added tests recently which do not have issue numbers attached to them?

alanking: Right, I'm in agreement about the issue numbers. I just don't know that there *is* an issue number to associate here. #7468 is the closest thing I can come up with, and it's not strictly speaking related to all (if any) of these tests. Plus, I already wrote a test which directly addresses that issue in `test_itrim.Test_Itrim.test_itrim_does_not_remove_the_last_replica__issue_7468`.

Here's a recent-ish commit in which I introduced several tests with no issue number: 78c3dca6dfdc423cc8e091be35adda7a08fc1ffe

Again, I definitely think we should add issue numbers. Just looking for suggestions on an issue to link to in this case since it's not clear to me which issue would make sense here.

korydraughn: I see. You could create a new issue that captures the reasons for adding the tests. You could even link other related issues to the new issue so there's a link between them.

scripts/irods/test/test_itrim.py
korydraughn: Excellent. Thanks.

scripts/irods/test/test_itrim.py
alanking: Created https://github.com/irods/irods/issues/7515 and included the issue in the title.

alanking:
Oh, and I will update the commit title on "Add itrim tests: no complete removal, and magic" to link with #7515 as well as #7468. Or, I could break it in two to keep them separated.

korydraughn:
> Oh, and I will update the commit title on "Add itrim tests: no complete removal, and magic" to link with #7515 as well as #7468. Or, I could break it in two to keep them separated.

If splitting them into multiple commits makes it easier for readers to understand what's happening, then let's do that.

alanking:
I think I will leave them together. Doesn't add much value, and I can just add more words to the message body for clarification.

alanking:
Done, I think

korydraughn (approved):
Pound and merge when ready.

alanking:
Done. Mergin

Issue #7512, opened 2024-02-16
CMake: clean up remaining references to `CMAKE_SOURCE_DIR`


SwooshyCueb:
- [x] main
- [x] 4-3-stable

------

We still have a few references to `CMAKE_SOURCE_DIR`:
https://github.com/irods/irods/blob/e6df49f58759d6609650b82ce4a907d82ee10881/plugins/api/CMakeLists.txt#L70
https://github.com/irods/irods/blob/e6df49f58759d6609650b82ce4a907d82ee10881/plugins/auth/CMakeLists.txt#L44
https://github.com/irods/irods/blob/e6df49f58759d6609650b82ce4a907d82ee10881/plugins/auth_legacy/CMakeLists.txt#L45

Let's clean them up.

Issue #7511, opened 2024-02-16
get_delay_rule_info unit test cmake needs updating


SwooshyCueb:
- [x] main
- [x] 4-3-stable

------

The `get_delay_rule_info` unit test cmake file still contains references to `CMAKE_BINARY_DIR`, `CMAKE_SOURCE_DIR`, and `IRODS_EXTERNALS_FULLPATH_CATCH2`, and `IRODS_EXTERNALS_FULLPATH_JSON`:
https://github.com/irods/irods/blob/e6df49f58759d6609650b82ce4a907d82ee10881/unit_tests/cmake/test_config/irods_get_delay_rule_info.cmake#L6-L17

Let's give this some attention.

Issue #7510, opened 2024-02-16
process_stash unit test cmake needs updating


SwooshyCueb:
- [x] main
- [x] 4-3-stable

------

The `process_stash` unit test cmake file still contains references to `CMAKE_BINARY_DIR`, `CMAKE_SOURCE_DIR`, and `IRODS_EXTERNALS_FULLPATH_CATCH2`:
https://github.com/irods/irods/blob/e6df49f58759d6609650b82ce4a907d82ee10881/unit_tests/cmake/test_config/irods_process_stash.cmake#L6-L9

Let's give this some attention.

Issue #7509, opened 2024-02-16
CMake: do not call `target_include_directories` on `irods_server`


SwooshyCueb:
- [x] main
- [x] 4-3-stable

------

We make a `target_include_directories` call on `irods_server` here:
https://github.com/irods/irods/blob/e6df49f58759d6609650b82ce4a907d82ee10881/server/CMakeLists.txt#L73-L82

This call is not needed, as `irods_server` is an amalgamation of object libraries and has no source files.

PR #7514, opened 2024-02-16
[#6251,#6256,#7398,#7506,#7507,#7509,#7510,#7511,#7512] some CMake TLC (main)

SwooshyCueb:
In service of #6251
In service of #6256
Addresses #7398
Addresses #7506
Addresses #7507
Addresses #7509
Addresses #7510
Addresses #7511
Addresses #7512

I started preparing for spdlog's transition from static library to shared library, and figured I this was a good opportunity to move it over to `find_package` as well. That meant also moving fmt to `find_package`, and in the process of doing so I found a bunch of other things I could fix at the same time.

Not tested yet.


korydraughn:
Excellent. Happy to see things move towards `find_package`.

lib/core/CMakeLists.txt
```diff
@@ -79,6 +79,7 @@ target_link_libraries(
   PUBLIC
   nlohmann_json::nlohmann_json
   fmt::fmt
+  spdlog::spdlog_header_only # TODO: switch to spdlog::spdlog once shared lib is in externals
```
alanking, line 82: RE TODO: We have an issue for this already, I assume?

alanking (commented):
Nice. Looking more official by the day.

Should this commit message...
"[7508] CMake: do not call target_include_directories on irods_server"
...be linked to #7509 instead?

Also, the commit "[6256] CMake: use find_package for spdlog" has a few things that look like they belong to other commits, if that matters. e.g. Some more `CMAKE_SOURCE_DIR` and `fmt::fmt` cleanup efforts.

lib/core/CMakeLists.txt
SwooshyCueb: No, I'm just going to keep #6256 open

SwooshyCueb:
I'm going to have to re-think my approach to #7398 and #7505, as both would necessitate downstream projects calling `find_package(IRODS` *after* `project`. This is troublesome, as our CMake module that sets the compiler needs to be included *before* the `project` call.

SwooshyCueb:
> Should this commit message... "[7508] CMake: do not call target_include_directories on irods_server" ...be linked to #7509 instead?

Yup, nice catch.

> Also, the commit "[6256] CMake: use find_package for spdlog" has a few things that look like they belong to other commits, if that matters. e.g. Some more `CMAKE_SOURCE_DIR` and `fmt::fmt` cleanup efforts.

I'll clean that up too, nice catch.

SwooshyCueb:
I'm dropping #7505 from this PR, as we can't define targets in our config file until we have moved away from setting the compilers in our cmake files.

lib/core/CMakeLists.txt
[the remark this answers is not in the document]
SwooshyCueb: Covered by #6256

korydraughn (approved):
Pound it.

SwooshyCueb:
\#'d

alanking (approved):

Issue #7380, opened 2023-10-13
`iadmin rmuser` and `iadmin rmgroup` can function for each other


mstfdkmn:
- [x] main
- [x] 4-3-stable

---

### iRODS Version, OS and Version
4.3.0, almalinux8

### What did you try to do?
I accidentally called `iadmin rmuser` to remove a group.

### Expected behavior
Not to expect the group to be removed and I expected to see an informative error.

### Observed behavior (including steps to reproduce, if applicable)
The group was removed like a user and I didnt see any error. I see the same behavior for `iadmin rmgroup user`

trel:
groups *are* users in the system/database... but... this still feels like a bug.

FifthPotato:
> groups _are_ users in the system/database... but... this still feels like a bug.

Funny thing about that...

https://github.com/irods/irods_client_icommands/blob/d2c409184ff4d4a59aed41c7674a0d7c992bfefc/src/iadmin.cpp#L1626-L1630

iadmin actually does this on purpose. Indeed...

https://github.com/irods/irods/blob/17422abcce63ccd8ae0ed5a5fe711e15d0456658/server/api/src/rsGeneralAdmin.cpp#L1136

`rsGeneralAdmin` doesn't look like it even has a special block for removing groups.

Does this mean we should add an additional case to `rsGeneralAdmin`?

alanking:
Yes, I think we will need to modify `iadmin` as well as the GeneralAdmin API such that they are separated. It is possible that the database plugin will need to grow a new thing as well, but that remains to be seen, I guess.

korydraughn:
Agreed.

At the very least, `rsGeneralAdmin` should detect whether the entity is a user/group and respond accordingly.

@FifthPotato Please investigate which database PEP is invoked for this operation and what information is passed to it. That should help us determine if the database plugin needs a new operation.

trel:
without looking - i think i would expect the admin endpoint to detect a mismatch and say 'no' to the requesting client.

whether there should be an additional/different mechanism *in* the server feels like a different/separate question.

korydraughn:
That's fair.

alanking:
@FifthPotato  - Please close if complete. Thanks!

korydraughn:
Re-opening.

@FifthPotato Please update the user administration library. Two of the unit tests are failing. The junit output will tell you where the issue is (i.e. avoid passing `--discard-logs` to get the junit output).

PR #7533, opened 2024-02-27
Server-side code to only allow removal of appropriate entity (main)

FifthPotato:
For #7380.
One issue is that this breaks any client currently using the original `rmgroup`, since every client has sent it with the `user` arg1. Taking suggestions on how to resolve that.
Haven't run tests.

### DIFF
--- a/scripts/irods/test/test_iadmin.py
+++ b/scripts/irods/test/test_iadmin.py
@@ -485,6 +485,22 @@ class Test_Iadmin(resource_suite.ResourceBase, unittest.TestCase):
         self.admin.assert_icommand("iadmin mkuser lando invalid_user_type",
                                    'STDERR_SINGLELINE', "CAT_INVALID_USER_TYPE")  # should be rejected
 
+    @unittest.skipIf(IrodsConfig().version_tuple < (4, 3, 2), "Only applies to 4.3.2 and above.")
+    def test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380(self):
+        user_name = "test_user_mismatch_entity_issue_7380"
+        group_name = "test_group_mismatch_entity_issue_7380"
+        self.admin.assert_icommand("iadmin mkuser %s rodsuser" % user_name)
+        self.admin.assert_icommand("iadmin mkgroup %s" % group_name)
+
+        try:
+            self.admin.assert_icommand("iadmin rmuser %s" % group_name,
+                                       'STDERR_SINGLELINE', "rcGeneralAdmin failed with error -827000 CAT_INVALID_USER")
+            self.admin.assert_icommand("iadmin rmgroup %s" % user_name,
+                                       'STDERR_SINGLELINE', "rcGeneralAdmin failed with error -829000 CAT_INVALID_GROUP")
+        finally:
+            self.admin.assert_icommand("iadmin rmuser %s" % user_name)
+            self.admin.assert_icommand("iadmin rmgroup %s" % group_name)
+
     # =-=-=-=-=-=-=-
     # REBALANCE
     def test_rebalance_for_invalid_data__ticket_3147(self):
@@ -1437,7 +1453,7 @@ class Test_Iadmin(resource_suite.ResourceBase, unittest.TestCase):
         self.admin.assert_icommand(['iadmin', 'lrg'], 'STDERR_SINGLELINE', 'Resource groups are deprecated.')
 
     def test_group_membership(self):
-        self.admin.assert_icommand(["iadmin", "rmgroup", "g1"], 'STDERR_SINGLELINE', 'CAT_INVALID_USER')
+        self.admin.assert_icommand(["iadmin", "rmgroup", "g1"], 'STDERR_SINGLELINE', 'CAT_INVALID_GROUP')
         self.admin.assert_icommand(["iadmin", "mkgroup", "g1"])
         self.admin.assert_icommand(["iadmin", "atg", "g1", self.user0.username])
         self.admin.assert_icommand(["iadmin", "atg", "g1a", self.user0.username], 'STDERR_SINGLELINE', 'CAT_INVALID_GROUP')
--- a/server/api/src/rsGeneralAdmin.cpp
+++ b/server/api/src/rsGeneralAdmin.cpp
@@ -22,6 +22,10 @@
 #include "irods/irods_logger.hpp"
 #include "irods/user_validation_utilities.hpp"
 #include "irods/rs_set_delay_server_migration_info.hpp"
+#include "irods/version.hpp"
+
+#define IRODS_USER_ADMINISTRATION_ENABLE_SERVER_SIDE_API
+#include "irods/user_administration.hpp"
 
 #include <fmt/format.h>
 
@@ -1135,9 +1139,39 @@ _rsGeneralAdmin( rsComm_t *rsComm, generalAdminInp_t *generalAdminInp ) {
     }
     if ( strcmp( generalAdminInp->arg0, "rm" ) == 0 ) {
         if ( strcmp( generalAdminInp->arg1, "user" ) == 0 ) {
-            return irods::remove_user(*rsComm,
-                                      generalAdminInp->arg2 ? generalAdminInp->arg2 : "",
-                                      generalAdminInp->arg3 ? generalAdminInp->arg3 : "");
+            const auto* user_name = generalAdminInp->arg2 ? generalAdminInp->arg2 : "";
+            const auto* zone_name = generalAdminInp->arg3 ? generalAdminInp->arg3 : "";
+            namespace iua = irods::experimental::administration;
+            // new functionality only enforced on versions >= 4.3.2
+            const auto client_version = irods::to_version(rsComm->cliVersion.relVersion);
+            constexpr irods::version v4320{4, 3, 2};
+            if (client_version && client_version.value() >= v4320) {
+                const auto user_and_zone = irods::user::validate_name(user_name);
+                if (user_and_zone) {
+                    std::string derived_zone = std::get<1>(user_and_zone.value());
+                    if (derived_zone.empty()) {
+                        derived_zone = std::string{zone_name};
+                    }
+
+                    if (!iua::server::exists(*rsComm, iua::user{std::get<0>(user_and_zone.value()), derived_zone})) {
+                        log_api::error("{}: Cannot remove nonexistent user [{}#{}]", __func__, user_name, zone_name);
+                        return CAT_INVALID_USER;
+                    }
+                }
+            }
+            return irods::remove_user(*rsComm, user_name, zone_name);
+        }
+        if (strcmp(generalAdminInp->arg1, "group") == 0) {
+            const auto* group_name = generalAdminInp->arg2 ? generalAdminInp->arg2 : "";
+            const auto* zone_name = generalAdminInp->arg3 ? generalAdminInp->arg3 : "";
+            namespace iua = irods::experimental::administration;
+
+            if (!iua::server::exists(*rsComm, iua::group{group_name})) {
+                log_api::error("{}: Cannot remove nonexistent group [{}]", __func__, group_name);
+                return CAT_INVALID_GROUP;
+            }
+
+            return irods::remove_user(*rsComm, group_name, zone_name);
         }
         if ( strcmp( generalAdminInp->arg1, "dir" ) == 0 ) {
             memset( ( char* )&collInfo, 0, sizeof( collInfo ) );


server/api/src/rsGeneralAdmin.cpp
```diff
@@ -1135,9 +1138,29 @@ _rsGeneralAdmin( rsComm_t *rsComm, generalAdminInp_t *generalAdminInp ) {
     }
     if ( strcmp( generalAdminInp->arg0, "rm" ) == 0 ) {
         if ( strcmp( generalAdminInp->arg1, "user" ) == 0 ) {
+            const auto user_name = generalAdminInp->arg2 ? generalAdminInp->arg2 : "";
```
korydraughn, line 1141: When using `auto` to deduce a raw pointer, prefer `auto*`.

`auto` by itself is correct/acceptable, but it's not as clear as `auto*`.

_Please apply change to the other lines as well._

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -1135,9 +1138,29 @@ _rsGeneralAdmin( rsComm_t *rsComm, generalAdminInp_t *generalAdminInp ) {
...
+            const auto zone_name = generalAdminInp->arg3 ? generalAdminInp->arg3 : "";
+            namespace iua = irods::experimental::administration;
+
+            if(!iua::server::exists(*rsComm, iua::user { user_name, zone_name } )) {
+                return CAT_INVALID_USER;
```
korydraughn, line 1146: Let's log a message for the admin.

_This applies to the group logic below too._

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -1135,9 +1138,29 @@ _rsGeneralAdmin( rsComm_t *rsComm, generalAdminInp_t *generalAdminInp ) {
     }
     if ( strcmp( generalAdminInp->arg0, "rm" ) == 0 ) {
         if ( strcmp( generalAdminInp->arg1, "user" ) == 0 ) {
+            const auto user_name = generalAdminInp->arg2 ? generalAdminInp->arg2 : "";
+            const auto zone_name = generalAdminInp->arg3 ? generalAdminInp->arg3 : "";
```
korydraughn, line 1142: Should we check for empty strings or does the `iua::server::exists` call cover that?

korydraughn (commented):
The changes are mostly there. Just a few more things to consider/tweak and we can get this merged.

Needs a couple tests too. **test_iadmin.py** is the script that you'll want to add the tests to.

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -1135,9 +1138,29 @@ _rsGeneralAdmin( rsComm_t *rsComm, generalAdminInp_t *generalAdminInp ) {
...
+                                      user_name,
+                                      zone_name);
+        }
+        if ( strcmp( generalAdminInp->arg1, "group" ) == 0 ) {
+            const auto group_name = generalAdminInp->arg2 ? generalAdminInp->arg2 : "";
+            const auto zone_name = generalAdminInp->arg3 ? generalAdminInp->arg3 : "";
+            namespace iua = irods::experimental::administration;
+
+            if(!iua::server::exists(*rsComm, iua::group { group_name } )) {
+                return CAT_INVALID_USER;
```
korydraughn, line 1158: So we don't have a `CAT_INVALID_GROUP`?

Please add it if that's true.

korydraughn:
> For #7380. One issue is that this breaks any client currently using the original `rmgroup`, since every client has sent it with the `user` arg1. Taking suggestions on how to resolve that. Haven't run tests.

It's not clear to me what you mean by **_since every client has sent it with the `user` arg1_**. Please elaborate.

alanking:
I think he's saying that if any clients (e.g., `iadmin rmgroup` version 4.3.1) send "user" as arg1 for removing a group, as they have done historically, it will not work anymore. In other words, this is an interface change. Might affect something like PRC or ZMT, as well.

trel:
That seems... good/thorough.      Old things get old behavior.    New things get new, narrower behavior.   And we make sure that the things we ship categorize themselves as new things.

FifthPotato:
> > For #7380. One issue is that this breaks any client currently using the original `rmgroup`, since every client has sent it with the `user` arg1. Taking suggestions on how to resolve that. Haven't run tests.
> 
> It's not clear to me what you mean by **_since every client has sent it with the `user` arg1_**. Please elaborate.

Suppose a someone were using an old client (icommands or otherwise). If they call an `iadmin rmgroup ...` (or equivalent...), it sends the api call with `arg1 = "user"`, since that's how it works right now.

With this change merged, though, it seems like that would error. Seems like a breaking change...is that acceptable?

FifthPotato:
Added the version check/old behavior and made sure it worked (had to learn how to temporarily bump version too so I could test it).
Working on the rest of the review items now.

server/api/src/rsGeneralAdmin.cpp
FifthPotato: From what I can tell, `iua::server::exists` just tosses the input into a genquery, which I expect to be totally fine with empty strings. Calling it end-to-end with an empty string in my docker container seems to return the expected result as well.

FifthPotato:
In the process of writing the tests. I noticed while testing (the tests) that some of the existing tests expect CAT_INVALID_USER to be thrown when trying to remove a nonexistent group. (It throws CAT_INVALID_GROUP now) Do I...
1. Change all relevant unit tests to instead check for CAT_INVALID_GROUP
2. Make group removal throw CAT_INVALID_USER in the C++ code when client version is < 4.3.2
3. Do something else?

trel:
2 seems good and correct for backwards compatibility.

and

1 seems correct moving forward.

korydraughn:
Another thing to keep in mind - clients are not required to pass a version, yet. If a client doesn't include that piece of information, fallback to 4.3.1 behavior.

scripts/irods/test/test_iadmin.py
```diff
@@ -486,6 +486,19 @@ def test_iadmin_mkuser(self):
         self.admin.assert_icommand("iadmin mkuser lando invalid_user_type",
                                    'STDERR_SINGLELINE', "CAT_INVALID_USER_TYPE")  # should be rejected
 
+    @unittest.skipIf(IrodsConfig().version_tuple < (4, 3, 2), "Only applies to 4.3.2 and above.")
```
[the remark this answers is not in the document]
trel: can't hurt - keeps us honest about 'other' clients out there...

FifthPotato:
About the CAT_INVALID_GROUP/USER thing, my previous comment is wrong. So, as it stands...
1. None of the new tests will run until the version is bumped to 4.3.2. 
2. After 4.3.2, the test will fail until the icommands are updated to pass `group` instead of `user`. 
3. After that is updated, other tests will fail until they're updated to expect `CAT_INVALID_GROUP`.

Will the problems with 2 and 3 be issues opened on the icommands repo?

FifthPotato:
> need more tests for the 'bad' cases

Which bad cases are those?

trel:
I see logic for >4.3.2 for user, but not for group - don't we need to return the old `CAT_INVALID_USER` error to a pre432 or undeclaredversion client?

korydraughn:
> About the CAT_INVALID_GROUP/USER thing, my previous comment is wrong. So, as it stands...
> 
> 1. None of the new tests will run until the version is bumped to 4.3.2.
> 2. After 4.3.2, the test will fail until the icommands are updated to pass `group` instead of `user`.
> 3. After that is updated, other tests will fail until they're updated to expect `CAT_INVALID_GROUP`.
> 
> Will the problems with 2 and 3 be issues opened on the icommands repo?

No need to open new issues for icommands. The changes to icommands can be linked to the same issue since they're all related.

The main branch was recently bumped to 4.90.0. I'd imagine the tests should run after rebasing. Please confirm.

scripts/irods/test/test_iadmin.py
```diff
@@ -486,6 +486,20 @@ def test_iadmin_mkuser(self):
...
+    def test_new_clients_cannot_rmuser_groups_nor_rmgroup_users__issue_7380(self):
```
korydraughn, line 490: `new_clients` is a little misleading. Consider rephrasing.

Could be something like ... `test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380`.

scripts/irods/test/test_iadmin.py
```diff
@@ -486,6 +486,20 @@ def test_iadmin_mkuser(self):
...
+        self.admin.assert_icommand("iadmin mkuser testuser rodsuser")
+        self.admin.assert_icommand("iadmin mkgroup testgroup")
+
+        try:
+            self.admin.assert_icommand("iadmin rmuser testgroup",
+                                       'STDERR_SINGLELINE', "CAT_INVALID_USER")
+            self.admin.assert_icommand("iadmin rmgroup testuser",
+                                       'STDERR_SINGLELINE', "CAT_INVALID_GROUP")
```
korydraughn, line 498: Consider including more characters in the expected output string. As it is, if something like `OTHER_CAT_INVALID_USER_STUFF` was returned, the assertion for L495 would pass.

This helps make the test more resilient to future changes.

FifthPotato:
> > About the CAT_INVALID_GROUP/USER thing, my previous comment is wrong. So, as it stands...
> > 
> > 1. None of the new tests will run until the version is bumped to 4.3.2.
> > 2. After 4.3.2, the test will fail until the icommands are updated to pass `group` instead of `user`.
> > 3. After that is updated, other tests will fail until they're updated to expect `CAT_INVALID_GROUP`.
> > 
> > Will the problems with 2 and 3 be issues opened on the icommands repo?
> 
> No need to open new issues for icommands. The changes to icommands can be linked to the same issue since they're all related.
> 
> The main branch was recently bumped to 4.90.0. I'd imagine the tests should run after rebasing. Please confirm.

I rebased and got the tests to work after some fighting. As expected, they do run, and a handful of failures crop up because of these changes. Should I include fixes to that as part of this PR?

trel:
If merged changes are already in and have caused issues, outside of this PR... then a new PR should be created to clean up any existing problems.

If those merged changes are requiring updates to this PR to work as expected, then just make any changes to this PR to get this new code ready for merging.

korydraughn:
Right. Include a commit which focuses on updates to the tests which were affected by your change. For example:
- One commit to address the assigned issue _(should include one or more **new** tests)_
- Another commit to fix other existing tests affected by the previous commit

The commits would share the same issue number.

Don't forget to open a PR with the changes for icommands. Leave a comment in the icommands PR stating its relationship with this PR. If you need an example, look at one of @SwooshyCueb's PRs.

FifthPotato:
Force-pushed since I had to rebase for new main branch. Added some stuff to account for zone specification in username (e.g. username#zone syntax).

korydraughn:
We're so close. Go go go!

scripts/irods/test/test_iadmin.py
```diff
@@ -485,6 +485,20 @@ def test_iadmin_mkuser(self):
         self.admin.assert_icommand("iadmin mkuser lando invalid_user_type",
                                    'STDERR_SINGLELINE', "CAT_INVALID_USER_TYPE")  # should be rejected
 
+    @unittest.skipIf(IrodsConfig().version_tuple < (4, 3, 2), "Only applies to 4.3.2 and above.")
+    def test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380(self):
+        self.admin.assert_icommand("iadmin mkuser testuser rodsuser")
+        self.admin.assert_icommand("iadmin mkgroup testgroup")
```
korydraughn, line 491: `testuser` and `testgroup` are very generic. Consider including the issue number in the names so there's less of a chance for colliding with other tests.

You could store the test user/group names in variables and use those throughout the test too. That would make it easy to update the test names later if needed.

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -1135,9 +1139,43 @@ _rsGeneralAdmin( rsComm_t *rsComm, generalAdminInp_t *generalAdminInp ) {
...
-                                      generalAdminInp->arg2 ? generalAdminInp->arg2 : "",
-                                      generalAdminInp->arg3 ? generalAdminInp->arg3 : "");
+            const auto* user_name = generalAdminInp->arg2 ? generalAdminInp->arg2 : "";
+            const auto* zone_name = generalAdminInp->arg3 ? generalAdminInp->arg3 : "";
+            namespace iua = irods::experimental::administration;
+            const auto user_and_zone = irods::user::validate_name(user_name);
+            // new functionality only enforced on versions >= 4.3.2
+            const auto client_version = irods::to_version(rsComm->cliVersion.relVersion);
+            constexpr irods::version v4320{4, 3, 2};
+            if (user_and_zone != std::nullopt && client_version != std::nullopt && client_version >= v4320) {
```
korydraughn, line 1149: When checking if an optional contains a value, prefer `.has_value()` or treating it like a pointer (e.g. `if (user_and_zone && ...)`).

korydraughn, line 1149: Avoid implicit conversions when interacting with the contents of an optional.

Instead of using `client_version >= v4320`, prefer one of the following:
- `*client_version >= v4320`
- `client_version.value() >= v4320`

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -1135,9 +1139,43 @@ _rsGeneralAdmin( rsComm_t *rsComm, generalAdminInp_t *generalAdminInp ) {
...
+                std::string derived_zone = std::get<1>(user_and_zone.value());
+                bool user_exists;
```
korydraughn, line 1151: Missing initializer.

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -1135,9 +1139,43 @@ _rsGeneralAdmin( rsComm_t *rsComm, generalAdminInp_t *generalAdminInp ) {
...
+                if (!derived_zone.empty()) {
+                    user_exists =
+                        iua::server::exists(*rsComm, iua::user{std::get<0>(user_and_zone.value()), derived_zone});
+                }
+                else {
+                    user_exists =
+                        iua::server::exists(*rsComm, iua::user{std::get<0>(user_and_zone.value()), zone_name});
```
korydraughn, line 1158: Consider isolating the branches to modifying the zone argument. That is, move the call to `exists()` out of the if-ladder and update the `derived_zone` variable.

That removes the chance for the calls to `exists()` becoming out of sync.

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -1135,9 +1139,43 @@ _rsGeneralAdmin( rsComm_t *rsComm, generalAdminInp_t *generalAdminInp ) {
     }
     if ( strcmp( generalAdminInp->arg0, "rm" ) == 0 ) {
         if ( strcmp( generalAdminInp->arg1, "user" ) == 0 ) {
-            return irods::remove_user(*rsComm,
-                                      generalAdminInp->arg2 ? generalAdminInp->arg2 : "",
-                                      generalAdminInp->arg3 ? generalAdminInp->arg3 : "");
+            const auto* user_name = generalAdminInp->arg2 ? generalAdminInp->arg2 : "";
+            const auto* zone_name = generalAdminInp->arg3 ? generalAdminInp->arg3 : "";
+            namespace iua = irods::experimental::administration;
+            const auto user_and_zone = irods::user::validate_name(user_name);
```
korydraughn, line 1145: What happens if `user_name` is invalid?
Does it throw or return an empty optional?
Should we care?

korydraughn: That sounds correct.

The call to `validate_name` should only happen if we're dealing with a 4.3.2 server. You agree?

FifthPotato: Good point. I'll add that.

korydraughn:
What's the status of this PR?
Are the tests passing?

Needs more clang-format too.

FifthPotato:
Tests are running! Two seem to have failed, but might be spurious, so rerunning right now.
For what it's worth, the iadmin tests run fine.

FifthPotato:
All tests did eventually pass, despite taking around 4 tries. Working on implementing server sniffing for https://github.com/irods/irods_client_icommands/pull/447 now.

korydraughn:
Did you capture the tests that failed?
Are the failures different across runs?
Do you feel they are related to your change?

FifthPotato:
I didn't record which ones failed, but a lot of them seemed pretty irrelevant, and went away with a few more re-runs. Should I rerun from the start just to be sure?

alanking:
I don't think you need to run them again. It would just be good to note which ones fail in the future (particularly if they are failing repeatedly) so that we can put some more eyeballs on those tests. Just my two cents.

korydraughn:
Guaranteed there are clients which do not send the correct version. Jargon always sends a value identifying itself as iRODS v3.

`to_version` returns an optional, which the code checks. Unless I'm misunderstanding, I think the server is safe.

korydraughn:

FifthPotato:
Unless someone else objects, don't think so!

korydraughn:
Seems good to me. Please squash to taste.

FifthPotato:
Squished

trel:
it's so clean!

korydraughn (approved):
Pound it!

FifthPotato:
Octothorpe'd

alanking:

Issue #7516, opened 2024-02-19
Remove "recursive pre-scan" text from output in several iCommands with recursive options


alanking:
- [ ] main
- [ ] 4-3-stable
---
```bash
$ iput -r log
Running recursive pre-scan... pre-scan complete... transferring data...
```
This text should only be shown as verbose output at best, but I think we should remove it entirely.

Unsure about the milestone.

trel:
agree with the sentiment - verbose feels okay/fine.   it could be useful/helpful when it's a deep directory.

can happen anytime in 4.3 i think

alanking:
The text is now available for viewing with the "veryVerbose" option (-V).

Issue #7242, opened 2023-08-01
Address Python test suite inefficiencies


SwooshyCueb:
We have several test suites with expensive `setUp()`/`tearDown()` methods, inherited or otherwise, that could be trimmed down and/or partially or completely changed to `setUpClass()`/`tearDownClass()` methods. This slows down testing quite significantly.

Given the size of our collection of test suites, and the relatively low priority of this effort, the likelihood that this can be started and completed in a single milestone is unlikely. Therefore, I am putting it in the "ongoing" milestone.

Issue #5359, opened 2021-01-21
irsync does not honor "ignore symlinks flag --link", when symlink is broken


vitlacil:
- [x] main
- [x] 4-3-stable
---
## Bug Report

Expected behavior:  To ignore all symlinks, broken or not. 
Observed behavior  When `irsync` reaches a broken symlink, it fails to ignore it and aborts with error.

For example:
**in absence of broken symlink irsync executes as expected:**

```
[vitlacil@t04n27 symlinks]$ ls -lah mydir/
  total 857K
  drwxrwxr-x 2 vitlacil vitlacil 2.0K Jan 21 13:49 .
  drwxrwxr-x 3 vitlacil vitlacil 2.0K Jan 21 13:49 ..
  lrwxr-xr-x 1 vitlacil vitlacil   59 Jan 20 14:21 inTSM_1.txt.copy -> /afs/pdc.kth.se/home/v/vitlacil/Public/HPR/inTSM_1.txt.copy
  -rw-rw-r-- 1 vitlacil vitlacil 851K Jan 20 14:08 inTSM_3.txt
  lrwxr-xr-x 1 vitlacil vitlacil   54 Jan 20 14:08 inTSM_4.txt -> /afs/pdc.kth.se/home/v/vitlacil/Public/HPR/inTSM_4.txt

[vitlacil@t04n27 symlinks]$ ls -lh /afs/pdc.kth.se/home/v/vitlacil/Public/HPR/inTSM_1.txt.copy
  -rw-rw-r-- 1 vitlacil vitlacil 0 Jan 20 16:25 /afs/pdc.kth.se/home/v/vitlacil/Public/HPR/inTSM_1.txt.copy

[vitlacil@t04n27 symlinks]$ ls -lh /afs/pdc.kth.se/home/v/vitlacil/Public/HPR/inTSM_4.txt
  -rw-rw-rw- 1 vitlacil games 66K Oct 26  2010 /afs/pdc.kth.se/home/v/vitlacil/Public/HPR/inTSM_4.txt

[vitlacil@t04n27 symlinks]$  irsync -vlr --link mydir i:/snic.se/home/s_dejvi/test/symlinks
Running recursive pre-scan... pre-scan complete... transferring data...
C- /snic.se/home/s_dejvi/test/symlinks:
mydir/inTSM_3.txt   871134   N
   inTSM_3.txt                     0.831 MB | 0.000 sec | 0 thr | 16615.562 MB/s
[...]

PR #7534, opened 2024-02-27
Fix handling of `--link` in `iput` and `irsync`, remove "pre-flight" output (main)

alanking:
Addresses #5359
In service of #7242 
Addresses #7516 

Companion PR: https://github.com/irods/irods_client_icommands/pull/443

---

Most of the changes in this PR have to do with tests. The C++ changes are only in the client library and *should* only affect `iput` and `irsync` as they are the only ones using `--link`.

I have seen all the core tests passed through much toil and many tears, but I will run them again once more to make sure.


scripts/irods/test/test_icommands_file_operations.py
korydraughn, line 1: The first statement of the commit message for 7242 appears to have a grammatical error.

korydraughn, line 1: For commit 7242, it looks like some tests were moved, but it's clear to me whether some of the contents of all/some tests changed as well.

Are there any specific lines we should pay attention to?

scripts/irods/test/test_symlink_operations.py
```diff
@@ -246,24 +246,28 @@ def test_irsync_iput_recursive_dangling_symlink_issues_3072_3988(self):
...
+                    else:
+                        self.assertIn(recurse_fail_string, stdout,
+                                      '{0}: Expected stdout: "...{1}...", got: "{2}"'.format(cmd, recurse_fail_string, stdout))
 
-                # Make sure the single valid file under dir1
-                # was not moved into the vault
-                self.user0.assert_icommand('ils -l {target_collection_path}/newname'.format(**locals()),
-                            'STDERR_SINGLELINE',
-                        '{target_collection_path}/newname does not exist or user lacks access permission'.format(**locals())
-                )
```
alanking, line 267: If you're wondering why this assertion was changed, please see the following history...

The test was introduced in this commit and originally asserted that the dangling symlink does not appear in iRODS:
https://github.com/irods/irods/commit/bb4cb508e822794e3aec8eb43140cea3c829f96c

The test was changed in this commit to assert that the valid file would not be uploaded:
https://github.com/irods/irods/commit/e37c529f2beabc66cbda529bd2e06f55f5537008

I'm not really sure how this was passing before because from my changes I do not see how this is preventing the "pre-flight" scan from checking for the dangling symlink. What we have now is *partial success* because the dangling symlink just fails to upload later. We can discuss whether this is the expected behavior and I can look into threading that particular needle.

scripts/irods/test/test_iput.py
```diff
@@ -613,6 +613,215 @@ def test_recursive_bulk_overwrite_is_not_allowed__issue_7110(self):
...
+
+        # These are the normal files and directories. These will be added so that SOMETHING is in iRODS at the end.
+        existent_file_path = os.path.join(put_dir, 'this_file_will_exist')
+        existent_subdir_path = os.path.join(put_dir, 'this_dir_will_exist')
+
+        # These are the data objects and collections which should exist for the normal files and directories.
+        existent_data_object_path = os.path.join(collection_path, os.path.basename(existent_file_path))
+        existent_subcollection_path = os.path.join(collection_path, os.path.basename(existent_subdir_path))
+
+        file_size = 1024 # I just made this up... it has no particular significance
```
korydraughn, line 663: Consider rephrasing to not use "I".

trel:
first commit message
"to a several tests" -> "to several tests"

scripts/irods/test/test_iput.py
```diff
@@ -613,6 +613,215 @@ def test_recursive_bulk_overwrite_is_not_allowed__issue_7110(self):
...
+
+        # These are the normal files and directories.
+        existent_file_path = os.path.join(put_dir, 'this_file_will_exist')
+        existent_subdir_path = os.path.join(put_dir, 'this_dir_will_exist')
+
+        # These are the data objects and collections which should exist for the normal files and directories.
+        existent_data_object_path = os.path.join(collection_path, os.path.basename(existent_file_path))
+        existent_subcollection_path = os.path.join(collection_path, os.path.basename(existent_subdir_path))
+
+        file_size = 1024 # I just made this up... it has no particular significance
```
korydraughn, line 780: Consider rephrasing to not use "I".

scripts/irods/test/test_iput.py
```diff
@@ -613,6 +613,215 @@ def test_recursive_bulk_overwrite_is_not_allowed__issue_7110(self):
...
+
+            # Create files and symlinks to the files.
+            lib.make_file(good_symlink_target_file_path, file_size)
+            self.assertTrue(os.path.exists(good_symlink_target_file_path))
+            lib.make_file(existent_file_path, file_size)
+            self.assertTrue(os.path.exists(existent_file_path))
+            os.symlink(good_symlink_target_file_path, good_symlink_file_path)
+            self.assertTrue(os.path.exists(good_symlink_file_path))
+
+            # Run irsync between the local directory and the collection (which is basically an iput). Use the --link
```
korydraughn, line 803: irsync -> iput?

lib/core/src/irods_path_recursion.cpp
```diff
@@ -323,25 +321,16 @@ irods::file_system_sanity_check( irods::recursion_map_t& pathmap,
             return USER_INPUT_OPTION_ERR;
         }
 
-        if (DISPLAY_PREFLIGHT) { std::cout << "Running recursive pre-scan... " << std::flush; }
```
alanking, line 326: Oh, right, we had discussed making this appear when "verbose" is in use. I can add this if desired.

scripts/irods/test/test_iput.py
```diff
@@ -613,6 +613,215 @@ def test_recursive_bulk_overwrite_is_not_allowed__issue_7110(self):
...
+
+            # Run iput to upload the directory. Use the --link option to instruct that symlinks are to be ignored,
+            # broken or not.
+            self.user.assert_icommand(['iput', '--link', '-r', put_dir, collection_path])
+
+            # Confirm that the main collection was created - no funny business there.
+            self.assertTrue(lib.collection_exists(self.user, collection_path))
+
+            # Confirm that a subcollection was created for the existent subcollection, but that no subcollection was
+            # created for the symlink or the collection to which it points because irsync was supposed to ignore
```
korydraughn, line 712: irsync -> iput?

scripts/irods/test/test_iput.py
```diff
@@ -613,6 +613,215 @@ def test_recursive_bulk_overwrite_is_not_allowed__issue_7110(self):
...
+            # symlinks.
+            self.assertTrue(lib.collection_exists(self.user, existent_subcollection_path))
+            self.assertFalse(lib.collection_exists(self.user, good_symlink_collection_path))
+            self.assertFalse(lib.collection_exists(self.user, good_symlink_target_dir_subcollection_path))
+            self.assertFalse(lib.replica_exists(self.user, good_symlink_collection_path, 0))
+            self.assertFalse(lib.replica_exists(self.user, good_symlink_target_dir_subcollection_path, 0))
+
+            # Confirm that a subcollection was not created for the nonexistent subdirectory, and that no subcollection
+            # was created for the symlink because irsync was supposed to ignore symlinks. Also make sure a data object
```
korydraughn, line 721: irsync -> iput?

scripts/irods/test/test_iput.py
```diff
@@ -613,6 +613,215 @@ def test_recursive_bulk_overwrite_is_not_allowed__issue_7110(self):
...
+            # was not created.
+            self.assertFalse(lib.collection_exists(self.user, nonexistent_subcollection_path))
+            self.assertFalse(lib.collection_exists(self.user, broken_symlink_collection_path))
+            self.assertFalse(lib.replica_exists(self.user, broken_symlink_collection_path, 0))
+
+            # Confirm that a data object was not created for the nonexistent file, and that no data object was created
+            # for the symlink because irsync was supposed to ignore symlinks.
```
korydraughn, line 728: irsync -> iput?

scripts/irods/test/test_iput.py
```diff
@@ -613,6 +613,215 @@ def test_recursive_bulk_overwrite_is_not_allowed__issue_7110(self):
...
+            self.assertFalse(lib.replica_exists(self.user, nonexistent_data_object_path, 0))
+            self.assertFalse(lib.replica_exists(self.user, broken_symlink_data_object_path, 0))
+
+            # Confirm that a data object was created for the existent file, but that no data object was created for the
+            # symlink because irsync was supposed to ignore symlinks.
```
korydraughn, line 733: irsync -> iput?

lib/core/src/irods_path_recursion.cpp
trel: i think that makes sense

scripts/irods/test/test_iput.py
```diff
@@ -613,6 +613,215 @@ def test_recursive_bulk_overwrite_is_not_allowed__issue_7110(self):
...
+            # Create files and symlinks to the files.
+            lib.make_file(good_symlink_target_file_path, file_size)
+            self.assertTrue(os.path.exists(good_symlink_target_file_path))
+            lib.make_file(existent_file_path, file_size)
+            self.assertTrue(os.path.exists(existent_file_path))
+            os.symlink(good_symlink_target_file_path, good_symlink_file_path)
+            self.assertTrue(os.path.exists(good_symlink_file_path))
+
+            # Run irsync between the local directory and the collection (which is basically an iput). Use the --link
+            # option to instruct that symlinks are to be ignored, broken or not.
```
alanking, line 804:
```suggestion
            # Run iput to upload the directory.
```

scripts/irods/test/test_irsync.py
```diff
@@ -439,3 +439,211 @@ def test_irsync_r_symlink(self):
...
+            # Create files and symlinks to the files.
+            lib.make_file(good_symlink_target_file_path, file_size)
+            self.assertTrue(os.path.exists(good_symlink_target_file_path))
+            lib.make_file(existent_file_path, file_size)
+            self.assertTrue(os.path.exists(existent_file_path))
+            os.symlink(good_symlink_target_file_path, good_symlink_file_path)
+            self.assertTrue(os.path.exists(good_symlink_file_path))
+
+            # Run irsync between the local directory and the collection (which is basically an iput). Use the --link
+            # option to instruct that symlinks are to be ignored, broken or not.
```
alanking, line 629:
```suggestion
           # Run irsync between the local directory and the collection (which is basically an iput).
```

scripts/irods/test/test_symlink_operations.py
```diff
@@ -246,24 +246,28 @@ def test_irsync_iput_recursive_dangling_symlink_issues_3072_3988(self):
...
 
-                cmd = teststring.format(**locals())
-                stdout,stderr,_ = self.user0.run_icommand(cmd)
+                    cmd = teststring.format(**locals())
+                    stdout,stderr,_ = self.user0.run_icommand(cmd)
 
-                self.assertIn(recurse_fail_string, stdout,
-                              '{0}: Expected stdout: "...{1}...", got: "{2}"'.format(cmd, recurse_fail_string, stdout))
+                    if ignore_symlinks:
+                        self.assertEqual(str(), stdout, '{}: Expected stdout: "", got: "{}"'.format(cmd, stdout))
```
korydraughn, line 256: Why is `str()` used instead of `""`? Is it triggering some conversion or something?

korydraughn (commented):
Seems good to me overall.

scripts/irods/test/test_iput.py
alanking: Done

scripts/irods/test/test_irsync.py
alanking: Done

scripts/irods/test/test_icommands_file_operations.py
alanking: Done. That's why I force-pushed, by the way. All code changes remain in separate commits.

scripts/irods/test/test_iput.py
trel: just 'No particular significance'

scripts/irods/test/test_icommands_file_operations.py
alanking: The contents of the moved tests did not change. The contents of the consolidated tests changed in that some extra assertions were carried over from `test_icommands_file_operations` to `test_irsync` but are only executed when not running the tests in topology because files in the vault are checked.

scripts/irods/test/test_iput.py
alanking: Done

scripts/irods/test/test_iput.py
alanking: Done

scripts/irods/test/test_iput.py
alanking: Done

scripts/irods/test/test_iput.py
alanking: Done

scripts/irods/test/test_iput.py
alanking: Done

scripts/irods/test/test_iput.py
alanking: Done

scripts/irods/test/test_iput.py
alanking: Done

scripts/irods/test/test_symlink_operations.py
alanking: No real reason. I have updated all of these to just use `''`'.

scripts/irods/test/test_icommands_file_operations.py
korydraughn: Cool cool.

lib/core/src/irods_path_recursion.cpp
alanking: Done

scripts/irods/test/test_symlink_operations.py
alanking: After some discussion we have determined that this assertion should remain for the `ignore_symlinks == False` case. Update to follow...

alanking: As it turns out, this is not a problem with the new changes but with the test. In the past, this test never uploaded anything due to the incorrect handling of `--link` and so never needed to clean up after itself between sub-tests. Now that `--link` is being handled correctly, tests which use `--link` will upload things and the test needs to clean up the test collection between runs.

alanking: Done

trel:
how do you feel about possibly making a new option named better? ...
`--ignore-symlinks` as an alias for `--link`, and then we stop advertising `--link`...

i suppose that's a new issue...

alanking:
> how do you feel about possibly making a new option named better? ... `--ignore-symlinks` as an alias for `--link`, and then we stop advertising `--link`...
> 
> i suppose that's a new issue...

I feel that it is possibly not worth the effort. If we offer equivalent functionality in the iRODS CLI, we can name the option better there. I will, however, update the option name for `iput` and `irsync` and deprecate the old one if instructed to do so. :)

trel:
let's think on it - would be a separate issue.   so carry on.

alanking:
Okay, I created https://github.com/irods/irods/issues/7537 for that effort.

korydraughn (commented):
Looks like everything is in order.

Please squash to taste if you're happy with it.

alanking:
Squashed. I combined the commit for adding new tests and updating the old since I didn't see much value in keeping them separate at this point.

trel (approved):
good good.

Issue #7554, opened 2024-03-07
deprecate GeneralUpdate API endpoint


trel:
- [x] main
- [x] 4-3-stable
---
Before removing GeneralUpdate in #2956 - we need to declare it deprecated.

Need discussion/investigation about what will get removed and if there is any calling code within the server.

alanking:
> NTS: Investigate deprecating `rxGeneralRowInsert` and `rxGeneralRowPurge` as well.


Do we have an issue or issues for this?

korydraughn:
We don't. Will create. Thanks for the reminder.

PR #7568, opened 2024-03-14
[#7554] Deprecate `rxGeneralUpdate` API endpoint (main)

korydraughn:
Not yet tested.

Depending on how things go, I may expand this to `rxGeneralRowInsert` and `rxGeneralRowPurge`.

### DIFF
--- a/lib/api/include/irods/generalUpdate.h
+++ b/lib/api/include/irods/generalUpdate.h
@@ -1,12 +1,18 @@
-#ifndef GENERAL_UPDATE_H__
-#define GENERAL_UPDATE_H__
+#ifndef IRODS_GENERAL_UPDATE_H
+#define IRODS_GENERAL_UPDATE_H
 
 #include "irods/rcConnect.h"
 #include "irods/rodsGeneralUpdate.h"
 
 #ifdef __cplusplus
-extern "C"
+extern "C" {
 #endif
-int rcGeneralUpdate( rcComm_t *conn, generalUpdateInp_t *generalUpdateInp );
 
+int rcGeneralUpdate(rcComm_t* conn, generalUpdateInp_t* generalUpdateInp)
+    __attribute__((deprecated("GeneralUpdate is deprecated. Its use should be avoided.")));
+
+#ifdef __cplusplus
+} // extern "C"
 #endif
+
+#endif // IRODS_GENERAL_UPDATE_H
--- a/lib/core/include/irods/rodsGeneralUpdate.h
+++ b/lib/core/include/irods/rodsGeneralUpdate.h
@@ -24,10 +24,13 @@
 #define UPDATE_NOW_TIME "update_now_time"
 
 #define GeneralUpdateInp_PI "int type; struct InxValPair_PI;"
-typedef struct GeneralUpdateInp {
+
+// NOLINTNEXTLINE(modernize-use-using)
+__attribute__((deprecated("GeneralUpdate is deprecated. Its use should be avoided."))) typedef struct __attribute__((
+    deprecated("GeneralUpdate is deprecated. Its use should be avoided."))) GeneralUpdateInp
+{
     int type;  /* GEN_UPDATE_INSERT or DELETE */
     inxValPair_t values;  /* Column IDs (from rodsGenQuery.h) and values */
 } generalUpdateInp_t;
-
 
 #endif	// RODS_GENERAL_UPDATE_H__
--- a/server/api/include/irods/rsGeneralUpdate.hpp
+++ b/server/api/include/irods/rsGeneralUpdate.hpp
@@ -1,10 +1,15 @@
-#ifndef RS_GENERAL_UPDATE_HPP
-#define RS_GENERAL_UPDATE_HPP
+#ifndef IRODS_RS_GENERAL_UPDATE_HPP
+#define IRODS_RS_GENERAL_UPDATE_HPP
 
 #include "irods/rcConnect.h"
 #include "irods/rodsGeneralUpdate.h"
 
-int rsGeneralUpdate( rsComm_t *rsComm, generalUpdateInp_t *generalUpdateInp );
-int _rsGeneralUpdate( generalUpdateInp_t *generalUpdateInp );
+int rsGeneralUpdate(rsComm_t* rsComm, generalUpdateInp_t* generalUpdateInp)
+    __attribute__((deprecated("GeneralUpdate is deprecated. Its use should be avoided.")));
 
-#endif
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+int _rsGeneralUpdate( generalUpdateInp_t *generalUpdateInp );
+#pragma clang diagnostic pop
+
+#endif // IRODS_RS_GENERAL_UPDATE_HPP
--- a/server/api/src/rsGeneralUpdate.cpp
+++ b/server/api/src/rsGeneralUpdate.cpp
@@ -54,11 +54,10 @@ rsGeneralUpdate( rsComm_t *rsComm, generalUpdateInp_t *generalUpdateInp ) {
     return status;
 }
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
 int
 _rsGeneralUpdate( generalUpdateInp_t *generalUpdateInp ) {
-    int status;
-
-    status  = chlGeneralUpdate( *generalUpdateInp );
-
-    return status;
+    return chlGeneralUpdate(*generalUpdateInp);
 }
+#pragma clang diagnostic pop
--- a/lib/api/include/irods/apiNumberData.h
+++ b/lib/api/include/irods/apiNumberData.h
@@ -129,6 +129,7 @@ API_NUMBER(MOD_AVU_METADATA_AN,                     706)
 API_NUMBER(MOD_ACCESS_CONTROL_AN,                   707)
 API_NUMBER(RULE_EXEC_MOD_AN,                        708)
 API_NUMBER(GET_TEMP_PASSWORD_AN,                    709)
+// DEPRECATED: GeneralUpdate is deprecated. Its use should be avoided.
 API_NUMBER(GENERAL_UPDATE_AN,                       710)
 API_NUMBER(READ_COLLECTION_AN,                      713)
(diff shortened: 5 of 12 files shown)


server/api/include/irods/rsGeneralUpdate.hpp
```diff
@@ -1,10 +1,14 @@
...
 
 #include "irods/rcConnect.h"
 #include "irods/rodsGeneralUpdate.h"
 
+[[deprecated("rsGeneralUpdate is deprecated and will be removed in a later release. "
+             "If you are reaching for this, consider revising your approach.")]]
 int rsGeneralUpdate( rsComm_t *rsComm, generalUpdateInp_t *generalUpdateInp );
+
+// TODO Leading underscores are not allowed in the global namespace.
 int _rsGeneralUpdate( generalUpdateInp_t *generalUpdateInp );
```
korydraughn, line 12: Sooo, C and C++ do not allow leading underscores in the global namespace and the last time I checked, iRODS uses leading underscores on functions everywhere.

Issue coming soon.

korydraughn: See https://clang.llvm.org/extra/clang-tidy/checks/bugprone/reserved-identifier.html.

alanking: Niiiiice

alanking: Niiice

lib/api/include/irods/generalUpdate.h
```diff
@@ -1,12 +1,19 @@
...
 
 #ifdef __cplusplus
-extern "C"
+extern "C" {
 #endif
-int rcGeneralUpdate( rcComm_t *conn, generalUpdateInp_t *generalUpdateInp );
 
+int rcGeneralUpdate(rcComm_t* conn, generalUpdateInp_t* generalUpdateInp)
+    __attribute__((deprecated("rcGeneralUpdate is deprecated and will be removed in a later release. If you are "
+                              "reaching for this, consider revising your approach.")));
```
alanking, line 13: First of all, this formatting lol

Secondly, I feel like the wording could be a little more helpful. Is there any way to offer alternatives, or is this too "general"? Even just suggesting that other endpoints likely exist that will meet the desired use case could be helpful, I think.

alanking (commented):
Do we need to return a deprecation message in the API as well?

lib/api/include/irods/generalUpdate.h
korydraughn: > First of all, this formatting lol

I originally had the attribute above the function prototype, but clang-format wanted to format it in a weird way. So, here we are :-)

> Secondly, I feel like the wording could be a little more helpful. Is there any way to offer alternatives, or is this too "general"? Even just suggesting that other endpoints likely exist that will meet the desired use case could be helpful, I think.

Based on the issue, this API should not exist because it allows users to change data in the database directly. Suggesting an alternative would depend on what the caller was attempting to do. For that reason, I don't think we can offer good alternatives.

alanking: Fair enough. I felt like the message reads a little bit strangely, but maybe it's just me. I'll just resolve this.

korydraughn:
> Do we need to return a deprecation message in the API as well?

I don't think we need to return anything to the client. Mostly because the protocol doesn't provide a well paved / dedicated path for it. Using the rError stack feels wrong to me even though it could work.

I think the only thing we can do safely is to document it well and possibly log a message. The admin will see that and can then report it for next steps.

alanking:
Seems fine to me. Are we awaiting next steps on the open conversation?

korydraughn:
Not really.

I still need to write up an issue for the open conversation. I also need to run the tests.

server/api/include/irods/rsGeneralUpdate.hpp
korydraughn: Created #7579.

Removed TODO.

lib/api/include/irods/generalUpdate.h
korydraughn: If you feel the wording isn't good enough, I'm happy to hear alternatives.

I could remove the second statement or remove the message entirely. There's nothing stating we must give devs a reason for the deprecation. And this is a special case given there isn't an alternative API to point devs to.

Reopening conversation so I don't forget to revisit this.

trel: i think it's okay.   no good hints to give.  and this text goes away in 5.0.

alanking (commented):
Seems good for squashing. Please notify when ready

alanking (commented):
Yep, still seems good to me.

korydraughn:
Getting ready to launch those tests. Will see what happens.

korydraughn:
All unit tests pass. This did require a tweak to the user administration lib (@FifthPotato is working on the fix). I'll rebase this PR once @FifthPotato's fix lands.

Running core tests.

korydraughn:
All core tests passed.

korydraughn:
Depends on #7589

lib/api/include/irods/generalUpdate.h
korydraughn: Here are some last minute alternatives:
- rcGeneralUpdate is deprecated. Its use is highly discouraged.
- rcGeneralUpdate is deprecated. Please revise any code using this API.
- rcGeneralUpdate is deprecated. Its use should be avoided.

Thoughts?

alanking: I like "rcGeneralUpdate is deprecated. Its use should be avoided."

trel: Agreed - we don't necessarily know the clients that are using it... and that's what deprecation means most closely... avoid.

korydraughn: Will use that.

korydraughn: Changed to `Its use should be avoided`. May change it to `Use of this API should be avoided.` since that reads well in the code and in the compiler output.

The message about the API being deprecated is produced by the compiler so I removed that.

korydraughn: Leaving as is. The deprecation message is designed for the compiler and consumers of the API, rather than people reading the code.

korydraughn:
Rerunning all unit/core tests.

korydraughn:
Rebased on top of main.

All unit/core tests passed.

server/api/src/rsGeneralUpdate.cpp
```diff
@@ -11,6 +11,13 @@
 
 int
 rsGeneralUpdate( rsComm_t *rsComm, generalUpdateInp_t *generalUpdateInp ) {
+    // Deprecation messages must be handled by doing the following.
+    // The native rule engine may erase all messages in the rError array.
+    // The only way to guarantee that messages are received by the client
+    // is to add them to the rError array when the function returns.
+    const irods::at_scope_exit add_deprecated_api_warning_message{
+        [&] { addRErrorMsg(&rsComm->rError, DEPRECATED_API, "rxGeneralUpdate has been deprecated."); }};
```
alanking, line 19: I think the "Its use should be avoided" clause would be helpful here.

alanking (commented):
Just saw a few inconsistencies in the messages and wanted to make sure it was on purpose / as to why.

server/api/src/rsGeneralUpdate.cpp
korydraughn: Agreed.

lib/api/include/irods/generalUpdate.h
```diff
@@ -1,12 +1,18 @@
...
 #include "irods/rcConnect.h"
 #include "irods/rodsGeneralUpdate.h"
 
 #ifdef __cplusplus
-extern "C"
+extern "C" {
 #endif
-int rcGeneralUpdate( rcComm_t *conn, generalUpdateInp_t *generalUpdateInp );
 
+__attribute__((deprecated("Its use should be avoided."))) int rcGeneralUpdate(rcComm_t* conn,
```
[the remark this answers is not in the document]
trel: perhaps just `deprecated("Avoid use of this function.")` ?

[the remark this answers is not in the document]
korydraughn: Here's an alternative: `GeneralUpdate is deprecated. Its use should be avoided.`

That could work since `GeneralUpdate` can be viewed as a group of related things.

Thoughts?

lib/core/include/irods/rodsGeneralUpdate.h
```diff
@@ -24,10 +24,12 @@
 #define UPDATE_NOW_TIME "update_now_time"
 
 #define GeneralUpdateInp_PI "int type; struct InxValPair_PI;"
-typedef struct GeneralUpdateInp {
+
+// NOLINTNEXTLINE(modernize-use-using)
+__attribute__((deprecated)) typedef struct __attribute__((deprecated)) GeneralUpdateInp
```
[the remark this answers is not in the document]
korydraughn: Will add.

lib/api/include/irods/generalUpdate.h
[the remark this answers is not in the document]
alanking: That sounds good. This is in line with how `SimpleQuery` was deprecated: https://github.com/irods/irods/commit/185302df12b2883497bcf7d0928fbc39e04a4668

server/api/src/rsGeneralUpdate.cpp
alanking: Please update this change in light of our deprecation rules: https://github.com/irods/irods_docs/issues/124#issuecomment-2023055718

lib/api/include/irods/generalUpdate.h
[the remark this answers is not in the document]
korydraughn: Done.

server/api/src/rsGeneralUpdate.cpp
korydraughn: Done.

alanking (commented):
Seems ripe for squashing.

korydraughn:
Rebased/Squashed. Still passes unit tests.

This is ready.

alanking (approved):
Nice, no diffs. Let's # it.

korydraughn:
#'d

Issue #7562, opened 2024-03-14
Deprecate `msiSendMail`


korydraughn:
- [x] main
- [x] 4-3-stable
---
## Deprecation

`msiSendMail` provides very basic functionality for sending email. iRODS exposes mechanisms for building much better email-related tools (e.g. `msiExecCmd` and Python).

This microservice also adds an unnecessary dependency on the `mail` binary.

With that said, `msiSendMail` should be deprecated.

PR #7569, opened 2024-03-14
[#7562] Deprecate msiSendMail (main)

korydraughn:


trel:
looks good

server/re/src/mailMS.cpp
```diff
@@ -14,12 +14,16 @@
  *
  * \brief Sends email
  *
+ * This microservice is deprecated and will be removed in a later release. Consider using the Python rule engine plugin
+ * or exposing your own mail utility via msiExecCmd.
```
alanking, line 18: Consider using the `\deprecated` tag?

alanking (commented):
We used to log a message for the microservices, but... I don't see much value in that, so I think this is good. Just had one suggestion

server/re/src/mailMS.cpp
korydraughn: Will do.

alanking (commented):
Seems good to me. Squash em

korydraughn:
Squashed and pounded.

alanking (approved):

PR #7587, opened 2024-03-19
Server-side code to only allow removal of appropriate entity (4-3-stable)

FifthPotato:
Cherry-pick of https://github.com/irods/irods/pull/7533

PR #7589, opened 2024-03-20
Fix user_administration library to account for new API changes (main)

FifthPotato:
user_administration is also a rsGeneralAdmin client (d'oh!) and I didn't update it for the new [changes](https://github.com/irods/irods/pull/7587).

This fixes that. 

Question: does this also need to server sniff, maybe?

Unit tests pass; haven't tried the whole suite yet.

### DIFF
--- a/lib/administration/user/src/user_administration.cpp
+++ b/lib/administration/user/src/user_administration.cpp
@@ -6,6 +6,7 @@
 
 #include "irods/irods_query.hpp"
 #include "irods/query_builder.hpp"
+#include "irods/version.hpp"
 
 namespace irods::experimental::administration::NAMESPACE_IMPL
 {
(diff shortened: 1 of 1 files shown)


korydraughn:
> Question: does this also need to server sniff, maybe?

Yes. Keep in mind the lib handles RcComm and RsComm objects. You'll need to use the preprocessor to handle them. Here's an example.
https://github.com/irods/irods/blob/8bf8637b33a27f42029b58a56dc507b1ffc0309d/lib/administration/user/src/user_administration.cpp#L16-L25

> Unit tests pass; haven't tried the whole suite yet.

Once you think you have it working, run the unit tests and the core tests.

Also, you're missing the issue number in the commit message.

FifthPotato:
Core and unit tests pass.

alanking:
Great! Are we still integrating the version sniff stuff? Don't quite recall where we landed after the team discussion the other day

FifthPotato:
> Great! Are we still integrating the version sniff stuff? Don't quite recall where we landed after the team discussion the other day

I think so. I did end up doing it for icommands.

alanking:
Okey dokey. Awaiting those changes and I think we'll be good

FifthPotato:
> Okey dokey. Awaiting those changes and I think we'll be good

I did a dumb. Forgot to actually push up the commit with that change. I was testing against this version.

lib/administration/user/src/user_administration.cpp
```diff
@@ -6,6 +6,7 @@
 
 #include "irods/irods_query.hpp"
 #include "irods/query_builder.hpp"
+#include <irods/version.hpp>
```
alanking, line 9: Please use quotation marks instead of the angle brackets here:
```suggestion
#include "irods/version.hpp"
```

alanking (approved):
Looks good to me. I say squash and # it.

FifthPotato:
Octothorpe'd

alanking:
@FifthPotato - Please cherry-pick to 4-3-stable. Thanks!

PR #7595, opened 2024-03-22
[6610] Deprecated RBUDP (4-3-stable)

JustinKyleJames:
I am opening this so that everyone can take a look.  I am not sure if I did this correctly or not.  When someone calls the various -Q options via icommands (iput, iget, icp, irepl), they will get a warning message that states the following:

```
Warning: RBUDP (-Q) is deprecated. Its use should be avoided.
```
The following will show up in the server log:

```
Warning: RBUDP transfer option was called.  This has been deprecated.
```
I am not sure I implemented the client side the correct way.  Furthermore, I am not sure that this code has any effect.

```
    const irods::at_scope_exit add_deprecated_api_warning_message{[&] {
        addRErrorMsg(&rsComm->rError, DEPRECATED_API, "RBUDP has been deprecated.");
    }};
```

I was hoping this would make it back to the client so that this warning could be seen by non-icommands.

Finally, there are many methods that are called by the ones deprecated here that could also be deprecated.  Not sure how far I should go with that.

### DIFF
--- a/lib/core/src/irods_parse_command_line_options.cpp
+++ b/lib/core/src/irods_parse_command_line_options.cpp
@@ -40,7 +40,7 @@ static int parse_program_options(
         ("num_threads,N", po::value<int>(), "numThreads - the number of threads to use for the transfer. A value of 0 means no threading. By default (-N option not used) the server decides the number of threads to use.")
         ("physical_path,p", po::value<std::string>(), "physicalPath - the absolute physical path of the uploaded file on the server")
         ("progress,P", "output the progress of the upload.")
-        ("rbudp,Q", "use RBUDP (datagram) protocol for the data transfer")
+        ("rbudp,Q", "[Deprecated] use RBUDP (datagram) protocol for the data transfer")
         ("recursive,r", "recursive - store the whole subdirectory")
         ("dest_resc,R", po::value<std::string>(), "resource - specifies the resource to store to. This can also be specified in your environment or via a rule set up by the administrator")
         ("ticket,t", po::value<std::string>(), "ticket - ticket (string) to use for ticket-based access")
--- a/server/core/src/miscServerFunct.cpp
+++ b/server/core/src/miscServerFunct.cpp
@@ -439,7 +439,10 @@ svrPortalPutGet( rsComm_t *rsComm ) {
 
     if ( getUdpPortFromPortList( thisPortList ) != 0 ) {
         /* rbudp transfer */
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
         retVal = svrPortalPutGetRbudp( rsComm );
+#pragma clang diagnostic pop
         return retVal;
     }
 
@@ -1319,7 +1322,6 @@ rbudpRemLocCopy( dataCopyInp_t *dataCopyInp ) {
         status = getFileToPortalRbudp( portalOprOut, NULL,
                                        FileDesc[destL3descInx].fd,
                                        veryVerbose, packetSize );
-
     }
     else {
         int srcL3descInx = dataOprInp->srcL3descInx;
@@ -1361,7 +1363,10 @@ int remLocCopy(rsComm_t *rsComm, dataCopyInp_t *dataCopyInp)
 
     if ( getUdpPortFromPortList( &portalOprOut->portList ) != 0 ) {
         /* rbudp transfer */
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
         return rbudpRemLocCopy( dataCopyInp );
+#pragma clang diagnostic pop
     }
 
     if ( numThreads > MAX_NUM_CONFIG_TRAN_THR || numThreads <= 0 ) {
--- a/lib/api/src/rcDataObjCopy.cpp
+++ b/lib/api/src/rcDataObjCopy.cpp
@@ -59,11 +59,11 @@
  *    \n DEST_RESC_NAME_KW - The resource to store this data object
  *    \n FILE_PATH_KW - The physical file path for this data object if the
  *             normal resource vault is not used.
- *    \n RBUDP_TRANSFER_KW - use RBUDP for data transfer. This keyWd has no
+ *    \n RBUDP_TRANSFER_KW - (Deprecated) use RBUDP for data transfer. This keyWd has no
  *             value
- *    \n RBUDP_SEND_RATE_KW - the number of RBUDP packet to send per second
+ *    \n RBUDP_SEND_RATE_KW - (Deprecated) the number of RBUDP packet to send per second
  *	    The default is 600000
- *    \n RBUDP_PACK_SIZE_KW - the size of RBUDP packet. The default is 8192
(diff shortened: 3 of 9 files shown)


server/core/src/miscServerFunct.cpp
```diff
@@ -2023,8 +2032,18 @@ longNoSupport( ... ) {
...
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
 int
 svrPortalPutGetRbudp( rsComm_t *rsComm ) {
+#pragma clang diagnostic pop
+
+    const irods::at_scope_exit add_deprecated_api_warning_message{[&] {
+        addRErrorMsg(&rsComm->rError, DEPRECATED_API, "RBUDP has been deprecated.");
+    }};
+    
+    irods::experimental::log::server::warn("Warning: RBUDP transfer option was called.  This has been deprecated.");
```
alanking, line 2045: I'm not sure that we need to log a message for this. The rError message should be enough.

alanking (commented):
I think that the rError message you implemented is fine. What makes you think it won't make it back to the client? Since the `RcComm` is part of communicating with the server, I *think* most clients will have some form of `rError` stack and the message will be in there, even for non-icommands.

server/core/src/miscServerFunct.cpp
alanking: That being said, I'm not opposed to it. I defer to others.

JustinKyleJames:
> I think that the rError message you implemented is fine. What makes you think it won't make it back to the client? Since the `RcComm` is part of communicating with the server, I _think_ most clients will have some form of `rError` stack and the message will be in there, even for non-icommands.

I am not seeing that logged when I run icommands with the -Q option.  That is why I implemented specific warning log messages in the parse command line methods.

alanking:
Oh, I see. And you do see the warning message in the log? Hmm... I wonder if the message isn't making it back because it establishes a separate connection with the client and the rError stack isn't copied back to the original 1247 connection.

Perhaps we should just print a message in the iCommands which use the option. I'm not sure whether any other clients have RBUDP implemented, but given that traditional parallel transfer is implemented for very few, if any, I would guess that RBUDP is not either.

Thoughts?

JustinKyleJames:
> Perhaps we should just print a message in the iCommands which use the option. I'm not sure whether any other clients have RBUDP implemented, but given that traditional parallel transfer is implemented for very few, if any, I would guess that RBUDP is not either.

I have that message printed in the iCommands so that should be covered.

lib/core/src/irods_parse_command_line_options.cpp
```diff
@@ -151,6 +151,7 @@ static int parse_program_options(
         _rods_args.progressFlag = 1;
     }
     if ( global_prog_ops_var_map.count( "rbudp" ) ) {
+        printf("Warning: RBUDP (-Q) is deprecated. Its use should be avoided.\n");
```
korydraughn, line 154: Let's remove this since the help text states the option is deprecated.

lib/core/src/parseCommandLine.cpp
```diff
@@ -418,6 +418,7 @@ parseCmdLineOpt( int argc, char **argv, const char *optString, int includeLong,
             rodsArgs->queryStr = optarg;
             break;
         case 'Q':
+            printf("Warning: RBUDP (-Q) is deprecated. Its use should be avoided.\n");
```
korydraughn, line 421: Let's remove this since the help text states the option is deprecated.

alanking:
> I have that message printed in the iCommands so that should be covered.

Oh, I see it now. But you said that you're not seeing the messages when the options are used with iCommands. Are you only referring to the rError message from the API?

korydraughn (commented):
Nothing strange caught my eye.

lib/core/src/irods_parse_command_line_options.cpp
alanking: The rules have changed: https://github.com/irods/irods_docs/issues/124#issuecomment-2023055718

JustinKyleJames: I removed all warning messages.

lib/core/src/parseCommandLine.cpp
JustinKyleJames: done

server/core/src/miscServerFunct.cpp
JustinKyleJames: Done

server/core/src/miscServerFunct.cpp
```diff
@@ -2023,8 +2032,18 @@ longNoSupport( ... ) {
     return ( rodsLong_t ) SYS_NOT_SUPPORTED;
 }
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
 int
 svrPortalPutGetRbudp( rsComm_t *rsComm ) {
+#pragma clang diagnostic pop
```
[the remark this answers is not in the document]
JustinKyleJames: It wasn't necessary.  I removed these pragmas.

server/core/src/miscServerFunct.cpp
```diff
@@ -1280,8 +1283,12 @@ remToLocPartialCopy( portalTransferInp_t *myInput ) {
 /* rbudpRemLocCopy - The rbudp version of remLocCopy.
  */
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
 int
 rbudpRemLocCopy( dataCopyInp_t *dataCopyInp ) {
+#pragma clang diagnostic pop
```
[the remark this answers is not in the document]
JustinKyleJames: It wasn't necessary.  I removed these pragmas.

alanking (commented):
Seems good to me. Please squash to taste

korydraughn:
Please run clang-format on the changes.

korydraughn:
So there's no `rc/rs` APIs for RBUDP? It's fine if the answer is yes.

alanking:
Well... I'm not sure how one deprecates a whole directory, but we do have lib/rbudp which has some... stuff... Not sure how far to go with it.

korydraughn:
If that directory contains implementation details then we don't need to touch those. We should focus on the public items.

If there aren't any rc/rs APIs then we're ready to go. And considering the release notes will contain a statement about deprecating RBUDP, we're good.

korydraughn:
Please update the commit message to say `Deprecated RBUDP` so it covers all RBUDP related things.

JustinKyleJames:
> Please update the commit message to say `Deprecated RBUDP` so it covers all RBUDP related things.

Done

JustinKyleJames:
> If that directory contains implementation details then we don't need to touch those. We should focus on the public items.
> 
> If there aren't any rc/rs APIs then we're ready to go. And considering the release notes will contain a statement about deprecating RBUDP, we're good.

I could not find any rc/rs API's for it.

alanking (approved):
Seems good to me. Please await one more approval before #'ing.

korydraughn (approved):
Pound it!

korydraughn:
Don't forget to run clang-format.

JustinKyleJames:
> Pound it!

Done

korydraughn:
Please cherry-pick to the main branch, no #'s

Issue #7371, opened 2023-10-09
Ensure irods-dev pulls in proper curl-dev dependencies


SwooshyCueb:
- [x] main
- [x] 4-3-stable

-----

Should be fairly trivial, but not urgent enough (IMO) to put in 4.3.1.

We want to encourage downstream projects to use the same version of curl that irods is built against, and since debian/ubuntu provides three to choose from (openssl, gnutls, nss), we should ensure that the curl development package used during build winds up in the dependencies of `irods-dev`. However, this should be a "recommends" declaration, as a hard requirement may cause issues for other packages.

PR #7596, opened 2024-03-23
[#7371] Add curl to development package dependencies (main)

SwooshyCueb:
Addresses #7371


cmake/packaging.cmake
```diff
@@ -54,40 +54,33 @@ endif()
 
 set(CPACK_ARCHIVE_COMPONENT_INSTALL ON)
 
+# identify proper curl packages for debian
```
alanking, line 57: Is "debian" referring to the distro or the package format (which would include Ubuntu)?

alanking (commented):
Just had one clarifying question. Seems fine

cmake/packaging.cmake
SwooshyCueb: package format

alanking (approved):
Seems good to me. I think we can go ahead and # this.

SwooshyCueb:
\#'d

Issue #7390, opened 2023-10-20
test_password_extend_lifetime_set_to_false_invalidates_other_authentications_on_expiration can fail when run with the Python REP on Ubuntu 22.04 


korydraughn:
- [x] main
- [x] 4-3-stable
---
## Bug Report
https://github.com/irods/irods/blob/97eb33f130349db5e01a4b85e89dd1da81460345/scripts/irods/test/test_pam_password_authentication.py#L563-L566

The assertion above sometimes fails with the following:
```
Level 0: Error occurred while authenticating user [irodsauthuser] [CAT_INVALID_AUTHENTICATION: failed to perform request
```

The failure feels very related to timing.

trel:
sometimes it passes?   usually?

korydraughn:
It's been passing, but there's no reason to think it can't fail because the test relies on sleeping to lead to certain assertions.

trel:
ah, then most certainly your assessment is correct.

alanking:
Even if I can reproduce this, I'm not sure if I can do anything about it. The test is inherently timing-based as it is testing for a password expiration. I would like to propose linking this issue with https://github.com/irods/irods/issues/7344 as it seems to be the cause of the problem.

A possible workaround for the time being would be to accept either error message in this test and fix it properly in #7344, putting back the correct assertions. See also: 
https://github.com/irods/irods/blob/06ed23911a6fc3f6f4a35d81cf917bfa3e366a1a/scripts/irods/test/test_pam_password_authentication.py#L328-L329
https://github.com/irods/irods/blob/06ed23911a6fc3f6f4a35d81cf917bfa3e366a1a/scripts/irods/test/test_pam_password_authentication.py#L454-L456
https://github.com/irods/irods/blob/06ed23911a6fc3f6f4a35d81cf917bfa3e366a1a/scripts/irods/test/test_pam_password_authentication.py#L570-L572
https://github.com/irods/irods/blob/06ed23911a6fc3f6f4a35d81cf917bfa3e366a1a/scripts/irods/test/test_native_authentication.py#L172-L173
https://github.com/irods/irods/blob/06ed23911a6fc3f6f4a35d81cf917bfa3e366a1a/scripts/irods/test/test_native_authentication.py#L265-L267
https://github.com/irods/irods/blob/06ed23911a6fc3f6f4a35d81cf917bfa3e366a1a/scripts/irods/test/test_native_authentication.py#L346-L348

korydraughn:
[...]

PR #7609, opened 2024-03-28
[#7390] Auth tests: Loosen error message assertions (main)

alanking:
Addresses https://github.com/irods/irods/issues/7390

Just ran the updated pam_password test and it passed like 20 times in a row, so probably okay. The native auth version was updated but the test is being skipped right now as the feature in question is not supported for that auth scheme.


korydraughn (approved):
Looks good. Pound it.

We linking the code snippets to issue 7344?

Issue #7319, opened 2023-09-13
`msiSetResource` crashes agent


alanking:
- [x] main (msiSetResource)
- [x] 4-3-stable (msiSetResource)
- [x] main (msiSetDefaultResc)
- [x] 4-3-stable (msiSetDefaultResc)
---

### iRODS Version, OS and Version
4.3.0, probably all versions

### What did you try to do?
Use `msiSetResource` in `pep_api_data_obj_put_pre`:
```c
pep_api_data_obj_put_pre(*INSTANCE_NAME, *COMM, *DATAOBJINP, *BUFFER, *PORTAL_OPR_OUT) {
    msiSetResource("testResc");
}
```
Then perform a simple `iput`.

### Expected behavior
I am not sure what `msiSetResource` does, but I would expect the agent to survive the invocation.

### Observed behavior (including steps to reproduce, if applicable)
`SYS_INTERNAL_ERR` comes back to the client, and a very long stacktrace appears in the log, caused by signal 11 from `snprintf`:
```json
[...]

PR #7610, opened 2024-03-28
[#7319] Make `msiSetResource` not crash agent

MartinFlores751:
This provides a simple fix to `msiSetResource(...)` by adding a null check before dereference.

### DIFF
--- a/scripts/irods/test/test_native_rule_engine_plugin.py
+++ b/scripts/irods/test/test_native_rule_engine_plugin.py
@@ -831,6 +831,30 @@ OUTPUT ruleExecOut
         finally:
             os.unlink(rule_file)
 
+    @unittest.skipIf(plugin_name == 'irods_rule_engine_plugin-python', 'rule language only')
+    def test_msiSetResource__issue_7319(self):
+        rule_map = {
+            'irods_rule_engine_plugin-irods_rule_language': textwrap.dedent('''
+                test_msiSetResource {{
+                    msiSetResource("testResc");
+                }}
+                OUTPUT ruleExecOut
+                ''')
+        }
+
+        rule_file = 'test_msiSetResource.r'
+        with open(rule_file, 'w') as f:
+            f.write(rule_map[self.plugin_name])
+
+        try:
+            rep_name = 'irods_rule_engine_plugin-irods_rule_language-instance'
+
+            self.admin.assert_icommand(['irule', '-r', rep_name, '-F', rule_file], 'STDERR', 'SYS_INTERNAL_NULL_INPUT_ERR', desired_rc=4)
+            self.user0.assert_icommand(['irule', '-r', rep_name, '-F', rule_file], 'STDERR', 'SYS_INTERNAL_NULL_INPUT_ERR', desired_rc=4)
+
+        finally:
+            os.remove(rule_file)
+
     def test_re_serialization_for_RsComm_ptr__issue_5950(self):
         pep_map = {
             'irods_rule_engine_plugin-irods_rule_language': textwrap.dedent('''
--- a/server/re/src/icatGeneralMS.cpp
+++ b/server/re/src/icatGeneralMS.cpp
@@ -6,8 +6,14 @@
 #include "irods/miscServerFunct.hpp"
 #include "irods/modAccessControl.h"
 #include "irods/irods_configuration_keywords.hpp"
+#include "irods/irods_logger.hpp"
 #include "irods/rsModAccessControl.hpp"
 #include "irods/irods_re_structs.hpp"
+
+namespace
+{
+    using log_msi = irods::experimental::log::microservice;
+} // anonymous namespace
 
 /**
  * \fn msiGetIcatTime (msParam_t *timeOutParam, msParam_t *typeInParam, ruleExecInfo_t *)
@@ -138,11 +144,14 @@ msiQuota( ruleExecInfo_t *rei ) {
  * \sa none
 **/
 int  msiSetResource( msParam_t* xrescName, ruleExecInfo_t *rei ) {
-    char *rescName;
+    char* rescName{static_cast<char*>(xrescName->inOutStruct)};
 
-    rescName = ( char * ) xrescName->inOutStruct;
+    if (nullptr == rei->doi) {
+        log_msi::error("{}: Cannot set resource, no DOI (DataObjInfo) attached to REI (RuleExecInfo).", __func__);
+        return SYS_INTERNAL_NULL_INPUT_ERR;
+    }
 
-    snprintf( rei->doi->rescName, sizeof( rei->doi->rescName ), "%s", rescName );
+    std::snprintf(rei->doi->rescName, sizeof(rei->doi->rescName), "%s", rescName);
     return 0;
 }
 


server/re/src/icatGeneralMS.cpp
```diff
@@ -138,11 +138,14 @@ msiQuota( ruleExecInfo_t *rei ) {
  * \sa none
 **/
 int  msiSetResource( msParam_t* xrescName, ruleExecInfo_t *rei ) {
-    char *rescName;
+    char* rescName{static_cast<char *>(xrescName->inOutStruct)};
 
-    rescName = ( char * ) xrescName->inOutStruct;
+    if (rei->doi == nullptr) {
+        rodsLog(LOG_ERROR, "Cannot set resource, DOI not found.");
```
korydraughn, line 144: Use the new logger.
```cpp

namespace
{
    using log_msi = irods::experimental::log::microservice;
} // anonymous namespace

// Replace rodsLog with ...
log_msi::error("your message ...");
```

See https://docs.irods.org/4.3.1/developers/logging/ for more info on the logging library.

server/re/src/icatGeneralMS.cpp
```diff
@@ -138,11 +138,14 @@ msiQuota( ruleExecInfo_t *rei ) {
...
+        return ACTION_FAILED_ERR;
```
korydraughn, line 145: Leave yourself a comment about updating this error code so you don't forget.

server/re/src/icatGeneralMS.cpp
```diff
@@ -138,11 +138,14 @@ msiQuota( ruleExecInfo_t *rei ) {
  * \sa none
 **/
 int  msiSetResource( msParam_t* xrescName, ruleExecInfo_t *rei ) {
-    char *rescName;
+    char* rescName{static_cast<char *>(xrescName->inOutStruct)};
 
-    rescName = ( char * ) xrescName->inOutStruct;
+    if (rei->doi == nullptr) {
```
korydraughn, line 143: Swap the operands for extra protection.

scripts/irods/test/test_native_rule_engine_plugin.py
```diff
@@ -831,6 +831,53 @@ def test_msiSegFault(self):
...
+
+            local_file = os.path.join(self.admin.local_session_dir, 'random_file')
+            lib.make_file(local_file, 2048, 'arbitrary')
+
+            dest_path = os.path.join(self.admin.session_collection, 'random_file')
+
+            self.admin.assert_icommand(['iput', local_file, dest_path], desired_rc=0)
+
+    @unittest.skipIf(plugin_name == 'irods_rule_engine_plugin-python', 'rule language only')
+    def test_msiSetResource__7319(self):
```
korydraughn, line 855: `__7319` -> `__issue_7319`

korydraughn (commented):
Looks good.

server/re/src/icatGeneralMS.cpp
MartinFlores751: Done.

server/re/src/icatGeneralMS.cpp
MartinFlores751: What's the benefit of the anonymous namespace? Does it keep log_msi better localized?

server/re/src/icatGeneralMS.cpp
MartinFlores751: I'm assuming this is for preventing accidental `=` instead of `==`, right?

korydraughn: Correct.

server/re/src/icatGeneralMS.cpp
korydraughn: Correct. The compiler will mark the symbols as private to the translation unit. It's not required. It's just one way to do things.

If you prefer to move the alias into the function directly, that's perfectly fine too.

scripts/irods/test/test_native_rule_engine_plugin.py
MartinFlores751: Done.

server/re/src/icatGeneralMS.cpp
MartinFlores751: I was mostly just curious, since I don't think I've seen it used for this before.  Done.

server/re/src/icatGeneralMS.cpp
```diff
@@ -141,7 +147,7 @@ int  msiSetResource( msParam_t* xrescName, ruleExecInfo_t *rei ) {
     char* rescName{static_cast<char *>(xrescName->inOutStruct)};
 
     if (rei->doi == nullptr) {
-        rodsLog(LOG_ERROR, "Cannot set resource, DOI not found.");
+        log_msi::error("Cannot set resource, DOI not found.");
```
korydraughn, line 150: `DOI` -> `DOI (DataObjInfo)`

korydraughn, line 150: `"Cannot ..."` -> `"{}: Cannot ...", __func__`

MartinFlores751: Should we just make this in plain english, "Data Object Info"?

MartinFlores751: Done.

korydraughn: No. `DataObjInfo` is a real thing in iRODS and using plain english to try and refer to it would only lead to more confusion.

MartinFlores751: Done.

server/re/src/icatGeneralMS.cpp
```diff
@@ -138,11 +144,14 @@ msiQuota( ruleExecInfo_t *rei ) {
  * \sa none
 **/
 int  msiSetResource( msParam_t* xrescName, ruleExecInfo_t *rei ) {
-    char *rescName;
+    char* rescName{static_cast<char *>(xrescName->inOutStruct)};
 
-    rescName = ( char * ) xrescName->inOutStruct;
+    if (nullptr == rei->doi) {
+        log_msi::error("{}: Cannot set resource, DOI (DataObjInfo) not found.", __func__);
```
korydraughn, line 150: `"not found."` seems a little misleading.

How do you feel about `"Cannot set resource, no DOI (DataObjInfo) attached to REI."`?

That captures the situation a bit better IMO. Thoughts?

korydraughn: Or is that too specific?

MartinFlores751: I'm fine with the more verbose description. Should we 'full name' the REI like we did with DOI?

korydraughn: Sure. It wouldn't hurt to put it in parens.

MartinFlores751: Done.

korydraughn:
Seems good to me.

Let's run clang-format.

Does it compile? Do the unit/core tests pass?

korydraughn:
Thinking about this more, I don't think we need to run the unit/core tests.

MartinFlores751:
It does compile. At the very least, I did run the new test and it passed.

Can run all tests if desired.

alanking (commented):
I think the new test passing is good enough. Probably ready to squash

MartinFlores751:
Squashed.

alanking (approved):
Seems good to me. Let's wait for one more approval before #'ing.

korydraughn (approved):
Pound it!

MartinFlores751:
Pounded!

korydraughn:
Nice. Let's get that cherry-pick to 4-3-stable.

PR #7624, opened 2024-04-02
[7540] Fix double-free in rsCollRepl (main)

FifthPotato:
Funny-looking PR. Debugged with asan.

`rsCloseCollection` calls `freeCollHandle` calls `clearCollHandle` calls `clearCollSqlResult`, which frees the `collSqlResult_t`'s members. 

However, `rsReadCollection` calls `readCollection` calls `getNextCollMetaInfo`/`getNextDataObjMetaInfo`, and pointers from the `collSqlResult_t` are assigned to the `collEnt_t`'s members (no copying!), so freeing the members of the `collEnt_t` results in a double-free (since `clearCollSqlResult` should have already freed it).

The only reason this worked before (but not in error cases) was because the double-free was evaded by `collEnt` being set to `NULL` a few lines above (as the final operation of the `while` loop). This means `freeCollEnt` effectively becomes a no-op, since it instantly returns if `NULL` (and doesn't call `clearCollEnt`, the offending function). 

However, in error cases, like when `rsDataObjRepl` fails, the `break` immediately breaks out of the loop. This means `collEnt` is not null, so `clearCollEnt` is called.

Finally, my suggested change doesn't double-free on normal operation because after the while-loop terminates, `collEnt` is freed and set to `NULL` in `rsReadCollection` when `status < 0`. Per the [C standard](https://www.open-std.org/JTC1/SC22/wg14/www/docs/n1124.pdf) (7.20.3.2/2), `free(NULL);` does nothing.

Additionally, I have confirmed with asan that not freeing `collEnt` here (or somewhere) will cause a leak in error cases.

Will run tests shortly.

Whew, I think that's all of it. A lot of words for a one line commit.

### DIFF
--- a/server/api/src/rsCollRepl.cpp
+++ b/server/api/src/rsCollRepl.cpp
@@ -138,7 +138,7 @@ rsCollRepl( rsComm_t *rsComm, collInp_t *collReplInp,
         collEnt = NULL;
     }
     rsCloseCollection( rsComm, &handleInx );
-    freeCollEnt( collEnt );
+    free(collEnt);
 
     return savedStatus;
 }


korydraughn:
Excellent explanation.

Consider adding a summary of your explanation to the body of the commit message.

Let's make clang-tidy happy by adding a `// NOLINT(checks...)` directive. There should be plenty of examples throughout the codebase.

korydraughn:
Let us know how the results of the tests once they're done.

alanking (commented):
So... the issue is that all of the contents of `collEnt` have already been freed in the error case (and apparently also in the success case), and we just need to free `collEnt` itself in the case you changed? That's wild.

alanking:
Agreed, please put some version of your comment in OP for this PR in the commit message itself because it may help future perusers in Git.

FifthPotato:
Core and unit tests pass.

alanking (commented):
Very nice. Go ahead and update the commit message and we will have one last look before #'ing.

FifthPotato:
Commit message updated. Let me know if terse/verbose enough.

korydraughn (approved):
Looks good to me.

Let's wait for another approval before pounding.

FifthPotato:
Octothorpe'd

korydraughn:

Issue #7602, opened 2024-03-27
Deprecate `msiSetResource`


alanking:
- [x] main
- [x] 4-3-stable

---

This microservice is not tested and is known to crash agents (see https://github.com/irods/irods/issues/7319). `msiSetDefaultResc` is likely the replacement?

PR #7625, opened 2024-04-02
[#7602] Deprecate msiSetResource (main)

MartinFlores751:
Update documentation on `msiSetResource(...)` to note its deprecation.

### DIFF
--- a/server/re/src/icatGeneralMS.cpp
+++ b/server/re/src/icatGeneralMS.cpp
@@ -121,6 +121,8 @@ msiQuota( ruleExecInfo_t *rei ) {
  *
  * \since pre-2.1
  *
+ * \deprecated Deprecated in 4.3.2. Its use should be avoided.
+ *
  *
  * \note  This microservice sets the resource as part of a workflow execution.
  *
--- a/doxygen/microservices.cpp
+++ b/doxygen/microservices.cpp
@@ -167,7 +167,7 @@
   - #msiSetDataObjAvoidResc - Specifies the copy to avoid
   - #msiSetGraftPathScheme - Sets the scheme for composing the physical path in the vault to GRAFT_PATH
   - #msiSetRandomScheme - Sets the the scheme for composing the physical path in the vault to RANDOM
-  - #msiSetResource  - sets the resource from default
+  - #msiSetResource  - (Deprecated) sets the resource from default
   - #msiSetNumThreads - specify the parameters for determining the number of threads to use for data transfer
   - #msiNoChkFilePathPerm - Does not check file path permission when registering a file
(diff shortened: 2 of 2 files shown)


server/re/src/icatGeneralMS.cpp
```diff
@@ -121,6 +121,8 @@ msiQuota( ruleExecInfo_t *rei ) {
  *
  * \since pre-2.1
  *
+ * \deprecated Deprecated in 4.3.2. Use alternative, such as msiSetDefaultResource(...) instead.
```
korydraughn, line 124: This assumes `msiSetDefaultResource` works. Does it?

korydraughn (commented):
Looks fine overall.

server/re/src/icatGeneralMS.cpp
MartinFlores751: True, mostly threw it in there as a 'it would be nice to suggest a different solution'. Can drop it, since I'm not sure that it would cover the same case.

korydraughn: I agree that offering an alternative is good. We also have to prove that use-case is indeed covered though.

Please investigate and share what you find here.

trel: And check the name. I think it’s either Def or Resc.

alanking: `msiSetDefaultResc`!

alanking (commented):
Seems fine to me. Are we waiting on anything here? Confirmation of `msiSetDefaultResc` being a valid replacement?

korydraughn:
> Seems fine to me. Are we waiting on anything here? Confirmation of `msiSetDefaultResc` being a valid replacement?

Yes. Just waiting on confirmation.

server/re/src/icatGeneralMS.cpp
MartinFlores751: I think it's better to drop the suggestion. At the very least, they do set different properties, i.e., `rei->doi->rescName` vs `rei->rescName` if I recall correctly.

alanking: Agreed. Maybe just say "Its use should be avoided." or something to that effect

Citing the fancy new docs: https://github.com/irods/irods_docs/blob/main/docs/developers/deprecation.md#how-do-i-deprecate-a-feature

MartinFlores751: Done.

alanking (commented):
I think we are ready to squash

MartinFlores751:
Squashed.

alanking (approved):
\# it.

MartinFlores751:
#ed.

alanking:

Issue #7619, opened 2024-04-01
Server does not terminate agents after graceful shutdown period


korydraughn:
- [x] main
- [x] 4-3-stable

### iRODS Version, OS and Version

iRODS 4.3.1, Ubuntu 20.04

This occurs on tip of main too.

### What did you try to do?

Launch a long running operation and watch the server kill the agent after the shutdown grace period.
```
kory $ istream write foo
```

In a separate terminal, attempt to shutdown the server.
```
irods $ ./irodsctl -v stop
```

### Expected behavior

All iRODS processes are shutdown. The agent handling the istream operation is killed.

### Observed behavior (including steps to reproduce, if applicable)

The server shuts down, but the agent handling the istream operation still exists.

alanking:
Confirmed that this issue exists in 4.3.0 and *not* 4.2.12.

Sidebar: The data object created by `istream` in this instance is left stuck in the intermediate state when this happens. I don't think there's anything we can do about that given that the process responsible for finalizing it is supposed to be killed.

alanking:
Small update here... Here's a table of platforms where this issue exists. All of this was done in the irods_testing_environment:

| platform | commitish | issue exists? |
| ---- | ---- | ---- |
| Centos 7 | 4.3.1 | no |
| Centos 7 | 4.3.0 | no |
| Ubuntu 18.04 | 4.3.1 | no |
| Ubuntu 18.04 | 4.3.0 | no |
| Ubuntu 20.04 | 4.3.1 | yes |
| Ubuntu 20.04 | 4.3.0 | yes |
| Ubuntu 22.04 | 4.3.1 | yes |
| Almalinux 8 | 4.3.1 | no |
| Almalinux 8 | 4.3.0 | no |
| Debian 11 | 4.3.1 | yes |
| Debian 11 | 4.3.0 | yes |
| Debian 12 | 4.3.1 | yes |
| RockyLinux 9 | 4.3.1 | yes |

So it seems that the newer platforms demonstrate the issue while the older ones do not.

Investigation ongoing...

alanking:
I put this in my commit message, but I'll put it here too just in case...

The way the server is being shut down was reworked in commit https://github.com/alanking/irods/commit/1a5b0399141de3448390a2d875290c60f669342e.

[...]

PR #7629, opened 2024-04-03
[#7619] irodsctl stop: Fix orphan agent cleanup (main)

alanking:
Addresses #7619

I've confirmed the fix and test validity on Ubuntu 22.04. I would like to test on Centos 7, too, because we will be releasing for that platform for 4.3.2 and it uses python2 by default.

Need to run core tests as well.

### DIFF
--- a/scripts/irods/controller.py
+++ b/scripts/irods/controller.py
@@ -293,7 +293,7 @@ def capture_process_tree(server_proc, server_descendants, candidate_binaries=Non
 
     if server_proc.is_running():
         try:
-            cur_descendants = filter(should_return_proc, server_proc.children(recursive=True))
+            cur_descendants = set(filter(should_return_proc, server_proc.children(recursive=True)))
             orphaned_descendants = server_descendants.difference(cur_descendants)
             server_descendants.update(cur_descendants)
         except (psutil.NoSuchProcess):
--- a/scripts/irods/test/test_irodsctl.py
+++ b/scripts/irods/test/test_irodsctl.py
@@ -1,21 +1,20 @@
+import unittest
+
+from subprocess import PIPE
 import json
 import os
 import shutil
 import sys
 import tempfile
 
-if sys.version_info < (2, 7):
-    import unittest2 as unittest
-else:
-    import unittest
-
-from .. import test
+from . import session
 from . import settings
-from ..test.command import assert_command
 from .. import lib
 from .. import paths
-from ..controller import IrodsController
+from .. import test
 from ..configuration import IrodsConfig
+from ..controller import IrodsController, capture_process_tree
+from ..test.command import assert_command
 
 class Test_Irodsctl(unittest.TestCase):
     def test_re_shm_creation(self):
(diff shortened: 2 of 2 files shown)


scripts/irods/test/test_irodsctl.py
```diff
@@ -43,6 +42,69 @@ def test_configuration_schema_validation_from_file(self):
...
+    def test_agents_kept_alive_by_connected_clients_are_cleaned_up_after_failed_graceful_shutdown__issue_7619(self):
+        # Get the server process so that we can monitor the child processes.
+        server_proc = IrodsController().get_server_proc()
+        self.assertIsNotNone(server_proc)
+
+        with session.make_session_for_existing_admin() as admin_session:
+            user = session.mkuser_and_return_session('rodsuser', 'bilbo', 'bpass', lib.get_hostname())
+            logical_path = f'{user.session_collection}/the_replica_that_never_should_have_been'
+
+            # Use a nested try-except here because we want to ensure that the user session is cleaned up.
```
alanking, line 54: Whoops. Remove this.

korydraughn (commented):
Seems good to me.

Let's get another set of eyes on this.

alanking:
I have confirmed that using `filter()` in python2 yields the same result as `list(filter())` in the case with which we are concerned in this issue:
```python
>>> l = [1,2,3,4]
>>> print filter(lambda x: 0 == x % 2, l)
[2, 4]
>>> print list(filter(lambda x: 0 == x % 2, l))
[2, 4]
```

scripts/irods/test/test_irodsctl.py
alanking: Done

korydraughn:
Which means you don't have to worry about version sniffing, correct?

alanking:
Correct.

alanking:
Core tests passed.

korydraughn:
Are you still testing centos 7?

alanking:
I cannot build on main for Centos 7 right now, so I guess what I can do is go ahead and cherry-pick to 4-3-stable and try building there. If we still can't build then we have a problem. Will write back once I get some results.

korydraughn:
Sounds good.

MartinFlores751 (approved):
Everything looks good to me.

scripts/irods/controller.py
```diff
@@ -293,7 +293,7 @@ def should_return_proc(_p):
 
     if server_proc.is_running():
         try:
-            cur_descendants = filter(should_return_proc, server_proc.children(recursive=True))
+            cur_descendants = list(filter(should_return_proc, server_proc.children(recursive=True)))
```
FifthPotato, line 296: If we're using `difference` below, does that mean `server_descendants` is a `set`? Would it make more sense to use a `set` here to compute `set` difference?

FifthPotato (commented):
Looks pretty good to me. One possible nitpick is all I got!

scripts/irods/controller.py
alanking: `server_descendants` is indeed a `set`. I'll look into usin `set` here instead.

alanking: Yep, that works. Good catch

alanking:
I have confirmed that the existing fix works on Centos 7 (which also means that 4-3-stable builds for Centos 7, hooray). I also confirmed that the `set` instead of `list` works.

korydraughn:
Seems ready to go. Is there anything left to do for this?

alanking:
I don't think so. I'll go ahead and squash with no #s.

alanking (commented):
Hmm... wait a second, something weird happened. Hold please...

scripts/irods/controller.py
```diff
@@ -293,7 +293,7 @@ def should_return_proc(_p):
 
     if server_proc.is_running():
         try:
-            cur_descendants = filter(should_return_proc, server_proc.children(recursive=True))
+            cur_descendants = set(filter(should_return_proc, server_proc.children(recursive=True)))
```
alanking, line 296: This is t

alanking: Fixed.

alanking:
Okay, fixed the weird thing and confirmed that the test still passes.

korydraughn (approved):
Seems ready to pound.

Go for it.

alanking:
#'d, and mergin.

Issue #7570, opened 2024-03-14
Absorb GenQuery2 parser into iRODS server


korydraughn:
- [x] main
- [x] 4-3-stable
---
## Enhancement

The first step to replacing the original GenQuery parser is to make it available in the next release of iRODS. This means the following:
- Exposing a new API endpoint
- Exposing new microservices
- Adding options to `iquest` to allow use of the new parser
    - Or, providing a new icommand (i.e. `iquery`)

The new parser would exist as an alternative to GenQuery1 until the issues are worked out.

All APIs would continue to use GenQuery1.

This work would result in the GenQuery2 repo being archived.

korydraughn:
This work is now complete.

For those running GenQuery2 built from https://github.com/irods/irods_api_plugin_genquery2, you need to do the following before upgrading to 4.3.2.
- Uninstall the GenQuery2 package
- Remove the GenQuery2 rule engine plugin from your server_config.json
    - iRODS 4.3.2 does not require the need for a rule engine plugin
- Replace all calls to the GenQuery2 microservices with the 4.3.2 variants

If this isn't done, the upgrade process will halt and report conflicts with the GenQuery2 package. This is intentional and serves as a means for telling the admin they need to examine their deployment.

Use of the GenQuery2 repo/package with 4.3.2 and later should be avoided.

alanking:
I'm guessing we will add something like that to the release notes?

korydraughn:
Absolutely.

PR #7644, opened 2024-04-08
[#7570] Declare server and dev package conflict with genquery2 package (main)

SwooshyCueb:
In service of #7570 

Adds the genquery2 plugin package to the "conflicts" delcarations for dev and server packages.


korydraughn (commented):
Changes look fine.

What does REPLACES and OBSOLETES do if the genquery2 package is installed prior to 4.3.2?

korydraughn:
If someone has policy which depends on things in the genquery2 package, would this PR break their deployment?

SwooshyCueb:
> What does REPLACES and OBSOLETES do if the genquery2 package is installed prior to 4.3.2?

The genquery2 packages will be uninstalled when the irods packages are updated to 4.3.2.

> If someone has policy which depends on things in the genquery2 package, would this PR break their deployment?

I have no idea.

Given that we never actually released the genquery2 plugin, I don't think it's of much concern.

korydraughn:
> Given that we never actually released the genquery2 plugin, I don't think it's of much concern.

I see what you're saying, but I'm not all the way there yet. If an admin deploys policy/rules which use names only available in the genquery2 and we automatically uninstall the package on upgrade, their system could break when those rules are encountered. It feels wrong to uninstall the package without giving the admin a heads up.

Let's say the PR handled this via the CONFLICTS property ...
- What happens if the admin attempts to install the genquery2 package on a 4.3.2 server?
- What happens if genquery2 is installed on a 4.3.1 server and the admin upgrades to 4.3.2?

SwooshyCueb:
> > Given that we never actually released the genquery2 plugin, I don't think it's of much concern.
> 
> I see what you're saying, but I'm not all the way there yet. If an admin deploys policy/rules which use names only available in the genquery2 and we automatically uninstall the package on upgrade, their system could break when those rules are encountered. It feels wrong to uninstall the package without giving the admin a heads up.

Makes sense

> Let's say the PR handled this via the CONFLICTS property ...
>     * What happens if the admin attempts to install the genquery2 package on a 4.3.2 server?

The package manager won't allow it.

>     * What happens if genquery2 is installed on a 4.3.1 server and the admin upgrades to 4.3.2?

My understanding is that it depends on the package manager frontend being used. Some may require that the user manually uninstall the genquery2 package first, and some will automatically make removing the genquery2 package part of the same transaction.

korydraughn:
Ok. I'm leaning toward the packaging reporting a conflict unless there's a better way to do it. At the very least that gives the admin a chance to decide how they want to proceed.

Thoughts?

SwooshyCueb:
I think I'm in agreement.

korydraughn (approved):
Pound it.

SwooshyCueb:
\#'d

korydraughn:
I don't see the pound?

alanking (approved):

SwooshyCueb:
\#'d, for real this time

alanking:
There we go. Mergin

PR #7649, opened 2024-04-09
[_7319] Correct nullptr dereference in `msiSetDefaultResc(...)` (main)

MartinFlores751:
This PR fixes a bug in `msiSetDefaultResc(...)` where a `nullptr` may be dereferenced, causing the agent to crash.

### DIFF
--- a/scripts/irods/test/test_native_rule_engine_plugin.py
+++ b/scripts/irods/test/test_native_rule_engine_plugin.py
@@ -855,6 +855,30 @@ OUTPUT ruleExecOut
         finally:
             os.remove(rule_file)
 
+    @unittest.skipIf(plugin_name == 'irods_rule_engine_plugin-python', 'rule language only')
+    def test_msiSetDefaultResc__issue_7319(self):
+        rule_map = {
+            'irods_rule_engine_plugin-irods_rule_language': textwrap.dedent('''
+                test_msiSetDefaultResc {{
+                    msiSetDefaultResc('TestResc', 'preferred');
+                }}
+                OUTPUT ruleExecOut
+                ''')
+        }
+
+        rule_file = 'test_msiSetDefaultResc.r'
+        with open(rule_file, 'w') as f:
+            f.write(rule_map[self.plugin_name])
+
+        try:
+            rep_name = 'irods_rule_engine_plugin-irods_rule_language-instance'
+
+            self.admin.assert_icommand(['irule', '-r', rep_name, '-F', rule_file], desired_rc=0)
+            self.user0.assert_icommand(['irule', '-r', rep_name, '-F', rule_file], desired_rc=0)
+
+        finally:
+            os.remove(rule_file)
+
     def test_re_serialization_for_RsComm_ptr__issue_5950(self):
         pep_map = {
             'irods_rule_engine_plugin-irods_rule_language': textwrap.dedent('''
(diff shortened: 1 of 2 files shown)


scripts/irods/test/test_native_rule_engine_plugin.py
```diff
@@ -855,6 +855,30 @@ def test_msiSetResource__issue_7319(self):
...
+
+        rule_file = 'test_msiSetDefaultResc.r'
+        with open(rule_file, 'w') as f:
+            f.write(rule_map[self.plugin_name])
+
+        try:
+            rep_name = 'irods_rule_engine_plugin-irods_rule_language-instance'
+
+            self.admin.assert_icommand(['irule', '-r', rep_name, '-F', rule_file], desired_rc=0)
+            self.user0.assert_icommand(['irule', '-r', rep_name, '-F', rule_file], desired_rc=0)
```
alanking, line 877: `irods::set_default_resource` is supposed to return `USER_NO_RESC_INPUT_ERR` when `_cond_input` is `nullptr`:
https://github.com/irods/irods/blob/5916f2ff07fdffe32580bace51dd241b84699f4f/server/core/src/irods_resource_backport.cpp#L304-L307

This should then set `rei->status` to `USER_NO_RESC_INPUT_ERR`. Do we know why this isn't being returned? Can/Should the microservice return the error, or is that up to the implementation of the rule?

Keep in mind that the response before your fix was for the agent to crash, so returning an error and/or halting the rule's execution wouldn't necessarily be much of a change in behavior from the perspective of the calling rule :)

As things now stand, it appears that it must be failing silently? If there is some way to show that the `nullptr` check had some effect other than "the agent didn't crash", that could be helpful, I think.

alanking (commented):
Seems fine. I had one concern about the return value/failure behavior of the microservice.

scripts/irods/test/test_native_rule_engine_plugin.py
MartinFlores751: `_resc_list` would be set to `TestResc`, so it's not empty. Execution would continue, likely down to the following lines: https://github.com/irods/irods/blob/5916f2ff07fdffe32580bace51dd241b84699f4f/server/core/src/irods_resource_backport.cpp#L383-L415

I didn't run the debugger while the test was running, but running separately, it does show it reaches this point (note, the 'debug resource' may be different from that of the 'test resource').

Not sure how important `_cond_input` is, but there does seem to be some bonus info there: https://github.com/irods/irods/blob/5916f2ff07fdffe32580bace51dd241b84699f4f/server/core/src/irods_resource_backport.cpp#L321-L336

alanking: Ah, you're right. We don't need to worry about that then. Resolving

alanking (commented):
I think this looks good. Let's wait for one more set of eyeballs before we proceed

korydraughn:
Changes look fine. Squash it, no #'s

MartinFlores751:
Squashed.

alanking (approved):
\# it, I say

MartinFlores751:
It's got the '#' now.

alanking:
@MartinFlores751 - Excellent. Please cherry-pick to 4-3-stable.

Issue #7636, opened 2024-04-05
Change naming scheme used by feature test macros


korydraughn:
- [x] main
- [x] 4-3-stable

## Enhancement

This is primarily about the macros having names which resemble booleans.

The new naming scheme will be `IRODS_LIBRARY_FEATURE_<phrase>`. `IRODS_FEATURE_<phrase>` will be left as a reserved naming scheme in case the iRODS server needs more generic macro names.

Existing feature test macros will be left as is. Instead, new macros will be introduced as aliases for consistency.

See conversation starting here: https://github.com/irods/irods/pull/7582#discussion_r1532075368

PR #7661, opened 2024-04-09
[#7636] Change naming scheme used by feature test macros (main)

korydraughn:
The `IRODS_LIBRARY_FEATURE_` prefix is fine/good. That's the namespace for the macros.

I'm not a big fan of the `_LIBRARY_`, `_API_ENDPOINT_` component though.

Is there a better naming scheme for those?
Would a shorter name be better? i.e. `_LIB_`, `_API_`
Does it make since to continue namespacing the macros after the `IRODS_LIBRARY_FEATURE_` prefix?
Would the names be better if removed?

I think I can reuse the documentation for the original macros. Will investigate.


SwooshyCueb:
> I'm not a big fan of the `_LIBRARY_`, `_API_ENDPOINT_` component though.

I'm not a fan of `_LIBRARY_` either.

> Is there a better naming scheme for those? 
> Would a shorter name be better? i.e. `_LIB_`, `_API_` 

I like both of these. However...

> Does it make since to continue namespacing the macros after the `IRODS_LIBRARY_FEATURE_` prefix?
> Would the names be better if removed?

I think removing the namespacing/prefixing past `IRODS_LIBRARY_FEATURE_` makes sense. Since we're using numbers that can be incremented, I don't think there is as much need for that kind of granularity. Should we ever need to differentiate, that can be done on a case-by-case basis.

korydraughn:
Updated.

Need to render the docs to see how things look now.

lib/core/include/irods/library_features.h
```diff
@@ -21,42 +21,84 @@
...
+/// \sa Prefer IRODS_LIBRARY_FEATURE_LIB_TICKET_ADMINISTRATION
+///
 /// \since 4.3.1
-#define IRODS_HAS_LIBRARY_TICKET_ADMINISTRATION                              202209L
+#define IRODS_HAS_LIBRARY_TICKET_ADMINISTRATION                               202209L
+
+/// An alias for IRODS_HAS_LIBRARY_TICKET_ADMINISTRATION.
+///
+/// \since 4.3.2
+#define IRODS_LIBRARY_FEATURE_LIB_TICKET_ADMINISTRATION                       202209L
```
alanking, line 32: Would it make sense to have the alias macros defined by the other macro? In other words...
```c
#define IRODS_LIBRARY_FEATURE_LIB_TICKET_ADMINISTRATION                       202209L
#define IRODS_HAS_LIBRARY_TICKET_ADMINISTRATION                               IRODS_LIBRARY_FEATURE_LIB_TICKET_ADMINISTRATION
```
- OR -
```c
#define IRODS_HAS_LIBRARY_TICKET_ADMINISTRATION                               202209L
#define IRODS_LIBRARY_FEATURE_LIB_TICKET_ADMINISTRATION                       IRODS_HAS_LIBRARY_TICKET_ADMINISTRATION
```

alanking (commented):
Just had one minor suggestion

lib/core/include/irods/library_features.h
korydraughn: I considered that and decided to keep them separate in case they end up diverging (on purpose).

korydraughn: I'm going to make them aliases.

It just makes more sense from a maintainability perspective.

SwooshyCueb (commented):
Was there further discussion about namespacing/prefixing past `IRODS_LIBRARY_FEATURE_` that I missed? I see `_LIB_` and `_API_`, but the last discussion about it in this PR was @korydraughn agreeing with my statement that removing the added prefixing/namespacing makes sense.

korydraughn:
There wasn't really any additional conversation following ours. @alanking made a suggestion about macro aliasing.

Did I misinterpret what you meant in regard to `_LIB_` and `_API_`?

korydraughn:
After some discussion, the decision is to remove the internal namespacing (i.e. `_LIB_`, `_API_`) and only keep the `IRODS_LIBRARY_FEATURE_` prefix.

korydraughn:
Internal/Nested namespaces removed.

Running the unit test.

korydraughn:
The library features unit test passed.

Squashing now.

korydraughn:
Squashed.

SwooshyCueb (approved):

korydraughn:
#'d

Issue #7651, opened 2024-04-09
Make `msiSendMail` opt-in


korydraughn:
- [x] 4-3-stable
- [x] irods/irods_docs

## Deprecation

- Add private variable for controlling whether `msiSendMail` returns early
- Add new MSI for changing the value of the variable
    - Requires rodsadmin privileges 
    - Potential names for the MSI
       - `msi_use_deprecated_msiSendMail`
       - `msi_allow_deprecated_msiSendMail`
       - `msi_enable_deprecated_msiSendMail`
- `msiSendMail` checks private variable before sending any mail
    - Consider whether use of `msiSendMail` should reset the private variable

korydraughn:
The final solution resulted in the following:
- A new server_config.json option: `/advanced_settings/enable_deprecated_msiSendMail`
    - Set to `true` to enable the microservice
    - Set to `false` to disable it
    - If not present in the config file, the behavior is equivalent to setting the option to `false`
- `msiSendMail` is disabled by default _(due to the option not being defined)_
- `msiSendMail` will return `SYS_NOT_ALLOWED` when NOT enabled

Forgot to update the documentation.

PR #7662, opened 2024-04-09
[#7651] Make use of msiSendMail an opt-in (main)

korydraughn:
The design may be a little weird. That's due to different ideas on how to go about exposing this functionality.

I'm leaning towards the following:
- Making `msiSendMail` always reset the internal variable upon use
- Making the new MSI always set the internal variable to true
    - Removes the need for an input argument

Those changes disallow putting the agent in a state where anyone can invoke the MSI. I'm thinking about long-running agents here. It also forces all calls to `msiSendMail` to be addressed/reviewed.

Thoughts?

### DIFF
--- a/doxygen/microservices.cpp
+++ b/doxygen/microservices.cpp
@@ -114,7 +114,7 @@
   - #msiSubstr   - Returns a substring of the given string
 
  \subsection msiemail Email Microservices
-  - #msiSendMail   - (Deprecated) Sends email
+  - #msiSendMail - (Deprecated) Sends email
 
  \subsection msikv Key-Value (Attr-Value) Microservices
   - #writeKeyValPairs - Writes key-value pairs to stdout or stderr and with given separator
--- a/scripts/irods/test/test_all_rules.py
+++ b/scripts/irods/test/test_all_rules.py
@@ -265,6 +265,10 @@ class Test_AllRules(resource_suite.ResourceBase, unittest.TestCase):
                 # print("skipping " + rulefile + " ----- tested separately")
                 return False
 
+            # msiSendMail is deprecated and disabled by default as of 4.3.2.
+            if "rulemsiSendMail" in rulefile:
+                return False
+
             return True
 
         for rulefile in filter(filter_rulefiles, sorted(os.listdir(rulesdir))):
@@ -275,6 +279,26 @@ class Test_AllRules(resource_suite.ResourceBase, unittest.TestCase):
                                                      ,'STDOUT_SINGLELINE', "completed successfully")
                 return test
 
+            # Returns a test with the following name:
+            #
+            #     test_<filename>_r
+            #
+            # Dots (.) are replaced with underscores (_).
+            #
+            # The rule files which are used to generate the tests normally follow
+            # the naming scheme below (following the scheme is not required):
+            #
+            #     rule<microservice_name>.r.
+            #
+            # So, given a file named rulemsiSendMail.r, the line below will
+            # generate the following test name:
+            #
+            #     test_rulemsiSendMail_r
+            #
+            # Individual tests can be run using the generated name like so:
+            #
+            #     test_all_rules.Test_AllRules.test_<filename>_r
+            #
             yield 'test_' + rulefile.replace('.', '_'), make_test(rulefile)
 
     def test_rulemsiDataObjRsync(self):
--- a/scripts/irods/test/test_misc.py
+++ b/scripts/irods/test/test_misc.py
@@ -16,13 +16,14 @@ from ..configuration import IrodsConfig
 from ..controller import IrodsController
 from ..core_file import temporary_core_file
 
-class Test_Misc(session.make_sessions_mixin([('otherrods', 'rods')], []), unittest.TestCase):
+class Test_Misc(session.make_sessions_mixin([('otherrods', 'rods')], [('alice', 'apass')]), unittest.TestCase):
 
     plugin_name = IrodsConfig().default_rule_engine_plugin
 
     def setUp(self):
         super(Test_Misc, self).setUp()
         self.admin = self.admin_sessions[0]
+        self.user = self.user_sessions[0]
 
     def tearDown(self):
         super(Test_Misc, self).tearDown()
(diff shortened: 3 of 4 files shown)


alanking:
> - Making the new MSI always set the internal variable to true
>   - Removes the need for an input argument

I agree. No need for the input since it will be disabled by default and the microservice is only useful to enable it. If it wasn't disabled after using `msiSendMail` I might feel differently.

server/re/src/mailMS.cpp
```diff
@@ -70,8 +61,8 @@ int msi_enable_msiSendMail(MsParam* _value, RuleExecInfo* _rei)
  * \note This microservice sends e-mail using the mail command in the unix system. No attachments are supported. The
  *       sender of the e-mail is the unix user-id running the irodsServer.
  *
- * \warning This microservice must be enabled via msi_enable_msiSendMail first, otherwise the call to msiSendMail will
- * fail.
+ * \warning All invocations of this microservice must be preceded by a call to msi_enable_msiSendMail, otherwise the
+ * call to msiSendMail will fail.
```
trel, line 65: will it fail?  or just exit silently and 'early'?

trel: oh, i see, it's an actual error now.

server/re/src/mailMS.cpp
```diff
@@ -47,6 +87,16 @@
  **/
 int msiSendMail( msParam_t* xtoAddr, msParam_t* xsubjectLine, msParam_t* xbody, ruleExecInfo_t* ) {
 
+    if (!enable_msiSendMail) {
+        log_msi::error("{}: msiSendMail is disabled. To enable, call msi_enable_msiSendMail() first.", __func__);
+        return SYS_NOT_ALLOWED;
+    }
+
+    // Reset the option to force the admin to consider all locations where
+    // this microservice is used. This guarantees use is always narrowly scoped.
```
trel, line 96: does this preclude rodsusers from ever calling msiSendMail directly?    i think it does.

i also think i'm okay with that.

server/re/src/mailMS.cpp
korydraughn: Yep. If the MSI to enable use isn't called first, `msiSendMail` will report an error.

server/re/src/mailMS.cpp
korydraughn: Indirectly, yes.

Thinking about this more, the rodsadmin requirement means the MSIs can't be in policy triggered by non-admins. As soon as the enable-MSI is hit, an error will occur due to the lack of rodsadmin privileges.

Hmm.

trel: correct... by going admin-only ... we've also effectively 'broken' any existing deployments with no workaround available.

does this suggest we need a server-wide variable, that only admins can set (a kind of 'mode')... and it doesn't get 'reset' after each use?

korydraughn: That won't work either due to the enable-MSI being callable via clients. Crap.

korydraughn: @trel Yes, a server-wide option is the only other (easy) way I can see this working. An environment variable in server_config.json is pretty easy to do without requiring adjustments to JSON schemas.

I'd like to disallow non-admins from being able to call the MSI via `irule` though.

trel: i'm suggesting the enable-MSI be admin-only, and msiSendMail not reset the value.

trel: or a server_config.json variable, and msiSendMail depend on that.

korydraughn: > i'm suggesting the enable-MSI be admin-only, and msiSendMail not reset the value.

I doubt this is doable. How do you envision the admin enabling things?

> or a server_config.json variable, and msiSendMail depend on that.

I'd like to avoid this since that's a change in the schema. And the sendmail MSIs are deprecated.

trel: - no enable-MSI
- new advanced_setting for allow_msiSendMail
- server defaults to false
- msiSendMail check this value, exit early if false

korydraughn: No change to server_config.json.template since a schema bump isn't going to happen for this.

The server will default to false if the option isn't defined in server_config.json.

Done.

korydraughn:
Review comments addressed. Checking it in the testing environment now.

korydraughn:
I've seen success at the bench. Will add a test.

doxygen/microservices.cpp
```diff
@@ -112,7 +112,8 @@
   - #msiSubstr   - Returns a substring of the given string
 
  \subsection msiemail Email Microservices
-  - #msiSendMail   - (Deprecated) Sends email
+  - #msi_enable_msiSendMail - Enables the use of msiSendMail.
```
trel, line 115: remove - no msi anymore

server/re/src/mailMS.cpp
```diff
@@ -46,6 +56,24 @@
  * \sa none
  **/
 int msiSendMail( msParam_t* xtoAddr, msParam_t* xsubjectLine, msParam_t* xbody, ruleExecInfo_t* ) {
+    const auto enabled = [] {
+        try {
+            return irods::get_advanced_setting<bool>("enable_msiSendMail");
```
trel, line 61: what if it is set to 'apples'... will that set `enabled` to be true enough for line 71 below?

korydraughn: Setting the value to apples will cause an exception to be thrown due to a type mismatch.

Resulting in `enabled` being set to false.

trel, line 61: we had originally named this `enable_deprecated_msiSendMail`.      in case that wasn't an intentional choice to shorten it.

korydraughn: Will change.

doxygen/microservices.cpp
korydraughn: Done

server/re/src/mailMS.cpp
korydraughn: Done

korydraughn:
Test added.

Running core tests. I suspect one failure to occur due to `test/rules/rulemsiSendMail.r` in the icommands repo.

Will address that test after I see it fail.

scripts/irods/test/test_misc.py
```diff
@@ -471,3 +472,31 @@ def test_tcp_keepalive_probes__issue_2533_3824(self):
...
+    def test_enabling_and_disabling_of_msiSendMail__issue_7651(self):
+        config = IrodsConfig()
+        rep_name = f'{self.plugin_name}-instance'
+        rule = 'msiSendMail("x", "y", "z")'
+
+        # Show use of msiSendMail is blocked when the configuration option is not defined.
+        self.assertNotIn('enable_deprecated_msiSendMail', config.server_config['advanced_settings'])
+        self.user.assert_icommand(['irule', '-r', rep_name, rule, 'null', 'ruleExecOut'], 'STDERR', ['-169000 SYS_NOT_ALLOWED'])
+
+        # Show use of msiSendMail is blocked when the configuration option is set to false.
```
trel, line 488: please add another assertion when set to a non-True-or-False string, aka apples or bananas.

korydraughn: Done.

korydraughn:
The addition to the test passed at the bench.

Waiting for core tests to finish so I can address failures.

trel:
looking good.

alanking (commented):
Looks good. I say squash when ready. Will of course await the results of the tests

korydraughn:
All tests passed except `test_all_rules.py` (this is expected).

May add the `msiSendMail` rule file to the skip list or adjust the rule file in the icommands repo.

scripts/irods/test/test_all_rules.py
```diff
@@ -265,6 +265,10 @@ def filter_rulefiles(rulefile):
                 # print("skipping " + rulefile + " ----- tested separately")
                 return False
 
+            # msiSendMail is deprecated and disabled by default.
```
trel, line 268: could say 'since 4.3.2'

korydraughn: Done

korydraughn:
Rerunning test_all_rules.py.

korydraughn:
test_all_rules.py passed.

Squashing.

korydraughn:
Squashed.

trel (approved):

korydraughn:
Wait, we labelled this as being 4-3-stable only.

That means this PR should not be merged. Instead, we need to pound the 4-3-stable PR.

korydraughn:
Closing this PR since the MSI will be removed in iRODS 5.0.

See #7667.

PR #7668, opened 2024-04-10
[7413] Serialize function for structFileExtAndReg

FifthPotato:
Still feels drafty. Tried my hand at making something that could avoid having to unpack stuff element-by-element. Instead, it'll just try calling whatever's defined in `the_map` for each sub-type, and merging all of them back into the output `std::map` with a prefix. Potentially magical.
Tested by hand, and seems to behave mostly as expected.

Broken:
Trying to write the following into core.re:
```
*w = *STRUCTINP.collection;
msiModAVUMetadata('-C', '{self.user.session_collection}', 'add', '{avu_prefix}_1', *w, '');
```
...gives me an unsolvable typing error (`unit < (PATH)`, whatever that means), which is pretty baffling to me. The line it gets mad on is the first line.
I'm not sure if the NREP type solver is trying to automagically parse out a type, but then gets mad at something? Or maybe the name `collection` is reserved somehow?

### DIFF
--- a/scripts/irods/test/test_dynamic_peps.py
+++ b/scripts/irods/test/test_dynamic_peps.py
@@ -497,7 +497,6 @@ class Test_Dynamic_PEPs(session.make_sessions_mixin([('otherrods', 'rods')], [('
     def test_if_ExecMyRuleInp_is_exposed__issue_7552(self):
         with temporary_core_file() as core:
             avu_prefix = 'test_if_ExecMyRuleInp_is_exposed__issue_7552'
-
             # Add a rule to core.re which when triggered will add AVUs to the user's
             # session collection.
             core.add_rule(dedent('''
@@ -548,3 +547,73 @@ class Test_Dynamic_PEPs(session.make_sessions_mixin([('otherrods', 'rods')], [('
                 self.user.run_icommand(['imeta', 'rm', '-C', self.user.session_collection, f'{avu_prefix}_4', "0"])
                 self.user.run_icommand(['imeta', 'rm', '-C', self.user.session_collection, f'{avu_prefix}_5', "@external rule {{ *C=*A++*B }}"])
                 self.user.run_icommand(['imeta', 'rm', '-C', self.user.session_collection, f'{avu_prefix}_6', "*C="])
+
+    @unittest.skipIf(plugin_name != 'irods_rule_engine_plugin-irods_rule_language' or test.settings.RUN_IN_TOPOLOGY, "Requires NREP and single node zone.")
+    def test_if_StructFileExtAndRegInp_is_exposed__issue_7413(self):
+        with temporary_core_file() as core:
+            avu_prefix = 'test_if_StructFileExtAndRegInp_is_exposed__issue_7413'
+
+            # Add a rule to core.re which when triggered will add AVUs to the user's
+            # session collection.
+            core.add_rule(dedent('''
+            pep_api_struct_file_ext_and_reg_post(*INSTANCE, *COMM, *STRUCTINP) {{
+                *v = *STRUCTINP.obj_path;
+                msiModAVUMetadata('-C', '{self.user.session_collection}', 'add', '{avu_prefix}_0', *v, '');
+
+                *v = *STRUCTINP.collection_path;
+                msiModAVUMetadata('-C', '{self.user.session_collection}', 'add', '{avu_prefix}_1', *v, '');
+
+                *v = *STRUCTINP.opr_type;
+                msiModAVUMetadata('-C', '{self.user.session_collection}', 'add', '{avu_prefix}_2', *v, '');
+
+                *v = *STRUCTINP.flags;
+                msiModAVUMetadata('-C', '{self.user.session_collection}', 'add', '{avu_prefix}_3', *v, '');
+
+                *v = *STRUCTINP.defRescName;
+                msiModAVUMetadata('-C', '{self.user.session_collection}', 'add', '{avu_prefix}_4', *v, '');
+
+                if('' == *STRUCTINP.forceFlag) {{
+                    msiModAVUMetadata('-C', '{self.user.session_collection}', 'add', '{avu_prefix}_5', 'forceFlagcheck', '');
+                }}
+            }}
+            '''.format(**locals())))
+
+            TEST_COLLECTION_NAME = "%s/%s" % (self.user.session_collection, 'testcoll')
+            TEST_TARFILE_NAME = "%s/%s" % (self.user.session_collection, 'test.tar')
+
+            try:
+                self.user.assert_icommand(['imkdir', TEST_COLLECTION_NAME], 'STDOUT')
+                # Make some test files...
+                for x in range(0, 5):
+                    self.user.assert_icommand(['itouch', "%s/file%s" % (TEST_COLLECTION_NAME, x)])
+
+                self.user.assert_icommand(['ibun', '-cDtar', "%s/%s" % (self.user.session_collection, 'test.tar'), TEST_COLLECTION_NAME])
+
+                self.user.assert_icommand(['imkdir', TEST_COLLECTION_NAME + "_backup"], 'STDOUT')
+
+                # Should trigger PEP
+                self.user.assert_icommand(['ibun', '-x', '-f', TEST_TARFILE_NAME, TEST_COLLECTION_NAME + "_backup"])
+
+                # Show the session collection has the expected AVUs attached to it.
+                self.user.assert_icommand(['imeta', 'ls', '-C', self.user.session_collection], 'STDOUT', [
+                    f'attribute: {avu_prefix}_0\nvalue: {TEST_TARFILE_NAME}\n',
+                    f'attribute: {avu_prefix}_1\nvalue: {TEST_COLLECTION_NAME + "_backup"}\n',
+                    f'attribute: {avu_prefix}_2\nvalue: 0\n',
+                    f'attribute: {avu_prefix}_3\nvalue: 0\n',
+                    f'attribute: {avu_prefix}_4\nvalue: {self.user.default_resource}\n',
+                    f'attribute: {avu_prefix}_5\nvalue: forceFlagcheck\n',
+                ])
+
+            finally:
+                # Remove the AVUs
+                self.user.run_icommand(['imeta', 'rm', '-C', self.user.session_collection, f'{avu_prefix}_0', TEST_TARFILE_NAME])
+                self.user.run_icommand(['imeta', 'rm', '-C', self.user.session_collection, f'{avu_prefix}_1', TEST_COLLECTION_NAME])
+                self.user.run_icommand(['imeta', 'rm', '-C', self.user.session_collection, f'{avu_prefix}_2', '0'])
+                self.user.run_icommand(['imeta', 'rm', '-C', self.user.session_collection, f'{avu_prefix}_3', '0'])
+                self.user.run_icommand(['imeta', 'rm', '-C', self.user.session_collection, f'{avu_prefix}_4', self.user.default_resource])
+                self.user.run_icommand(['imeta', 'rm', '-C', self.user.session_collection, f'{avu_prefix}_5', 'forceFlagcheck'])
+
+                # Cleanup
+                self.user.run_icommand(['irm', '-rf', TEST_COLLECTION_NAME])
+                self.user.run_icommand(['irm', '-rf', TEST_COLLECTION_NAME + "_backup"])
+                self.user.run_icommand(['irm', '-f', TEST_TARFILE_NAME])
--- a/server/re/src/irods_re_serialization.cpp
+++ b/server/re/src/irods_re_serialization.cpp
@@ -1296,6 +1296,29 @@ namespace irods::re_serialization
         }
     } // serialize_execMyRuleInp_ptr
 
+    static irods::error serialize_structFileExtAndRegInp_ptr(boost::any _p, serialized_parameter_t& _out)
+    {
+        try {
+            log_re::trace(__func__);
+            const auto* casted_ptr = boost::any_cast<structFileExtAndRegInp_t*>(_p);
+            _out["obj_path"] = casted_ptr->objPath;
+            _out["collection_path"] = casted_ptr->collection;
+            _out["opr_type"] = std::to_string(casted_ptr->oprType);
+            _out["flags"] = std::to_string(casted_ptr->flags);
+            serialize_keyValPair(casted_ptr->condInput, _out);
+            return SUCCESS();
+        }
+        catch (const boost::bad_any_cast& e) {
+            return ERROR(
+                INVALID_ANY_CAST,
+                fmt::format("{}: failed to cast pointer to [structFileExtAndRegInp_t*]: {}", __func__, e.what()));
+        }
+        catch (const std::exception& e) {
+            return ERROR(SYS_LIBRARY_ERROR,
+                         fmt::format("{}: failed to serialize [structFileExtAndRegInp_t*]: {}", __func__, e.what()));
+        }
+    }
+
 #if 0
     static error serialize_XXXX_ptr(
             boost::any               _p,
@@ -1353,7 +1376,8 @@ namespace irods::re_serialization
             {std::type_index(typeid(Genquery2Input*)), serialize_Genquery2Input_ptr},
             {std::type_index(typeid(const std::vector<std::string>*)), serialize_const_vector_of_strings_ptr},
             {std::type_index(typeid(std::vector<std::string>*)), serialize_vector_of_strings_ptr},
-            {std::type_index(typeid(execMyRuleInp_t*)), serialize_execMyRuleInp_ptr}};
+            {std::type_index(typeid(execMyRuleInp_t*)), serialize_execMyRuleInp_ptr},
+            {std::type_index(typeid(structFileExtAndRegInp_t*)), serialize_structFileExtAndRegInp_ptr}};
         return the_map;
 
     } // get_serialization_map


korydraughn (commented):
What's the motivation behind this particular implementation?

Why not just serialize the individual members of `StructFileExtAndRegInp`?
https://github.com/irods/irods/blob/81871db9123316b9ad16576234c70666fe064eef/lib/api/include/irods/structFileExtAndReg.h#L7-L13

There's a lot happening, but it's not clear to me why.

server/re/src/irods_re_serialization.cpp
```diff
@@ -35,6 +40,21 @@ namespace irods::re_serialization
         }
     }
 
+    static error serialize_keyValPair(boost::any _p, serialized_parameter_t& _out)
+    {
+        try {
+            const keyValPair_t& kvp = boost::any_cast<keyValPair_t>(_p);
+            serialize_keyValPair(kvp, _out);
```
trel, line 47: c++ noob here... how could this work?
if it succeeded in casting to a keyValPair... then we call this same function again on the result?

server/re/src/irods_re_serialization.cpp
```diff
@@ -35,6 +40,21 @@ namespace irods::re_serialization
         }
     }
 
+    static error serialize_keyValPair(boost::any _p, serialized_parameter_t& _out)
```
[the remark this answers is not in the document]
korydraughn: Ignore this question.

server/re/src/irods_re_serialization.cpp
korydraughn: I'll give you a hint.

Look at the function signature for the other `keyValPair_t` function. It's above this one.

trel: it's a ref!

as you were.

korydraughn: The compiler will pick the function overload which best matches the types involved.

FifthPotato:
> What's the motivation behind this particular implementation?
> 
> Why not just serialize the individual members of `StructFileExtAndRegInp`?
> 
> https://github.com/irods/irods/blob/81871db9123316b9ad16576234c70666fe064eef/lib/api/include/irods/structFileExtAndReg.h#L7-L13
> 
> There's a lot happening, but it's not clear to me why.

I guess it's an effort to try and reduce the amount of copy-paste/boilerplate in the implementation. I noticed a lot of the serialize functions effectively boiled down to some form of "put each field-and-subfield of thing in output serialized_parameter_t"
I thought since `the_map` already defines how to serialize a bunch of base types, why not re-use that to serialize types that are effectively just collections of leaf types w/ a prefix? That way, future serialize function implementations would only need to focus on leaf types, and serializing of the top-level object would be mostly abstracted away with this.
If you think it's a dumb/bad idea, of course, I could do the other implementation.

korydraughn:
I see. The goal is sound, but getting 4.3.2 released for UGM is top priority.

Let's save what you have in a separate branch and consider it for a later release. UGM is approaching quickly and there's still a lot to do so let's resolve the serialization issues by using the simpler approach.

FifthPotato:
Force-pushed to the simpler implementation...the core.re thing is still broken, though.

server/re/src/irods_re_serialization.cpp
```diff
@@ -1199,6 +1199,29 @@ namespace irods::re_serialization
         }
     } // serialize_vector_of_strings_ptr
 
+    static irods::error serialize_structFileExtAndRegInp_ptr(boost::any _p, serialized_parameter_t& _out)
+    {
+        try {
+            log_re::trace(__func__);
+            auto* casted_ptr = boost::any_cast<structFileExtAndRegInp_t*>(_p);
```
korydraughn, line 1206: `casted_ptr` can be marked `const` too.

scripts/irods/test/test_dynamic_peps.py
```diff
@@ -492,3 +492,59 @@ def test_if_Genquery2Input_is_exposed__issue_7570(self):
...
+    @unittest.skipIf(plugin_name != 'irods_rule_engine_plugin-irods_rule_language' or test.settings.RUN_IN_TOPOLOGY, "Requires NREP and single node zone.")
+    def test_if_StructFileExtAndRegInp_is_exposed__issue_7413(self):
+        with temporary_core_file() as core:
+            avu_prefix = 'test_if_StructFileExtAndRegInp_is_exposed__issue_7413'
+
+            # Add a rule to core.re which when triggered will add AVUs to the user's
+            # session collection.
+            core.add_rule(dedent('''
+            pep_api_struct_file_ext_and_reg_post(*INSTANCE, *COMM, *STRUCTINP) {{
+                writeLine("serverLog",*STRUCTINP);
```
korydraughn, line 505: Remove?

scripts/irods/test/test_dynamic_peps.py
```diff
@@ -492,3 +492,59 @@ def test_if_Genquery2Input_is_exposed__issue_7570(self):
...
+                *v = *STRUCTINP.objPath;
+                msiModAVUMetadata('-C', '{self.user.session_collection}', 'add', '{avu_prefix}_0', *v, '');
+
+                *v = *STRUCTINP.collection;
+                msiModAVUMetadata('-C', '{self.user.session_collection}', 'add', '{avu_prefix}_1', *v, '');
+
+                *w = *STRUCTINP.oprType;
+                msiModAVUMetadata('-C', '{self.user.session_collection}', 'add', '{avu_prefix}_2', *w, '');
```
korydraughn, line 513: No `.flags` or keyValPair_t AVUs?

scripts/irods/test/test_dynamic_peps.py
```diff
@@ -492,3 +492,59 @@ def test_if_Genquery2Input_is_exposed__issue_7570(self):
...
+            '''.format(**locals())))
+
+            TEST_COLLECTION_NAME = "%s/%s" % (self.user.session_collection, 'testcoll')
+            TEST_TARFILE_NAME = "%s/%s" % (self.user.session_collection, 'test.tar')
+
+            try:
+                self.user.assert_icommand(['imkdir', TEST_COLLECTION_NAME], 'STDOUT')
+                # Make some test files...
+                for x in range(0, 5):
+                    self.user.assert_icommand(['itouch', "%s/file%s" % (TEST_COLLECTION_NAME, x)], 'STDOUT')
```
korydraughn, line 524: `STDOUT` can be removed since `itouch` doesn't produce any output on success.

scripts/irods/test/test_dynamic_peps.py
```diff
@@ -492,3 +492,59 @@ def test_if_Genquery2Input_is_exposed__issue_7570(self):
...
+
+                self.user.assert_icommand(['ibun', '-cDtar', "%s/%s" % (self.user.session_collection, 'test.tar'), TEST_COLLECTION_NAME], 'STDOUT')
+
+                self.user.assert_icommand(['imkdir', TEST_COLLECTION_NAME + "_backup"], 'STDOUT')
```
korydraughn, line 528: Same here. If the icommand doesn't produce any output on success, `STDOUT` isn't needed.

korydraughn (commented):
Looks good.

scripts/irods/test/test_dynamic_peps.py
```diff
@@ -492,3 +492,59 @@ def test_if_Genquery2Input_is_exposed__issue_7570(self):
...
+            finally:
+                # Remove the AVUs
+                self.user.run_icommand(['imeta', 'rm', '-C', self.user.session_collection, f'{avu_prefix}_0', TEST_TARFILE_NAME])
+                self.user.run_icommand(['imeta', 'rm', '-C', self.user.session_collection, f'{avu_prefix}_1', TEST_COLLECTION_NAME])
+                self.user.run_icommand(['imeta', 'rm', '-C', self.user.session_collection, f'{avu_prefix}_2', '0'])
+
+                # Cleanup
+                self.user.run_icommand(['irm', '-r', TEST_COLLECTION_NAME])
+                self.user.run_icommand(['irm', '-r', TEST_COLLECTION_NAME + "_backup"])
+                self.user.run_icommand(['irm', TEST_TARFILE_NAME])
```
alanking, line 550: Any reason to not use the force flag here?

alanking (commented):
Yep, seems fine overall. Please mark as ready for review if/when true.

scripts/irods/test/test_dynamic_peps.py
FifthPotato: Did `.flags` and one of the KVPs-- is that ok, or should I try to hit both KVPs?

trel: Looks like you've exercised all five parts of the struct.   I think you're good to go.

What is the 'other' KVP you're referencing in your comment?

scripts/irods/test/test_dynamic_peps.py
```diff
@@ -492,3 +492,67 @@ def test_if_Genquery2Input_is_exposed__issue_7570(self):
...
+                # Make some test files...
+                for x in range(0, 5):
+                    self.user.assert_icommand(['itouch', "%s/file%s" % (TEST_COLLECTION_NAME, x)])
+
+                self.user.assert_icommand(['ibun', '-cDtar', "%s/%s" % (self.user.session_collection, 'test.tar'), TEST_COLLECTION_NAME])
+
+                self.user.assert_icommand(['imkdir', TEST_COLLECTION_NAME + "_backup"], 'STDOUT')
+
+                # Should trigger PEP
+                self.user.run_icommand(['ibun', '-x', TEST_TARFILE_NAME, TEST_COLLECTION_NAME + "_backup"])
```
korydraughn, line 536: `run_icommand` -> `assert_icommand`

korydraughn (commented):
Looking good. Does the test pass now?

scripts/irods/test/test_dynamic_peps.py
FifthPotato: I'm not sure why, but this command returns nonzero. When we were testing this in discush last week, we saw a nonzero return despite it still working as expected.
When this line is as-is, it passes. Otherwise, I think it errors.

korydraughn: Hmm. We definitely have tests around `ibun -x`.

Have you tried running `ibun -x` manually and seeing what is returned by `$?`?

scripts/irods/test/test_dynamic_peps.py
korydraughn: I agree. The test looks complete to me.

I second @trel's question.

FifthPotato: Sorry, I meant another element in the KVP. If just one is fine, then I think we're good.

server/re/src/irods_re_serialization.cpp
```diff
@@ -1199,6 +1199,29 @@ namespace irods::re_serialization
...
     } // serialize_vector_of_strings_ptr
 
+    static irods::error serialize_structFileExtAndRegInp_ptr(boost::any _p, serialized_parameter_t& _out)
+    {
+        try {
+            log_re::trace(__func__);
+            const auto* casted_ptr = boost::any_cast<structFileExtAndRegInp_t*>(_p);
+            _out["objPath"] = casted_ptr->objPath;
+            _out["collection_path"] = casted_ptr->collection;
+            _out["oprType"] = boost::lexical_cast<std::string>(casted_ptr->oprType);
```
korydraughn, line 1209: Two of these need to be changed slightly to guard against overwrites (see rodsKeyWdDef.h).

- `objPath` -> `obj_path`
- `oprType` -> `opr_type`

korydraughn (commented):
Almost there. Just one final push and it'll be ready.

scripts/irods/test/test_dynamic_peps.py
korydraughn: Let's include one more element in the KVP so we're absolutely sure things are working.

If it's not clear which one to include, try adding `-f` to the `ibun` call and seeing if the force flag appears in the KVP. The force flag should appear as `"forceFlag"`.

https://github.com/irods/irods/blob/f51b33a1f83ac612536e2b6093209367f79dd2ed/lib/core/include/irods/rodsKeyWdDef.h#L11

Which is used in bunUtil.cpp _(i.e. the workhorse of `ibun`)_.

https://github.com/irods/irods/blob/f51b33a1f83ac612536e2b6093209367f79dd2ed/lib/core/src/bunUtil.cpp#L153-L155

scripts/irods/test/test_dynamic_peps.py
alanking: Right, we should confirm that `ibun -x` behaves the same way before and after this change. I don't remember seeing that in our discussion (sorry) but that doesn't seem like what we want. If that is how `ibun -x` behaves before these changes, then we should make a separate issue to look into that later.

server/re/src/irods_re_serialization.cpp
```diff
@@ -1199,6 +1199,29 @@ namespace irods::re_serialization
...
+            _out["flags"] = boost::lexical_cast<std::string>(casted_ptr->flags);
```
alanking, line 1210: clang-tidy says `std::to_string` is preferred to `boost::lexical_cast<std::string>`. I think I agree...

trel: if std works, seems good.

scripts/irods/test/test_dynamic_peps.py
FifthPotato: Huh, false alarm. Just re-ran it with assert and it seems ok now?

alanking: Well, that's good. It shall come out in testing in the run up to release if not. Hooray

scripts/irods/test/test_dynamic_peps.py
```diff
@@ -502,20 +502,22 @@ def test_if_StructFileExtAndRegInp_is_exposed__issue_7413(self):
...
+                *v = *STRUCTINP.opr_type;
                 msiModAVUMetadata('-C', '{self.user.session_collection}', 'add', '{avu_prefix}_2', *v, '');
 
                 *v = *STRUCTINP.flags;
                 msiModAVUMetadata('-C', '{self.user.session_collection}', 'add', '{avu_prefix}_3', *v, '');
 
                 *v = *STRUCTINP.defRescName;
                 msiModAVUMetadata('-C', '{self.user.session_collection}', 'add', '{avu_prefix}_4', *v, '');
+
+                *v = *STRUCTINP.forceFlag;
```
alanking, line 520: Needs an `msiModAVUMetadata` call after this, I think

scripts/irods/test/test_dynamic_peps.py
```diff
@@ -533,7 +535,7 @@ def test_if_StructFileExtAndRegInp_is_exposed__issue_7413(self):
                 self.user.assert_icommand(['imkdir', TEST_COLLECTION_NAME + "_backup"], 'STDOUT')
 
                 # Should trigger PEP
-                self.user.assert_icommand(['ibun', '-x', TEST_TARFILE_NAME, TEST_COLLECTION_NAME + "_backup"])
+                self.user.assert_icommand(['ibun', '-x', '-f', TEST_TARFILE_NAME, TEST_COLLECTION_NAME + "_backup"])
```
alanking, line 538: Just to confirm, the `-f` is to exercise the `forceFlag`, right?

alanking (commented):
Closer...

scripts/irods/test/test_dynamic_peps.py
FifthPotato: Yes!

scripts/irods/test/test_dynamic_peps.py
FifthPotato: In my testing, `forceFlag` didn't actually get set to a value; it was just there or not there. I thought the additional AVU would be unnecessary because we're just checking if `forceFlag` exists and not its value-- if it didn't, then this rule would error out with `UNMATCHED_KEY_OR_INDEX` and fail the assertion later.

alanking: Oh, I see. The value is empty (because we use the force keyword as a flag rather than a key-value pair) so it can't annotate the empty value via `msiModAVUMetadata`. Is that right? Let's leave a comment saying something along those lines for future readers.

scripts/irods/test/test_dynamic_peps.py
alanking: Okey dokey. Resolving

scripts/irods/test/test_dynamic_peps.py
korydraughn: I believe you can test for the existence of `forceFlag`. If it's there, set a custom AVU. Then the rest of the test checks to see if the custom AVU is there.

FifthPotato: > I believe you can test for the existence of `forceFlag`. If it's there, set a custom AVU. Then the rest of the test checks to see if the custom AVU is there.

What would the NREP syntax for this be?
```
if(*STRUCTINP.forceFlag) {
...
}
```
or such?

alanking: Seems reasonable to me

korydraughn: You can check for an empty string.
```
if ('' == *STRUCTINP.forceFlag) {
```

alanking (commented):
Excellent. Looks like all the review conversations have been resolved. Let's go ahead and rebase to make sure all the conflicts get worked out okay.

FifthPotato:
Fought through the merge conflicts...should've worked, I think?

korydraughn:
I don't see any conflicts now. Looks like it worked.

Review the changes and make sure everything is as it should be.

FifthPotato:
Couldn't see anything out of place. Ready for testing, then?

korydraughn:
Sure. I can run the tests for this as well. I'm half way through them for the other PR.

If there's other PRs you need to focus on, please do so.

korydraughn:
All unit tests passed.

Running core tests.

korydraughn:
All core tests passed.

Please rebase and squash, no #'s.

FifthPotato:
Rebased, squished.

korydraughn (approved):
Pound it.

korydraughn:
Cherry-pick to 4-3-stable.

FifthPotato:
Octothorpe'd

korydraughn:
Merging.

PR #7665, opened 2024-04-10
[#7590] irods_setup: Prevent local resource creation when user says no (main)

alanking:
Addresses #7590 

Tested manually on Debian 12.

Issue #7590, opened 2024-03-20
Setup iRODS Consumer with no local resource


0x010C:
- [x] main
- [x] 4-3-stable

---

### iRODS Version, OS and Version
iRODS 4.3.1 on Debian 12

### What did you try to do?
I wanted to setup a new iRODS Consummer without any local resource. So after installing irods-server package, I run the setup_irods.py script, choosing "no" when asked for a "Local storage on this server" and giving my already existing resource for the "Default resource".
```
$ sudo /var/lib/irods/scripts/setup_irods.py
iRODS user [irods]:
iRODS group [irods]:
iRODS server's role:
1. provider
2. consumer
Please select a number or choose 0 to enter a new value [1]: 2
Local storage on this server [yes]: no
Default resource [demoResc]: mainResc
...
```

### Expected behavior
I would have expected the script to setup an iRODS server configured without a local resource.

### Observed behavior (including steps to reproduce, if applicable)
```
Traceback (most recent call last):
  File "/var/lib/irods/scripts/setup_irods.py", line 537, in <module>
    sys.exit(main())
             ^^^^^^
  File "/var/lib/irods/scripts/setup_irods.py", line 523, in main
    setup_server(irods_config,
  File "/var/lib/irods/scripts/setup_irods.py", line 141, in setup_server
    irods.lib.execute_command(['iadmin', 'mkresc', default_resource_name, 'unixfilesystem', ':'.join([irods.lib.get_hostname(), default_resource_directory]), ''])
                                                                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: sequence item 1: expected str instance, NoneType found
```
Note that using the non-interactive --json_configuration_file option of setup_irods.py, if the "default_resource_directory" field is left empty the result is the same.

trel:
Hmm, interesting - that definitely should work.  We'll work to reproduce this ASAP.

alanking:
I was able to reproduce this. I'll post a fix here so that you can modify the setup script in order to proceed. It will of course be included in 4.3.2 as well.

alanking:
We just need to check for `NoneType` on `default_resource_directory` before attempting to create local storage. This is how `setup_storage()` returns the fact that the user answered "no" (or... rather, anything other than "yes" or "y", case-insensitive) to the question about "Local storage on this server". Here's the diff that fixed it for me:
```diff
[...]

PR #7669, opened 2024-04-11
[#7590] setup_irods: Prevent local resource creation when user says no (4-3-stable)

alanking:
Addresses #7590 

Cherry-picked from https://github.com/irods/irods/pull/7665


trel:
ship it

Issue #7576, opened 2024-03-08
proxied groupadmin cannot make groups


luijs:
- [x] main
- [x] 4-3-stable

---

## Bug Report

Not sure if this is a feature request or a bug report, but a groupadmin cannot create groups(didtn scheck users yet), while via icommadns a groupadmin can create users.  The same command does work with a rodsadmin user.

### iRODS Version, OS and Version
irods 4.3.0
ubuntu 20.04
http-api 0.2.0

### What did you try to do?
Create a group via the api with a groupadmin:

``` 
curl https://our.dns.nl/irods-http-api/0.2.0/users-groups -H 'Authorization: Bearer sometoken'   --data-urlencode 'op=create_group' --data-urlencode 'name=BINSQL'
```

### Expected behavior
Group is created

### Observed behavior (including steps to reproduce, if applicable)
No priviledge to create group

logging in irods:
```
 {"acl_operation":"{\"acl\":\"read\",\"entity_name\":\"BINSQL\"}","log_category":"api","log_facility":"local0","log_level":"error","log_message":"iRODS Exception:\n    file: /irods_source/
plugins/api/src/atomic_apply_acl_operations.cpp\n    function: void (anonymous namespace)::throw_if_invalid_entity_id((anonymous namespace)::id_type)\n    line: 202\n    code: -130000 (SYS
_INVALID_INPUT_PARAM)\n    message:\n        Invalid entity\nstack trace:\n--------------\n 0# irods::stacktrace::dump() const in /lib/libirods_common.so.4.3.0\n 1# irods::exception::assem
ble_full_display_what() const in /lib/libirods_common.so.4.3.0\n 2# irods::exception::what() const in /lib/libirods_common.so.4.3.0\n 3# 0x00007F7A8BEEDC7C in /usr/lib/irods/plugins/api/li
birods_atomic_apply_acl_operations_server.so\n 4# 0x00007F7A8BEEC658 in /usr/lib/irods/plugins/api/libirods_atomic_apply_acl_operations_server.so\n 5# irods::experimental::catalog::execute
_transaction(nanodbc::connection&, std::__1::function<int (nanodbc::transaction&)>) in /lib/libirods_server.so.4.3.0\n 6# 0x00007F7A8BEEB000 in /usr/lib/irods/plugins/api/libirods_atomic_a
[...]

PR #7688, opened 2024-04-12
[#7576] Allow proxied groupadmin to make groups (main)

MartinFlores751:
This PR should allow groupadmins to make groups as a proxied user.

When discussing with @korydraughn yesterday, there may be some merit, either in this PR or a separate, to testing all the groupadmin features as a proxied user to ensure we are using the appropriate endpoints.

### DIFF
--- a/lib/administration/user/src/user_administration.cpp
+++ b/lib/administration/user/src/user_administration.cpp
@@ -54,6 +54,22 @@ namespace irods::experimental::administration::NAMESPACE_IMPL
             zone = get_local_zone(_comm);
         }
 
+        const auto current_user{user{_comm.clientUser.userName, _comm.clientUser.rodsZone}};
+        const auto user_type{type(_comm, current_user)};
+
+        if (user_type == irods::experimental::administration::user_type::groupadmin) {
+            // clang-format off
+            userAdminInp_t input{.arg0 = "mkuser",
+                                 .arg1 = const_cast<char*>(name.data()),
+                                 .arg3 = const_cast<char*>(zone.data())};
+            // clang-format on
+            if (const auto ec{rxUserAdmin(&_comm, &input)}; ec != 0) {
+                // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+                THROW(ec, fmt::format("Failed to add user [{}].", name));
+            }
+            return;
+        }
+
         GeneralAdminInput input{};
         input.arg0 = "add";
         input.arg1 = "user";
@@ -86,6 +102,22 @@ namespace irods::experimental::administration::NAMESPACE_IMPL
     auto add_group(RxComm& _comm, const group& _group) -> void
     {
         const auto zone = get_local_zone(_comm);
+
+        const auto current_user{user{_comm.clientUser.userName, _comm.clientUser.rodsZone}};
+        const auto user_type{type(_comm, current_user)};
+
+        if (user_type == irods::experimental::administration::user_type::groupadmin) {
+            userAdminInp_t input{.arg0 = "mkgroup",
+                                 .arg1 = const_cast<char*>(_group.name.data()),
+                                 .arg2 = "rodsgroup",
+                                 .arg3 = const_cast<char*>(zone.data())};
+
+            if (const auto ec{rxUserAdmin(&_comm, &input)}; ec != 0) {
+                // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+                THROW(ec, fmt::format("Failed to add group [{}].", _group.name));
+            }
+            return;
+        }
 
         GeneralAdminInput input{};
         input.arg0 = "add";
@@ -122,6 +154,25 @@ namespace irods::experimental::administration::NAMESPACE_IMPL
 
     auto add_user_to_group(RxComm& _comm, const group& _group, const user& _user) -> void
     {
+        const auto name{local_unique_name(_comm, _user)};
+        const auto current_user{user{_comm.clientUser.userName, _comm.clientUser.rodsZone}};
+        const auto user_type{type(_comm, current_user)};
+
+        if (user_type == irods::experimental::administration::user_type::groupadmin) {
+            userAdminInp_t input{.arg0 = "modify",
+                                 .arg1 = "group",
+                                 .arg2 = const_cast<char*>(_group.name.data()),
+                                 .arg3 = "add",
+                                 .arg4 = const_cast<char*>(_user.name.data()),
+                                 .arg5 = const_cast<char*>(_user.zone.data())};
+
+            if (const auto ec{rxUserAdmin(&_comm, &input)}; ec != 0) {
+                // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+                THROW(ec, fmt::format("Failed to add user [{}] to group [{}].", name, _group.name));
+            }
+            return;
+        }
+
         GeneralAdminInput input{};
         input.arg0 = "modify";
         input.arg1 = "group";
@@ -131,7 +182,6 @@ namespace irods::experimental::administration::NAMESPACE_IMPL
         input.arg5 = _user.zone.data();
 
         if (const auto ec = rxGeneralAdmin(&_comm, &input); ec != 0) {
-            const auto name = local_unique_name(_comm, _user);
             // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
             THROW(ec, fmt::format("Failed to add user [{}] to group [{}].", name, _group.name));
         }
@@ -139,6 +189,25 @@ namespace irods::experimental::administration::NAMESPACE_IMPL
 
     auto remove_user_from_group(RxComm& _comm, const group& _group, const user& _user) -> void
     {
+        const auto name{local_unique_name(_comm, _user)};
+        const auto current_user{user{_comm.clientUser.userName, _comm.clientUser.rodsZone}};
+        const auto user_type{type(_comm, current_user)};
+
+        if (user_type == irods::experimental::administration::user_type::groupadmin) {
+            userAdminInp_t input{.arg0 = "modify",
+                                 .arg1 = "group",
+                                 .arg2 = const_cast<char*>(_group.name.data()),
+                                 .arg3 = "remove",
+                                 .arg4 = const_cast<char*>(_user.name.data()),
+                                 .arg5 = const_cast<char*>(_user.zone.data())};
+
+            if (const auto ec{rxUserAdmin(&_comm, &input)}; ec != 0) {
+                // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+                THROW(ec, fmt::format("Failed to remove user [{}] from group [{}].", name, _group.name));
+            }
+            return;
+        }
+
         GeneralAdminInput input{};
         input.arg0 = "modify";
         input.arg1 = "group";
@@ -148,7 +217,6 @@ namespace irods::experimental::administration::NAMESPACE_IMPL
         input.arg5 = _user.zone.data();
 
         if (const auto ec = rxGeneralAdmin(&_comm, &input); ec != 0) {
-            const auto name = local_unique_name(_comm, _user);
             // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
             THROW(ec, fmt::format("Failed to remove user [{}] from group [{}].", name, _group.name));
         }
--- a/unit_tests/src/test_user_administration.cpp
+++ b/unit_tests/src/test_user_administration.cpp
@@ -18,6 +18,12 @@ TEST_CASE("user group administration")
 
     irods::experimental::client_connection conn;
 
+    // Create a shared, generic test user
+    adm::user generic_user{"lagging-train"};
+    REQUIRE_NOTHROW(adm::client::add_user(conn, generic_user));
+    irods::at_scope_exit generic_user_cleanup{
+        [&conn, &generic_user] { CHECK_NOTHROW(adm::client::remove_user(conn, generic_user)); }};
+
     SECTION("local user operations")
     {
         adm::user test_user{"unit_test_test_user"};
@@ -222,5 +228,50 @@ TEST_CASE("user group administration")
         irods::experimental::client_connection user_conn{
             irods::experimental::defer_authentication, env.rodsHost, env.rodsPort, {test_user.name, env.rodsZone}};
         CHECK(clientLoginWithPassword(static_cast<RcComm*>(user_conn), test_user_prop.value.data()) == 0);
+    }
+
+    SECTION("#7576: ensure no limitations on proxied groupadmin")
+    {
+        rodsEnv env{};
+        _getRodsEnv(env);
+
+        REQUIRE_NOTHROW(
+            adm::client::modify_user(conn, generic_user, adm::user_type_property{adm::user_type::groupadmin}));
+        REQUIRE(adm::client::type(conn, generic_user) == adm::user_type::groupadmin);
+
+        // Connect to iRODS
+        irods::experimental::client_connection user_conn{env.rodsHost,
+                                                         env.rodsPort,
+                                                         {env.rodsUserName, env.rodsZone}, // (proxy) rodsadmin user
+                                                         {generic_user.name, env.rodsZone}}; // groupadmin user
+
+        // Try to add a group as the proxied groupadmin user.
+        adm::group test_group{"bad_bad_water_group"};
+        REQUIRE_NOTHROW(adm::client::add_group(user_conn, test_group));
+        REQUIRE(adm::client::exists(user_conn, test_group));
+        irods::at_scope_exit group_cleanup{
+            [&conn, &test_group] { CHECK_NOTHROW(adm::client::remove_group(conn, test_group)); }};
+
+        // Add self to group
+        REQUIRE_NOTHROW(adm::client::add_user_to_group(user_conn, test_group, generic_user));
+        REQUIRE(adm::client::user_is_member_of_group(user_conn, test_group, generic_user));
+        irods::at_scope_exit groupadmin_self_add_cleanup{[&user_conn, &test_group, &generic_user] {
+            CHECK_NOTHROW(adm::client::remove_user_from_group(user_conn, test_group, generic_user));
+        }};
+
+        // Create and remove user using groupadmin
+        adm::user groupadmin_created_user{"mr_flips"};
+        REQUIRE_NOTHROW(adm::client::add_user(user_conn, groupadmin_created_user));
+        REQUIRE(adm::client::exists(user_conn, groupadmin_created_user));
+        irods::at_scope_exit created_user_cleanup{[&conn, &groupadmin_created_user] {
+            CHECK_NOTHROW(adm::client::remove_user(conn, groupadmin_created_user));
+        }};
+
+        // Add and remove user from group using groupadmin
+        REQUIRE_NOTHROW(adm::client::add_user_to_group(user_conn, test_group, groupadmin_created_user));
+        REQUIRE(adm::client::user_is_member_of_group(user_conn, test_group, groupadmin_created_user));
+
+        // Begin cleanup. Previous 'irods::at_scope_exit's will run after the following line
+        CHECK_NOTHROW(adm::client::remove_user_from_group(user_conn, test_group, groupadmin_created_user));
     }
 }
--- a/lib/administration/user/include/irods/user_administration.hpp
+++ b/lib/administration/user/include/irods/user_administration.hpp
@@ -6,15 +6,18 @@
 #undef NAMESPACE_IMPL
 #undef RxComm
 #undef rxGeneralAdmin
+#undef rxUserAdmin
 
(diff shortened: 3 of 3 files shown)


unit_tests/src/test_client_connection.cpp
```diff
@@ -300,3 +300,40 @@ TEST_CASE("#7192: all connections store unique session signatures", "client_conn
...
+                                                {"group_user", "tempZone"}}; // groupadmin user
+
+    // Authenticate the rodsadmin.
+    char pw[] = "rods";
+    clientLoginWithPassword((RcComm*) conn, pw);
+
+    REQUIRE(ia::client::type(conn, group_user) == ia::user_type::groupadmin);
+
+    // Try to add a group as the proxied groupadmin user.
+    REQUIRE_NOTHROW(ia::client::add_group(conn, ia::group{"in_group"}));
```
MartinFlores751, line 338: Consider adding a bonus check to ensure the group is created.
Also, don't forget to clean up the group you just created!

lib/administration/user/src/user_administration.cpp
```diff
@@ -85,18 +85,29 @@ namespace irods::experimental::administration::NAMESPACE_IMPL
...
+                                 .arg3 = const_cast<char*>(zone.data())};
+
+            if (const auto ec{rxUserAdmin(&_comm, &input)}; ec != 0) {
+                // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+                THROW(ec, fmt::format("Failed to add group [{}].", _group.name));
+            }
+        }
+        else {
+            GeneralAdminInput input{
+                .arg0 = "add", .arg1 = "user", .arg2 = _group.name.data(), .arg3 = "rodsgroup", .arg4 = zone.data()};
```
MartinFlores751, line 106: If we keep this style, consider manual formatting? Think it's a bit unreadable as is.

MartinFlores751 (commented):
Some words.

lib/administration/user/src/user_administration.cpp
```diff
@@ -85,18 +85,29 @@ namespace irods::experimental::administration::NAMESPACE_IMPL
...
+            userAdminInp_t input{.arg0 = "mkgroup",
+                                 .arg1 = const_cast<char*>(_group.name.data()),
+                                 .arg2 = "rodsgroup",
+                                 .arg3 = const_cast<char*>(zone.data())};
+
+            if (const auto ec{rxUserAdmin(&_comm, &input)}; ec != 0) {
+                // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+                THROW(ec, fmt::format("Failed to add group [{}].", _group.name));
+            }
+        }
```
korydraughn, line 103: If you insert a `return` statement at the end of this block, the rest of the code can remain as it was (i.e. no additional indentation necessary) and the new block then sticks out more to the reader as a special case.

lib/administration/user/src/user_administration.cpp
```diff
@@ -85,18 +85,29 @@ namespace irods::experimental::administration::NAMESPACE_IMPL
...
+        const auto user_type{type(_comm, current_user)};
 
-        if (const auto ec = rxGeneralAdmin(&_comm, &input); ec != 0) {
-            // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
-            THROW(ec, fmt::format("Failed to add group [{}].", _group.name));
+        if (user_type == irods::experimental::administration::user_type::groupadmin) {
+            userAdminInp_t input{.arg0 = "mkgroup",
+                                 .arg1 = const_cast<char*>(_group.name.data()),
+                                 .arg2 = "rodsgroup",
+                                 .arg3 = const_cast<char*>(zone.data())};
```
korydraughn, line 97: Are there other members in that struct?
Are they guaranteed to be initialized?
Are we required by the C++ standard to initialize all members?

unit_tests/src/test_client_connection.cpp
```diff
@@ -300,3 +300,40 @@ TEST_CASE("#7192: all connections store unique session signatures", "client_conn
...
+    // Connect to iRODS without authenticating.
+    irods::experimental::client_connection conn{irods::experimental::defer_authentication,
+                                                env.rodsHost,
+                                                env.rodsPort,
+                                                {env.rodsUserName, env.rodsZone}, // (proxy) rodsadmin user
+                                                {"group_user", "tempZone"}}; // groupadmin user
+
+    // Authenticate the rodsadmin.
+    char pw[] = "rods";
+    clientLoginWithPassword((RcComm*) conn, pw);
```
korydraughn, line 333: `REQUIRE(clientLoginWithPassword(static_cast<RcComm*>(conn), pw) == 0);`

unit_tests/src/test_client_connection.cpp
```diff
@@ -300,3 +300,40 @@ TEST_CASE("#7192: all connections store unique session signatures", "client_conn
...
+
+    // Connect to iRODS without authenticating.
+    irods::experimental::client_connection conn{irods::experimental::defer_authentication,
+                                                env.rodsHost,
+                                                env.rodsPort,
+                                                {env.rodsUserName, env.rodsZone}, // (proxy) rodsadmin user
+                                                {"group_user", "tempZone"}}; // groupadmin user
+
+    // Authenticate the rodsadmin.
+    char pw[] = "rods";
```
korydraughn, line 332: This assumes the test can only be run by a rodsadmin with a password of "rods".

Probably want to create another rodsadmin user and set their password within the test.

unit_tests/src/test_client_connection.cpp
```diff
@@ -300,3 +300,40 @@ TEST_CASE("#7192: all connections store unique session signatures", "client_conn
...
+        [&admin_conn, group_user] { ia::client::remove_user(admin_conn, group_user); }};
+
+    REQUIRE_NOTHROW(ia::client::modify_user(admin_conn, group_user, ia::user_type_property{ia::user_type::groupadmin}));
+
+    // Connect to iRODS without authenticating.
+    irods::experimental::client_connection conn{irods::experimental::defer_authentication,
+                                                env.rodsHost,
+                                                env.rodsPort,
+                                                {env.rodsUserName, env.rodsZone}, // (proxy) rodsadmin user
+                                                {"group_user", "tempZone"}}; // groupadmin user
```
korydraughn, line 329: `"group_user"` -> `group_user.name`
`"tempZone"` -> `env.rodsZone`

unit_tests/src/test_client_connection.cpp
```diff
@@ -300,3 +300,40 @@ TEST_CASE("#7192: all connections store unique session signatures", "client_conn
 
     CHECK(std::all_of(std::begin(pairs), std::end(pairs), has_unique_session_signatures));
 }
+
+TEST_CASE("#7576: ensure no limitaitons on proxied groupadmin", "client_connection")
```
korydraughn, line 304: This unit test file isn't the best fit for the test. test_user_administration.cpp is a much better fit since the bug is in that library.

unit_tests/src/test_client_connection.cpp
```diff
@@ -300,3 +300,40 @@ TEST_CASE("#7192: all connections store unique session signatures", "client_conn
...
+                                                env.rodsHost,
+                                                env.rodsPort,
+                                                {env.rodsUserName, env.rodsZone}, // (proxy) rodsadmin user
+                                                {"group_user", "tempZone"}}; // groupadmin user
+
+    // Authenticate the rodsadmin.
+    char pw[] = "rods";
+    clientLoginWithPassword((RcComm*) conn, pw);
+
+    REQUIRE(ia::client::type(conn, group_user) == ia::user_type::groupadmin);
```
korydraughn, line 335: This can be moved just below L322 and asserted using the `admin_conn`.

unit_tests/src/test_client_connection.cpp
korydraughn, line 304: limitaitons -> limitations

unit_tests/src/test_client_connection.cpp
```diff
@@ -300,3 +300,40 @@ TEST_CASE("#7192: all connections store unique session signatures", "client_conn
...
+    // Creates a connection to an iRODS server based on the user's
+    // $HOME/.irods/irods_environment.json file.
+    irods::experimental::client_connection admin_conn;
+
+    namespace ia = irods::experimental::administration;
+    const ia::user group_user{"group_user"};
+    REQUIRE_NOTHROW(ia::client::add_user(admin_conn, group_user));
+
+    irods::at_scope_exit remove_defer_user{
+        [&admin_conn, group_user] { ia::client::remove_user(admin_conn, group_user); }};
```
korydraughn, line 320: Missing `&` on `group_user`?

unit_tests/src/test_client_connection.cpp
korydraughn: Yes please.

korydraughn (commented):
Good stuff.

Going to need that server-side unit test as well. It could be identical in behavior to the client-side version.

See the following for examples of server-side unit tests.
https://github.com/irods/irods/blob/669a726968dcdd7c4dab7d52761095119511c182/plugins/microservices/src/test_user_administration.cpp

https://github.com/irods/irods/blob/669a726968dcdd7c4dab7d52761095119511c182/plugins/microservices/src/test_scoped_permission.cpp

And here's a commit which demonstrates how to add server-side tests in full.
https://github.com/irods/irods/commit/8a3ff6c9ec863db97958fe5ec745136110291f9f

lib/administration/user/src/user_administration.cpp
korydraughn: Not against this style, but there's definitely C++ rules about proper usage. I don't know what they are off the top of my head though.

If you use `input.argN` then formatting is easier, we don't have to worry about new C++ rules, and it's easier to maintain and understand.

lib/administration/user/src/user_administration.cpp
MartinFlores751: > Are there other members in that struct?

Yes, the struct goes from arg0 to arg9, much like `GeneralAdminInput`.

> Are they guaranteed to be initialized?

Relevant Page: https://en.cppreference.com/w/cpp/language/aggregate_initialization

Partial quote:

> ...
> For a non-union aggregate, each element that is not an explicitly initialized element is initialized as follows:
> - If the element has a default member initializer, the element is initialized from that initializer. (since C++11)
> - Otherwise, if the element is not a reference, the element is copy-initialized from an empty initializer list.
> - Otherwise, the program is ill-formed. 
>
> ...

So in this case, since the struct contains 10 `char*`, the ones not explicitly defined behave as if initialized as follows: `char* arg# {};`.

That would mean it does zero initialization: https://en.cppreference.com/w/cpp/language/zero_initialization.

This would then would apply the following rule, as a pointer is a scalar type:
> ...
> The effects of zero-initialization are:
> - If T is a scalar type, the object is initialized to the value obtained by explicitly converting the integer literal ​0​ (zero) to T.
>
> ...

Very good question, made me learn a good bit about how this works too. 👍 

> Are we required by the C++ standard to initialize all members?

I think the question before answered this one, but not initializing the pointers might mean we try to access something that doesn't exist, right?

Anywho, I think by doing `userAdminInp_t input;`, then initializing the elements thereafter would avoid initializing all elements.

korydraughn: Good. Check the compiler implementation status for clang 13.0.1 to be sure it fully implements the feature.

> Anywho, I think by doing `userAdminInp_t input;`, then initializing the elements thereafter would avoid initializing all elements.

We'd make it clear its members by adding a pair of curly braces to it (e.g. `userAdminInp_t input{}`).

FifthPotato, line 97: Are you `const_cast`ing away `const` here? Are you guaranteed that none of these pointers are ever modified downstream? Could be UB otherwise.

FifthPotato (commented):
Not much to say, looks good to me! Just one comment.

unit_tests/src/test_client_connection.cpp
MartinFlores751: Done.

unit_tests/src/test_client_connection.cpp
MartinFlores751: Done.

unit_tests/src/test_client_connection.cpp
MartinFlores751: Done.

unit_tests/src/test_client_connection.cpp
MartinFlores751: Done.

unit_tests/src/test_client_connection.cpp
MartinFlores751: Done.

lib/administration/user/src/user_administration.cpp
MartinFlores751: Supported since clang 10, it's a subset(?) of aggregate initialization called Designated initializers. https://en.cppreference.com/w/cpp/compiler_support/20

lib/administration/user/src/user_administration.cpp
MartinFlores751: Done.

unit_tests/src/test_client_connection.cpp
MartinFlores751: Done.

lib/administration/user/src/user_administration.cpp
trel: And GCC8 it looks like

korydraughn: Good. I don't have a strong preference on this. Seems fine to me.

unit_tests/src/test_client_connection.cpp
MartinFlores751: Done.

unit_tests/src/test_user_administration.cpp
```diff
@@ -223,4 +229,26 @@ TEST_CASE("user group administration")
...
+        irods::experimental::client_connection user_conn{env.rodsHost,
+                                                         env.rodsPort,
+                                                         {env.rodsUserName, env.rodsZone}, // (proxy) rodsadmin user
+                                                         {generic_user.name, env.rodsZone}}; // groupadmin user
+
+        // Try to add a group as the proxied groupadmin user.
+        adm::group test_group{"bad_bad_water_group"};
+        REQUIRE_NOTHROW(adm::client::add_group(user_conn, test_group));
+        irods::at_scope_exit group_cleanup{
+            [&user_conn, &test_group] { CHECK_NOTHROW(adm::client::remove_group(user_conn, test_group)); }};
```
MartinFlores751, line 252: `remove_group(...)` produces a `SYS_NO_API_PRIV`, going to patch this shortly.

MartinFlores751: Never mind, a `groupadmin` cannot do this to begin with.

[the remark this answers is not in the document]
trel: groupadmins cannot remove groups

https://docs.irods.org/4.3.1/icommands/groupadmin/

alanking:
Please mark as ready for review when true

unit_tests/src/test_user_administration.cpp
```diff
@@ -223,4 +229,39 @@ TEST_CASE("user group administration")
...
+        REQUIRE_NOTHROW(adm::client::add_user(user_conn, groupadmin_created_user));
+        irods::at_scope_exit created_user_cleanup{[&conn, &groupadmin_created_user] {
+            CHECK_NOTHROW(adm::client::remove_user(conn, groupadmin_created_user));
+        }};
+
+        // Add and remove user from group using groupadmin
+        REQUIRE_NOTHROW(adm::client::add_user_to_group(user_conn, test_group, groupadmin_created_user));
+        irods::at_scope_exit group_add_cleanup{[&user_conn, &test_group, &groupadmin_created_user] {
+            CHECK_NOTHROW(adm::client::remove_user_from_group(user_conn, test_group, groupadmin_created_user));
+        }};
```
MartinFlores751, line 265: I think I found a bug? Looking at `test_igroupadmin.py` for reference, it looks like we want group admins to be bound to a group before they can manage it.

Here I can add a user to a group, but cannot remove it, since the groupadmin was never added to the group.

korydraughn: Hmm, that does sound like a bug.

Is it only the library that's violating the rules or the library and the icommands that's violating the rules?

alanking: If you mean a bug in this test, please disregard the rest of this. :)

If you mean a bug with groupadmins... I think that this is not a bug.

groupadmins can add users to a group in exactly 2 situations:
 1. When the groupadmin is a member of the group
 2. When the group has no members

So the very first thing a groupadmin should do after creating a group to manage is add that groupadmin user to the group so that the groupadmin can manage the new group. This is described in the `igroupadmin mkgroup` help text (https://docs.irods.org/4.3.1/icommands/groupadmin/#mkgroup):

> You will need to add yourself to the new group to then
be able to add and remove others (when the group is empty groupadmins are
allowed to add themselves.)

If the groupadmin is not the first member of the new group, the groupadmin cannot manage the group until a user with permission to manage the group adds that groupadmin. So, the test should first add the groupadmin to the group before adding the other user.

trel: from the help text - not sure if this clarifies what you're seeing.

the groupadmin can add themselves only to empty groups.

```
mkgroup
 mkgroup groupName[#Zone] (make a new group)
Make a new group.  You will need to add yourself to the new group to then
be able to add and remove others (when the group is empty groupadmins are
allowed to add themselves.)
```

trel: jinx.  and i lost.

korydraughn: Excellent save! Forgot about that.

MartinFlores751: Is it fine to allow adding a non-groupadmin to a group first?

trel: You can do that - but the groupadmin will not be able to add themselves to a non-empty group.

trel: rodsadmin can do whatever they want in whatever order.

MartinFlores751:
Mini-discuss with Kory, decided to remove the microservice tests. Server-side tests aren't necessary, since we're already in the server at that point.

Revert code changes regarding that, as well as a few others that modify additional code not directly relevant to the issue.

lib/administration/user/src/user_administration.cpp
MartinFlores751: Reverted this part.

lib/administration/user/src/user_administration.cpp
MartinFlores751: Think this is fine for new code, closing now.

unit_tests/src/test_user_administration.cpp
MartinFlores751: Thinking of creating an issue to make a test and solidify this as a feature?

lib/administration/user/src/user_administration.cpp
MartinFlores751: Going to close this for now, unless someone else has issues with the const_cast.

unit_tests/src/test_user_administration.cpp
korydraughn: Sure. Put it in an issue and we'll get it handled in 4.3.3.

unit_tests/src/test_user_administration.cpp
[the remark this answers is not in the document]
MartinFlores751: > I'm a little confused by the at_scope_exit appearing at the end of the scope here... Maybe this could just be the remove_group call, sans at_scope_exit?

Can do. Would it still be a CHECK_NOTHROW or does it get upgraded to a REQUIRE_NOTHROW?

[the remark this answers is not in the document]
alanking: CHECK_NOTHROW is fine. Since it's the last thing in the test I suppose it doesn't matter much

unit_tests/src/test_client_connection.cpp
MartinFlores751: Done.

unit_tests/src/test_user_administration.cpp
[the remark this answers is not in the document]
MartinFlores751: Done. Kept it as a CHECK.

unit_tests/src/test_user_administration.cpp
MartinFlores751: Issue created, see #7707.

korydraughn:
All review comments are now shown to be resolved. What's next? Testing?

korydraughn:
Cool cool.

MartinFlores751:
Unit tests passed, running core.

MartinFlores751:
All tests passed.

korydraughn (commented):
Rebase and squash to taste, no #'s.

MartinFlores751:
Squashed.

trel (approved):
looks good to me.

MartinFlores751:
Found a mistake in the commit note, going to fix it n push it up.

korydraughn (approved):
Pound it.

MartinFlores751:
Pounded. The mistake in the commit message was referencing `remove_user(...)` instead of `remove_group(...)`.

PR #7691, opened 2024-04-16
[7553] Serialize function for MsParamArray (main)

FifthPotato:
First attempt. Couple of things that need to be checked:
Is the type `MsParamArray**` or `MsParamArray*`? The two api calls in the api table that seem to use it both use this double-pointer. The other place where it seems like it could be used is `ExecMyRuleInp`'s member `inpParamArray`, which is a `MsParamArray*`
What is the minimum functionality that could trigger something for this? I tried doing a pep with `pep_api_exec_my_rule_post`, but couldn't seem to get any entries.

Still needs a test, once I know what to test for.

PR #7693, opened 2024-04-17
[7552] Serialize ExecMyRuleInp

FifthPotato:
Still drafty, because I'd like some assistance in figuring out a real test case because I'm a NREP noob. `writeLine` at a PEP seems to work, but a lot of the members seem to not have values at first blush. It does print the rule text, though, which is nice, I guess.
Code for the `MsParamArray` is mostly copied from https://github.com/irods/irods/pull/7691, and will probably be updated in lockstep with the other.

### DIFF
--- a/server/re/src/irods_re_serialization.cpp
+++ b/server/re/src/irods_re_serialization.cpp
@@ -2,6 +2,7 @@
 
 #include "irods/irods_logger.hpp"
 #include "irods/irods_plugin_context.hpp"
+#include "irods/msParam.h"
 #include "irods/rodsErrorTable.h"
 
 #include <boost/lexical_cast.hpp>
@@ -1199,6 +1200,102 @@ namespace irods::re_serialization
         }
     } // serialize_vector_of_strings_ptr
 
+    static irods::error serialize_execMyRuleInp_ptr(boost::any _p, serialized_parameter_t& _out)
+    {
+        try {
+            log_re::trace(__func__);
+            const auto* casted_ptr = boost::any_cast<execMyRuleInp_t*>(_p);
+            if (!casted_ptr) {
+                _out["execMyRuleInp_ptr"] = "nullptr";
+                return SUCCESS();
+            }
+            _out["myRule"] = casted_ptr->myRule;
+            _out["addr_hostAddr"] = casted_ptr->addr.hostAddr;
+            _out["addr_zoneName"] = casted_ptr->addr.zoneName;
+            _out["addr_portNum"] = boost::lexical_cast<std::string>(casted_ptr->addr.portNum);
+            _out["addr_dummyInt"] = boost::lexical_cast<std::string>(casted_ptr->addr.dummyInt);
+            serialize_keyValPair(casted_ptr->condInput, _out);
+            _out["outParamDesc"] = casted_ptr->outParamDesc;
+
+            const auto* msparamarray_ptr = casted_ptr->inpParamArray;
+            if (!msparamarray_ptr) {
+                _out["inpParamArray"] = "nullptr";
+                return SUCCESS();
+            }
+            _out["inpParamArray_len"] = boost::lexical_cast<std::string>(msparamarray_ptr->len);
+            _out["inpParamArray_oprType"] = boost::lexical_cast<std::string>(msparamarray_ptr->oprType);
+            for (int i = 0; i < msparamarray_ptr->len; i++) {
+                std::string prefix = fmt::format("inpParamArray_{}_", i);
+                const auto* lbl = msparamarray_ptr->msParam[i]->label;
+                const auto* typ = msparamarray_ptr->msParam[i]->type;
+                _out[prefix + "label"] = lbl ? lbl : "nullptr";
+                _out[prefix + "type"] = typ ? typ : "nullptr";
+                const auto* msparam_ptr = msparamarray_ptr->msParam[i];
+
+                if (msparam_ptr == nullptr || msparam_ptr->inOutStruct == nullptr || !typ) {
+                    _out[prefix + "inOutStruct"] = "nullptr";
+                }
+                else if (strcmp(msparam_ptr->type, STR_MS_T) == 0) {
+                    _out[prefix + "inOutStruct"] = static_cast<char*>(msparam_ptr->inOutStruct);
+                }
+                else if (strcmp(msparam_ptr->type, INT_MS_T) == 0) {
+                    _out[prefix + "inOutStruct"] = std::to_string(parseMspForPosInt(const_cast<MsParam*>(msparam_ptr)));
+                }
+                else if (strcmp(msparam_ptr->type, DOUBLE_MS_T) == 0) {
+                    double out = 0.0;
+                    if (parseMspForDouble(const_cast<MsParam*>(msparam_ptr), &out)) {
+                        THROW(SYS_INVALID_INPUT_PARAM, "Failed to convert microservice argument to float.");
+                    }
+                    _out[prefix + "inOutStruct"] = std::to_string(out);
+                }
+                else if (strcmp(msparam_ptr->type, FLOAT_MS_T) == 0) {
+                    float out = 0.0;
+                    if (parseMspForFloat(const_cast<MsParam*>(msparam_ptr), &out)) {
+                        THROW(SYS_INVALID_INPUT_PARAM, "Failed to convert microservice argument to float.");
+                    }
+                    _out[prefix + "inOutStruct"] = std::to_string(out);
+                }
+                else {
+                    _out[prefix + "inOutStruct"] = fmt::format("unrepresentable_type_{}", msparam_ptr->type);
+                }
+                const auto* ptr = msparamarray_ptr->msParam[i]->inpOutBuf;
+                if (ptr) {
+                    _out[prefix + "inpOutBuf_len"] = boost::lexical_cast<std::string>(ptr->len);
+                    std::string& str = _out[prefix + "inpOutBuf_buf"];
+
+                    for (int j = 0; j < ptr->len; j++) {
+                        unsigned char c = *(static_cast<unsigned char*>(ptr->buf) + j);
+
+                        if (isprint(c) || isspace(c) || isblank(c)) {
+                            str += c;
+                        }
+                        else {
+                            str += fmt::format("\\x{0:02x}", c);
+                        }
+                    }
+                }
+                else {
+                    _out[prefix + "null_value"] = "null_value";
+                }
+            }
+            return SUCCESS();
+        }
+        catch (const boost::bad_any_cast& e) {
+            return ERROR(INVALID_ANY_CAST,
+                         fmt::format("{}: failed to cast pointer to [execMyRuleInp_t*]: {}", __func__, e.what()));
+        }
+        catch (const irods::exception& e) {
+            return ERROR(
+                e.code(),
+                fmt::format(
+                    "{}: failed to cast pointer to [execMyRuleInp_t*]: {}, ec: {}", __func__, e.what(), e.code()));
+        }
+        catch (const std::exception& e) {
+            return ERROR(
+                SYS_LIBRARY_ERROR, fmt::format("{}: failed to serialize [execMyRuleInp_t*]: {}", __func__, e.what()));
+        }
+    } // serialize_execMyRuleInp_ptr
+
 #if 0
     static error serialize_XXXX_ptr(
             boost::any               _p,
@@ -1255,7 +1352,8 @@ namespace irods::re_serialization
             {std::type_index(typeid(bytesBuf_t*)), serialize_bytesBuf_ptr},
             {std::type_index(typeid(Genquery2Input*)), serialize_Genquery2Input_ptr},
             {std::type_index(typeid(const std::vector<std::string>*)), serialize_const_vector_of_strings_ptr},
-            {std::type_index(typeid(std::vector<std::string>*)), serialize_vector_of_strings_ptr}};
+            {std::type_index(typeid(std::vector<std::string>*)), serialize_vector_of_strings_ptr},
+            {std::type_index(typeid(execMyRuleInp_t*)), serialize_execMyRuleInp_ptr}};
         return the_map;
 
     } // get_serialization_map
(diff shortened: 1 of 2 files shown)


server/re/src/irods_re_serialization.cpp
```diff
@@ -1199,6 +1199,77 @@ namespace irods::re_serialization
...
+
+            const auto* msparamary_ptr = casted_ptr->inpParamArray;
+            if(!msparamary_ptr) {
+                _out["inpParamArray"] = "nullptr";
+                return SUCCESS();
+            }
+            _out["inpParamArray_len"] = boost::lexical_cast<std::string>(msparamary_ptr->len);
+            _out["inpParamArray_oprType"] = boost::lexical_cast<std::string>(msparamary_ptr->oprType);
+            for (int i = 0; i < msparamary_ptr->len; i++) {
+                std::string prefix = "inpParamArray_" + std::to_string(i) + "_";
```
korydraughn, line 1227: Use `fmt::format()` or `+=` to concat strings. This avoids unnecessary temporaries.

server/re/src/irods_re_serialization.cpp
```diff
@@ -1199,6 +1199,77 @@ namespace irods::re_serialization
...
+                _out[prefix + "label"] = msparamary_ptr->msParam[i]->label;
+                _out[prefix + "type"] = msparamary_ptr->msParam[i]->type;
+                const auto* msparam_ptr = msparamary_ptr->msParam[i];
+
+                if (msparam_ptr == NULL || msparam_ptr->inOutStruct == NULL) {
```
korydraughn, line 1232: NULL -> nullptr

server/re/src/irods_re_serialization.cpp
```diff
@@ -1199,6 +1199,77 @@ namespace irods::re_serialization
...
+                    _out[prefix + "inOutStruct"] = "null";
+                }
+                else if (strcmp(msparam_ptr->type, STR_MS_T) != 0) {
+                    _out[prefix + "inOutStruct"] = "non_string_value";
+                }
+                else {
+                    _out[prefix + "inOutStruct"] = static_cast<char*>(msparam_ptr->inOutStruct);
+                }
```
korydraughn, line 1240: Probably should add support for a few other types as well.

See the following.
https://github.com/irods/irods/blob/669a726968dcdd7c4dab7d52761095119511c182/server/core/src/json_serialization.cpp#L143-L157

server/re/src/irods_re_serialization.cpp
```diff
@@ -1199,6 +1199,77 @@ namespace irods::re_serialization
...
+                            str += c;
+                        }
+                        else {
+                            str += fmt::format("\\x{0:02x}", c);
+                        }
+                    }
+                }
+                else {
+                    _out[prefix + "null_value"] = "null_value";
+                }
```
korydraughn, line 1259: Is this code path ever hit? If yes, what information does it add to the serialized struct?

server/re/src/irods_re_serialization.cpp
```diff
@@ -1199,6 +1199,77 @@ namespace irods::re_serialization
...
+        }
+        catch (const boost::bad_any_cast& e) {
+            return ERROR(INVALID_ANY_CAST,
+                         fmt::format("{}: failed to cast pointer to [execMyRuleInp_t*]: {}", __func__, e.what()));
+        }
+        catch (const std::exception& e) {
+            return ERROR(
+                SYS_LIBRARY_ERROR, fmt::format("{}: failed to serialize [execMyRuleInp_t*]: {}", __func__, e.what()));
+        }
+    } //serialize_execMyRuleInp_ptr
```
trel, line 1271: add a space
"//s" -> "// s"

server/re/src/irods_re_serialization.cpp
```diff
@@ -1199,6 +1199,77 @@ namespace irods::re_serialization
...
+                _out["execMyRuleInp_ptr"] = "nullptr";
+                return SUCCESS();
+            }
+            _out["myRule"] = casted_ptr->myRule;
+            _out["addr_hostAddr"] = casted_ptr->addr.hostAddr;
+            _out["addr_zoneName"] = casted_ptr->addr.zoneName;
+            _out["addr_portNum"] = boost::lexical_cast<std::string>(casted_ptr->addr.portNum);
+            _out["addr_dummyInt"] = boost::lexical_cast<std::string>(casted_ptr->addr.dummyInt);
+            serialize_keyValPair(casted_ptr->condInput, _out);
+            _out["outParamDesc"] = casted_ptr->outParamDesc;
```
korydraughn, line 1217: I'm guessing none of these need to be checked for null because they are fixed size buffers (no allocation necessary)?

server/re/src/irods_re_serialization.cpp
```diff
@@ -1199,6 +1199,77 @@ namespace irods::re_serialization
...
+
+            const auto* msparamary_ptr = casted_ptr->inpParamArray;
+            if(!msparamary_ptr) {
+                _out["inpParamArray"] = "nullptr";
```
korydraughn, line 1221: Perhaps this should be set to `"null"` just like on L1233?

server/re/src/irods_re_serialization.cpp
```diff
@@ -1199,6 +1199,77 @@ namespace irods::re_serialization
...
+                else {
+                    _out[prefix + "inOutStruct"] = static_cast<char*>(msparam_ptr->inOutStruct);
+                }
+                const auto* ptr = msparamary_ptr->msParam[i]->inpOutBuf;
+                if (ptr) {
+                    _out[prefix + "inpOutBuf_len"] = boost::lexical_cast<std::string>(ptr->len);
+                    std::string& str = _out[prefix + "inpOutBuf_buf"];
+
+                    for (int j = 0; j < ptr->len; j++) {
+                        unsigned char c = *((unsigned char*) (ptr->buf) + j);
```
korydraughn, line 1247: Prefer C++ casts.

trel:
Can you see the values when calling `irule -F`?

You can pass inputs and rule text in, and ... should be able to see them via writeLine?

server/re/src/irods_re_serialization.cpp
```diff
@@ -1199,6 +1199,77 @@ namespace irods::re_serialization
...
+                else if (strcmp(msparam_ptr->type, STR_MS_T) != 0) {
+                    _out[prefix + "inOutStruct"] = "non_string_value";
+                }
+                else {
+                    _out[prefix + "inOutStruct"] = static_cast<char*>(msparam_ptr->inOutStruct);
+                }
+                const auto* ptr = msparamary_ptr->msParam[i]->inpOutBuf;
+                if (ptr) {
+                    _out[prefix + "inpOutBuf_len"] = boost::lexical_cast<std::string>(ptr->len);
+                    std::string& str = _out[prefix + "inpOutBuf_buf"];
```
korydraughn, line 1244: Computing the key twice?

Perhaps `insert_or_assign()` is a better option? See the following.
- https://en.cppreference.com/w/cpp/container/map/insert_or_assign.

server/re/src/irods_re_serialization.cpp
```diff
@@ -1199,6 +1199,77 @@ namespace irods::re_serialization
...
+                        if (isprint(c) || isspace(c) || isblank(c)) {
+                            str += c;
+                        }
+                        else {
+                            str += fmt::format("\\x{0:02x}", c);
+                        }
+                    }
+                }
+                else {
+                    _out[prefix + "null_value"] = "null_value";
```
korydraughn, line 1258: Can this result in overwriting previous mappings?

Also, `= "null_value"` -> `= "null"`?

korydraughn (commented):
Let's link up and do some live debugging today.

server/re/src/irods_re_serialization.cpp
alanking: Agreed, we should pick one. We are setting `execMyRuleInp_ptr` to "nullptr" on 1208. Seems we are using "nullptr" for all other serializations (at least that I found), so I think we should stick with that and change L1233. Thoughts?

server/re/src/irods_re_serialization.cpp
FifthPotato: I believe so. In order...
`myRule` is `char[META_STR_LEN]`
`addr` is `rodsHostAddr_t` (not a pointer)
`hostAddr` is `char[LONG_NAME_LEN]`
`zoneName` is `char[NAME_LEN]`
`portNum` and `dummyInt` are `int`
`outParamDesc` is `char[LONG_NAME_LEN]`

server/re/src/irods_re_serialization.cpp
FifthPotato: I don't see the first key computation-- am I missing something?

server/re/src/irods_re_serialization.cpp
FifthPotato: I do not believe any previous key is named `null_value`, so I don't think a collision is possible here.
As for the assigned value, this block was mostly lifted from the `bytesBuf_t` serializer elsewhere in this file. I thought to keep it the same to be consistent, but I can change this accordingly if you think it would be better otherwise.

server/re/src/irods_re_serialization.cpp
korydraughn: Oops. I misread the string literals.

server/re/src/irods_re_serialization.cpp
korydraughn: That's fine. Leave it as is.

alanking (commented):
Please mark as ready for review when true

server/re/src/irods_re_serialization.cpp
FifthPotato: I couldn't get it to trigger under normal testing. From what I can tell, it happens when one of the `MsParam` has something in its `inpOutBuf`, which is of type `bytesBuf_t`, which...I'm not entirely sure when this happens?

server/re/src/irods_re_serialization.cpp
FifthPotato: Added these, but I don't know how to verify if it works. `irule` seems to want to pass in all arguments as strings. Any way around that?

trel: I don't think there is.  I think another client could potentially submit other types - but I think testing for the strings is probably the 95%+ use case.

korydraughn: Right. Covering the string case is good enough. The extra types are just me wanting to cover potential cases.

server/re/src/irods_re_serialization.cpp
korydraughn: That's fine.

Did the implementation come from `serialize_bytesBuf_ptr`?

server/re/src/irods_re_serialization.cpp
```diff
@@ -1199,6 +1200,94 @@ namespace irods::re_serialization
...
+            if (!msparamarray_ptr) {
+                _out["inpParamArray"] = "nullptr";
+                return SUCCESS();
+            }
+            _out["inpParamArray_len"] = boost::lexical_cast<std::string>(msparamarray_ptr->len);
+            _out["inpParamArray_oprType"] = boost::lexical_cast<std::string>(msparamarray_ptr->oprType);
+            for (int i = 0; i < msparamarray_ptr->len; i++) {
+                std::string prefix = fmt::format("inpParamArray_{}_", i);
+                _out[prefix + "label"] = msparamarray_ptr->msParam[i]->label;
+                _out[prefix + "type"] = msparamarray_ptr->msParam[i]->type;
```
korydraughn, line 1230: `->label` and `->type` need to be checked for null. If they are null, use "nullptr" or an empty string.

server/re/src/irods_re_serialization.cpp
```diff
@@ -1199,6 +1200,94 @@ namespace irods::re_serialization
...
+                const auto* msparam_ptr = msparamarray_ptr->msParam[i];
+
+                if (msparam_ptr == nullptr || msparam_ptr->inOutStruct == nullptr) {
+                    _out[prefix + "inOutStruct"] = "nullptr";
+                }
+                else if (strcmp(msparam_ptr->type, STR_MS_T) == 0) {
```
korydraughn, line 1236: `->type` can be null. Need to check before passing to `strcmp`.

_This applies to the other cases that follow too._

server/re/src/irods_re_serialization.cpp
```diff
@@ -1199,6 +1200,94 @@ namespace irods::re_serialization
...
+                }
+                else if (strcmp(msparam_ptr->type, FLOAT_MS_T) == 0) {
+                    float out = 0.0;
+                    if (parseMspForFloat(const_cast<MsParam*>(msparam_ptr), &out)) {
+                        THROW(SYS_INVALID_INPUT_PARAM, "Failed to convert microservice argument to float.");
+                    }
+                    _out[prefix + "inOutStruct"] = std::to_string(out);
+                }
+                else {
+                    _out[prefix + "inOutStruct"] = fmt::format("unrepresentable_type_{}", msparam_ptr->type);
```
korydraughn, line 1257: May be a problem if `->type` is null.

server/re/src/irods_re_serialization.cpp
```diff
@@ -1199,6 +1200,94 @@ namespace irods::re_serialization
...
+                        }
+                    }
+                }
+                else {
+                    _out[prefix + "null_value"] = "null_value";
+                }
+            }
+            return SUCCESS();
+        }
+        catch (const boost::bad_any_cast& e) {
```
korydraughn, line 1281: Needs a catch block for `const irods::exception&` due to the use of `THROW(...)` above.

Use `e.code()` to get the iRODS error code from the exception.

server/re/src/irods_re_serialization.cpp
FifthPotato: Yes.

korydraughn: Good enough for me.

Don't worry about testing that part of the code.

server/re/src/irods_re_serialization.cpp
```diff
@@ -1199,6 +1200,102 @@ namespace irods::re_serialization
...
+            return ERROR(INVALID_ANY_CAST,
+                         fmt::format("{}: failed to cast pointer to [execMyRuleInp_t*]: {}", __func__, e.what()));
+        }
+        catch (const irods::exception& e) {
+            return ERROR(
+                SYS_LIBRARY_ERROR,
```
korydraughn, line 1289: `SYS_LIBRARY_ERROR` -> `e.code()`

korydraughn (commented):
Found one last item.

Does the new test pass in isolation?
Is this ready for a full run of the test suite?

FifthPotato:
> Found one last item.
> 
> Does the new test pass in isolation? Is this ready for a full run of the test suite?

Yes, the new test passes in isolation. I think it's ready for a full suite run.

korydraughn:
I've rebased my local branch/copy of this PR on top of 5e5057bdfff711163dcc6d5a26315997da1058a9.

All unit tests passed.

Running core tests.

korydraughn:
All core tests passed.

@FifthPotato Please rebase and squash to taste. No #'s.

FifthPotato:
Rebased, squished

korydraughn (approved):
Pound it!

Good job.

FifthPotato:
Octothorpe'd

korydraughn:
Merging.

alanking (approved):

PR #7694, opened 2024-04-17
[7027] Add null checks to msiSetKeyValuePairsToObj (main)

FifthPotato:
Potentially related to the issue? Not sure if this is exactly what Alan saw. Reproducible via:
`irule "msiSetKeyValuePairsToObj(*x,'$(ipwd)/afile.txt','-d')" null null` (note the nonexistence of `*x`)
Just checking the pointers themselves is not enough-- in testing, I found that the above call actually passes valid pointers into `msiSetKeyValuePairsToObj`, but their contents are empty, so the proceeding `strcmp` calls appear to segfault.

### DIFF
--- a/scripts/irods/test/test_all_rules.py
+++ b/scripts/irods/test/test_all_rules.py
@@ -1754,6 +1754,27 @@ OUTPUT ruleExecOut
         finally:
             self.admin.run_icommand(['iadmin', 'rmgroup', group])
 
+    @unittest.skipUnless(plugin_name == 'irods_rule_engine_plugin-irods_rule_language', "Not implemented for other REPs.")
+    def test_msiSetKeyValuePairsToObj_does_not_crash__issue_7027(self):
+        TEST_FILE = "%s/test_msiSetKeyValuePairsToObj_does_not_crash__issue_7027" % self.admin.session_collection
+        # Create testfile
+        self.admin.assert_icommand(['itouch', TEST_FILE])
+
+        # Invalid argument sets: none of them should crash the agent
+        # First argument is invalid
+        self.admin.assert_icommand(['irule', '-r', 'irods_rule_engine_plugin-irods_rule_language-instance', f"msiSetKeyValuePairsToObj(*x,'{TEST_FILE}','-d')", 'null', 'null'], 'STDERR', ['-316000 USER__NULL_INPUT_ERR'])
+
+        # Second argument is invalid
+        self.admin.assert_icommand(['irule', '-r', 'irods_rule_engine_plugin-irods_rule_language-instance', "*x.key='val'; msiSetKeyValuePairsToObj(*x,*y,'-d')", 'null', 'null'], 'STDERR', ['-316000 USER__NULL_INPUT_ERR'])
+
+        # Third argument is invalid
+        self.admin.assert_icommand(['irule', '-r', 'irods_rule_engine_plugin-irods_rule_language-instance', f"*x.key='val'; msiSetKeyValuePairsToObj(*x,{TEST_FILE},*y)", 'null', 'null'], 'STDERR', ['-316000 USER__NULL_INPUT_ERR'])
+
+        # Valid
+        self.admin.assert_icommand(['irule', '-r', 'irods_rule_engine_plugin-irods_rule_language-instance', f"*x.key='val'; msiSetKeyValuePairsToObj(*x,{TEST_FILE},'-d')", 'null', 'null'])
+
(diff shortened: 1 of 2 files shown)


korydraughn:
Do your changes stop the agent from segfaulting?

What happens if you replace the first `null` with `'*x='`?
what happens if you replace `*x` with an empty string?

FifthPotato:
> Do your changes stop the agent from segfaulting?
> 
> What happens if you replace the first `null` with `'*x='`? what happens if you replace `*x` with an empty string?

Yes, there is no more segfault-- it hits the check and returns an error reported in the logs.

1. Replacing first null with `'*x='` results in USER_PARAM_TYPE_ERR, and fails without crash.
2. Replacing `*x` with an empty string also results the same as 1.

korydraughn:
Because the issue doesn't include information about how to trigger the crash, we'll assume the underlying cause to be lack of null pointer checks.

Let's get a test written for this.

alanking:
> Potentially related to the issue? Not sure if this is exactly what Alan saw.

Unfortunately, Alan from 1 year ago was not particularly helpful in that regard:

> I don't know what I did to make the agent crash

If you found a way to reliably reproduce the agent crash, then that's great and should be the test that we use (in addition to, perhaps, the other things that @korydraughn suggested above). Good work

trel:
potentially `test_irule`?     not a blocker, regardless.

korydraughn (commented):
Let's move the test to test_all_rules.py. Add it to the `Test_AllRules` class.

Once the code changes are applied and the test is moved, I think this will be ready for testing.

scripts/irods/test/test_all_rules.py
```diff
@@ -2408,3 +2429,6 @@ def test_admin_keyword(self):
             admin_session.assert_icommand(['irule', '-r', self.rep_instance, rule_text, 'null', 'ruleExecOut'])
         self.assertEqual(self.new_size, int(lib.get_replica_size(self.user, self.data_name, 0)))
         self.assertEqual(len(self.original_contents), int(lib.get_replica_size(self.user, self.data_name, 1)))
+
+
+
```
korydraughn, line 2434: Remove empty lines.

scripts/irods/test/test_all_rules.py
```diff
@@ -1754,6 +1754,27 @@ def main(rule_args, callback, rei):
...
+        # Second argument is invalid
+        self.admin.assert_icommand(['irule', '-r', 'irods_rule_engine_plugin-irods_rule_language-instance', "*x.key='val'; msiSetKeyValuePairsToObj(*x,*y,'-d')", 'null', 'null'], 'STDERR', ['-316000 USER__NULL_INPUT_ERR'])
+
+        # Third argument is invalid
+        self.admin.assert_icommand(['irule', '-r', 'irods_rule_engine_plugin-irods_rule_language-instance', f"*x.key='val'; msiSetKeyValuePairsToObj(*x,{TEST_FILE},*y)", 'null', 'null'], 'STDERR', ['-316000 USER__NULL_INPUT_ERR'])
+
+        # Valid
+        self.admin.assert_icommand(['irule', '-r', 'irods_rule_engine_plugin-irods_rule_language-instance', f"*x.key='val'; msiSetKeyValuePairsToObj(*x,{TEST_FILE},'-d')", 'null', 'null'])
+
+        self.admin.assert_icommand(['imeta', 'ls', '-d', TEST_FILE], 'STDOUT', [ 'attribute: key\nvalue: val\n', ])
```
korydraughn, line 1776: Trailing comma in list argument (the expected output list).

korydraughn (commented):
Found two things.

Once those are handled, is there anything else left to do for this PR?

FifthPotato:
> Found two things.
> 
> Once those are handled, is there anything else left to do for this PR?

I think this one's pretty much ready for testing.

korydraughn:
Good. Starting up the tests.

korydraughn:
I've rebased my local branch/copy on top of main.

All unit tests passed.

Running core tests.

korydraughn:
All core tests passed.

Please rebase on top of main and squash to taste, no #'s.

korydraughn:
Also, cherry-pick to 4-3-stable.

FifthPotato:
Rebased, squished

korydraughn (approved):
Pound it!

FifthPotato:
Octothorpe'd

korydraughn:
Merging.

korydraughn:
Cherry-pick to 4-3-stable.

Issue #7704, opened 2024-04-18
`irule` crashes when handling a poorly written rule


MartinFlores751:
- [x] main (irods)
- [x] 4-3-stable (irods)
- [x] main (icommands)
- [x] 4-3-stable (icommands)

---

### iRODS Version, OS and Version
Main branch (working towards 4.3.2), Ubuntu 20.04 docker container

### What did you try to do?
I tried to make a simple rule from memory, and ended up with the following `non.r` rule:

```
bleh {
        writeLine("Did I do this right?");
}
OUTPUT
```

Then ran the following command to invoke irule: `irule -F /tmp/non.r`

### Expected behavior
irule gets upset at the improper rule, logs a message if possible, and then exits with a non-zero code.

### Observed behavior (including steps to reproduce, if applicable)
irule segfaults.

#### Following is some detailed information from the core dump:
Backtrace:
```
(lldb) bt
* thread #1, name = 'irule', stop reason = signal SIGSEGV
  * frame #0: 0x000072214324fdfd libc.so.6`___lldb_unnamed_symbol2869 + 589
    frame #1: 0x0000722143c39e19 libirods_common.so.4.90.0`::trimPrefix(str="OUTPUT") at rcMisc.cpp:4421:5
    frame #2: 0x000000000042ead4 irule`main(argc=0, argv=0x0000000000000003) at irule.cpp:329:37
```

There is no info for frame 0.

---
Frame 1:
```
(lldb) frame select 1
frame #1: 0x0000722143c39e19 libirods_common.so.4.90.0`::trimPrefix(str="OUTPUT") at rcMisc.cpp:4421:5
   4418     while ( str[i] == ' ' ) {
   4419         i++;
   4420     }
-> 4421     memmove( str, str + i, strlen( str ) + 1 - i );
   4422     return str;
   4423 }
   4424
(lldb) frame variable
(char *) str = 0x00007ffc6bdceb10 "OUTPUT"
(int) i = 8
```
---
Frame 2:
```
(lldb) frame select 2
frame #2: 0x000000000042ead4 irule`main(argc=0, argv=0x0000000000000003) at irule.cpp:329:37
   326                      }
   327                      else if ( startsWith( buf, "OUTPUT" ) || startsWith( buf, "output" ) ) {
   328                          gotRule = 2;
-> 329                          trimSpaces( trimPrefix( buf ) );
   330                      }
   331                  }
   332 
(lldb) frame variable
(int) argc = 0
[...]

PR #7811, opened 2024-06-13
[#7704] Ensure `trimPrefix(...)` doesn't go past end of string (main)

MartinFlores751:
Minimal patch to fix going past the end of the string.

Can't switch to using `std::isspace(...)` or others, as they would expand what this function originally did, I believe.
Table for reference:
https://en.cppreference.com/w/cpp/string/byte/isspace

### DIFF
--- a/lib/core/include/irods/rcMisc.h
+++ b/lib/core/include/irods/rcMisc.h
@@ -360,9 +360,15 @@ int getAttriInAttriArray(const char* objPath,
                          int* outDataMode,
                          char** outChksum);
 
+// clang-format off
+__attribute__((deprecated))
 char* trimSpaces(char* str);
+// clang-format on
 
+// clang-format off
+__attribute__((deprecated))
 char* trimPrefix(char* str);
+// clang-format on
 
 int convertListToMultiString(char* strInput, int input);
 
--- a/scripts/irods/test/test_irule.py
+++ b/scripts/irods/test/test_irule.py
@@ -1,4 +1,5 @@
 from __future__ import print_function
+import os
 import sys
 if sys.version_info >= (2, 7):
     import unittest
@@ -53,3 +54,18 @@ class Test_Irule(ResourceBase, unittest.TestCase):
         self.assertNotIn( "[1]", stdout )
         self.assertIn( "badInput format error", stderr )
         self.assertNotEqual( rc, 0 )
+
+    @unittest.skipUnless(plugin_name == 'irods_rule_engine_plugin-irods_rule_language', 'only applicable for irods_rule_language REP')
+    def test_irule_does_not_crash_on_bad_rule_file__issue_7740(self):
+        bad_rule = '''
+        test_irule_does_not_crash_on_bad_rule_file__issue_7740 {
+            writeLine("Did I do this right?");
+        }
+        OUTPUT
+        '''
+        path_to_file = os.path.join(self.admin.local_session_dir, 'issue_7740.r')
+
+        with open(path_to_file, 'w') as f:
+            f.write(bad_rule)
+
+        self.admin.assert_icommand_fail(['irule', '-F', path_to_file, '-r', 'irods_rule_engine_plugin-irods_rule_language-instance'], desired_rc=2)


MartinFlores751:
Need to rerun to verify it works as expected, since it's been a while. Also fix up commit message spelling.

trel:
is there a test that can be included for this?   something that breaks before the fix, and would now pass?

MartinFlores751:
Previously it broke in the icommand itself, so I think it's just manual testing. Doesn't seem to be used anywhere inside irods.

alanking:
I think that we should write a test to ensure that the iCommand (`irule`?) itself does not crash.

korydraughn:
You can add C++ unit tests for the function too. That will cover the function's behavior.

MartinFlores751:
Passes in python/icommands test (i.e. no segfault), segfaults in unit test. Think we can keep the unit test only, unless the python test is desired.

trel:
what test?  where?

korydraughn:
Having more tests isn't a bad thing.

I second @trel's question.

MartinFlores751:
> what test? where?

Secret test :nerd_face:

Push up in bit.

korydraughn:
I did a search across the org to see what uses `trimPrefix`. Only `irule` and a project which hasn't been touched in 9 years appeared.

What if we just deprecated the function and updated `irule` to use something safer?

I don't see much value in keeping it since there are better/safer tools available (i.e. the std library, boost, etc).

scripts/irods/test/test_irule.py
```diff
@@ -53,3 +54,19 @@ def test_irule_printVariables_on_stdout_4189(self):
...
+            writeLine("Did I do this right?");
+        }
+        OUTPUT
+        '''
+        path_to_file = os.path.join(self.admin.local_session_dir, 'non.r')
+
+        with open(path_to_file, 'w') as f:
+            f.write(bad_rule)
+
+        self.admin.assert_icommand_fail(['irule', '-F', path_to_file, '-r', 'irods_rule_engine_plugin-irods_rule_language-instance'], desired_rc=1)
```
korydraughn, line 72: Test is fine/good. Just need to tweak a few words.

unit_tests/src/test_process_stash.cpp
```diff
@@ -102,3 +103,10 @@ TEST_CASE("process_stash::erase_if")
     CHECK(irods::process_stash::find(h2));
     CHECK(irods::process_stash::handles().size() == 1);
 }
+
+// NOLINTNEXTLINE(readability-function-cognitive-complexity)
+TEST_CASE("trimPrefix does not go past end of string") {
+	char* testString{"OUTPUT"};
+	trimPrefix(testString);
+	REQUIRE(*testString == '\0');
+}
```
korydraughn, line 112: If we deprecate the function, then we don't need a c++ unit test.

MartinFlores751:
> I did a search across the org to see what uses `trimPrefix`. Only `irule` and a project which hasn't been touched in 9 years appeared.
> 
> What if we just deprecated the function and updated `irule` to use something safer?
> 
> I don't see much value in keeping it since there are better/safer tools available (i.e. the std library, boost, etc).

I think that's fine. ~~Would it matter if the replacement function does more than the original did? Asking since I think most "trim" functions handle more types of whitespace than the original.~~

I think `boost::algorithm::trim_left_copy(...)` can be a drop in replacement. See docs: https://www.boost.org/doc/libs/1_83_0/doc/html/string_algo/usage.html#id-1.3.3.5.5

Would there be a benefit to the tests then? We're not testing the "function in question" at this point, especially if we're deprecating/replacing it.

korydraughn:
That's the correct answer. No need for a C++ unit test.

MartinFlores751:
Still need a test that will execute the new icommands code.

lib/core/include/irods/rcMisc.h
```diff
@@ -360,9 +360,15 @@ int getAttriInAttriArray(const char* objPath,
...
 
+// clang-format off
+__attribute__((deprecated("trimSpaces is deprecated. Use boost's trim instead.")))
 char* trimSpaces(char* str);
+// clang-format on
 
+// clang-format off
+__attribute__((deprecated))
 char* trimPrefix(char* str);
+// clang-format on
```
MartinFlores751, line 371: While this works in this case, do we prefer "comment-based" deprecation?

MartinFlores751 (commented):
Just one question.

unit_tests/src/test_process_stash.cpp
MartinFlores751: Removed.

scripts/irods/test/test_irule.py
MartinFlores751: Which words need tweaking?

lib/core/include/irods/rcMisc.h
```diff
@@ -360,9 +360,15 @@ int getAttriInAttriArray(const char* objPath,
                          int* outDataMode,
                          char** outChksum);
 
+// clang-format off
+__attribute__((deprecated("trimSpaces is deprecated. Use boost's trim instead.")))
```
korydraughn, line 364: We can drop the message since the compiler will state that information for us. There's also no need to speak of boost's trim function since the person may be working in a C-only environment.

scripts/irods/test/test_irule.py
korydraughn: - bleh
- non.r

Consider replacing them with something involving the issue number. That helps us debug test failures.

lib/core/include/irods/rcMisc.h
MartinFlores751: > What do you mean by "comment-based" deprecation?

Something like this: https://github.com/irods/irods/blob/cf4744ecd539f3ea3847f957d8d1121a468100d9/lib/core/include/irods/rcMisc.h#L350-L356

korydraughn: I see. Check the deprecation rules.
- https://docs.irods.org/4.3.2/developers/deprecation/

I don't know if they cover this situation. Let me think about this and get back to you.

korydraughn: Probably need to tweak the deprecation notes regarding always providing a message since deprecation implies avoiding usage (see https://github.com/irods/irods/pull/7811#discussion_r1645232619).

I'll write up an issue for that.

scripts/irods/test/test_irule.py
```diff
@@ -53,3 +54,19 @@ def test_irule_printVariables_on_stdout_4189(self):
         self.assertNotIn( "[1]", stdout )
         self.assertIn( "badInput format error", stderr )
         self.assertNotEqual( rc, 0 )
+
+    @unittest.skipIf(test.settings.TOPOLOGY_FROM_RESOURCE_SERVER or plugin_name=='irods_rule_engine_plugin-python',
+                     'Skip for topology testing from resource server: reads server log')
```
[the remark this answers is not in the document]
MartinFlores751: Set it to only skip based on plugin name. `skipIf(...)` comment/message removed.

lib/core/include/irods/rcMisc.h
MartinFlores751: Removed message.

scripts/irods/test/test_irule.py
MartinFlores751: Renamed.

lib/core/include/irods/rcMisc.h
MartinFlores751: Believe this is answered, leaving open just in case though.

MartinFlores751:
Believe we're just waiting on one test and probably a format commit. For the test, `test_all_rules` failed when ran by itself, but when split (i.e. each test function queued separately) passed.

Rerunning the test to see if I can catch the issue.

lib/core/include/irods/rcMisc.h
alanking: Looks like it has been addressed, yep.

MartinFlores751:
Turns out I wasn't exactly replicating the behavior of `trimSpaces(...)` correctly. I initially thought that it behaved like a 'trim left', but it turns out it trims both sides. `test_all_rules` passed after making the change to use `boost::algorithm::trim_copy_if(...)`.

Rerunning tests in case the change may cause issues elsewhere.

MartinFlores751:
`test_irule` failed from a bad `skipIf(...)` at the top of the test. Fixed it and the test passes now. All core tests passed.

lib/core/include/irods/rcMisc.h
MartinFlores751: Resolving.

korydraughn (commented):
Good. Squash to taste, no #'s.

MartinFlores751:
Squashed.

korydraughn (approved):
Issue number and commit message look good. Pound it.

alanking (approved):

MartinFlores751:
Pounded.

alanking:
Will merge with https://github.com/irods/irods_client_icommands/pull/487 so builds are not interrupted by the deprecation.

korydraughn:
Don't forget to cherry-pick to 4-3-stable.

PR #7815, opened 2024-06-13
[5210] Rewrite make_copy so it stores rule files in memory

FifthPotato:
Instead of what `make_copy` used to do, it stores copies of rule files in-memory with an `unordered_map` mapping `irb` to the content.
`make_copy`'s copies seem to be accessed in two ways:
1. `readRuleSetFromLocalFile` uses `fopen` and friends. Copied the function, except with `fmemopen` for the new version.
2. `hash_rules` uses `std::ifstream` and friends. Copied the function, except with a `std::stringstream`

Seems to work ok at first blush. Taking suggestions on anything.

### DIFF
--- a/plugins/rule_engines/irods_rule_language/src/configuration.cpp
+++ b/plugins/rule_engines/irods_rule_language/src/configuration.cpp
@@ -21,10 +21,17 @@
 #include "irods/private/re/reFuncDefs.hpp"
 #include "irods/Hasher.hpp"
 #include "irods/irods_hasher_factory.hpp"
+#include "irods/irods_exception.hpp"
 
 #include "irods/irods_get_full_path_for_config_file.hpp"
 #include "irods/irods_log.hpp"
 #include <boost/filesystem.hpp>
+#include <memory>
+#include <sstream>
+#include <unordered_map>
+
+using log_re = irods::experimental::log::rule_engine;
+constexpr auto err_buf_len = ERR_MSG_LEN * 1024;
 
 Cache::Cache() : address(NULL), /* unsigned char *address */
     pointers(NULL), /* unsigned char *pointers */
@@ -418,22 +425,59 @@ int hash_rules(const std::vector<std::string> &irbs, const int pid, std::string 
     return 0;
 }
 
-class make_copy {
-public:
-        make_copy(const std::vector<std::string>& _irbs, const int _pid) : irbs_(_irbs), pid_(_pid) {
-            for(auto const &irb : _irbs) {
-                boost::filesystem::copy_file(get_rule_base_path(irb), get_rule_base_path_copy(irb, _pid));
+class in_memory_rulebases
+{
+  public:
+    explicit in_memory_rulebases(const std::vector<std::string>& _irods_rule_bases)
+    {
+        for (auto const& irb : _irods_rule_bases) {
+            std::ifstream ifs(get_rule_base_path(irb));
+            if (!ifs) {
+                log_re::error("in_memory_rulebases: input file stream [{}] failed", get_rule_base_path(irb));
+                // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+                THROW(RULES_FILE_READ_ERROR, fmt::format("Failed to load rulebase [{}] into memory", irb));
             }
+            rule_bases_.insert(
+                {irb, std::string({std::istreambuf_iterator<char>(ifs), std::istreambuf_iterator<char>()})});
         }
-        ~make_copy() {
-            for(auto const &irb : irbs_) {
-                boost::filesystem::remove(get_rule_base_path_copy(irb, pid_));
-            }
-        }
-private:
-        const std::vector<std::string> irbs_;
-        const int pid_;
+    }
+
+    const std::string& get_rulebase(const std::string& irb)
+    {
+        return rule_bases_.at(irb);
+    }
+
+  private:
+    std::unordered_map<std::string, std::string> rule_bases_;
 };
+
+int hash_rules_with_copy(const std::vector<std::string>& irods_rule_bases,
+                         std::string& digest,
+                         in_memory_rulebases& copy)
+{
+    irods::Hasher hasher;
+    irods::error ret = irods::getHasher("md5", hasher);
+    if (!ret.ok()) {
+        int status = static_cast<int>(ret.code());
+        log_re::error("hash_rules_with_copy: cannot get hasher, status = {}", status);
+        return status;
+    }
+
+    for (auto const& irb : irods_rule_bases) {
+        try {
+            hasher.update(copy.get_rulebase(irb));
+        }
+        catch (std::out_of_range& ex) {
+            log_re::warn("hash_rules_with_copy: attempted to access nonexistent rulebase [{}]", irb);
+        }
+    }
+
+    hasher.digest(digest);
+
+    log_re::debug("hash_rules_with_copy: rulebases = [{}]; digest = [{}]", fmt::join(irods_rule_bases, ", "), digest);
+
+    return 0;
+}
 
 int load_rules(const char* irbSet, const std::vector<std::string> &irbs, const int pid, const time_type timestamp) {
                 generateRegions();
@@ -443,33 +487,51 @@ int load_rules(const char* irbSet, const std::vector<std::string> &irbs, const i
                     getSystemFunctions( ruleEngineConfig.sysFuncDescIndex->current, ruleEngineConfig.sysRegion );
                 }
 
-                make_copy copy_rule_base_files(irbs, pid);
-                for(auto const & irb: irbs) {
-                    int i = readRuleStructAndRuleSetFromFile( irb.c_str(), get_rule_base_path_copy(irb, pid).c_str() );
+                try {
+                    in_memory_rulebases stored_rulebases(irbs);
+                    for (auto const& irb : irbs) {
+                        try {
+                            // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
+                            auto* rule_ptr = const_cast<char*>(&stored_rulebases.get_rulebase(irb)[0]);
+                            int i = readRuleStructAndRuleSetFromBuffer(irb.c_str(), rule_ptr);
 
-                    if ( i != 0 ) {
-                        ruleEngineConfig.ruleEngineStatus = INITIALIZED;
-                        return i;
+                            if (i != 0) {
+                                ruleEngineConfig.ruleEngineStatus = INITIALIZED;
+                                return i;
+                            }
+                        }
+                        catch (std::out_of_range& ex) {
+                            log_re::warn("{}: attempted to access nonexistent rulebase [{}]", __func__, irb);
+                        }
                     }
-                }
 
-                createCoreRuleIndex( );
+                    createCoreRuleIndex();
 
-                /* set max timestamp */
-                time_type_set( ruleEngineConfig.timestamp, timestamp );
-                std::string hash;
-                int ret = hash_rules(irbs, pid, hash);
-                if (ret >= 0) {
-                    rstrcpy(ruleEngineConfig.hash, hash.c_str(), CHKSUM_LEN);
-                } else {
+                    /* set max timestamp */
+                    time_type_set(ruleEngineConfig.timestamp, timestamp);
+                    std::string hash;
+                    int ret = hash_rules_with_copy(irbs, hash, stored_rulebases);
+                    if (ret >= 0) {
+                        rstrcpy(static_cast<char*>(ruleEngineConfig.hash), hash.c_str(), CHKSUM_LEN);
+                    }
+                    else {
+                        ruleEngineConfig.ruleEngineStatus = INITIALIZED;
+                        return ret;
+                    }
+                    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-vararg)
+                    snprintf(
+                        static_cast<char*>(ruleEngineConfig.ruleBase), sizeof(ruleEngineConfig.ruleBase), "%s", irbSet);
                     ruleEngineConfig.ruleEngineStatus = INITIALIZED;
-                    return ret;
                 }
-                snprintf( ruleEngineConfig.ruleBase, sizeof( ruleEngineConfig.ruleBase ), "%s", irbSet );
-                ruleEngineConfig.ruleEngineStatus = INITIALIZED;
+                catch (const irods::exception& e) {
+                    log_re::error(
+                        "{} encountered an exception in source file {}:{}", __PRETTY_FUNCTION__, __FILE__, e.what());
+                    return static_cast<int>(e.code());
+                }
                 return 0;
 }
 
+// NOLINTNEXTLINE(readability-function-cognitive-complexity)
 int loadRuleFromCacheOrFile( const char* inst_name, const char *irbSet ) {
     int res = 0;
     auto irbs = parse_irbSet(irbSet);
@@ -522,35 +584,47 @@ int loadRuleFromCacheOrFile( const char* inst_name, const char *irbSet ) {
                     rodsLog( LOG_ERROR, "Failed to restore cache." );
                 } else {
                     int diffIrbSet = strcmp( cache->ruleBase, irbSet ) != 0;
+                    try {
+                        in_memory_rulebases stored_rulebases(irbs);
+                        std::string hash;
+                        int ret = hash_rules_with_copy(irbs, hash, stored_rulebases);
 
-                    make_copy copy_rule_base_files(irbs, pid);
-                    std::string hash;
-                    int ret = hash_rules(irbs, pid, hash);
-
-                    int diffHash = ret < 0 || hash != cache->hash;
-                    if ( diffIrbSet ) {
-                        rodsLog( LOG_DEBUG, "Rule base set changed, old value is %s", cache->ruleBase );
-                    }
-
-                    if ( diffIrbSet || time_type_gt( timestamp, cache->timestamp ) || diffHash ) {
-                        update = 1;
-                        free( cache->address );
-                        rodsLog( LOG_DEBUG, "Rule base set or rule files modified, force refresh." );
-                    } else {
-                        cache->cacheStatus = INITIALIZED;
-                        ruleEngineConfig = *cache;
-
-                        /* generate extRuleSet */
-                        generateRegions();
-                        generateRuleSets();
-                        generateFunctionDescriptionTables();
-                        if ( ruleEngineConfig.ruleEngineStatus == UNINITIALIZED ) {
-                            getSystemFunctions( ruleEngineConfig.sysFuncDescIndex->current, ruleEngineConfig.sysRegion );
+                        int diffHash = static_cast<int>(ret < 0 || hash != static_cast<char*>(cache->hash));
+                        if (static_cast<bool>(diffIrbSet)) {
+                            log_re::debug("Rule base set changed, old value is {}", cache->ruleBase);
                         }
 
-                        ruleEngineConfig.ruleEngineStatus = INITIALIZED;
+                        if (static_cast<bool>(diffIrbSet) || time_type_gt(timestamp, cache->timestamp) ||
+                            static_cast<bool>(diffHash)) {
+                            update = 1;
+                            // NOLINTNEXTLINE(cppcoreguidelines-no-malloc,cppcoreguidelines-owning-memory)
+                            free(cache->address);
+                            log_re::debug("Rule base set or rule files modified, force refresh.");
+                        }
+                        else {
+                            cache->cacheStatus = INITIALIZED;
+                            ruleEngineConfig = *cache;
 
-                        return res;
+                            /* generate extRuleSet */
+                            generateRegions();
+                            generateRuleSets();
+                            generateFunctionDescriptionTables();
+                            if (ruleEngineConfig.ruleEngineStatus == UNINITIALIZED) {
+                                getSystemFunctions(
+                                    ruleEngineConfig.sysFuncDescIndex->current, ruleEngineConfig.sysRegion);
+                            }
+
+                            ruleEngineConfig.ruleEngineStatus = INITIALIZED;
+
+                            return res;
+                        }
+                    }
+                    catch (const irods::exception& e) {
+                        log_re::error("{} encountered an exception in file {}:{}",
+                                      __PRETTY_FUNCTION__,
+                                      __FILE__,
+                                      e.client_display_what());
+                        return static_cast<int>(e.code());
                     }
                 }
         }
@@ -576,8 +650,7 @@ int loadRuleFromCacheOrFile( const char* inst_name, const char *irbSet ) {
     return res;
 }
 int readRuleStructAndRuleSetFromFile( const char *ruleBaseName, const char *rulesBaseFile ) {
-
-    int errloc;
+    int errloc = 0;
     rError_t errmsgBuf;
     errmsgBuf.errMsg = NULL;
     errmsgBuf.len = 0;
@@ -592,6 +665,30 @@ int readRuleStructAndRuleSetFromFile( const char *ruleBaseName, const char *rule
         }
     free( buf );
     freeRErrorContent( &errmsgBuf );
+    return res;
+}
+
+int readRuleStructAndRuleSetFromBuffer(const char* ruleBaseName, char* ruleBase)
+{
+    int errloc = 0;
+    rError_t errmsgBuf;
+    errmsgBuf.errMsg = nullptr;
+    errmsgBuf.len = 0;
+
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays,modernize-avoid-c-arrays)
+    auto buf = std::make_unique<char[]>(err_buf_len * sizeof(char));
+    int res = readRuleSetFromBuffer(ruleBaseName,
+                                    ruleBase,
+                                    ruleEngineConfig.coreRuleSet,
+                                    ruleEngineConfig.coreFuncDescIndex,
+                                    &errloc,
+                                    &errmsgBuf,
+                                    ruleEngineConfig.coreRegion);
+    if (res != 0) {
+        errMsgToString(&errmsgBuf, buf.get(), err_buf_len);
+        log_re::error(buf.get());
+    }
+    freeRErrorContent(&errmsgBuf);
     return res;
 }
 
--- a/plugins/rule_engines/irods_rule_language/src/rules.cpp
+++ b/plugins/rule_engines/irods_rule_language/src/rules.cpp
@@ -61,6 +61,41 @@ int readRuleSetFromLocalFile( const char *ruleBaseName, const char *rulesFileNam
 
     return 0;
 }
+int readRuleSetFromBuffer(const char* ruleBaseName,
+                          char* ruleBody,
+                          RuleSet* ruleSet,
+                          Env* funcDesc,
+                          int* errloc,
+                          rError_t* errmsg,
+                          Region* r)
+{
+    FILE* file = fmemopen(ruleBody, strlen(ruleBody), "r");
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays,modernize-avoid-c-arrays)
+    char errbuf[ERR_MSG_LEN];
+    if (file == nullptr) {
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-type-vararg)
+        snprintf(static_cast<char*>(errbuf), ERR_MSG_LEN, "readRuleSetFromBuffer() could not read rule from buffer");
+        addRErrorMsg(errmsg, RULES_FILE_READ_ERROR, static_cast<char*>(errbuf));
+        return RULES_FILE_READ_ERROR;
+    }
+    Pointer* e = newPointer(file, ruleBaseName);
+    int ret = parseRuleSet(e, ruleSet, funcDesc, errloc, errmsg, r);
+    deletePointer(e);
+    if (ret < 0) {
+        return ret;
+    }
+
+    Node* errnode{};
+    ExprType* restype = typeRuleSet(ruleSet, errmsg, &errnode, r);
+    if (getNodeType(restype) == T_ERROR) {
+        if (nullptr != errnode) {
+            *errloc = NODE_EXPR_POS(errnode);
+        }
+        return RE_TYPE_ERROR;
+    }
+
+    return 0;
+}
 
 int parseAndComputeMsParamArrayToEnv( msParamArray_t *var, Env *env, ruleExecInfo_t *rei, int reiSaveFlag, rError_t *errmsg, Region *r ) {
     int i;
--- a/plugins/rule_engines/irods_rule_language/include/irods/private/re/configuration.hpp
+++ b/plugins/rule_engines/irods_rule_language/include/irods/private/re/configuration.hpp
@@ -141,6 +141,7 @@ int clearResources( int resources );
 int clearCoreRuleIndex( );
 int clearAppRuleIndex( );
 int readRuleStructAndRuleSetFromFile( const char *ruleBaseName, const char *rulesFileName );
+int readRuleStructAndRuleSetFromBuffer(const char* ruleBaseName, char* ruleBase);
 int loadRuleFromCacheOrFile( const char*, const char *irbSet );
 int createCoreRuleIndex( );
 int createAppRuleIndex( );
--- a/plugins/rule_engines/irods_rule_language/include/irods/private/re/rules.hpp
+++ b/plugins/rule_engines/irods_rule_language/include/irods/private/re/rules.hpp
@@ -13,6 +13,13 @@
 
 int readRuleSetFromFile( const char *ruleBaseName, RuleSet *ruleSet, Env *funcDesc, int* errloc, rError_t *errmsg, Region *r );
 int readRuleSetFromLocalFile( const char *ruleBaseName, const char *fileName, RuleSet *ruleSet, Env *funcDesc, int *errloc, rError_t *errmsg, Region *r );
+int readRuleSetFromBuffer(const char* ruleBaseName,
+                          char* ruleBody,
+                          RuleSet* ruleSet,
+                          Env* funcDesc,
+                          int* errloc,
+                          rError_t* errmsg,
+                          Region* r);
 int parseAndComputeMsParamArrayToEnv( msParamArray_t *msParamArray, Env *global, ruleExecInfo_t *rei, int reiSaveFlag, rError_t *errmsg, Region *r );
 int parseAndComputeRule( char *expr, Env *env, ruleExecInfo_t *rei, int reiSaveFlag, rError_t *errmsg, Region *r );
 int parseAndComputeRuleNewEnv( char *expr, ruleExecInfo_t *rei, int reiSaveFlag, msParamArray_t *msParamArray, rError_t *errmsg, Region *r );


plugins/rule_engines/irods_rule_language/src/configuration.cpp
```diff
@@ -420,20 +422,60 @@ int hash_rules(const std::vector<std::string> &irbs, const int pid, std::string
 
 class make_copy {
 public:
-        make_copy(const std::vector<std::string>& _irbs, const int _pid) : irbs_(_irbs), pid_(_pid) {
-            for(auto const &irb : _irbs) {
-                boost::filesystem::copy_file(get_rule_base_path(irb), get_rule_base_path_copy(irb, _pid));
-            }
+  make_copy(const std::vector<std::string>& _irbs)
```
korydraughn, line 425: What does `irbs` stand for? in-memory rulebases?

plugins/rule_engines/irods_rule_language/src/configuration.cpp
```diff
@@ -420,20 +422,60 @@ int hash_rules(const std::vector<std::string> &irbs, const int pid, std::string
...
+      : irbs_(_irbs)
+  {
+      for (auto const& irb : _irbs) {
+          std::string rulebase;
```
korydraughn, line 429: Is this used?

plugins/rule_engines/irods_rule_language/src/configuration.cpp
```diff
@@ -420,20 +422,60 @@ int hash_rules(const std::vector<std::string> &irbs, const int pid, std::string
...
+          std::ifstream ifs(get_rule_base_path(irb));
+          rule_bases.insert(
+              {irb, std::string({std::istreambuf_iterator<char>(ifs), std::istreambuf_iterator<char>()})});
+      }
+  }
+
+  std::string& get_rulebase(const std::string& irb)
+  {
+      return rule_bases.at(irb);
```
korydraughn, line 438: This will throw an exception if `irb` doesn't exist in the vector. Are the calls to `get_rulebase` wrapped in a try-catch block?

plugins/rule_engines/irods_rule_language/src/configuration.cpp
```diff
@@ -420,20 +422,60 @@ int hash_rules(const std::vector<std::string> &irbs, const int pid, std::string
...
+  {
+      for (auto const& irb : _irbs) {
+          std::string rulebase;
+          std::ifstream ifs(get_rule_base_path(irb));
+          rule_bases.insert(
+              {irb, std::string({std::istreambuf_iterator<char>(ifs), std::istreambuf_iterator<char>()})});
+      }
+  }
+
+  std::string& get_rulebase(const std::string& irb)
```
korydraughn, line 436: Are the strings returned by this function meant to be mutable?

plugins/rule_engines/irods_rule_language/src/configuration.cpp
```diff
@@ -420,20 +422,60 @@ int hash_rules(const std::vector<std::string> &irbs, const int pid, std::string
...
+  {
+      return rule_bases.at(irb);
+  }
+
+private:
+  std::unordered_map<std::string, std::string> rule_bases;
```
korydraughn, line 442: `rule_bases` -> `rule_bases_`

plugins/rule_engines/irods_rule_language/src/configuration.cpp
```diff
@@ -420,20 +422,60 @@ int hash_rules(const std::vector<std::string> &irbs, const int pid, std::string
...
-        ~make_copy() {
-            for(auto const &irb : irbs_) {
-                boost::filesystem::remove(get_rule_base_path_copy(irb, pid_));
+
+        if (in_file.eof()) {
+            if (in_file.gcount() > 0) {
+                buffer_read.resize(in_file.gcount());
+                hasher.update(buffer_read);
             }
         }
```
korydraughn, line 472: If the EOF isn't reached, what does that mean for the rulebase?
Does anything care about the hash of these files?
Perhaps we should log a warning?

plugins/rule_engines/irods_rule_language/src/configuration.cpp
korydraughn, line 425: Can we come up with a name for this class which better expresses its intent?

plugins/rule_engines/irods_rule_language/src/configuration.cpp
```diff
@@ -594,6 +636,32 @@ int readRuleStructAndRuleSetFromFile( const char *ruleBaseName, const char *rule
     freeRErrorContent( &errmsgBuf );
     return res;
 }
+int readRuleStructAndRuleSetFromBuffer(const char* ruleBaseName, char* ruleBase)
+{
+    int errloc;
+    rError_t errmsgBuf;
+    errmsgBuf.errMsg = NULL;
```
korydraughn, line 643: `NULL` -> `nullptr`

plugins/rule_engines/irods_rule_language/src/configuration.cpp
```diff
@@ -594,6 +636,32 @@ int readRuleStructAndRuleSetFromFile( const char *ruleBaseName, const char *rule
...
+                                     ruleEngineConfig.coreRuleSet,
+                                     ruleEngineConfig.coreFuncDescIndex,
+                                     &errloc,
+                                     &errmsgBuf,
+                                     ruleEngineConfig.coreRegion)) == 0)
+    {
+    }
+    else {
+        errMsgToString(&errmsgBuf, buf, ERR_MSG_LEN * 1024);
+        rodsLog(LOG_ERROR, "%s", buf);
```
korydraughn, line 659: Let's use the new logger instead of `rodsLog`. `rodsLog` is going to be deprecated soon.

plugins/rule_engines/irods_rule_language/src/configuration.cpp
```diff
@@ -594,6 +636,32 @@ int readRuleStructAndRuleSetFromFile( const char *ruleBaseName, const char *rule
...
     return res;
 }
+int readRuleStructAndRuleSetFromBuffer(const char* ruleBaseName, char* ruleBase)
+{
+    int errloc;
+    rError_t errmsgBuf;
+    errmsgBuf.errMsg = NULL;
+    errmsgBuf.len = 0;
+
+    char* buf = (char*) malloc(ERR_MSG_LEN * 1024 * sizeof(char));
```
korydraughn, line 646: Let's use a C++ container for memory mgmt or `irods::at_scope_exit` to free the memory. That'll keep this code future proof.

plugins/rule_engines/irods_rule_language/src/configuration.cpp
```diff
@@ -594,6 +636,32 @@ int readRuleStructAndRuleSetFromFile( const char *ruleBaseName, const char *rule
...
+    int res = 0;
+    if ((res = readRuleSetFromBuffer(ruleBaseName,
```
korydraughn, line 648: Let's move the assignment out of the if-stmt and let the if-stmt handle the condition. That will reduce complexity for other readers.

plugins/rule_engines/irods_rule_language/src/configuration.cpp
```diff
@@ -594,6 +636,32 @@ int readRuleStructAndRuleSetFromFile( const char *ruleBaseName, const char *rule
...
+                                     ruleBase,
+                                     ruleEngineConfig.coreRuleSet,
+                                     ruleEngineConfig.coreFuncDescIndex,
+                                     &errloc,
+                                     &errmsgBuf,
+                                     ruleEngineConfig.coreRegion)) == 0)
+    {
+    }
```
korydraughn, line 656: How about `!= 0)`?

plugins/rule_engines/irods_rule_language/src/configuration.cpp
```diff
@@ -594,6 +636,32 @@ int readRuleStructAndRuleSetFromFile( const char *ruleBaseName, const char *rule
...
+    else {
+        errMsgToString(&errmsgBuf, buf, ERR_MSG_LEN * 1024);
```
korydraughn, line 658: I saw `ERR_MSG_LEN * 1024` this earlier. Consider capturing that in a `constexpr` variable and using that throughout the impl.

plugins/rule_engines/irods_rule_language/src/configuration.cpp
```diff
@@ -594,6 +636,32 @@ int readRuleStructAndRuleSetFromFile( const char *ruleBaseName, const char *rule
...
+        rodsLog(LOG_ERROR, "%s", buf);
+    }
+    free(buf);
+    freeRErrorContent(&errmsgBuf);
```
korydraughn, line 662: This should be wrapped in an `irods::at_scope_exit` object as well.

plugins/rule_engines/irods_rule_language/src/configuration.cpp
```diff
@@ -594,6 +636,32 @@ int readRuleStructAndRuleSetFromFile( const char *ruleBaseName, const char *rule
     freeRErrorContent( &errmsgBuf );
     return res;
 }
+int readRuleStructAndRuleSetFromBuffer(const char* ruleBaseName, char* ruleBase)
```
korydraughn, line 639: Let's add an empty line above this function.

plugins/rule_engines/irods_rule_language/src/rules.cpp
```diff
@@ -61,6 +61,40 @@ int readRuleSetFromLocalFile( const char *ruleBaseName, const char *rulesFileNam
...
+                          RuleSet* ruleSet,
+                          Env* funcDesc,
+                          int* errloc,
+                          rError_t* errmsg,
+                          Region* r)
+{
+    FILE* file;
+    char errbuf[ERR_MSG_LEN];
+    file = fmemopen(ruleBody, strlen(ruleBody), "r");
+    if (file == NULL) {
```
korydraughn, line 75: `NULL` -> `nullptr`

plugins/rule_engines/irods_rule_language/src/rules.cpp
```diff
@@ -61,6 +61,40 @@ int readRuleSetFromLocalFile( const char *ruleBaseName, const char *rulesFileNam
...
 }
+int readRuleSetFromBuffer(const char* ruleBaseName,
+                          char* ruleBody,
+                          RuleSet* ruleSet,
+                          Env* funcDesc,
+                          int* errloc,
+                          rError_t* errmsg,
+                          Region* r)
+{
+    FILE* file;
```
korydraughn, line 72: Combine with L74.

plugins/rule_engines/irods_rule_language/src/rules.cpp
```diff
@@ -61,6 +61,40 @@ int readRuleSetFromLocalFile( const char *ruleBaseName, const char *rulesFileNam
...
+    int ret = parseRuleSet(e, ruleSet, funcDesc, errloc, errmsg, r);
+    deletePointer(e);
+    if (ret < 0) {
+        return ret;
+    }
+
+    Node* errnode{};
+    ExprType* restype = typeRuleSet(ruleSet, errmsg, &errnode, r);
+    if (getNodeType(restype) == T_ERROR) {
+        if (NULL != errnode) {
```
korydraughn, line 90: `nullptr`

korydraughn (commented):
It's getting there. If you've grown an understanding of the things you've touched, please add comments explaining what's happening. That'll help the next reader of the code a lot.

plugins/rule_engines/irods_rule_language/src/configuration.cpp
trel: probably iRODS rule bases

trel: definitely do not need to keep that nomenclature... strong proposals welcome.

FifthPotato: I believe it's a vector of strings that correspond to the various `.re` files in `/etc/irods` (e.g. `core` for `core.re`), so if I had to hazard a guess, it'd be something like `irods rule bases`? I mostly copied the naming scheme from the existing implementation.

plugins/rule_engines/irods_rule_language/src/configuration.cpp
FifthPotato: No, but originally marking it `const` caused issues with trying to pass it into `fmemopen`, since it expects a non-const input pointer, even if I'm opening it read-only.
I was originally worried I might hit a weird UB case if I took address-of a `const` variable, but some research says it's technically not illegal, so I could do it and maybe just `const_cast` away the const for `fmemopen`. Would that be preferable?

plugins/rule_engines/irods_rule_language/src/configuration.cpp
korydraughn: Should be easy enough to check. Could use a debugger for that.

plugins/rule_engines/irods_rule_language/src/configuration.cpp
korydraughn: No. I believe casting away const on a const object is UB. A comment explaining why the string is a mutable reference is good enough.

FifthPotato: If I recall correctly, we've previously allowed it because it isn't UB unless you access it as if it were non-const. Does that change your mind about the outcome?

plugins/rule_engines/irods_rule_language/src/configuration.cpp
FifthPotato: True, originally this was so I could quickly diagnose if something went wrong. There's no try-catch, so I'm thinking maybe instead of throwing here, I log an error and return an empty string? I never hit this case while testing, and it seems somewhat rare to hit (a function requests a rulebase from the copy that doesn't exist?)
What are your thoughts?

plugins/rule_engines/irods_rule_language/src/configuration.cpp
FifthPotato: `free(buf);` or `freeRErrorContent(...);`?

plugins/rule_engines/irods_rule_language/src/configuration.cpp
FifthPotato: Absolutely! Maybe `stored_rulebases` or `rulebase_files` or `rulebase_map`? Or something else?

plugins/rule_engines/irods_rule_language/src/configuration.cpp
FifthPotato: Personally, it's hard to imagine a case where a `stringstream` doesn't hit EOF-- maybe it means the reads were interrupted above, somehow? I feel like the program's probably in a bad state if it can't read from something stored in memory...Could log a warning for that.

plugins/rule_engines/irods_rule_language/src/configuration.cpp
FifthPotato: Did a quick debug print. Seems to map 1-to-1 to the "re_rulebase_set" array in `server_config.json`. So `iRODS rule bases` definitely seems right here.

trel: excellent.   yes, please use all the letters for a new name, so any future readers of the code don't have to think about this again.

korydraughn: Agreed.

plugins/rule_engines/irods_rule_language/src/rules.cpp
```diff
@@ -61,6 +61,39 @@ int readRuleSetFromLocalFile( const char *ruleBaseName, const char *rulesFileNam
...
+                          Env* funcDesc,
+                          int* errloc,
+                          rError_t* errmsg,
+                          Region* r)
+{
+    FILE* file = fmemopen(ruleBody, strlen(ruleBody), "r");
+    char errbuf[ERR_MSG_LEN];
+    if (file == nullptr) {
+        snprintf(errbuf, ERR_MSG_LEN, "readRuleSetFromBuffer() could not read rule from buffer");
+        addRErrorMsg(errmsg, RULES_FILE_READ_ERROR, errbuf);
```
korydraughn, line 76: Why do we use `addRErrorMsg` instead of the logger here?

plugins/rule_engines/irods_rule_language/src/configuration.cpp
korydraughn: I don't think you can use an empty string because a rulebase file is allowed to be empty.

I think wrapping the calls to `get_rulebase` is a stronger approach. That gives you more options on how to proceed.

Alternatively, you make the function return `std::optional<std::string>` and check the optional at each call site. This approach would solve the `const` situation too, but at the cost of a copy. _The copy can be worked around via `std::ref` or a pointer._

plugins/rule_engines/irods_rule_language/src/rules.cpp
FifthPotato: From what I can tell, it looks to be sort of an old fashioned "stack trace," where the call starting from `readRuleStructAndRuleSetFromBuffer()` down to this function will append messages to the list, and print them out using `rodsLog` if an error occurs. The stack doesn't seem _too_ deep, so I could replace them one-by-one; or, we could keep this and I could just replace the rodsLog at the top. Thoughts?

plugins/rule_engines/irods_rule_language/src/configuration.cpp
```diff
@@ -418,23 +424,54 @@ int hash_rules(const std::vector<std::string> &irbs, const int pid, std::string
...
-                boost::filesystem::copy_file(get_rule_base_path(irb), get_rule_base_path_copy(irb, _pid));
-            }
-        }
-        ~make_copy() {
-            for(auto const &irb : irbs_) {
-                boost::filesystem::remove(get_rule_base_path_copy(irb, pid_));
-            }
-        }
+  in_memory_rulebases(const std::vector<std::string>& _irods_rule_bases)
+      : irods_rule_bases_(_irods_rule_bases)
```
alanking, line 430: Do we need to store `irods_rule_bases_` in this class at all? We are using the passed in `_irods_rule_bases` to populate the `Key`s of `rule_bases_` and never end up using `irods_rule_bases_` anywhere. Even if we did, all of its contents are stored as the `Key`s in the map, so they are still available.

plugins/rule_engines/irods_rule_language/src/configuration.cpp
```diff
@@ -418,23 +424,54 @@ int hash_rules(const std::vector<std::string> &irbs, const int pid, std::string
...
 
+int hash_rules_with_copy(const std::vector<std::string>& irods_rule_bases, std::string& digest, in_memory_rulebases& copy)
+{
+    int status;
+    irods::Hasher hasher;
+    irods::error ret = irods::getHasher("md5", hasher);
+    if (!ret.ok()) {
+        status = ret.code();
+        logger::rule_engine::error("hash_rules_with_copy: cannot get hasher, status = {}", status);
+        return status;
```
alanking, line 457: I think we could remove `status` if we return `ret.code()` here.

Or, we can reduce its scope since I now see it is being used to construct the log message.

plugins/rule_engines/irods_rule_language/src/configuration.cpp
```diff
@@ -418,23 +424,54 @@ int hash_rules(const std::vector<std::string> &irbs, const int pid, std::string
...
+        try {
+            hasher.update(copy.get_rulebase(irb));
+        } catch(std::out_of_range& ex) {
+            logger::rule_engine::error("hash_rules_with_copy: attempted to access nonexistent rulebase [{}]", irb);
+        }
+    }
+    
+    hasher.digest(digest);
+
+    logger::rule_engine::debug("hash_rules_with_copy: rulebases = {}, digest = {}", fmt::join(irods_rule_bases, ","), digest);
```
alanking, line 470: Feel free to ignore this since it's just a debug message...

Consider using a semi-colon or something to separate the two items in the log message here since the rulebases are being joined with commas. In other words...
```suggestion
    logger::rule_engine::debug("hash_rules_with_copy: rulebases = {}; digest = {}", fmt::join(irods_rule_bases, ","), digest);
```

plugins/rule_engines/irods_rule_language/src/configuration.cpp
```diff
@@ -418,23 +424,54 @@ int hash_rules(const std::vector<std::string> &irbs, const int pid, std::string
...
+        status = ret.code();
+        logger::rule_engine::error("hash_rules_with_copy: cannot get hasher, status = {}", status);
+        return status;
+    }
+
+    for (auto const& irb : irods_rule_bases) {
+        try {
+            hasher.update(copy.get_rulebase(irb));
+        } catch(std::out_of_range& ex) {
+            logger::rule_engine::error("hash_rules_with_copy: attempted to access nonexistent rulebase [{}]", irb);
```
alanking, line 464: Is this error serious enough to bail out here? If not, I wonder if this should be a warning message rather than an error. Just a thought.

plugins/rule_engines/irods_rule_language/src/configuration.cpp
```diff
@@ -443,13 +480,18 @@ int load_rules(const char* irbSet, const std::vector<std::string> &irbs, const i
                     getSystemFunctions( ruleEngineConfig.sysFuncDescIndex->current, ruleEngineConfig.sysRegion );
                 }
 
-                make_copy copy_rule_base_files(irbs, pid);
+                in_memory_rulebases copy_rule_base_files(irbs);
```
alanking, line 483: These rulebases are no longer being stored in files, so consider giving this a different name.

plugins/rule_engines/irods_rule_language/src/configuration.cpp
```diff
@@ -443,13 +480,18 @@ int load_rules(const char* irbSet, const std::vector<std::string> &irbs, const i
...
 
-                    if ( i != 0 ) {
-                        ruleEngineConfig.ruleEngineStatus = INITIALIZED;
-                        return i;
+                        if ( i != 0 ) {
+                            ruleEngineConfig.ruleEngineStatus = INITIALIZED;
+                            return i;
+                        }
+                    } catch(std::out_of_range& ex) {
+                        logger::rule_engine::error("load_rules: attempted to access nonexistent rulebase [{}]", irb);
```
alanking, line 494: Same question here about whether this is serious enough to return early with an error. Maybe the answer is no.

plugins/rule_engines/irods_rule_language/src/rules.cpp
alanking: It is a strange usage of an rError stack, but it seems to be correct here. The messages are logged in the calling function, like @FifthPotato said. I'm not sure how I feel about it, but since we are just copying an existing function that is tried and true, maybe we just let this ride for now.

alanking (commented):
Good job navigating this change. This area can get pretty tricky

plugins/rule_engines/irods_rule_language/src/rules.cpp
trel: if we're not sure, and want to keep moving - consider a new issue pointing to this conversation... and we can unravel / discuss there.

FifthPotato: > It is a strange usage of an rError stack, but it seems to be correct here. The messages are logged in the calling function, like @FifthPotato said. I'm not sure how I feel about it, but since we are just copying an existing function that is tried and true, maybe we just let this ride for now.

I'm ambivalent. On one hand, it seems to only extend 1-2 levels deep at most, meaning replacing them all shouldn't be too bad. It also saves a `malloc` and `free`, which counts as modernization. On the other hand, it is tried-and-true. What's the overall risk of replacing these here?

alanking: I think the main risk would be a change in behavior when running rules. The error information returned would need to be done *somehow* to be a valid replacement for the `rError` stack, and even then, the output may differ. I'm not sure if our tests depend on that behavior or not, but that would be the first place we would see it (if there is indeed a change in behavior).

I'm thinking that this implementation should be put into the upcoming release and we make a new issue, like @trel said. That being said, feel free to play around with it and see if the behavior changes.

plugins/rule_engines/irods_rule_language/src/configuration.cpp
```diff
@@ -418,23 +424,56 @@ int hash_rules(const std::vector<std::string> &irbs, const int pid, std::string
...
-            for(auto const &irb : irbs_) {
-                boost::filesystem::remove(get_rule_base_path_copy(irb, pid_));
-            }
+class in_memory_rulebases
+{
+  public:
+    in_memory_rulebases(const std::vector<std::string>& _irods_rule_bases)
+    {
+        for (auto const& irb : _irods_rule_bases) {
+            std::ifstream ifs(get_rule_base_path(irb));
```
korydraughn, line 433: Check `ifs` for errors?
```cpp
if (!ifs) {
    // Handle error.
}
```

plugins/rule_engines/irods_rule_language/src/configuration.cpp
```diff
@@ -418,23 +424,56 @@ int hash_rules(const std::vector<std::string> &irbs, const int pid, std::string
...
+        }
+        catch (std::out_of_range& ex) {
+            logger::rule_engine::warn("hash_rules_with_copy: attempted to access nonexistent rulebase [{}]", irb);
+        }
+    }
+
+    hasher.digest(digest);
+
+    logger::rule_engine::debug(
+        "hash_rules_with_copy: rulebases = {}; digest = {}", fmt::join(irods_rule_bases, ","), digest);
```
korydraughn, line 472: Does `irods_rule_bases` hold filenames or the rule code of each file?

If it holds rule code, then let's change this to list only the filenames. Syslog will truncate the message if it exceeds the max message size.

plugins/rule_engines/irods_rule_language/src/rules.cpp
korydraughn: If we're unsure of that line, then let's do what @trel said and keep it moving.

plugins/rule_engines/irods_rule_language/src/configuration.cpp
FifthPotato: In this context, it should be just rulebase names (e.g. `core`), so it should be fine.

FifthPotato:
All core and unit tests pass (on the last commit, before this most recent one.) I think that's a good vote of confidence, but can re-run if we think it's needed.

plugins/rule_engines/irods_rule_language/src/configuration.cpp
korydraughn: Good.

plugins/rule_engines/irods_rule_language/src/configuration.cpp
```diff
@@ -418,23 +424,60 @@ int hash_rules(const std::vector<std::string> &irbs, const int pid, std::string
...
+    in_memory_rulebases(const std::vector<std::string>& _irods_rule_bases)
+    {
+        for (auto const& irb : _irods_rule_bases) {
+            std::ifstream ifs(get_rule_base_path(irb));
+            if (!ifs) {
+                log_re::warn(
+                    "in_memory_rulebases: input file stream [{}] failed, errno = {}", get_rule_base_path(irb), errno);
             }
+            rule_bases_.insert(
+                {irb, std::string({std::istreambuf_iterator<char>(ifs), std::istreambuf_iterator<char>()})});
```
korydraughn, line 439: What is the outcome of the istreambuf iterators (on L439) when `ifs` is in a bad state?

korydraughn:
> All core and unit tests pass (on the last commit, before this most recent one.) I think that's a good vote of confidence, but can re-run if we think it's needed.

What SHA are you referring to when you say **before this most recent one**? 51eec32?

FifthPotato:
> > All core and unit tests pass (on the last commit, before this most recent one.) I think that's a good vote of confidence, but can re-run if we think it's needed.
> 
> What SHA are you referring to when you say **before this most recent one**? [51eec32](https://github.com/irods/irods/commit/51eec32718bcd15f578cb122772ffac6e192253a)?

[53604d7](https://github.com/irods/irods/pull/7815/commits/53604d7cab75b6c9ee002c1654adc3ff7ccbdd81), the one right before the last review changes I made.

plugins/rule_engines/irods_rule_language/src/configuration.cpp
FifthPotato: Oops, good catch. Fixing...

korydraughn:
> 53604d7, the one right before the last review changes I made.

Got it.

I noticed the addition of the `continue` instruction in the constructor. I'm trying to think through whether the agent (or server?) should continue to run or fail immediately. If I configure a server and one of the files cannot be loaded for whatever reason, my server will not be in the state I expected.

For that reason, I think we have to terminate.

Thoughts?

plugins/rule_engines/irods_rule_language/src/configuration.cpp
```diff
@@ -572,35 +581,43 @@ int loadRuleFromCacheOrFile( const char* inst_name, const char *irbSet ) {
...
 
-                        ruleEngineConfig.ruleEngineStatus = INITIALIZED;
+                            ruleEngineConfig.ruleEngineStatus = INITIALIZED;
 
-                        return res;
+                            return res;
+                        }
+                    }
+                    catch (const irods::exception& e) {
+                        logger::rule_engine::error(
```
trel, line 618: I see `log_re::error` used above?   Do we want / need consistency?

plugins/rule_engines/irods_rule_language/src/configuration.cpp
```diff
@@ -487,36 +488,44 @@ int load_rules(const char* irbSet, const std::vector<std::string> &irbs, const i
...
+                    }
+                    snprintf(ruleEngineConfig.ruleBase, sizeof(ruleEngineConfig.ruleBase), "%s", irbSet);
                     ruleEngineConfig.ruleEngineStatus = INITIALIZED;
-                    return ret;
                 }
-                snprintf( ruleEngineConfig.ruleBase, sizeof( ruleEngineConfig.ruleBase ), "%s", irbSet );
-                ruleEngineConfig.ruleEngineStatus = INITIALIZED;
+                catch (const irods::exception& e) {
+                    logger::rule_engine::error(
+                        "{} encountered an exception in file {}:{}", __PRETTY_FUNCTION__, __FILE__, e.what());
```
alanking, line 526: A couple of things to consider here...

1. `irods::exception::what()` prints out a whole stacktrace and some other pieces that do not format particularly well in our new JSON-based logs. If you think it is helpful to include the stacktrace here for purposes of debugging (especially for user deployments) in the future, then this is fine. If you just want to print the message from the exception and don't care about the other bits, we can use `irods::exception::client_display_what()`.
2. Could we rephrase this log message slightly so that it's clear(er) that the "file" being referred to here is the source code? Maybe it's just me, but I thought this could have perhaps been referring to a rulebase file. In context, it clearly is not, but I figured I would mention it anyway. Also, if we end up using `what()` instead of `client_display_what()`, I believe it already prints out the file and line number in that extra information.

FifthPotato: >     2. Could we rephrase this log message slightly so that it's clear(er) that the "file" being referred to here is the source code?

Do you have any wording suggestions? Maybe something like "...in source file..." or "...in server source file..."?

plugins/rule_engines/irods_rule_language/src/configuration.cpp
korydraughn: Yes, please replace.

korydraughn:
Have we seen how iRODS behaves when the constructor of `in_memory_rulebases` throws an exception?

We need a test to make sure API requests fail immediately following rulebase initialization failures.

FifthPotato:
> Have we seen how iRODS behaves when the constructor of `in_memory_rulebases` throws an exception?
> 
> We need a test to make sure API requests fail immediately following rulebase initialization failures.

I ran a few tests. If I cause a condition where my code is hit, it indeed prints the messages I expect it to, but does not stop the API from responding-- things such as `ils` still respond.
That being said, I caused a similar failure condition while testing on tip-of-main. I added a nonexistent rule into `re_rulebase_set`, but things like `ils` still function.
This is me being an iRODS noob question for sure, but given the NREP's status as a plugin, would it be expected for an exceptional condition to be a showstopper for the whole server?

plugins/rule_engines/irods_rule_language/src/configuration.cpp
```diff
@@ -474,8 +474,7 @@ int hash_rules_with_copy(const std::vector<std::string>& irods_rule_bases,
 
     hasher.digest(digest);
 
-    logger::rule_engine::debug(
-        "hash_rules_with_copy: rulebases = {}; digest = {}", fmt::join(irods_rule_bases, ","), digest);
+    log_re::debug("hash_rules_with_copy: rulebases = {}; digest = {}", fmt::join(irods_rule_bases, ","), digest);
```
korydraughn, line 477: Please wrap the curly braces in square brackets (i.e. `{}` -> `[{}]`). That helps detect empty strings and has become a convention throughout iRODS code.

korydraughn, line 477: Add space after comma for second argument to `fmt::join`. Improves readability.
```cpp
fmt::join(irods_rule_bases, ",")` -> `fmt::join(irods_rule_bases, ", ")
```

plugins/rule_engines/irods_rule_language/src/configuration.cpp
```diff
@@ -501,7 +500,7 @@ int load_rules(const char* irbSet, const std::vector<std::string> &irbs, const i
                             }
                         }
                         catch (std::out_of_range& ex) {
-                            logger::rule_engine::warn("load_rules: attempted to access nonexistent rulebase [{}]", irb);
+                            log_re::warn("load_rules: attempted to access nonexistent rulebase [{}]", irb);
```
korydraughn, line 503: Let's use `__func__` or `__FUNCTION__`, etc for capturing the function name in log messages. That helps keep the log message and function name in sync (in case of renames).

korydraughn:
> This is me being an iRODS noob question for sure, but given the NREP's status as a plugin, would it be expected for an exceptional condition to be a showstopper for the whole server?

Yes. If you configure a system to do something and it cannot initialize to the expected/proper state, that means you cannot be sure of its behavior. For iRODS, even though the NREP is a plugin, it is a crucial component to the operation of the system.

Happy to hear counter-arguments.

alanking:
FWIW, if you configure a non-existent rulebase in 4.3.2 as released, the server appears to continue to function, which is consistent with what @FifthPotato reported.

korydraughn:
Sure. I don't think we should continue with that behavior though.

FifthPotato:
> Sure. I don't think we should continue with that behavior though.

Since this is the behavior that is currently observed, do we consider these changes to be in scope for this issue? From what I can tell, the onus for "exiting with failure" falls with wherever the main server loads the NREP plugin and not the plugin itself. If you think it is in scope, I can continue looking into implementing that.

korydraughn:
Opening up a new PR is perfectly fine.

Write up an issue capturing what the plan is and let's get it implemented while it's fresh in your mind.

The issue can link to comments in this PR for additional context too. That way we know what sparked the change.

With that said, I see one unresolved conversation. Let's get that taken care of and run the test suite - assuming we're ready for testing.

FifthPotato:
All unit and core tests pass. Last convo item was just discussing the use of rErrorStack, but we decided to let it go for now. I'll write up an issue for it.

alanking (commented):
To complete the loop, the new issues are https://github.com/irods/irods/issues/7862 and https://github.com/irods/irods/issues/7863.

Seems like this is ready for squashing, so please do so to your liking, and omit #s for now.

FifthPotato:
Squished.

korydraughn:
Please look through the clang-tidy report and resolve the error cases which involve lines you touched.

korydraughn:
Just one more application of clang-format and we may be ready.

plugins/rule_engines/irods_rule_language/src/configuration.cpp
```diff
@@ -418,23 +425,61 @@ int hash_rules(const std::vector<std::string> &irbs, const int pid, std::string
...
+class in_memory_rulebases
+{
+  public:
+    explicit in_memory_rulebases(const std::vector<std::string>& _irods_rule_bases)
+    {
+        for (auto const& irb : _irods_rule_bases) {
+            std::ifstream ifs(get_rule_base_path(irb));
+            if (!ifs) {
+                log_re::error(
+                    "in_memory_rulebases: input file stream [{}] failed, errno = {}", get_rule_base_path(irb), errno);
```
korydraughn, line 437: Can we truly rely on `errno` to reflect what happened with the C++ file stream?

korydraughn (commented):
Looks good. Just one question.

FifthPotato:
> Looks good. Just one question.

Updated with removed `errno`.

korydraughn (approved):
I think we're good. Let's pound it.

FifthPotato:
Octothorpe'd

alanking:
@FifthPotato - Please cherry-pick to 4-3-stable. Thanks!

Issue #3902, opened 2018-04-25
replace GenQuery ad-hoc parsing code with flex/bison parser


trel:
## Feature

This would replace the existing parsing code with a 'real' parser.  An example implementation can be found at https://github.com/kellerb/genquery-ast-example.

Replacing the parser would close issues:
- #3886 
- #3064
- https://github.com/irods/irods/issues/4697
- #4983 
- https://github.com/irods/irods/issues/5734
- https://github.com/irods/irods/issues/5992

trel:
Should define the parser via ANTLR...
 - https://tomassetti.me/getting-started-antlr-cpp/

trel:
Probably sticking with flex/bison...

korydraughn:
Latest reference implementation for GenQuery replacement can be found here:
- https://github.com/irods/irods_api_plugin_genquery2

korydraughn:
#7570 must be resolved before taking on this issue.

korydraughn:
The GenQuery2 parser was merged into upstream (#7570) but is too new to start replacing code within the server.

GenQuery1 will continue to exist for the duration of the 4.x series _(perhaps 5.x too for backwards compatibility reasons)_. Its usage will be replaced over time as confidence in the new parser grows.

korydraughn:
After some discussion, we've decided to ship #7819 post 4.3.3. We'll also introduce an environment variable to control whether the new parser is used. The environment variable would give admins a way to opt into the new parser (default would be off).

korydraughn:
I'm now starting to wonder if the opt-in solution is worse than switching to the new parser.

My current thoughts ...
- Environment variable can lead to misalignment between client and server
    - Server has environment variable set, client does not
- Do clients have to care about the opt-in flag?
    - Yes, if they have paths with embedded single quotes
- What if the client/server libs always hex-escaped single quotes? Are there any downsides to this?
    - If the client never receives a query with embedded single quotes, things work
        - The paths will not be altered
    - If the client sends a query with embedded single quotes and the server isn't using the new parser, an error will occur
        - The paths must be escaped for things to work
    - If the client sends a query with embedded single quotes and the server is using the new parser, things work
- What does this affect?
    - `iquest`
    - `irods::query`
    - `irods::query_builder`
    - clients using `fillGenQueryInpFromStrCond`
- Isn't this an issue for genquery2 as well?
    - We have to migrate to it eventually
    - And it will require the embedded single quotes to be escaped 
- Possible solutions
[...]

PR #7819, opened 2024-06-19
[#3902] Replace GenQuery1 ad-hoc parsing code with flex/bison parser (4-3-stable)

korydraughn:
I've now watched all unit/core tests pass with these changes. However, those test runs did not include Ubuntu 24 changes so I'll need to re-run the tests after updating my environment.

There are a few more things to handle/discuss:
- Review CMake (and understand why the `add_subdirectory` approach didn't work)
- Changes to error codes/messages returned by `iquest` (see tests)
- Update iquest tests to cover embedded single quotes and perhaps other special characters
- Add C++ unit tests for `irods::query[_builder]` demonstrating handling of embedded single quotes and perhaps other special characters
- Add doxygen documentation to the C function wrapping the new GenQuery1 parser

### An important note

This PR introduces a stronger parser which may result in some queries breaking. The reason for this is due to the parser correctly handling embedded single quotes, etc.

Scripts depending on the error code names will also fail due to how flex/bison works. I may be able to restore backwards compatibility, but it will take more studying of flex/bison to get it right. No guarantees though.


lib/genquery1/CMakeLists.txt
```diff
@@ -0,0 +1,71 @@
...
+#)
+#add_dependencies(irods_common_core flex_bison_generation)
+
+target_sources(
+  irods_common_core
+  PRIVATE
+  ${CMAKE_CURRENT_SOURCE_DIR}/src/genquery1_driver.cpp
+  ${${IRODS_FLEX_OUTPUTS}}
+  ${${IRODS_BISON_OUTPUTS}}
+)
```
korydraughn, line 34: Neither of these work because there's nothing to trigger the generation of the parser files.

lib/core/CMakeLists.txt
```diff
@@ -104,6 +108,36 @@ target_compile_definitions(
...
+set(IRODS_FLEX_OUTPUTS "FLEX_${IRODS_FLEX_LEXER_NAME}_OUTPUTS")
+set(IRODS_BISON_OUTPUTS "BISON_${IRODS_BISON_PARSER_NAME}_OUTPUTS")
+
+target_sources(
+  irods_common_core
+  PRIVATE
+  "${CMAKE_IRODS_SOURCE_DIR}/lib/genquery1/src/genquery1_driver.cpp"
+  ${${IRODS_FLEX_OUTPUTS}}
+  ${${IRODS_BISON_OUTPUTS}}
+)
```
korydraughn, line 138: This is just a temporary solution. See the comments for the genquery1 CMakeLists.txt file.

lib/CMakeLists.txt
```diff
@@ -14,6 +14,7 @@ add_subdirectory(administration)
 add_subdirectory(hasher)
 add_subdirectory(rbudp)
 add_subdirectory(metadata)
+#add_subdirectory(genquery1)
```
korydraughn, line 17: Don't forget to handle this. Requires discussion first.

lib/core/CMakeLists.txt
```diff
@@ -86,6 +87,9 @@ target_include_directories(
   "$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include>"
   "$<BUILD_INTERFACE:${CMAKE_IRODS_SOURCE_DIR}/lib/api/include>"
   "$<BUILD_INTERFACE:${CMAKE_IRODS_SOURCE_DIR}/lib/filesystem/include>"
+  #"$<BUILD_INTERFACE:${CMAKE_IRODS_BINARY_DIR}/lib/genquery1/dsl>" # For parser.hpp.
+  "$<BUILD_INTERFACE:${CMAKE_IRODS_BINARY_DIR}/lib/core/dsl>" # For parser.hpp.
+  "$<BUILD_INTERFACE:${CMAKE_IRODS_SOURCE_DIR}/lib/genquery1/include>"
```
korydraughn, line 92: Don't forget to handle this. Requires discussion first.

scripts/irods/test/test_iquest.py
```diff
@@ -108,14 +108,15 @@ def test_iquest_resc_hier_with_like__3714(self):
         self.admin.assert_icommand("irm -f " + filename)
         remove_large_hierarchy(self, LEAF_COUNT, directory)
 
+    # TODO This test can be removed now that iquest relies on a real parser for GenQuery1.
     def test_iquest_matches_names_containing_apostrophes_via_the_equals_operator__issue_4887(self):
         data_object = "data' object"
         self.admin.assert_icommand(['istream', 'write', data_object], input='hello, world')
-        self.admin.assert_icommand(['iquest', "select DATA_NAME where DATA_NAME = '{0}'".format(data_object)], 'STDOUT', ['DATA_NAME = ' + data_object])
+        self.admin.assert_icommand(['iquest', "select DATA_NAME where DATA_NAME = '{0}'".format(data_object.replace("'", '\\x27'))], 'STDOUT', ['DATA_NAME = ' + data_object])
```
korydraughn, line 115: Notice the use of `.replace`. This is a breaking change depending on how you view things.

scripts/irods/test/test_iquest.py
```diff
@@ -824,8 +825,8 @@ def test_invalid_operations(self):
                 for vop in valid_operations:
                     # First, with the invalid operation as the first clause of the ||...
                     query = f'{self.base_query} {iop} \'{pattern}\' || {vop} \'{pattern}\''
-                    self.user.assert_icommand(['iquest', '%s', query], 'STDERR', 'CAT_INVALID_ARGUMENT')
+                    self.user.assert_icommand(['iquest', '%s', query], 'STDERR', 'INPUT_ARG_NOT_WELL_FORMED_ERR')
```
korydraughn, line 828: This is a change in behavior. Users relying on the error message output instead of the code returned to the shell will need to adjust their scripts.

korydraughn:
I've rebased this work on top of 4-3-stable (39e875992ec9ee996ed8ef814c06d661d3b56b7b) and all unit/core tests passed against ub20/pg14.

lib/core/src/lsUtil.cpp
```diff
@@ -11,7 +11,7 @@
 #include "irods/irods_query.hpp"
 #include "irods/escape_utilities.hpp"
 
-#include <boost/format.hpp>
```
SwooshyCueb, line 13: This commit might go with issue 7763

korydraughn:
This PR will likely require changes to ALL plugins using `irods::query[_builder]` and any queries involving `DATA_NAME`/`COLL_NAME` based conditions _(i.e. we have to escape embedded single quotes)_.

An easy/safe way to provide this new parser is to introduce an environment variable for controlling which version of the GenQuery1 parser to use. Then this becomes an opt-in and we don't have to worry about introducing temporary server_config.json options.

trel:
all of that last comment makes me somewhat nervous as we're preparing for a release.

korydraughn:
After some discussion, we've decided to ship this post 4.3.3. We'll also introduce an environment variable to control whether the new parser is used. The environment variable would give admins a way to opt into the new parser (default would be off).

korydraughn:
Given the plan is to get this PR merged, we need to consider testing.

The following comes from my previous comment:

> We'll also introduce an environment variable to control whether the new parser is used. The environment variable would give admins a way to opt into the new parser (default would be off).

Exposing an opt-in option means the amount of testing needed for a release doubles. This also means the test infrastructure must learn about the option. Documentation would need to be updated/added too. Even if we labeled the new parser as experimental, we couldn't skip testing it because we eventually want to make it the default.

Thoughts?

alanking:
> Exposing an opt-in option means the amount of testing needed for a release doubles. This also means the test infrastructure must learn about the option. Documentation would need to be updated/added too. Even if we labeled the new parser as experimental, we couldn't skip testing it because we eventually want to make it the default.

It may be acceptable to create some "holes" in our test matrix through selective testing. Rather than adding a third dimension to the usual OS/DB combinations, we could ensure that each GenQuery1 parser is tested on each OS and each DB, but not necessarily on each OS/DB combination. The goal would be to ensure that each parser has been tested against each DB flavor at least once and each OS at least once. I think that would be sufficient.

korydraughn:
I think that's a reasonable approach. Once it becomes the default GenQuery1 parser, we'll have to make sure it works.

Here's another approach ... extract all string-based queries from the iRODS source and use those to confirm each parser produces matching results. The new parser would have more tests because it covers the special characters cases. The tests could also be adapted for GenQuery2 as well.

lib/genquery1/dsl/parser.y
```diff
@@ -0,0 +1,228 @@
...
+    return drv.lexer.yylex(drv);
+} // yylex
+
+auto find_column_id_by_name(std::string_view _column) -> int
+{
+    for (int i = 0; i < NumOfColumnNames; ++i) {
+        if (const auto& entry = columnNames[i]; _column == entry.columnName) {
+            return entry.columnId;
+        }
+    }
```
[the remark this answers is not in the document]
korydraughn: This is fine. By keeping this here, it makes it easier for the reader to find/read.

lib/genquery1/dsl/parser.y
```diff
@@ -0,0 +1,228 @@
...
+
+    auto find_column_id_by_name(std::string_view _column) -> int;
+
+    auto add_column_to_projection_list(irods::experimental::genquery1::driver& _drv,
+                                       std::string_view _column,
+                                       std::optional<std::string_view> _fn = std::nullopt) -> void;
+
+    auto add_condition(irods::experimental::genquery1::driver& _drv,
+                       std::string_view _column,
+                       std::string_view _condition) -> void;
```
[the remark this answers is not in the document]
korydraughn: These functions do not need comments. Their function names express their intent pretty well.

korydraughn:
Need to add environment variable check to tests and c++ code which uses the hex escaping function for single quotes.

korydraughn:
The work for GenQuery1 will not be part of the iRODS 4 series. This work will be part of iRODS 5 only.

See conversation starting at https://github.com/irods/irods/issues/3902#issuecomment-2539820047.

Will close this once the PR for the main branch exists.

korydraughn:
Opened #8089.

Closing.

Issue #7816, opened 2024-06-14
Regression: Replication fails for users in groups with permissions on data object but with no permissions for the user itself


alanking:
- [x] main
- [x] 4-3-stable

---

### iRODS Version, OS and Version
4.3.2

### What did you try to do?
Give `modify_object` permission to a group - let's call it `group_a`. As a user - let's say `alice` - with no permissions on the object but is a member of `group_a`, attempt to replicate the data object to another resource.

### Expected behavior
Replication occurs successfully.

### Observed behavior (including steps to reproduce, if applicable)
Replication fails with error `SYS_USER_NO_PERMISSION`.

This is due to a permissions check in the replication API which assumes that because `alice` has no permissions on the data object, `alice` cannot replicate the data object. The permissions check needs to take `alice`'s groups into account as well. Here is the check:
https://github.com/irods/irods/blob/cf4744ecd539f3ea3847f957d8d1121a468100d9/server/api/src/rsDataObjRepl.cpp#L480-L536

This behavior changed as a result of addressing #7444 and #7465.

Thanks to @deanlinssen for the help in finding this.

trel:
So we didn't have a test for this case...    good find!

Issue #5593, opened 2021-05-04
Investigate assertions made in test_iget_with_purgec


alanking:
- [x] main
- [x] 4-3-stable

---

I saw a failure for the test `test_resource_types.Test_Resource_ReplicationToTwoCompoundResourcesWithPreferArchive.test_iget_with_purgec` but upon closer inspection was not sure whether it was asserting the right thing:
```
...

 --- IrodsSession: icommand executed by [otherrods] [iput purgecgetfile.txt] --- 
Assert Command: iput purgecgetfile.txt
Expecting EMPTY: ['']
  stdout:
    | 
  stderr:
    | 
Output found

 --- IrodsSession: icommand executed by [otherrods] [iget -f --purgec purgecgetfile.txt] --- 
Assert Command: iget -f --purgec purgecgetfile.txt
Expecting EMPTY: ['']
  stdout:
    | 
  stderr:
    | 
Output found

 --- IrodsSession: icommand executed by [otherrods] [ils -L purgecgetfile.txt] --- 
Assert FAIL Command: ils -L purgecgetfile.txt
Expecting STDOUT_SINGLELINE: [' 0 ', 'purgecgetfile.txt']
  stdout:
    |   otherrods         0 demoResc;compResc2;cacheResc2           54 2021-05-03.20:40 & purgecgetfile.txt
    |         generic    /var/lib/irods/cacheResc2Vault/home/otherrods/2021-05-04Z00:40:06--irods-testing-Tyx6Eb/purgecgetfile.txt
    |   otherrods         1 demoResc;compResc2;archiveResc2           54 2021-05-03.20:40 & purgecgetfile.txt
    |         generic    /var/lib/irods/archiveResc2Vault/home/otherrods/2021-05-04Z00:40:06--irods-testing-Tyx6Eb/purgecgetfile.txt
    |   otherrods         3 demoResc;compResc1;archiveResc1           54 2021-05-03.20:40 & purgecgetfile.txt
    |         generic    /var/lib/irods/archiveResc1Vault/home/otherrods/2021-05-04Z00:40:06--irods-testing-Tyx6Eb/purgecgetfile.txt
  stderr:
    | 
Output found

FAILED TESTING ASSERTION
...
```

[...]

PR #7827, opened 2024-06-19
[#7816] Check group permissions for replication (main)

alanking:
Addresses #7816

New tests shown to fail appropriately before change and pass after change. Full core test suite running now.

Submitting as draft for now because I'm debating whether we need to add tests for federation or for the case where the user has multiple groups with differing permission levels on the same object (perhaps with only 1 being sufficient to replicate the data object). I just noticed that I had added some federation tests in the other changes so those could be adjusted similarly to how these tests were adjusted.


server/api/src/rsDataObjRepl.cpp
```diff
@@ -477,62 +481,88 @@ namespace
...
+                fmt::format("select DATA_ACCESS_TYPE where USER_NAME = '{}' and USER_ZONE = '{}' and DATA_ID = '{}'",
+                            user.name,
+                            user.zone,
+                            _obj->data_id());
+            auto user_permission_query = irods::query{&_comm, user_permission_query_string};
+            if (!user_permission_query.empty() &&
+                permission_is_sufficient_to_replicate_object(std::stoi(user_permission_query.front()[0])))
+            {
+                return true;
             }
```
korydraughn, line 526: Is there a reason the permission check couldn't be done by the database? I think you can add an additional condition (i.e `DATA_ACCESS_TYPE = '<int_value>'`) and avoid the use of `std::stoi` altogether.

Not required, just pointing it out because it reduces the number of operations and removes an exception case.

server/api/src/rsDataObjRepl.cpp
```diff
@@ -477,62 +481,88 @@ namespace
...
+
+            // If the user does not have sufficient permissions on the data object for replication, perhaps one of their
+            // groups does...
+            const auto groups = adm::server::groups(_comm, user);
+            std::vector<std::string> groups_with_quotes;
```
korydraughn, line 531: Add line to reserve the size of the vector containing quoted group names.
```cpp
groups_with_quotes.reserve(groups.size());
```

korydraughn (commented):
Looks correct.

Awaiting test results.

server/api/src/rsDataObjRepl.cpp
alanking: That's a good point. In that case, we also wouldn't need to loop through results for the group permissions because any result would indicate sufficient permissions.

I'm also now noticing that `permission_is_sufficient_to_replicate_object` does not account for `create_token`, `delete_token`, or `curate` permissions. However, `ichmod` does not advertise these permission levels and we don't seem to have tests for them, so I guess it's fine?

Anyway, with that in mind, do you have a preference between these two?

1. Leveraging the linear permission model:
```
DATA_ACCESS_TYPE >= '1120'
```

2. Checking for specific permission types:
```
DATA_ACCESS_TYPE = '1120' || = '1130' || = '1140' || = '1150' || = '1160' || = '1200'
```

I prefer 1, I think, because we assume the linear permission model in many places.

server/api/src/rsDataObjRepl.cpp
alanking: Sure thing.

By the way, I tried to make some `libfmt` and `std::vector` magic happen here so it would be a bit cleaner but this is the best I could come up with for now. :)

alanking (commented):
Saw a test failure, so I'll have to look into that. Very possibly will be fixed with the suggested changes

server/api/src/rsDataObjRepl.cpp
alanking: I went ahead and implemented 1. We can iterate on that.

server/api/src/rsDataObjRepl.cpp
alanking: Done

scripts/irods/test/test_irepl.py
```diff
@@ -2010,7 +2033,30 @@ def test_insufficient_permissions_results_in_no_replication_and_an_error(self):
...
+        user = self.user1.username
+        self.insufficient_permissions_results_in_no_replication_and_an_error_impl(user)
+
+    def test_insufficient_permissions_for_group_results_in_no_replication_and_an_error(self):
+        user = self.user1.username
+        group = 'testgroup'
+
+        try:
+            # Create a test group and add the test user to the group. The test group will have permissions on the data
+            # object while the user itself has none.
```
FifthPotato, line 2046: I think this comment is accurate, but it may be potentially misleading if you're checking for insufficient permissions. Maybe something like "The test group will be checked for permissions..." or similar?

FifthPotato (commented):
Looks pretty good! I only have minor things to comment on.

server/api/src/rsDataObjRepl.cpp
```diff
@@ -479,60 +483,51 @@ namespace
...
-            return false;
-        }
+        // If the user does not have sufficient permissions on the data object for replication, perhaps one of their
+        // groups does...
+        const auto groups = adm::server::groups(_comm, user);
+        std::vector<std::string> groups_with_quotes;
+        groups_with_quotes.reserve(groups.size());
+        std::for_each(groups.cbegin(), groups.cend(), [&groups_with_quotes](const auto& group) -> void {
+            groups_with_quotes.emplace_back(fmt::format("'{}'", group.name));
+        });
```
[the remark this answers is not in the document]
alanking: I wasn't quite sure how to get that work here. Can you provide a rough sketch of what that would look like?

scripts/irods/test/test_irepl.py
alanking: True, for the case where no permissions are granted to the group or the user. Does this wording make it better?
```suggestion
            # Create a test group and add the test user to the group. The test group will be checked for permissions on
            # the data object while the user itself will have no permissions on the data object.
```

trel: i like it

server/api/src/rsDataObjRepl.cpp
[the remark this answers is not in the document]
MartinFlores751: I think it would be:
```cpp
std::transform(std::cbegin(groups), std::cend(groups), std::back_inserter(groups_with_quotes), [](const auto& val) -> std::string {
  return fmt::format("'{}'", val);
});
```

Didn't try it though.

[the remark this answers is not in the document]
FifthPotato: > I think it would be:
> 
> ```c++
> std::transform(std::cbegin(groups), std::cend(groups), std::back_inserter(groups_with_quotes), [](const auto& val) -> std::string {
>   return fmt::format("'{}'", val);
> });
> ```
> 
> Didn't try it though.

Yeah, what Martin wrote looks right to me. Alternatively, if you want to avoid the auto resize with the vector, I think you can resize and then pass in the destination iterator.

```cpp
groups_with_quotes.reserve(groups.size());
std::transform(std::cbegin(groups), std::cend(groups), std::begin(groups_with_quotes), [](const auto& val) {
  return fmt::format("'{}'", val);
});
```
The benefit of this one is that I think a parallel implementation of `std::transform` can work better (though I doubt that'll ever be an issue...)

scripts/irods/test/test_irepl.py
FifthPotato: I like it too!

server/api/src/rsDataObjRepl.cpp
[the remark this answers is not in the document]
alanking: Ahh, this is actually just what I was looking for. I'll try this out and report back.

scripts/irods/test/test_irepl.py
alanking: Done

server/api/src/rsDataObjRepl.cpp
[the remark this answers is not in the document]
alanking: Given that this line is just putting apostrophes around some strings for a query (inexpensive relative to running the query), I don't think we need to worry about parallel implementations. After some discussion, I think the `std::back_inserter` solution will work fine here. I've created a separate issue for discussion in our copious free time for a deeper analysis of the approaches: https://github.com/irods/irods/issues/7832

And for future readers... The destination `std::vector` used with `std::transform` needs to have its elements constructed ahead of time using `resize()` rather than `reserve()`!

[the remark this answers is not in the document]
korydraughn: Agreed. Good investigation.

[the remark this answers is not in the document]
alanking: I've confirmed that the `std::back_inserter` / `std::vector::reserve` solution works, so I pushed that up. Thanks for the suggestion and analysis so far. Resolving.

alanking:
Unit tests, core tests, and topology-on-provider tests are passing. Topology-on-consumer is currently running, but I expect these to pass as well.

alanking:
Saw two failures in topology-on-consumer...

`test_resource_types.Test_Resource_ReplicationToTwoCompoundResourcesWithPreferArchive.test_iget_with_purgec`
`test_resource_types.Test_Resource_ReplicationToTwoCompoundResourcesWithPreferArchive.test_irepl_with_purgec`

I have reason to believe that both are related to this issue: https://github.com/irods/irods/issues/5593

alanking:
I think that the test failures I saw are unrelated to this change. I am thinking this is ready to go, pending any further review comments (and of course squashing).

server/api/src/rsDataObjRepl.cpp
```diff
@@ -479,60 +483,52 @@ namespace
...
+        const auto groups = adm::server::groups(_comm, user);
+        std::vector<std::string> groups_with_quotes;
+        groups_with_quotes.reserve(groups.size());
+        std::transform(std::cbegin(groups),
+                       std::cend(groups),
+                       std::back_inserter(groups_with_quotes),
+                       [](const auto& group) -> std::string { return fmt::format("'{}'", group.name); });
+        const auto group_permission_query_string =
+            fmt::format("select DATA_ACCESS_TYPE where USER_NAME in ({}) and USER_TYPE = 'rodsgroup' and DATA_ID = "
+                        "'{}' and DATA_ACCESS_TYPE >= '{}'",
```
korydraughn, line 524: Please add a test which proves the `in (group1, group2, ..., groupN)` case works for this if you haven't already.

alanking: I'll add a test especially for this.

alanking:
Force-push was a rebase / squash. Will push additional changes shortly.

server/api/src/rsDataObjRepl.cpp
alanking: I've added 2 tests which add the user to 10 groups. In one test, no group has sufficient permission to replicate the data object, but there exists a group of which the test user is not a member which does have sufficient permission to replicate the data object. In the other test, every group has sufficient permission to replicate the data object, and there exists a group of which the test user is not a member which does not have sufficient permission to replicate the data object. The test user itself has no permission on the data object at all in either test.

In this way, we show that groups unrelated to the client user do not impact its ability to replicate and that multiple groups having permissions (or not) are handled appropriately. Did I miss anything?

korydraughn: > In the other test, every group has sufficient permission to replicate the data object, and there exists a group of which the test user is not a member which does not have sufficient permission to replicate the data object.

I think this may need to be tweaked such that only one group has sufficient permissions to replicate the data object. That would further prove that the `(g1, g2, ..., gN)` syntax is working as expected. I don't think we can do more in the python because we can't control the order of the group names returned by the user administration lib.

Overall, what you've described is good.

_Thinking about this a little more, what I'm actually attempting to address in this test is whether GenQuery1 handles the IN-clause correctly and that's really out-of-scope for this PR._

alanking: > I think this may need to be tweaked such that only one group has sufficient permissions to replicate the data object. That would further prove that the `(g1, g2, ..., gN)` syntax is working as expected. I don't think we can do more in the python because we can't control the order of the group names returned by the user administration lib.

The existing tests could be modified to make the user a member of more than one group as we are doing in the 2 new tests. The existing tests are set up in such a way that only one group's permissions are being modified (we could pick one in the "middle"?) but that may satisfy the gap identified here.

However, I don't think the gap is necessarily exercising anything new. The query here always includes *every group* (perhaps we should include some limit here to prevent the query from getting too long, if anything?) and filters based on the access permission value in the same query. If any result is returned at all, that means at least one group of which the user is a member has sufficient permission to replicate the data object, and it doesn't matter which group or what permission level it is. In other words, the order of the results doesn't matter for this case. ...unless I'm missing the point.

alanking: I've created an issue for making a test for the `IN` clause specifically: https://github.com/irods/irods/issues/7851 Resolving this.

korydraughn: Excellent.

korydraughn (approved):
Squash to taste and pound it.

alanking:
Squashed, #'d, and mergin

PR #7830, opened 2024-06-21
[7760] templatize irodsServerMonPerf (4-3-stable)

trel:


alanking (commented):
Seems good.

We will also need to update the testing environment and/or `test_all_rules.Test_AllRules` to mv the `.template` file back to the regular path so it can be used for tests. Would like to put those changes in at about the same time so that testing isn't interrupted.

trel:
Probably cp rather than mv.   But, yes to the rest.

SwooshyCueb:
> We will also need to update the testing environment and/or `test_all_rules.Test_AllRules` to mv the `.template` file back to the regular path so it can be used for tests. Would like to put those changes in at about the same time so that testing isn't interrupted.

Here's the approach I would take:
- Remove [`rulemsiServerMonPerf.r`](https://github.com/irods/irods_client_icommands/blob/52023e6713b420b05a792cac8d6505adb3c21b60/test/rules/rulemsiServerMonPerf.r) from icommands
- Restore the `test_all_rules.Test_AllRules.test_msiServerMonPerf_default_3736` test removed in irods/irods@52023e6713b420b05a792cac8d6505adb3c21b60 and irods/irods@79e88b06e4912349aa53ace08379eea0079c2f57
- Add `msiServerMonPerf("verbose","default");` to `test_all_rules.Test_AllRules.test_msiServerMonPerf_default_3736`
- Have `test_all_rules.Test_AllRules.test_msiServerMonPerf_default_3736` copy the template before running `irule` (if needed), and then remove the copy when done in a `finally` block (if it wasn't there before)

alanking:
I can take a swing at @SwooshyCueb's approach. Seems reasonable.

alanking:
I was able to implement the solution, but ran into an issue when I looked into things a little more closely. I've captured the problem in a separate issue: https://github.com/irods/irods/issues/7842

Should I push commits on top of this PR or open new PRs which will act as companion PRs to this?

alanking:
I have created two PRs as companion PRs for this:
https://github.com/irods/irods_client_icommands/pull/490
https://github.com/irods/irods/pull/7843

All 3 PRs should be merged at around the same time so that testing is not interrupted.

alanking:
https://github.com/irods/irods_client_icommands/pull/490 and https://github.com/irods/irods/pull/7843 have been approved, so I think we are ready to add #s here.

trel:
#'d

Issue #7760, opened 2024-05-23
Disable-by-default `irodsServerMonPerf`


SwooshyCueb:
- [x] main
- [x] 4-3-stable

---

`msiServerMonPerf` is a microservice that invokes `irodsServerMonPerf`, a *perl* script that hasn't seen a lot of love in recent years. Unless I'm mistaken, this is the only perl in the entire irods project. It's very stinky, and has very little documentation as to what it does, how it works, how to use it, etc. It can also be broken in various ways that cause `msiServerMonPerf` to fail silently.

Let's consider taking one or more of the following actions:
- Deprecate `msiServerMonPerf` in 4.3 and remove both `msiServerMonPerf` and `irodsServerMonPerf` in 5.0.
- Replace `msiServerMonPerf` and/or `irodsServerMonPerf` with something better.
- Disable `msiServerMonPerf` by default.

trel:
At minimum, we will move `irodsServerMonPerf` to be a template file (off by default).

alanking:
`irodsServerMonPerf` is now disabled by default. #7757 and #7842 will address issues with the script and microservice. Closing this one.

PR #7843, opened 2024-06-26
[#7760] Stage irodsServerMonPerf for msiServerMonPerf test (4-3-stable)

alanking:
In service of #7760 

Companion PR: https://github.com/irods/irods/pull/7830
Companion PR: https://github.com/irods/irods_client_icommands/pull/490

### DIFF
--- a/scripts/irods/test/test_all_rules.py
+++ b/scripts/irods/test/test_all_rules.py
@@ -4,14 +4,11 @@ import json
 import os
 import shutil
 import socket
+import stat
 import sys
 import tempfile
 import textwrap
-
-if sys.version_info >= (2, 7):
-    import unittest
-else:
-    import unittest2 as unittest
+import unittest
 
 from .. import lib
 from .. import test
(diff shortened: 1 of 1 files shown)


scripts/irods/test/test_all_rules.py
```diff
@@ -489,6 +486,46 @@ def test_writeLine_config_first_3477(self):
             IrodsController().start()
 
 
+    @unittest.skipUnless(plugin_name == 'irods_rule_engine_plugin-irods_rule_language', 'only applicable for irods_rule_language REP')
+    def test_msiServerMonPerf_default_3736(self):
+        path_to_irodsServerMonPerf = '/'.join([paths.home_directory(), 'msiExecCmd_bin', 'irodsServerMonPerf'])
```
SwooshyCueb, line 491: Consider using [`os.path.join`](https://docs.python.org/3.5/library/os.path.html#os.path.join) instead.

alanking: Oh, good point. I'd gotten into this habit due to PRC running on Windows and then trying to construct iRODS logical paths. This is referring to a path on the local filesystem, so `os.path.join` should be better here.

alanking: Done

scripts/irods/test/test_all_rules.py
```diff
@@ -489,6 +486,46 @@ def test_writeLine_config_first_3477(self):
...
+    @unittest.skipUnless(plugin_name == 'irods_rule_engine_plugin-irods_rule_language', 'only applicable for irods_rule_language REP')
+    def test_msiServerMonPerf_default_3736(self):
+        path_to_irodsServerMonPerf = os.path.join(paths.home_directory(), 'msiExecCmd_bin', 'irodsServerMonPerf')
+        path_to_irodsServerMonPerf_template = path_to_irodsServerMonPerf + '.template'
+        irodsServerMonPerf_already_existed = os.path.exists(path_to_irodsServerMonPerf)
+
+        rule_file="test_msiServerMonPerf.r"
+        rule_string= '''
+msiTestServerMonPerf {{
+    msiServerMonPerf("verbose", "default");
```
SwooshyCueb, line 498: The test rule file in the icommands repo being replaced by this test runs it both ways, so we probably should here too.

```suggestion
    msiServerMonPerf("default", "default");
    msiServerMonPerf("verbose", "default");
```

alanking: Also a good point. Done.

korydraughn (approved):
Seems good to me.

alanking:
Added #s. Will merge with https://github.com/irods/irods/pull/7830

PR #7854, opened 2024-06-27
[#5593] Fix test_iget/irepl_with_purgec assumptions (main)

alanking:
Addresses #5593 

Only ran the modified tests (many times in a row, each) in order to confirm that they are now passing consistently.


korydraughn (approved):
If the tests are passing consistently, pound it.

Issue #7859, opened 2024-07-02
Invoking `msiDataObjChksum` via the PREP can lead to segfault following checksum verification


korydraughn:
- [x] main
- [x] 4-3-stable
---
## Bug Report

The following code can lead to a segfault when `chksum` is `NULL`. `chksum` will be set to `NULL` when verification is performed and everything is in sync.
https://github.com/irods/irods/blob/d7ff9f4f42d78b1480531af7a4805f51f2584554/server/re/src/reDataObjOpr.cpp#L1623-L1626

However, as the process makes its way back up the call stack, `convertFromMsParam` is invoked which is where things fall apart. Specifically at L66. Notice the argument passed to the `std::string` constructor. Passing a null pointer to `std::string` is undefined behavior.
https://github.com/irods/irods/blob/d7ff9f4f42d78b1480531af7a4805f51f2584554/server/re/src/irods_re_plugin.cpp#L63-L67 

### Solutions

There are two ways to deal with this.
- (A) We can handle the null pointer in `msiDataObjChksum`
    - This fixes the issue only for the MSI
- (B) Or, we adjust `convertFromMsParam` to handle null pointers
    - This affect several things and could lead to incorrect behavior

Option (A) is the safest approach.

_Found as a result of the conversation at https://groups.google.com/d/msgid/irod-chat/5af1c699-2f3d-4bcc-80fe-ae7a4c94c7edn%40googlegroups.com?utm_medium=email&utm_source=footer._

PR #7861, opened 2024-07-02
[#7859] Convert null pointer to empty string for checksum output parameter of msiDataObjChksum (4-3-stable)

korydraughn:
Tested manually at the bench.

Just need to confirm the test is good and run the full test suite.

### DIFF
--- a/scripts/irods/test/test_all_rules.py
+++ b/scripts/irods/test/test_all_rules.py
@@ -15,6 +15,7 @@ from .. import test
 from .. import paths
 from ..configuration import IrodsConfig
 from ..controller import IrodsController
+from ..core_file import temporary_core_file
 from . import metaclass_unittest_test_case_generator
 from . import resource_suite
 from . import session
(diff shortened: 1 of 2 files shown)


korydraughn:
Updated test.

Unit tests pass. Running core tests.

korydraughn:
Forgot to mention I verified the test fails when the fix is not included.

korydraughn:
All core tests pass.

scripts/irods/test/test_all_rules.py
```diff
@@ -1818,6 +1819,41 @@ def test_msiSetKeyValuePairsToObj_does_not_crash__issue_7027(self):
...
+
+            # Trigger the PEP.
+            # Before the fix, this would result in an error.
+            self.user0.assert_icommand(['itrim', data_object], 'STDOUT', ['Number of files trimmed = 0.'])
+
+            # Show the checksum verification was successful by searching for the expected attribute units.
+            self.user0.assert_icommand(
+                ['iquest', f"select META_DATA_ATTR_UNITS where META_DATA_ATTR_NAME = '{attr_key}' and META_DATA_ATTR_VALUE = '{attr_key}'"],
+                'STDOUT',
+                [f'META_DATA_ATTR_UNITS = {expected_attr_units}\n'])
```
alanking, line 1856: The following suggestion is completely optional. :)

There is a function called `metadata_attr_with_value_exists` in `lib.py` which does something *almost* like this, and I think could act as a valid replacement for these lines. The AVU being annotated in the rule for this test is using the same attribute and value string, and a unique-ish string in the units field. If we put the unique-ish string in the value field instead and drop the units, we could use `metadata_attr_with_value_exists` here, which selects the `META_DATA_ATTR_VALUE` instead of `META_DATA_ATTR_UNITS`. ...unless there's a reason for using the units field.

scripts/irods/test/test_all_rules.py
```diff
@@ -1818,6 +1819,41 @@ def test_msiSetKeyValuePairsToObj_does_not_crash__issue_7027(self):
...
+            def pep_api_data_obj_trim_pre(rule_args, callback, rei):
+                logical_path = str(rule_args[2].objPath)
+                result = callback.msiDataObjChksum(logical_path, 'verifyChksum=', '')
+                if result['arguments'][2] == '':
+                    callback.msiModAVUMetadata('-d', logical_path, 'set', '{attr_key}', '{attr_key}', '{expected_attr_units}')
+            '''))
+
+            # Create a new data object.
+            data_object = 'issue_7859_data_object'
+            self.user0.assert_icommand(['itouch', data_object])
```
alanking, line 1842: Do we need to clean up this data object or the metadata at the end?

alanking (commented):
Looks good. Just had suggestions for the test...

scripts/irods/test/test_all_rules.py
korydraughn: No, it's handled for us because the data object is automatically placed in the session collection.

scripts/irods/test/test_all_rules.py
korydraughn: I originally used that function, but decided to go in another direction because it doesn't include the target data object. AFAIK, that function only checks that the AVU exists in the database. It doesn't assert that the AVU is attached to the data object.

I plan to open an issue for that today.

alanking: Ah, I noticed that as well. However, this is also not asserting that the metadata is attached to the data object, though, so perhaps we should do that here?

korydraughn: You're right lol. Definitely need to include the data object.

korydraughn: Fixed.

alanking (commented):
Looks good. Squash em

korydraughn:
I just realized that the core test run did not cover this PR. I have to use the plugin builder and test hook for that.

Will have to re-run the test suite with the PREP to be sure nothing's broken.

korydraughn:
All PREP tests passed.

korydraughn:
#'d

alanking (approved):

PR #7878, opened 2024-07-06
[#7909,irods/irods_rule_engine_plugin_python#182] Add genquery2 tests for genquery python module (4-3-stable)

korydraughn:
Companion PR for irods/irods_rule_engine_plugin_python#206.

Leaving as draft until the PREP PR is done.

Current tests pass. Feel free to review them.

### DIFF
--- a/scripts/irods/test/test_prep_genquery_iterator.py
+++ b/scripts/irods/test/test_prep_genquery_iterator.py
@@ -832,3 +832,153 @@ class Test_Genquery_Iterator(resource_suite.ResourceBase, unittest.TestCase):
 
         super(Test_Genquery_Iterator, self).tearDown()
 
+    @unittest.skipUnless(plugin_name == 'irods_rule_engine_plugin-python', 'Requires PREP.')
+    def test_genquery_constructor_raises_exception_on_invalid_parser_argument__issue_7909(self):
+        with temporary_core_file() as core:
+            attr_name = 'test_genquery_constructor_raises_exception_on_invalid_parser_argument__issue_7909_error'
+
+            # The following rule triggers and attaches the exception message generated
+            # by the invalid parser argument to the Query constructor.
+            core.add_rule(dedent(f'''
+            def pep_api_touch_pre(rule_args, callback, rei):
+                try:
+                    import genquery
+                    _ = genquery.Query(callback, 'COLL_NAME', parser=None)
+                except ValueError as e:
+                    callback.msiModAVUMetadata('-C', '{self.admin.session_collection}', 'add', '{attr_name}', str(e), '')
+            '''))
+
+            # Trigger the PEP.
+            # This will cause the rule to add an AVU to the session collection.
+            self.admin.assert_icommand(['itouch', 'issue_7909'], 'STDERR')
+
+            # Show the AVU containing the error exists on the session collection.
+            self.admin.assert_icommand(['imeta', 'ls', '-C', self.admin.session_collection], 'STDOUT', [
+                f'attribute: {attr_name}\n',
+                'value: Invalid value for [parser]. Expected Parser.GENQUERY1 or Parser.GENQUERY2.\n'
+            ])
+
+    @unittest.skipUnless(plugin_name == 'irods_rule_engine_plugin-python', 'Requires PREP.')
+    def test_genquery_constructor_raises_exception_when_genquery2_syntax_is_passed_to_genquery1_parser__issue_7909(self):
+        with temporary_core_file() as core:
+            attr_name = 'test_genquery_constructor_raises_exception_when_genquery2_syntax_is_passed_to_genquery1_parser__issue_7909_error'
+
+            # The following rule triggers and attaches the exception message generated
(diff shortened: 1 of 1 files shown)


scripts/irods/test/test_prep_genquery_iterator.py
```diff
@@ -832,3 +832,117 @@ def tearDown(self):
...
+
+    @unittest.skipUnless(plugin_name == 'irods_rule_engine_plugin-python', 'Requires PREP.')
+    def test_genquery_iterator_supports_genquery2__issue_7909(self):
+        rule_file = f'{self.admin.local_session_dir}/test_genquery_iterator_supports_genquery2__issue_7909.r'
+
+        with open(rule_file, 'w') as rf:
+            rf.write(dedent(f'''
+            def main(rule_args, callback, rei):
+                from genquery import Query, Parser
+                for r in Query(callback, 'COLL_NAME', "COLL_NAME = '{self.admin.session_collection}'", parser=Parser.GENQUERY2):
```
alanking, line 870: Any value in trying to use some GenQuery2-only features in these tests?

Should we add a test which attempts to use GenQuery2-only features with the `Parser.GENQUERY1` parser? Not for the sake of testing the parser, but more to make sure the right parser is being used (in this case, by watching it break when the syntax for the other parser is in use).

Maybe there isn't any value to that.

scripts/irods/test/test_prep_genquery_iterator.py
```diff
@@ -832,3 +832,117 @@ def tearDown(self):
...
+            INPUT null
+            OUTPUT ruleExecOut
+            '''))
+
+        # Execute the rule.
+        rep_instance = IrodsConfig().default_rule_engine_plugin + '-instance'
+        expected_output = ['count: 7']
+        self.admin.assert_icommand(['irule', '-r', rep_instance, '-F', rule_file], 'STDOUT', expected_output)
+
+        # Test limit functionality.
```
alanking, line 907: I feel like it would be cleaner to split these tests up, but having them combined is also fine.

korydraughn: Oh, you're talking about separating the testing of `limit` and `offset`?

No problem. Will separate.

scripts/irods/test/test_prep_genquery_iterator.py
korydraughn: If the GenQuery2 interface was different, I think it would be good to test that. But since GenQuery2 just takes a single string, I don't think there's much value in testing that.

Most of the GenQuery2 logic in the module just ignores GenQuery1 stuff.

With that said, do you see value in adding a test for that? What would the test do?

scripts/irods/test/test_prep_genquery_iterator.py
korydraughn: Separated them into two distinct tests.

scripts/irods/test/test_prep_genquery_iterator.py
korydraughn: After some discussion, we've agreed that adding a test to verify the correct parser is being used has value.

scripts/irods/test/test_prep_genquery_iterator.py
alanking: Yes, sorry, I should have clarified. Thanks

scripts/irods/test/test_prep_genquery_iterator.py
```diff
@@ -832,3 +832,155 @@ def tearDown(self):
...
+            def main(rule_args, callback, rei):
+                from genquery import Query, Parser
+                q1 = Query(callback, 'COLL_NAME', "COLL_NAME = '{self.admin.session_collection}'", parser=Parser.GENQUERY2)
+                q2 = q1.copy()
+                q3 = q2.copy()
+                iterations = 0
+                for query in [q1, q2, q3]:
+                    for r in query:
+                        callback.writeLine('stdout', f'[iteration={{iterations}}] row: {{r}}')
+                    iterations += 1
```
alanking, line 975: If you want to get an automatic integer iterator value, you can use `enumerate` here:
```suggestion
                for iterations, query in enumerate([q1, q2, q3]):
                    for r in query:
                        callback.writeLine('stdout', f'[iteration={{iterations}}] row: {{r}}')
```

scripts/irods/test/test_prep_genquery_iterator.py
alanking: New test looks good

alanking (commented):
Changes look good. Had one suggestion, but otherwise this feels right

scripts/irods/test/test_prep_genquery_iterator.py
korydraughn: Nice. Will update

korydraughn: Done.

alanking (commented):
Looks good to me. I say squash em

korydraughn:
Squashed.

korydraughn:
Yep. Pounded.

Issue #7870, opened 2024-07-04
Improve documentations for `msi_atomic_apply_acl_operations` and `msi_atomic_apply_metadata_operations`


mstfdkmn:
- [x] main
- [x] 4-3-stable
---
### iRODS Version, OS and Version
4.3.2, almalinux8

### What did you try to do?
I tried to run these msis in my custom rules.

### Expected behavior
I expected to see examples like other msis in https://github.com/irods/irods_rule_engine_plugin_python/tree/main/python_rules and in https://github.com/irods/irods_client_icommands/tree/main/test/rules. Also the doxygen doesnt contain enough explanation for how to pass `admin_mode` and `_json_output`.

### Observed behavior (including steps to reproduce, if applicable)
I learned how to pass the parameters with trial and error learning (and looking into .h files).

PR #7891, opened 2024-07-10
[#7870] Improve documentation for atomic acls/metadata microservices (main)

korydraughn:
Example code works as intended.

### DIFF
--- a/plugins/microservices/src/atomic_apply_acl_operations.cpp
+++ b/plugins/microservices/src/atomic_apply_acl_operations.cpp
@@ -86,6 +86,7 @@ auto plugin_factory() -> irods::ms_table_entry*
 /// \p _json_input must have the following JSON structure:
 /// \code{.js}
 /// {
+///   "admin_mode": boolean,
 ///   "logical_path": string,
 ///   "operations": [
 ///     {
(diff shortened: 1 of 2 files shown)


plugins/microservices/src/atomic_apply_acl_operations.cpp
```diff
@@ -126,6 +127,26 @@ auto plugin_factory() -> irods::ms_table_entry*
 /// \retval 0        On success.
 /// \retval non-zero On failure.
 ///
+/// \b Example
+/// \code{.py}
+/// # The collection to apply the ACL operations to.
+/// *lpath = '/tempZone/home/alice';
```
trel, line 133: does a trailing slash matter here? is it worth noting if it DOES matter?

korydraughn: It's not supposed to matter, but I bet the current implementation of the API doesn't handle it. Will give it a try.

I don't think it's worth mentioning. The user shouldn't need to worry about trailing slashes. We have a function which takes care of that.

korydraughn: Confirmed the API doesn't handle trailing slashes.

Will write up and fix as part of this PR.

korydraughn: Well, I'll write an issue.

korydraughn: Created #7892.

Resolving since that is out-of-scope of this PR.

korydraughn:
#'d

korydraughn:
Merging.

PR #7893, opened 2024-07-11
[7408] Atomic ACLs consider zone of user (main)

FifthPotato:
First shot at this. Any comments welcome.

### DIFF
--- a/plugins/api/src/atomic_apply_acl_operations.cpp
+++ b/plugins/api/src/atomic_apply_acl_operations.cpp
@@ -83,7 +83,8 @@ namespace
 
     auto to_access_type_id(std::string_view _acl) -> int;
 
-    auto get_entity_id(nanodbc::connection& _db_conn, std::string_view _entity_name) -> id_type;
+    auto get_entity_id(nanodbc::connection& _db_conn, std::string_view _entity_name, std::string_view _entity_zone)
+        -> id_type;
 
     auto entity_has_acls_set_on_object(nanodbc::connection& _db_conn,
                                        const std::string_view _db_instance_name,
@@ -351,13 +352,15 @@ namespace
     }
 
     // TODO This function should probably deal with remote zones.
-    auto get_entity_id(nanodbc::connection& _db_conn, std::string_view _entity_name) -> id_type
+    auto get_entity_id(nanodbc::connection& _db_conn, std::string_view _entity_name, std::string_view _entity_zone)
+        -> id_type
     {
         nanodbc::statement stmt{_db_conn};
 
-        prepare(stmt, "select user_id from R_USER_MAIN where user_name = ?");
+        prepare(stmt, "select user_id from R_USER_MAIN where user_name = ? and zone_name = ?");
 
         stmt.bind(0, _entity_name.data());
+        stmt.bind(1, _entity_zone.data());
 
         if (auto row = execute(stmt); row.next()) {
             return row.get<id_type>(0);
@@ -378,7 +381,12 @@ namespace
             throw_if_invalid_acl(acl);
 
             log::api::trace("Retrieving entity ID ...");
-            const auto entity_id = get_entity_id(_db_conn, _op.at("entity_name").get<std::string>());
+
+            std::string_view zone = getLocalZoneName();
+            if (const auto iter = _op.find("zone_name"); iter != std::end(_op)) {
+                zone = iter->get_ref<const std::string&>();
+            }
+            const auto entity_id = get_entity_id(_db_conn, _op.at("entity_name").get<std::string>(), zone);
             throw_if_invalid_entity_id(entity_id);
 
             if (acl == "null") {
--- a/plugins/api/src/atomic_apply_acl_operations.cpp
+++ b/plugins/api/src/atomic_apply_acl_operations.cpp
@@ -351,7 +351,6 @@ namespace
         }
     }
 
-    // TODO This function should probably deal with remote zones.
     auto get_entity_id(nanodbc::connection& _db_conn, std::string_view _entity_name, std::string_view _entity_zone)
         -> id_type
     {
@@ -383,7 +382,7 @@ namespace
             log::api::trace("Retrieving entity ID ...");
 
             std::string_view zone = getLocalZoneName();
-            if (const auto iter = _op.find("zone_name"); iter != std::end(_op)) {
+            if (const auto iter = _op.find("zone"); iter != std::end(_op)) {
                 zone = iter->get_ref<const std::string&>();
             }
             const auto entity_id = get_entity_id(_db_conn, _op.at("entity_name").get<std::string>(), zone);
@@ -407,7 +406,8 @@ namespace
         }
         catch (const irods::exception& e) {
             log::api::error({{"log_message", e.what()}, {"acl_operation", _op.dump()}});
-            return {e.code(), irods::to_bytes_buffer(make_error_object(_op, _op_index, e.what()).dump())};
+            return {
+                e.code(), irods::to_bytes_buffer(make_error_object(_op, _op_index, e.client_display_what()).dump())};
         }
         catch (const fs::filesystem_error& e) {
             log::api::error({{"log_message", e.what()}, {"acl_operation", _op.dump()}});
--- a/unit_tests/src/test_atomic_apply_acl_operations.cpp
+++ b/unit_tests/src/test_atomic_apply_acl_operations.cpp
@@ -11,6 +11,7 @@
 #include "irods/irods_at_scope_exit.hpp"
 #include "irods/transport/default_transport.hpp"
 #include "irods/dstream.hpp"
+#include "irods/zone_administration.hpp"
 
 #include <nlohmann/json.hpp>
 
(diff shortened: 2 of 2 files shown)


plugins/api/src/atomic_apply_acl_operations.cpp
```diff
@@ -378,7 +379,11 @@ namespace
             throw_if_invalid_acl(acl);
 
             log::api::trace("Retrieving entity ID ...");
-            const auto entity_id = get_entity_id(_db_conn, _op.at("entity_name").get<std::string>());
+            auto zone = std::string{getLocalZoneName()};
+            if(_op.find("zone_name") != std::end(_op)) {
+                zone = _op.at("zone_name").get<std::string>();
+            }
```
korydraughn, line 385: Perhaps this can be done without allocating memory?

plugins/api/src/atomic_apply_acl_operations.cpp
```diff
@@ -378,7 +379,11 @@ namespace
...
+            const auto entity_id = get_entity_id(_db_conn, _op.at("entity_name").get<std::string>(), zone);
```
korydraughn, line 386: `.get` -> `.get_ref` ?

korydraughn (commented):
Looking good.

We'll need a test for this once you're convinced the implementation is correct.

plugins/api/src/atomic_apply_acl_operations.cpp
```diff
@@ -351,13 +351,14 @@ namespace
     }
 
     // TODO This function should probably deal with remote zones.
```
trel, line 353: this TODO can go... once we have a test to confirm remote zones are handled.

plugins/api/src/atomic_apply_acl_operations.cpp
FifthPotato: I think switching to `std::string_view` should work here-- it has a constructor for `char *` and shouldn't allocate any more memory.

plugins/api/src/atomic_apply_acl_operations.cpp
FifthPotato: Since I switched to `std::string_view` above, `.get_ref` doesn't work anymore.

plugins/api/src/atomic_apply_acl_operations.cpp
korydraughn: I'm nervous about replacing std::string with string_view on the `.get` call. The nlohmann JSON library offers `.get_ref` for situations where the data is readonly.

What do you think about this?
```cpp
std::string_view zone = getLocalZoneName();
if (const auto iter = _op.find("zone_name"); iter != std::end(_op)) {
    zone = iter->get_ref<const std::string&>();
}
```

plugins/api/src/atomic_apply_acl_operations.cpp
korydraughn: Check the API docs for nlohmann JSON to see if string_view is acceptable. If you can't find any documentation for it, open an issue in the nlohmann JSON github repo about using string_view with `.get` and `.get_ref`.

plugins/api/src/atomic_apply_acl_operations.cpp
FifthPotato: Seems good. I'll add it!

plugins/api/src/atomic_apply_acl_operations.cpp
korydraughn: @FifthPotato @FifthPotato Let's capture the need for a federation test in an issue. You can include a link to the TODO comment.

You can implement a non-federated test in c++.

korydraughn:
What's left for this PR?

FifthPotato:
> What's left for this PR?

Just tests. Trying to get a federation test going that adequately tests this.

plugins/api/src/atomic_apply_acl_operations.cpp
FifthPotato: > @FifthPotato @FifthPotato Let's capture the need for a federation test in an issue. You can include a link to the TODO comment.
> 
> You can implement a non-federated test in c++.

Oops, missed this comment. I'll pivot to getting a non-federated test going in C++, and make an issue for the federation test.

FifthPotato: Just to be sure, a "non-federated test in C++" in this context would imply a unit test, right?

korydraughn: Yes. You need to use the zone administration library to add a remote zone (w/o connection info) and then create a user in that zone using the user administration library. That will give you the tools to write a test for this PR.

korydraughn (commented):
Looking real good.

unit_tests/src/test_atomic_apply_acl_operations.cpp
```diff
@@ -177,6 +178,85 @@ TEST_CASE("atomic_apply_acl_operations")
...
+        const auto zone_info = ua::client::zone_info(conn, test_zone_name);
+        REQUIRE(zone_info.has_value());
+
+        irods::at_scope_exit remove_test_zone{[&conn, &test_zone_name] {
+            if (ua::client::zone_exists(conn, test_zone_name)) {
+                ua::client::remove_zone(conn, test_zone_name);
+            }
+        }};
+
+        ua::user user_a{"atomic_acl_zone_user", test_zone_name};
```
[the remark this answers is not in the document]
alanking: So you're suggesting making a user in the local zone with the same username?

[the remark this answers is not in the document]
korydraughn: Yep.

unit_tests/src/test_atomic_apply_acl_operations.cpp
```diff
@@ -177,6 +178,85 @@ TEST_CASE("atomic_apply_acl_operations")
...
+                {"operations", json::array({{{"entity_name", user_a.name}, {"zone", test_zone_name}, {"acl", "own"}}})}}
+                .dump();
+
+        char* json_error_string{};
+        irods::at_scope_exit free_memory{[&json_error_string] { std::free(json_error_string); }};
+
+        // should fail: remote zone with no zone param
+        REQUIRE_FALSE(rc_atomic_apply_acl_operations(
+                          static_cast<rcComm_t*>(conn), failing_json_input.c_str(), &json_error_string) == 0);
+        REQUIRE(json_error_string != "{}"s);
```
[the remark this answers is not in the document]
FifthPotato: Any best practices for when the message is extremely verbose?

unit_tests/src/test_atomic_apply_acl_operations.cpp
```diff
@@ -512,6 +434,205 @@ TEST_CASE("#7338")
...
+        {"logical_path", user_home},
+        {"operations", json::array({
+            {
+                {"entity_name", acl_user_name},
+                {"acl", "read"}
+            }
+        })}
+    }.dump();
+
+    // clang-format off
```
alanking, line 574:
```suggestion
    // clang-format on
```

unit_tests/src/test_atomic_apply_acl_operations.cpp
```diff
@@ -512,6 +434,205 @@ TEST_CASE("#7338")
...
+    REQUIRE(rc_atomic_apply_acl_operations(
+                static_cast<rcComm_t*>(conn), failing_json_input.c_str(), &json_error_string) == SYS_INVALID_INPUT_PARAM);
+    REQUIRE(json_error_string != "{}"s);
+    std::free(json_error_string);
+
+    // should succeed: remote zone only user with zone param
+    REQUIRE(rc_atomic_apply_acl_operations(
+                static_cast<rcComm_t*>(conn), remote_only_success_json_input.c_str(), &json_error_string) == 0);
+    REQUIRE(json_error_string == "{}"s);
+    std::free(json_error_string);
```
alanking, line 588: Perhaps it doesn't matter because this is a test, but... if the `REQUIRE` on the assertion on the above line fails, `json_error_string` will leak, right?

alanking (commented):
Unit test looks fine overall. Had a couple of observations

unit_tests/src/test_atomic_apply_acl_operations.cpp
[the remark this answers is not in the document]
korydraughn: Not really. What does the JSON output look like when the API call fails?

unit_tests/src/test_atomic_apply_acl_operations.cpp
FifthPotato: I think what I did just now should take care of that. Could you double-check my work?

unit_tests/src/test_atomic_apply_acl_operations.cpp
```diff
@@ -432,6 +433,210 @@ TEST_CASE("#7338")
...
+    // two users with identical names but differing zones
+    ua::user remote_zone_user{acl_user_name, test_zone_name};
+    ua::user local_zone_user{acl_user_name};
+    // one user with unique name in remote zone only
+    ua::user remote_only_user{remote_only_user_name, test_zone_name};
+
+    // clean up users after end
+    irods::at_scope_exit remove_users{[&] {
+        for (auto&& u : {remote_zone_user, local_zone_user, remote_only_user}) {
+            ua::client::remove_user(conn, u);
```
korydraughn, line 470: If `remove_user` can throw, consider wrapping it in `CHECK_NOTHROW`.

unit_tests/src/test_atomic_apply_acl_operations.cpp
alanking: I don't remember exactly what situation causes this, but pushing addresses onto `std::vector`s can lead to problems if the `std::vector` reallocates. I think because the `std::vector` is not taking ownership of the memory, this is okay, but it may be in our best interests to use `std::vector::reserve` just in case. I wish I could explain this better lol

unit_tests/src/test_atomic_apply_acl_operations.cpp
```diff
@@ -432,6 +433,210 @@ TEST_CASE("#7338")
...
+        }
+    }};
+
+    if (!ua::client::exists(conn, remote_zone_user)) {
```
korydraughn, line 474: Let's remove these checks.

If a requirement of the test isn't satisfied, we need the test to report that. The check for the existence of the user can lead to a prior test getting by without cleaning up.

unit_tests/src/test_atomic_apply_acl_operations.cpp
```diff
@@ -432,6 +433,210 @@ TEST_CASE("#7338")
...
+        ua::client::add_user(conn, remote_zone_user, ua::user_type::rodsuser, ua::zone_type::remote);
```
korydraughn, line 475: I'm pretty sure these can throw. If they do, wrap them in `REQUIRE_NOTHROW`.

unit_tests/src/test_atomic_apply_acl_operations.cpp
alanking: Alternatively, we could use a `std::array<char*, 4>` since we know the number of items. But like I said, I think what you have should be safe/okay.

SwooshyCueb:
Please include the branch name (e.g. main or 4-3-stable) in the PR title.
That's my nit, and I've picked it.

unit_tests/src/test_atomic_apply_acl_operations.cpp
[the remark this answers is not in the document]
FifthPotato: The message is an exception with full backtrace, so it's pretty hefty.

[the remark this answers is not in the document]
korydraughn: Please post the JSON output so we have something to comment on.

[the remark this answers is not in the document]
FifthPotato:
```
{"error_message":"iRODS Exception:\n    file: /irods_source/plugins/api/src/
  atomic_apply_acl_operations.cpp\n    function: void (anonymous namespace)::
  throw_if_invalid_entity_id((anonymous namespace)::id_type)\n    line: 204\n
  code: -130000 (SYS_INVALID_INPUT_PARAM)\n    message:\n        Invalid
  entity\nstack trace:\n--------------\n 0# irods::stacktrace::dump[abi:cxx11]
  () const in /lib/libirods_common.so.4.90.0\n 1# irods::exception::
  assemble_full_display_what() const in /lib/libirods_common.so.4.90.0\n 2#
  irods::exception::what() const in /lib/libirods_common.so.4.90.0\n 3#
  0x000073B59D5916B2 in /usr/lib/irods/plugins/api/
  libirods_atomic_apply_acl_operations_server.so\n 4# 0x000073B59D58F006 in /
  usr/lib/irods/plugins/api/libirods_atomic_apply_acl_operations_server.so\n 5#
  irods::experimental::catalog::execute_transaction(nanodbc::connection&amp;, std::
  function&lt;int (nanodbc::transaction&amp;)>) in /lib/libirods_server.so.4.90.0\n 6#
  0x000073B59D58D6A2 in /usr/lib/irods/plugins/api/
  libirods_atomic_apply_acl_operations_server.so\n 7# irods::api_call_adaptor
  &lt;BytesBuf*, BytesBuf**>::operator()(irods::plugin_context&amp;, RsComm*,
  BytesBuf*, BytesBuf**) in /usr/lib/irods/plugins/api/
  libirods_atomic_apply_acl_operations_server.so\n 8# std::_Function_handler
  &lt;irods::error (irods::plugin_context&amp;, RsComm*, BytesBuf*, BytesBuf**),
  irods::api_call_adaptor&lt;BytesBuf*, BytesBuf**> >::_M_invoke(std::_Any_data
  const&amp;, irods::plugin_context&amp;, RsComm*&amp;&amp;, BytesBuf*&amp;&amp;, BytesBuf**&amp;&amp;) in /
  usr/lib/irods/plugins/api/libirods_atomic_apply_acl_operations_server.so\n 9#
  int irods::api_entry::call_handler&lt;BytesBuf*, BytesBuf**>(RsComm*, BytesBuf*,
  BytesBuf**) in /usr/lib/irods/plugins/api/
  libirods_atomic_apply_acl_operations_server.so\n10# rsApiHandler(RsComm*,
  int, BytesBuf*, BytesBuf*) in /lib/libirods_server.so.4.90.0\n11#
  readAndProcClientMsg(RsComm*, int) in /lib/libirods_server.so.4.90.0\n12#
  agentMain(RsComm*) in /lib/libirods_server.so.4.90.0\n13#
  runIrodsAgentFactory(sockaddr_un) in /lib/libirods_server.so.4.90.0\n14#
  main::$_5::operator()() const at rodsServer.cpp:?\n15# main in /usr/sbin/
  irodsServer\n16# __libc_start_main in /lib/x86_64-linux-gnu/libc.so.6\n17#
  _start in /usr/sbin/irodsServer\n\n","operation":{"acl":"own","entity_name":
  "atomic_acl_remote_zone_user"},"operation_index":0}
  ```

File: plugins/api/src/atomic_apply_acl_operations.cpp (hunk 3 above)
Comment (alanking) [reply plugins/api/src/atomic_apply_acl_operations.cpp:354]: Seems we are still awaiting an issue for writing a test for this so that we can update the TODO comment with it. Does that sound about right?

File: unit_tests/src/test_atomic_apply_acl_operations.cpp (hunk 5 above)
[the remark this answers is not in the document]
Comment (alanking) [reply unit_tests/src/test_atomic_apply_acl_operations.cpp]: Yikes. Maybe we just want to assert that there is an "error_message" member and that it contains `SYS_INVALID_INPUT_PARAM` and/or "Invalid entity"? Don't need to match all of those contents...

File: plugins/api/src/atomic_apply_acl_operations.cpp (hunk 3 above)
Comment (korydraughn) [reply plugins/api/src/atomic_apply_acl_operations.cpp:354]: I believe the tests included in this PR cover the TODO.

I don't think we have an icommand which uses the atomic ACLs API?

@FifthPotato Thoughts?

File: unit_tests/src/test_atomic_apply_acl_operations.cpp (hunk 5 above)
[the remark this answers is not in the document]
Comment (korydraughn) [reply unit_tests/src/test_atomic_apply_acl_operations.cpp]: That stacktrace is just noise. We should change the API to use `.client_display_what()` instead of `.what()`.

I agree with @alanking on the what needs to be asserted. The error code is very important. The contents of the message, a nice to have so we can detect when things change.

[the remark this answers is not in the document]
Comment (FifthPotato) [reply unit_tests/src/test_atomic_apply_acl_operations.cpp]: How does my most recent approach look?

File: plugins/api/src/atomic_apply_acl_operations.cpp (hunk 3 above)
Comment (korydraughn) [reply plugins/api/src/atomic_apply_acl_operations.cpp:354]: After some discussion, please do the following:
- Remove TODO
- Open an issue stating we need a federation test for rx_atomic_apply_acl_operations

File: unit_tests/src/test_atomic_apply_acl_operations.cpp (hunk 5 above)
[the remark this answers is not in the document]
Comment (korydraughn) [reply unit_tests/src/test_atomic_apply_acl_operations.cpp]: After some discussion, please do the following:
- Change the JSON output structure to use `client_display_what()` instead of `what()`
- Update unit test to check the JSON output structure using function at the bottom of the unit test source file

File: unit_tests/src/test_atomic_apply_acl_operations.cpp [hunk 11]
```diff
@@ -575,8 +575,11 @@ TEST_CASE("Atomic acls respect entity zone")
     REQUIRE(rc_atomic_apply_acl_operations(static_cast<rcComm_t*>(conn),
                                            failing_json_input.c_str(),
                                            &json_error_string) == SYS_INVALID_INPUT_PARAM);
+    json parsed_error_string = json::parse(json_error_string);
     err_strs.push_back(json_error_string);
```
alanking, line 579: I *think* the `json_error_string` would leak here if `json::parse` fails (because it will `throw`: https://json.nlohmann.me/api/basic_json/parse/#exceptions). Maybe just swap the order of these two lines:
```suggestion
    err_strs.push_back(json_error_string);
    json parsed_error_string = json::parse(json_error_string);
```

I think this is safe because of the exception safety guarantee outlined in the `nlohmann::json::parse` documentation linked above:
> Strong guarantee: if an exception is thrown, there are no changes in the JSON value.

korydraughn:
Don't forget to resolve review comments as you handle them.

FifthPotato:
Should be everything, I think? Will run tests now.

korydraughn:
Good. There are four clang-tidy errors to resolve. Please take care of those.

FifthPotato:
> Good. There are four clang-tidy errors to resolve. Please take care of those.

Done. Unit and core tests pass as well.

unit_tests/src/test_atomic_apply_acl_operations.cpp
```diff
@@ -432,6 +433,209 @@ TEST_CASE("#7338")
...
 }
 
+TEST_CASE("Atomic acls respect entity zone")
+{
+    using namespace std::string_literals;
+
+    rodsEnv env;
+    _getRodsEnv(env);
+    irods::experimental::client_connection conn;
+    const auto user_home = fs::path{static_cast<char*>(env.rodsHome)}.string();
```
korydraughn, line 443: Based on the test, it looks like L443 can be replaced with the following:
```cpp
const auto* const user_home = env.rodsHome;
```
Or ...
```cpp
const std::string user_home = env.rodsHome;
```

Please verify and rerun this one unit test to make sure it still passes.

unit_tests/src/test_atomic_apply_acl_operations.cpp
```diff
@@ -432,6 +433,209 @@ TEST_CASE("#7338")
     CHECK(json_error_string == nullptr);
 }
 
+TEST_CASE("Atomic acls respect entity zone")
```
korydraughn, line 436: Please add the issue number to the beginning of the string.
```
#7408: Atomic acls respect entity zone
```

korydraughn (commented):
Looks good.

I noticed two small things. Once you've resolved those and seen the unit test pass, squash it, no #'s.

FifthPotato:
Squashed!

korydraughn (approved):
If the test passes and you're happy with it, pound it.

alanking (approved):

FifthPotato:
Octothorpe'd

Issue #7759, opened 2024-05-23
Sister projects with decoupled or semi-decoupled version numbers should include irods version number in package revision


SwooshyCueb:
Many of our plugins and clients are no longer chained to a specific version of iRODS, instead specifying a minimum version (or no version requirement at all).
Projects with coupled version numbers that no longer require an exact iRODS version match and instead specify a minimum iRODS version now use the minimum iRODS version as the coupled component of the version number. (I call this a semi-coupled version number.)

In order to differentiate between packages built against different versions of iRODS, we should put the iRODS version number in the package revision.

PR #7895, opened 2024-07-11
[#7759] Stamp iRODS version number on sister project packages (4-3-stable)

SwooshyCueb:
Addresses #7759


korydraughn (commented):
Changes look good/normal.

I take it this only adds the iRODS version to the package metadata, correct?

SwooshyCueb:
package metadata and package filename

trel:
So when someone specifies a particular version to install... would they have to type/specify this new information?

aka `apt install irods-server=xxxxxx`

SwooshyCueb:
> So when someone specifies a particular version to install... would they have to type/specify this new information?
> 
> aka `apt install irods-server=xxxxxx`

Only in cases where the package revision was already needed. If they weren't typing `~jammy` or `.el8` before, they won't have to worry about this.
Also, this doesn't affect the irods-server package.

trel:
Right, I realized irods-server was immune as soon as I hit enter.

Going farther afield...  is there a future where they don't have to specify the longer `~jammy` or `.el8` suffix, and can get back to just the dotted-number?

korydraughn:
@trel Any additional comments to add for this?

So in the future, users who want to install a specific version of a sister project will need to include the `+<irods_version>` suffix in the version string.

I'm wondering if that's worth documenting in docs.irods.org or whether the release notes are enough.

@SwooshyCueb Are there conventions for handling that?

SwooshyCueb:
> So in the future, users who want to install a specific version of a sister project will need to include the `+<irods_version>` suffix in the version string.

Only on Debian and Ubuntu.

> I'm wondering if that's worth documenting in docs.irods.org or whether the release notes are enough.

apt has always required the full package version, including package revision string when specifying a version of a package to install. We started adding proper package revision strings to our package versions a while back, so I don't think any new documentation is needed. We can say something about the change in our release notes, but, in my opinion, if someone wants to learn how their package manager works, they should consult their package manager's documentation, not ours.

korydraughn:
Yep, makes sense.

The release notes are the place to inform people of the change.

korydraughn (approved):
Let's get one more approval before adding the pound.

trel (approved):
looks complete

korydraughn:
Pound it.

SwooshyCueb:
\#'d

korydraughn:
Merging.

PR #7906, opened 2024-07-17
[#7907] Add trace level logging to low-level ODBC code (4-3-stable) 

korydraughn:
This was used to figure out where the post-install test was failing when using MariaDB ODBC drivers to communicate with MariaDB 11.4.

### DIFF
--- a/plugins/database/src/low_level_odbc.cpp
+++ b/plugins/database/src/low_level_odbc.cpp
@@ -31,12 +31,21 @@
 #include "irods/private/low_level_odbc.hpp"
 
 #include "irods/irods_log.hpp"
+#include "irods/irods_logger.hpp"
 #include "irods/irods_error.hpp"
 #include "irods/irods_stacktrace.hpp"
 #include "irods/irods_server_properties.hpp"
 
 #include <cctype>
 #include <string>
+
+namespace
+{
+    using log_db = irods::experimental::log::database;
+
+    // NOLINTNEXTLINE(cppcoreguidelines-macro-usage)
+#define IRODS_DB_TRACE(msg, ...) log_db::trace("{}@{}: " msg, __func__, __LINE__ __VA_OPT__(,) __VA_ARGS__);
+} // anonymous namespace
 
 int _cllFreeStatementColumns( icatSessionStruct *icss, int statementNumber );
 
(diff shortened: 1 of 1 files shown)


korydraughn:
Forgot to mention that the second commit presents alternative wording for one log message.

Which one do you feel is better?

trel:
i like the updated message

plugins/database/src/low_level_odbc.cpp
```diff
@@ -31,13 +31,22 @@
...
 
 #include <cctype>
 #include <string>
 
+namespace
+{
+    using log_db = irods::experimental::log::database;
+
+    // NOLINTNEXTLINE(cppcoreguidelines-macro-usage)
+#define IRODS_DB_TRACE(msg, ...) log_db::trace("{}@{}: " msg, __func__, __LINE__ __VA_OPT__(, ) __VA_ARGS__);
```
korydraughn, line 47: `(, )` -> `(,)`

korydraughn: Done

korydraughn:
I felt people would like that wording better. Will use that.

korydraughn:
Squashed.

plugins/database/src/low_level_odbc.cpp
```diff
@@ -31,13 +31,22 @@
...
 
 #include <cctype>
 #include <string>
 
+namespace
+{
+    using log_db = irods::experimental::log::database;
+
+    // NOLINTNEXTLINE(cppcoreguidelines-macro-usage)
+#define IRODS_DB_TRACE(msg, ...) log_db::trace("{}@{}: " msg, __func__, __LINE__ __VA_OPT__(,) __VA_ARGS__);
```
trel, line 47: and just asking... `#define` has to be on left edge (no indentation), even though it's within the namespace block?

alanking (commented):
Seems good. Awaiting issue number

plugins/database/src/low_level_odbc.cpp
korydraughn: No, it's not required. Making preprocessor logic follow its own indentation helps to make it more noticeable, which is good since it's handled at a different stage during compilation.

Preprocessor exists in its own world.

korydraughn:
Issue number assigned.

plugins/database/src/low_level_odbc.cpp
alanking: This is also the formatting enforced by our clang-format configuration (on purpose).

alanking (approved):
\# it if happy with it. I suppose we should wrap up the open conversation first.

plugins/database/src/low_level_odbc.cpp
korydraughn: Correct.

@trel Does that answer your question?

korydraughn:
Ignoring clang-format. It wants to add extra whitespace which is unnecessary.

plugins/database/src/low_level_odbc.cpp
trel: it does.   resolved!

korydraughn:
#'d

PR #7916, opened 2024-07-22
[?] Use error information in THROW expression on failed connection (4-3-stable)

korydraughn:
The connection pool library will be handled in a later release since the changes are not trivial and will likely result in a breaking change.

All existing unit tests pass. I'm currently thinking through how to write tests which exercise the changes.

These changes appear to be completely safe due to the rErrMsg_t structure.

This PR is in draft until I have tests, but feel free to review.


alanking (commented):
Seems quite reasonable so far

korydraughn:
Added tests. They aren't exhaustive, but prove the implementation is correct.

These are ready now.

korydraughn:
Will tweak the commit message to only cover the `irods::client_connection` class.

unit_tests/src/test_client_connection.cpp
```diff
@@ -300,3 +307,108 @@ TEST_CASE("#7192: all connections store unique session signatures", "client_conn
...
+    }
+
+    SECTION("invalid port")
+    {
+        try {
+            ix::client_connection{env.rodsHost, 8080, {env.rodsUserName, env.rodsZone}};
+        }
+        catch (const irods::exception& e) {
+            // Assertion fails with including the embedded errno value.
+            CHECK(e.code() == (USER_SOCK_CONNECT_ERR - ECONNREFUSED));
```
alanking, line 337: I suppose that this implicitly tests for `ECONNREFUSED` inherently. Pretty slick.

If we wanted to explicitly assert the parts of the error code (i.e. the iRODS error code and the `errno` value), we could use the functions in `irods/system_error.hpp` to extract each bit of information and make assertions. What you have now works, too. Because math.

alanking (commented):
Just had one suggestion but it's totally ignorable

unit_tests/src/test_client_connection.cpp
korydraughn: True. Doing the math was easier than pulling in the header lol.

Can change if you feel it's better to use `system_error.hpp`. Either is fine with me.

Just noticed the comment on L336 is weird. Will fix that.

unit_tests/src/test_client_connection.cpp
```diff
@@ -333,8 +332,9 @@ TEST_CASE("#7155: report underlying error info from server on failed connection"
             ix::client_connection{env.rodsHost, 8080, {env.rodsUserName, env.rodsZone}};
         }
         catch (const irods::exception& e) {
-            // Assertion fails with including the embedded errno value.
-            CHECK(e.code() == (USER_SOCK_CONNECT_ERR - ECONNREFUSED));
+            // Extract the iRODS error code, ignoring the embedded errno value.
```
alanking, line 335: Is there any value in asserting the errno value? Is it important?

korydraughn: I'm not sure. It's trivial to add, so will so. Someone may find it valuable in the future.

korydraughn: Done.

unit_tests/src/test_client_connection.cpp
korydraughn: Switched to using system_error.hpp.

korydraughn:
Running unit tests.

alanking (commented):
Seems good to me. If there's nothing left to do, let's go ahead and squash em

EDIT: Oh, after unit tests, of course.

korydraughn:
Tests pass.

korydraughn:
Squashed. Tweaked commit message and applied clang-format.

alanking (approved):
\# it

korydraughn:
#'d

Issue #7909, opened 2024-07-17
Add GenQuery2 tests for PREP GenQuery python module


korydraughn:
- [x] main
- [x] 4-3-stable
---
## Testing

This issue exists so we have an issue number to tie to the tests.

The PREP issue numbers are behind the irods/irods issue numbers.

Related:
- irods/irods_rule_engine_plugin_python#182
- irods/irods_rule_engine_plugin_python#206

PR #7917, opened 2024-07-22
[#7909,irods/irods_rule_engine_plugin_python#182] Add genquery2 order-by test for genquery python module (4-3-stable)

korydraughn:
Depends on irods/irods_rule_engine_plugin_python#206.

This PR covers the following GenQuery2 features for the PREP genquery.py module.
- order-by clause
- `close()` member function

The new tests pass.

### DIFF
--- a/scripts/irods/test/test_prep_genquery_iterator.py
+++ b/scripts/irods/test/test_prep_genquery_iterator.py
@@ -982,3 +982,90 @@ class Test_Genquery_Iterator(resource_suite.ResourceBase, unittest.TestCase):
             f"[iteration=1] row: ['{self.admin.session_collection}']",
             f"[iteration=2] row: ['{self.admin.session_collection}']"
         ])
+
+    @unittest.skipUnless(plugin_name == 'irods_rule_engine_plugin-python', 'Requires PREP.')
+    def test_genquery_iterator_supports_genquery2_order_by_clause__issue_7909(self):
+        iterations = 5
+
+        # Create some data objects and build a single string which captures the expected order
+        # of the rows returned by GenQuery2.
+        data_name_prefix = 'test_genquery_iterator_supports_genquery2_order_by_clause__issue_7909'
+        expected_output = ''
+        for i in reversed(range(iterations)):
+            self.admin.assert_icommand(['itouch', f'{data_name_prefix}.{i}'])
+            expected_output += f"['{data_name_prefix}.{i}']\n"
+
+        # Create a rule file which, when executed, lists the data objects in the admin's
+        # session collection in reverse order.
+        rule_file = f'{self.admin.local_session_dir}/test_genquery_iterator_supports_genquery2_order_by_clause__issue_7909.r'
+        with open(rule_file, 'w') as rf:
+            rf.write(dedent(f'''
+            def main(rule_args, callback, rei):
+                from genquery import Query, Parser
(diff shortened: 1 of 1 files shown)


scripts/irods/test/test_prep_genquery_iterator.py
```diff
@@ -982,3 +982,88 @@ def main(rule_args, callback, rei):
...
+                from genquery import Query, Parser
+                for r in Query(callback, 'DATA_NAME', "COLL_NAME = '{self.admin.session_collection}'", order_by='DATA_NAME desc', parser=Parser.GENQUERY2):
+                    callback.writeLine('stdout', f'{{r}}')
+            INPUT null
+            OUTPUT ruleExecOut
+            '''))
+
+        # Execute the rule.
+        rep_instance = IrodsConfig().default_rule_engine_plugin + '-instance'
+        self.admin.assert_icommand(['irule', '-r', rep_instance, '-F', rule_file], 'STDOUT', expected_output)
```
alanking, line 1013: Does this actually care about the order of the items in `expected_output`?

scripts/irods/test/test_prep_genquery_iterator.py
```diff
@@ -982,3 +982,88 @@ def main(rule_args, callback, rei):
...
+
+            # Show the AVU containing the error exists on the session collection.
+            self.admin.assert_icommand(['imeta', 'ls', '-C', self.admin.session_collection], 'STDOUT', [
+                f'attribute: {attr_name}\n',
+                '[iRods__Error__Code:-130000]',
+                '[SYS_INVALID_INPUT_PARAM]'
+            ])
+
+    @unittest.skipUnless(plugin_name == 'irods_rule_engine_plugin-python', 'Requires PREP.')
+    def test_genquery_iterator_allows_close_member_function_to_be_called_multiple_times__issue_7909(self):
```
alanking, line 1043: Should we also add a test for when `GENQUERY1` parser is used since `close()` can be used by any `Query` object?

alanking (commented):
Looks good overall

scripts/irods/test/test_prep_genquery_iterator.py
korydraughn: Yes. Do you feel you've spotted a bug?

scripts/irods/test/test_prep_genquery_iterator.py
korydraughn: Done.

korydraughn:
PREP GenQuery2 tests still pass.

scripts/irods/test/test_prep_genquery_iterator.py
alanking: I wasn't sure if `assert_icommand` was just checking for each string in the list in the stdout contents, or if it is actually checking to see that they appear in order. Maybe worth reversing/scrambling the order of `expected_output` and making sure that the assertion here fails?

korydraughn: The `expected_output` is a single string containing the list of data object names in reversed order. Will double check.

korydraughn: Correct. I've confirmed `expected_output` is working as expected. Printing the string and removing the call to `reversed()` causes the test to fail.

korydraughn:
Squashed.

korydraughn:
#'d

PR #7964, opened 2024-09-24
[7947] make checksum buffer size configurable (4-3-stable)

JustinKyleJames:
This is a change to allow the setting of the read buffer size for checksum in server_config.json.

- The test case simply puts a file in iRODS, reads the checksum, removed the data object, updates the buffer size in the configuration, restarts the server, puts the same file, reads the checksum, compares the two, then cleans up.  I don't know of any way to verify that the buffer size is actually updated at runtime although I have manually tested and verified this with extra logging.

-  I updated the configuration template to add `checksum_read_buffer_size_in_bytes` with the same default that was originally hardcoded.  I started up a server to verify that a new installation has this configuration.  I also verified that during the execution of the test the configuration was in fact updated.

### DIFF
--- a/packaging/server_config.json.template
+++ b/packaging/server_config.json.template
@@ -3,6 +3,7 @@
     "schema_version": "v4",
     "advanced_settings": {
         "agent_factory_watcher_sleep_time_in_seconds": 5,
+        "checksum_read_buffer_size_in_bytes": 1048576,
         "default_number_of_transfer_threads": 4,
         "default_temporary_password_lifetime_in_seconds": 120,
         "delay_rule_executors": [],
--- a/scripts/irods/test/test_ichksum.py
+++ b/scripts/irods/test/test_ichksum.py
@@ -10,6 +10,8 @@ import shutil
 import re
 import hashlib
 import base64
+import json
+import psutil
 
 from . import session
 from . import settings
@@ -466,3 +468,100 @@ C- {5}:
         # Show that ichksum detects the size inconsistency between the catalog and the object in storage.
         self.admin.assert_icommand(['ichksum', '-f', data_object], 'STDERR', ['-512000 UNIX_FILE_READ_ERR'])
 
+    def test_ichksum_server_honors_server_config_option_checksum_read_buffer_size_in_bytes__issue_7947(self):
+
+        object_name = 'test_ichksum_server_honors_server_config_option_checksum_read_buffer_size_in_bytes__issue_7947'
+        server_config_filename = paths.server_config_path()
+
+        # load server_config.json to inject a new rule base
+        with open(server_config_filename) as f:
+            svr_cfg = json.load(f)
+
+        try:
+            # create a 100 MiB file
+            file_size = 100*1024*1024
+            local_file = os.path.join(self.admin.local_session_dir, object_name)
+            lib.make_file(local_file, file_size)
+            file_checksum = lib.file_digest(local_file, 'sha256', encoding='base64')
+
+            with lib.file_backed_up(server_config_filename):
+
+                for buffer_size in [10*1024*1024, file_size - 1, file_size, file_size + 1,  2**31-1]:
+
+                    # make sure we have plenty of available virtual memory to run this iteration
+                    if psutil.virtual_memory().available < 2 * buffer_size:
+                        continue
+
+                    # update the checksum_read_buffer_size_in_bytes
+                    svr_cfg['advanced_settings']['checksum_read_buffer_size_in_bytes'] = buffer_size
+                    new_server_config = json.dumps(svr_cfg, sort_keys=True, indent=4, separators=(',', ': '))
+
+                    # repave the existing server_config.json
+                    with open(server_config_filename, 'w') as f:
+                        f.write(new_server_config)
+
+                    # put the file with new server configuration, run ichksum, and compare checksums
+                    self.admin.assert_icommand(['iput', local_file, object_name])
+                    self.admin.assert_icommand(['ichksum', '-f', object_name], 'STDOUT_SINGLELINE', file_checksum)
+                    self.assertEqual(lib.get_replica_checksum(self.admin, object_name, 0), f'sha2:{file_checksum}')
+
+                    # cleanup for this loop
+                    self.admin.assert_icommand(['irm', '-f', object_name])
+
+                # test a buffer size of 1 with a much smaller file
+                file_size = 1024
+                local_file = os.path.join(self.admin.local_session_dir, object_name)
+                lib.make_file(local_file, file_size)
+                file_checksum = lib.file_digest(local_file, 'sha256', encoding='base64')
+
+                buffer_size = 1
+                svr_cfg['advanced_settings']['checksum_read_buffer_size_in_bytes'] = buffer_size;
+                new_server_config = json.dumps(svr_cfg, sort_keys=True, indent=4, separators=(',', ': '))
+
+                # repave the existing server_config.json
+                with open(server_config_filename, 'w') as f:
+                    f.write(new_server_config)
+
+                # put the file with new server configuration, run ichksum, and compare checksums
+                self.admin.assert_icommand(['iput', local_file, object_name])
+                self.admin.assert_icommand(['ichksum', '-f', object_name], 'STDOUT_SINGLELINE', file_checksum)
+                self.assertEqual(lib.get_replica_checksum(self.admin, object_name, 0), f'sha2:{file_checksum}')
+
+        finally:
+            os.unlink(local_file)
+            self.admin.assert_icommand(['irm', '-f', object_name])
+
+    def test_ichksum_server_rejects_invalid_server_config_options_for_checksum_read_buffer_size_in_bytes__issue_7947(self):
+        object_name = 'test_ichksum_server_rejects_invalid_server_config_options_for_checksum_read_buffer_size_in_bytes__issue_7947'
+        server_config_filename = paths.server_config_path()
+
+        # load server_config.json to inject a new rule base
+        with open(server_config_filename) as f:
+            svr_cfg = json.load(f)
+
+        try:
+            # create a small file
+            file_size = 10
+            local_file = os.path.join(self.admin.local_session_dir, object_name)
+            lib.make_file(local_file, file_size)
+
+            # put the file that we will be trying to calculate the checksum on
+            self.admin.assert_icommand(['iput', local_file, object_name])
+
+            with lib.file_backed_up(server_config_filename):
+
+                for buffer_size in [-2**63-1, -1, 0, 2**63]:  # first and last are outside range of 64 bit signed integers
+
+                    svr_cfg['advanced_settings']['checksum_read_buffer_size_in_bytes'] = buffer_size;
+                    new_server_config = json.dumps(svr_cfg, sort_keys=True, indent=4, separators=(',', ': '))
+
+                    # repave the existing server_config.json
+                    with open(server_config_filename, 'w') as f:
+                        f.write(new_server_config)
+
+                    # run icksum on the file
+                    self.admin.assert_icommand_fail(['ichksum', '-f', object_name], 'STDOUT_SINGLELINE', 'CONFIGURATION_ERROR')
+
+        finally:
+            os.unlink(local_file)
+            self.admin.assert_icommand(['irm', '-f', object_name])
--- a/scripts/irods/test/test_iput.py
+++ b/scripts/irods/test/test_iput.py
@@ -3,6 +3,7 @@ import json
 import os
 import sys
 import tempfile
+import psutil
 
 if sys.version_info < (2, 7):
     import unittest2 as unittest
@@ -885,6 +886,100 @@ class test_iput_with_checksums(session.make_sessions_mixin(rodsadmins, rodsusers
     def test_large_put(self):
         self.object_name = 'test_large_put'
         self.run_tests(self.object_name, 5120000)
+
+    def test_server_honors_server_config_option_checksum_read_buffer_size_in_bytes__issue_7947(self):
+        self.object_name = 'test_iput_with_updated_checksum_read_buffer_size_in_bytes__issue_7947'
+        server_config_filename = paths.server_config_path()
+
+        # load server_config.json to inject a new rule base
+        with open(server_config_filename) as f:
+            svr_cfg = json.load(f)
+
+        try:
+            # create a 100 MiB file
+            file_size = 100*1024*1024
+            local_file = os.path.join(self.admin.local_session_dir, self.object_name)
+            lib.make_file(local_file, file_size)
+            file_checksum = lib.file_digest(local_file, 'sha256', encoding='base64')
+
+            with lib.file_backed_up(server_config_filename):
+
+                for buffer_size in [10*1024*1024, file_size - 1, file_size, file_size + 1,  2**31-1]:
+
+                    # make sure we have plenty of available virtual memory to run this iteration
+                    if psutil.virtual_memory().available < 2 * buffer_size:
+                        continue
+
+                    svr_cfg['advanced_settings']['checksum_read_buffer_size_in_bytes'] = buffer_size
+                    new_server_config = json.dumps(svr_cfg, sort_keys=True, indent=4, separators=(',', ': '))
+
+                    # repave the existing server_config.json
+                    with open(server_config_filename, 'w') as f:
+                        f.write(new_server_config)
+
+                    # put the file with new server configuration and compare checksum
+                    self.admin.assert_icommand(['iput', '-R', self.resource, '-k', local_file, self.object_name])
+                    self.assertEqual(self.get_checksum(self.object_name).rstrip(), f'sha2:{file_checksum}')
+
+                    # cleanup for this loop
+                    self.admin.assert_icommand(['irm', '-f', self.object_name])
+
+                # test a buffer size of 1 with a much smaller file
+                file_size = 1024
+                local_file = os.path.join(self.admin.local_session_dir, self.object_name)
+                lib.make_file(local_file, file_size)
+                file_checksum = lib.file_digest(local_file, 'sha256', encoding='base64')
+
+                buffer_size = 1
+                svr_cfg['advanced_settings']['checksum_read_buffer_size_in_bytes'] = buffer_size;
+                new_server_config = json.dumps(svr_cfg, sort_keys=True, indent=4, separators=(',', ': '))
+
+                # repave the existing server_config.json
+                with open(server_config_filename, 'w') as f:
+                    f.write(new_server_config)
+
+                # put the file with new server configuration and compare checksum
+                self.admin.assert_icommand(['iput', '-R', self.resource, '-k', local_file, self.object_name])
+                self.assertEqual(self.get_checksum(self.object_name).rstrip(), f'sha2:{file_checksum}')
+
+        finally:
+            # self.object_name deleted in teardown
+            os.unlink(local_file)
+
+    def test_server_rejects_invalid_server_config_options_for_checksum_read_buffer_size_in_bytes__issue_7947(self):
+        self.object_name = 'test_server_rejects_invalid_server_config_options_for_checksum_read_buffer_size_in_bytes__issue_7947'
+        server_config_filename = paths.server_config_path()
+
+        # load server_config.json to inject a new rule base
+        with open(server_config_filename) as f:
+            svr_cfg = json.load(f)
+
+        try:
+            # create a small file
+            file_size = 10
+            local_file = os.path.join(self.admin.local_session_dir, self.object_name)
+            lib.make_file(local_file, file_size)
+
+            with lib.file_backed_up(server_config_filename):
+
+                for buffer_size in [-2**63-1, -1, 0, 2**63]:  # first and last are outside range of 64 bit signed integers
+
+                    svr_cfg['advanced_settings']['checksum_read_buffer_size_in_bytes'] = buffer_size
+                    new_server_config = json.dumps(svr_cfg, sort_keys=True, indent=4, separators=(',', ': '))
+
+                    # repave the existing server_config.json
+                    with open(server_config_filename, 'w') as f:
+                        f.write(new_server_config)
+
+                    # put the file and verify we get an error due to invalid configuration
+                    self.admin.assert_icommand(['iput', '-R', self.resource, '-k', local_file, self.object_name], 'STDERR_SINGLELINE', 'CONFIGURATION_ERROR')
+
+                    # clean up for next iteration
+                    self.admin.assert_icommand(['irm', '-f', self.object_name])
+
+        finally:
+            # self.object_name deleted in teardown
+            os.unlink(local_file)
 
     def get_checksum(self, object_name):
         iquest_result,_,ec = self.admin.run_icommand(['iquest', '%s', '"select DATA_CHECKSUM where DATA_NAME = \'{}\'"'.format(object_name)])
--- a/scripts/irods/test/test_irepl.py
+++ b/scripts/irods/test/test_irepl.py
@@ -2,6 +2,7 @@ from __future__ import print_function
 import os
 import sys
 import json
+import psutil
 
 if sys.version_info < (2, 7):
     import unittest2 as unittest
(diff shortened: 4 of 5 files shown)


server/api/src/rsFileChksum.cpp
```diff
@@ -190,14 +189,21 @@ int fileChksum(rsComm_t* rsComm,
         irods::getHasher( irods::MD5_NAME, hasher );
     }
 
+    // get the buffer size from server configuration - default to 1 MiB
+    int read_buffer_size;
+    try {
+        read_buffer_size = irods::get_advanced_setting<int>("checksum_read_buffer_size_in_bytes");
+    }
+    catch (const irods::exception&) {
+        rodsLog(LOG_DEBUG, "%s :: checksum_read_buffer_size_in_bytes not set.  Using default of 1 MB.", __func__);
```
trel, line 198:
```suggestion
        rodsLog(LOG_DEBUG, "%s :: checksum_read_buffer_size_in_bytes not set.  Using default of 1 MiB.", __func__);
```

server/api/src/rsFileChksum.cpp
```diff
@@ -339,11 +345,21 @@ int file_checksum(RsComm* _comm,
...
     }};
 
-    char buffer[SVR_MD5_BUF_SZ];
+    // get the buffer size from server configuration - default to 1 MiB
+    int read_buffer_size;
+    try {
+        read_buffer_size = irods::get_advanced_setting<int>("checksum_read_buffer_size_in_bytes");
+    }
+    catch (const irods::exception&) {
+        rodsLog(LOG_DEBUG, "%s :: checksum_read_buffer_size_in_bytes not set.  Using default of 1 MB.", __func__);
```
trel, line 354:
```suggestion
        rodsLog(LOG_DEBUG, "%s :: checksum_read_buffer_size_in_bytes not set.  Using default of 1 MiB.", __func__);
```

trel (commented):
please remove mention of clang-format from the commit message itself.

trel:
please update the irods_docs for unattended_install to include this setting as well...

https://github.com/irods/irods_docs/blob/7f17699711bdfdf382c4cd01b6c3a09061285586/docs/getting_started/installation.md?plain=1#L278

and

https://github.com/irods/irods_docs/blob/7f17699711bdfdf382c4cd01b6c3a09061285586/docs/getting_started/installation.md?plain=1#L442

scripts/irods/test/test_iput.py
```diff
@@ -886,6 +887,50 @@ def test_large_put(self):
...
+            irodsctl.restart()
+
+            # put the file again with new server configuration and compare checksum
+            self.admin.assert_icommand(['iput', '-R', self.resource, '-k', local_file, self.object_name])
+            self.assertEqual(self.get_checksum(self.object_name), tmp_checksum)
+
+        # self.object_name deleted in teardown
+
+        # Bounce server to get back original settings
+        irodsctl.restart()
```
korydraughn, line 932: The restart instructions can be removed since every agent reloads server_config.json after a fork.

scripts/irods/test/test_iput.py
```diff
@@ -886,6 +887,50 @@ def test_large_put(self):
...
+        # update the checksum_read_buffer_size_in_bytes to 10 MiB
+        svr_cfg['advanced_settings']['checksum_read_buffer_size_in_bytes'] = 10*1024*1024
+
+        # dump to a string to repave the existing server_config.json
+        new_server_config = json.dumps(svr_cfg, sort_keys=True, indent=4, separators=(',', ': '))
+
+        # create a 100 MiB file
+        file_size = 100*1024*1024
+        local_file = os.path.join(self.admin.local_session_dir, self.object_name)
+        lib.make_file(local_file, file_size)
```
korydraughn, line 908: File needs to be removed when the test ends (regardless of success or failure).

scripts/irods/test/test_iput.py
```diff
@@ -886,6 +887,50 @@ def test_large_put(self):
         self.object_name = 'test_large_put'
         self.run_tests(self.object_name, 5120000)
 
+    def test_iput_with_updated_checksum_read_buffer_size_in_bytes__issue_7947(self):
```
korydraughn, line 890: Please rework the test's name to describe what it does. For example:
```py
def test_server_honors_server_config_option_checksum_read_buffer_size_in_bytes__issue_7947(self):
```

korydraughn, line 890: Hmm, `test_iput.py` feels like the wrong place to put this. Given this change applies to multiple commands, we should find a python script that communicates that better.

We should add a test for `ichksum` to prove that the option applies to other commands too.

server/api/src/rsFileChksum.cpp
korydraughn, line 198: Replace use of `rodsLog` with new logger.

You can add the following near the top of the file.
```cpp

```

And then replace `rodsLog` with `log_api::debug`.

server/api/src/rsFileChksum.cpp
korydraughn, line 354: Replace `rodsLog` with new logger.

korydraughn (commented):
Looks good so far.

The configuration section at docs.irods.org will need an update too.

- https://docs.irods.org/4.3.3/system_overview/configuration/
- https://github.com/irods/irods_docs/blob/7f17699711bdfdf382c4cd01b6c3a09061285586/docs/system_overview/configuration.md?plain=1#L13

server/api/src/rsFileChksum.cpp
JustinKyleJames: done

server/api/src/rsFileChksum.cpp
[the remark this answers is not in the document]
JustinKyleJames: done

server/api/src/rsFileChksum.cpp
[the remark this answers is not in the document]
JustinKyleJames: done

scripts/irods/test/test_iput.py
JustinKyleJames: done

scripts/irods/test/test_iput.py
```diff
@@ -886,6 +887,50 @@ def test_large_put(self):
...
+        # clean up this file
+        self.admin.assert_icommand(['irm', '-f', self.object_name])
+
+        with lib.file_backed_up(server_config_filename):
+            # repave the existing server_config.json
+            with open(server_config_filename, 'w') as f:
+                f.write(new_server_config)
+
+            # Bounce server to apply setting
+            irodsctl.restart()
```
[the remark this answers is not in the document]
JustinKyleJames: done

scripts/irods/test/test_iput.py
JustinKyleJames: Missed that.  Done.

scripts/irods/test/test_iput.py
JustinKyleJames: done

server/api/src/rsFileChksum.cpp
JustinKyleJames: done

server/api/src/rsFileChksum.cpp
JustinKyleJames: done

scripts/irods/test/test_iput.py
alanking: I would suggest `test_icommands_file_operations` but I hate that file, so maybe each iCommand gets its own version of this test...

alanking (commented):
Looks good. Just a couple things in the test which may or may not be valid

scripts/irods/test/test_iput.py
korydraughn: After some discussion, we decided the following:
- The test must show that invalid values (e.g. negative integers) are detected and result in an error being returned to the client

> I would suggest `test_icommands_file_operations` but I hate that file, so maybe each iCommand gets its own version of this test...

That sounds reasonable to me.

packaging/server_config.json.template
```diff
@@ -24,7 +24,8 @@
         "number_of_concurrent_delay_rule_executors": 4,
         "stacktrace_file_processor_sleep_time_in_seconds": 10,
         "transfer_buffer_size_for_parallel_transfer_in_megabytes": 4,
-        "transfer_chunk_size_for_parallel_transfer_in_megabytes": 40
+        "transfer_chunk_size_for_parallel_transfer_in_megabytes": 40,
+        "checksum_read_buffer_size_in_bytes": 1048576
```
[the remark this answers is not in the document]
JustinKyleJames: done

scripts/irods/test/test_iput.py
```diff
@@ -886,6 +887,50 @@ def test_large_put(self):
...
+            svr_cfg = json.load(f)
+
+        # update the checksum_read_buffer_size_in_bytes to 10 MiB
+        svr_cfg['advanced_settings']['checksum_read_buffer_size_in_bytes'] = 10*1024*1024
+
+        # dump to a string to repave the existing server_config.json
+        new_server_config = json.dumps(svr_cfg, sort_keys=True, indent=4, separators=(',', ': '))
+
+        # create a 100 MiB file
+        file_size = 100*1024*1024
```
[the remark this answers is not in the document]
JustinKyleJames: I have not been able to reproduce the error.  I think it requires a sufficiently fast network to an S3 implementation that also generates the "reduce your request rate" error.

I have seen that error in AWS but don't remember what scenario created it.  When I try with a very small buffer size to AWS it does a lot of read request but not at a rate that would trigger this error.  It just takes a long time to read the full file.  It would probably require some type of parallel request which we don't do in this instance.

I also tried with minio but I don't think it generates that type of error.

But to answer the question, in general I can put a 100 MiB file to S3 with checksum without creating the error.

server/api/src/rsFileChksum.cpp
```diff
@@ -11,14 +11,17 @@
...
 #include <fmt/format.h>
 
 #include <algorithm>
 #include <string>
 #include <string_view>
+#include <vector>
+#include <climits>
 
-#define SVR_MD5_BUF_SZ (1024*1024)
+using log_api = irods::experimental::log::server;
```
korydraughn, line 24: `::server` -> `::api`

server/api/src/rsFileChksum.cpp
```diff
@@ -339,11 +349,29 @@ int file_checksum(RsComm* _comm,
...
+    int read_buffer_size;
+    try {
+        read_buffer_size = irods::get_advanced_setting<int>("checksum_read_buffer_size_in_bytes");
+    }
+    catch (const irods::exception&) {
+        log_api::debug("{} :: checksum_read_buffer_size_in_bytes not set. Using default of 1 MiB.", __func__);
+        read_buffer_size = 1024 * 1024;
+    }
+
+    if (read_buffer_size < 0 || read_buffer_size > INT_MAX) {
```
korydraughn, line 362: If `read_buffer_size` is an int, it can never exceed `INT_MAX`.

Given that we have Boost, consider looking into the safe numerics library. See https://gcc.godbolt.org/z/aq8Gojv9z for an example.

korydraughn, line 362: Is there a reason this logic isn't in `fileChksum`?

scripts/irods/test/test_iput.py
```diff
@@ -14,6 +14,7 @@
 from .. import paths
 from .. import test
 from ..test.command import assert_command
+from ..controller import IrodsController
```
korydraughn, line 17: This can be removed now.

scripts/irods/test/test_iput.py
```diff
@@ -886,6 +887,47 @@ def test_large_put(self):
         self.object_name = 'test_large_put'
         self.run_tests(self.object_name, 5120000)
 
+    def test_server_honors_server_config_option_checksum_read_buffer_size_in_bytes__issue_7947(self):
+        self.object_name = 'test_iput_with_updated_checksum_read_buffer_size_in_bytes__issue_7947'
+        irodsctl = IrodsController()
```
korydraughn, line 892: This can be removed now.

server/api/src/rsFileChksum.cpp
alanking: Should the first condition be `read_buffer_size <= 0`?

JustinKyleJames: > Should the first condition be `read_buffer_size <= 0`?

yes

JustinKyleJames: > If `read_buffer_size` is an int, it can never exceed `INT_MAX`.
> 
> Given that we have Boost, consider looking into the safe numerics library. See https://gcc.godbolt.org/z/aq8Gojv9z for an example.

I was thinking that myself since it is an int but added it because we said to have that check in there.  That said, I put a really long sequence of digits into that field in the json and it triggered the condition in 362.  Not sure how unless it somehow was parsed as a negative integer.

scripts/irods/test/test_iput.py
JustinKyleJames: missed it

JustinKyleJames: done

server/api/src/rsFileChksum.cpp
JustinKyleJames: I just missed it / forgot about it.

It's done.

server/api/src/rsFileChksum.cpp
JustinKyleJames: done

scripts/irods/test/test_iput.py
JustinKyleJames: done

scripts/irods/test/test_iput.py
JustinKyleJames: I added a test case to check invalid configuration options.  I have not added these for ichksum so leaving this unresolved for now.

scripts/irods/test/test_iput.py
```diff
@@ -928,6 +926,39 @@ def test_server_honors_server_config_option_checksum_read_buffer_size_in_bytes__
...
+
+        try:
+            # create a small file 
+            file_size = 10
+            local_file = os.path.join(self.admin.local_session_dir, self.object_name)
+            lib.make_file(local_file, file_size)
+
+            with lib.file_backed_up(server_config_filename):
+
+                for file_size in [-1000, -1, 0]:
```
alanking, line 946: `file_size` -> `buffer_size`?

Should we add some overflow/underflow values? e.g. 2^64, -1*2^64

scripts/irods/test/test_iput.py
```diff
@@ -928,6 +926,39 @@ def test_server_honors_server_config_option_checksum_read_buffer_size_in_bytes__
             # self.object_name deleted in teardown
             os.unlink(local_file)
 
+    def test_server_rejects_invalid_server_config_options_for_checksum_read_buffer_size_in_bytes__issue_7947(self):
+        self.object_name = 'test_iput_with_updated_checksum_read_buffer_size_in_bytes__issue_7947'
```
alanking, line 930: This doesn't matter too much, but it seems the object name still reflects the old test name.

scripts/irods/test/test_iput.py
JustinKyleJames: I didn't think this would be possible because we are going through a Python layer first but I just tried "2**64" and "2**64" in Python and they gave me back a number.  I didn't know that that was possible.

JustinKyleJames: I used -2\*\*63-1 and 2\*\*63 which both are just outside the range of 64 bit signed integers.

scripts/irods/test/test_iput.py
JustinKyleJames: done

scripts/irods/test/test_iput.py
```diff
@@ -943,9 +943,9 @@ def test_server_rejects_invalid_server_config_options_for_checksum_read_buffer_s
 
             with lib.file_backed_up(server_config_filename):
 
-                for file_size in [-1000, -1, 0]:
+                for buffer_size in [-2**63-1, -1, 0, 2**63]:  # first and last are outside range of 64 bit signed integers
 
-                    svr_cfg['advanced_settings']['checksum_read_buffer_size_in_bytes'] = file_size; 
+                    svr_cfg['advanced_settings']['checksum_read_buffer_size_in_bytes'] = buffer_size;
```
alanking, line 948: Trailing whitespace here

scripts/irods/test/test_iput.py
```diff
@@ -886,6 +886,79 @@ def test_large_put(self):
...
+        self.object_name = 'test_server_rejects_invalid_server_config_options_for_checksum_read_buffer_size_in_bytes__issue_7947'
+        server_config_filename = paths.server_config_path()
+
+        # load server_config.json to inject a new rule base
+        with open(server_config_filename) as f:
+            svr_cfg = json.load(f)
+
+
+        try:
+            # create a small file
```
alanking, line 939: Trailing whitespace here

alanking (commented):
Seems good to me. Don't forget that the #+1/2/3/etc. in the commit messages are making real links to issues in GitHub. Not a big deal, just a reminder

scripts/irods/test/test_iput.py
JustinKyleJames: Added irepl thinking that that is what we want.  I will add ichksum.

JustinKyleJames: I added ichksum tests.  I also noticed that there was a library function to calculate the checksum in the format desired for the file itself so I modified iput/irepl tests to calculate the checksum on the file rather to writing the file to iRODS without server_config.json updates and then reading it.  I now just do the checksum comparison to what I calculated from the posix file.

scripts/irods/test/test_iput.py
JustinKyleJames: Also semicolon.  Both removed.

JustinKyleJames:
> Seems good to me. Don't forget that the #+1/2/3/etc. in the commit messages are making real links to issues in GitHub. Not a big deal, just a reminder

Oops

scripts/irods/test/test_ichksum.py
```diff
@@ -466,3 +467,81 @@ def test_ichksum_detects_when_the_size_in_storage_is_less_than_the_size_in_catal
...
 
+    def test_ichksum_server_honors_server_config_option_checksum_read_buffer_size_in_bytes__issue_7947(self):
+
+        def get_checksum(data_object, replica_number, expected_length_of_checksum=49):
+            gql = "select DATA_CHECKSUM where COLL_NAME = '{0}' and DATA_NAME = '{1}' and DATA_REPL_NUM = '{2}'"
+            checksum, err, ec = self.admin.run_icommand(['iquest', '%s', gql.format(self.admin.session_collection, data_object, replica_number)])
+            self.assertEqual(ec, 0)
+            self.assertEqual(len(err), 0)
+            self.assertEqual(len(checksum.strip()), expected_length_of_checksum)
+            return checksum.strip()
```
alanking, line 478: FWIW, we have a function called `get_replica_checksum` in lib: https://github.com/irods/irods/blob/bc2a9bf0d8d3587d1aad8bf6a9b3db650e0d6944/scripts/irods/lib.py#L651-L654 Unfortunately, it only takes a `DATA_NAME` for some reason but that shouldn't be a problem for your particular data object. If we use that, we won't need so many copies of the same function.

server/api/src/rsFileChksum.cpp
alanking, line 1: clang-tidy is expressing some opinions in this file which we may want to consider addressing.

scripts/irods/test/test_iput.py
```diff
@@ -886,6 +886,72 @@ def test_large_put(self):
...
+
+                    svr_cfg['advanced_settings']['checksum_read_buffer_size_in_bytes'] = buffer_size
+                    new_server_config = json.dumps(svr_cfg, sort_keys=True, indent=4, separators=(',', ': '))
+
+                    # repave the existing server_config.json
+                    with open(server_config_filename, 'w') as f:
+                        f.write(new_server_config)
+
+                    # put the file and verify we get an error due to invalid configuration
+                    self.admin.assert_icommand(['iput', '-f', '-R', self.resource, '-k', local_file, self.object_name], 'STDERR_SINGLELINE', 'CONFIGURATION_ERROR')
```
alanking, line 949: The `-f` makes me itchy here. Maybe we should just `irm` the data object each iteration?

scripts/irods/test/test_irepl.py
```diff
@@ -212,6 +212,104 @@ def test_irepl_a_updates_stale_zero_byte_replicas__issue_6285(self):
...
+
+        ufs_resources = ['issue_7947_ufs1', 'issue_7947_ufs2']
+
+        try:
+            # create two resources
+            for resc_name in ufs_resources:
+                lib.create_ufs_resource(self.admin, resc_name)
+
+            # create a 100 MiB file
+            file_size = 100*1024*1024
```
alanking, line 246: Thinking on this some more, we should probably include tests for the following:

- Smallest valid size (1 byte)
- Largest valid size (2^31-1(?))
- buffer size == size of file
- buffer size == size of file - 1
- buffer size == size of file + 1

We can punt on this by creating a new issue for it. Up to you.

scripts/irods/test/test_ichksum.py
```diff
@@ -466,3 +467,81 @@ def test_ichksum_detects_when_the_size_in_storage_is_less_than_the_size_in_catal
...
+            file_checksum = lib.file_digest(local_file, 'sha256', encoding='base64')
+
+            with lib.file_backed_up(server_config_filename):
+                # repave the existing server_config.json
+                with open(server_config_filename, 'w') as f:
+                    f.write(new_server_config)
+
+                # put the file with new server configuration, run ichksum, and compare checksums
+                self.admin.assert_icommand(['iput', local_file, object_name])
+                self.admin.assert_icommand(['ichksum', '-f', object_name], 'STDOUT_SINGLELINE', file_checksum)
```
alanking, line 507: Is the `-f` needed here in order to register the checksum?

alanking (commented):
Sorry for extra comments - noticed some other stuff. It's mostly in the testing though

scripts/irods/test/test_irepl.py
JustinKyleJames: Aren't we asking for problems if we try a buffer size of 2 GiB?

alanking: Mmmm probably. We don't have to do that one. Or maybe we can skip the test based on the amount of memory available on the machine (might be running in a k8s pod!).

scripts/irods/test/test_ichksum.py
JustinKyleJames: I made this change for both test_ichksum.py and test_irepl.py.  I left test_iput.py alone since it already had that method as a class variable.

scripts/irods/test/test_iput.py
JustinKyleJames: The problem is that that class deletes self.object_name in its tearDown() method so if I delete every iteration that file will not exist to clean up.

scripts/irods/test/test_ichksum.py
JustinKyleJames: I believe so.

scripts/irods/test/test_iput.py
alanking: That's okay because `irm -f` doesn't produce an error in the event that the data object doesn't exist. Alternatively, you could wrap the try-finally with the loop, but I think it's better how it is now.

Speaking of which, each iteration of the loop could be made into a `unittest.subTest` if you're feeling up to it.

scripts/irods/test/test_irepl.py
JustinKyleJames: I added it in with a check on each loop to make sure we have at least twice the memory and put the 1 byte buffer by itself with a much smaller file.

scripts/irods/test/test_iput.py
JustinKyleJames: I removed the -f and deleted the object on each iteration

server/api/src/rsFileChksum.cpp
JustinKyleJames: Maybe we can discuss.  I'm not sure I agree with them.

1. The narrowing from unsigned long to signed long is unavoidable I believe.  We are never going to be close to this being a problem.
2. The uninitialized buffer size seems picky as it is set immediately following.  Either it is read from configuration or set to the default - all immediately after declaration.  Setting it initially to zero seems worse than waiting to set it as we would never want that to be zero.\
3. Not sure about the complexity complaint.

korydraughn: Sure thing. Adding to discush list.

JustinKyleJames: I made the clang-tidy changes.

server/api/src/rsFileChksum.cpp
```diff
@@ -23,6 +23,21 @@
...
+    constexpr int default_read_buffer_size = 1024 * 1024;
+    int read_buffer_size = default_read_buffer_size;
+    try {
+        read_buffer_size = irods::get_advanced_setting<int>("checksum_read_buffer_size_in_bytes");
+    }
+    catch (const irods::exception&) {
+        log_api::debug("{} :: checksum_read_buffer_size_in_bytes not set. Using default of 1 MiB.", __func__);
+        read_buffer_size = default_read_buffer_size;
+    }
+    return read_buffer_size;
```
korydraughn, line 38: This can be rewritten like so:
```cpp
try {
    return irods::get_advanced_setting<int>("checksum_read_buffer_size_in_bytes");
}
catch (...) {
    log_api::debug("{} :: checksum_read_buffer_size_in_bytes not set. Using default of 1 MiB.", __func__);
    constexpr int default_read_buffer_size = 1024 * 1024;
    return default_read_buffer_size;
}
```

server/api/src/rsFileChksum.cpp
```diff
@@ -11,14 +11,32 @@
...
 #include <algorithm>
 #include <string>
 #include <string_view>
+#include <vector>
+#include <climits>
 
-#define SVR_MD5_BUF_SZ (1024*1024)
+using log_api = irods::experimental::log::api;
+
+static int get_read_buffer_size()
```
alanking, line 26: Consider using an anonymous namespace. But it's fine as-is. Up to you.

server/api/src/rsFileChksum.cpp
```diff
@@ -11,14 +11,32 @@
...
 #include "irods/irods_at_scope_exit.hpp"
+#include "irods/irods_logger.hpp"
 
 #include <fmt/format.h>
 
 #include <algorithm>
 #include <string>
 #include <string_view>
+#include <vector>
+#include <climits>
```
alanking, line 22: Sort, please

alanking (commented):
Just a couple final thoughts, and clang-format had one opinion as well. Otherwise, seems like we're pretty much there.

korydraughn (commented):
Have one suggestion on the checksum buffer size function.

Once addressed, I think we may be ready to squash _(assuming the tests are passing)_.

server/api/src/rsFileChksum.cpp
JustinKyleJames: I made the change.

server/api/src/rsFileChksum.cpp

server/api/src/rsFileChksum.cpp
JustinKyleJames: I put it into an anonymous namespace.

JustinKyleJames:
> Have one suggestion on the checksum buffer size function.
> 
> Once addressed, I think we may be ready to squash _(assuming the tests are passing)_.

Commits have been squashed.

alanking (approved):
I think this is ready. Await one more approval and then # it.

korydraughn (approved):
Pound it!

JustinKyleJames:
> Pound it!

done

PR #7966, opened 2024-09-26
[7913] Add support for same range of permissions ichmod allows in atomic ACLs (main)

FifthPotato:
Haven't run full test suite, but did run the individual unit test to make sure it works. 
Question: do we need to throw in `to_access_type_id` if `throw_if_invalid_acl` is called at the top of `execute_acl_operation`? 
Question 2: I noticed some of the permissions were called `X_object` (e.g. `write_object`) but the `_object` was omitted in previous iterations of this plugin. Do we standardize all of them to `X_object`? I didn't do that in this PR, but worth a thought.


trel:
The `_object` notation is the correct / modern one.   The 'read' and 'write' aliases were just short-forms.

korydraughn (commented):
> Question: do we need to throw in `to_access_type_id` if `throw_if_invalid_acl` is called at the top of `execute_acl_operation?`

If `throw_if_invalid_acl` already protects the server from invalid client values, then it's fine to remove the check/throw.

plugins/api/src/atomic_apply_acl_operations.cpp
```diff
@@ -207,14 +216,12 @@ namespace
...
-        if (_acl == "null")  { return 1000; }
-        // clang-format on
-
-        THROW(SYS_INVALID_INPUT_PARAM, fmt::format("Invalid ACL [acl={}]", _acl));
+        if (acl_to_id.find(_acl) == acl_to_id.cend()) {
+            // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+            THROW(SYS_INVALID_INPUT_PARAM, fmt::format("Invalid ACL [acl={}]", _acl));
+        }
+
+        return acl_to_id.at(_acl);
```
alanking, line 224: Minor improvement: We could store the iterator returned by `find` and use that instead of calling `at` in the `return` line to save an extra search.

Otherwise, we could use `throw_if_invalid_acl` here since L219-L222 pretty much duplicates that function.

alanking (commented):
Looks good

plugins/api/src/atomic_apply_acl_operations.cpp
alanking: Oh, I just noticed your question. Please ignore suggestion as appropriate

alanking:
Also, feel free to remove the centos:7 builder GitHub Action... Or, we can just ignore the failure for now.

korydraughn:
@FifthPotato We still need to accept `read` and `write` as ACLs for backwards compatibility.

alanking (commented):
Seems good to me

unit_tests/src/test_atomic_apply_acl_operations.cpp
```diff
@@ -609,9 +609,9 @@ TEST_CASE("#7408: Atomic acls respect entity zone")
     b = std::begin(perms);
     e = std::end(perms);
 
-    // success indicator: should find unique permission "write" with remote zone and shared username
+    // success indicator: should find unique permission "delete_metadata" with remote zone and shared username
     REQUIRE(std::find_if(b, e, [&acl_user_name, &test_zone_name](const fs::entity_permission& p) {
-                return p.name == acl_user_name && p.zone == test_zone_name && p.prms == fs::perms::write;
+                return p.name == acl_user_name && p.zone == test_zone_name && p.prms == fs::perms::delete_metadata;
```
korydraughn, line 614: Let's avoid changing existing tests if they aren't broken.

Adding a new test will allow you to produce something that focuses on the issue.

trel: agreed - consider even making a test that walks through them all(?).

unit_tests/src/test_atomic_apply_acl_operations.cpp
```diff
@@ -636,6 +636,109 @@ TEST_CASE("#7408: Atomic acls respect entity zone")
...
+    // clang-format off
+
+    const std::array new_acls_and_perms{ std::make_pair("read_metadata", fs::perms::read_metadata), 
+                                         std::make_pair("read_object", fs::perms::read),
+                                         std::make_pair("create_metadata", fs::perms::create_metadata),
+                                         std::make_pair("modify_metadata", fs::perms::modify_metadata),
+                                         std::make_pair("delete_metadata", fs::perms::delete_metadata),
+                                         std::make_pair("create_object", fs::perms::create_object),
+                                         std::make_pair("modify_object", fs::perms::write),
+                                         std::make_pair("delete_object", fs::perms::delete_object) };
```
alanking, line 669: Is the reason that we are only testing the "new" permissions because the others (null, read, write, and own) are being tested elsewhere? Should we include them here anyway, if only to avoid future confusion?

unit_tests/src/test_atomic_apply_acl_operations.cpp
```diff
@@ -636,6 +636,109 @@ TEST_CASE("#7408: Atomic acls respect entity zone")
...
+
+            REQUIRE(rc_atomic_apply_acl_operations(
+                        static_cast<rcComm_t*>(conn), json_input.c_str(), &json_error_string) == 0);
+            REQUIRE(json_error_string == "{}"s);
+
+            const auto perms = fs::client::status(conn, user_home).permissions();
+            auto b = std::begin(perms);
+            auto e = std::end(perms);
+            REQUIRE(std::find_if(
+                        b, e, [&acl_user](const fs::entity_permission& p) { return p.name == acl_user.name; }) == e);
```
alanking, line 720: If we made a test collection to manipulate rather than changing permissions on the local client environment user's home collection, it would become less imperative that this succeeds (since it would only affect test data). Thoughts?

unit_tests/src/test_atomic_apply_acl_operations.cpp
```diff
@@ -636,6 +636,109 @@ TEST_CASE("#7408: Atomic acls respect entity zone")
...
+        }};
+
+        char* json_error_string{};
+
+        REQUIRE(rc_atomic_apply_acl_operations(static_cast<rcComm_t*>(conn), json_input.c_str(), &json_error_string) ==
+                0);
+        REQUIRE(json_error_string == "{}"s);
+        err_strs.push_back(json_error_string);
```
alanking, line 728: Do we need to put this before the `REQUIRE(json_error_string == "{}"s);`? In the event that the assertion fails, the string will leak.

unit_tests/src/test_atomic_apply_acl_operations.cpp
```diff
@@ -636,6 +636,109 @@ TEST_CASE("#7408: Atomic acls respect entity zone")
...
+    _getRodsEnv(env);
+    irods::experimental::client_connection conn;
+    const auto* const user_home = env.rodsHome;
+    const auto* const acl_user_name = "atomic_acl_user";
+
+    ua::user acl_user{acl_user_name};
+
+    // clean up users after end
+    irods::at_scope_exit remove_users{[&] {
+            CHECK_NOTHROW(ua::client::remove_user(conn, acl_user));
```
alanking, line 653:
```suggestion
        CHECK_NOTHROW(ua::client::remove_user(conn, acl_user));
```

unit_tests/src/test_atomic_apply_acl_operations.cpp
```diff
@@ -636,6 +636,109 @@ TEST_CASE("#7408: Atomic acls respect entity zone")
...
+            // clang-format off
+        const auto json_input = json{
+            {"logical_path", user_home},
+            {"operations", json::array({
+                {
+                    {"entity_name", acl_user.name},
+                    {"acl", "null"}
+                },
+            })}
+        }.dump();
```
alanking, line 704:
```suggestion
            const auto json_input = json{
                {"logical_path", user_home},
                {"operations", json::array({
                    {
                        {"entity_name", acl_user.name},
                        {"acl", "null"}
                    },
                })}
            }.dump();
```

unit_tests/src/test_atomic_apply_acl_operations.cpp
```diff
@@ -636,6 +636,109 @@ TEST_CASE("#7408: Atomic acls respect entity zone")
             }) != e);
 }
 
+TEST_CASE("#7966: Atomic acls accept all acls")
```
alanking, line 639:
```suggestion
TEST_CASE("#7913: Atomic acls accept all acls")
```

unit_tests/src/test_atomic_apply_acl_operations.cpp
alanking: Actually, will the string be populated in the event that `rc_atomic_apply_acl_operations` returns non-zero? We might need to use an `at_scope_exit` in the loop to ensure the string is freed.

unit_tests/src/test_atomic_apply_acl_operations.cpp
trel: i think i like that idea.

unit_tests/src/test_atomic_apply_acl_operations.cpp
trel: yes please, just loop all of them.   could even 'replace' the older test if that helps clarify.

unit_tests/src/test_atomic_apply_acl_operations.cpp
trel: new test added... resolving.

alanking:
Oh, and clang-format is angry

unit_tests/src/test_atomic_apply_acl_operations.cpp
```diff
@@ -636,6 +636,109 @@ TEST_CASE("#7408: Atomic acls respect entity zone")
...
+    // clean up users after end
+    irods::at_scope_exit remove_users{[&] {
+            CHECK_NOTHROW(ua::client::remove_user(conn, acl_user));
+    }};
+
+    REQUIRE_NOTHROW(ua::client::add_user(conn, acl_user));
+
+    REQUIRE(ua::client::exists(conn, acl_user));
+
+    // clang-format off
```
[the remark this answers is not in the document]
FifthPotato: I'll play around with it a bit to see where to put this better...

unit_tests/src/test_atomic_apply_acl_operations.cpp
FifthPotato: I could initialize it to 0, add the `char*` to the vector beforehand and then check that it's nonzero before freeing in the outer `at_scope_exit`. Do you think that would work?

alanking: Yes, I think that works.

unit_tests/src/test_atomic_apply_acl_operations.cpp
```diff
@@ -636,6 +636,112 @@ TEST_CASE("#7408: Atomic acls respect entity zone")
...
+
+            REQUIRE(rc_atomic_apply_acl_operations(
+                        static_cast<rcComm_t*>(conn), json_input.c_str(), &json_error_string) == 0);
+            REQUIRE(json_error_string == "{}"s);
+
+            const auto perms = fs::client::status(conn, user_home).permissions();
+            auto b = std::begin(perms);
+            auto e = std::end(perms);
+            REQUIRE(std::find_if(
+                        b, e, [&acl_user](const fs::entity_permission& p) { return p.name == acl_user.name; }) == e);
```
alanking, line 723: I think this could be `std::none_of`, similar to the other spot where we replaced the `std::find_if` with `std::any_of`. Up to you.

unit_tests/src/test_atomic_apply_acl_operations.cpp
```diff
@@ -657,7 +657,11 @@ TEST_CASE("#7913: Atomic acls accept all acls")
 
     REQUIRE(ua::client::exists(conn, acl_user));
 
-    // clang-format off
+    // create test collection to modify ACLs on
+    const auto test_collection = fs::path{user_home} / "test_collection";
```
trel, line 661: please put 7913 in the test collection name somewhere

korydraughn:
Clang-Tidy report needs some attention.

Feel free to handle #7968 in this PR. Up to you.

korydraughn (commented):
Very nice.

Please check the doxygen for the API and MSI. The doxygen for the MSI will be located at the bottom of the cpp file.

Given you've only expanded the range of perms, I think it's enough to just run the C++ unit tests suite for the atomic ACLs. If all the tests pass, squash it.

alanking (commented):
Look at all that green...

unit_tests/src/test_atomic_apply_acl_operations.cpp
```diff
@@ -636,6 +636,110 @@ TEST_CASE("#7408: Atomic acls respect entity zone")
...
+
+    // create test collection to modify ACLs on
+    const auto test_collection = fs::path{user_home} / "acl_test_collection_issue_7913";
+    REQUIRE_NOTHROW(fs::client::create_collection(conn, test_collection));
+    irods::at_scope_exit remove_test_collection{
+        [&] { REQUIRE(fs::client::remove_all(conn, test_collection, fs::remove_options::no_trash)); }};
+
+    const std::array new_acls_and_perms{std::make_pair("read_metadata", fs::perms::read_metadata),
+                                        std::make_pair("read", fs::perms::read),
+                                        std::make_pair("read_object", fs::perms::read),
```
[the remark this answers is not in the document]
FifthPotato: This one is interesting-- since the enums are actually distinct in the `perms` enum, when you `stat` for the permissions, it will actually give you `fs::perms::read` specifically, not `read_object`. If mapped to `read_object`, this test will fail. This also applies to the `modify_object` case!

unit_tests/src/test_atomic_apply_acl_operations.cpp
```diff
@@ -636,6 +636,110 @@ TEST_CASE("#7408: Atomic acls respect entity zone")
...
+                                        std::make_pair("create_metadata", fs::perms::create_metadata),
+                                        std::make_pair("modify_metadata", fs::perms::modify_metadata),
+                                        std::make_pair("delete_metadata", fs::perms::delete_metadata),
+                                        std::make_pair("create_object", fs::perms::create_object),
+                                        std::make_pair("write", fs::perms::write),
+                                        std::make_pair("modify_object", fs::perms::write),
```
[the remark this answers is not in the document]
FifthPotato: As above.

unit_tests/src/test_atomic_apply_acl_operations.cpp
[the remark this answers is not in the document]
korydraughn: Got it.

FifthPotato:
> Very nice.
> 
> Please check the doxygen for the API and MSI. The doxygen for the MSI will be located at the bottom of the cpp file.
> 
> Given you've only expanded the range of perms, I think it's enough to just run the C++ unit tests suite for the atomic ACLs. If all the tests pass, squash it.

I couldn't find a doxygen thing for the API that mentions any perms specifically, only for the MSI. Is that fine?

korydraughn:
See the following:
- https://github.com/irods/irods/blob/bc2a9bf0d8d3587d1aad8bf6a9b3db650e0d6944/lib/api/include/irods/atomic_apply_acl_operations.h#L42-L45
- https://github.com/irods/irods/blob/bc2a9bf0d8d3587d1aad8bf6a9b3db650e0d6944/server/api/include/irods/rs_atomic_apply_acl_operations.hpp#L42-L45

FifthPotato:
Unit tests for atomic ACLs pass. Will squash soon.

FifthPotato:
Squished.

alanking (approved):
Looks good to me. I say # it.

FifthPotato:
Octothorpe'd

alanking:

PR #7976, opened 2024-10-10
[7967] Change if statement so it only triggers on first pass (main)

FifthPotato:
This change makes the ticket failure only trigger when the uses count is updated. Ran just the test for `iticket`, but not sure if this breaks anything else (haven't run full test suite yet).

### DIFF
--- a/plugins/database/src/mid_level_routines.cpp
+++ b/plugins/database/src/mid_level_routines.cpp
@@ -15,6 +15,7 @@
 #include "irods/private/low_level.hpp"
 #include "irods/irods_stacktrace.hpp"
 #include "irods/irods_log.hpp"
+#include "irods/irods_logger.hpp"
 #include "irods/irods_virtual_path.hpp"
 #include "irods/rodsErrorTable.h"
 #include "irods/rcMisc.h"
@@ -27,6 +28,8 @@
 
 /* Size of the R_OBJT_AUDIT comment field;must match table column definition */
 #define AUDIT_COMMENT_MAX_SIZE       1000
+
+using log_db = irods::experimental::log::database;
 
 extern int logSQL_CML;
 
@@ -1600,10 +1603,6 @@ int checkObjIdByTicket(const char* dataId,
 
     iUsesLimit = atoi( usesLimit );
     if ( iUsesLimit > 0 ) {
-        iUsesCount = atoi( usesCount );
-        if ( iUsesCount > iUsesLimit ) {
-            return CAT_TICKET_USES_EXCEEDED;
-        }
 
         intDataId = atoll( dataId );
 
@@ -1611,6 +1610,20 @@ int checkObjIdByTicket(const char* dataId,
         // Given this function can be called multiple times within the same operation, checking to
         // see if the data id is different keeps the server from updating the ticket information multiple times.
         if ( previousDataId2 != intDataId ) {
+            errno = 0;
+            //NOLINTNEXTLINE(cppcoreguidelines-avoid-magic-numbers,readability-magic-numbers)
+            iUsesCount = static_cast<int>(strtol(usesCount, nullptr, 10));
+            if (errno != 0) {
+                log_db::error("[{}]: Error converting usesCount to int. [usesCount: [{}], dataId: [{}], errno: [{}]]",
+                              __func__,
+                              usesCount,
+                              dataId,
+                              errno);
+                return SYS_INTERNAL_ERR;
+            }
+            if (iUsesCount >= iUsesLimit) {
+                return CAT_TICKET_USES_EXCEEDED;
+            }
             iUsesCount++;
             snprintf( myUsesCount, sizeof myUsesCount, "%d", iUsesCount );
             cllBindVars[cllBindVarCount++] = myUsesCount;
@@ -1625,7 +1638,6 @@ int checkObjIdByTicket(const char* dataId,
             if ( status != 0 ) {
                 return status;
             }
-
 #ifndef ORA_ICAT
             cllCheckPending( "", 2, icss->databaseType );
 #endif
--- a/scripts/irods/test/test_iticket.py
+++ b/scripts/irods/test/test_iticket.py
@@ -1122,6 +1122,25 @@ class Test_Iticket(SessionsMixin, unittest.TestCase):
         self.assertGreater(modify_ts, create_ts)
         self.assertGreater(modify_ts, previous_ts)
 
+    def test_exceeding_ticket_use_count_returns_correct_error_code__issue_7967(self):
+        ticket_string = "issue_7967_uses_count"
+        data_object = "issue_7967_uses_filename"
+        self.user.assert_icommand(['itouch', data_object])
+
+        self.user.assert_icommand(['iticket', 'create', 'read', data_object, ticket_string])
+
+        # Modify ticket usage amount
+        self.user.assert_icommand(['iticket', 'mod', ticket_string, 'uses', '3'])
+        self.user.assert_icommand(['iticket', 'ls', ticket_string], 'STDOUT', ['uses count: 0', 'uses limit: 3'])
+        self.user.assert_icommand(['iget', '-f', '-t', ticket_string, data_object])
+        self.user.assert_icommand(['iticket', 'ls', ticket_string], 'STDOUT', ['uses count: 1'])
+        self.user.assert_icommand(['iget', '-f', '-t', ticket_string, data_object])
+        self.user.assert_icommand(['iticket', 'ls', ticket_string], 'STDOUT', ['uses count: 2'])
+        self.user.assert_icommand(['iget', '-f', '-t', ticket_string, data_object])
+        self.user.assert_icommand(['iticket', 'ls', ticket_string], 'STDOUT', ['uses count: 3'])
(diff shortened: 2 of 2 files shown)


trel:
yes, the change to >= may have broken the write scenario... please make sure we have tests that bracket both the read and write scenarios.

scripts/irods/test/test_iticket.py
```diff
@@ -1122,6 +1122,21 @@ def test_ticket_mtime_is_updated_when_ticket_properties_are_modified_by_non_admi
         self.assertGreater(modify_ts, create_ts)
         self.assertGreater(modify_ts, previous_ts)
 
+    def test_exceeding_ticket_uses_returns_correct_error_code__issue_7967(self):
```
korydraughn, line 1125: `uses` -> `use_count` ?

scripts/irods/test/test_iticket.py
```diff
@@ -1122,6 +1122,21 @@ def test_ticket_mtime_is_updated_when_ticket_properties_are_modified_by_non_admi
...
+        self.user.assert_icommand(['itouch', file_name])
+
+        self.user.assert_icommand(['iticket', 'create', 'read', file_name, ticket_string])
+
+        # Modify ticket usage amount
+        self.user.assert_icommand(['iticket', 'mod', ticket_string, 'uses', '3'])
+        self.user.assert_icommand(['iget', '-f', '-t', ticket_string, file_name])
+        self.user.assert_icommand(['iget', '-f', '-t', ticket_string, file_name])
+        self.user.assert_icommand(['iget', '-f', '-t', ticket_string, file_name])
+        self.user.assert_icommand(['iget', '-f', '-t', ticket_string, file_name],'STDERR', ['CAT_TICKET_USES_EXCEEDED'])
```
korydraughn, line 1137: Let's assert the use-count is still 3 after this line.

scripts/irods/test/test_iticket.py
```diff
@@ -1122,6 +1122,21 @@ def test_ticket_mtime_is_updated_when_ticket_properties_are_modified_by_non_admi
...
+
+
```
korydraughn, line 1139: Remove empty line.

scripts/irods/test/test_iticket.py
korydraughn, line 1137: Add other parts of the expected error message to future proof the test from things like `BLAH_CAT_TICKET_USES_EXCEEDED_BLAH`. Including the error code with its symbolic name is a good way to achieve that.

scripts/irods/test/test_iticket.py
```diff
@@ -1122,7 +1122,7 @@ def test_ticket_mtime_is_updated_when_ticket_properties_are_modified_by_non_admi
         self.assertGreater(modify_ts, create_ts)
         self.assertGreater(modify_ts, previous_ts)
 
-    def test_exceeding_ticket_uses_returns_correct_error_code__issue_7967(self):
+    def test_exceeding_ticket_use_count_returns_correct_error_code__issue_7967(self):
         ticket_string = "issue_7967_uses_count"
         file_name = "issue_7967_uses_filename"
```
korydraughn, line 1127: Consider changing `file_name` to `data_object`.

korydraughn (commented):
Looks good.

Had one recommendation.

Has this PR been run against the unit and core test suites? If not, please do that once you feel it's ready.

Is there anything else left to do for this PR?

FifthPotato:
Core and unit tests pass.

korydraughn:
Good. Squash it. No pounds.

FifthPotato:
Squished.

alanking (commented):
Please make some mention of tickets in the commit title so it's clearer with which feature the change is related. Otherwise, I think this looks good.

FifthPotato:
> Please make some mention of tickets in the commit title so it's clearer with which feature the change is related. Otherwise, I think this looks good.

Done!

korydraughn:
There's one clang-tidy issue that needs to be addressed.

plugins/database/src/mid_level_routines.cpp
```diff
@@ -1600,17 +1600,17 @@ int checkObjIdByTicket(const char* dataId,
...
-            return CAT_TICKET_USES_EXCEEDED;
-        }
 
         intDataId = atoll( dataId );
 
         // Don't update a second time if this id matches the last one.
         // Given this function can be called multiple times within the same operation, checking to
         // see if the data id is different keeps the server from updating the ticket information multiple times.
         if ( previousDataId2 != intDataId ) {
+            iUsesCount = static_cast<int>(strtol(usesCount, nullptr, 0));
```
korydraughn, line 1610: Need to set errno to 0 before calling strtol().
Need to check errno too.

See https://en.cppreference.com/w/c/string/byte/strtol

FifthPotato: What should I do in the case `errno` is nonzero after the call?
1. Log a message (nothing else in `mid_level_routines.cpp` logs without the switch being enabled, so this is maybe weird?)
2. Return `CAT_TICKET_USES_EXCEEDED`
3. All of the above

FifthPotato: ...or 4, something else?

korydraughn: Log a message and return an error.

We shouldn't return `CAT_TICKET_USES_EXCEEDED` unless that's what happened. Perhaps one of the following could work:
- SYS_INTERNAL_ERR
- SYS_LIBRARY_ERROR
- Does rodsErrorTable.h define any conversion/parsing error?

If we don't have an error code that describes the situation well, we can talk about introducing a new one.

trel:
don't pre-squash.   want to see the diffs...

FifthPotato:
> can you add a couple more `iticket ls` assertions just before and after the threshold is hit?

Done!

korydraughn:
Please run this through the test suite again to see if the use of `strtol` reveals something unexpected.

Let us know how it goes.

trel:
new assertions look good.  thanks.

scripts/irods/test/test_iticket.py
```diff
@@ -1122,6 +1122,25 @@ def test_ticket_mtime_is_updated_when_ticket_properties_are_modified_by_non_admi
...
+        self.user.assert_icommand(['iticket', 'mod', ticket_string, 'uses', '3'])
+        self.user.assert_icommand(['iticket', 'ls', ticket_string], 'STDOUT', ['uses count: 0', 'uses limit: 3'])
+        self.user.assert_icommand(['iget', '-f', '-t', ticket_string, data_object])
+        self.user.assert_icommand(['iticket', 'ls', ticket_string], 'STDOUT', ['uses count: 1'])
+        self.user.assert_icommand(['iget', '-f', '-t', ticket_string, data_object])
+        self.user.assert_icommand(['iticket', 'ls', ticket_string], 'STDOUT', ['uses count: 2'])
+        self.user.assert_icommand(['iget', '-f', '-t', ticket_string, data_object])
+        self.user.assert_icommand(['iticket', 'ls', ticket_string], 'STDOUT', ['uses count: 3'])
+        self.user.assert_icommand(['iget', '-f', '-t', ticket_string, data_object], 'STDERR', ['-892000 CAT_TICKET_USES_EXCEEDED'])
+        self.user.assert_icommand(['iticket', 'ls', ticket_string], 'STDOUT', ['uses count: 3'])
```
trel, line 1142: does this mean it no longer goes to 4?     good if true.

FifthPotato: Sure does!

trel: excellent.

alanking (commented):
The can of worms has opened, apparently. Sorry. lol Hoping my questions are dumb and/or invalid.

plugins/database/src/mid_level_routines.cpp
```diff
@@ -1600,17 +1603,26 @@ int checkObjIdByTicket(const char* dataId,
...
-        }
 
         intDataId = atoll( dataId );
 
         // Don't update a second time if this id matches the last one.
         // Given this function can be called multiple times within the same operation, checking to
         // see if the data id is different keeps the server from updating the ticket information multiple times.
         if ( previousDataId2 != intDataId ) {
+            errno = 0;
+            iUsesCount = static_cast<int>(strtol(usesCount, nullptr, 0));
```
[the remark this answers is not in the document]
korydraughn: The easiest thing to do is replace use of strtol with stol and rely on exceptions. I bet those cover all the bases.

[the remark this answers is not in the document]
alanking: I was just thinking about the case where you feed it garbage (e.g. "junk") and it fails to convert it to a value, effectively meaning that the uses count is 0.

I guess the choices for that case would be... 

a. the ticket is always usable
b. the ticket is never usable

This would fall under (a), which seems okay, I suppose. This only really matters if you can actually set the uses count, though. So maybe none of this matters (or at least is not particularly concerning).

FifthPotato:
Latest commit passes unit tests. Commit before the base 10 change passes core tests.

alanking (commented):
Seems good to me. I say squash em.

korydraughn:
Agreed. Squash it.

FifthPotato:
Squished.

alanking (approved):
\# it

korydraughn (approved):

FifthPotato:
Octothorpe'd

PR #7981, opened 2024-10-21
[7967] Change if statement so ticket use count is only updated on first pass (4-3-stable)

FifthPotato:
Cherry-pick of #7976


alanking (approved):
Looks identical. # it.

Issue #7410, opened 2023-11-01
Don't put libraries into `/usr/lib` if that's not where they go


SwooshyCueb:
Related: #4232 

Currently, we install libraries and plugins into `/usr/lib` regardless of distribution. We should instead follow the standard practice for the distribution. For example, Debian/Ubuntu expects libraries and plugins to be in `/usr/lib/<triplet>` and EL expects `/usr/lib<qual>`.

korydraughn:
Can you provide an example for `<triplet>` and `<qual>`?

SwooshyCueb:
Some examples for Ubuntu/Debian:
`/usr/lib/x86_64-linux-gnu`
`/usr/lib/aarch64-linux-gnu`
`/usr/lib/s390x-linux-gnu`
`/usr/lib/riscv64-linux-gnu`
`/usr/lib/powerpc64le-linux-gnu`

Some examples for EL:
`/usr/lib64`
`/usr/lib32`
`/usr/libx32`

PR #7982, opened 2024-10-22
[#7410] Use default value for `CMAKE_INSTALL_LIBDIR` (main)

SwooshyCueb:
With this PR, `InstallDirs.cmake` no longer overrides the default value for `CMAKE_INSTALL_LIBDIR`, which means the by default, our libraries and plugins will now be installed in the proper locations depending on distribution directory structure.

Since `CMAKE_INSTALL_LIBDIR` is a cache variable on which other variables depend, a clean cmake build directory is recommended.

So far testing hasn't revealed anything surprising, but I've got more tests to run, so I'm leaving this in draft mode for now.

### DIFF
--- a/cmake/InstallDirs.cmake
+++ b/cmake/InstallDirs.cmake
@@ -14,13 +14,6 @@ if(CMAKE_HOST_UNIX)
   endif()
 endif()
 
-# We don't do multiarch triples on debian, nor do we use the lib64 dir on centos and
-# opensuse (though we probably should). We can revisit this later; for now let's just
-# set our own default for CMAKE_INSTALL_LIBDIR
-if(NOT DEFINED CMAKE_INSTALL_LIBDIR)
-  set(CMAKE_INSTALL_LIBDIR "lib" CACHE PATH "Object code libraries (lib)")
-endif()
-
 include(GNUInstallDirs)
 
 if(NOT DEFINED IRODS_INCLUDE_DIRS)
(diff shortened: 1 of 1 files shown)


SwooshyCueb:
Seems to behave as intended. Ready for review.

alanking (approved):
Seems good to me. Please await one more approval and I think this should be good to go

cmake/InstallDirs.cmake
```diff
@@ -1,3 +1,19 @@
+# We used to override CMAKE_INSTALL_LIBDIR to always default to "lib", but starting with 5.0, we now
+# use the distribution-defaults for library installation directories. Since CMAKE_INSTALL_LIBDIR is
+# a cache variable, let's try and detect if an old default is in the cache.
+function(warn_for_old_libdir_cache)
+  if(CMAKE_INSTALL_LIBDIR STREQUAL "lib"
+     AND EXISTS "${CMAKE_BINARY_DIR}/CMakeCache.txt"
+     AND NOT IRODS_INSTALL_LIBDIR_GOOD_DEFAULT)
```
korydraughn, line 7: So `IRODS_INSTALL_LIBDIR_GOOD_DEFAULT` is a way to tell the build to skip this check?

SwooshyCueb: Correct.

korydraughn (approved):
Pound it.

SwooshyCueb:
\#'d

Issue #7440, opened 2023-12-19
adding duplicate AVU results in ODBC Function Sequence Error


trel:
### iRODS Version, OS and Version

4.3.1

### What did you try to do?

add the same AVU to the same data object, twice

### Expected behavior

i think... an error saying this already exists on this data object and would be a duplicate

### Observed behavior (including steps to reproduce, if applicable)

```
irods$ itouch foo
irods$ imeta add -d foo a b c
irods$ imeta add -d foo a b c
remote addresses: 127.0.0.1 ERROR: rcModAVUMetadata failed with error -806000 CAT_SQL_ERR 
```
and in the logs...

```
{                                                                                     
  "log_category": "legacy",                                                           
  "log_level": "info",                                                                
  "log_message": "bindVar[1]=10819",                                                  
  "request_api_name": "MOD_AVU_METADATA_AN",
  "request_api_number": 706,    
  "request_api_version": "d",   
  "request_client_user": "rods",         
  "request_host": "127.0.0.1",            
  "request_proxy_user": "rods",           
  "request_release_version": "rods4.3.1",                                             
  "server_host": "xxx.xxx.xxx.xxx",                                            
  "server_pid": 1481377,                                                              
  "server_timestamp": "2023-12-19T15:15:29.727Z",
  "server_type": "agent",  
  "server_zone": "tempZone"
}                          
{                                                                                     
  "log_category": "legacy",                                                                                                                                                 
  "log_level": "info",                                                                                                                                                      
  "log_message": "bindVar[2]=10623",                                                  
  "request_api_name": "MOD_AVU_METADATA_AN",
  "request_api_number": 706,    
  "request_api_version": "d",   
  "request_client_user": "rods",         
  "request_host": "127.0.0.1",            
  "request_proxy_user": "rods",           
  "request_release_version": "rods4.3.1",                                             
  "server_host": "xxx.xxx.xxx.xxx",                                            
  "server_pid": 1481377,                                                              
  "server_timestamp": "2023-12-19T15:15:29.727Z",
  "server_type": "agent",                  
  "server_zone": "tempZone"                
}                                          
{                                          
  "log_category": "legacy",                                                           
  "log_level": "info",                                                                
  "log_message": "bindVar[3]=01702998929",                                            
  "request_api_name": "MOD_AVU_METADATA_AN",                                          
  "request_api_number": 706,               
  "request_api_version": "d",              
  "request_client_user": "rods",           
  "request_host": "127.0.0.1",             
[...]

PR #7987, opened 2024-10-30
[7599] Rework logic in _cllExecSqlNoResult to return proper error codes (main)

FifthPotato:
In the "moving existing code around" stage. 
Works for all the test cases listed in the issue. 
Still needs new test(s) to box in the behavior, as well as running of all the other tests.

### DIFF
--- a/plugins/database/src/low_level_odbc.cpp
+++ b/plugins/database/src/low_level_odbc.cpp
@@ -86,6 +86,8 @@ SQLINTEGER columnLength[MAX_TOKEN];  /* change me ! */
 #ifndef ORA_ICAT
 static int didBegin = 0;
 #endif
+// Stores the affected row count of queries made through cllExecSqlNoResult
+// Read by cllGetRowCount
 static int noResultRowCount = 0;
 
 // =-=-=-=-=-=-=-
@@ -463,6 +465,7 @@ _cllExecSqlNoResult(
     HDBC myHdbc = icss->connectPtr;
     HSTMT myHstmt;
     SQLRETURN stat = SQLAllocHandle( SQL_HANDLE_STMT, myHdbc, &myHstmt );
+    SQL_INT_OR_LEN rowCount = 0;
     if ( stat != SQL_SUCCESS ) {
         rodsLog( LOG_ERROR, "_cllExecSqlNoResult: SQLAllocHandle failed for statement: %d", stat );
         return -1;
@@ -475,9 +478,6 @@ _cllExecSqlNoResult(
     rodsLogSql( sql );
 
     stat = SQLExecDirect( myHstmt, ( unsigned char * )sql, strlen( sql ) );
-    SQL_INT_OR_LEN rowCount = 0;
-    auto ec = SQLRowCount(myHstmt, (SQL_INT_OR_LEN*) &rowCount); // NOLINT(google-readability-casting)
-    IRODS_DB_TRACE("SQLRowCount returned [{}] and set [rowCount] to [{}].", ec, rowCount);
     switch ( stat ) {
     case SQL_SUCCESS:
         rodsLogSqlResult( "SUCCESS" );
@@ -502,6 +502,9 @@ _cllExecSqlNoResult(
     if ( stat == SQL_SUCCESS ||
             stat == SQL_SUCCESS_WITH_INFO ||
             stat == SQL_NO_DATA_FOUND ) {
+        const auto ec = SQLRowCount(myHstmt, (SQL_INT_OR_LEN*) &rowCount); // NOLINT(google-readability-casting)
+        IRODS_DB_TRACE("SQLRowCount returned [{}] and set [rowCount] to [{}].", ec, rowCount);
+
         cllCheckPending( sql, 0, icss->databaseType );
         result = 0;
         if ( stat == SQL_NO_DATA_FOUND ) {
@@ -534,8 +537,17 @@ _cllExecSqlNoResult(
         }
         rodsLog( LOG_NOTICE, "_cllExecSqlNoResult: SQLExecDirect error: %d sql:%s",
                  stat, sql );
+
+        // logPsgError returns an iRODS-specific error code for duplicate entries in the catalog
+        // Returns -2 otherwise.
         result = logPsgError( LOG_NOTICE, icss->environPtr, myHdbc, myHstmt,
                               icss->databaseType );
+        // On error cases, do not call SQLRowCount: it will cause a Function sequence error
+        // However, calling functions may read the return value of SQLRowCount via
+        // static variable noResultRowCount, which is set below.
+        // Thus, since SQLRowCount returns -1 on error cases, we emulate the result
+        // without calling SQLRowCount directly.
+        rowCount = -1;
     }
 
     stat = SQLFreeHandle( SQL_HANDLE_STMT, myHstmt );
@@ -543,6 +555,7 @@ _cllExecSqlNoResult(
         rodsLog( LOG_ERROR, "_cllExecSqlNoResult: SQLFreeHandle for statement error: %d", stat );
     }
 
+    // Saves the affected row count to a static variable
     noResultRowCount = rowCount;
 
     return result;
--- a/scripts/irods/test/test_iadmin.py
+++ b/scripts/irods/test/test_iadmin.py
@@ -1669,6 +1669,22 @@ class Test_Iadmin(resource_suite.ResourceBase, unittest.TestCase):
     def test_non_admins_are_not_allowed_to_invoke_iadmin_get_delay_server_info__issue_5249(self):
         self.user0.assert_icommand(['iadmin', 'get_delay_server_info'], 'STDERR', ['-830000 CAT_INSUFFICIENT_PRIVILEGE_LEVEL'])
 
+    def test_iadmin_mkuser_returns_correct_error_on_duplicate_user__issue_7599(self):
+        test_user_name = 'duplicate_user_issue_7599'
+        try:
+            self.admin.assert_icommand(['iadmin', 'mkuser', test_user_name, 'rodsuser'])
+            self.admin.assert_icommand(['iadmin', 'mkuser', test_user_name, 'rodsuser'], 'STDERR', ['-809000 CATALOG_ALREADY_HAS_ITEM_BY_THAT_NAME'])
+        finally:
+            self.admin.run_icommand(['iadmin', 'rmuser', test_user_name])
+
+    def test_iadmin_mkgroup_returns_correct_error_on_duplicate_group__issue_7599(self):
+        test_group_name = 'duplicate_group_issue_7599'
+        try:
+            self.admin.assert_icommand(['iadmin', 'mkgroup', test_group_name])
+            self.admin.assert_icommand(['iadmin', 'mkgroup', test_group_name], 'STDERR', ['-809000 CATALOG_ALREADY_HAS_ITEM_BY_THAT_NAME'])
+        finally:
+            self.admin.run_icommand(['iadmin', 'rmgroup', test_group_name])
+
 class Test_Iadmin_Resources(resource_suite.ResourceBase, unittest.TestCase):
 
     def setUp(self):
--- a/scripts/irods/test/test_imeta_error_handling.py
+++ b/scripts/irods/test/test_imeta_error_handling.py
@@ -71,10 +71,7 @@ class Test_Imeta_Error_Handling(unittest.TestCase):
         self.admin.assert_icommand(['imeta', 'addw', '-d', self.test_data_paths_wildcard, attr, val], 'STDOUT', desired_rc=0)
         _, err, ec = self.admin.run_icommand(['imeta', 'addw', '-d', self.test_data_paths_wildcard, attr, val])
         self.assertEqual(4, ec)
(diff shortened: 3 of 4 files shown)


alanking:
Is this PR ready for review? If not, please put it in draft mode. Thanks

FifthPotato:
> Is this PR ready for review? If not, please put it in draft mode. Thanks

Good point! I'll do so. What I currently have does work and is OK to look at, but maybe in the interest of everyone's time, it's easier for me to leave it in draft until the tests are in so there's the option to review everything in one sweep.

plugins/database/src/low_level_odbc.cpp
```diff
@@ -502,6 +500,9 @@ _cllExecSqlNoResult(
     if ( stat == SQL_SUCCESS ||
             stat == SQL_SUCCESS_WITH_INFO ||
             stat == SQL_NO_DATA_FOUND ) {
+        auto ec = SQLRowCount(myHstmt, (SQL_INT_OR_LEN*) &rowCount); // NOLINT(google-readability-casting)
```
korydraughn, line 503: Is `ec` ever checked after this call?

scripts/irods/test/test_iadmin.py
```diff
@@ -1669,6 +1669,22 @@ def test_non_admins_are_not_allowed_to_invoke_iadmin_set_delay_server__issue_524
     def test_non_admins_are_not_allowed_to_invoke_iadmin_get_delay_server_info__issue_5249(self):
         self.user0.assert_icommand(['iadmin', 'get_delay_server_info'], 'STDERR', ['-830000 CAT_INSUFFICIENT_PRIVILEGE_LEVEL'])
 
+    def test_iadmin_mkuser_returns_correct_error_on_duplicate_user__issue_7599(self):
+        test_user_name = 'duplicate_user_issue_7599'
+        try:
+            self.admin.assert_icommand(['iadmin', 'mkuser', test_user_name, 'rodsuser'])
+            self.admin.assert_icommand(['iadmin', 'mkuser', test_user_name, 'rodsuser'], 'STDERR', ['-809000 CATALOG_ALREADY_HAS_ITEM_BY_THAT_NAME'])
+        finally:
+            self.admin.assert_icommand(['iadmin', 'rmuser', test_user_name])
```
korydraughn, line 1678: Let's use `.run_icommand` instead of `assert_icommand` for clean-up code.

This allows more tests to run and possibly pass, reducing the number of tests we have to rerun and check.

scripts/irods/test/test_iadmin.py
```diff
@@ -1669,6 +1669,22 @@ def test_non_admins_are_not_allowed_to_invoke_iadmin_set_delay_server__issue_524
...
+
+    def test_iadmin_mkgroup_returns_correct_error_on_duplicate_group__issue_7599(self):
+        test_group_name = 'duplicate_group_issue_7599'
+        try:
+            self.admin.assert_icommand(['iadmin', 'mkgroup', test_group_name])
+            self.admin.assert_icommand(['iadmin', 'mkgroup', test_group_name], 'STDERR', ['-809000 CATALOG_ALREADY_HAS_ITEM_BY_THAT_NAME'])
+        finally:
+            self.admin.assert_icommand(['iadmin', 'rmgroup', test_group_name])
```
korydraughn, line 1686: Let's use `.run_icommand` instead of `assert_icommand` for clean-up code.

This allows more tests to run and possibly pass, reducing the number of tests we have to rerun and check.

scripts/irods/test/test_imeta_error_handling.py
```diff
@@ -74,7 +74,7 @@ def test_imeta_addw_return_code(self):
         # In unixODBC versions > 2.3.5, the database plugin does not return CATALOG_ALREADY_HAS_ITEM_BY_THAT_NAME
         # but instead it returns CAT_SQL_ERR. We support platforms that use both 2.3.4 and 2.3.6, so either error
         # code is acceptable for this test.
-        self.assertTrue('CATALOG_ALREADY_HAS_ITEM_BY_THAT_NAME' in err or 'CAT_SQL_ERR' in err)
+        self.assertTrue('CATALOG_ALREADY_HAS_ITEM_BY_THAT_NAME' in err)
```
korydraughn, line 77: Prefer the more targeted versions of the assert functions over `assertTrue` and `assertFalse` as they lead to clearer failure messages when they occur.

`assertTrue` -> `assertIn`

scripts/irods/test/test_imeta_error_handling.py
```diff
@@ -86,7 +86,7 @@ def test_imeta_add_return_code(self):
...
```
korydraughn, line 89: Prefer the more targeted versions of the assert functions over `assertTrue` and `assertFalse` as they lead to clearer failure messages when they occur.

`assertTrue` -> `assertIn`

korydraughn (commented):
Very nice.

Has this been run against the unit/core tests?

plugins/database/src/low_level_odbc.cpp
```diff
@@ -536,6 +537,7 @@ _cllExecSqlNoResult(
                  stat, sql );
         result = logPsgError( LOG_NOTICE, icss->environPtr, myHdbc, myHstmt,
                               icss->databaseType );
+        rowCount = -1;
```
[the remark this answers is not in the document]
trel: So we're already in an error `else` block... and we're setting `rowCount = -1` because a calling function expects to see `-1` when there is an error?

But... where is `rowCount` used after this line other than to set `noResultRowCount`... which is ... not used?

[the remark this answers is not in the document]
trel: and what is `result` in this case?   it gets returned below without checking it?

plugins/database/src/low_level_odbc.cpp
FifthPotato: Nope, it's only used in the logging statement in the next line (same as the original code-- I made no deletions here). If that's no longer necessary, it can be removed.

plugins/database/src/low_level_odbc.cpp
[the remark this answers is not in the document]
FifthPotato: Tried to better-ify the comment. Did it work?

plugins/database/src/low_level_odbc.cpp
korydraughn: Got it. Please mark it `const` since it's not modified.

plugins/database/src/low_level_odbc.cpp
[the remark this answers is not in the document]
korydraughn: That's better.

Add a separate commit which does the following:
- Adds a comment to the static global variable explaining its purpose and how it relates to this function and others
- Adds comments throughout this function explaining what it does and why

This is all about helping the next person who enters this function. You probably understand this function better than anyone now and including that knowledge here will keep the next dev from needing to rediscover what you've learned.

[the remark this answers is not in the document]
FifthPotato: Added more comments. Are those sufficient, or are there more unclear things?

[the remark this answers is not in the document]
korydraughn: Looks good.

korydraughn:
All review comments have been addressed.

Is there anything else left to do for this?

alanking:
Consider modifying this test to only test: https://github.com/irods/irods/blob/344f9c70a97d5adfe50bd23e65bbe9547cf78e40/scripts/irods/test/test_imeta_error_handling.py#L79-L89

It currently accepts `CAT_SQL_ERR` as a value but it should only allow `CATALOG_ALREADY_HAS_ITEM_BY_THAT_NAME`. If that works, you might be able to link and close https://github.com/irods/irods/issues/7440 as well. Thoughts?

alanking:
Oh, it looks like you already addressed that test - my bad. I still think we might still be able to link #7440, though. Will defer to others. Also, that comment right above about unixODBC versions > 2.3.5 can probably be removed now that you've fixed the underlying issue.

korydraughn:
@FifthPotato Please investigate whether this PR addresses #7440 and post your findings here.

FifthPotato:
> @FifthPotato Please investigate whether this PR addresses #7440 and post your findings here.

Yes, I am no longer able to reproduce #7440 with these changes in place.

korydraughn:
Good. Please update that test accordingly.

Also, see if there are other tests which assert on CAT_SQL_ERR and CATALOG_ALREADY_HAS_ITEM_BY_NAME.

trel:
please make sure / assert here that you CAN reproduce #7440 without these changes.

FifthPotato:
Apologies for the force push, I had rebased my branch onto latest main to try and debug the test failure.

FifthPotato:
Core tests (except for Test_Resource_Unixfilesystem) pass without [2e31a0b](https://github.com/irods/irods/pull/7987/commits/2e31a0b68af2643e4d3c73e03bf7a0b37945bc4c). After [2e31a0b](https://github.com/irods/irods/pull/7987/commits/2e31a0b68af2643e4d3c73e03bf7a0b37945bc4c), all of Test_Resource_Unixfilesystem passes. Also, all unit tests pass.

FifthPotato:
> please make sure / assert here that you CAN reproduce #7440 without these changes.

I rebuilt from tip-of-main and can confirm this is reproducible.

trel:
>  I rebuilt from tip-of-main and can confirm this is reproducible.

Excellent.

I think I'm ready to see the squashing.

scripts/irods/test/test_resource_types.py
```diff
@@ -786,7 +786,7 @@ def test_detached_mode_host_not_in_host_list__issue_4421(self):
             lib.make_file(file1, 100)
 
             # put small file
-            self.admin.assert_icommand("iput -R %s %s" % (resource_name, file1), 'STDERR', 'USER_SOCK_CONNECT_ERR')  # iput
+            self.admin.assert_icommand("iput -R %s %s" % (resource_name, file1), 'STDERR', '|'.join(['-347000 USER_SOCK_CONNECT_TIMEDOUT', '-305000 USER_SOCK_CONNECT_ERR']), use_regex = True)  # iput
```
korydraughn, line 789: We can make this clearer. Consider the following:
```py
_, err, ec = self.admin.run_icommand(<cmd>)
self.assertNotEqual(ec, 0)
self.assertTrue('-347000 USER_SOCK_CONNECT_TIMEDOUT' in err or '-305000 USER_SOCK_CONNECT_ERR' in err)
```

That's probably easier for everyone to understand. This applies to the other calls too.

Thoughts?

korydraughn: Agreed.

FifthPotato:
All wrapped up now. Re-ran the modified test to make sure it still works.

scripts/irods/test/test_resource_types.py
```diff
@@ -786,8 +786,9 @@ def test_detached_mode_host_not_in_host_list__issue_4421(self):
             lib.make_file(file1, 100)
 
             # put small file
-            self.admin.assert_icommand("iput -R %s %s" % (resource_name, file1), 'STDERR', 'USER_SOCK_CONNECT_ERR')  # iput
-            self.admin.assert_icommand("ils -L %s" % file1, 'STDERR_SINGLELINE', "does not exist")  # should not be listed
+            _, err, ec = self.admin.run_icommand("iput -R %s %s" % (resource_name, file1))
+            self.assertNotEqual(ec, 0)
+            self.assertTrue('-347000 USER_SOCK_CONNECT_TIMEDOUT' in err or '-305000 USER_SOCK_CONNECT_ERR' in err)
```
trel, line 791: > And more explanation as to why both are acceptable, please.

Please add an explanation.   Here, and copy/paste twice again below.

FifthPotato:
Squished to 2.

scripts/irods/test/test_resource_types.py
```diff
@@ -786,8 +786,12 @@ def test_detached_mode_host_not_in_host_list__issue_4421(self):
             lib.make_file(file1, 100)
 
             # put small file
-            self.admin.assert_icommand("iput -R %s %s" % (resource_name, file1), 'STDERR', 'USER_SOCK_CONNECT_ERR')  # iput
-            self.admin.assert_icommand("ils -L %s" % file1, 'STDERR_SINGLELINE', "does not exist")  # should not be listed
+            _, err, ec = self.admin.run_icommand("iput -R %s %s" % (resource_name, file1))
+            self.assertNotEqual(ec, 0)
+
+            # Both are okay, since either case is possible when connecting to an invalid host
+            # It depends on if the host rejects the connection or ignores it (i.e. a "stealth" port)
```
korydraughn, line 793: Please add periods to the ends of these sentences so no one is confused later.

_Apply the same change to the other locations._

korydraughn (commented):
Once the review comment is resolved, squash it and then we'll give the approval after one final look.

FifthPotato:
Added the periods, and re-squashed.

alanking (approved):
Are we going to link #7440 to one of these commits? I think that's the last thing that I see. Change itself looks good.

FifthPotato:
> Are we going to link #7440 to one of these commits? I think that's the last thing that I see. Change itself looks good.

Done!

alanking:
Nice. # it

korydraughn (approved):

FifthPotato:
Octothorpe'd.

alanking:

PR #8030, opened 2024-11-14
[7734] Add tests for new iadmin behavior (main)

FifthPotato:
Depends on https://github.com/irods/irods_client_icommands/pull/506 
Works for what it tests. 

Still needs a full test run.

### DIFF
--- a/scripts/irods/test/test_iadmin.py
+++ b/scripts/irods/test/test_iadmin.py
@@ -1685,6 +1685,12 @@ class Test_Iadmin(resource_suite.ResourceBase, unittest.TestCase):
         finally:
             self.admin.run_icommand(['iadmin', 'rmgroup', test_group_name])
 
+    def test_iadmin_exits_nonzero_on_nonexistent_command__issue_7734(self):
+        # Nonexistent command
+        ec, _, _ = self.admin.assert_icommand(['iadmin', 'nonexistentcommand'], 'STDERR');
+        self.assertNotEqual(ec, 0)
+
+
 class Test_Iadmin_Resources(resource_suite.ResourceBase, unittest.TestCase):
 
     def setUp(self):
@@ -2132,6 +2138,34 @@ class Test_Iadmin_modrepl(resource_suite.ResourceBase, unittest.TestCase):
         finally:
             if os.path.exists(original_file_path):
                 os.unlink(original_file_path)
+
+    def test_iadmin_modrepl_exits_nonzero_on_failure__issue_7734(self):
+        # Bad parameters to modrepl
+        ec, _, _ = self.admin.assert_icommand(['iadmin', 'modrepl', 'foo', 'bar', 'baz'], 'STDERR');
+        self.assertNotEqual(ec, 0)
+
+        # Nonexistent data object
+        ec, _, _ = self.admin.assert_icommand(['iadmin', 'modrepl', 'logical_path', '/tempZone/object/does/not/exist', 'replica_number', '0', 'DATA_COMMENTS', 'something'], 'STDERR');
+        self.assertNotEqual(ec, 0)
+
+        test_file_name = 'test_file_name_issue_7734'
+        self.admin.assert_icommand(['itouch', test_file_name])
+
+        # Nonexistent replica number
+        ec, _, _ = self.admin.assert_icommand(['iadmin', 'modrepl', 'logical_path', self.admin.session_collection + '/' + test_file_name, 'replica_number', '123', 'DATA_COMMENTS', 'something'], 'STDERR');
+        self.assertNotEqual(ec, 0)
+
+        # Invalid replica number
+        ec, _, _ = self.admin.assert_icommand(['iadmin', 'modrepl', 'logical_path', self.admin.session_collection + '/' + test_file_name, 'replica_number', 'abc', 'DATA_COMMENTS', 'something'], 'STDERR');
+        self.assertNotEqual(ec, 0)
+
+        # Invalid data object ID
+        ec, _, _ = self.admin.assert_icommand(['iadmin', 'modrepl', 'data_id', 'abc', 'replica_number', '123456789', 'DATA_COMMENTS', 'something'], 'STDERR');
+        self.assertNotEqual(ec, 0)
+
+        # Nonexistent data object ID
+        ec, _, _ = self.admin.assert_icommand(['iadmin', 'modrepl', 'data_id', '1234567', 'replica_number', '123456789', 'DATA_COMMENTS', 'something'], 'STDERR');
+        self.assertNotEqual(ec, 0)
 
 class test_moduser_user(unittest.TestCase):
     """Test modifying a user."""


scripts/irods/test/test_iadmin.py
```diff
@@ -1685,6 +1685,43 @@ def test_iadmin_mkgroup_returns_correct_error_on_duplicate_group__issue_7599(sel
         finally:
             self.admin.run_icommand(['iadmin', 'rmgroup', test_group_name])
 
+    def test_iadmin_exits_nonzero_on_error_cases__issue_7734(self):
+        # Nonexistent command
+        out, err, ec = self.admin.run_icommand(['iadmin', 'nonexistentcommand']);
```
korydraughn, line 1690: Replace unused-assigned variables with `_` so that it's clear they aren't used.
```py
_, _, ec = self.admin.run_icommand(...)
```

scripts/irods/test/test_iadmin.py
```diff
@@ -1685,6 +1685,43 @@ def test_iadmin_mkgroup_returns_correct_error_on_duplicate_group__issue_7599(sel
...
+
+        # Invalid data object ID
+        out, err, ec = self.admin.run_icommand(['iadmin', 'modrepl', 'data_id', 'abc', 'replica_number', '123456789', 'DATA_COMMENTS', 'something']);
+        self.assertNotEqual(ec, 0)
+
+
+        # Nonexistent data object ID
+        out, err, ec = self.admin.run_icommand(['iadmin', 'modrepl', 'data_id', '1234567', 'replica_number', '123456789', 'DATA_COMMENTS', 'something']);
+        self.assertNotEqual(ec, 0)
+
```
alanking, line 1724:
```suggestion
        # Invalid data object ID
        out, err, ec = self.admin.run_icommand(['iadmin', 'modrepl', 'data_id', 'abc', 'replica_number', '123456789', 'DATA_COMMENTS', 'something']);
        self.assertNotEqual(ec, 0)

        # Nonexistent data object ID
        out, err, ec = self.admin.run_icommand(['iadmin', 'modrepl', 'data_id', '1234567', 'replica_number', '123456789', 'DATA_COMMENTS', 'something']);
        self.assertNotEqual(ec, 0)


```

scripts/irods/test/test_iadmin.py
korydraughn, line 1690: I don't think `.run_icommand()` prints the executed command to the junit xml output file.

By all means, please take a minute or two to confirm that. If it turns out that `.run_icommand()` doesn't include the executed command in the junit xml output, replace it with `.assert_icommand()`. We know that prints the commands to the file.

**Keep in mind that `ec` becomes the first return value for `.assert_icommand()`.**

To keep `.assert_icommand()` from failing the test, you can give it things to check that are guaranteed to exist in the output. For example:
```py
# For situations where you expect output on stdout, but don't care what it is.
ec, _, _ = self.admin.assert_icommand(cmd, 'STDOUT')

# For situations where you expect output on stderr, but don't care what it is.
ec, _, _ = self.admin.assert_icommand(cmd, 'STDERR')

# For situations where you expect output on stdout and need to pass it something
# that will appear in that output. Same thing for stderr.
ec, _, _ = self.admin.assert_icommand(cmd, 'STDOUT', [' '])
```

Check out some of the other tests for examples. Should be easy to spot them in newer tests.

scripts/irods/test/test_iadmin.py
```diff
@@ -1685,6 +1685,43 @@ def test_iadmin_mkgroup_returns_correct_error_on_duplicate_group__issue_7599(sel
...
+        out, err, ec = self.admin.run_icommand(['iadmin', 'modrepl', 'foo', 'bar', 'baz']);
+        self.assertNotEqual(ec, 0)
+
+        # Nonexistent data object 
+        out, err, ec = self.admin.run_icommand(['iadmin', 'modrepl', 'logical_path', '/tempZone/object/does/not/exist', 'replica_number', '0', 'DATA_COMMENTS', 'something']);
+        self.assertNotEqual(ec, 0)
+
+        try:
+            test_file_name = 'test_file_name_issue_7734'
+            self.admin.assert_icommand(['itouch', self.admin.home_collection + '/' + test_file_name])
```
korydraughn, line 1703: This is _almost_ equivalent to:
```py
self.admin.assert_icommand(['itouch', test_file_name])
```

The benefit is that by NOT including the path prefix, the data object is created in `self.admin.session_collection`. Prefer that collection where possible because it is automatically cleaned up by the test harness, meaning the `irm` in the `finally` block can be removed.

scripts/irods/test/test_iadmin.py
```diff
@@ -1685,6 +1685,43 @@ def test_iadmin_mkgroup_returns_correct_error_on_duplicate_group__issue_7599(sel
...
+            self.assertNotEqual(ec, 0)
+
+        finally:
+            self.admin.run_icommand(['irm', '-f', self.admin.home_collection + '/' + test_file_name])
+
+        # Invalid data object ID
+        out, err, ec = self.admin.run_icommand(['iadmin', 'modrepl', 'data_id', 'abc', 'replica_number', '123456789', 'DATA_COMMENTS', 'something']);
+        self.assertNotEqual(ec, 0)
+
+
```
korydraughn, line 1720: Collapse empty lines to one.

korydraughn (commented):
Looking good.

scripts/irods/test/test_iadmin.py
```diff
@@ -1685,6 +1685,43 @@ def test_iadmin_mkgroup_returns_correct_error_on_duplicate_group__issue_7599(sel
...
+            # Nonexistent replica number
+            out, err, ec = self.admin.run_icommand(['iadmin', 'modrepl', 'logical_path', self.admin.home_collection + '/' + test_file_name, 'replica_number', '123', 'DATA_COMMENTS', 'something']);
+            self.assertNotEqual(ec, 0)
+
+            # Invalid replica number
+            out, err, ec = self.admin.run_icommand(['iadmin', 'modrepl', 'logical_path', self.admin.home_collection + '/' + test_file_name, 'replica_number', 'abc', 'DATA_COMMENTS', 'something']);
+            self.assertNotEqual(ec, 0)
+
+        finally:
+            self.admin.run_icommand(['irm', '-f', self.admin.home_collection + '/' + test_file_name])
```
alanking, line 1714: I believe this is covered in part by `test_iadmin.Test_Iadmin_modrepl.test_input_parameters_invalid` and/or these assertions would fit nicely into that test. Perhaps we can move this test into that class or modify the existing test to include assertions around the return code since this seems mainly to deal with `iadmin modrepl` (except for the first assertion, which should perhaps stay in a test on its own). Thoughts?

alanking (commented):
Have you tried out `iadmin` in "interactive" mode to see how this change affects that? Currently, interactive mode does not quit when an unrecognized command is fed to it. I'm not sure that it's worth the effort to make an automated test for that, but maybe just something we can confirm.

scripts/irods/test/test_iadmin.py
korydraughn: Yeah, moving this test into the class sounds good to me. I like the idea of keeping these assertions together since they target a specific issue.

trel:
please update the commit message to have more detail / specifics, not sure what 'new iadmin behavior' means...

alanking (commented):
Looks about right to me. Tests are all passing and whatnot?

And yes, commit message and/or body could stand to be more detailed.

FifthPotato:
> Looks about right to me. Tests are all passing and whatnot?
> 
> And yes, commit message and/or body could stand to be more detailed.

This one works-- haven't run the full test suite yet. Wanted to make sure there weren't any more major changes before doing that.
I'll touch up the commit message.

alanking:
Okay. Please confirm that interactive mode behaves and I'll be satisfied (with other tests passing as well, of course).

FifthPotato:
Just verified that it doesn't break interactive mode (unrecognized command will still prompt you afterward).
Will run tests now!

FifthPotato:
All core and unit tests passed.

FifthPotato:
Touched up the commit message, too.

trel:
test separation and commit message looks good.

anyone have strong python opinions about the asserts?   right now just nonzero, which seems to cover the main request/complaint.

korydraughn:
No. The assertions were addressed by earlier review comments.

I think we're in a good spot. Let's squash it.

FifthPotato:
Squished.

alanking (approved):
\# it

korydraughn:
Please cherry pick to 4-3-stable

PR #8035, opened 2024-11-22
[8017] Modify schema file to enforce required attributes (main)

FifthPotato:
Not sure if anything else is required for this.

### DIFF
--- a/schemas/configuration/v4/server_config.json.in
+++ b/schemas/configuration/v4/server_config.json.in
@@ -66,7 +66,8 @@
                     "items": {"type": "string"},
                     "uniqueItems": true
                 }
-            }
+            },
+            "required": ["control_type", "users"]
         },
         "maximum_connections": {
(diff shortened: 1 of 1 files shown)


korydraughn:
See if you can capture this in a test. We want to avoid going backwards as much as possible.

I think IrodsController and IrodsConfig are probably what you want.

FifthPotato:
> See if you can capture this in a test. We want to avoid going backwards as much as possible.
> 
> I think IrodsController and IrodsConfig are probably what you want.

I think we've now agreed not to do a test for this. In that case, is there anything else necessary for this?

alanking (approved):
Possibly include some language in the commit message about *which* required attributes are being enforced? I'm not so hung up on that. Up to you.

alanking:
@FifthPotato - Also, please cherry-pick to 4-3-stable. Thanks

PR #8053, opened 2024-12-02
[#7410] Provide CMake cache variable for package configuration directory

SwooshyCueb:
Despite what CMake's documentation might say, getting `find_package` to look in `lib64` for CMake package configuration files is a very ugly business that we would rather try to avoid. This PR puts our CMake package configuration back in `lib` for EL and provides a cache variable `IRODS_INSTALL_CMAKEDIR` should the builder need to tweak things.

Part of #7410, as this work should've been in the PR for changing where our libraries end up.

Still testing, so draft for now.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -286,10 +286,12 @@ add_subdirectory(unit_tests)
 include("${CMAKE_CURRENT_SOURCE_DIR}/cmake/packaging.cmake")
 
 include(CMakePackageConfigHelpers)
+set(IRODS_INSTALL_CMAKECONFDIR "${IRODS_INSTALL_CMAKEDIR}/irods-${IRODS_VERSION_MAJOR}.${IRODS_VERSION_MINOR}")
+
 configure_package_config_file(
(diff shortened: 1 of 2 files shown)


SwooshyCueb:
Looks like this works on all distros we support, marking ready for review

alanking (approved):
Seems good to me. Maybe get one more approval before #'ing. @korydraughn / @FifthPotato / @MartinFlores751

SwooshyCueb:
I've done some digging in CMake's issue tracker and it appears that the reason we're running into problems here is because we're (intentionally) calling `find_package` before any languages are enabled, so CMake hasn't yet inspected the toolchain to determine the target architecture.
https://gitlab.kitware.com/cmake/cmake/-/issues/25157#note_1417979

This is a problem that will eventually go away as we approach Normal and Boring packaging, but for now, this PR provides a suitable workaround.

korydraughn:
You were able to reproduce what I was seeing before this change, correct?

SwooshyCueb:
> You were able to reproduce what I was seeing before this change, correct?

Yes, but only on EL8 and EL9

korydraughn (approved):
Good. Pound it.

SwooshyCueb:
\#'d

Issue #8042, opened 2024-12-02
Deprecate unused globals in resource manager implementation


korydraughn:
## Deprecation

https://github.com/irods/irods/blob/a6d646c37423f4e8fc25b21d4e15c2025eaed984/server/core/include/irods/irods_resource_manager.hpp#L22-L24

_This applies to 4-3-stable only._

PR #8062, opened 2024-12-06
[#8042] Deprecate unused global variables in resource manager implementation (4-3-stable)

korydraughn:
Will compile server and icommands shortly.


korydraughn:
Compiles and passes C++ unit tests.

Adding pound.

korydraughn:
#'d

Issue #8018, opened 2024-11-14
Deprecate `plugin_home` property in irods_environment.json


korydraughn:
## Deprecation

- Add `plugin_directory` as the preferred property
- Update documentation - configuration page (irods/irods_docs)
- Print/Log a message to stderr when `plugin_home` is used

_This applies to 4-3-stable only._

alanking:
Looks like this is about to removed in main, so we probably only need a checkbox for 4-3-stable, correct?

korydraughn:
Yes, that's correct. Good catch.

korydraughn:
To avoid a potential ABI break, this issue will only do the following:
- Add deprecation attribute
- Document deprecation in release notes

The code will remain as is for 4.3. The option will be replaced in iRODS 5.

korydraughn:
Marking the `irodsPluginHome` member variable in `RodsEnvironment`/`rodsEnv` as deprecated requires modifications to the POD.

To keep things simple, we'll handle this through documentation.

PR #8063, opened 2024-12-06
[#8018] Deprecate plugins_home property in irods_environment.json (4-3-stable)

korydraughn:
I'm still thinking about this, but feel free to leave comments.

### DIFF
--- a/lib/core/include/irods/getRodsEnv.h
+++ b/lib/core/include/irods/getRodsEnv.h
@@ -55,13 +55,19 @@ typedef struct RodsEnvironment {
 
     // =-=-=-=-=-=-=-
     // override of plugin installation directory
-    char irodsPluginHome[MAX_NAME_LEN];
+    __attribute__((deprecated("Use irodsPluginDirectory"))) char irodsPluginHome[MAX_NAME_LEN];
+    // TODO Hmm, adding a new member will result in an ABI break. Users will have to
+    // recompile their C/C++ clients. If we don't add the new member, then users can't
+    // migrate their clients to the new variable.
 
     // =-=-=-=-=-=-=-
     // TCP keepalive configurations
     int tcp_keepalive_intvl;
     int tcp_keepalive_probes;
     int tcp_keepalive_time;
+
+    // TODO This will result in an ABI break.
+    char irodsPluginDirectory[MAX_NAME_LEN];
 } rodsEnv;
 
 #ifdef __cplusplus
--- a/lib/core/src/getRodsEnv.cpp
+++ b/lib/core/src/getRodsEnv.cpp
@@ -348,8 +348,13 @@ extern "C" {
             irods::KW_CFG_IRODS_CONNECTION_POOL_REFRESH_TIME,
             _env->irodsConnectionPoolRefreshTime );
 
-        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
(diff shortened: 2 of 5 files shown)


lib/core/include/irods/getRodsEnv.h
```diff
@@ -55,13 +55,19 @@ typedef struct RodsEnvironment {
 
     // =-=-=-=-=-=-=-
     // override of plugin installation directory
-    char irodsPluginHome[MAX_NAME_LEN];
+    __attribute__((deprecated("Use irodsPluginDirectory"))) char irodsPluginHome[MAX_NAME_LEN];
+    // TODO Hmm, adding a new member will result in an ABI break. Users will have to
+    // recompile their C/C++ clients. If we don't add the new member, then users can't
+    // migrate their clients to the new variable.
```
alanking, line 61: Dumb question: I thought an ABI break only occurred when members are removed or added in the middle of the struct? Hmm... I guess if the `RodsEnvironment` is a member of another struct it would cause a misalignment...

korydraughn: Any changes to a data structure can cause ABI issues because there's a change in the layout, size, etc.

I want to make the iRODS 5 member available, but I'm not confident it's possible without breaking the ABI (forcing people to recompile code) or using come special compiler flag (yuck).

lib/core/src/getRodsEnv.cpp
```diff
@@ -348,8 +348,13 @@ extern "C" {
             irods::KW_CFG_IRODS_CONNECTION_POOL_REFRESH_TIME,
             _env->irodsConnectionPoolRefreshTime );
 
-        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
-        capture_string_property(irods::KW_CFG_IRODS_PLUGINS_HOME, _env->irodsPluginHome, MAX_NAME_LEN);
+        capture_string_property(irods::KW_CFG_IRODS_PLUGIN_DIRECTORY, _env->irodsPluginHome, MAX_NAME_LEN);
+
+        // If the "plugin_directory" is configured, ignore "plugins_home".
+        if (std::strlen(_env->irodsPluginHome) == 0) {
```
korydraughn, line 354: Use `is_empty_string`.

lib/core/src/getRodsEnv.cpp
```diff
@@ -642,10 +647,14 @@ extern "C" {
...
 
-        env_var = irods::KW_CFG_IRODS_PLUGINS_HOME;
-        capture_string_env_var(
-            env_var,
-            _env->irodsPluginHome );
+        env_var = irods::KW_CFG_IRODS_PLUGIN_DIRECTORY;
+        capture_string_env_var(env_var, _env->irodsPluginHome);
+
+        // If the "plugin_directory" is configured, ignore "plugins_home".
+        if (std::strlen(_env->irodsPluginHome) == 0) {
```
korydraughn, line 654: Use `is_empty_string`.

lib/core/include/irods/getRodsEnv.h
```diff
@@ -55,13 +55,19 @@ typedef struct RodsEnvironment {
...
 
     // =-=-=-=-=-=-=-
     // TCP keepalive configurations
     int tcp_keepalive_intvl;
     int tcp_keepalive_probes;
     int tcp_keepalive_time;
+
+    // TODO This will result in an ABI break.
+    char irodsPluginDirectory[MAX_NAME_LEN];
```
korydraughn, line 70: Handle TODO

Issue #8001, opened 2024-11-08
Investigate correctness of `ips` output


korydraughn:
- [ ] main
- [ ] 4-3-stable
---
### Investigation

The first code block writes the information to a file that is to be consumed for `ips` information.

https://github.com/irods/irods/blob/759a039e73b93f0904344d72edbefca8b9df739a/server/core/src/procLog.cpp#L92-L95

This code block reads the file and returns the information to the client (e.g. `ips`).

https://github.com/irods/irods/blob/759a039e73b93f0904344d72edbefca8b9df739a/server/core/src/procLog.cpp#L146-L151

Notice the order of the first few inputs/outputs between to the two code blocks. They do not align. Is this correct/expected?

Please investigate.

korydraughn:
For iRODS 5, the relevant source lines to review are located here:
- https://github.com/irods/irods/pull/7988#discussion_r1833205589

Those lines address this issue by swizzling the order of the inputs, but the change needs to be checked for correctness.

korydraughn:
After looking at the implementation of `ips` and the functions in the server which support it, I'm sure the implementation is incorrect.

To fix this, `rsComm->clientUser.userName` and `rsComm->proxyUser.userName` need to be swapped.

korydraughn:
#7988 is merged, resolving this issue for the main checkbox.

PR #8067, opened 2024-12-06
[#8001] Write client and proxy user info to ips data file in correct order (4-3-stable)

korydraughn:
I'll confirm at the bench whether this works.

This doesn't require running the full suite since the same data is being presented, just in a different order.


korydraughn:
Ran `ips` with other agents active and the output worked as expected.

This is now ready.

alanking (approved):
Nice. # it

korydraughn:
#'d

Issue #6239, opened 2022-03-14
rodsadmin user moves folder


tedgin:
### iRODS Version, OS and Version
iRODS 4.2.8, CentOS 7

### What did you try to do?
A  user has a collection named `/iplant/home/user/ Ss  /XXXX_vX.X.X_analysis1-2021-10-27-12-19-57.5`. The parent collection ` Ss  ` has a single leading blank and two trailing blanks. The collection contains a child collection named `index`, which contains a data object named `ht2`. Using the Jargon library, the admin user `de-irods` renames the collection to `/iplant/trash/home/de-irods/user/XXXX_vX.X.X_analysis1-2021-10-27-12-19-57.5.0aUcVvx`. It does NOT use the built in move to trash functionality.  It uses a standard move/rename.  Notices that the appends an extension to the collection's original name.

### Expected behavior

I expect the collection and all of its contents to be renamed appropriately, and I expect the physical path of the replica to change similarly.

### Observed behavior (including steps to reproduce, if applicable)

The collection is renamed correctly. However the child collection and data object replica path aren't.  The replica path remains unchaged, but the path to the child collection, `index`,  gets mangled, orphaning it. Here's what's in the ICAT.

```
parent_coll_name | /iplant/home/user/ Ss  //XXXX_vX.X.X_analysis1-2021-10-27-12-19-57.5.0aUcVvx
coll_name        | /iplant/home/user/ Ss  //XXXX_vX.X.X_analysis1-2021-10-27-12-19-57.5.0aUcVvxindex
```

There are no errors in the logs.

trel:
So, is this a Jargon issue?  Or do you think this is in the server?

tedgin:
There might be a Jargon issue too, but the server shouldn't allow children to become orphans in the ICAT.

trel:
Agreed - will try to reproduce.   Is the rodsadmin part necessary?   Can a rodsuser move something into trash manually and see the same problem?

tedgin:
I'm not certain.  The DE doesn't provide a way for the user to move something to trash as the user. 

This issue has existed for a while, but this is the first time I observed it in 4.2.8. When I observed it in iRODS 4.1.10, I didn't report it, because iRODS 4.1.10 was EOL at the time. 

[...]

PR #8076, opened 2024-12-11
[6239] fix for mangled subcollection name upon rename (main)

tsmeele:
Use RDBMS internal function to calculate stringlength as input to substr() operations, rather than relying on a client-calculated literal. This avoids differences in interpretation of string length related to unicode characters that span multiple bytes.


trel:
is this safe across mysql/mariadb and postgresql?

tsmeele:
The length() function is standard across SQL implementations.  Note that chosen string encoding in the RDBMS influences the actual result returned by length().  The proposed solution takes advantage of the alignment of length(), substr() and string encoding within the RDBMS of the vendor, rather than requiring the client to anticipate on differences across vendors and encodings.

SwooshyCueb:
> The proposed solution takes advantage of the alignment of length(), substr() and string encoding within the RDBMS of the vendor, rather than requiring the client to anticipate on differences across vendors and encodings.

I certainly like the sound of that

trel:
Excellent.

I see clang-format has an opinion...

Otherwise, I think this should turn green on the next push.

korydraughn (commented):
Very nice.

We'll need to add a test based on this: https://github.com/irods/irods/issues/6239#issuecomment-2535223519

We'll take care of that.

plugins/database/src/db_plugin.cpp
```diff
@@ -11375,17 +11375,17 @@ irods::error db_move_object_op(
         len = strlen( collNameSlash );
         snprintf( collNameSlashLen, 10, "%d", len );
         cllBindVars[cllBindVarCount++] = newCollName;
-        cllBindVars[cllBindVarCount++] = ocLenStr;
+        cllBindVars[cllBindVarCount++] = oldCollName;
         cllBindVars[cllBindVarCount++] = newCollName;
-        cllBindVars[cllBindVarCount++] = ocLenStr;
-        cllBindVars[cllBindVarCount++] = collNameSlashLen;
```
alanking, line 11381: We might be able to remove `ocLenStr` and `collNameSlashLen` as they appear to only be used for this query.

tsmeele: Good point, I have added a change to remove the unused declarations related to length calculation

alanking (commented):
Looks good. I'll try to get a test written for this so we can get this in ASAP

alanking:
Just wanted to echo the comment I made in https://github.com/irods/irods/issues/6239#issuecomment-2551869538...

After trying this out manually and writing some automated tests, it appears that the issue concerning renaming collections with multi-byte characters in the path remains even with these changes. Will leave this open while we look for another solution...

tsmeele:
The method in principle should fix the issue, yet I had overlooked a subtle increment of the old ocLen during conversion to string.  Now added the increment to the SQL.   You may want to add a test without multi-byte as well to ensure behavior is correct on single-byte unicode characters as well.

korydraughn:
Sounds good. We'll confirm things work as intended.

Issue #8088, opened 2024-12-12
Deprecate GenQuery1 string-based parser, `fillGenQueryInpFromStrCond`


korydraughn:
This applies to iRODS 4 only.

PR #8098, opened 2024-12-17
[#8088] Deprecate `fillGenQueryInpFromStrCond()` (4-3-stable)

korydraughn:
Code changes compile.


alanking (approved):
This function is kind of the only provided way of using GenQuery1 (besides populating the structs oneself), so the message recommending using GenQuery2 makes sense. # it

korydraughn:
Pounded.

PR #8105, opened 2024-12-20
[7778] User removal now also removes ACLs (main)

FifthPotato:
Taking suggestions, as always

### DIFF
--- a/plugins/database/src/db_plugin.cpp
+++ b/plugins/database/src/db_plugin.cpp
@@ -4505,11 +4505,23 @@ irods::error db_del_user_re_op(
                  &icss );
     if ( status != 0 && status != CAT_SUCCESS_BUT_WITH_NO_INFO ) {
         char errMsg[MAX_NAME_LEN + 40];
-        log_db::info("chlDelUserRE delete password failure {}", status);
+        log_db::error("chlDelUserRE delete password failure {}", status);
         snprintf( errMsg, sizeof errMsg, "Error removing password entry" );
         addRErrorMsg( &_ctx.comm()->rError, 0, errMsg );
         _rollback( "chlDelUserRE" );
         return ERROR( status, "Error removing password entry" );
+    }
+
+    // Remove any ACLs associated with the user
+    cllBindVars[cllBindVarCount++] = iValStr;
+
+    status = cmlExecuteNoAnswerSql("delete from R_OBJT_ACCESS where user_id=?", &icss);
+
+    if (status != 0 && status != CAT_SUCCESS_BUT_WITH_NO_INFO) {
+        log_db::error("chlDelUserRE delete ACL failure {}", status);
+        addRErrorMsg(&_ctx.comm()->rError, 0, "Error removing ACLs");
+        _rollback("chlDelUserRE");
+        return ERROR(status, "Error removing ACLs");
     }
 
     /* Remove both the special user_id = group_user_id entry and any
--- a/scripts/irods/test/test_iadmin.py
+++ b/scripts/irods/test/test_iadmin.py
@@ -1708,6 +1708,138 @@ class Test_Iadmin(resource_suite.ResourceBase, unittest.TestCase):
         ec, _, _ = self.admin.assert_icommand(['iadmin', 'nonexistentcommand'], 'STDERR');
         self.assertNotEqual(ec, 0)
 
+    def test_iadmin_rmuser_removes_associated_acls__issue_7778(self):
+        test_user_name = "test_user_issue_7778"
+        test_group_name = "test_group_issue_7778"
+        test_collection_name = "test_collection_issue_7778"
+        test_dataobj_name = "test_data_object_issue_7778"
+        try:
+            self.admin.assert_icommand(["imkdir", test_collection_name])
+            self.admin.assert_icommand(["itouch", test_dataobj_name])
+            self.admin.assert_icommand(["iadmin", "mkgroup", test_group_name])
+            self.admin.assert_icommand(["iadmin", "mkuser", test_user_name, "rodsuser"])
+            self.admin.assert_icommand(
+                ["iadmin", "atg", test_group_name, test_user_name]
+            )
+            self.admin.assert_icommand(
+                ["ichmod", "read_object", test_group_name, test_collection_name]
+            )
+            self.admin.assert_icommand(
+                ["ichmod", "modify_object", test_user_name, test_collection_name]
+            )
+            self.admin.assert_icommand(
+                ["ichmod", "read_object", test_group_name, test_dataobj_name]
+            )
+            self.admin.assert_icommand(
+                ["ichmod", "modify_object", test_user_name, test_dataobj_name]
+            )
+
+            # Collect IDs of user and group
+            _, out, _ = self.admin.assert_icommand(
+                [
+                    "iquest",
+                    "%s",
+                    "SELECT USER_ID WHERE USER_NAME = '%s' || = '%s'"
+                    % (test_user_name, test_group_name),
+                ],
+                "STDOUT_MULTILINE",
+                r"^[0-9]+$",
+                use_regex=True,
+            )
+
+            # Create iquest query string for the IDs-- we don't care about order, we just want both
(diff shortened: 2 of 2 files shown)


scripts/irods/test/test_iadmin.py
```diff
@@ -1708,6 +1708,36 @@ def test_iadmin_exits_nonzero_on_nonexistent_command__issue_7734(self):
         ec, _, _ = self.admin.assert_icommand(['iadmin', 'nonexistentcommand'], 'STDERR');
         self.assertNotEqual(ec, 0)
 
+    def test_iadmin_rmuser_removes_associated_data__issue_7778(self):
+        test_user_name = 'test_user_issue_7778'
+        test_group_name = 'test_group_issue_7778'
+        test_collection_name = 'test_collection_issue_7778'
```
alanking, line 1714: Should we add a test (or add to this test) for data objects as well? I know from the implementation that it "should" work, but, y'know... "blackbox testing" and all that...

scripts/irods/test/test_iadmin.py
```diff
@@ -1708,6 +1708,36 @@ def test_iadmin_exits_nonzero_on_nonexistent_command__issue_7734(self):
...
+            _, out, _ = self.admin.assert_icommand(['iquest', '%s', "SELECT USER_ID WHERE USER_NAME = '%s' || = '%s'" % (test_user_name, test_group_name) ], 'STDOUT_MULTILINE', r'^[0-9]+$', use_regex=True)
+
+            # Create iquest query string for the IDs-- we don't care about order, we just want both
+            query_string = "= '%s' || = '%s'" % tuple(out.strip().split('\n'))
+
+            # Should see the permissions as expected
+            self.admin.assert_icommand(['iquest', "SELECT COLL_ACCESS_USER_ID, COLL_ACCESS_NAME WHERE COLL_NAME = '%s' AND COLL_ACCESS_USER_ID %s" % (os.path.join(self.admin.session_collection, test_collection_name), query_string)], 'STDOUT', ['COLL_ACCESS_NAME = read_object', 'COLL_ACCESS_NAME = modify_object'])
+
+            self.admin.assert_icommand(['iadmin', 'rmuser', test_user_name])
+            self.admin.assert_icommand(['iadmin', 'rmgroup', test_group_name])
```
alanking, line 1733: Even though I know the removal of ACLs is tied to the user ID... It feels like it may be worth checking ACLs in between removal of the user and the group to make sure the user's ACLs are gone but the group's ACLs remain. Thoughts?

Also, let's make sure the data itself remains (the name of the test concerns me :p).

plugins/database/src/db_plugin.cpp
```diff
@@ -4512,6 +4512,20 @@ irods::error db_del_user_re_op(
...
 
+    // Remove any ACLs associated with the user
+    cllBindVars[cllBindVarCount++] = iValStr;
+
+    status = cmlExecuteNoAnswerSql("delete from R_OBJT_ACCESS where user_id=?", &icss);
+
+    if (status != 0 && status != CAT_SUCCESS_BUT_WITH_NO_INFO) {
+        char errMsg[MAX_NAME_LEN + 40];
+        log_db::info("chlDelUserRE delete ACL failure {}", status);
+        snprintf(errMsg, sizeof errMsg, "Error removing ACLs");
```
alanking, line 4523: Can we just... construct a string here?

alanking (commented):
Good find. Most of my suggestions pertain to the test. I realize that the new change is based on the existing code, so... up to you if you want to take that suggestion.

scripts/irods/test/test_iadmin.py
FifthPotato: Sure, I can look into that. More test cases = better, right?

trel:
looking pretty good to me.

FifthPotato:
Core and unit tests pass.

scripts/irods/test/test_iadmin.py
```diff
@@ -1708,6 +1708,137 @@ def test_iadmin_exits_nonzero_on_nonexistent_command__issue_7734(self):
...
         self.assertNotEqual(ec, 0)
 
+    def test_iadmin_rmuser_removes_associated_acls__issue_7778(self):
+        test_user_name = "test_user_issue_7778"
+        test_group_name = "test_group_issue_7778"
+        test_collection_name = "test_collection_issue_7778"
+        test_dataobj_name = "test_data_object_issue_7778"
+        try:
+            self.admin.assert_icommand(["imkdir", test_collection_name])
+            self.admin.assert_icommand(["itouch", test_dataobj_name])
```
alanking, line 1718: Is this data object cleaned up because it's in the session collection? Just making sure since I noticed we are running `irmdir` on the test collection but not removing the test data object.

plugins/database/src/db_plugin.cpp
```diff
@@ -4512,6 +4512,18 @@ irods::error db_del_user_re_op(
         return ERROR( status, "Error removing password entry" );
     }
 
+    // Remove any ACLs associated with the user
+    cllBindVars[cllBindVarCount++] = iValStr;
+
+    status = cmlExecuteNoAnswerSql("delete from R_OBJT_ACCESS where user_id=?", &icss);
+
+    if (status != 0 && status != CAT_SUCCESS_BUT_WITH_NO_INFO) {
+        log_db::info("chlDelUserRE delete ACL failure {}", status);
```
korydraughn, line 4521: What's the reason for logging as _info_ vs _error?_

alanking (commented):
Looks good to me. I have one last question for something I noticed

scripts/irods/test/test_iadmin.py
```diff
@@ -1708,6 +1708,137 @@ def test_iadmin_exits_nonzero_on_nonexistent_command__issue_7734(self):
...
+                    % (test_dataobj_name, self.admin.session_collection, query_string),
+                ],
+                "STDOUT",
+                ["DATA_ACCESS_NAME = read_object"],
+            )
+            self.assertNotIn("DATA_ACCESS_NAME = modify_object", out)
+
+            self.admin.assert_icommand(["iadmin", "rmgroup", test_group_name])
+
+            # Perms should vanish after deletion of associated users
```
korydraughn, line 1812: `of associated users` -> `of associated group`?

korydraughn (commented):
Good stuff.

plugins/database/src/db_plugin.cpp
FifthPotato: I'm basing this off of some of the other code in the same function (e.g. line 4508). If I had to wager a guess, I think it'd be because the primary error is bubbled up from the rError, so this log message could technically be considered redundant? I'm down to change it if you don't agree, though.

FifthPotato: Done and done.

scripts/irods/test/test_iadmin.py
```diff
@@ -1708,6 +1708,138 @@ def test_iadmin_exits_nonzero_on_nonexistent_command__issue_7734(self):
...
+                    % (test_dataobj_name, self.admin.session_collection, query_string),
+                ],
+                "STDOUT_SINGLELINE",
+                "CAT_NO_ROWS_FOUND: Nothing was found matching your query",
+            )
+        finally:
+            self.admin.run_icommand(["iadmin", "rmuser", test_user_name])
+            self.admin.run_icommand(["iadmin", "rmgroup", test_group_name])
+            self.admin.run_icommand(["irmdir", test_collection_name])
+            self.admin.run_icommand(["irm", test_dataobj_name])
```
alanking, line 1842: This will move the object to the trash. Let's just delete it outright:
```suggestion
            self.admin.run_icommand(["irm", "-f", test_dataobj_name])
```

alanking (commented):
Sorry, one more thing. Then, I think we're good to squash?

FifthPotato:
Squished

alanking (approved):
\# it

korydraughn (approved):

FifthPotato:
Octothorpe'd.

alanking:
@FifthPotato - Please cherry-pick to 4-3-stable.

PR #8118, opened 2025-01-16
[7961] Update test for imeta deprecation message (4-3-stable)

FifthPotato:
Test should no longer fail after this change.

### DIFF
--- a/scripts/irods/test/test_imeta_interactive.py
+++ b/scripts/irods/test/test_imeta_interactive.py
@@ -71,6 +71,7 @@ class Test_Imeta_Interactive(unittest.TestCase):
         cls.test_data_paths_wildcard = cls.test_data_path_base + '%'
         cls.test_univ_attr = 'some_attr'
         cls.test_data_qty = 5
+        cls.deprecation_message = 'NOTE: Interactive mode is deprecated and will be removed in an upcoming release.\n'
         for n in range(cls.test_data_qty):
             data_path = cls.test_data_path_base + str(n)
             cls.admin.assert_icommand(['itouch', data_path])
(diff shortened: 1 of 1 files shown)


scripts/irods/test/test_imeta_interactive.py
```diff
@@ -101,7 +102,7 @@ def tearDown(self):
 
     def test_prompt(self):
         (out, err) = self.get_output()
-        self.assertEqual(out, 'imeta>')
+        self.assertEqual(out, f'{self.deprecation_message}imeta>')
```
alanking, line 105: I see another `imeta>` prompt on L118 - does this not have the same problem?

alanking (commented):
Please confirm that all tests are passing.

FifthPotato:
Ran the test and it does seem to pass now, when it previously didn't.

scripts/irods/test/test_imeta_interactive.py
FifthPotato: Nope, because the deprecation message only prints when you first enter interactive mode. So only this one, which tests the initial prompt, is an issue. Anything else that calls `self.get_output()` (and doesn't check its output) at the top of the test shouldn't be affected.

alanking: Okay, cool. Thanks

alanking (approved):
\# it

korydraughn (commented):
Is that the only test that failed or are there others?

FifthPotato:
> Is that the only test that failed or are there others?

According to Martin's PR, this seems to be the only one-- do you want me to rerun the whole suite on this to check?

alanking:
It would probably be in our best interest to at least confirm that all of `test_imeta_interactive` is passing.

korydraughn:
I agree with @alanking. Running that one test suite is enough.

FifthPotato:
`test_imeta_interactive` passes as expected.

alanking:
Very good. Please add #.

FifthPotato:
Octothorpe added.

PR #8127, opened 2025-01-23
[7835] Deprecate update_deprecated_database_columns.py (4-3-stable)

FifthPotato:
Should be all, I think? Let me know if I missed anything.

### DIFF
--- a/scripts/update_deprecated_database_columns.py
+++ b/scripts/update_deprecated_database_columns.py
@@ -131,4 +131,5 @@ row to be updated. This script must be run on the catalog provider.
         print('Failed getting database connection. Note: This script should be run on the iRODS catalog provider.')
 
 if __name__ == "__main__":
+    print(f'NOTE: {__file__.split("/")[-1]} is deprecated and will be removed in an upcoming version of iRODS.')
     scrub_main()


scripts/update_deprecated_database_columns.py
```diff
@@ -131,4 +131,5 @@ def scrub_main():
         print('Failed getting database connection. Note: This script should be run on the iRODS catalog provider.')
 
 if __name__ == "__main__":
+    print(f'NOTE: {__file__.split("/")[-1]} is deprecated and will be removed in an upcoming version of iRODS.')
```
korydraughn, line 134: What does that print to the terminal?
Why not hard-code the script name into the string instead of parsing it?

FifthPotato: It should print the last slash-delimited element of the absolute path of the filename. (I thought about doing this with os.path, but it feels silly to import more stuff just for a deprecation message.) This is in the off-chance a user renames the script before running it.

alanking: I really don't mind what is done here either way, but would "**This script** is deprecated and..." suffice? No parsing, no script name?

korydraughn: I'm fine with either direction. The script isn't called by anything AFAIK.

@FifthPotato Have you confirmed the script produces the expected output? Please post the terminal output here showing success.

@FifthPotato The decision is yours.

FifthPotato: Yeah, I built and checked the output and it prints as expected: called from same dir and different dir. I thought this way was safest in the very very slim chance that:
1. Someone renames the script
2. Someone calls it from their own stuff
3. All of the above, somehow
This way it's unequivocally clear what is deprecated in that case...but if we have strong opinions on it, I don't mind changing it.

korydraughn: Keep what you have.

korydraughn (approved):
Pound it.

FifthPotato:
Octothorpe'd.

PR #8128, opened 2025-01-23
[6234] Remove some nonexistent includes (4-3-stable)

FifthPotato:
Think I got them all. Let me know if I missed any.

### DIFF
--- a/lib/core/src/obf.cpp
+++ b/lib/core/src/obf.cpp
@@ -53,12 +53,6 @@
 #include "irods/rodsPath.h"
 #include "irods/authenticate.h"
 
-#ifdef _WIN32
-#include "irods/Unix2Nt.hpp"  /* May need something like this for Windows */
-#include "irods/iRODSNtUtil.hpp"    /* May need something like this for Windows */
-#include "irods/rcGlobalExtern.h"
-#endif
-
 #include <boost/filesystem.hpp>
 #include <boost/system/error_code.hpp>
 #include <boost/filesystem/operations.hpp>
--- a/lib/core/src/rcConnect.cpp
+++ b/lib/core/src/rcConnect.cpp
@@ -2,10 +2,6 @@
 #include "irods/rcGlobalExtern.h"
 #include "irods/rcMisc.h"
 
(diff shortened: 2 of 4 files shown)


korydraughn (commented):
Looks correct to me.

Does the server compile?
Do the C++ unit tests pass?

lib/core/src/obf.cpp
```diff
@@ -54,8 +54,6 @@
 #include "irods/authenticate.h"
 
 #ifdef _WIN32
-#include "irods/Unix2Nt.hpp"  /* May need something like this for Windows */
-#include "irods/iRODSNtUtil.hpp"    /* May need something like this for Windows */
 #include "irods/rcGlobalExtern.h"
 #endif
```
alanking, line 58: Can we remove this whole block because `_WIN32`?

FifthPotato: I think so? Since the header wasn't one of the ones listed, I wasn't sure, but if there are no plans to port to windows, I think it'd be fine. 
In the original issue, Terrell specifically said there were no plans to run "the server" on Windows, but this isn't necessarily server-only code. So, I'd like his input on it before I remove it.

FifthPotato:
> Looks correct to me.
> 
> Does the server compile? Do the C++ unit tests pass?

Yup, compiles and unit tests pass.

lib/core/src/obf.cpp
trel: we'd have so much work to do on windows... i think we just assume linux for now.  windows wins will come from python or go...

FifthPotato: Ok, removed.

alanking (commented):
Sweet. Squash em

FifthPotato:
Squished

alanking (approved):
\# it

FifthPotato:
Octothorpe'd.

lib/core/src/obf.cpp
korydraughn: Agreed.

Issue #7942, opened 2024-08-21
`irods_assert.hpp` declares `irods::assert()` but it is never defined


SwooshyCueb:
`lib/core/include/irods/irods_assert.hpp` includes a declaration for `bool irods::assert(bool, bool)`:
https://github.com/irods/irods/blob/bc2a9bf0d8d3587d1aad8bf6a9b3db650e0d6944/lib/core/include/irods/irods_assert.hpp#L4-L10

I cannot find any definition for this function in the repository. The closest thing is `void assert(int)` from `lib/core/src/region.cpp`, but it is in the global namespace, does not return anything, and takes different arguments:
https://github.com/irods/irods/blob/bc2a9bf0d8d3587d1aad8bf6a9b3db650e0d6944/lib/core/src/region.cpp#L169-L173

This is the only thing in this header, so I think the best course of action would be to remove the declaration and deprecate/remove the header.

Issue #1278, opened 2013-11-26
create an assert function


trel:
_Originally Opened: @jasoncoposky (2013-03-13 10:55:40_)
_Originally Closed: 2013-03-13 10:56:53_

create an eirods::assert which will call stacktrace if the given expression is false, and possibly call exit(-1) if the exit flag is set ( false by default )
## 

From: https://code.renci.org/gf/project/e-irods/tracker/?action=TrackerItemEdit&tracker_item_id=1278

PR #8130, opened 2025-01-24
[7942] Remove irods_assert.hpp (4-3-stable)

FifthPotato:
In the original issue, Markus recommended that we remove the declaration then deprecate the header, but I'm not sure what it means to deprecate a header-- so I deprecated the declaration. I think that's fine? If there's any better ways to do it, I'm open to suggestions as always.


alanking:
Is this the issue? https://github.com/irods/irods/issues/7942

I don't think we have an official procedure for deprecating a whole header file... There are some decent suggestions here: https://stackoverflow.com/questions/68306977/how-can-i-deprecate-a-c-header Initializing a `constexpr` variable with the `[[deprecated]]` attribute is almost too clever by half, but could do what we want.

Otherwise, your existing deprecation of the function seems fine. Please satisfy clang-format.

alanking:
Might be more proper to use a `pragma` for Clang: https://clang.llvm.org/docs/UsersManual.html#controlling-diagnostics-via-pragmas And a `#warning` directive for GCC: https://gcc.gnu.org/onlinedocs/cpp/Diagnostics.html Apparently, `#warning` is now part of the C++23/C23 standard.

Deferring to @korydraughn and @SwooshyCueb.

korydraughn:
If nothing relies on that file and it doesn't have a definition, remove the file.

It's a C++ source file, so I doubt anyone would be using it outside of us.

Confirm whether the server compiles w/o that file and passes the C++ unit test suite.

SwooshyCueb:
It's in a header, so there is the possibility of downstream projects `#include`-ing it. If we think this is unlikely, I'd be fine with removing it. Otherwise, I'd remove the declaration since there is no matching definition, and use a `#pragma GCC warning` in the header, as this is compatible with both GCC and clang.

trel:
initially... https://github.com/irods/irods/issues/1278

FifthPotato:
Built/tested it after removing it. Seems fine; unit tests pass.
Do we know what we want to end up doing with this?

korydraughn:
The fact that there's no definition for that function and no one has reported it as an issue indicates no one is using it.

Torch it.

korydraughn:
Squash it and add the issue number. No pound.

FifthPotato:
Squished, numbered

korydraughn (approved):
Pound it.

FifthPotato:
Octothorpe'd.

alanking (approved):

alanking:
I believe we need to cherry-pick these changes to main as we will want the same thing there.

PR #8137, opened 2025-01-29
[8120] Avoid segfault in rcDisconnect (main)

FifthPotato:
Running tests. Ideally, nothing calls this twice, but this should avoid segfault if it is done so by accident.

### DIFF
--- a/unit_tests/src/test_rcConnect.cpp
+++ b/unit_tests/src/test_rcConnect.cpp
@@ -2,6 +2,7 @@
 
 #include "irods/rcConnect.h"
 #include "irods/rodsErrorTable.h"
+#include "irods/irods_threads.hpp"
 
 #include <algorithm>
 #include <array>
@@ -40,5 +41,19 @@ TEST_CASE("set_session_signature_client_side")
 
         CHECK(set_session_signature_client_side(&comm, nullptr, buf.size()) == SYS_INVALID_INPUT_PARAM);
         CHECK(std::strncmp(comm.session_signature, "", sig_size) == 0);
+    }
+}
+
+TEST_CASE("rcDisconnect")
+{
+    SECTION("rcDisconnect does not segfault when called twice")
+    {
+        rcComm_t* otherconn = static_cast<rcComm_t*>(malloc(sizeof(rcComm_t)));
+        memset(otherconn, 0, sizeof(rcComm_t));
+        otherconn->thread_ctx = static_cast<thread_context*>(malloc(sizeof(thread_context)));
+        memset(otherconn->thread_ctx, 0, sizeof(thread_context));
+
+        rcDisconnect(otherconn);
+        rcDisconnect(otherconn);
     }
 }
--- a/lib/core/src/rcConnect.cpp
+++ b/lib/core/src/rcConnect.cpp
@@ -286,7 +286,7 @@ int rcDisconnect(
 
 
     _conn->exit_flg = true; //
-    if ( _conn->thread_ctx->reconnThr ) {
+    if (_conn->thread_ctx && _conn->thread_ctx->reconnThr) {
         try {
             // force an interruption point
             _conn->exit_flg = true;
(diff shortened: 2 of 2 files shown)


korydraughn:
Please add a C++ unit test for this.

FifthPotato:
Made some more changes after writing the test; got nailed by some double-frees. `cleanRcComm` used to `return 0;` in every code path, making its return value essentially useless. I changed it so it would `return 1` if any of the `rcComm_t` members are `free`'d, implying that the `rcComm_t` itself also needs to be `free`'d. This return value is read in `freeRcComm` to determine whether or not to free the `rcComm_t` ptr itself.
I also used a malloc'ed struct to simulate calling `rcConnect` in the test. Let me know if I've missed anything.

MartinFlores751:
Some UBSAN output that might be related:

```console
/irods_source/lib/core/src/rcConnect.cpp:268:22: runtime error: load of value 48761872, which is not a valid value for type 'irodsProt_t' (aka 'iRODSProtocol')
/irods_source/lib/core/src/rcConnect.cpp:293:29: runtime error: load of misaligned address 0x657320726f727265 for type 'boost::thread *', which requires 8 byte alignment
/irods_source/lib/core/src/rcConnect.cpp:293:29: runtime error: member access within misaligned address 0x657320726f727265 for type 'struct thread_context', which requires 8 byte alignment
```

korydraughn:
We can investigate that once this work and the UBsan work are merged.

FifthPotato:
> Changing what's returned by `cleanRcComm` changes the behavior of a public library function.
> 
> The change makes sense, but could it break any existing user code? Is that function called throughout the iRODS codebase? Would it be safer to add a better version of that function and replace use of the old one?

In an extreme case, this could break a user's code for which they read and care about the return value of `cleanRcComm`. (This would also require them to bypass `rcDisconnect` as well for whatever they're doing, because I couldn't find any references to this function outside of this file in any of our stuff).
However, since `cleanRcComm` has only ever returned 0, this would mean that caring about the return value would be making an assumption on what this return value means, which we've never really made clear.
Implicitly, we usually have a "status" where nonzero is error and zero is "OK," but `cleanRcComm` seems to not make use of that, given that it only ever returned 0. So for this to break someone's code, they would have to
1. Not use rcDisconnect and instead try to clean up the structs on their own
2. Read the return value of `cleanRcComm` and
3. assume that it is significant, even though it has only historically ever returned 0 (which means they would be "assumption-based future-proofing" in case we added an error case to it)

So, while I think the chance of this breaking someone's code is extremely small, I'm willing to go an alternative route if you think it's worth it; I'll leave it up to you.

korydraughn:
Excellent.

I cannot see any reason why a user of the library would want to call `cleanRcComm` themselves and bypass `rcDisconnect`.

And the fact that the library has never asserted anything about `cleanRcComm` (because there's no docs) makes me agree with you.

By giving meaning to the return value, we now have to document it. Please add doxygen documentation to that function explaining its usage and behavior.

Good use of the iRODS C API means using `rcDisconnect` at the appropriate time. No one should be calling `cleanRcComm` in client-side code.

The change in behavior is approved.

unit_tests/src/test_rcConnect.cpp
```diff
@@ -41,4 +42,19 @@ TEST_CASE("set_session_signature_client_side")
         CHECK(set_session_signature_client_side(&comm, nullptr, buf.size()) == SYS_INVALID_INPUT_PARAM);
         CHECK(std::strncmp(comm.session_signature, "", sig_size) == 0);
     }
+
+}
+
+TEST_CASE("rcDisconnect segfault")
```
korydraughn, line 48: Let's remove segfault.

This can cover all `rcDisconnect` test situations moving forward.

unit_tests/src/test_rcConnect.cpp
```diff
@@ -41,4 +42,19 @@ TEST_CASE("set_session_signature_client_side")
...
+{
+    SECTION("does not segfault when calling rcDisconnect twice")
```
korydraughn, line 50: Consider rephrasing to:
```
rcDisconnect does not segfault when called twice
```

korydraughn (commented):
Have you seen your test fail in the same way without the fix?

FifthPotato:
> Have you seen your test fail in the same way without the fix?

By "the fix," do you mean the first commit (`thread_ctx` not checked for `NULL`) or the second one (double-frees I found because of the unit test)?

korydraughn:
I was referring to the double-frees, but now I'm wondering if you've reproduced both of them.

FifthPotato:
> I was referring to the double-frees, but now I'm wondering if you've reproduced both of them.

The test failed on tip-of-main, but I had assumed that was because of the `thread_ctx` thing (because of the debugging I did on the `iqmod` segfault). I didn't specifically dig into that test failure. The second failure was because of the double-free, which I then resolved.

korydraughn (commented):
Looking at the code, I think we're good.

If the tests are passing, squash to taste. No pounds.

korydraughn:
I see all of the review comments have been addressed.

What's the status of this? Are we running tests?

FifthPotato:
> I see all of the review comments have been addressed.
> 
> What's the status of this? Are we running tests?

Yep, running core on it.

FifthPotato:
Core and unit pass. Squished.

alanking:
Oh, and clang-format, please

FifthPotato:
Updated with suggested changes and clang-format.

korydraughn (approved):
If the tests are passing and all review comments have been resolved, pound it.

alanking (approved):

FifthPotato:
Octothorpe'd.

alanking:

Issue #8090, opened 2024-12-13
Add option to enable Undefined Behavior Sanitizer


MartinFlores751:
- [x] main (irods)
- [x] main (iCommands)
- [x] 4-3-stable (irods)
- [x] 4-3-stable (iCommands)
---

See for UBSAN documentation at: irods/irods_client_http_api#390.

If an ignore list is needed, see: irods/irods_client_http_api#392.

korydraughn:
Related to #6613.

alanking:
@MartinFlores751 - Please checkbox and close if complete. Thanks

PR #8138, opened 2025-01-29
[#8090] Add Undefined Behavior Sanitizer to Build Options (4-3-stable)

MartinFlores751:
Still need to improve the ignore list (e.g. hashing functions have expected overflow behavior.)

Found some issues, planning to address easy fixes in this PR too.

I do have a question. Based on the current configuration, files like ssl.cpp, sockComm.cpp, and packStruct.cpp produce a lot of errors on implicit conversions. These conversions are not undefined behavior, but a reminder to check "is this is what you really want".

Do we want to make these conversions explicit? If we don't, do we make an ignore for these?

### DIFF
--- a/lib/core/src/msParam.cpp
+++ b/lib/core/src/msParam.cpp
@@ -1189,9 +1189,18 @@ namespace
 
         const auto& bbuf = _func(*execCmdOut);
 
-        *_out = static_cast<char*>(std::malloc(bbuf.len + 1));
-        std::memcpy(*_out, static_cast<char*>(bbuf.buf), bbuf.len);
-        (*_out)[bbuf.len] = '\0';
+        const auto out_len{bbuf.len + 1};
+        // NOLINTNEXTLINE(cppcoreguidelines-owning-memory,cppcoreguidelines-no-malloc)
+        *_out = static_cast<char*>(std::malloc(out_len));
+
+        if (nullptr != bbuf.buf) {
+            std::memcpy(*_out, static_cast<char*>(bbuf.buf), bbuf.len);
+            // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic)
+            (*_out)[bbuf.len] = '\0';
+        }
+        else {
+            std::memset(*_out, 0, out_len);
+        }
 
         return 0;
     }
--- a/lib/filesystem/src/path.cpp
+++ b/lib/filesystem/src/path.cpp
@@ -368,7 +368,7 @@ namespace irods::experimental::filesystem
             return *this;
         }
 
-        if (is_separator(fp[pos_ - 1])) {
+        if (pos_ > 0 && is_separator(fp[pos_ - 1])) {
             // We've reached the root separator of the path.
             if (0 == pos_ - 1)
             {
--- a/server/re/src/printMS.cpp
+++ b/server/re/src/printMS.cpp
@@ -171,10 +171,18 @@ int writeBytesBuf( msParam_t* where, msParam_t* inBuf, ruleExecInfo_t *rei ) {
     }
 
     if ( inBuf->inpOutBuf ) {
+        const auto writeStr_len{inBuf->inpOutBuf->len + 1};
         /* Buffer might no be null-terminated */
-        writeStr = ( char* )malloc( inBuf->inpOutBuf->len + 1 );
-        strncpy( writeStr, ( char* )inBuf->inpOutBuf->buf, inBuf->inpOutBuf->len );
-        writeStr[inBuf->inpOutBuf->len] = '\0';
+        // NOLINTNEXTLINE(cppcoreguidelines-owning-memory,cppcoreguidelines-no-malloc)
+        writeStr = static_cast<char*>(malloc(writeStr_len));
+        if (nullptr != inBuf->inpOutBuf->buf) {
+            strncpy(writeStr, static_cast<char*>(inBuf->inpOutBuf->buf), inBuf->inpOutBuf->len);
+            // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic)
+            writeStr[inBuf->inpOutBuf->len] = '\0';
+        }
+        else {
+            std::memset(writeStr, 0, writeStr_len);
+        }
     }
     else {
         writeStr = strdup( inBuf->label );
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,6 +8,8 @@ include(RequireOutOfSourceBuild)
 
 option(IRODS_DISABLE_COMPILER_OPTIMIZATIONS "Disables compiler optimizations by setting -O0." OFF)
 option(IRODS_ENABLE_ADDRESS_SANITIZER "Enables detection of memory leaks and other features provided by Address Sanitizer." OFF)
+option(IRODS_ENABLE_UNDEFINED_BEHAVIOR_SANITIZER "Enables detection of undefined behavior provided by Undefined Behavior Sanitizer." OFF)
+option(IRODS_ENABLE_UNDEFINED_BEHAVIOR_SANITIZER_IMPLICIT_CONVERSION_CHECK "Enables detection of implicit conversions by Undefined Behavior Sanitizer." OFF)
 option(IRODS_ENABLE_ALL_TESTS "Enables all tests (e.g. C/C++ unit tests, microservices specifically designed for testing)." OFF)
 
 if (NOT CMAKE_CONFIGURATION_TYPES AND NOT CMAKE_BUILD_TYPE)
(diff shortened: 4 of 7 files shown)


MartinFlores751:
There can be ignores for specific functions, but there are some limitations (must use mangled names).

See: https://releases.llvm.org/13.0.1/tools/clang/docs/SanitizerSpecialCaseList.html

For clang-tidy reporting, I didn't see it reporting the same issues that UBSAN is reporting (e.g. use of `htonl(...)` in packStruct).

Maybe making the check off by default is enough, or providing a script to filter out those reports?

korydraughn:
Let's not filter the output. Exposing an option to turn on implicit conversion checks is good enough for now.

MartinFlores751:
> There can be ignores for specific functions, but there are some limitations (must use mangled names).
> 
> See: https://releases.llvm.org/13.0.1/tools/clang/docs/SanitizerSpecialCaseList.html
> 
> For clang-tidy reporting, I didn't see it reporting the same issues that UBSAN is reporting (e.g. use of `htonl(...)` in packStruct).
> 
> Maybe making the check off by default is enough, or providing a script to filter out those reports?

Should probably note, I think the ignorelist applies to both ASAN and UBSAN. The current ignorelist disables both for anything in `/opt/irods-externals/*`.

MartinFlores751:
> Looking at the impl, it looks like it only applies to UBsan. The ignorelist appears to be set only when UBsan is enabled.
> 
> Am I missing a step in the current logic?

If both are enabled, then the ignorelist applies to both.

> Also, I believe the rule is to never compile with multiple sanitizers enabled.

Is that a rule? I thought ASAN and UBSAN could be used at the same time. If that is as rule, we should enforce that in CMake if possible.

korydraughn:
I read that somewhere in the clang compiler docs when I was learning how to use Asan in iRODS. Add investigating that to your TODO list. Can even open an issue for it so it's not forgotten.

With that said, it's not urgent. Let's get this PR wrapped up so we can focus on other issues.

MartinFlores751:
> I read that somewhere in the clang compiler docs when I was learning how to use Asan in iRODS. Add investigating that to your TODO list. Can even open an issue for it so it's not forgotten.
> 
> With that said, it's not urgent. Let's get this PR wrapped up so we can focus on other issues.

Issue created, we'll deal with it later.

lib/filesystem/src/path.cpp
```diff
@@ -368,7 +368,7 @@ namespace irods::experimental::filesystem
             return *this;
         }
 
-        if (is_separator(fp[pos_ - 1])) {
+        if (pos_ != 0 && is_separator(fp[pos_ - 1])) {
```
korydraughn, line 371: `> 0` instead of `!= 0`?

korydraughn: Hmm, I guess `!=` is fine too since the impl is supposed to guard against range errors.

korydraughn:
What's left for this PR?

lib/filesystem/src/path.cpp
korydraughn: Definitely need to run the tests to make sure this change doesn't break anything.

MartinFlores751:
Nothing else if we're fine with our current ignorelist. It would just report some unsigned overflows from hash functions.

Will take out of draft.

korydraughn:
Ok. Please run the test suites and let us know how it goes.

lib/core/src/msParam.cpp
```diff
@@ -1190,7 +1190,10 @@ namespace
         const auto& bbuf = _func(*execCmdOut);
 
         *_out = static_cast<char*>(std::malloc(bbuf.len + 1));
-        std::memcpy(*_out, static_cast<char*>(bbuf.buf), bbuf.len);
+
+        if (nullptr != bbuf.buf) {
+            std::memcpy(*_out, static_cast<char*>(bbuf.buf), bbuf.len);
+        }
         (*_out)[bbuf.len] = '\0';
```
alanking, line 1197: Would it make sense to zero out the structure, or put the `\0` in `(*_out)[0]` if `bbuf.buf == nullptr` (i.e. the `else` case)? As it stands, there could be some garbage string in there since we just `malloc` above.

lib/filesystem/src/path.cpp
korydraughn: Good to know. Can't remember if the check for 0 was intentionally left out. Either it's an oversight or it was determined unnecessary by design.

server/re/src/printMS.cpp
```diff
@@ -173,7 +173,9 @@ int writeBytesBuf( msParam_t* where, msParam_t* inBuf, ruleExecInfo_t *rei ) {
     if ( inBuf->inpOutBuf ) {
         /* Buffer might no be null-terminated */
         writeStr = ( char* )malloc( inBuf->inpOutBuf->len + 1 );
-        strncpy( writeStr, ( char* )inBuf->inpOutBuf->buf, inBuf->inpOutBuf->len );
+        if (nullptr != inBuf->inpOutBuf->buf) {
+            strncpy(writeStr, (char*) inBuf->inpOutBuf->buf, inBuf->inpOutBuf->len);
+        }
         writeStr[inBuf->inpOutBuf->len] = '\0';
```
alanking, line 179: Same thing here - wondering whether we should put `\0` in `writeStr[0]` in the `else` case to avoid a garbage string.

alanking (commented):
Same question in two places. Not sure if valid

lib/filesystem/src/path.cpp
MartinFlores751: Made this change following the UBSAN report if it helps:

```
...
/irods_source/lib/filesystem/src/path.cpp:371:34: runtime error: unsigned integer overflow: 0 - 1 cannot be represented in type 'unsigned long'
...
```

lib/core/src/msParam.cpp
MartinFlores751: I was assuming if `bbuf.buf` is `nullptr`, then `bbuf.len` is set to zero, but it's better to have the buffer zeroed out. I'll do that.

MartinFlores751: Done.

server/re/src/printMS.cpp
MartinFlores751: Done.

lib/filesystem/src/path.cpp
MartinFlores751: Updated to `> 0`.

MartinFlores751:
Core and Unit Tests passed before the last two commits. Rerunning to see if they caused any changes.

server/re/src/printMS.cpp
```diff
@@ -171,12 +171,16 @@ int writeBytesBuf( msParam_t* where, msParam_t* inBuf, ruleExecInfo_t *rei ) {
     }
 
     if ( inBuf->inpOutBuf ) {
+        const auto writeStr_len{inBuf->inpOutBuf->len + 1};
         /* Buffer might no be null-terminated */
-        writeStr = ( char* )malloc( inBuf->inpOutBuf->len + 1 );
+        writeStr = (char*) malloc(writeStr_len);
         if (nullptr != inBuf->inpOutBuf->buf) {
             strncpy(writeStr, (char*) inBuf->inpOutBuf->buf, inBuf->inpOutBuf->len);
```
alanking: Let's go ahead and give these the `static_cast` treatment while we're in here.
```suggestion
        writeStr = static_cast<char*>(malloc(writeStr_len));
        if (nullptr != inBuf->inpOutBuf->buf) {
            strncpy(writeStr, static_cast<char*>(inBuf->inpOutBuf->buf), inBuf->inpOutBuf->len);
```

alanking (commented):
One more suggestion. I don't think that change requires re-running tests, so don't interrupt your current run(s).

server/re/src/printMS.cpp
MartinFlores751: Sounds good

MartinFlores751: Done.

korydraughn:
Looks like we're just waiting for test results now?

MartinFlores751:
Yep, core is still running with unit tests queued up.

MartinFlores751:
All passed!

alanking (commented):
Squash em

MartinFlores751:
I'm thinking of keeping the impl and the fixes separate. Should the fixes be a new issue? Something like "Fix UBSAN issues"?

alanking:
> I'm thinking of keeping the impl and the fixes separate. Should the fixes be a new issue? Something like "Fix UBSAN issues"?

Sounds fine to me. Do the fixes here represent all of the issues found by UBSAN? If not, maybe mention the issues being addressed specifically in the issue and perhaps open yet another issue to address the outstanding issues (just so we don't lose them).

MartinFlores751:
> > I'm thinking of keeping the impl and the fixes separate. Should the fixes be a new issue? Something like "Fix UBSAN issues"?
> 
> Sounds fine to me. Do the fixes here represent all of the issues found by UBSAN?

This only fixes a minor "quick fix" subset of the issues. Some of them are much more involved.

> If not, maybe mention the issues being addressed specifically in the issue and perhaps open yet another issue to address the outstanding issues (just so we don't lose them).

I'll create the separate issue rn, but only mention to run the tests to generate the UB list. It's better to generate the list at the time of fixing so you don't chase bugs that may have already been fixed.

alanking:
That makes sense.

MartinFlores751:
Created, see #8153.

alanking:
Very good. Please update commit message with the new issue number.

MartinFlores751:
Updated commit.

korydraughn:
Don't forget to clang-format.

MartinFlores751:
> Don't forget to clang-format.

What's reported in actions is a job failure. Shows up clean on my side.

See:
```console
W: Failed to fetch https://unstable.irods.org/apt/dists/noble/InRelease  Could not connect to unstable.irods.org:443 (152.54.5.113), connection timed out
W: Some index files failed to download. They have been ignored, or old ones used instead.
E: Unable to locate package irods-externals-clang13.0.1-0
E: Couldn't find any package by glob 'irods-externals-clang13.0.1-0'
E: Couldn't find any package by regex 'irods-externals-clang13.0.1-0'
Error: Process completed with exit code 100.
```

korydraughn:
Relaunched the clang-format action!

korydraughn:
Anything left to do for this?

MartinFlores751:
Nothing left for this issue. Any further work already has its own separate issue.

korydraughn (approved):
Pound it and cherry-pick to main.

MartinFlores751:
Updated message on UBSAN commit.

alanking (approved):
\# it

MartinFlores751:
Pounded.

korydraughn:
@MartinFlores751 Please cherry-pick to main.

Issue #8106, opened 2024-12-24
irodsServer does not free the memory when irsync transfer a large number files


cardenas-in2p3:
- [x] main (irods)
- [x] main (icommands)
- [x] 4-3-stable (irods)
- [x] 4-3-stable (icommands)

---

## Bug Report
irodsServer does not free the memory when irsync transfer a large number of small files then the irodsServer is killed by oom-killer

### iRODS Version, OS and Version

server: providers and consumers iRODS_4.3.2/CentOS 7.9
client: icommands on WSL/MS Windows 10 

### What did you try to do?
transfer a large number of small files with  irsync  

### Expected behavior

synchronization (transfer) of data from the client's local file system to iRODS is unsuccessful/incomplete 

### Observed behavior (including steps to reproduce, if applicable)

client command on MS Windows with WSL: 
  % date 
  Thu Dec 19 12:51:25 CET 2024
  % irsync -N 0 -r -V sourceDirectory i:targetDirectory
  
sourceDirectory
  size 117GiB
  number of files:64414  (averege size 1.85 MiB)

server side behavior (process 1067) increase memory usage during transfer time 

provider% date  
Thu Dec 19 15:07:32 CET 2024

provider% ips  -z zone -a 

Server: provider.in2p3.fr
 […] truncated lines 
     1067 rods#zone  22:03:14  irsync  _IP_ADDR_ 
 […] truncated lines    

provider% top -p 1067

  PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND
  
 1067 irodsmgr  20   0   26.2g  25.7g  36052 S  33.0 41.2  24:17.37 irodsServer

 1067 irodsmgr  20   0   32.8g  32.3g  36052 S  35.2 51.7  28:18.59 irodsServer 
 
 1067 irodsmgr  20   0   37.4g  36.9g  36052 S   6.7 59.2  31:08.67 irodsServer
 
 1067 irodsmgr  20   0   42.9g  42.4g  36052 S  34.3 67.9  34:29.87 irodsServer
                                                                                        
 1067 irodsmgr  20   0   45.9g  45.4g  36052 R  44.0 72.7  36:24.72 irodsServer
 
 1067 irodsmgr  20   0   49.6g  49.1g  36052 R  33.7 78.7  38:42.59 irodsServer
 […]
 
provider% free -g 
              total        used        free      shared  buff/cache   available
Mem:             62          60           0           0           1           0
[...]

PR #8141, opened 2025-01-29
[#8106] Fix memory leaks found in irsync tests (4-3-stable)

MartinFlores751:
This PR has a companion fix in icommands. The icommands fix are optional, as in there is no real growing memory leak, but it does address ASAN comments.

### DIFF
--- a/lib/core/src/irods_parse_command_line_options.cpp
+++ b/lib/core/src/irods_parse_command_line_options.cpp
@@ -277,7 +277,7 @@ static int build_irods_path_structure(const path_list_t& _path_list,
         return USER__NULL_INPUT_ERR;
     }
 
-    memset( _rods_paths, 0, sizeof( rodsPathInp_t ) );
+    freeRodsPathInpMembers(_rods_paths);
 
     if ( _path_list.size() <= 0 ) {
         if ( ( _flag & ALLOW_NO_SRC_FLAG ) == 0 ) {
--- a/lib/core/src/miscUtil.cpp
+++ b/lib/core/src/miscUtil.cpp
@@ -199,16 +199,17 @@ rmdirR( char *startDir, char *destDir ) {
 int
 getRodsObjType( rcComm_t *conn, rodsPath_t *rodsPath ) {
     int status;
-    dataObjInp_t dataObjInp;
-    rodsObjStat_t *rodsObjStatOut = NULL;
 
     if ( rodsPath == NULL ) {
         return USER__NULL_INPUT_ERR;
     }
 
-    memset( &dataObjInp, 0, sizeof( dataObjInp ) );
+    dataObjInp_t dataObjInp{};
+    irods::at_scope_exit clearData{[&dataObjInp] { clearDataObjInp(&dataObjInp); }};
 
     rstrcpy( dataObjInp.objPath, rodsPath->outPath, MAX_NAME_LEN );
+
+    rodsObjStat_t* rodsObjStatOut{};
     status = rcObjStat( conn, &dataObjInp, &rodsObjStatOut );
 
     if ( status < 0 ) {
@@ -248,6 +249,8 @@ getRodsObjType( rcComm_t *conn, rodsPath_t *rodsPath ) {
             rstrcpy( rodsPath->chksum, rodsObjStatOut->chksum, NAME_LEN );
         }
     }
+    // NOLINTNEXTLINE(cppcoreguidelines-no-malloc,cppcoreguidelines-owning-memory)
+    std::free(rodsPath->rodsObjStat);
     rodsPath->rodsObjStat = rodsObjStatOut;
 
     return rodsPath->objState;
--- a/lib/core/src/msParam.cpp
+++ b/lib/core/src/msParam.cpp
@@ -570,6 +570,14 @@ int clearMsParam(msParam_t* msParam, int freeInOutStruct)
                 clearExecCmdOut(msParam->inOutStruct);
                 std::free(msParam->inOutStruct);
             }
+            // Do not free KeyValPair_MS_T unconditionally as this will lead
+            // to a use-after-free error. Additionally, this behavior matches the default
+            // of checking freeInOutStruct before freeing the data.
+            else if (std::strcmp(msParam->type, KeyValPair_MS_T) == 0 && freeInOutStruct > 0) {
+                clearKeyVal(static_cast<keyValPair_t*>(msParam->inOutStruct));
+                // NOLINTNEXTLINE(cppcoreguidelines-no-malloc,cppcoreguidelines-owning-memory)
+                std::free(msParam->inOutStruct);
+            }
             // This else-block must always be the final block in this if-ladder.
             // Changing the order of these if-blocks can result in memory leaks.
             else if (freeInOutStruct > 0) {
--- a/lib/core/src/rodsPath.cpp
+++ b/lib/core/src/rodsPath.cpp
@@ -555,6 +555,30 @@ clearRodsPath( rodsPath_t *rodsPath ) {
     return;
 }
 
+auto freeRodsPathInpMembers(rodsPathInp_t* path) -> void
+{
+    if (nullptr == path) {
+        return;
+    }
+
+    for (int i{}; i < path->numSrc; i++) {
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic)
+        clearRodsPath(&path->srcPath[i]);
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic)
+        clearRodsPath(&path->targPath[i]);
+    }
+    clearRodsPath(path->destPath);
+
+    // NOLINTNEXTLINE(cppcoreguidelines-no-malloc,cppcoreguidelines-owning-memory)
+    std::free(path->srcPath);
+    // NOLINTNEXTLINE(cppcoreguidelines-no-malloc,cppcoreguidelines-owning-memory)
+    std::free(path->destPath);
+    // NOLINTNEXTLINE(cppcoreguidelines-no-malloc,cppcoreguidelines-owning-memory)
+    std::free(path->targPath);
+
+    std::memset(path, 0, sizeof(rodsPathInp_t));
+} // freeRodsPathInpMembers
+
 char* escape_path(const char* _path)
 {
     static const std::map<char, std::string> special_char_mappings{
--- a/lib/core/include/irods/rodsPath.h
+++ b/lib/core/include/irods/rodsPath.h
@@ -69,6 +69,7 @@ int
 getFileType( rodsPath_t *rodsPath );
 void
 clearRodsPath( rodsPath_t *rodsPath );
+void freeRodsPathInpMembers(rodsPathInp_t* path);
 
 // Returns a new path with the following special characters escaped:
 //   - '\f'
--- a/lib/core/src/chksumUtil.cpp
+++ b/lib/core/src/chksumUtil.cpp
@@ -5,6 +5,7 @@
 #include "irods/rodsLog.h"
 #include "irods/chksumUtil.h"
 #include "irods/rcGlobalExtern.h"
+#include "irods/irods_at_scope_exit.hpp"
 
 #include <sys/time.h>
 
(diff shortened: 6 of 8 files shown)


lib/core/src/irods_parse_command_line_options.cpp
```diff
@@ -277,7 +277,7 @@ static int build_irods_path_structure(const path_list_t& _path_list,
         return USER__NULL_INPUT_ERR;
     }
 
-    memset( _rods_paths, 0, sizeof( rodsPathInp_t ) );
+    clearRodsPathInp(_rods_paths);
```
korydraughn, line 280: Does the clear operation reset/memset the input for reuse?

lib/core/src/miscUtil.cpp
```diff
@@ -248,6 +249,7 @@ getRodsObjType( rcComm_t *conn, rodsPath_t *rodsPath ) {
             rstrcpy( rodsPath->chksum, rodsObjStatOut->chksum, NAME_LEN );
         }
     }
+    std::free(rodsPath->rodsObjStat);
```
korydraughn, line 252: I take it `rodsPath->rodsObjStat` is always valid at this point?

Should it be free'd on lines 221 and 226 too?

lib/core/src/rodsPath.cpp
```diff
@@ -555,6 +555,39 @@ clearRodsPath( rodsPath_t *rodsPath ) {
...
 }
 
+auto clearRodsPathInp( rodsPathInp_t* path ) -> void {
+  if (nullptr == path) {
+    return;
+  }
+
+  for (int i{}; i < path->numSrc; i++) {
+    clearRodsPath(&path->srcPath[i]);
+    clearRodsPath(&path->targPath[i]);
```
korydraughn, line 565: So `targPath` is an array?

korydraughn (commented):
Good stuff.

Please run the unit/core tests.

lib/core/src/rodsPath.cpp
```diff
@@ -555,6 +555,39 @@ clearRodsPath( rodsPath_t *rodsPath ) {
...
+  }
+  clearRodsPath(path->destPath);
+
+  memset(path, 0, sizeof(rodsPathInp_t));
+}
```
alanking, line 570:
```suggestion
} // clearRodsPathInp
```

lib/core/src/rodsPath.cpp
```diff
@@ -555,6 +555,39 @@ clearRodsPath( rodsPath_t *rodsPath ) {
...
+  }
+  clearRodsPath(path->destPath);
+
+  free(path->srcPath);
+  path->srcPath = nullptr;
+  free(path->destPath);
+  path->destPath = nullptr;
+  free(path->targPath);
+  path->targPath = nullptr;
+}
```
alanking, line 589:
```suggestion
} // freeRodsPathInpMembers
```

alanking, line 589: Should we `memset` the structure at the end here as is being done in `clearRodsPathInp`?

Alternatively, should `freeRodsPathInpMembers` and `clearRodsPathInp` be combined with an option to free the internal members? Perhaps the separate functions you have now are cleaner...

alanking (commented):
Exciting stuff. Please satisfy clang-format

lib/core/src/irods_parse_command_line_options.cpp
MartinFlores751: `clearRodsPathInp(...)` does memset the input. I based it off of the other `clear...` functions.

lib/core/src/miscUtil.cpp
MartinFlores751: Since the next line replaces `rodsPath->rodsObjStat`, we need to free the value held there. The other returns do not override `rodsPath->rodsObjStat`.

However, `rodsObjStatOut` _might_ leak in those cases. If `rcObjStat(...)` still creates a `rodsObjStat_t` when it returns a `status < 0`, then it may leak.

lib/core/src/msParam.cpp
```diff
@@ -570,6 +570,10 @@ int clearMsParam(msParam_t* msParam, int freeInOutStruct)
                 clearExecCmdOut(msParam->inOutStruct);
                 std::free(msParam->inOutStruct);
             }
+	    else if (std::strcmp(msParam->type, KeyValPair_MS_T) == 0 && freeInOutStruct > 0) {
```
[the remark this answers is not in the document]
MartinFlores751: I don't really understand it. If you look a bit higher up, none of the other functions check `freeInOutStruct`, and always free if the `std::strcmp(...) == 0`.

My best guess is that `freeInOutStruct` is a catch-all generic free, which is only executed if specified, as seen in the code following the added `else if` block.

As for why use it in this case, I believe unconditionally freeing the KVP leads to a use-after-free, if I recall.

lib/core/src/rodsPath.cpp
MartinFlores751: If I understand the struct correctly, every `srcPath` has a paired `targPath`, which is probably based off the `basePath`. The last part is a guess, but I did see `srcPath` and `targPath` paired together, so `numSrc` should represent the length for both.

lib/core/src/irods_parse_command_line_options.cpp
korydraughn: Got it.

lib/core/src/rodsPath.cpp
MartinFlores751: I think I made the clear function first based off the existing clear functions, followed by the free function if that helps explain anything.

I think we could `memset(...)` it, or at the very least set the `numSrc` to zero in here too.

If we do memset in the free function, maybe it would be worth combining them and adding a `free_members` arg?

alanking: Ah, it's not really *supposed* to be the same as `clearRodsPathInp`. Well, I still think it's helpful to go ahead and clear things while we're freeing. As for combining the functions, I defer. Keeping them separated is fine; combining is fine.

lib/core/src/miscUtil.cpp
MartinFlores751: > Should we check the pointer before freeing it?

The only check I can think of is to check if `rodsPath->rodsObjStat == rodsObjStatOut`? I have no idea if `rcObjStat(...)` would return it's input like that...

Otherwise, if it's a `nullptr`, there is no need to check as free is safe with `nullptr`. If there's a value that needs to be freed, it will be freed. If it's a pointer to an already freed structure, there's no check for that, and it's UB (catchable by ASAN).

lib/core/src/msParam.cpp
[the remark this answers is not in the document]
MartinFlores751: Unconditionally freeing the KVP leads to a use-after-free, if I recall. Having the `freeInOutStruct` check avoids the unconditional free.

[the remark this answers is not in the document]
korydraughn: Ok. Add that as a comment. Then no one will have to relearn that.

lib/core/src/miscUtil.cpp
korydraughn: Got it.

lib/core/src/rodsPath.cpp
MartinFlores751, line 570: I think this function is incorrect. We lose the data pointed by `srcPath`, `targPath`, and `destPath`.

The only place this function is used should call `freeRodsPathInpMembers` instead.

MartinFlores751 (commented):
Think I found something.

lib/core/src/rodsPath.cpp
MartinFlores751: This might also explain why there are smaller leaks still. Will replace and check.

MartinFlores751:
> Can't remove the clear function in 4-3-stable since it's part of the public API.
> 
> Mark it as deprecated and remove it in the main (5.0) branch.

I added that function in my first commit for this PR: https://github.com/irods/irods/pull/8141/commits/c151c0a76d6489c9c81c032021aecbef57d21c76

I think a revert commit would have made that clearer. I'll do that next time.

alanking:
> I think a revert commit would have made that clearer. I'll do that next time.

Perhaps for purposes of code review, sure. For the final set of commits... probably just squash it out and pretend like it didn't happen. Just to be clear. :) Agree?

Also, don't forget about clang-format and resolving conversations if they have been addressed. Looks like we're rounding the bend on this one.

korydraughn:
Either approach is fine with me as long as it's clear or communicated via an appropriate comment.

FifthPotato (commented):
I've stared at it really hard and I can't find much wrong with it. Going to say it looks good to me!

lib/core/src/rodsPath.cpp
MartinFlores751: Done.

lib/core/src/rodsPath.cpp
MartinFlores751: Removed function.

lib/core/src/rodsPath.cpp
MartinFlores751: `memset(...)` is now at the end of `freeRodsPathInpMembers(...)`. One function removed, so merging them is not a concern anymore.

lib/core/src/rodsPath.cpp
MartinFlores751: This does not fix the persistent few leaks, but it does allow us to only have one new function (rather than two).

Will not be including the `clearRodsPathInp(...)` function.

lib/core/src/msParam.cpp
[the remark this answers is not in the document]
MartinFlores751: Comment added.

lib/core/src/msParam.cpp
```diff
@@ -570,6 +570,13 @@ int clearMsParam(msParam_t* msParam, int freeInOutStruct)
                 clearExecCmdOut(msParam->inOutStruct);
                 std::free(msParam->inOutStruct);
             }
+            // Do not free KeyValPair_MS_T unconditionally as this will lead
+            // to a use-after-free. Additionally, this behavior matches the default
```
korydraughn, line 574: Feels like we're missing a word after `... to a use-after-free.`.

MartinFlores751:
Running the core and unit tests.

lib/core/src/msParam.cpp
MartinFlores751: Jokingly: `... this well lead to a contretemps; a use-after-free error.`

I feel like `use-after-free` is 'noun enough' to end on, but perhaps ending in `error` or `memory error` may help?

korydraughn: Appending `error` is good enough.

Think about future readers of the code. Try not to leave room for confusion.

MartinFlores751: `error` added.

alanking (commented):
Looks good to me. Please squash to taste and adjust commit message with the issue number (no #).

MartinFlores751:
Rebased onto the latest 4-3-stable commit.

A bug was introduced that causes the tests to fail quite often. Currently checking to see if the removal of the clear function is what caused it.

MartinFlores751:
Update, will need to do some more looking into this PR to see what is causing the issue.

For the time, I'll push other PRs through, just so testing/debugging of this PR doesn't block the others.

MartinFlores751:
Bisecting to find the problem commit.

korydraughn:
Given https://github.com/irods/irods_client_icommands/pull/519#issuecomment-2664068013, what's the status of this PR?

Has it been tested without the icommands PR?

MartinFlores751:
This PR has been tested without the icommands PR, and the core and unit tests passed.

korydraughn:
Good. Please squash to taste if this PR resolves the target issue.

MartinFlores751:
Squashing without the latest two commits.

MartinFlores751:
Mostly squashed. Waiting for okay on the last two commits.

korydraughn (approved):
Commits look good.

Squash and pound it.

MartinFlores751:
Squashed and pounded.

alanking (approved):

Issue #8133, opened 2025-01-26
GenQuery1 column macro defined twice


korydraughn:
- [x] main
- [x] 4-3-stable
---
## Bug Report

Please remove one of the following:

https://github.com/irods/irods/blob/0f13ac4c15602758b7b676352abd16b6c183c2f1/lib/core/include/irods/rodsGenQuery.h#L364-L365

trel:
almost 16y!

https://github.com/irods/irods-legacy/commit/ad769c70acc186f5d8b3454311283c0b15189275

PR #8169, opened 2025-02-05
[#8133] Remove duplicate macro definition for GenQuery1 (4-3-stable)

korydraughn:
This doesn't require testing since the macro is already defined once. The removal of the second definition does not change existing behavior.


korydraughn:
Pounded.

Issue #8111, opened 2025-01-08
`ruleEngineConfig` should not be `memset(...)`


MartinFlores751:
https://github.com/irods/irods/blob/89d8b33487d6f122ec6a3f5dc1e48479a82b51ea/plugins/rule_engines/irods_rule_language/src/configuration.cpp#L71-L81

`Cache` is not a trivial class, and should not be `memset(...)`.

Turn into a POD or swap with a default constructed class in `clearRuleEngineConfig()`.

alanking:
@MartinFlores751 - Please checkbox and close if complete. Thanks

PR #8173, opened 2025-02-06
[#8111] Fix: Remove constructor from `Cache` in NREP to allow `Cache` to be memset (4-3-stable)

MartinFlores751:
Previously, `Cache` was not a trivial class, meaning that it should not be `memset(...)`.

This PR removes the constructor, allowing the structure to be `memset(...)`.

### DIFF
--- a/plugins/rule_engines/irods_rule_language/src/configuration.cpp
+++ b/plugins/rule_engines/irods_rule_language/src/configuration.cpp
@@ -33,42 +33,10 @@
 using log_re = irods::experimental::log::rule_engine;
 constexpr auto err_buf_len = ERR_MSG_LEN * 1024;
 
-Cache::Cache() : address(NULL), /* unsigned char *address */
-    pointers(NULL), /* unsigned char *pointers */
-    dataSize(0), /* size_t dataSize */
-    cacheSize(0), /* size_t cacheSize */
-    coreRuleSetStatus(UNINITIALIZED), /* RuleEngineStatus coreRuleSetStatus */
-    appRuleSetStatus(UNINITIALIZED), /* RuleEngineStatus appRuleSetStatus */
-    extRuleSetStatus(UNINITIALIZED), /* RuleEngineStatus extRuleSetStatus */
-    sysFuncDescIndexStatus(UNINITIALIZED), /* RuleEngineStatus sysFuncDescIndexStatus */
-    coreFuncDescIndexStatus(UNINITIALIZED), /* RuleEngineStatus coreFuncDescIndexStatus */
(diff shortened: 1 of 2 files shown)


MartinFlores751:
Need to run tests. Leaving in draft until tests pass.

MartinFlores751:
Running tests.

MartinFlores751:
Core and Unit tests passed.

plugins/rule_engines/irods_rule_language/src/configuration.cpp
```diff
@@ -33,42 +33,7 @@
...
-    extRegion(NULL), /* Region *extRegion */
-    tvarNumber(0), /* int tvarNumber */
-    clearDelayed(0), /* int clearDelayed */
-    timestamp(time_type_initializer), /* time_type timestamp */
-    logging(0), /* int logging */
-    ruleBase(""), /* char ruleBase[RULE_SET_DEF_LENGTH] */
-    hash("") /* char *hash */{}
-
-Cache ruleEngineConfig;
+Cache ruleEngineConfig{};
```
alanking, line 36: Consider putting a `NOLINT` here to satisfy clang-tidy. Bonus points if you leave an explanatory comment :) OR, don't, so it's not hidden from view. Up to you.

alanking (commented):
Very nice. Just had one minor thing so we get all the green checkmarks, but it's optional.

plugins/rule_engines/irods_rule_language/src/configuration.cpp
MartinFlores751: Hmm, is Cache a `const` candidate? Can you `memset(...)` a const struct (or is that confusing/bad practice)? I forget...

If you can, we should probably just do that now.
If you can't, we can add the ignore.

alanking: I see many places where `ruleEngineConfig` is being updated, so I don't think this should be `const`.

MartinFlores751: I'll add an ignore and a comment shortly then.

MartinFlores751: Done.

alanking (commented):
Seems good to me. Squash em

MartinFlores751:
Squashed.

korydraughn (approved):
Pound it.

korydraughn:
You can remove the `fix:` prefix.

MartinFlores751:
Pounded and removed 'fix:' from the message.

alanking (approved):

alanking:
Please cherry-pick to main. Thanks

Issue #7722, opened 2024-04-26
Inconsistent behavior between server and iadmin if zone name contains special chars


ll4strw:
- [x] main
- [x] 4-3-stable
---
## Bug Report

OS: Ubuntu 22.04.3 LTS
iRODS: v4.3.1

This issue was first mentioned in https://github.com/irods/irods_api_plugin_genquery2/issues/97#issuecomment-2077471668

While the iRODS server does not allow at start time to have a zone which name contains special chars

```
irods.exceptions.IrodsError: ERROR: Validation Failed for [/etc/irods/server_config.json]:
	against [file:///var/lib/irods/configuration_schemas/v4/server_config.json]
	ValidationError: 'überZone' does not match '^[A-Za-z0-9_\\.]+$'

Failed validating 'pattern' in schema['properties']['zone_name']:
    {'maxLength': 63, 'pattern': '^[A-Za-z0-9_\\.]+$', 'type': 'string'}

On instance['zone_name']:
    'überZone'
```

iadmin does. For instance,

```
#!/bin/env bash
zone=$(printf '\xc3\xbcberZone')
iadmin modzone testZone name $zone
```
succeeds and iRODS works as expected. Would it be possible to have a consistent behavior in the treatment of special chars, please?
This would make sure that an admin that modified a zone name to contain special chars will still be able to restart the server in case of need.

PR #8174, opened 2025-02-07
[#7722] Validate and reject invalid zone names (4-3-stable)

korydraughn:
Running tests.

### DIFF
--- a/plugins/database/src/db_plugin.cpp
+++ b/plugins/database/src/db_plugin.cpp
@@ -1,3 +1,4 @@
+#include "irods/administration_utilities.hpp"
 #include "irods/authenticate.h"
 #include "irods/catalog.hpp"
 #include "irods/catalog_utilities.hpp"
@@ -937,33 +938,6 @@ bool _userInRUserAuth( const char* userName, const char* zoneName, const char* a
         return true;
     }
 }
-
-// =-=-=-=-=-=-=-
-/// @brief function which determines if a char is allowed in a zone name
-static bool allowed_zone_char( const char _c ) {
-    return ( !std::isalnum( _c ) &&
-             !( '.' == _c )      &&
-             !( '_' == _c ) );
-} // allowed_zone_char
-
-// =-=-=-=-=-=-=-
-/// @brief function for validating the name of a zone
-irods::error validate_zone_name(
-    std::string _zone_name ) {
-    std::string::iterator itr = std::find_if( _zone_name.begin(),
-                                _zone_name.end(),
-                                allowed_zone_char );
-    if ( itr != _zone_name.end() || _zone_name.length() >= NAME_LEN ) {
-        std::stringstream msg;
-        msg << "validate_zone_name failed for zone [";
-        msg << _zone_name;
-        msg << "]";
-        return ERROR( SYS_INVALID_INPUT_PARAM, msg.str() );
-    }
-
-    return SUCCESS();
-
-} // validate_zone_name
 
 /* delCollection (internally called),
    does not do the commit.
@@ -5214,12 +5188,9 @@ irods::error db_reg_zone_op(
         return ERROR( CAT_INVALID_ARGUMENT, "Currently, only zones of type 'remote' are allowed" );
     }
 
-    // =-=-=-=-=-=-=-
-    // validate the zone name does not include improper characters
-    ret = validate_zone_name( _zone_name );
-    if ( !ret.ok() ) {
-        irods::log( ret );
-        return PASS( ret );
+    if (!irods::is_zone_name_valid(_zone_name)) {
+        log_db::error("{}: Zone name [{}] does not satisfy requirements.", __func__, _zone_name);
+        return ERROR(SYS_INVALID_INPUT_PARAM, "Zone name contains invalid characters");
     }
 
     // =-=-=-=-=-=-=-
@@ -5393,17 +5364,10 @@ irods::error db_mod_zone_op(
             return ERROR( CAT_INVALID_ARGUMENT, "cannot rename localzone" );
         }
 
-        // =-=-=-=-=-=-=-
-        // validate the zone name does not include improper characters
-        ret = validate_zone_name( _option_value );
-        if ( !ret.ok() ) {
-            irods::log( ret );
-            std::string msg( "zone name is invalid [" );
-            msg += _option_value;
-            msg += "]";
-            addRErrorMsg( &_ctx.comm()->rError, 0,
-                          msg.c_str() );
-            return PASS( ret );
+        if (!irods::is_zone_name_valid(_option_value)) {
+            log_db::error("{}: Zone name [{}] does not satisfy requirements.", __func__, _option_value);
+            addRErrorMsg(&_ctx.comm()->rError, 0, fmt::format("zone name is invalid [{}]", _option_value).c_str());
+            return ERROR(SYS_INVALID_INPUT_PARAM, "Zone name contains invalid characters");
         }
 
         cllBindVars[cllBindVarCount++] = _option_value;
--- a/scripts/irods/test/test_iadmin.py
+++ b/scripts/irods/test/test_iadmin.py
@@ -1823,6 +1823,67 @@ class Test_Iadmin(resource_suite.ResourceBase, unittest.TestCase):
             self.admin.run_icommand(["irmdir", test_collection_name])
             self.admin.run_icommand(["irm", "-f", test_dataobj_name])
 
+    def test_modzone_does_not_accept_invalid_zone_names_for_the_local_zone__issue_7722(self):
+        # Show that "iadmin modzone name" no longer accepts invalid zone names.
+        invalid_zone_name = 'überZone'
+        ec, _, _ = self.admin.assert_icommand(
+            ['iadmin', 'modzone', self.admin.zone_name, 'name', invalid_zone_name],
+            'STDERR',
+            ['-130000 SYS_INVALID_INPUT_PARAM'],
+            input='yes')
+        self.assertNotEqual(ec, 0)
+
+        # Show the zone name was not updated.
+        _, out, err = self.admin.assert_icommand(['iadmin', 'lz'], 'STDOUT')
+        self.assertNotIn(invalid_zone_name, out)
+        self.assertIn(self.admin.zone_name, out)
+        self.assertEqual(len(err), 0)
+
+        # Show the zone is still functional.
+        self.admin.assert_icommand(['ils'], 'STDOUT', [self.admin.session_collection + ':'])
+
+    def test_mkzone_does_not_accept_invalid_zone_names__issue_7722(self):
+        # Show that "iadmin mkzone" no longer accepts invalid zone names.
+        invalid_zone_name = 'überZone'
+        ec, _, _ = self.admin.assert_icommand(
+            ['iadmin', 'mkzone', invalid_zone_name, 'remote', 'localhost:1247'], 'STDERR', ['-130000 SYS_INVALID_INPUT_PARAM'])
+        self.assertNotEqual(ec, 0)
+
+        # Show the zone was not added.
+        _, out, err = self.admin.assert_icommand(['iadmin', 'lz'], 'STDOUT')
+        self.assertNotIn(invalid_zone_name, out)
+        self.assertIn(self.admin.zone_name, out)
+        self.assertEqual(len(err), 0)
+
+    def test_modzone_does_not_accept_invalid_zone_names_for_remote_zones__issue_7722(self):
+        good_zone_name = 'goodZoneIssue7722'
+
+        try:
+            self.admin.assert_icommand(['iadmin', 'mkzone', good_zone_name, 'remote', 'localhost:1247'])
+
+            # Show that "iadmin modzone name" no longer accepts invalid zone names.
+            # Modification of a zone name takes different code paths depending on whether
+            # the zone name matches the local zone or a remote zone.
+            invalid_zone_name = 'überZone'
+            ec, _, _ = self.admin.assert_icommand(
+                ['iadmin', 'modzone', good_zone_name, 'name', invalid_zone_name],
+                'STDERR',
+                ['-130000 SYS_INVALID_INPUT_PARAM'],
+                input='yes')
+            self.assertNotEqual(ec, 0)
+
+            # Show the zone name was not updated.
+            _, out, err = self.admin.assert_icommand(['iadmin', 'lz'], 'STDOUT')
+            self.assertNotIn(invalid_zone_name, out)
+            self.assertIn(good_zone_name, out)
+            self.assertEqual(len(err), 0)
+
+            # Show the local zone is still functional.
+            self.admin.assert_icommand(['ils'], 'STDOUT', [self.admin.session_collection + ':'])
+
+        finally:
+            self.admin.run_icommand(['iadmin', 'rmzone', good_zone_name])
+
 class Test_Iadmin_Resources(resource_suite.ResourceBase, unittest.TestCase):
 
     def setUp(self):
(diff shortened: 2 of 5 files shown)


plugins/database/src/db_plugin.cpp
```diff
@@ -938,31 +939,20 @@ bool _userInRUserAuth( const char* userName, const char* zoneName, const char* a
...
-        msg << _zone_name;
-        msg << "]";
-        return ERROR( SYS_INVALID_INPUT_PARAM, msg.str() );
+static irods::error validate_zone_name(const std::string& _zone_name)
+{
+    // Verify the zone name satisfies the requirements defined by the configuration
+    // schema. It would be nice to get the regular expression from the schema file, but
+    // we can't do that because iRODS 4 supports remote schema files. However, hard-coding
+    // the regular expression is also fine because no one should modify the schema (even
+    // though it's modifiable).
```
MartinFlores751, line 948: The comment on the regex is a bit confusing to me.

If the comment is to highlight potential future issues by hard-coding the regex, the comment should be simplified to that. State the assumption of a default iRODS 4 regex and that issues may arise if the regex file is changed (if I understood the comment).

scripts/irods/test/test_iadmin.py
```diff
@@ -1823,6 +1823,41 @@ def test_iadmin_rmuser_removes_associated_acls__issue_7778(self):
...
 
+    def test_modzone_does_not_accept_invalid_zone_names__issue_7722(self):
+        # Show that "iadmin modzone name" no longer accepts invalid zone names.
+        invalid_zone_name = 'überZone'
+        ec, _, _ = self.admin.assert_icommand(
+            ['iadmin', 'modzone', self.admin.zone_name, 'name', invalid_zone_name],
+            'STDERR',
+            ['-130000 SYS_INVALID_INPUT_PARAM'],
+            input='yes')
+        self.assertNotEqual(ec, 0)
```
MartinFlores751, line 1834: Do we expect the same ec every time from the `iadmin ...` command? If we do, we should use `desired_rc` in the `assert_icommand(...)` function.

This may catch changes that may break expectations (e.g. scripting).

scripts/irods/test/test_iadmin.py
```diff
@@ -1823,6 +1823,41 @@ def test_iadmin_rmuser_removes_associated_acls__issue_7778(self):
...
+
+        # Show the zone name was not updated.
+        ec, out, err = self.admin.assert_icommand(['iadmin', 'lz'], 'STDOUT')
+        self.assertEqual(ec, 0)
```
MartinFlores751, line 1838: Consider using `desired_rc` in the `assert_icommand(...)` function.

The logging provided by the function may help catch issues later.

scripts/irods/test/test_iadmin.py
```diff
@@ -1823,6 +1823,41 @@ def test_iadmin_rmuser_removes_associated_acls__issue_7778(self):
...
+        self.assertEqual(len(err), 0)
+
+        # Show the zone is still functional.
+        self.admin.assert_icommand(['ils'], 'STDOUT', [self.admin.session_collection + ':'])
+
+    def test_mkzone_does_not_accept_invalid_zone_names__issue_7722(self):
+        # Show that "iadmin mkzone" no longer accepts invalid zone names.
+        invalid_zone_name = 'überZone'
+        ec, out, err = self.admin.assert_icommand(['iadmin', 'mkzone', invalid_zone_name, 'remote', 'localhost:1247'], 'STDERR')
+        self.assertNotEqual(ec, 0)
```
MartinFlores751, line 1850: Consider using `desired_rc` in the `assert_icommand(...)` function.

The logging provided by the function may help catch issues later.

scripts/irods/test/test_iadmin.py
```diff
@@ -1823,6 +1823,41 @@ def test_iadmin_rmuser_removes_associated_acls__issue_7778(self):
...
+        print('stdout =', out)
+        print('stderr =', err)
```
MartinFlores751, line 1852: Are these prints necessary? Is the logging provided by `assert_icommand(...)` not working for this case?

scripts/irods/test/test_iadmin.py
```diff
@@ -1823,6 +1823,41 @@ def test_iadmin_rmuser_removes_associated_acls__issue_7778(self):
...
+
+        # Show the zone was not added.
+        ec, out, err = self.admin.assert_icommand(['iadmin', 'lz'], 'STDOUT')
+        self.assertEqual(ec, 0)
```
MartinFlores751, line 1856: Consider using `desired_rc` in the `assert_icommand(...)` function.

The logging provided by the function may help catch issues later.

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -1026,6 +1027,20 @@ _rsGeneralAdmin( rsComm_t *rsComm, generalAdminInp_t *generalAdminInp ) {
             rei.uoip = &rsComm->proxyUser;
             args[0] = generalAdminInp->arg2;
             args[1] = generalAdminInp->arg3;
+
+            // Verify the zone name satisfies the requirements defined by the configuration
+            // schema. It would be nice to get the regular expression from the schema file, but
+            // we can't do that because iRODS 4 supports remote schema files. However, hard-coding
+            // the regular expression is also fine because no one should modify the schema (even
+            // though it's modifiable).
```
MartinFlores751, line 1035: I have the same comment on this comment in my previous comment.

If the comment is to highlight potential future issues by hard-coding the regex, the comment should be simplified to that. State the assumption of a default iRODS 4 regex and that issues may arise if the regex file is changed (if I understood the comment).

server/core/src/administration_utilities.cpp
```diff
@@ -163,4 +167,9 @@ namespace irods
 
         return 0;
     } // remove_user
+
+    auto is_valid_zone_name(const std::string_view _zone_name) -> bool
+    {
+        return std::regex_match(std::begin(_zone_name), std::end(_zone_name), std::regex{R"(^[A-Za-z0-9\.]+$)"});
```
MartinFlores751, line 173: Should std::regex be `static` so it's not built on every function call?

If this function is called often, I think it might be worth considering.

MartinFlores751 (commented):
Looks like a good change. Most of my comments are in the python code, particularly on `assert_icommand(...)`.

plugins/database/src/db_plugin.cpp
```diff
@@ -938,31 +939,20 @@ bool _userInRUserAuth( const char* userName, const char* zoneName, const char* a
...
-    if ( itr != _zone_name.end() || _zone_name.length() >= NAME_LEN ) {
-        std::stringstream msg;
-        msg << "validate_zone_name failed for zone [";
-        msg << _zone_name;
-        msg << "]";
-        return ERROR( SYS_INVALID_INPUT_PARAM, msg.str() );
+static irods::error validate_zone_name(const std::string& _zone_name)
+{
+    // Verify the zone name satisfies the requirements defined by the configuration
+    // schema. It would be nice to get the regular expression from the schema file, but
```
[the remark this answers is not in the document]
korydraughn: I'm going to remove this function wrapper since it doesn't add anything to the implementation.

plugins/database/src/db_plugin.cpp
korydraughn: Will revisit wording.

scripts/irods/test/test_iadmin.py
korydraughn: The actual error code returned to the OS by the icommands aren't important. Unless otherwise stated, their only purpose is to indicate 0 (success) or non-zero (failure).

scripts/irods/test/test_iadmin.py
korydraughn: See https://github.com/irods/irods/pull/8174/files#r1946983500.

scripts/irods/test/test_iadmin.py
korydraughn: https://github.com/irods/irods/pull/8174/files#r1946983500.

scripts/irods/test/test_iadmin.py
korydraughn: I forgot to remove these.

Will remove.

scripts/irods/test/test_iadmin.py
korydraughn: This check is actually redundant.

Will remove it.

scripts/irods/test/test_iadmin.py
korydraughn: This is redundant.

Will remove.

server/core/src/administration_utilities.cpp
korydraughn: I don't expect users to rename their zone often, if ever.

We recommend they spend time thinking of a good zone name during initial setup of the zone and sticking with it.

korydraughn: However, given this is a free function, I'm now leaning toward making it static since we don't know how people will use it.

Will think about this a bit more.

plugins/database/src/db_plugin.cpp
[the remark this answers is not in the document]
MartinFlores751: Yeah, skimming the code, I think that the main reason to leave the function is to not have to change the call-sites.

NP if you remove the function or don't, since there's only a few locations this is called (from a quick grep).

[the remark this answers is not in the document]
korydraughn: Removed function.

plugins/database/src/db_plugin.cpp
korydraughn: Removed wording in favor of TODO comment.

Issue coming soon.

scripts/irods/test/test_iadmin.py
korydraughn: Correct.

korydraughn: Removed assertion.

scripts/irods/test/test_iadmin.py
korydraughn: Removed.

scripts/irods/test/test_iadmin.py
korydraughn: Removed.

server/api/src/rsGeneralAdmin.cpp
korydraughn: Removed comment.

server/core/src/administration_utilities.cpp
korydraughn: Changed it to be static.

server/core/src/administration_utilities.cpp
```diff
@@ -163,4 +167,11 @@ namespace irods
 
         return 0;
     } // remove_user
+
+    auto is_zone_name_valid(const std::string_view _zone_name) -> bool
+    {
+        // TODO(#NNNN): This should come from the JSON schema file to avoid drift.
```
korydraughn, line 173: Needs issue number.

scripts/irods/test/test_iadmin.py
korydraughn: The only change to this was replacing `out` and `err` with underscores.

server/core/src/administration_utilities.cpp
korydraughn: Created #8175.

Will update TODO comment.

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -36,6 +36,7 @@
 #include <cctype>
 #include <iostream>
 #include <optional>
+#include <regex>
```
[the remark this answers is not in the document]
korydraughn: Done.

server/core/src/administration_utilities.cpp
korydraughn: Done.

korydraughn:
First run of unit tests failed due to missing underscore in regex.

Running unit tests again.

korydraughn:
Unit tests passed.
```
-----
All tests passed! :)
time elapsed: [ 160.8136]seconds ([   0]hours [ 2.6802]minutes)

2025-02-07 14:43:52,861 - version:[4.3.3]
2025-02-07 14:43:52,861 - sha:[86124e9635a1b4bd12fa698ec460cf7cbd6abb16]
```

Running core tests.

alanking (commented):
Seems good to me. Martin hit all the comments I was going to make :) Just one more question from me

scripts/irods/test/test_iadmin.py
```diff
@@ -1823,6 +1823,37 @@ def test_iadmin_rmuser_removes_associated_acls__issue_7778(self):
...
+        self.assertIn(self.admin.zone_name, out)
+        self.assertEqual(len(err), 0)
+
+        # Show the zone is still functional.
+        self.admin.assert_icommand(['ils'], 'STDOUT', [self.admin.session_collection + ':'])
+
+    def test_mkzone_does_not_accept_invalid_zone_names__issue_7722(self):
+        # Show that "iadmin mkzone" no longer accepts invalid zone names.
+        invalid_zone_name = 'überZone'
+        ec, _, _ = self.admin.assert_icommand(['iadmin', 'mkzone', invalid_zone_name, 'remote', 'localhost:1247'], 'STDERR')
```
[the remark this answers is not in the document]
korydraughn: Just forgot to include it.

Will update.

korydraughn:
Core tests passed.

Will rerun that one test after addressing the review comment.

scripts/irods/test/test_iadmin.py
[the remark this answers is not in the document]
korydraughn: Updated.

korydraughn:
Changes to test referenced at https://github.com/irods/irods/pull/8174#discussion_r1947115350 pass.

This is ready.

scripts/irods/test/test_iadmin.py
```diff
@@ -1823,6 +1823,38 @@ def test_iadmin_rmuser_removes_associated_acls__issue_7778(self):
...
             self.admin.run_icommand(["irm", "-f", test_dataobj_name])
 
+    def test_modzone_does_not_accept_invalid_zone_names__issue_7722(self):
+        # Show that "iadmin modzone name" no longer accepts invalid zone names.
+        invalid_zone_name = 'überZone'
+        ec, _, _ = self.admin.assert_icommand(
+            ['iadmin', 'modzone', self.admin.zone_name, 'name', invalid_zone_name],
+            'STDERR',
+            ['-130000 SYS_INVALID_INPUT_PARAM'],
+            input='yes')
```
alanking, line 1833: Although I admire the level of trust, I wonder whether we should avoid modifying the zone which is running the tests. If a regression is introduced, there's no recovering from this outside of modifying the database. It's not as big of a deal as in the past since all of our testing is done in relatively short-lived containers now, but I'm wondering whether creating a zone with a valid name and modifying that zone would be better. Up to you

alanking (commented):
Had one suggestion on further reflection, but it's minor, and deals exclusively with the tests. Up to you if you want to address it. If not, go ahead and squash.

scripts/irods/test/test_iadmin.py
korydraughn: That's a good point. The problem is that working with a remote zone doesn't trigger the code path we're interested in.

Your comment does point out the lack of a test for modifying the name of remote zones. I'll add a test for that.

I also think the implementation for renaming zones should be reviewed and made more atomic and recoverable. I'll open an issue for that.

korydraughn: Added a new test for renaming a remote zone.

Will test it shortly.

korydraughn:
The new test passed.

No need to rerun the unit/core test suites.

This is ready.

alanking (commented):
Changes look good to me. I think it's ready for squashing.

korydraughn:
Squashed.

korydraughn:
Pounded.

Issue #7685, opened 2024-04-11
Compiler warnings building genquery2 parser and lexer


alanking:
- [x] main
- [x] 4-3-stable

---

I noticed a couple of warnings from the compiler while building:

```
[524/1166] [BISON][irods_parser] Building parser with bison 3.8.2
/irods_source/server/genquery2/dsl/parser.y:165.5-10: warning: empty rule for typed nonterminal, and no action [-Wother]
  165 |     %empty
      |     ^~~~~~
[525/1166] [FLEX][irods_lexer] Building scanner with flex 2.6.4
/irods_source/server/genquery2/dsl/lexer.l:109: warning, -s option given but default rule can be matched
```

This is pretty low priority from what I can tell, but figured I would write it up.

Here's a permalink in case something moves in the future...
https://github.com/irods/irods/blob/669a726968dcdd7c4dab7d52761095119511c182/server/genquery2/dsl/parser.y#L164-L166

alanking:
Looks like #7847 might have fixed the warning from bison. Still seeing the flex warning.

PR #8179, opened 2025-02-10
[#7685] Fix flex warning for GenQuery2 lexer rules (4-3-stable)

korydraughn:
All unit/core tests pass.

### DIFF
--- a/unit_tests/src/test_rc_genquery2.cpp
+++ b/unit_tests/src/test_rc_genquery2.cpp
@@ -32,6 +32,10 @@ TEST_CASE("rc_genquery2")
         input.zone = nullptr;
 
         char* output{};
+        irods::at_scope_exit_unsafe free_output{[&output] {
+            // NOLINTNEXTLINE(cppcoreguidelines-owning-memory,cppcoreguidelines-no-malloc)
+            std::free(output);
+        }};
         REQUIRE(rc_genquery2(static_cast<RcComm*>(conn), &input, &output) == 0);
         REQUIRE_FALSE(output == nullptr);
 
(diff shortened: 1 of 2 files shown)


alanking (commented):
What is the result of the commit with no issue number? Will that need a test?

korydraughn:
I haven't decided on whether to create a new issue or link it to the housekeeping one yet.

korydraughn:
Added issue number to commit.

alanking (commented):
Just to make sure... Should we add a test for the multi-whitespace capturing change? Other than this question, changes look good to me.

korydraughn:
I don't think we need one, but it doesn't hurt to add a test for that.

Adding test now.

korydraughn:
Test added and it passes.

unit_tests/src/test_rc_genquery2.cpp
```diff
@@ -118,4 +118,30 @@ TEST_CASE("rc_genquery2")
...
+
+        		where       	COLL_NAME
+=      				'{}')", env.rodsHome);
+
+        Genquery2Input input{};
+        input.query_string = query_string.data();
+
+        char* output{};
+        REQUIRE(rc_genquery2(static_cast<RcComm*>(conn), &input, &output) == 0);
+        REQUIRE_FALSE(output == nullptr);
```
alanking, line 139: Will `output` leak once we get past this assertion?

korydraughn: It sure will.

Nice catch.

Will fix this one and check the other tests too.

korydraughn: Fixed.

alanking:
Looks good to me... Once we've got an issue number for that last commit, please squash to taste.

korydraughn:
Tests still pass.

Squashing.

korydraughn:
Squashed.

alanking (approved):
\# it

korydraughn:
Pounded.

Issue #8166, opened 2025-02-05
`istream --count` does not work correctly for values 2^22 and higher.


ghuls:
- [x] main
- [x] 4-3-stable
---
## Bug Report

`istream --count` does not work correctly for values 2^22 (4194304) and higher as from that value, it downloads the rest of the fie.

```bash
$ ils -l /bla/bli/very_big_file.tar
  ghuls          0 default;netapp 1275496427520 2024-12-27.20:02 & very_big_file.tar

# Till count of 4194303, the output file has the correct size.
$ istream read -o 24919971840 -c 4194303 /bla/bli/very_big_file.tar > subset.working.tar

# From count of 4194304, the output file will continue downloading (likely to the end of the file, as it is more than 1TB, I killed it before)
$ istream read -o 24919971840 -c 4194304 /bla/bli/very_big_file.tar > subset.not_working.tar
```

PR #8180, opened 2025-02-10
[#8166] Add test for `istream read --count` (4-3-stable)

korydraughn:
Companion PR for irods/irods_client_icommands#530.

All unit/core tests pass.

### DIFF
--- a/scripts/irods/lib.py
+++ b/scripts/irods/lib.py
@@ -786,6 +786,16 @@ def make_arbitrary_file(f_name, f_size, buffer_size=32*1024*1024):
 
             bytes_written += to_write
 
+def create_ascii_printable_file(filename, filesize, seed=None):
+    """Creates a file with pseudo-random ASCII printable characters."""
+    if seed is not None:
+        random.seed(seed)
+    import string
+    # This avoids control sequences.
+    safe_chars = string.digits + string.ascii_letters + string.punctuation + " "
+    with open(filename, "w") as f:
+        f.write("".join(random.choices(safe_chars, k=filesize)))
+
 def get_resource_parent_id(session, resource_name):
     return session.run_icommand(['iquest', '%s',
         "select RESC_PARENT where RESC_NAME = '{}'"
--- a/scripts/irods/test/test_istream.py
+++ b/scripts/irods/test/test_istream.py
@@ -1,4 +1,3 @@
-from __future__ import print_function
 import os
 import sys
 import tempfile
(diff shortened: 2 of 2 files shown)


scripts/irods/lib.py
```diff
@@ -786,6 +786,19 @@ def make_arbitrary_file(f_name, f_size, buffer_size=32*1024*1024):
...
+    
+    - Does NOT use OS entropy (no `os.urandom` or `secrets`).
+    - Uses `random` for deterministic pseudo-randomness.
+    """
+    import string
+    # This avoids control sequences.
+    safe_chars = string.digits + string.ascii_letters + string.punctuation + " "
+    with open(filename, "w") as f:
+        for _ in range(filesize):
+            f.write(random.choice(safe_chars))
```
MartinFlores751, line 800: I think we could use `choices(...)` here instead.
This would remove the for loop.

See: https://docs.python.org/3/library/random.html#random.choices

scripts/irods/lib.py
```diff
@@ -786,6 +786,19 @@ def make_arbitrary_file(f_name, f_size, buffer_size=32*1024*1024):
 
             bytes_written += to_write
 
+def create_ascii_printable_file(filename, filesize):
+    """Creates a file with pseudo-random ASCII printable characters.
+    
+    - Does NOT use OS entropy (no `os.urandom` or `secrets`).
+    - Uses `random` for deterministic pseudo-randomness.
+    """
```
MartinFlores751, line 794: I think we should seed the pseudo-randomizer like in `make_arbitrary_file(...)`. (i.e., `random.seed(5)`)

Otherwise, we end up with a pseudo-randomizer with an unknown start state. This would mean the state would now depend on the order the tests are executed and the number of parallel test containers used.

scripts/irods/test/test_istream.py
```diff
@@ -486,3 +485,37 @@ def test_sibling_replicas_marked_stale_on_create__issue_5187(self):
...
+                    data = f.read(count)
+                    self.assertEqual(data, out)
+
+            # 4194304 is special because it is what causes istream to enter a code
+            # path which leads to reading all the replica's data.
+            do_test(offset=0, count=4194304)
+
+            # These cases exist to help guard against future modifications.
+            do_test(offset=10000000, count=4194304)
+            do_test(offset=local_filesize - 100, count=100)
```
alanking, line 517: Consider making these separate tests or using the `unittest.TestCase.subTest` feature so that they're named and failures are a little clearer.

scripts/irods/lib.py
MartinFlores751: I think you may have to change `write(...)` to `writelines(...)` too if you decide to use `choices(...)`.

MartinFlores751 (commented):
Changes looking good, only have some suggestions on `create_ascii_printable_file(...)`.

scripts/irods/lib.py
korydraughn: Will add a parameter to allow the caller to set a seed.

scripts/irods/lib.py
korydraughn: Updated.

scripts/irods/lib.py
korydraughn: Done.

scripts/irods/test/test_istream.py
korydraughn: Updated.

korydraughn:
Test passes with changes.

scripts/irods/lib.py
```diff
@@ -786,6 +786,16 @@ def make_arbitrary_file(f_name, f_size, buffer_size=32*1024*1024):
...
 
+def create_ascii_printable_file(filename, filesize, seed=None):
+    """Creates a file with pseudo-random ASCII printable characters."""
+    if seed is not None:
+        random.seed(seed)
+    import string
+    # This avoids control sequences.
+    safe_chars = string.digits + string.ascii_letters + string.punctuation + " "
+    with open(filename, "w") as f:
+        f.write("".join(random.choices(safe_chars, k=filesize)))
```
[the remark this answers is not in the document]
korydraughn: Will test the writelines function to see what it produces.

alanking (commented):
Looks good to me... I think it's ready for squashing.

alanking (approved):
\# it

korydraughn:
Pounded.

Issue #7795, opened 2022-03-07
iquest parses queries with tabs in the where clause incorrectly.


tedgin:
- [x] main
- [x] 4-3-stable

---

I'm using iRODS iCommands 4.2.8.

Where there is a horizontal tab in the where clause of a generalized query, iquest doesn't return any results. In the following example there is a horizontal tab immediately before `and`,

```bash
prompt> printf -v query $'select DATA_NAME, DATA_SIZE where COLL_NAME = \'/iplant/home/tedgin\'\tand DATA_NAME like \'test%%\''
prompt> iquest '%s %s' "$query"
CAT_NO_ROWS_FOUND: Nothing was found matching your query
```

If the horizontal tab is replaced by a space, iquest works as expected.

```bash
prompt> printf -v query $'select DATA_NAME, DATA_SIZE where COLL_NAME = \'/iplant/home/tedgin\' and DATA_NAME like \'test%%\''
prompt> iquest '%s %s' "$query"
test.tgz 110084063
test0B 0
test100GiB 107374182400
```

alanking:
`iquery` works!

```bash
$ printf -v query $'select DATA_NAME, DATA_SIZE where COLL_NAME = \'/tempZone/home/alan\'\tand DATA_NAME like \'test%%\''
$ ils -l
/tempZone/home/alan:
  alan              0 demoResc          186 2024-06-12.16:20 & test1
$ iquest "$query"
CAT_NO_ROWS_FOUND: Nothing was found matching your query
$ iquery "$query"
[["test1","186"]]
```

korydraughn:
Let's make this a test so we avoid regressions.

Issue #6421, opened 2022-05-25
`SYS_AGENT_INIT_ERR` when attempting to `iget` a file accross federation


kript:
- [x] main
- [x] 4-3-stable

---

### iRODS Version, OS and Version

4.2.7
Ubuntu 18.04

### What did you try to do?

`iget` a file from a remote zone where the user account doesn't have a federated user, but _only_ a local user of the same name.

e.g.
jc18#RemoteZone from the federated `RemoteZone` attempts to download a file from `localZone`.
The ACL for the file has jc18 (and thus an implied `jc18#localZone`).
There is *no* Federated user account (i.e. `jc18#RemoteZone` created in the `localZone`.

### Expected behavior

`CAT_NO_ACCESS_PERMISSION`

### Observed behaviour (including steps to reproduce, if applicable)

```
iget -V --progress /localZone/projects/file
0/1 -  0.00% of files done   0.000/760.332 MB -  0.00% of file sizes done
Processing file - 760.332 MB   2022-05-25.10:07:53
remote addresses: 10.160.2.32 ERROR: getUtil: get error for ./file status = -38000 SYS_AGENT_INIT_ERR
```

kript:
Just hit this again, but this time there was _no_ corresponding user on the remote zone, so I would have expected an error relating to that.

trel:
Is there anything interesting/related in either the remoteZone rodsLog or the localZone rodsLog?

I would have expected `CAT_INVALID_USER` before `CAT_NO_ACCESS_PERMISSION` for an undefined user.

kript:
I _think_ this is the pertinent entries;

```
irods@irods-humgen-ies1:/var/lib/irods/log$ grep "pid:31545" rodsLog.2022.06.11
Jun 15 17:03:55 pid:31545 NOTICE: _rsGenQuery: genQuery status = -818000
Jun 15 17:03:55 pid:31545 NOTICE: rsGenQuery: _rsGenQuery failed, status = -818000
Jun 15 17:03:55 pid:31545 NOTICE: getDataObjInfo: rsGenQuery error, status = -818000
Jun 15 17:03:55 pid:31545 NOTICE: _rsGenQuery: genQuery status = -818000
Jun 15 17:03:55 pid:31545 NOTICE: rsGenQuery: _rsGenQuery failed, status = -818000
Jun 15 17:03:55 pid:31545 NOTICE: getDataObjInfo: rsGenQuery error, status = -818000
[...]

Issue #4984, opened 2020-06-24
Unencrypted parallel upload fails when server-to-server SSL enabled 


tedgin:
### iRODS Version, OS and Version

iRODS 4.1.10, CentOS 7

### What did you try to do?

I have an IES with a separate resource server.  I enabled TLS encryption on them by modifying their configuration as follows.

`core.re`:
```
acPreConnect(*OUT) { *OUT="CS_NEG_DONT_CARE"; }
```

`irods_environment.json`:
```
  "irods_client_server_negotiation": "request_server_negotiation",
  "irods_client_server_policy": "CS_NEG_REQUIRE",
  "irods_encryption_algorithm": "AES-256-CBC",
  "irods_encryption_num_hash_rounds": 16,
  "irods_encryption_key_size": 32,
  "irods_encryption_salt_size": 8,
  "irods_ssl_certificate_chain_file": "/etc/pki/tls/certs/cyverse.rocks-chain.crt",
  "irods_ssl_certificate_key_file": "/etc/pki/tls/private/cyverse.rocks.key",
  "irods_ssl_dh_params_file": "/var/lib/irods/.irods/dhparams.pem",
  "irods_ssl_verify_server": "cert",
```

From a third computer acting as a client, I used `iput` to upload a file large enough to trigger parallel a upload. The client was not configured to use TLS.  

### Expected behavior

I expected the file to be uploaded successfully.

### Observed behavior (including steps to reproduce, if applicable)

```
rods@dev-play ~? iput test64MiB
ERROR: rcPartialDataPut: toWrite 4194304, bytesWritten 183138, errno = 104 status = -27104 SYS_COPY_LEN_ERR, Connection reset by peer
ERROR: rcPartialDataPut: toWrite 4194304, bytesWritten 183138, errno = 104 status = -27104 SYS_COPY_LEN_ERR, Connection reset by peer
ERROR: rcPartialDataPut: toWrite 4194304, bytesWritten 183138, errno = 104 status = -27104 SYS_COPY_LEN_ERR, Connection reset by peer
ERROR: putUtil: put error for /cyverse.dev/home/rods/test64MiB, status = -27104 status = -27104 SYS_COPY_LEN_ERR, Connection reset by peer
```

In the IES, the following entries were logged.

```
Jun 24 11:41:52 pid:10132 NOTICE: Agent process 16153 started for puser=rods and cuser=rods from 128.196.65.179
[...]

PR #8195, opened 2025-02-13
Fix unencrypted parallel transfer with server encryption enabled, test iget over federation (4-3-stable)

alanking:
Addresses #4984 
In service of #6421
In service of #7795

Unit tests and federation tests passed. Core tests are running now.

There's no way to automatically test the scenario described in #4984 with our existing test infrastructure. As such, there are no new tests for that scenario, and I confirmed manually that things are working for Puts and Gets.

### DIFF
--- a/scripts/irods/test/test_federation.py
+++ b/scripts/irods/test/test_federation.py
@@ -2393,6 +2393,89 @@ class test_irepl_all_permission_levels__issue_7444_7465(unittest.TestCase):
 					self.user0.assert_icommand(['irm', '-f', logical_path])
 
 
+class test_iget_data_in_remote_zone(unittest.TestCase):
+	@classmethod
+	def setUpClass(self):
+		# Create a session as the administrator for the remote zone so we can do things in the remote zone.
+		self.remote_admin = session.make_session_for_existing_user(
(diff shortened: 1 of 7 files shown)


korydraughn:
Got it. Please open an issue to add support for testing this in an automated fashion.

alanking:
I think https://github.com/irods/irods_testing_environment/issues/239 captures this pretty well. Basically we want a way to configure a zone to one's specifications and in the case of https://github.com/irods/irods_testing_environment/issues/239, a JSON input generated from a zone report. Doesn't have to just be a zone report, but would allow for things like mixed-SSL zones, using `CS_NEG_DONT_CARE`, etc.

korydraughn:
That's very true. Don't worry about opening an issue for that.

alanking:
Sliding in one more while waiting on the other core tests to finish. New test passes.

korydraughn:
Will review soon.

korydraughn:
Let's keep it. The test in the other PR tests the API directly. Your test covers `iquery`.

korydraughn (approved):
Looks ready to me.

Pound it if everything's passing.

alanking:
Core tests passed. #'ing and whatnot shortly

Issue #7392, opened 2023-10-25
iRODS 4.3.1 fails to start after upgrade because of missing required parameter in service account environment


cookie33:
- [x] irods_docs
- [x] Confirm upgrades from iRODS 4.2.x to 4.3.3 do not run into the schema validation issue (it does, but situationally - https://github.com/irods/irods/issues/7392#issuecomment-2669579446)

---

### iRODS Version, OS and Version

iRODS 4.3.1
CentOS Linux release 7.9.2009 (Core)

### What did you try to do?

Upgrade iRODS from 4.3.0 to 4.3.1

### Expected behavior

iRODS starts after upgrade

### Observed behavior (including steps to reproduce, if applicable)

iRODS does not start after upgrade. It complains about a missing required variable `irods_connection_pool_refresh_time_in_seconds` in ~/.irods/irods_environment.json.

Following is output of `/var/lib/irods/log/control_log.txt`:
```
...
2023-10-24T12:43:14.780Z -   DEBUG -             json_validation.py:  59 - {'$id': 'file:///var/lib/irods/configuration_schemas/v4/service_account_environment.json',
 '$schema': 'https://json-schema.org/draft/2020-12/schema',
 'allOf': [{'$ref': 'client_environment.json'}],
 'properties': {'irods_authentication_file': {'type': 'string'},
                'irods_authentication_scheme': {'type': 'string'},
                'irods_client_server_negotiation': {'enum': ['request_server_negotiation',
                                                             'off']},
                'irods_client_server_policy': {'enum': ['CS_NEG_REFUSE',
                                                        'CS_NEG_REQUIRE',
                                                        'CS_NEG_DONT_CARE']},
                'irods_connection_pool_refresh_time_in_seconds': {'type': 'integer'},
                'irods_cwd': {'type': 'string'},
                'irods_debug': {'type': 'string'},
                'irods_default_hash_scheme': {'type': 'string'},
                'irods_default_number_of_transfer_threads': {'type': 'integer'},
                'irods_default_resource': {'type': 'string'},
                'irods_encryption_algorithm': {'type': 'string'},
                'irods_encryption_key_size': {'type': 'integer'},
                'irods_encryption_num_hash_rounds': {'type': 'integer'},
                'irods_encryption_salt_size': {'type': 'integer'},
                'irods_gsi_server_dn': {'type': 'string'},
                'irods_home': {'type': 'string'},
                'irods_log_level': {'type': 'integer'},
[...]

PR #8209, opened 2025-02-18
[#7392] fix: Add missing required `irods_environment.json` option (4-3-stable)

MartinFlores751:
Quick link to the issue: #7392.

This PR adds `irods_connection_pool_refresh_time_in_seconds` to the configuration upgrade script, defaulting to 300 if no previous value is found.

---

Needs to be tested.

Needs a companion PR to update documentation.

### DIFF
--- a/scripts/irods/upgrade_configuration.py
+++ b/scripts/irods/upgrade_configuration.py
@@ -465,7 +465,8 @@ def convert_legacy_configuration_to_json(irods_config):
             'irods_transfer_buffer_size_for_parallel_transfer_in_megabytes': server_config_v2['advanced_settings']['transfer_buffer_size_for_parallel_transfer_in_megabytes'],
             "irods_ssl_verify_server": os.getenv("irodsSSLVerifyServer", 'hostname'),
             "irods_ssl_ca_certificate_file": os.getenv("irodsSSLCACertificateFile", ''),
-            "irods_ssl_ca_certificate_path": os.getenv("irodsSSLCACertificatePath", '')
+            "irods_ssl_ca_certificate_path": os.getenv("irodsSSLCACertificatePath", ''),
+            "irods_connection_pool_refresh_time_in_seconds": int(legacy_irods_environment.get('irodsConnectionPoolRefreshTime', 300))
             }
 
     #optional irods environment keys


MartinFlores751 (commented):
Update the commit message, there are some errors.

scripts/irods/upgrade_configuration.py
```diff
@@ -465,7 +465,8 @@ def convert_legacy_configuration_to_json(irods_config):
             'irods_transfer_buffer_size_for_parallel_transfer_in_megabytes': server_config_v2['advanced_settings']['transfer_buffer_size_for_parallel_transfer_in_megabytes'],
             "irods_ssl_verify_server": os.getenv("irodsSSLVerifyServer", 'hostname'),
             "irods_ssl_ca_certificate_file": os.getenv("irodsSSLCACertificateFile", ''),
-            "irods_ssl_ca_certificate_path": os.getenv("irodsSSLCACertificatePath", '')
+            "irods_ssl_ca_certificate_path": os.getenv("irodsSSLCACertificatePath", ''),
+            "irods_connection_pool_refresh_time_in_seconds": int(legacy_irods_environment.get('irodsConnectionPoolRefreshTime', 300))
```
korydraughn, line 469: Hmm, not sure this is quite right.

I think you want to update this as if you're working on preparing a 4.3.1 release. That means, the upgrade logic needs to apply this change based on the active schema version number.

The problem with what I just said is that the schema version didn't change between 4.3.0 and 4.3.1, meaning there's no place to do this work. You can see what I mean by looking at `schema_version` in version.json.dist.in for each release.

Hmm, need to think about this a little more.

MartinFlores751: Pointers would be appreciated. Would this change then be in `previous_version` or some other dictionary (assuming the change is in a dictionary)?

See: https://github.com/irods/irods/blob/eebc7854dc5ef54eefedf8f0164e2621a02f6554/scripts/irods/upgrade_configuration.py#L617-L627

---

I think I see what you're getting at. The current change would be right if we had it there the first time. Do we have a part of the update code that checks `irods_version` for updates?

korydraughn: After some discussion, we've agreed to do the following:
- Remove "(Optional)" from the configs in irods/irods_docs
- Confirm upgrades from iRODS 4.2.x to 4.3.3 do not run into the schema validation issue

We are not going to bump the schema version for 4.3 (remains at v4). We expect iRODS 5 to ship with a schema version of v5.

MartinFlores751:
Tested and reproduce the issue: https://github.com/irods/irods/issues/7392#issuecomment-2669579446

Closing.

PR #8211, opened 2025-02-19
Revert #8195, add error handling to legacy parallel transfer (4-3-stable)

alanking:
In service of #4984 

Partially reverts #8195

`SYS_SOCK_READ_ERR` is the closest thing I could find to what I wanted from an existing error code; open to other suggestions. Couldn't find anything related to SSL/TLS that came close. Maybe we need a new error code.

Unit tests passed. Core tests running. Will mark Ready for Review once they're passing.

### DIFF
--- a/server/api/src/rsDataGet.cpp
+++ b/server/api/src/rsDataGet.cpp
@@ -2,7 +2,6 @@
 
 #include "irods/dataGet.h"
 #include "irods/dataObjInpOut.h"
-#include "irods/irods_client_server_negotiation.hpp"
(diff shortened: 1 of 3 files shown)


trel:
> So, I think I'm actually satisfied with the current choice of error.

I'm fine with that.

korydraughn:
Same here. Squash to taste.

korydraughn (approved):
Pound it.

alanking:
#'d, mergin

Issue #7953, opened 2024-09-11
Make `rsDataObjUnlink` warning less noisy


alanking:
- [x] main
- [x] 4-3-stable

---

## Feature

`rsDataObjUnlink` prints something like the following warning message when a replica is being targeted that is not in an iRODS resource vault:

> Replica is not in a vault. Unregistering replica and leaving it on disk as-is [data_object=/tempZone/home/rods/ufs/dir09/file08, physical_object=/data/ufs/dir09/file08, vault_path=/var/lib/irods/Vault].

Here is the code:
https://github.com/irods/irods/blob/bc2a9bf0d8d3587d1aad8bf6a9b3db650e0d6944/server/api/src/rsDataObjUnlink.cpp#L543-L551

The `oprType` is being set to `UNREG_OPR`. This happens even if the `oprType` is already `UNREG_OPR`.

This is observed when using the python-irodsclient's `DataObjectManager.unregister` method:
https://github.com/irods/python-irodsclient/blob/527417da41a997198e514e0e211b5422585a97a0/irods/manager/data_object_manager.py#L620-L628

As you can see, `OPR_TYPE_KW` is being set in the `**options` which is then used in `DataObjectManager.unlink` to set the `oprType` in the API call:
https://github.com/irods/python-irodsclient/blob/527417da41a997198e514e0e211b5422585a97a0/irods/manager/data_object_manager.py#L597-L611

Marking as a feature / enhancement because it's not really a bug - just annoying. Willing to upgrade to bug.

trel:
There's no change in behavior that you're proposing... right?

So... setting it to `logger::api::debug` is perhaps sufficient to not be as annoying?

alanking:
No, we want to show the warning in the event that we attempt to unlink a replica which is not in an iRODS resource vault because we are changing the behavior requested by the client to avoid data loss (i.e. we are changing the operation to an unregister from an unlink).

I am proposing that we do not show this warning message when `UNREG_OPR` is already set because we are already unregistering the replica at the request of the client and the warning message is just noise in that case.

PR #8230, opened 2025-02-21
[#7953] Do not log message unless oprType is changed to UNREG_OPR (4-3-stable)

korydraughn:
I've confirmed at the bench that the code change works as intended. No test necessary since it's just logging.

The message is logged if the oprType is actually changed (i.e. `irm` vs `iunreg`).

### DIFF
--- a/server/api/src/rsDataObjUnlink.cpp
+++ b/server/api/src/rsDataObjUnlink.cpp
@@ -540,7 +540,7 @@ int dataObjUnlinkS(rsComm_t* rsComm,
                 return err.code();
             }
 
-            if (!has_prefix(dataObjInfo->filePath, vault_path.data())) {
+            if (UNREG_OPR != dataObjUnlinkInp->oprType && !has_prefix(dataObjInfo->filePath, vault_path.data())) {
                 dataObjUnlinkInp->oprType = UNREG_OPR;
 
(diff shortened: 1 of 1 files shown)


server/api/src/rsDataObjUnlink.cpp
```diff
@@ -541,13 +541,17 @@ int dataObjUnlinkS(rsComm_t* rsComm,
             }
 
             if (!has_prefix(dataObjInfo->filePath, vault_path.data())) {
```
alanking, line 543: We could just skip the whole block by checking here:
```suggestion
            if (!has_prefix(dataObjInfo->filePath, vault_path.data()) && UNREG_OPR != dataObjUnlinkInp->oprType) {
```
No need to set the oprType if it's already the right value. That said, what you have is fine.

alanking (commented):
Suggestion is optional. Up to you.

server/api/src/rsDataObjUnlink.cpp
korydraughn: Updated.

korydraughn:
Squashed and verified change does the right thing.

alanking (approved):
Looks good. # it

korydraughn:
pounded

Issue #6843, opened 2023-01-13
Deprecate acAclPolicy{msiAclPolicy('STANDARD');} (strictACL as a knob)


trel:
## Deprecation

`acAclPolicy{}` was set to `STRICT` by default in 2012 (https://github.com/irods/irods/commit/043d51f0f64761a8a87c9a5ecbf83d8db134436a)

`acAclPolicy {msiAclPolicy("STRICT"); }`

https://github.com/irods/irods/blob/7a8ea89af09a3ab222f8009825dda0b207328735/packaging/core.re.template#L61-L87

There is little/no reason to keep `STANDARD` as an option - it could allow surprising access to data objects within the system and is much harder to reason about for users and administrators.

Deprecating the policy/rule, but keeping the effect to `STRICT` will make future iRODS systems safer and less-likely to be configured incorrectly.

A downside of deprecation -> removal ...
- existing 'STANDARD' deployments will need to inspect their policies before upgrading

Also related - `iadmin modzonecollacl`.

---

Please share use cases for continued use/availability of `acAclPolicy{'STANDARD'}`.

trel:
related to https://github.com/irods/irods/issues/813

trel:
renaming to 'deprecate'.   will create a new issue for 'removal' for 5.0.0.

trel:
Removal is now https://github.com/irods/irods/issues/7985

korydraughn:
The steps to deprecate this are:
- Update documentation for rule in core.re to include words about it being deprecated
- Consider logging a warning on server startup if `STRICT` is not used

trel:
`core.re.template` should get the update

and yes on the warning to the log.

korydraughn:
The effort to log a warning for this has led to the log file becoming very noisy, so we've decided to make this strictly a documentation effort.

The new steps are:
- Document deprecation in the core.re.template
- Document deprecation in the release notes

korydraughn:
This will be covered in the release notes and is independent of this issue.

Closing as completed.

PR #8231, opened 2025-02-21
[#6843] Deprecate modification of ACL policy (4-3-stable)

korydraughn:
The presented implementation logs a warning about setting the ACL policy to something other than STRICT. The problem is that it happens on every fork of an agent due to how 4.3 works. That makes the log file very noisy, especially with the delay server migration check happening every 5 seconds.

As of right now, it seems the solution for the noisy log is to use more IPC. I'd really like to avoid that, but I'm not sure that I can. So, here's some ideas I'm thinking about:
- Use a boost interprocess mutex as a flag
- Or, use a file as a flag

The more I think about this, the more I'm leaning towards removing the edits to the C++ and letting this be purely a job for core.re.template and the release notes.

Thoughts?

### DIFF
--- a/packaging/core.re.template
+++ b/packaging/core.re.template
@@ -59,14 +59,20 @@ acRenameLocalZone(*oldZone, *newZone) {
 #acGetUserByDN(*arg,*OUT) {msiExecCmd("t","*arg","null","null","null",*OUT); }
 #
 # --------------------------------------------------------------------------
+# IMPORTANT: Setting the ACL policy to anything other than "STRICT"
+# is deprecated and will be disallowed in later versions of iRODS. It
+# is highly recommended that deployments relying on an ACL policy other
+# than "STRICT" develop a migration plan before upgrading to later
+# versions of iRODS.
+#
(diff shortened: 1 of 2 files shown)


korydraughn:
I've removed the C++ changes. All we have to do now is agree on the wording in core.re.template.

trel:
I think the wording is fine as you have it.   We need to have something similar/identical in the irods_docs wherever it is most helpful/relevant.

korydraughn:
Agreed. I'm looking around for a good spot for that as well.

alanking (approved):
Wording seems good to me.

korydraughn:
Need to deprecate the MSI as well since there's no point in providing it in iRODS 5.

korydraughn:
Added a deprecation directive to the doxygen for the MSI.

Let me know if the wording needs a tweak.

trel:
seems good.

korydraughn:
Ok. Squashing and pounding.

korydraughn:
pounded

alanking (approved):

Issue #7752, opened 2024-05-22
`json_validation.py`: reference to undefined variable `name`


SwooshyCueb:
- [x] 4-3-stable (no json_validation.py in main)

---

This method in `json_validation.py` references an undefined variable `name`:
https://github.com/irods/irods/blob/e4b4e8a7266360bb38ee97d7f2eda69249567dff/scripts/irods/json_validation.py#L99-L105

Apparently it's been this way for at least eight years.

trel:
oh, perfect.

PR #8234, opened 2025-02-21
[#7752] Remove `name` from exception (4-3-stable)

MartinFlores751:
Issue quick link: #7752.

---

`name` is not in scope at the point the exception is created.


alanking (approved):
\# it

MartinFlores751:
Pounded.

PR #8236, opened 2025-02-21
[7986] Add SSL info to imiscsvrinfo (main)

FifthPotato:
Leaving as draft for now. Still needs:
- Freeing of the bytesbuf buffer
- Determine what we actually want in the output
- Exception handling
- A test

For the icommands: https://github.com/irods/irods_client_icommands/pull/539

### DIFF
--- a/scripts/irods/test/test_imiscsvrinfo.py
+++ b/scripts/irods/test/test_imiscsvrinfo.py
@@ -1,0 +1,30 @@
+import unittest
+
+from . import session
+from .. import test
+
+class Test_Imiscsvrinfo(session.make_sessions_mixin([], [('alice', 'apass')]), unittest.TestCase):
+
+    def setUp(self):
+        super(Test_Imiscsvrinfo, self).setUp()
+        self.user = self.user_sessions[0]
+
+    def tearDown(self):
+        super(Test_Imiscsvrinfo, self).tearDown()
+
+    @unittest.skipUnless(test.settings.USE_SSL, 'Test case assumes use of SSL/TLS')
+    def test_imiscsvrinfo_prints_ssl_info_with_ssl_enabled__issue_7986(self):
+        _, out, _ = self.user.assert_icommand(['imiscsvrinfo'], 'STDOUT')
+        expected_outputs = ['enabled: true', 'issuer_name', 'not_after', 'not_before', 'public_key', 'subject_alternative_names', 'subject_name']
+        for output in expected_outputs:
+            self.assertIn(output, out);
+
+    @unittest.skipIf(test.settings.USE_SSL, 'Test case assumes absence of SSL/TLS')
+    def test_imiscsvrinfo_prints_ssl_info_with_ssl_disabled__issue_7986(self):
+
+        # Attributes should be missing with SSL disabled
+        expected_missing = ['enabled: true', 'issuer_name', 'not_after', 'not_before', 'public_key', 'subject_alternative_names', 'subject_name']
+        _, out, _ = self.user.assert_icommand(['imiscsvrinfo'], 'STDOUT', 'enabled: false')
+
+        for output in expected_missing:
+            self.assertNotIn(output, out);
--- a/server/api/src/rsGetMiscSvrInfo.cpp
+++ b/server/api/src/rsGetMiscSvrInfo.cpp
@@ -6,11 +6,29 @@
 
 /* script generated code */
 #include "irods/getMiscSvrInfo.h"
+#include "irods/irods_client_server_negotiation.hpp"
 #include "irods/irods_server_properties.hpp"
 #include "irods/irods_log.hpp"
+#include "irods/irods_logger.hpp"
 #include "irods/rodsVersion.h"
 #include "irods/miscServerFunct.hpp"
 #include "irods/rsGetMiscSvrInfo.hpp"
+
+#include <nlohmann/json.hpp>
+#include <openssl/asn1.h>
+#include <openssl/bio.h>
+#include <openssl/evp.h>
+#include <openssl/objects.h>
+#include <openssl/ssl.h>
+#include <openssl/x509.h>
+#include <openssl/x509v3.h>
+
+#include <array>
+#include <cstdlib>
+#include <cstring>
+#include <vector>
+
+using log_api = irods::experimental::log::api;
 
 int
 rsGetMiscSvrInfo( rsComm_t *rsComm, miscSvrInfo_t **outSvrInfo ) {
@@ -57,6 +75,168 @@ rsGetMiscSvrInfo( rsComm_t *rsComm, miscSvrInfo_t **outSvrInfo ) {
     snprintf( myOutSvrInfo->rodsZone, NAME_LEN, "%s", zone_name.c_str() );
     if ( ( tmpStr = getenv( SERVER_BOOT_TIME ) ) != NULL ) {
         myOutSvrInfo->serverBootTime = atoi( tmpStr );
+    }
+
+    nlohmann::json certinfo_json;
+    try {
+        if (irods::CS_NEG_USE_SSL == rsComm->negotiation_results && rsComm->ssl_ctx) {
+            certinfo_json["ssl_enabled"] = true;
+
+            const auto* cert = SSL_CTX_get0_certificate(rsComm->ssl_ctx);
+            if (cert) {
+                std::tm tmp_tm;
+                std::array<char, 40> timebuf{};
+                // Set notAfter
+                auto* asn_time_tmp = X509_get0_notAfter(cert);
+                if (asn_time_tmp) {
+                    const auto success = ASN1_TIME_to_tm(asn_time_tmp, &tmp_tm);
+                    if (success) {
+                        std::strftime(&timebuf[0], 40, "%F %T %Z", &tmp_tm);
+                        certinfo_json["not_after"] = std::string(timebuf.data());
+                    }
+                    else {
+                        log_api::warn("{}: Failure in ASN1_TIME_to_tm for not_after: time parse failure.", __func__);
+                    }
+                }
+                else {
+                    log_api::error("{}: Failure in X509_get0_notAfter: nullptr returned.", __func__);
+                }
+
+                // Set notBefore
+                asn_time_tmp = X509_get0_notBefore(cert);
+                if (asn_time_tmp) {
+                    const auto success = ASN1_TIME_to_tm(asn_time_tmp, &tmp_tm);
+                    if (success) {
+                        std::strftime(&timebuf[0], 40, "%F %T %Z", &tmp_tm);
+                        certinfo_json["not_before"] = std::string(timebuf.data());
+                    }
+                    else {
+                        log_api::warn("{}: Failure in ASN1_TIME_to_tm for not_before: time parse failure.", __func__);
+                    }
+                }
+                else {
+                    log_api::error("{}: Failure in X509_get0_notBefore: nullptr returned.", __func__);
+                }
+
+                // Set pubkey
+                auto* bio = BIO_new(BIO_s_mem());
+                irods::at_scope_exit free_bio{[&bio] { BIO_free(bio); }};
+                if (bio) {
+                    char* biodata;
+                    long biolen;
+                    const auto* pubkey = X509_get0_pubkey(cert);
+                    if (pubkey) {
+                        const auto ec = EVP_PKEY_print_public(bio, pubkey, 0, nullptr);
+                        if (ec > 0) {
+                            biolen = BIO_get_mem_data(bio, &biodata);
+                            if (!biodata) {
+                                log_api::error("{}: Failure in BIO_get_mem_data: nullptr returned.", __func__);
+                            }
+                            else if (std::isspace(biodata[biolen - 1])) {
+                                // strip off trailing newline
+                                certinfo_json["public_key"] = std::string(biodata, biolen - 1);
+                            }
+                            else {
+                                certinfo_json["public_key"] = std::string(biodata, biolen);
+                            }
+                        }
+                        else {
+                            log_api::warn("{}: Failure in EVP_PKEY_print_public with code [{}].", __func__, ec);
+                        }
+                    }
+                    else {
+                        log_api::error("{}: Failure in X509_get0_pubkey: nullptr returned.", __func__);
+                    }
+                    // Set issuer name
+                    BIO_reset(bio);
+                    const auto* issuer_name = X509_get_issuer_name(cert);
+                    if (issuer_name) {
+                        if (X509_NAME_print(bio, issuer_name, 0)) {
+                            biolen = BIO_get_mem_data(bio, &biodata);
+                            certinfo_json["issuer_name"] = std::string(biodata, biolen);
+                        }
+                        else {
+                            log_api::warn("{}: Failure in X509_NAME_print for issuer_name.", __func__);
+                        }
+                    }
+                    else {
+                        log_api::error("{}: Failure in X509_get_issuer_name: nullptr returned.", __func__);
+                    }
+
+                    // Set subject name
+                    BIO_reset(bio);
+                    const auto* subject_name = X509_get_subject_name(cert);
+                    if (subject_name) {
+                        if (X509_NAME_print(bio, subject_name, 0)) {
+                            biolen = BIO_get_mem_data(bio, &biodata);
+                            certinfo_json["subject_name"] = std::string(biodata, biolen);
+                        }
+                        else {
+                            log_api::warn("{}: Failure in X509_NAME_print for subject_name.", __func__);
+                        }
+                    }
+                    else {
+                        log_api::error("{}: Failure in X509_get_subject_name: nullptr returned.", __func__);
+                    }
+                }
+                else {
+                    log_api::error("{}: Failed to create new BIO.", __func__);
+                }
+
+                // Set signature algorithm
+                certinfo_json["signature_algorithm"] = std::string(OBJ_nid2ln(X509_get_signature_nid(cert)));
+
+                // Set subject alternative name
+                GENERAL_NAMES* gen_names =
+                    static_cast<GENERAL_NAMES*>(X509_get_ext_d2i(cert, NID_subject_alt_name, nullptr, nullptr));
+                std::vector<std::string> san_vector;
+                if (gen_names) {
+                    unsigned char* tempStr;
+                    for (int i = 0; i < sk_GENERAL_NAME_num(gen_names); i++) {
+                        GENERAL_NAME* gen_name = sk_GENERAL_NAME_value(gen_names, i);
+                        if (gen_name->type != GEN_DNS) {
+                            log_api::debug("{}: Non-dNSName Subject Alternative Name found with type [{}].",
+                                           __func__,
+                                           gen_name->type);
+                            // TODO(#8244): support more than just dNSName
+                            continue;
+                        }
+                        int len = ASN1_STRING_to_UTF8(&tempStr, gen_name->d.dNSName);
+                        if (len > -1) {
+                            auto san_str = std::string(reinterpret_cast<char*>(tempStr), len);
+                            san_str.insert(0, "DNS:");
+                            san_vector.push_back(san_str);
+                        }
+                        else {
+                            log_api::warn("{}: Failure in ASN1_STRING_to_UTF8 with code [{}].", __func__, len);
+                        }
+                    }
+                }
+                if (!san_vector.empty()) {
+                    certinfo_json["subject_alternative_names"] = san_vector;
+                }
+                else {
+                    log_api::debug("{}: Subject alternative names has no entries.", __func__);
+                }
+            }
+            else {
+                log_api::error("{}: Failure in SSL_CTX_get0_certificate: nullptr returned.", __func__);
+            }
+        }
+        else {
+            certinfo_json["ssl_enabled"] = false;
+        }
+
+        const auto json_str = certinfo_json.dump();
+        myOutSvrInfo->certinfo.buf = malloc(json_str.size());
+        myOutSvrInfo->certinfo.len = json_str.size();
+        memcpy(myOutSvrInfo->certinfo.buf, json_str.c_str(), myOutSvrInfo->certinfo.len);
+    }
+    catch (const nlohmann::json::exception& e) {
+        log_api::error("{}: Failed to serialize SSL/TLS info - {}", __func__, e.what());
+    }
+    catch (const std::exception& e) {
+        log_api::error("{}: Failed while fetching SSL/TLS info - {}", __func__, e.what());
     }
 
     return 0;
--- a/lib/api/include/irods/apiTable.hpp
+++ b/lib/api/include/irods/apiTable.hpp
@@ -323,7 +323,7 @@ static irods::apidef_t client_api_table_inp[] = {
         GET_MISC_SVR_INFO_AN, RODS_API_VERSION, NO_USER_AUTH, NO_USER_AUTH,
         NULL, 0, "MiscSvrInfo_PI", 0,
         boost::any(std::function<int(rsComm_t*,miscSvrInfo_t**)>(RS_GET_MISC_SVR_INFO)),
-        "api_get_misc_svr_info", irods::clearInStruct_noop, irods::clearOutStruct_noop,
+        "api_get_misc_svr_info", irods::clearInStruct_noop, clearMiscSvrInfo,
         (funcPtr)CALL_MISCSVRINFOOUT
     },
     {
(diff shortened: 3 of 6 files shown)


server/api/src/rsGetMiscSvrInfo.cpp
```diff
@@ -59,5 +63,42 @@ rsGetMiscSvrInfo( rsComm_t *rsComm, miscSvrInfo_t **outSvrInfo ) {
         myOutSvrInfo->serverBootTime = atoi( tmpStr );
     }
 
+    nlohmann::json certinfo_json;
+    if(irods::CS_NEG_USE_SSL == rsComm->negotiation_results && rsComm->ssl_ctx) {
+        certinfo_json["ssl_enabled"] = true;
+
+        const auto* cert = SSL_CTX_get0_certificate(rsComm->ssl_ctx);
```
korydraughn, line 70: Can this return a nullptr?

server/api/src/rsGetMiscSvrInfo.cpp
```diff
@@ -59,5 +63,42 @@ rsGetMiscSvrInfo( rsComm_t *rsComm, miscSvrInfo_t **outSvrInfo ) {
...
+        std::tm tmp_tm;
+        char timebuf[40] = { 0 };
```
korydraughn, line 72: This could be rewritten as:
```cpp
char timebuf[40]{};
```
Or ...
```cpp
#include <array>
...
std::array<char, 40> timebuf{};
// This gives you the following:
// - iterators
// - .size()
// - .data() for C API compatibility
// - safety checks in debug mode
```

server/api/src/rsGetMiscSvrInfo.cpp
```diff
@@ -59,5 +63,42 @@ rsGetMiscSvrInfo( rsComm_t *rsComm, miscSvrInfo_t **outSvrInfo ) {
...
+        // set notAfter
+        auto asn_time_tmp = X509_get0_notAfter(cert);
+        ASN1_TIME_to_tm(asn_time_tmp, &tmp_tm);
+        std::strftime(&timebuf[0], 40, "%F %T %Z", &tmp_tm);
```
korydraughn, line 76: Did you come up with this or did it come from the OpenSSL devs (as an example)?

Just curious if the OpenSSL people recommended this.

server/api/src/rsGetMiscSvrInfo.cpp
```diff
@@ -59,5 +63,42 @@ rsGetMiscSvrInfo( rsComm_t *rsComm, miscSvrInfo_t **outSvrInfo ) {
...
+        certinfo_json["notAfter"] = std::string(timebuf);
```
korydraughn, line 77: Is `timebuf` guaranteed to be null terminated?

korydraughn (commented):
Very nice.

I think we're missing the clear function for the API table entry / definition. That lives in rcMisc.h/.cpp. There are many examples in those two files.

Please open a PR for the icommand and link it to this PR so we know it's a companion of this PR. Include words stating that so it's clear to everyone.

Also, post the formatted JSON output in this PR or the icommands PR so it's easier to discuss what information to include in the output.

server/api/src/rsGetMiscSvrInfo.cpp
FifthPotato: I was mostly relying on the fact the buffer size is 40 and 0-initialized. Wasn't sure how much craziness could happen, but I couldn't see how `YYYY-MM-DD HH:MM:SS TZ` could exceed that...maybe if somehow there were 2 bytes per character? I could make the buffer even bigger to avoid that.

server/api/src/rsGetMiscSvrInfo.cpp
FifthPotato: I came up with this after reading the OpenSSL docs and cobbling that together with whatever knowledge I had regarding C. I am open to suggestions though, if this is ugly or awkward.

server/api/src/rsGetMiscSvrInfo.cpp
korydraughn: No. I just wanted to confirm the thought process. I agree that 40 bytes is likely enough.

It's fine as is.

server/api/src/rsGetMiscSvrInfo.cpp
FifthPotato: Yes, good catch.

FifthPotato: Well, per the function spec anyways-- I'm not sure what could cause iRODS to do this (somehow SSL-connected without a cert?) but I think adding the null check is good anyways.

korydraughn: Agreed. I don't know that we truly need to worry about it, but having the check there makes the system safer.

server/api/src/rsGetMiscSvrInfo.cpp
korydraughn: No, it's fine. As long as we have tests, it'll be fine.

FifthPotato:
Opened https://github.com/irods/irods_client_icommands/pull/539 for the icommands.

server/api/src/rsGetMiscSvrInfo.cpp
```diff
@@ -59,5 +63,44 @@ rsGetMiscSvrInfo( rsComm_t *rsComm, miscSvrInfo_t **outSvrInfo ) {
...
+
+        const auto* cert = SSL_CTX_get0_certificate(rsComm->ssl_ctx);
+        if(cert) {
+            std::tm tmp_tm;
+            char timebuf[40] = { 0 };
+            // set notAfter
+            auto asn_time_tmp = X509_get0_notAfter(cert);
+            ASN1_TIME_to_tm(asn_time_tmp, &tmp_tm);
+            std::strftime(&timebuf[0], 40, "%F %T %Z", &tmp_tm);
+            certinfo_json["notAfter"] = std::string(timebuf);
```
korydraughn, line 78: notAfter -> not_after

server/api/src/rsGetMiscSvrInfo.cpp
```diff
@@ -59,5 +63,44 @@ rsGetMiscSvrInfo( rsComm_t *rsComm, miscSvrInfo_t **outSvrInfo ) {
...
+
+            // set notBefore
+            asn_time_tmp = X509_get0_notBefore(cert);
+            ASN1_TIME_to_tm(asn_time_tmp, &tmp_tm);
+            std::strftime(&timebuf[0], 40, "%F %T %Z", &tmp_tm);
+            certinfo_json["notBefore"] = std::string(timebuf);
```
korydraughn, line 84: notBefore -> not_before

server/api/src/rsGetMiscSvrInfo.cpp
```diff
@@ -59,5 +63,44 @@ rsGetMiscSvrInfo( rsComm_t *rsComm, miscSvrInfo_t **outSvrInfo ) {
...
+
+            // set pubkey
+            auto* bio = BIO_new(BIO_s_mem());
+            irods::at_scope_exit free_bio {[&bio] { BIO_free(bio); } };
+            char* biodata;
+            const auto* pubkey = X509_get0_pubkey(cert);
+            EVP_PKEY_print_public(bio, pubkey, 0, NULL);
+            const auto biolen = BIO_get_mem_data(bio, &biodata);
+            certinfo_json["pubkey"] = std::string(biodata, biolen);
```
korydraughn, line 93: pubkey -> public_key

server/api/src/rsGetMiscSvrInfo.cpp
```diff
@@ -59,5 +64,44 @@ rsGetMiscSvrInfo( rsComm_t *rsComm, miscSvrInfo_t **outSvrInfo ) {
...
+            // set pubkey
+            auto* bio = BIO_new(BIO_s_mem());
+            irods::at_scope_exit free_bio {[&bio] { BIO_free(bio); } };
+            char* biodata;
+            const auto* pubkey = X509_get0_pubkey(cert);
+            EVP_PKEY_print_public(bio, pubkey, 0, NULL);
+            const auto biolen = BIO_get_mem_data(bio, &biodata);
+            certinfo_json["public_key"] = std::string(biodata, biolen);
+        }
+
```
korydraughn, line 96: Remove empty line.

server/api/src/rsGetMiscSvrInfo.cpp
```diff
@@ -32,6 +36,7 @@ rsGetMiscSvrInfo( rsComm_t *rsComm, miscSvrInfo_t **outSvrInfo ) {
         return ret.code();
     }
 
+
```
korydraughn, line 39: Remove empty line.

server/api/src/rsGetMiscSvrInfo.cpp
```diff
@@ -6,12 +6,16 @@
...
 #include "irods/getMiscSvrInfo.h"
+#include "irods/irods_client_server_negotiation.hpp"
 #include "irods/irods_server_properties.hpp"
 #include "irods/irods_log.hpp"
 #include "irods/rodsVersion.h"
 #include "irods/miscServerFunct.hpp"
 #include "irods/rsGetMiscSvrInfo.hpp"
 
+#include <array>
+#include <nlohmann/json.hpp>
```
korydraughn, line 17:
```cpp
#include <nlohmann/json.hpp>

#include <array>
```

server/api/src/rsGetMiscSvrInfo.cpp
```diff
@@ -59,5 +64,61 @@ rsGetMiscSvrInfo( rsComm_t *rsComm, miscSvrInfo_t **outSvrInfo ) {
...
+            certinfo_json["not_after"] = std::string(timebuf.data());
+
+            // set notBefore
+            asn_time_tmp = X509_get0_notBefore(cert);
+            ASN1_TIME_to_tm(asn_time_tmp, &tmp_tm);
+            std::strftime(&timebuf[0], 40, "%F %T %Z", &tmp_tm);
+            certinfo_json["not_before"] = std::string(timebuf.data());
+
+            // set pubkey
+            auto* bio = BIO_new(BIO_s_mem());
```
korydraughn, line 88: Consider if the pointers need to be checked for null before usage.

korydraughn, line 88: Do any of these SSL pointers need to be cleaned up / free'd?

korydraughn (commented):
For the test, I believe you can enable them using the skipIf/skipUnless annotations and the test.settings.USE_SSL flag.
https://github.com/irods/irods/blob/6dbbc1fe8c36704736dc679b751d529383f4350c/scripts/irods/test/settings.py#L7

I'm sure there's several examples throughout the codebase of this.

I think two tests + a maybe are what we need to cover this new feature. The tests are as follows:
- One to confirm the icmd reports when no SSL is used
- One to confirm the icmd reports when SSL is used
- Maybe one to confirm behavior when the SSL is misconfigured

Thoughts?

server/api/src/rsGetMiscSvrInfo.cpp
FifthPotato: As far as I can tell, all the functions I call return pointers that must not be freed. The only exception is the BIO, which I free with the `at_scope_exit`.

korydraughn:
What's the status of this PR?

server/api/src/rsGetMiscSvrInfo.cpp
```diff
@@ -65,100 +65,104 @@ rsGetMiscSvrInfo( rsComm_t *rsComm, miscSvrInfo_t **outSvrInfo ) {
...
-    const auto json_str = certinfo_json.dump();
-    myOutSvrInfo->certinfo.buf = malloc(json_str.size());
-    myOutSvrInfo->certinfo.len = json_str.size();
-    memcpy(myOutSvrInfo->certinfo.buf, json_str.c_str(), myOutSvrInfo->certinfo.len);
+        const auto json_str = certinfo_json.dump();
+        myOutSvrInfo->certinfo.buf = malloc(json_str.size());
+        myOutSvrInfo->certinfo.len = json_str.size();
+        memcpy(myOutSvrInfo->certinfo.buf, json_str.c_str(), myOutSvrInfo->certinfo.len);
+        } catch(const nlohmann::json::exception& e) {
+            irods::log(LOG_ERROR, fmt::format("[{}] Failed to serialize SSL/TLS data - {}", __FUNCTION__, e.what()));
```
korydraughn, line 164: `__FUNCTION__` -> `__func__`

Prefer `__func__` since it's defined by the standard.

korydraughn, line 164: Prefer the new logger over rodsLog and irods::log.
```cpp

log_api::error("{}: Failed to serialize SSL/TLS data: {}", __func__, e.what());
```

server/api/src/rsGetMiscSvrInfo.cpp
```diff
@@ -65,100 +65,104 @@ rsGetMiscSvrInfo( rsComm_t *rsComm, miscSvrInfo_t **outSvrInfo ) {
...
-                    } else {
-                        certinfo_json["public_key"] = std::string(biodata, biolen);
+                // Set pubkey
+                auto* bio = BIO_new(BIO_s_mem());
+                irods::at_scope_exit free_bio {[&bio] { BIO_free(bio); } };
+                char* biodata;
+                long biolen;
+                const auto* pubkey = X509_get0_pubkey(cert);
+                if(pubkey) {
+                    if(EVP_PKEY_print_public(bio, pubkey, 0, NULL) > 0) {
```
korydraughn, line 99: Replace use of `NULL` with `nullptr`.

server/api/src/rsGetMiscSvrInfo.cpp
```diff
@@ -65,100 +65,104 @@ rsGetMiscSvrInfo( rsComm_t *rsComm, miscSvrInfo_t **outSvrInfo ) {
...
+            if(cert) {
+                std::tm tmp_tm;
+                std::array<char, 40> timebuf{};
+                // Set notAfter
+                auto* asn_time_tmp = X509_get0_notAfter(cert);
+                if(asn_time_tmp) {
+                    ASN1_TIME_to_tm(asn_time_tmp, &tmp_tm);
+                    std::strftime(&timebuf[0], 40, "%F %T %Z", &tmp_tm);
+                    certinfo_json["not_after"] = std::string(timebuf.data());
+                }
```
korydraughn, line 82: Could add else-blocks and log a debug/trace message for things that weren't found?

Not required, but could be helpful in tracking down bugs if any exist.

server/api/src/rsGetMiscSvrInfo.cpp
```diff
@@ -65,100 +65,104 @@ rsGetMiscSvrInfo( rsComm_t *rsComm, miscSvrInfo_t **outSvrInfo ) {
...
+                        }
+                        int len = ASN1_STRING_to_UTF8(&tempStr, gen_name->d.dNSName);
+                        if(len > -1) {
+                            auto san_str = std::string(reinterpret_cast<char*>(tempStr), len);
+                            san_str.insert(0, "DNS:");
+                            san_vector.push_back(san_str);
+                        }
                     }
                 }
+                if(san_vector.size() > 0) {
```
korydraughn, line 151: Could be `if (!san_vector.empty()) {` to make it even clearer to the reader.

server/api/src/rsGetMiscSvrInfo.cpp
```diff
@@ -65,100 +65,104 @@ rsGetMiscSvrInfo( rsComm_t *rsComm, miscSvrInfo_t **outSvrInfo ) {
...
-                    if(len > -1) {
-                        auto san_str = std::string(reinterpret_cast<char*>(tempStr), len);
-                        san_str.insert(0, "DNS:");
-                        san_vector.push_back(san_str);
+                // Set signature algorithm
+                certinfo_json["signature_algorithm"] = std::string(OBJ_nid2ln(X509_get_signature_nid(cert)));
+
+                // Set subject alternative name
+                GENERAL_NAMES* gen_names = static_cast<GENERAL_NAMES*>(X509_get_ext_d2i(cert, NID_subject_alt_name, nullptr, nullptr));
+                std::vector<std::string> san_vector;
```
korydraughn, line 134: Missing `#include <vector>`

server/api/src/rsGetMiscSvrInfo.cpp
```diff
@@ -65,100 +65,104 @@ rsGetMiscSvrInfo( rsComm_t *rsComm, miscSvrInfo_t **outSvrInfo ) {
...
-            if(asn_time_tmp) {
-                ASN1_TIME_to_tm(asn_time_tmp, &tmp_tm);
-                std::strftime(&timebuf[0], 40, "%F %T %Z", &tmp_tm);
-                certinfo_json["not_after"] = std::string(timebuf.data());
-            }
+    try {
+        if(irods::CS_NEG_USE_SSL == rsComm->negotiation_results && rsComm->ssl_ctx) {
+            certinfo_json["ssl_enabled"] = true;
+
+            const auto* cert = SSL_CTX_get0_certificate(rsComm->ssl_ctx);
```
korydraughn, line 72: Missing headers for these SSL functions and types.

server/api/src/rsGetMiscSvrInfo.cpp
```diff
@@ -65,100 +65,104 @@ rsGetMiscSvrInfo( rsComm_t *rsComm, miscSvrInfo_t **outSvrInfo ) {
...
-    } else {
-        certinfo_json["ssl_enabled"] = false;
-    }
 
-    const auto json_str = certinfo_json.dump();
-    myOutSvrInfo->certinfo.buf = malloc(json_str.size());
-    myOutSvrInfo->certinfo.len = json_str.size();
-    memcpy(myOutSvrInfo->certinfo.buf, json_str.c_str(), myOutSvrInfo->certinfo.len);
+        const auto json_str = certinfo_json.dump();
+        myOutSvrInfo->certinfo.buf = malloc(json_str.size());
```
korydraughn, line 160: Missing `#include <cstdlib>`

server/api/src/rsGetMiscSvrInfo.cpp
```diff
@@ -65,100 +65,104 @@ rsGetMiscSvrInfo( rsComm_t *rsComm, miscSvrInfo_t **outSvrInfo ) {
...
+        myOutSvrInfo->certinfo.len = json_str.size();
+        memcpy(myOutSvrInfo->certinfo.buf, json_str.c_str(), myOutSvrInfo->certinfo.len);
```
korydraughn, line 162: Missing `#include <cstring>`

server/api/src/rsGetMiscSvrInfo.cpp
```diff
@@ -65,100 +65,104 @@ rsGetMiscSvrInfo( rsComm_t *rsComm, miscSvrInfo_t **outSvrInfo ) {
...
+        } catch(const nlohmann::json::exception& e) {
+            irods::log(LOG_ERROR, fmt::format("[{}] Failed to serialize SSL/TLS data - {}", __FUNCTION__, e.what()));
+        }
```
korydraughn, line 165: Add catch block for `const std::exception&`.

korydraughn (commented):
Almost there!

scripts/irods/test/test_imiscsvrinfo.py
```diff
@@ -0,0 +1,32 @@
+import sys
+
+if sys.version_info < (2, 7):
+    import unittest2 as unittest
+else:
```
korydraughn, line 5: This can be dropped since we're Python 3 only.

scripts/irods/test/test_imiscsvrinfo.py
```diff
@@ -0,0 +1,32 @@
...
+    import unittest
+
+from . import session
+from .. import lib
+from .. import paths
+from .. import test
```
korydraughn, line 11: Please remove any imports which aren't used.

scripts/irods/test/test_imiscsvrinfo.py
```diff
@@ -0,0 +1,32 @@
...
+
+class Test_Imiscsvrinfo(session.make_sessions_mixin([('otherrods', 'rods')], [('alice', 'apass')]), unittest.TestCase):
```
korydraughn, line 13: Could remove the tuple containing otherrods since `self.admin` isn't used.

scripts/irods/test/test_imiscsvrinfo.py
```diff
@@ -0,0 +1,32 @@
...
+    @unittest.skipUnless(test.settings.USE_SSL, 'Test case assumes use of SSL/TLS')
+    def test_imiscsvrinfo_prints_ssl_info_with_ssl_enabled__issue_7986(self):
+        _, out, _ = self.user.assert_icommand(['imiscsvrinfo'], 'STDOUT')
+        expected_outputs = ['enabled: true', 'issuer_name', 'not_after', 'not_before', 'public_key', 'subject_alternative_names', 'subject_name']
+        for output in expected_outputs:
+            self.assertIn(output, out);
+
+    @unittest.skipIf(test.settings.USE_SSL, 'Test case assumes absence of SSL/TLS')
+    def test_imiscsvrinfo_prints_ssl_info_with_ssl_disabled__issue_7986(self):
+        self.user.assert_icommand(['imiscsvrinfo'], 'STDOUT', 'enabled: false')
```
korydraughn, line 32: Let's add a few assertions about the negative space. For example, assert `issuer_name` and `not_after` aren't in the output.

The list doesn't have to span all properties though.

scripts/irods/test/test_imiscsvrinfo.py
trel: python3 only.... only in main.   right?

korydraughn: Python 3 for 4.3 as well.

korydraughn: https://docs.irods.org/4.3.3/release_notes/

> Notable updates include:
>
> - Python3 Compliance - All control scripts and the Python Rule Engine Plugin are now Python3 only. Please test any existing Python rules before upgrading.

trel: we did it.

scripts/irods/test/test_imiscsvrinfo.py
FifthPotato: I think I got them all?

server/api/src/rsGetMiscSvrInfo.cpp
FifthPotato: I tried hitting almost every else block where something could go wrong with a a log message. Anything else you think should be in here?

FifthPotato:
> Test looks good. I have a few suggestions for it though.
> 
> Does the new test pass in the testing environment?
> 
> I just realized this PR targets the main branch. As soon as this PR lands, we'll need to cherry-pick to 4-3-stable and confirm the new test passes in the testing environment.

To answer your question, the test passes as expected (with and without SSL) in my testing environment.

scripts/irods/test/test_imiscsvrinfo.py
korydraughn: See https://github.com/irods/irods/pull/8236#discussion_r1970350310

server/api/src/rsGetMiscSvrInfo.cpp
korydraughn: What you have is fine. We can tweak things in a later release if needed.

Resolving this.

korydraughn:
Your logic makes sense to me.

If the new tests pass with the icommand changes, then let's squash it. Please check that all review comments have been resolved before doing so.

server/api/src/rsGetMiscSvrInfo.cpp
```diff
@@ -65,100 +65,104 @@ rsGetMiscSvrInfo( rsComm_t *rsComm, miscSvrInfo_t **outSvrInfo ) {
...
+
+                // Set subject alternative name
+                GENERAL_NAMES* gen_names = static_cast<GENERAL_NAMES*>(X509_get_ext_d2i(cert, NID_subject_alt_name, nullptr, nullptr));
+                std::vector<std::string> san_vector;
+                if(gen_names) {
+                    unsigned char* tempStr;
+                    for(int i = 0; i < sk_GENERAL_NAME_num(gen_names); i++) {
+                        GENERAL_NAME* gen_name = sk_GENERAL_NAME_value(gen_names, i);
+                        if(gen_name->type != GEN_DNS) {
+                            // TODO: support more than just dNSName
```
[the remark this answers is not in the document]
korydraughn: Don't forget to open an issue for this and update the TODO comment.

korydraughn:
Looks like all review comments have been addressed and the code has been formatted. You can ignore the clang-tidy report.

With the formatting updated, please confirm the new test (with SSL enabled) passes. If it does, squash to taste. No pounds.

FifthPotato:
New test passes with SSL enabled.

Issue #8221, opened 2025-02-20
`test_ichmod.Test_ichmod.test_ichmod_access_levels` fails due to segfault in `freeRodsPathInpMembers`


SwooshyCueb:
- [x] main
- [x] 4-3-stable
---
Many, many tests fail on tip of 4-3-stable currently. I haven't tested on tip of main, but I assume tests are failing there, too.

I picked `test_ichmod.Test_ichmod.test_ichmod_access_levels` to look at, mostly arbitrarily, and was able to determine that it is failing due to iget segfaulting  in `clearRodsPath` in `freeRodsPathInpMembers`.

```
#0  0x00007ffff79dd060 in clearRodsPath (rodsPath=0x52435f4c4c4f431c) at /irods_source/lib/core/src/rodsPath.cpp:547
#1  0x00007ffff79dd124 in freeRodsPathInpMembers (path=0x7fffffffde48) at /irods_source/lib/core/src/rodsPath.cpp:570
#2  0x00007ffff790f66a in build_irods_path_structure (_path_list=..., _rods_args=..., _rods_env=0x7fffffffad88, _src_type=0, _dst_type=3, _flag=0, _rods_paths=0x7fffffffde48) at /irods_source/lib/core/src/irods_parse_command_line_options.cpp:280
#3  0x00007ffff790cdf9 in parse_opts_and_paths (_argc=4, _argv=0x7fffffffe648, _rods_args=..., _rods_env=0x7fffffffad88, _src_type=0, _dst_type=3, _flag=0, _rods_paths=0x7fffffffde48) at /irods_source/lib/core/src/irods_parse_command_line_options.cpp:404
#4  0x0000000000408915 in main (argc=4, argv=0x7fffffffe648) at /icommands_source/src/iget.cpp:43
```

Once this is fixed, let's verify that tests are passing again. I've only just started looking into this, so there may be other causes of test failures.

alanking:
Please provide additional information about the setup so we can try to reproduce the failures.

Looks like this may be a debug build? There's at least one other issue with tests failing for debug builds: https://github.com/irods/irods/issues/7412 Hmm...

MartinFlores751:
[...]

PR #8237, opened 2025-02-21
[#8221] Clear rodsPathInp_t instead of freeing memory (4-3-stable)

korydraughn:
The test referenced in the issue passed at the bench using a debug build. I only tested the one test case reported in the issue.

Running unit/core tests.

@MartinFlores751 Do you remember if this change was important in the memory leak fix which introduced `freeRodsPathInpMembers`?

### DIFF
--- a/lib/core/src/irods_parse_command_line_options.cpp
+++ b/lib/core/src/irods_parse_command_line_options.cpp
@@ -277,7 +277,7 @@ static int build_irods_path_structure(const path_list_t& _path_list,
         return USER__NULL_INPUT_ERR;
     }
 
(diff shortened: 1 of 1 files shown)


korydraughn:
All unit tests passed.

Running core tests.

MartinFlores751:
I don't believe this affects the server, only the clients, so it should be fine.

korydraughn:
Excellent.

alanking (commented):
Will approve once tests come back green

korydraughn:
Debug build passed all core tests.

korydraughn:
Pounded.

Merging.

Issue #8247, opened 2025-02-27
Skip or update test for topology environments


korydraughn:
## Bug Report / Testing

The following test is not designed for topology testing. It modifies the local server's config file and then attempts to upload a file to a resource. The problem is that the test creates the resource on a remote server, effectively ignoring the modifications to the local server_config.json.

https://github.com/irods/irods/blob/41312c68b06fe1278353ce4f23a32f76d371891c/scripts/irods/test/test_iput.py#L949-L982

The test either needs to be skipped or updated for topology environments. Given we don't have tools to modify configuration on remote servers (from within tests) yet, the best thing to do is to skip the test and open an issue to update it once we have such tools.

_This applies to 4-3-stable only._

korydraughn:
Related to #6835.

PR #8248, opened 2025-02-27
[#8247] Skip test until updated for topology environments (4-3-stable)

korydraughn:
I've watched all core topology tests + provider pass with this change and irods/irods_testing_environment#248.

Running the core tests in a non-topology environment to make sure that's still in a good spot.


scripts/irods/test/test_iput.py
```diff
@@ -946,6 +946,7 @@ def test_server_honors_server_config_option_checksum_read_buffer_size_in_bytes__
             # self.object_name deleted in teardown
             os.unlink(local_file)
 
+    @unittest.skipIf(test.settings.RUN_IN_TOPOLOGY, "Skip for Topology Testing")
```
korydraughn, line 949: Consider changing reason to state what needs to happen for the skip annotation to be removed.

korydraughn: Duplicate message from the main branch.

https://github.com/irods/irods/blob/fdf38573008cbb972675b939a6588945382ffa0a/scripts/irods/test/test_iput.py#L950-L951

korydraughn:
Core test run passed.

Updating and merging.

scripts/irods/test/test_iput.py
korydraughn: Done.

PR #8250, opened 2025-02-28
[#8249] Add support for `order_asc` to GenQuery1 string parser (4-3-stable)

korydraughn:
The test which triggered this fix passed for the PREP.

Issue #8249, opened 2025-02-28
Python REP test fails with INVALID_GENQUERY_AGGREGATE_FUNCTION


korydraughn:
- [x] main
- [x] 4-3-stable
---
## Bug Report

https://github.com/irods/irods/blob/e4ab2602acb427c552fc2e9047a9fb0bbdbca0fe/scripts/irods/test/test_prep_genquery_iterator.py#L318-L320

L320 fails when the following rule is executed because the parser which converts a GenQuery1 string to a GenQuery1 input struct doesn't recognize the `order_asc` symbol at L273.

https://github.com/irods/irods/blob/e4ab2602acb427c552fc2e9047a9fb0bbdbca0fe/scripts/irods/test/test_prep_genquery_iterator.py#L269-L283

The code of interest lives here. Notice the code looks for `"order"` instead of `"order_asc"`. So we can conclude the test wasn't quite right. Changing the test to use `order` allows it to pass.

https://github.com/irods/irods/blob/e4ab2602acb427c552fc2e9047a9fb0bbdbca0fe/lib/core/src/rcMisc.cpp#L3106-L3111

The test also needs to target the python REP. As it is right now, it is executing the rule text against all REPs _(and that's fine since the python REP is the only REP configured during testing - we shouldn't assume that to always be the case though)_.

korydraughn:
`order` is documented at docs.irods.org so we only need to add `order_asc` and make sure the documentation lists both forms.

PR #8251, opened 2025-02-28
[#8249] Add support for `order_asc` to GenQuery1 string parser (main)

korydraughn:
Port of #8250.

Will confirm this works at the bench soonish.


MartinFlores751 (approved):
Looks good. Same comment on style applies here as in the 4-3-stable PR on the string vs list use of `run_icommand(...)`. Not a blocker.

korydraughn:
Excellent. Pounding and merging.

korydraughn:
Confirmed the change works at the bench.

Needs to be updated to return the same error code as 4.3. Will open an issue for that.

korydraughn:
Created #8252.

PR #8255, opened 2025-03-05
[858] update github issue templates (main only)

trel:
this provides two templates for new issues.

submitters will have to pick from one of these two.   thoughts welcome on whether we should make 'more'... possibly 'build' or 'tools' related... or a blank one...

also possible to auto-tag things, but that felt bold.

also... possible checkboxes for branches and stuff... but again, not sure i want to lean into all the fancy things just yet - as some of this backend is still getting rolled out and polished at github itself.


.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -0,0 +1,20 @@
+---
+name: Bug Report
+about: Describe an iRODS bug
+title: ''
+labels: ''
```
alanking, line 5: Noticed that the enhancement template has an enhancement label whereas this does not have any labels. Is that on purpose?

trel: yes, i think we like to determine/verify if things are actually bugs, before adding the label.

enhancements are... well, safer to label early i think.

alanking: Okay. I think we're going to end up touching the labels in any case as we triage things, so I could see it either way. Up to you

trel: yeah, happy to ship this first with zero auto labeling.   will update the feature request one.

korydraughn (approved):
No concerns from me.

I say squash/pound it if you're happy with it. Can always tweak this without breaking things.

trel:
squashed and pounded

trel:
'new issue' now looks like this...

new-issue

Issue #7937, opened 2024-08-08
`KW_CFG_PLUGIN_TYPE_AUTHENTICATION` is "auth" but configuration JSON schema requires "authentication"


alanking:
- [x] irods/irods
- [x] irods/irods_client_icommands
- [x] pam_interactive

---

This is mostly an internal thing (I hope), but I noticed that the value of `KW_CFG_PLUGIN_TYPE_AUTHENTICATION` is "auth":
https://github.com/irods/irods/blob/bc2a9bf0d8d3587d1aad8bf6a9b3db650e0d6944/lib/core/src/irods_configuration_keywords.cpp#L174

And the JSON schema for server_config.json requires "authentication" for the auth plugin configuration key:
https://github.com/irods/irods/blob/bc2a9bf0d8d3587d1aad8bf6a9b3db650e0d6944/schemas/configuration/v4/server_config.json.in#L128-L134

The `KW_CFG_PLUGIN_TYPE_*` keywords appear to be mainly used for getting/setting configurations in server_config. So why does this keyword have a different value from the schema? Let's change it to "authentication".

alanking:
The "auth" string is required to be "auth" because of the install location for the authentication plugins. e.g. Ubuntu 24: 
```bash
$ ls /usr/lib/irods/plugins/auth/
libirods_auth_plugin-native_client.so  libirods_auth_plugin-pam_password_client.so  libnative_client.so  libosauth_client.so  libpam_client.so
libirods_auth_plugin-native_server.so  libirods_auth_plugin-pam_password_server.so  libnative_server.so  libosauth_server.so  libpam_server.so
```
We would either need to change the directory name to use all of the letters, or change server_config to match and be "auth" instead of "authentication". I'm not sure which is easier. Regardless, I no longer see this as a refactor.

korydraughn:
Nice observation.

Hmm, changing the directory name could break some people's setup/tools if they expect it to be named auth. For that reason, changing the schema seems safer.

What properties go inside the authentication stanza? Weren't they moved into R_GRID_CONFIGURATION for 4.3.x?

alanking:
`pam_interactive` uses the authentication stanza to configure insecure_mode and the name of the PAM stack to use.

korydraughn:
In that case, I'm now thinking it's better to change the directory name since that's handled by the packaging. Knowing where the plugins are installed isn't a requirement for using the software. The dev library and cmake define the install directories, etc.

I've convinced myself that changing the directory name is better.

[...]

Issue #6229, opened 2022-03-08
Remove `ASSERT_ERROR` and `ASSERT_PASS`


alanking:
Deprecated in #3727

PR #8256, opened 2025-03-05
Remove ASSERT_ERROR and SimpleQuery (main)

alanking:
Addresses #6229 
Addresses #5247 
~In service of #7937~

Core tests are running, so leaving in draft until those come back.

Some general notices...

 - Everything related to SimpleQuery - including the API number and its entry in the API table, and packing instructions for its related data structures - has been removed.
- ~The auth plugins now live in the "authentication" sub-directory of `IRODS_PLUGINS_DIRECTORY`. `pam_interactive` will need to be updated accordingly, as well. I'm not sure how / whether to remove the shared objects in the "auth" sub-directory which will no longer be used on upgrade from 4.x.~
- ~The legacy auth plugins will remain under the "auth" sub-directory of `IRODS_PLUGINS_DIRECTORY`.~


alanking:
I'm tempted to drop the clang-format commit and ignore clang-tidy and Codacy for this PR because I didn't even touch the lines it's complaining about. It's only pointing them out because the parent directory was renamed so it thinks it's a new file, I think.

lib/api/include/irods/apiNumberData.h
```diff
@@ -37,8 +37,6 @@ API_NUMBER(DATA_PUT_AN,                             607)
 API_NUMBER(DATA_OBJ_GET_AN,                         608)
 API_NUMBER(DATA_GET_AN,                             609)
 API_NUMBER(DATA_COPY_AN,                            611)
-// DEPRECATED: SimpleQuery is deprecated. Use GenQuery or SpecificQuery instead.
-API_NUMBER(SIMPLE_QUERY_AN,                         614)
```
trel, line 41: just checking...  we remove API numbers?     leaving them here would prevent accidental reuse?

alanking: Well, we apparently have done so in the past. Notice that API number 610 is missing. The prevention of accidental reuse is the only real reason to leave it there.

If we leave it, we cannot change the signature or the structure of the inputs and outputs (meaning we have to continue maintaining all the deprecated data structures associated with SimpleQuery). If we remove it, the caller will just an error back saying no API exists by that number. I think I prefer the latter.

alanking: As a compromise, we could leave it here commented out (or just a descriptive comment explaining what 614 used to be).

korydraughn: We also have `is_api_number_supported`.
https://github.com/irods/irods/blob/8013bd306e7b4fb54a7af6ff8695faf1e698f8f5/server/core/src/rsApiHandler.cpp#L67-L79

The XMSG API numbers are still present in apiNumberData.h.
https://github.com/irods/irods/blob/8013bd306e7b4fb54a7af6ff8695faf1e698f8f5/lib/api/include/irods/apiNumberData.h#L76-L78
https://github.com/irods/irods/blob/8013bd306e7b4fb54a7af6ff8695faf1e698f8f5/server/core/src/irods_api_number_validator.cpp#L13-L24

So we have a tool for disallowing use of the API while allowing removal of the data types and functions.

With that said, I also prefer the comment out approach. The code I linked feels like bloat and should be removed since this effort is effectively establishing the rules for how to remove an API.

alanking:
Seeing two core test failures...

`test_native_rule_engine_plugin.Test_Native_Rule_Engine_Plugin.test_network_pep` - I ran this a couple of times and it failed, then ran it many more times and it passed every time. This test checks messages in the logs, so I'm not too concerned about it.

This one is more interesting: `test_misc.Test_Misc.test_user_administration__issue_7208`

It fails to authenticate as a user after a modification that happens in the test. The log indicates that it failed to load an authentication plugin - it was looking for `libnative_server.so` in the new `plugins/authentication` directory. `libnative_server.so` is the *legacy authentication* shared object.

I also ran the unit tests and found that several of these tests are failing (namely, irods_atomic_apply_acl_operations, irods_rc_switch_user, irods_rcTicketAdmin, irods_connection_pool, irods_user_administration, and irods_client_connection). The common thread among all of these tests is usage of `clientLoginWithPassword`. `clientLoginWithPassword` uses legacy authentication plugins.

I'm thinking that we should probably back out changes related to #7937 until we decide how to proceed with that.

alanking:
On thinking about this further, the latest changes seem to resolve the issue. We introduce a new keyword with the old string for legacy authentication in the legacy authentication machinery (the auth manager) and replace the old keyword with the new string for standard authentication. It's a little weird because we will likely deprecate the new keyword soon (see #8257), but it seems to be a decent shim.

lib/api/include/irods/apiNumberData.h
alanking: Okay, I left it in, but commented out, and included a comment saying that the API has been removed.

alanking:
Unit tests passed. Running core tests again. Once those pass, I'll take this out of draft.

alanking:
`test_native_rule_engine_plugin.Test_Native_Rule_Engine_Plugin.test_network_pep` continues to fail. I'm going to go ahead and take this out of draft because I'm fairly certain this has nothing to do with my changes. I'll run the tests again tip of main without any changes and see what happens.

korydraughn (approved):
Changes look good.

Let's remove the body of the ASSERT-based commit.

Once done, pound it if that test failure is unrelated.

alanking:
Removed triumphal message, #'d, mergin

Issue #6766, opened 2022-12-09
remove `imeta rmw` (and `addw`?) (and `lsw`?) (and `adda`?)


mstfdkmn:
### iRODS Version, OS and Version
iRODS 4.3.0, iCommands 4.3.0, centos7

### What did you try to do?
I tried to remove metadata attached by using wildcard option - `imeta rwm` 

```
[root@localhost irods]# imeta add -d test.py atr val
[root@localhost irods]# imeta ls -d test.py
AVUs defined for dataObj /hpc_icts/home/vsc33586/test.py:
attribute: atr
value: val
units:
```

irods info:
```
[root@localhost irods]# imiscsvrinfo
RCAT_ENABLED
relVersion=rods4.3.0
apiVersion=d
rodsZone=hpc_icts
up 1 days, 8:17
```
iCommands:
`iRODS Version 4.3.0                ihelp`

### Expected behavior
To see `rmw is deprecated` or `Unrecognized subcommand`

### Observed behavior (including steps to reproduce, if applicable)
```
[root@localhost irods]# imeta rmw -d test.py % %
[root@localhost irods]# imeta ls -d test.py
AVUs defined for dataObj /hpc_icts/home/vsc33586/test.py:
None
```

korydraughn:
That is correct. No message is printed to the terminal, but the help text does mention deprecation.

See #6417

mstfdkmn:
Ah okay thanks. Can we know possible removal time of `w(s)` in metadata?

korydraughn:
We haven't discussed when the wildcard operations will be removed. You expect that to happen in a much later release.

The deprecation of the operation is there to signal that users should start to move away from using the wildcard operations.

trel:
do we need an 'remove imeta rmw' issue?   I couldn't find one.   If not, this could be it?

korydraughn:
I don't remember there being an issue for removal, so yes, I think we can use this issue for that.

mstfdkmn:
Btw, a small remark: my main concern is how to implement the logical quota for our production environment, that is why asked this regarding an issue exists at the metadata guard plugin.

trel:
renamed this issue.   not sure about `addw` or `lsw` either...

korydraughn:
`addw` is also marked as deprecated.

trel:
So, if we agree to removing all three - we should declare `lsw` deprecated beforehand as well.

korydraughn:
[...]

Issue #3683, opened 2017-10-17
feature request: iadmin modresc rebalance visibility


kript:
- [x] master
- [x] 4-2-stable
- [x] 4-1-stable
--------

Currently, when a rebalance operation is run on an IES, it is a manual operation. The command provides no feedback while it is running, apart from errors from the logs should it fail. 
in 4.1.11 onwards, logging will appear to show any dirty or single object rebalancing operations that occur. 

However, there is no way for an automated check, another sysadmin, or the iadmin tool itself to check if a rebalance is currently ongoing. The iadmin tool may well happily start a rebalance every time the command is invoked, and it is not clear what behaviour may result. 

This can lead to more than one rebalance being run at one time.

We have also seen some evidence where rebalance command fails, but, as per #3665 may continue to run 'in the background' in some form. 

As a sysadmin, I would like to be able to check if a rebalance was occurring, have a documented way of seeing what a rebalance would do before it was started, ideally have a way of seeing its progress, and a way to safely terminate it. When the rebalance fails, there should be a way to clear up any lock, while retaining the ability to correctly report on whether the rebalance was still running or not.

trel:
How does a metadata element on the resource being rebalanced sound?

A | V | U
-------- | ------- | -------
rebalance_operation | source_hostname | YYYYMMDDHHMMss

Any better suggestions for what to capture in the AVU?

This would allow a new process to check whether the catalog thinks there is a rebalance operation in flight.

kript:
I didn't realise the rebalance worked on a per-resource basis, I thought it was 'just' a list generated by a query it worked its way through, so I've learned something!

Other than that, is the suggestion that to determine if a rebalance is ongoing you check all the resources with imeta ls? Is there a quicker or simpler way to determine if a rebalance is in progress than iterating across all the resources, perhaps with iquest? Apologies - not familiar with using iquest to lookup metadata if its even possible..
It seems a lot of queries to be performed, although presumably only done at the start of a rebalance?

trel:
`iquest` can see the metadata information, so you can run a single query to get the metadata from all the resources that might be running a rebalance:

```
$ imeta add -R demoResc rebalance_operation `hostname`:$$ `date -u +%Y%m%dT%H%M%SZ`
$ imeta add -R r1 rebalance_operation somewhere.nice:$$ `date -u +%Y%m%dT%H%M%SZ`                                                                                                
[...]

PR #8263, opened 2025-03-07
[#6766] ModAVUMetadata / imeta: Remove adda, lsw, addw, and rmw (main)

alanking:
In service of #6766 

Companion PR: https://github.com/irods/irods_client_icommands/pull/541

Unit tests pass. Core tests running now, but they're almost done and are passing so far. Will take out of draft once complete.


scripts/irods/test/test_iadmin.py
```diff
@@ -2138,27 +2140,30 @@ def tearDown(self):
...
-            self.assertTrue(attempts != 0, msg="rebalance visibility metadata not in catalog")
-            out,_,_ = self.admin.run_icommand(['imeta','ls','-R','repl'])
-            if "rebalance_operation" not in out:
-                check_and_remove_rebalance_visibility_metadata(attempts - 1)
-            else:
-                self.admin.assert_icommand("imeta rmw -R repl rebalance_operation % %") # b/c #3683
+            self.assertGreater(attempts, 0, msg="rebalance visibility metadata not in catalog")
+            resource_metadata_query = "select META_RESC_ATTR_NAME, META_RESC_ATTR_VALUE, META_RESC_ATTR_UNITS " \
+                                      f"where RESC_NAME = '{resource_name}'"
+            resource_avus = json.loads(self.admin.run_icommand(["iquery", resource_metadata_query])[0].strip())
```
korydraughn, line 2149: Just want to put on your radar that GenQuery2 limits the resultset to 256 rows when the query doesn't define a limit.

server/api/src/rsGeneralAdmin.cpp
```diff
@@ -316,12 +325,12 @@ int _remove_rebalance_timestamp_avu_from_resource(
             free( modAVUMetadataInp.arg5 );
         }};
         // remove AVU from resource
-        modAVUMetadataInp.arg0 = strdup( "rmw" );
+        modAVUMetadataInp.arg0 = strdup("rm");
         modAVUMetadataInp.arg1 = strdup( "-R" );
         modAVUMetadataInp.arg2 = strdup( _resource_name.c_str() );
         modAVUMetadataInp.arg3 = strdup( "rebalance_operation" );
-        modAVUMetadataInp.arg4 = strdup( "%" );
-        modAVUMetadataInp.arg5 = strdup( "%" );
```
korydraughn, line 324: The original removed many AVUs. Does the new code still cover that?

Another question is - should it remove multiple AVUs?

korydraughn (commented):
Looks good.

Waiting for test results before approval.

alanking:
Core tests passed. Taking this out of draft.

scripts/irods/test/test_iadmin.py
alanking: Gotcha. If this ever has more than one AVU on it, then that's actually a problem. Perhaps I can insert an assertion to make sure of that here and we'll be good.

server/api/src/rsGeneralAdmin.cpp
alanking: I actually think this is more correct than removing all of the rebalance_operation AVUs. If multiple rebalance_operation AVUs appear on a resource, that means that multiple rebalance operations were started and each annotated its own metadata, which is a bug. The whole point of the AVU is to prevent multiple rebalances from occurring, so if that ever happens, this at least makes the bug more visible. Thoughts?

alanking: @trel may have more insight into this. Here's the original issue from which this feature came: https://github.com/irods/irods/issues/3683

I don't see any specific requirements around removing all of the rebalance_operation AVUs once a rebalance completes. Again, the fact that multiple rebalances can run simultaneously is itself a bug.

trel: Correct - the AVU itself is the 'lock' for a rebalance running.  So there should / cannot be more than one.

scripts/irods/test/test_iadmin.py
korydraughn: Sure. Doesn't hurt.

server/api/src/rsGeneralAdmin.cpp
korydraughn: All of that sounds great.

Resolving.

korydraughn:
Will circle back to this once that final review comment is resolved.

scripts/irods/test/test_iadmin.py
alanking: Done. Test passed

korydraughn (approved):
Squash to taste and pound it.

korydraughn:
Consider mentioning that admins should not rely on the db PEP for policy as it is not guaranteed to exist in later versions.

alanking:
> Consider mentioning that admins should not rely on the db PEP for policy as it is not guaranteed to exist in later versions.
> 
> For [4f17d56](https://github.com/irods/irods/commit/4f17d56284952b46e727c7cb9dcce194566ab0eb).

Well, the (worst?) thing about that is... `rmw` doesn't have a dedicated database PEP, but `addw` does (or... did, I guess). :( It is hidden in the `option` parameter for `db_del_avu_metadata_op`:
https://github.com/irods/irods/blob/a73a90cf0ba46c86f7cb6e9c5eae7e132faeb147/server/icat/src/icatHighLevelRoutines.cpp#L2573-L2576

So... we won't be removing a database PEP but rather the `1` value for `option` in that PEP.

korydraughn:
Gotcha. Forgot that's how that worked.

Is that part of this PR or is it planned for a later one?

alanking:
I was not planning to remove that as part of this PR because it's being used internally by the `set` operation to clear all the AVUs with a particular attribute before setting a new one with a particular value/units. I don't think re-implementing `set` is in scope for the removal of `rmw` from the ModAVUMetadata API. Thoughts?

korydraughn:
Agreed.

Let's capture that in an issue for later. We should avoid triggering policy related to removed APIs if possible. As you said, that's out of scope for this PR though.

alanking:
Created https://github.com/irods/irods/issues/8266 for that.

korydraughn:
Pound and merge it.

korydraughn:
Oops, noticed a leftover commit that needs to be squashed.

korydraughn:
See commit 4891b5a

alanking:
Squashed, #'d, mergin

Issue #7985, opened 2024-10-24
Remove acAclPolicy{msiAclPolicy('STANDARD');} (strictACL as a knob)


trel:
- [x] irods/irods
- [x] icommands
- [x] PREP
- [x] irods_docs

---

## Feature

Deprecation is https://github.com/irods/irods/issues/6843.

This issue is about removing the `STANDARD` option.

korydraughn:
Things to consider:
- Will the MSI continue to exist?
- How do we remove it from core.re? Do we remove it from core.re at all?
- If the MSI is removed, how does the NREP deal with the call still existing inside core.re?

trel:
I think the msi continues to exist, but we change the implementation to be a no-op that emits a warning.   So existing rulefiles will not throw errors?   Or maybe, it emits the warning for one major release before we remove the msi completely.

korydraughn:
So ...
- iRODS 4 prints a warning about the MSI's deprecation on server startup, not agent startup
- iRODS 5 does not provide the MSI, leading to server startup failing
  - A clear error message is logged for the admin
  - Config check could be used to verify the core.re file is good
    - i.e. Use the list rules function of the plugin

trel:
or... if we're a little nicer...
- iRODS 4 does as you say (somewhat noisy)
- iRODS 5 complains on every agent startup (more noisy)
- iRODS 6 does not start b/c syntax error / missing MSI error

korydraughn:
I can see it going either way, but I'm still in the camp of remove in iRODS 5 since it's a major release.

Will need to think about this a bit more.

trel:
sure, 6 is a major release too.    so yes, either is fine i think...

removing in 5, before we've even deprecated in 4 does give me a little pause...

korydraughn:
> removing in 5, before we've even deprecated in 4 does give me a little pause...

I agree. My comments assume the deprecation happens in 4 first. That's my bad. I should've mentioned that.

korydraughn:
Related to #7954.

alanking:
Need to confirm / merge https://github.com/irods/irods_rule_engine_plugin_python/pull/233, I think. Plus, take a pass over the docs.

alanking:
I think that's everything... @MartinFlores751 - If you agree, please close. Thanks

alanking:
Oh, and close #7954 as well

PR #8264, opened 2025-03-07
[#7985] Remove `msiAclPolicy(...)` and `acAclPolicy` static PEP (main)

MartinFlores751:
Quick link: #7985.

This PR removes both `msiAclPolicy(...)` and the static PEP `acAclPolicy`.

---

- [x] Remove any remaining `*AclPolicy` bits.
- [x] Check what relies on `STRICT_ACL_KW` and if anything needs to be done about that
- [x] Investigate if `DISABLE_STRICT_ACL_KW` needs any changes
- [x] Investigate if `irods::AGENT_CONN_KW` with `StrictACLOverride` causes an ACL change.
  - [x] See if removing the specific use broke anything.
- [x] Build and test the PR.


server/re/src/icatAdminMS.cpp
```diff
@@ -633,82 +633,6 @@ int msiRenameLocalZoneCollection(msParam_t* _new_zone_name, ruleExecInfo_t* _rei
...
-        }
-    }
-
-    // set a strict acl prop
-    try {
-        irods::set_server_property<std::string>(irods::STRICT_ACL_KW, strict);
-    } catch ( irods::exception& e ) {
-        irods::log(e);
-        return e.code();
-    }
```
korydraughn, line 705: You're going to need to port this into agent_main.cpp (i.e. the agent factory).

Notice the check on L686.

korydraughn: Port it, but without the options to change it from STRICT mode.

MartinFlores751: Spoiler for a commit I haven't pushed yet (going to right after). I put the following from the function you highlighted at the old call site of the static PEP:
```c++
if (irods::KW_CFG_SERVICE_ROLE_PROVIDER == svc_role) {
        chlGenQueryAccessControlSetup(NULL, NULL, NULL, 0, 2);
}
```

Do we set this up in agent main?

I'm still seeing more calls to `chlGenQueryAccessControlSetup(...)` in the static PEP call-site: https://github.com/irods/irods/blob/a73a90cf0ba46c86f7cb6e9c5eae7e132faeb147/server/api/src/rsGenQuery.cpp#L840-L855

korydraughn: > I put the following from the function you highlighted at the old call site of the static PEP.

Old call site of the static PEP? Got a link to a source file and line number?

> Do we set this up in agent main?

`rsGenQuery` is invoked quite early in the process and is the first function which triggers the PEP. I suspect that's why you're seeing multiple invocations of `chlGenQueryAccessControlSetup`.

MartinFlores751: > > I put the following from the function you highlighted at the old call site of the static PEP.
> 
> Old call site of the static PEP? Got a link to a source file and line number?

Here it is: https://github.com/irods/irods/blob/a73a90cf0ba46c86f7cb6e9c5eae7e132faeb147/server/api/src/rsGenQuery.cpp#L808-L834

> > Do we set this up in agent main?
> 
> `rsGenQuery` is invoked quite early in the process and is the first function which triggers the PEP. I suspect that's why you're seeing multiple invocations of `chlGenQueryAccessControlSetup`.

Yeah, kinda figured. Not sure what all the magic numbers at the end mean, so that's why I put the call at the old call site.

korydraughn: I think all of that can be removed in favor of setting the necessary values in agent_main.cpp. There's no reason to have GenQuery1 be responsible for that anymore.

korydraughn: Moving it to the agent factory's init logic means it will be invoked one time too.

korydraughn: Basically, somewhere before this point.

https://github.com/irods/irods/blob/a73a90cf0ba46c86f7cb6e9c5eae7e132faeb147/server/main_server/src/agent_main.cpp#L300

MartinFlores751: Can the `irods::KW_CFG_SERVICE_ROLE_PROVIDER == svc_role` check be done at that point?

korydraughn: It should be possible to do the check there.

MartinFlores751: Will try and then take out of draft for building and testing.

MartinFlores751: Done.

MartinFlores751:
Building and testing.

server/api/src/rsGenQuery.cpp
```diff
@@ -760,8 +760,6 @@ _rsGenQuery( rsComm_t *rsComm, genQueryInp_t *genQueryInp,
     int resc_hier_attr_pos = -1;
 
     static int ruleExecuted = 0;
```
alanking, line 762: Can `ruleExecuted` also be removed?

server/main_server/src/agent_main.cpp
```diff
@@ -297,6 +297,13 @@ auto main(int _argc, char* _argv[]) -> int
         // This allows rule engine plugins to setup state which doesn't change frequently.
         irods::re_plugin_globals->global_re_mgr.call_setup_operations();
 
+        std::string svc_role;
+        irods::error ret = get_catalog_service_role(svc_role);
```
alanking, line 301: Do we need to handle `!ret.ok()` here?

server/re/src/icatAdminMS.cpp
```diff
@@ -633,82 +633,6 @@ int msiRenameLocalZoneCollection(msParam_t* _new_zone_name, ruleExecInfo_t* _rei
...
-    else {
-        if( irods::KW_CFG_SERVICE_ROLE_PROVIDER == svc_role ) {
-            chlGenQueryAccessControlSetup( NULL, NULL, NULL, 0, 0 );
-        }
-    }
-
-    // set a strict acl prop
-    try {
-        irods::set_server_property<std::string>(irods::STRICT_ACL_KW, strict);
```
alanking, line 701: Not sure if we're allowed to remove it, but I see `DISABLE_STRICT_ACL_KW`. Should stop supporting this in the server and/or remove it, or at least check to see how it's being used?

alanking (commented):
Looking good so far. Saw a couple other things that we might be able to chop out

server/re/src/icatAdminMS.cpp
trel: yes, please let's not support something with that name

MartinFlores751:
Unit tests failed.

Have a `CHECK( rc_atomic_apply_acl_operations(static_cast&lt;RcComm*>(conn), json_input.c_str(), &amp;json_error_string) == ec )` which expands to `-830000 == -358000`.

There are some other errors too.

server/main_server/src/agent_main.cpp
MartinFlores751: Done.

server/api/src/rsGenQuery.cpp
```diff
@@ -805,31 +810,40 @@ _rsGenQuery( rsComm_t *rsComm, genQueryInp_t *genQueryInp,
...
+            // NOLINTNEXTLINE(bugprone-narrowing-conversions,cppcoreguidelines-narrowing-conversions)
+            status = ret.code();
+        }
 
+        if (irods::KW_CFG_SERVICE_ROLE_PROVIDER == svc_role) {
+            constexpr auto priv{0};
+            constexpr auto control_flag{2};
+
+            // Call directly to avoid additional policy
+            chl_gen_query_access_control_setup_impl(nullptr, nullptr, nullptr, priv, control_flag);
```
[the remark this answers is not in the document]
MartinFlores751: Building with a forward declared `chl_gen_query_access_control_setup_impl` doesn't work as is right now. Going to use `chlGenQueryAccessControlSetup` for now to tackle `DISABLE_STRICT_ACL_KW`, will revisit this afterward.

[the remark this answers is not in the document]
korydraughn: That's very surprising. Anyway, that can be addressed later.

Carry on.

server/api/src/rsGenQuery.cpp
MartinFlores751: Add check with `ruleExecuted` back in.

MartinFlores751:
Found this bit of code: https://github.com/irods/irods/blob/f11868c488e1eec26e9cce0af227dd0f860d9fbd/server/core/src/dataObjOpr.cpp#L820-L833

Is `irods::AGENT_CONN_KW` used anywhere outside of the server to make some decisions? Do they check for `StrictACLOverride` as the value?

---

I think this is used in the irods server in combination with the `DISABLE_STRICT_ACL_KEYWORD`, going to check for that...

server/re/src/icatAdminMS.cpp
MartinFlores751: Removed.

MartinFlores751:
Rebased onto main, I think the tests were failing due to some other tests. Updated the icommands companion PR as well.

Rebuilding and rerunning tests.

server/api/src/rsGenQuery.cpp
```diff
@@ -805,54 +794,50 @@ _rsGenQuery( rsComm_t *rsComm, genQueryInp_t *genQueryInp,
...
+            constexpr auto priv{0};
+            constexpr auto control_flag{2};
 
-    // =-=-=-=-=-=-=-
-    // detect if a request for disable of strict acls is made
-    int acl_val = -1;
-    char* dis_kw = getValByKey( &genQueryInp->condInput, DISABLE_STRICT_ACL_KW );
-    if ( agent_conn_flag && dis_kw ) {
-        acl_val = 0;
+            // Call directly to avoid additional policy
```
alanking, line 811: This call is still going to trigger at least one database PEP (`pep_database_gen_query_access_control_setup_*`), so I'm not sure that I understand this comment. Please explain.

alanking (commented):
Seems good on inspection. Hoping the tests pass

server/api/src/rsGenQuery.cpp
MartinFlores751: That comment was associated with this change: https://github.com/irods/irods/pull/8264#discussion_r1994092194

It was reverted since the code wouldn't compile. Will remove if we continue to use `chlGenQueryAccessControlSetup(...)`.

MartinFlores751:
Single test failure: `irods.test.test_native_rule_engine_plugin.Test_Native_Rule_Engine_Plugin.test_network_pep`. Multiple attempts to run fail.

All other unit and core tests pass.

Short log:
```
  File "/var/lib/irods/scripts/irods/test/test_native_rule_engine_plugin.py", line 543, in test_network_pep
    self.helper_test_pep(pep_map[self.plugin_name], "iput -f --metadata ATTR;VALUE;UNIT "+self.testfile)
  File "/var/lib/irods/scripts/irods/test/test_native_rule_engine_plugin.py", line 123, in helper_test_pep
    check_string_count_in_log_section (s,number_of_strings_to_look_for)
  File "/var/lib/irods/scripts/irods/test/test_native_rule_engine_plugin.py", line 112, in check_string_count_in_log_section
    lib.delayAssert(
  File "/var/lib/irods/scripts/irods/lib.py", line 552, in delayAssert
    raise AssertionError
AssertionError
```

alanking:
> Single test failure: `irods.test.test_native_rule_engine_plugin.Test_Native_Rule_Engine_Plugin.test_network_pep`. Multiple attempts to run fail.

I have been seeing this test fail occasionally on main as well. What OS / database are you using? I'll write up an issue for this and we can move this along.

server/api/src/rsGenQuery.cpp
```diff
@@ -805,54 +794,50 @@ _rsGenQuery( rsComm_t *rsComm, genQueryInp_t *genQueryInp,
...
 
     // =-=-=-=-=-=-=-
     // cache the old acl value for reuse later if necessary
-    int old_acl_val =  chlGenQueryAccessControlSetup(
-                           rsComm->clientUser.userName,
-                           rsComm->clientUser.rodsZone,
-                           rsComm->clientAddr,
-                           rsComm->clientUser.authInfo.authFlag,
-                           acl_val );
+    int old_acl_val = chlGenQueryAccessControlSetup(rsComm->clientUser.userName,
```
alanking, line 836: Codacy says... `old_acl_val` is never used (which is true).

MartinFlores751:
Reverting the most recent commit has all tests passing, with two retries on one test.

`irods.test.test_ichmod.Test_ichmod.test_ichmod_access_levels` failed twice with the following stack trace:

```
  File "/var/lib/irods/scripts/irods/test/test_ichmod.py", line 208, in test_ichmod_access_levels
    self.user.assert_icommand(params.get('cmd'), *check[2])
  File "/var/lib/irods/scripts/irods/test/session.py", line 162, in assert_icommand
    return assert_command(*args, **kwargs)
  File "/var/lib/irods/scripts/irods/test/command.py", line 76, in assert_command
    return _assert_helper(*args, should_fail=False, **kwargs)
  File "/var/lib/irods/scripts/irods/test/command.py", line 104, in _assert_helper
    assert result
AssertionError
```

The specific line that failed:
```
...
 --- IrodsSession: icommand executed by [alice#tempZone] [irmdir /tempZone/home/public/test_ichmod_collection_access] --- 
Assert Command: irmdir /tempZone/home/public/test_ichmod_collection_access
Expecting EMPTY: ['']
  stdout:
    | rmdirColl: rcRmColl failed with error -818000
  stderr:
    | 
Unexpected output on stdout

FAILED TESTING ASSERTION
...
```

alanking:
Thanks. Created https://github.com/irods/irods/issues/8299 to address the occasional failure for `test_native_rule_engine_plugin.Test_Native_Rule_Engine_Plugin.test_network_pep`.

As for `test_ichmod.Test_ichmod.test_ichmod_access_levels`, that one is quite unfortunate to not have employed a `subTest` so we can see what failed. I don't like that it failed and then started working after changing nothing. :) Please run a few more times in isolation and see if things improve.

MartinFlores751:
> As for `test_ichmod.Test_ichmod.test_ichmod_access_levels`, that one is quite unfortunate to not have employed a `subTest` so we can see what failed. I don't like that it failed and then started working after changing nothing. :) Please run a few more times in isolation and see if things improve.

This one behaved like this after reverting the latest commit. Still, it'll be tested again after addressing the remaining comments.

server/api/src/rsGenQuery.cpp
MartinFlores751: Done.

server/api/src/rsGenQuery.cpp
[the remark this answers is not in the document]
MartinFlores751: I'm not 100% sure, but I think the reason that this doesn't work is that `general_query.cpp`, where `chl_gen_query_access_control_setup_impl` is defined, is not being linked (or set as a dependency?) against the `irods_server_api` target in CMake.

Will resolve if we're fine with using `chlGenQueryAccessControlSetup` still.

server/api/src/rsGenQuery.cpp
MartinFlores751: Removed.

MartinFlores751:
> Thanks. Created #8299 to address the occasional failure for `test_native_rule_engine_plugin.Test_Native_Rule_Engine_Plugin.test_network_pep`.
> 
> As for `test_ichmod.Test_ichmod.test_ichmod_access_levels`, that one is quite unfortunate to not have employed a `subTest` so we can see what failed. I don't like that it failed and then started working after changing nothing. :) Please run a few more times in isolation and see if things improve.

Rebuilding and retesting.

server/api/src/rsGenQuery.cpp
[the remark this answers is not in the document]
MartinFlores751: In standup: We decided that we're going to create an issue for using `chl_gen_query_access_control_setup_impl` and not use it in this PR to get it moving.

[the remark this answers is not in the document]
MartinFlores751: Created #8304.

MartinFlores751:
Core and unit tests passed.

korydraughn:
Good. The question to answer now is - is a run of the unit/core tests enough to prove correctness?

Please review the documentation for STRICT to determine whether we can put other a test which bangs on the change directly.

MartinFlores751:
The direct documentation doesn't help much:

https://github.com/irods/irods_docs/blob/c9f5c10479e090c79c0517f36fc5c33d1da8d419/docs/system_overview/users_and_permissions.md?plain=1#L15-L37

But the changelogs show things that were/should be affected by the ACLs, here's one example:

https://github.com/irods/irods_docs/blob/c9f5c10479e090c79c0517f36fc5c33d1da8d419/docs/legacy_release_notes.md?plain=1#L93

---

I think it might be worth doing another check of what calls the init function, and figuring out what the magic numbers really mean, to make sure the fix  is correct. Put differently, making sure that whatever the magic input is (unless there's a special exception), the 'state' stays 'strict'.

I'll probably do this in a bit, unless others think it's unnecessary.

korydraughn:
After doing some review with @MartinFlores751, we've agreed to revert the changes which remove the `DISABLE_STRICT_ACL_KW` keyword/usage. We're going to leave it in until we have a better understanding of how it's used.

For now, our goal is to focus on disallowing modification of the ACL policy.

MartinFlores751:
Rebase onto tip of main.

MartinFlores751:
Building and running tests.

MartinFlores751:
Unit tests passed but core tests failed. Only `test_all_rules.Test_AllRules` failed.

Looking into the logs, we see:
```
 --- IrodsSession: icommand executed by [rods#tempZone] [irule -vF /var/lib/irods/clients/icommands/test/rules/rulemsiAclPolicy.r -r irods_rule_engine_plugin-irods_rule_language-instance] --- 
Assert Command: irule -vF /var/lib/irods/clients/icommands/test/rules/rulemsiAclPolicy.r -r irods_rule_engine_plugin-irods_rule_language-instance
Expecting STDOUT_SINGLELINE: ['completed successfully']
  stdout:
    | 
  stderr:
    |  ERROR: Incomplete rule input for /var/lib/irods/clients/icommands/test/rules/rulemsiAclPolicy.r
Unexpected output on stderr

FAILED TESTING ASSERTION
```

Whichs shows that I forgot to use the companion icommands PR... Which I checked and is true.

Will rerun that specific test with the companion PR just to see it pass.

MartinFlores751:
`test_all_rules.Test_AllRules` passed with the companion PR.

alanking (commented):
Squash em and will take one last pass over it

MartinFlores751:
Squashed.

alanking (approved):
\# it

MartinFlores751:
Pounded.

Issue #7834, opened 2024-06-24
Remove update_deprecated_columns.py script


alanking:
- [x] main
- [x] irods_docs

---

This script was created to help mitigate some problems for deployments upgrading from pre-4.2.0 to 4.2.0+.

Deployments upgrading from pre-4.2.0 to 5.0.0+ should be aware that older queries using the deprecated columns may produce unexpected / incorrect results as documented in the issue which spawned this script: #3981

Deprecated here: https://github.com/irods/irods/issues/7835

alanking:
@MartinFlores751 - Please close if complete. Thanks

PR #8271, opened 2025-03-10
[#7834] Remove `update_deprecated_columns.py` (main)

MartinFlores751:
Quick link to the issue: #7834.

Removes the deprecated (in 4.3.4) `update_deprecated_columns.py`.


trel:
and... tell me a story about someone upgrading directly from 4.2.3 or prior (before this script) to 5.0.0...

Are we okay?

alanking (commented):
As a side note, I think we should consider moving away from adding these known strings to the catalog. It's a small thing, but a catalog with 50 million objects at 2 replicas per object will use over 5GB in the catalog just for the strings in the 3 deprecated columns (combined 51 characters). Obviously doesn't need to happen any time soon since it's a comparatively small amount, but just thought I'd mention that.

MartinFlores751:
> And there's nothing depending on that script or invoking it, right?

Nothing shows up in a grep of irod/irods.

MartinFlores751:
> and... tell me a story about someone upgrading directly from 4.2.3 or prior (before this script) to 5.0.0...
> 
> Are we okay?

Would that story be a 4.2.5 (or less) -> 4.3.X -> 5.0? Wolud we have to document a required "stop" at version 4.3.X if you're going to upgrade to 5.0 and beyond?

MartinFlores751:
> As a side note, I think we should consider moving away from adding these known strings to the catalog. It's a small thing, but a catalog with 50 million objects at 2 replicas per object will use over 5GB in the catalog just for the strings in the 3 deprecated columns (combined 51 characters). Obviously doesn't need to happen any time soon since it's a comparatively small amount, but just thought I'd mention that.

Might be my tired brain, but this sounds like make an issue so we don't lose this (unless we already have this somewhere)?

alanking (approved):
Created https://github.com/irods/irods/issues/8276 for changing how we handle the deprecated columns in the future.

`update_deprecated_columns.py` was never run automatically on upgrade. If someone is upgrading from 4.2.3 to 5.0, they may run into this issue. They will either run into this issue and find the script, or adjust their queries to avoid the issue as the columns are unused. I think we're good.

trel:
> > and... tell me a story about someone upgrading directly from 4.2.3 or prior (before this script) to 5.0.0...
> > Are we okay?
> 
> Would that story be a 4.2.5 (or less) -> 4.3.X -> 5.0? Wolud we have to document a required "stop" at version 4.3.X if you're going to upgrade to 5.0 and beyond?

That's the question.   If someone directly upgrades from 4.2.x to 5.0.0+ (no 'stop')... is it going to work as designed/expected?

korydraughn:
> and... tell me a story about someone upgrading directly from 4.2.3 or prior (before this script) to 5.0.0...
> 
> Are we okay?

Running a 4.2.3 server implies an earlier OS and means they will not be able to upgrade to 5.0 without upgrading the OS. On top of that, our upgrade logic applies every intermediate change on the way to the target iRODS version.

For those reasons, I believe we're safe.

korydraughn:
Pound it.

Issue #8008, opened 2024-11-09
Consider removing `status()` function from controller.py and irodsctl


korydraughn:
Admins have `ips` and `ps`. `./irodsctl status` is not needed.

https://github.com/irods/irods/blob/13c08dab315ec05f953f76f3e082d9268a2f7283/scripts/irods/controller.py#L238

We should remove `status()` and anything related to it.

Related to #7996.

trel:
just for posterity...

it's a nice single-stop place to interact with the server...

modeled after apachectl

https://httpd.apache.org/docs/2.4/programs/apachectl.html

korydraughn:
With systemd supported by iRODS, admins now have a way to view all child processes spawned from the server.

And given that `irodsctl` no longer exists, there's little reason to keep the `status` function around.

korydraughn:
Searching the repo for calls to `.status()` shows that nothing uses it.

```bash
$ grep -Irn '\<status\>()'
server/api/src/rsDataObjTrim.cpp:47:        rstrcpy(info.statusString, obj.status().c_str(), sizeof(info.statusString));
server/api/src/rsDataObjOpen.cpp:1024:            replica.status()));
server/api/src/rsDataCopy.cpp:128:        return err.status();
server/api/src/rsDataCopy.cpp:135:        return err.status();
server/main_server/src/main.cpp:476:                        log_server::error("{}: {}", __func__, err.status());
server/main_server/src/main.cpp:479:                        fmt::print(stderr, "{}: {}\n", __func__, err.status());
server/core/src/dataObjOpr.cpp:514:             snprintf(data_obj_info->statusString, NAME_LEN, "%s", replica.status().c_str());
server/core/src/rsApiHandler.cpp:467:        return err.status();
server/core/src/rsApiHandler.cpp:474:        return err.status();
server/core/include/irods/irods_physical_object.hpp:69:            inline std::string status()          const {
plugins/microservices/src/update_unixfilesystem_resource_free_space.cpp:47:        return resource_manager_resolve_error.status();
plugins/microservices/src/update_unixfilesystem_resource_free_space.cpp:55:        return get_resource_type_error.status();
plugins/microservices/src/update_unixfilesystem_resource_free_space.cpp:68:        return get_resource_vault_path_error.status();
unit_tests/src/test_resource_administration.cpp:138:        REQUIRE(info->status() == adm::resource_status::unknown);
[...]

PR #8280, opened 2025-03-12
[#8008] Remove status function from IrodsController (main)

korydraughn:
Searching the repo for calls to `.status()` shows that nothing uses it.


korydraughn:
All unit and core tests passed.

Pounded.

Issue #7957, opened 2024-09-13
Absorb functionality of the Metadata Guard rule engine plugin into the server


korydraughn:
- [ ] main
- [ ] 4-3-stable
---
## Enhancement

The Metadata Guard rule engine plugin adds functionality that should just be a part of the iRODS server.

We need to investigate what it would take to make that happen.

This would remove the need for the metadata guard plugin. This also potentially improves performance of the server because the plugin logic lives directly in the server.

alanking:
Will need to include a page for metadata_guard in irods_docs as well, I'd imagine.

korydraughn:
Yes, please.

PR #8289, opened 2025-03-13
[#7957] Absorb metadata_guard REP (main)

alanking:
In service of #7957

Companion PR: https://github.com/irods/irods_testing_environment/pull/253

Made some tweaks from the code in the plugin repo which I will list here:
 - iRODS header includes get quotation marks instead of angle brackets
 - Sorted include files
 - Define "server side API" macros for `irods_query.hpp` and `user_administration.hpp`
 - Ran clang-format on the file
 - Added setup/teardown operation hooks (no-ops; can squash this commit if desired)

Core tests are running now.


scripts/irods/test/test_rule_engine_plugin_metadata_guard.py
```diff
@@ -0,0 +1,328 @@
+from __future__ import print_function
```
korydraughn, line 1: This can be removed.

Or we open an issue to remove python 2 code for a later release.

korydraughn (commented):
Good stuff.

We'll need an issue for optimizing things - i.e. moving the plugin logic into the API calls directly. That can happen in a later release though.

Will circle back once the test results are complete.

alanking:
I got `test_rule_engine_plugin_metadata_guard` to pass, but I can't get `test_rule_engine_plugin_metadata_guard_atomic` to run. I *think* this is because PRC is imported as the `irods` package and our tests live in `scripts/irods/test`, so it's getting confused. The reason it works for the plugin now is because it is run separately via the test hook and the script is getting installed in a different location from the rest of the test suite (it also seems to not depend on any of our usual test stuff). Not sure how we want to handle this... Also, not sure why we had to implement these tests using PRC to begin with...

Maybe we just remove it for now and circle back to this when atomic metadata operations are available via an iCommand or something?

scripts/irods/test/test_rule_engine_plugin_metadata_guard.py
alanking: Done

alanking:
I suppose once the metadata_guard behavior is properly integrated with the server, it won't need to rely on dynamic PEPs anymore and we will be able to write tests using rules which call the atomic metadata microservices.

korydraughn:
Correct. The PRC was used because we didn't have a tool which could trigger the atomic API PEPs. Since then, we've started building tools for testing which cover such cases (e.g. the storage tiering tools you wrote not too long ago). We can do the same thing here and remove the dependency on the PRC. Doing that would require rewriting the tests too.

As you said, this may have to wait until after 5.0.

alanking:
It just occurred to me that we may be able to write C++ unit tests for this now that it's in the server and the unit test framework is available. Anyway, I'm going to split out the atomic operations tests into a separate issue and remove that file from this PR so we can move this along.

korydraughn:
Good idea!

alanking:
Created https://github.com/irods/irods/issues/8293 for the atomic metadata operations tests and removed the python tests for atomic metadata operations.

plugins/rule_engines/src/metadata_guard.cpp
```diff
@@ -0,0 +1,368 @@
...
+        const auto gql = fmt::format("select META_COLL_ATTR_VALUE "
+                                     "where META_COLL_ATTR_NAME = 'irods::metadata_guard' and COLL_NAME = '/{}'",
+                                     _rei.rsComm->myEnv.rodsZone);
+
+        if (irods::query q{_rei.rsComm, gql}; q.size() > 0) {
+            try {
+                return json::parse(q.front()[0]);
+            }
+            catch (const json::exception&) {
+                const char* msg = "Cannot parse Rule Engine Plugin configuration";
```
trel, line 66: for your consideration...

```suggestion
                const char* msg = "Cannot parse Metadata Guard configuration";
```

plugins/rule_engines/src/metadata_guard.cpp
```diff
@@ -0,0 +1,368 @@
...
+        -> irods::error
+    {
+        try {
+            auto* input = boost::any_cast<modAVUMetadataInp_t*>(*std::next(std::begin(_rule_arguments), 2));
+
+            if (strcmp(input->arg0, "rmw") == 0) {
+                log_re::error({{"log_message", "rmw is deprecated and its usage is disallowed by metadata_guard"},
+                               {"rule_engine_plugin", "metadata_guard"}});
+                return ERROR(SYS_NOT_ALLOWED, "rmw is deprecated and its usage is disallowed by metadata_guard");
+            }
```
trel, line 161: may not be necessary in 5

scripts/irods/test/test_rule_engine_plugin_metadata_guard.py
```diff
@@ -0,0 +1,266 @@
...
+
+        with lib.file_backed_up(config.server_config_path):
+            self.enable_rule_engine_plugin(config)
+
+            # This should cause an exception to be thrown and logged, but users
+            # should still be able to proceed. Verify that the log contains a valid
+            # error message.
+            log_offset = lib.get_file_size_by_path(paths.server_log_path())
+            self.rods.assert_icommand(['imeta', 'set', '-C', root_coll, self.metadata_guard_attribute_name(), json_config])
+            count = lib.count_occurrences_of_string_in_log(paths.server_log_path(), 'Cannot parse Rule Engine Plugin configuration', log_offset)
```
trel, line 47:
```suggestion
            count = lib.count_occurrences_of_string_in_log(paths.server_log_path(), 'Cannot parse Metadata Guard configuration', log_offset)
```

FifthPotato (commented):
Looks good to me! 
(...I mean, I remember writing some of this, so does this count as me being biased?)

korydraughn:
> Looks good to me! (...I mean, I remember writing some of this, so does this count as me being biased?)

No. Given you were the last person to work on and improve the plugin, your opinion is valuable. Ask all the questions and raise concerns if you have any.

FifthPotato:
It looks pretty similar to how I left it, so (at the risk of sounding presumptuous) I think it looks good.

korydraughn:
Don't forget to mark it as approved if you feel it's good. Makes your stamp of approval more visible.

FifthPotato (approved):

Issue #8300, opened 2025-03-18
Remove osauth scheme


alanking:
- [x] main only

---

### Removal

The osauth scheme is a legacy auth scheme which can be completely replaced with PAM. We have been contemplating removal of this scheme for some time: https://groups.google.com/g/irod-chat/c/sSlBMvI-21o/m/fP7t6b4QAgAJ It is time to remove it.

alanking:
Note: Legacy auth plugins are deprecated in 5.0, so nobody should continue using this anyway. See #8257

PR #8305, opened 2025-03-19
[#8300] Remove osauth and genOSAuth (main)

alanking:
Addresses #8300 

Still need to run tests. We do not have tests for osauth scheme and do not use it anywhere, so I don't anticipate any failures provided everything builds and the server can stand up.


korydraughn:
Have you checked to see if the icommands contain anything related to osauth? I don't expect they would.

alanking:
> Have you checked to see if the icommands contain anything related to osauth? I don't expect they would.

I didn't see anything related to osauth in iCommands nor irods_docs. Also, iCommands build, so that's good.

alanking:
Unit tests passed. Core tests are running now.

alanking:
Unless, of course, we consider the server standing up as good enough...

korydraughn (approved):
If you're confident you removed all references to osauth in the tests, then I say go ahead and pound/merge it.

Issue #8259, opened 2025-03-06
GenQuery2 does not consider group permissions


korydraughn:
- [x] main
- [x] 4-3-stable
---
## Bug Report

The SQL generated by GenQuery2 does not take into account that a user may have permission to a data object or collection through a group.

### Example

```bash
$ iuserinfo
name: kory
id: 10014
type: rodsuser
zone: tempZone
info: 
comment: 
create time: 01739996609: 2025-02-19.15:23:29
modify time: 01741270385: 2025-03-06.09:13:05
member of group: kory
member of group: public
member of group: g0
$ itouch thefile
$ ichmod read g0 thefile
$ ils -A thefile
  /tempZone/home/kory/thefile
        ACL - g:g0#tempZone:read_object   kory#tempZone:own   
$ ichmod null kory thefile
$ ils -A thefile
  /tempZone/home/kory/thefile
        ACL - g:g0#tempZone:read_object   
$ # This query should return the data object since "kory" has read
$ # permission through the group.
$ iquery "select DATA_NAME where COLL_NAME = '/tempZone/home/kory' and DATA_NAME = 'thefile'"
[]
```

Here's the SQL created by GenQuery2.
```
$ iquery "select DATA_NAME where COLL_NAME = '/tempZone/home/kory' and DATA_NAME = 'thefile'" --sql-only | pg_format -
SELECT DISTINCT
    t0.data_name
FROM
    R_DATA_MAIN t0
    INNER JOIN R_COLL_MAIN t1 ON t1.coll_id = t0.coll_id
    INNER JOIN R_OBJT_ACCESS pdoa ON t0.data_id = pdoa.object_id
    INNER JOIN R_TOKN_MAIN pdt ON pdoa.access_type_id = pdt.token_id
    INNER JOIN R_USER_MAIN pdu ON pdoa.user_id = pdu.user_id
    INNER JOIN R_OBJT_ACCESS pcoa ON t1.coll_id = pcoa.object_id
    INNER JOIN R_TOKN_MAIN pct ON pcoa.access_type_id = pct.token_id
    INNER JOIN R_USER_MAIN pcu ON pcoa.user_id = pcu.user_id
WHERE
    t1.coll_name = ?
    AND t0.data_name = ?
    AND pdu.user_name = ?
    AND pdu.zone_name = 'tempZone'
    AND pcu.user_name = ?
    AND pcu.zone_name = 'tempZone'
    AND pdoa.access_type_id >= 1050
[...]

PR #8308, opened 2025-03-20
[#8259] Make GenQuery2 honor group permissions (4-3-stable)

korydraughn:
Cherry-pick of #8298.


alanking (approved):
Looks identical. My understanding is that GenQuery2 is identical between the two active branches, so that's good. # it.

korydraughn:
That's correct.

Added pound.

Issue #8335, opened 2025-03-27
Test involving `ips -v` failing in topology environment


korydraughn:
## Bug Report

The following assertion is failing consistently for topology tests + consumer. `ips` does not appear in the output.

https://github.com/irods/irods/blob/cf8c731db744ac08dea98981fba836cf1bd4e67b/scripts/irods/test/test_chunkydevtest.py#L58

This is what I'm seeing:
```
]]></system-out>                                     
        </testcase>                    
        <testcase classname="irods.test.test_resource_types.Test_Resource_CompoundWithMockarchive" name="test_beginning_from_devtest" time="16.621" timestamp="2025-03-27T15:43:40" file="scripts/irods/test/test_resource_types.py" line="25">
                <failure type="AssertionError" message=""><![CDATA[Traceback (most recent call last):
  File "/var/lib/irods/scripts/irods/test/test_chunkydevtest.py", line 58, in test_beginning_from_devtest
    self.admin.assert_icommand("ips -v", 'STDOUT_SINGLELINE', "ips")        
  File "/var/lib/irods/scripts/irods/test/session.py", line 162, in assert_icommand
    return assert_command(*args, **kwargs)
  File "/var/lib/irods/scripts/irods/test/command.py", line 76, in assert_command                                                                                                             
    return _assert_helper(*args, should_fail=False, **kwargs)
  File "/var/lib/irods/scripts/irods/test/command.py", line 104, in _assert_helper
    assert result
AssertionError
]]></failure>
                <system-out><![CDATA[ --- IrodsSession: icommand executed by [rods#tempZone] [iinit] --- 
Assert Command: iinit
Expecting STDOUT_SINGLELINE: ['']
  stdout:
    | WARNING: Error 25 disabling echo mode. Password will be displayed in plaintext.
    | Enter your current iRODS password:
  stderr:
    | 
Output found

 --- IrodsSession: icommand executed by [rods#tempZone] [iadmin modresc demoResc name origResc] --- 
Assert Command: iadmin modresc demoResc name origResc
Expecting STDOUT_SINGLELINE: ['rename']
  stdout:
    | If you modify a resource name, you and other users will need to
    | change your irods_environment.json files to use it, you may need to update
    | server_config.json and, if rules use the resource name, you'll need to
    | update the core rules (core.re).  This command will update various
    | tables with the new name.
--
Assert Command: ierror -14000
Expecting STDOUT_SINGLELINE: ['SYS_API_INPUT_ERR']
  stdout:
[...]

PR #8344, opened 2025-03-30
[#8335] Store client name in environment on agent initialization (main)

korydraughn:
Tested manually in a topology environment.

Will need to run topology/consumer tests to confirm the reported test failure is resolved.

The removal of dead code is bonus. I'm pretty sure that code path is dead due to the redesign of the startup/shutdown process for iRODS 5.


korydraughn:
All topology/consumer tests passed.
```
-----
All tests passed! :)
time elapsed: [9841.7950]seconds ([   2]hours [44.0299]minutes)

2025-03-30 02:40:32,151 - version:[4.90.0]
2025-03-30 02:40:32,151 - sha:[2482446cbe96af474ff7884057fbbade4e5457de]
```

server/core/src/initServer.cpp
```diff
@@ -723,121 +723,12 @@ int initRsCommWithStartupPack(rsComm_t* rsComm, startupPack_t* startupPack, bool
...
-            rstrcpy( rsComm->option, tmpStr, LONG_NAME_LEN );
-        }
 
-        if (std::getenv(irods::RODS_CS_NEG)) {
-            require_cs_neg = true;
+        // Set the SP_OPTION environment variable so that server-to-server connections
+        // forward the name of the connected client to remote servers. This information
+        // is used by the rcConnect library when establishing a connection to a server.
+        if (setenv(SP_OPTION, rsComm->option, /* overwrite */ 1) != 0) {
+            log_agent::warn("{}: Could not set environment variable [{}] for ips.", __func__, SP_OPTION);
```
trel, line 731: should we log the attempted `rsComm->option` value on failure as well?

korydraughn: I think that's what this line is doing. I feel I'm misunderstanding the question.

korydraughn: Oh oh, I see now. You're talking about logging the value of `->option`.

korydraughn: Sure, that sounds like a good idea.

trel: right, in case that's another clue for the admin.

korydraughn: Done.

server/core/src/initServer.cpp
```diff
@@ -723,121 +723,13 @@ int initRsCommWithStartupPack(rsComm_t* rsComm, startupPack_t* startupPack, bool
...
-        }
 
-        if (std::getenv(irods::RODS_CS_NEG)) {
-            require_cs_neg = true;
+        // Set the SP_OPTION environment variable so that server-to-server connections
+        // forward the name of the connected client to remote servers. This information
+        // is used by the rcConnect library when establishing a connection to a server.
+        if (setenv(SP_OPTION, rsComm->option, /* overwrite */ 1) != 0) {
+            log_agent::warn(
+                "{}: Could not set environment variable [{}] to [{}] for ips.", __func__, SP_OPTION, rsComm->option);
```
korydraughn, line 732: Should probably `ips`. It's not wrong, but it's more about the API powering `ips`.

korydraughn:
All unit tests passed.
All core tests passed except test_native_rule_engine_plugin. This is expected.

server/core/src/initServer.cpp
korydraughn: Updated.

server/core/src/initServer.cpp
```diff
@@ -723,121 +723,15 @@ int initRsCommWithStartupPack(rsComm_t* rsComm, startupPack_t* startupPack, bool
...
-        if (std::getenv(irods::RODS_CS_NEG)) {
-            require_cs_neg = true;
+        // Set the SP_OPTION environment variable so that server-to-server connections
+        // forward the name of the connected client to remote servers. This information
+        // is used by the rcConnect library when establishing a connection to a server.
+        if (setenv(SP_OPTION, rsComm->option, /* overwrite */ 1) != 0) {
+            log_agent::warn("{}: Could not set environment variable [{}] to [{}] for agent tracking APIs.",
+                            __func__,
+                            SP_OPTION,
+                            rsComm->option);
```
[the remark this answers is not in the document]
korydraughn: I've been wondering the same thing. Everything tells me it's safe to remove the `if (startupPack)` altogether, but I haven't pulled the trigger since that would lead to more changes which were out of scope of the target issue.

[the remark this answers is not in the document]
alanking: Okay. It seems to me that the server may not be in a valid state if the `else` case is not handled, so we might as well return an error or something. Just a thought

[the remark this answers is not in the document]
korydraughn: Went ahead and updated the function's design/implementation since it was trivial and also internal to the server.

korydraughn:
Unit tests pass.

The latest commit doesn't require rerunning the other test suites since `initRsCommWithStartupPack` is invoked during agent initialization.

alanking (commented):
Seems good. Squash em

korydraughn:
Squashed and pounded (by accident).

alanking (approved):
Still seems good. Mergin.

Issue #8260, opened 2025-03-06
Make server track access time


korydraughn:
## Enhancement

iRODS 5 needs to track access time. Doing that enables users to make more informed decisions about their data.

This requires the following:
- Adding a new column to R_DATA_MAIN - i.e. access_ts
- Adding support for the new column to GenQuery1 and GenQuery2

As for when/how to carry out the updates, we're leaning toward a deferred approach. That is, write the time of access down somewhere and update the catalog later. The motivation behind this is to reduce the number of catalog updates when the system is experiencing a high frequency of file accesses.

With that, there's some room for allowing administrators to configure the behavior of the system. Those options are likely going to be exposed as zone-wide options via the R_GRID_CONFIGURATION table.

One challenge for this is figuring out how to get the updates from a consumer into the catalog. Is it possible to do that without introducing another API endpoint?

trel:
I can see two configuration knobs off the top of my head (names are very preliminary)...
 - atime_batch_wakeup_time_in_seconds
    - how often the batch processor wakes up
 - atime_ignore_window_in_seconds
    - throws away any updates if existing atime is recent enough

trel:
> how to get the updates from a consumer into the catalog

if the queue only lives on the catalog provider... this is moot?

korydraughn:
That's true if all calls to open() hit the catalog provider.

@alanking Does opening a replica for reading result in a trip to the provider?

alanking:
If you mean DataObjOpen / replica_open, at least one catalog hit is incurred to get information about the data object being opened to populate the L1 descriptor.

korydraughn:
> If you mean DataObjOpen / replica_open, at least one catalog hit is incurred to get information about the data object being opened to populate the L1 descriptor.

Does that happen before or after the replica is actually opened?

alanking:
The catalog information is fetched before the physical open occurs.

korydraughn:
Got it. I was hoping you were going to say rsDataObjOpen updates the catalog after the physical open. I was thinking open() adjusted the replica status to say read-lock/write-lock on success.

korydraughn:
Hmm, I think we already have the means to do this without a new API endpoint (which is what I want to avoid).

Here's what I'm thinking:
- rxModDataObjMeta can be updated to support modification of the atime
- The main loop of the server maintains a queue of updates locally and when it's time to process the queue, it just invokes rxModDataObjMeta with the new atime

There are a few issues with this approach:
[...]

PR #8350, opened 2025-03-31
[#8260] Remove unnecessary assertion from access time test (main)

korydraughn:


alanking (approved):
I guess the goal mentioned in the commit message is in the comment of the removed code, so that seems good enough to me. \# it if test passes

korydraughn:
Pounded.

PR #8352, opened 2025-03-31
[7904] Remove server monitoring microservices (main)

FifthPotato:
As titled. Tests pass after removing relevant rule tests from icommands.


server/re/src/reIn2p3SysRule.cpp
```diff
@@ -39,133 +36,6 @@ static pthread_mutex_t my_mutex;
...
-        }
-    }
-
-    clearGenQueryInp( &genQueryInp );
-    freeGenQueryOut( &genQueryOut );
-
-    return 0;
-}
-
 int getListOfResc( rsComm_t *rsComm, char serverList[MAX_VALUE][MAX_NAME_LEN], int nservers,
```
korydraughn, line 39: Does anything use this function?

server/re/src/reIn2p3SysRule.cpp
korydraughn, line 1: Does anything use these?
- `MON_CFG_FILE`
- `OUTPUT_MON_PERF`

korydraughn, line 1: Are all the header includes being used?

korydraughn (commented):
Very nice.

server/re/src/reIn2p3SysRule.cpp
FifthPotato: As far as I can tell, no: recursive grep gives me nothing in either irods or icommands. Can remove if you also think it's not needed.

korydraughn: Yes, please remove anything that's being used. Add them as additional commits so the work you tested stays clean.

You can also grep the plugins for usage.

alanking (commented):
One open convo remaining

FifthPotato:
> One open convo remaining

Yeah, was just cleaning up the includes. Should be all resolved now?

alanking (commented):
I think this is ready for squashing.

korydraughn (commented):
Does it pass the C++ unit tests?

FifthPotato:
> Does it pass the C++ unit tests?

Just ran them-- all green.

korydraughn:
Squash to taste. No pounds.

FifthPotato:
Squished.

alanking (approved):
\# it

korydraughn (approved):

Issue #8257, opened 2025-03-05
Deprecate legacy auth plugins


alanking:
- [x] Remove GSI plugin stuff (last released version compatible with iRODS 4.2.6)
- [x] Remove Kerberos plugin stuff (last released version compatible with iRODS 4.2.12)
- [x] Remove legacy auth support from iinit
- [x] Document deprecation of legacy auth plugins, schemes, and API endpoints
- [x] Deprecate clientLoginWithPassword
- [x] Deprecate clientLoginPam
- [x] Deprecate clientLoginTTL
- [x] Replace clientLoginWithPassword in HTTP API (see https://github.com/irods/irods_client_http_api/issues/410)
- [x] Replace clientLoginWithPassword in S3 API (see https://github.com/irods/irods_client_s3_api/issues/142)
- [x] Deprecate clientLogin
- [x] Create new C library function to replace clientLogin for C clients
- [x] Replace clientLogin usage
- [x] #8408

---

The plugins under `plugins/auth_legacy` preceded the new auth plugins which use the new authentication plugin framework under `plugins/auth`. These were kept in order to give client libraries time to implement the new authentication plugin framework in their respective languages.

Now that the new auth plugins have proven themselves useful and stable, we should deprecate the legacy auth plugins.

Questions:
1. Does this deprecation include the auth plugin interface itself?
2. Should we return a deprecation message on the rError stack, or just deprecate types and such?
3. Should we deprecate `clientLogin*`?

alanking:
In support of deprecating `clientLoginTTL`, we have decided that the native auth client-side plugin will be responsible for generating the limited password and writing it out to the irodsA file in the event that TTL is used. This will require pam_password and pam_interactive to remove the TTL key from the request payload before using native authentication because they already record the randomly generated password in the irodsA file and do not use limited passwords. [1]

Because the plugin will record this information, we have also decided that the plugin will record an irodsA file for the "real" password as well. This is a departure for the historical behavior of the iCommands which only recorded irodsA files via `iinit`. If an iCommand was run with no irodsA file, the user would be prompted for a password, and that authentication would be good only for that iCommand call and no irodsA file would be recorded. Now, we are proposing that an irodsA file be recorded for any iCommand. This will also make native authentication behave more like pam_password and pam_interactive.

As a result, `iinit` will no longer need to know how / when to call `clientLoginTTL` nor will it be responsible for recording the irodsA file, making it much simpler and auth scheme-agnostic.

[...]

PR #8356, opened 2025-04-07
Change plugin dir from "auth" to "authentication", remove unsupported Kerberos and GSI auth schemes (main)

alanking:
Addresses #7937 
In service of #8257

Companion PR: https://github.com/irods/irods_client_icommands/pull/554
Companion PR: https://github.com/irods/irods_auth_plugin_pam_interactive/pull/63

Core tests passed

#8257 will involve more changes, but this one gets things started by removing some of the unsupported auth schemes and affiliated machinery. For reference, Kerberos hasn't shipped for 4.3, and GSI was last supported in iRODS 4.2.6. Wasn't sure how deep to cut (configurations, database tables, etc.) so I just started with the supporting code since this will have no effect. *There is no plugin to load so we don't need to deprecate these* - they are effectively deprecated already.


alanking:
Also, the only clang-format changes in this PR are on lines which I didn't necessarily touch, so I'm going to ignore these unless we just want to make a separate clang-format commit.

lib/core/include/irods/authentication_plugin_framework.hpp
```diff
@@ -162,6 +162,10 @@ namespace irods::experimental::auth
     /// \since 4.3.0
     inline void authenticate_client(RcComm& _comm, const RodsEnvironment& _env, const json& _ctx)
     {
+        if (_comm.loggedIn) {
+            return;
+        }
+
```
korydraughn, line 168: What's the reason for adding this? Did you notice something?

alanking: I had made some more extensive changes which were not included in this PR. I think this change either fixed an issue related to those changes or was just here as a way to save a few cycles. I left it in a separate commit in case anybody objected to its inclusion.

korydraughn: Seems fine to me.

This is about silently ignoring the call to `authenticate_client` if already logged in, correct? This is purely on the client side of things, right?

alanking: I would not say that it is pure because this is supposed to serve a similar purpose to `clientLogin`. Servers could call it when redirecting as well.

I suppose it technically causes the first operation to be skipped, but the plugins we have now would not really be affected. See the control loop below.

korydraughn: Got it.

korydraughn: One more question - do you feel that check is necessary/good?

alanking: It's an optimization - not necessary, but elides a (probably) unnecessary call into the plugin operation. Even if the `rcComm.loggedIn` flag is set erroneously, it would only get to call the start operation of the plugin below before returning and would do nothing.

korydraughn: Good enough for me.

alanking:
#'d, mergin

Issue #8362, opened 2025-04-08
`irods.lib.replica_exists` should escape single quotes in logical paths


alanking:
- [x] main

---

The GenQuery1 parser now requires escaping single quotes before using them as arguments in a query. Many of our testing utilities need to be updated to make sure the paths being used in their queries are properly escaped. This can be done by doubling up the single quotes / apostrophes, or replacing with its hexadecimal value in ASCII, `\x27`.

The title of this issue refers to `replica_exists` in `scripts/irods/lib.py`, but there are many other utilities which we may want to update.

Discovered this while working on https://github.com/irods/irods_rule_engine_plugin_logical_quotas/issues/125.

PR #8364, opened 2025-04-09
Escape single quotes in lib.py utilities, more ignoring -Wunused-but-set-variable from Boost (main)

alanking:
Addresses #8362

Unit and core tests passed


lib/core/src/dns_cache.cpp
```diff
@@ -1,14 +1,17 @@
...
 #include <boost/interprocess/containers/map.hpp>
 #include <boost/interprocess/containers/string.hpp>
-#include <boost/interprocess/containers/list.hpp>
-#include <boost/interprocess/allocators/allocator.hpp>
+#include <boost/interprocess/managed_shared_memory.hpp>
+#include <boost/interprocess/shared_memory_object.hpp>
 #include <boost/interprocess/sync/named_sharable_mutex.hpp>
 #include <boost/interprocess/sync/scoped_lock.hpp>
 #include <boost/interprocess/sync/sharable_lock.hpp>
+#pragma GCC diagnostic pop
```
korydraughn, line 14: Try wrapping only the following headers and see if that covers the compiler warning.

From the s3 resource plugin.

https://github.com/irods/irods_resource_plugin_s3/blob/55433030523b5ca8b290db39158a1bbe85340866/s3_transport/include/irods/private/s3_transport/multipart_shared_data.hpp#L9-L16

```cpp
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wunused-but-set-variable"
#include <boost/interprocess/allocators/allocator.hpp>
#include <boost/interprocess/containers/list.hpp>
#include <boost/interprocess/containers/map.hpp>
#include <boost/interprocess/containers/string.hpp>
#include <boost/interprocess/managed_shared_memory.hpp>
#pragma GCC diagnostic pop
```

korydraughn (commented):
Looks good.

If the build fails after making the proposed edits, then we'll move forward with your solution which wrapped all boost headers.

korydraughn (commented):
Excellent. Squash to taste.

alanking:
Squashed

korydraughn:
Is the commit message body for the boost headers still correct/true?

alanking:
Eh, not really. Fixed.

korydraughn (approved):
Pound it.

alanking:
#'d, mergin

Issue #8331, opened 2025-03-26
Switch to distro-provided catch2


SwooshyCueb:
Opening this here instead of in externals so that I can put it in a milestone.

iRODS 5.0 will drop support for EL8, Debian 11, and Ubuntu 20.04. Therefore, all distros we support for iRODS 5+ have a new-enough version of catch2 in their repositories. Let's switch.

Related: #7765

Issue #8330, opened 2025-03-26
Switch to distro-provided CMake


SwooshyCueb:
Opening this here instead of in externals so that I can put it in a milestone.

iRODS 5.0 will drop support for EL8, Debian 11, and Ubuntu 20.04. Therefore, all distros we support for iRODS 5+ have a newer version of CMake than is present in externals. Let's switch.

Supersedes irods/externals#265

Issue #8102, opened 2024-12-18
Adjust usage of `cmake_minimum_required`


SwooshyCueb:
In several locations we call `cmake_minimum_required` to set policy switches automatically. With newer versions of CMake, some of these calls are producing errors, as compatibility with older versions of CMake (older than 3.10) are slated to be torched soon. Let's adjust our usage of this function and make any other changes needed.

Issue #7765, opened 2024-06-04
Catch2 v3 compatibility


SwooshyCueb:
Catch2 changed a lot for v3. If we want to use distro-provided catch2 packages, we should have compatibility with both.
A [migration guide](https://github.com/catchorg/Catch2/blob/4e8d92bf02f7d1c8006a0e7a5ecabd8e62d98502/docs/migrate-v2-to-v3.md) has been provided.

PR #8366, opened 2025-04-09
[#6087,#7726,#7765,#8102,#8330,#8331,#8332,#8423,#8462,#8463] Migrate to system CMake, nlohmann-json, catch2, fmt, and spdlog (main)

SwooshyCueb:
In service of #6087
In service of #7726
In service of #7765
In service of #8102
In service of #8330
In service of #8331
In service of #8332
In service of #8423
In service of #8462
In service of #8463


irods_consortium_continuous_integration_build_hook.py
alanking, line 1: Should we just... remove this file? We don't seem to be using it in any of our existing dev processes.

alanking (commented):
~Didn't see anything unu~

EDIT: GitHub saved my unfinished thought before I changed my mind... :(

irods_consortium_continuous_integration_build_hook.py
korydraughn: Sounds like the right move to me. I don't see us leaning on it in the future. If we ever need it, we can always restore it.

I say let's remove it.

plugins/database/src/low_level_odbc.cpp
```diff
@@ -218,9 +218,9 @@ cllConnect( icatSessionStruct *icss ) {
         SQLSMALLINT length;
         while ( SQLError( icss->environPtr, myHdbc , 0, sqlstate, &sqlcode, buffer,
                           SQL_MAX_MESSAGE_LENGTH + 1, &length ) == SQL_SUCCESS ) {
-            log_db::error("{}:          SQLSTATE: {}\n", __func__, sqlstate);
+            log_db::error("{}:          SQLSTATE: {}\n", __func__, reinterpret_cast<char*>(sqlstate));
             log_db::error("{}:  Native Error Code: {}\n", __func__, sqlcode);
-            log_db::error("{}: {} \n", __func__, buffer);
```
SwooshyCueb, line 223: There is almost certainly a better way to do this.

plugins/database/src/low_level_odbc.cpp
```diff
@@ -218,9 +218,9 @@ cllConnect( icatSessionStruct *icss ) {
...
+            log_db::error("{}: {} \n", __func__, reinterpret_cast<char*>(buffer));
```
SwooshyCueb, line 223: There is almost certainly a better way to do this.

lib/core/src/irods_client_negotiation.cpp
```diff
@@ -60,7 +60,7 @@ namespace irods
 
                 const auto hostname_in_provider_hosts_list = std::any_of(
                     std::cbegin(*provider_hosts), std::cend(*provider_hosts), [&_host_name](const auto& _host) {
-                        return _host_name == _host;
+                        return _host_name == static_cast<std::string>(_host);
```
SwooshyCueb, line 63: Should I call `_host.get_ref<std::string>()` instead?

korydraughn: Use `_host.get_ref<const std::string&>()`.

Here's an example of the JSON structure. Taken from https://docs.irods.org/4.3.4/system_overview/federation/#zoneas-server_configjson.
```json
"federation": [
    {
        "catalog_provider_hosts": ["zoneB-provider.example.org"], 
        "negotiation_key": "_____32_byte_pre_shared_key_____",
        "zone_key": "ZONE_KEY_FOR_ALL_OF_ZONE_B", 
        "zone_name": "ZoneB", 
        "zone_port": 2247
    }
]
```

plugins/database/src/low_level_odbc.cpp
korydraughn: What type is `SQLCHAR` aliasing?

SwooshyCueb: unixODBC defines it as `unsigned char`, but my understanding is that this is implementation-defined.

SwooshyCueb: Worth noting:

I also tried
```c++
        SQLSCHAR sqlstate[SQL_SQLSTATE_SIZE + 1];
        SQLINTEGER sqlcode;
        SQLSCHAR buffer[SQL_MAX_MESSAGE_LENGTH + 1];
        SQLSMALLINT length;
        while ( SQLError( icss->environPtr, myHdbc , 0, renterpret_cast<SQLCHAR*>(sqlstate), &sqlcode, renterpret_cast<SQLCHAR*>(buffer),
                          SQL_MAX_MESSAGE_LENGTH + 1, &length ) == SQL_SUCCESS ) {
            log_db::error("{}:          SQLSTATE: {}\n", __func__, sqlstate);
            log_db::error("{}:  Native Error Code: {}\n", __func__, sqlcode);
            log_db::error("{}: {} \n", __func__, buffer);
        }
```
(`SQLSCHAR` being the explicitly signed variant of `SQLCHAR`)
but fmt didn't like this either, gave more or less the same error message, just said `signed char` where it previously said `unsigned char`.

Even if it had worked, I don't think it would've been the ideal solution either, just perhaps more ideal than the one currently in the PR.

irods_consortium_continuous_integration_build_hook.py
SwooshyCueb: I've opened #8421 for this

plugins/database/src/low_level_odbc.cpp
korydraughn: I suspect the `reinterpret_cast` is fine (for now) since you can always inspect the representation of any object using a char pointer. As I understand it, the cast is effectively short hand for:
```cpp
static_cast<T*>(static_cast<void*>(expression))
```

I'm going to do a little reading about this to get a better picture of the situation.

These may be useful:
- https://learn.microsoft.com/en-us/sql/odbc/reference/appendixes/sqlerror-mapping?view=sql-server-ver16
- https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqlgetdiagrec-function?view=sql-server-ver16
- https://learn.microsoft.com/en-us/sql/odbc/reference/appendixes/c-data-types?view=sql-server-ver16
- https://en.cppreference.com/w/cpp/language/reinterpret_cast
- https://en.cppreference.com/w/cpp/language/reinterpret_cast#Type_accessibility
- https://en.cppreference.com/w/cpp/language/implicit_conversion#Similar_types

SwooshyCueb: I suspect fmt just wants a specialization of some formatter class or method to be defined.

lib/core/src/irods_client_negotiation.cpp
SwooshyCueb: This worked in the development environment, but does not work in CI:
```
/__w/irods/irods/lib/core/src/irods_client_negotiation.cpp:63:52: error: use 'template' keyword to treat 'get_ref' as a dependent template name
                        return _host_name == _host.get_ref<const std::string&>();
                                                   ^
                                                   template 
```

That's concerning.

plugins/database/src/low_level_odbc.cpp
SwooshyCueb: One of my concerns is that there may be some configuration where `SQLCHAR` is defined to some type with a different size from `char`. If you know this isn't ever the case, though, I'd feel better about keeping the `reinterpet_cast`.

korydraughn: Totally agree.

I think we continue with `reinterpret_cast` until we encounter a platform which presents an issue.

For now, you can introduce a `static_assert` to catch size issues at compile-time.

SwooshyCueb:
Core and unit tests passing for all distros.
I've made all the clang-format-suggested changes I intend to.
Ready for review.

.clang-tidy
```diff
@@ -116,5 +121,11 @@ CheckOptions:
     value:           'false'
   - key:             google-readability-function-size.StatementThreshold
     value:           '800'
+  - key:             cppcoreguidelines-narrowing-conversions.WarnOnIntegerToFloatingPointNarrowingConversion
+    value:           'true'
```
korydraughn, line 125: The docs say this is true by default. Why is it necessary to explicitly set this?

https://releases.llvm.org/16.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/cppcoreguidelines/narrowing-conversions.html

.clang-tidy
```diff
@@ -116,5 +121,11 @@ CheckOptions:
...
+  - key:             readability-function-cognitive-complexity.DescribeBasicIncrements
+    value:           'false'
```
korydraughn, line 127: Why disable this? Does it improve the report in some way?

.clang-tidy
```diff
@@ -116,5 +121,11 @@ CheckOptions:
...
+  - key:             readability-function-cognitive-complexity.IgnoreMacros
+    value:           'true'
```
korydraughn, line 129: Why enable this? Does it improve the report in some way?

unit_tests/src/catch2_compat_include/catch2/catch_all.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+#include <catch2/catch.hpp>
+
+// NOLINTBEGIN(misc-unused-alias-decls)
+
+namespace Catch
+{
+    namespace Matchers
+    {
+
```
korydraughn, line 12: Remove empty line.

unit_tests/src/catch2_compat_include/catch2/catch_all.hpp
```diff
@@ -0,0 +1,28 @@
+#ifndef IRODS_CATCH2_V2_SHIM_H
+#define IRODS_CATCH2_V2_SHIM_H
```
korydraughn, line 2: `_H` -> `_HPP`

unit_tests/src/catch2_compat_include/catch2/catch_all.hpp
```diff
@@ -0,0 +1,28 @@
...
+        using Impl::MatcherBase;
+    } //namespace Matchers
+
+    namespace clara { }
+    namespace Clara = clara;
+} //namespace Catch
+
+// NOLINTEND(misc-unused-alias-decls)
+
+#endif // IRODS_CATCH2_V2_SHIM_H
```
korydraughn, line 28: `_H` -> `_HPP`

korydraughn (commented):
Looks good. Just have a few questions.

unit_tests/src/catch2_compat_include/catch2/catch_all.hpp
```diff
@@ -0,0 +1,28 @@
...
+        template <typename... Args>
+        constexpr auto ContainsSubstring(Args&&... args) -> decltype(Contains(std::forward<Args>(args)...))
+        {
+            return Contains(std::forward<Args>(args)...);
+        }
+
+        using Impl::MatcherBase;
+    } //namespace Matchers
+
+    namespace clara { }
```
korydraughn, line 22: Could remove the space between the braces and wrap this in `// clang-format off/on`.

Up to you.

korydraughn:
Seems clang-tidy thinks we need to review `enum error_enum`. Please put that part of the report in an issue for the 5.1.0 milestone so we remember to look at it.

.github/workflows/build-irods-debian.yml
```diff
@@ -14,25 +14,47 @@ jobs:
...
+                libxml2-dev \
+                nlohmann-json3-dev \
+                unixodbc-dev \
+                python3-psutil \
+                odbc-postgresql \
+                python3-distro \
+                flex \
+                bison \
+                libsystemd-dev \
+                libarchive-dev
```
trel, line 45: is it interesting/useful/good to alphabetize this list?

plugins/database/src/low_level_odbc.cpp
```diff
@@ -129,8 +129,15 @@ logPsgError( HENV henv, HDBC hdbc, HSTMT hstmt, int dbType ) {
                 errorVal = CATALOG_ALREADY_HAS_ITEM_BY_THAT_NAME;
             }
         }
-
-        log_sql::info("{}: SQLState: {}; SQLCODE: {}; SQL Error message: {}", __func__, sqlstate, sqlcode, psgErrorMsg);
+        // Some versions of fmt don't know how to deal with SQLCHAR[].
```
trel, line 132: can we mark what version of fmt is sufficient... so that once we clear that version everywhere, we can remove this workaround?

unit_tests/CMakeLists.txt
```diff
@@ -8,7 +8,9 @@ endif()
 
 set(IRODS_UNIT_TESTS_BUILD NO CACHE BOOL "${_IRODS_UNIT_TESTS_BUILD_DOCSTR}")
 set(IRODS_UNIT_TESTS_REPORTING_STYLE "junit" CACHE STRING "The style of output used for unit test reporting [console, compact, junit, xml]")
-set(IRODS_UNIT_TESTS_REPORT_FILENAME "report.xml" CACHE STRING "The filename of the unit test report")
+set(IRODS_UNIT_TESTS_REPORT_DIR "${CMAKE_CURRENT_BINARY_DIR}/reports" CACHE STRING "The filename prefix of the unit test report")
```
trel, line 11: ?

```suggestion
set(IRODS_UNIT_TESTS_REPORT_DIR "${CMAKE_CURRENT_BINARY_DIR}/reports" CACHE STRING "The directory of the unit test report")
```

lib/core/src/irods_client_negotiation.cpp
```diff
@@ -58,10 +58,12 @@ namespace irods
...
                 }
 
-                const auto hostname_in_provider_hosts_list = std::any_of(
-                    std::cbegin(*provider_hosts), std::cend(*provider_hosts), [&_host_name](const auto& _host) {
-                        return _host_name == _host;
-                    });
+                const auto hostname_in_provider_hosts_list =
+                    std::any_of(std::cbegin(*provider_hosts),
+                                std::cend(*provider_hosts),
+                                [&_host_name](const nlohmann::json& _host) {
```
trel, line 64: interesting you had to leave `auto`... and this is the only place this was required?   why was it required?

plugins/database/src/low_level_odbc.cpp
SwooshyCueb: It's a problem with newer versions, not older versions.

trel: oh, well that's upside down :)

let's make that clear in the comment - thanks.

.clang-tidy
SwooshyCueb: The content of a macro does not impact the perceived complexity of a function as far as lining should be concerned.

.clang-tidy
SwooshyCueb: Given that the check produces warnings that we often ignore, it's better to have this off by default to avoid filling up the report with information we usually don't care about. It can always be overridden via argument.

lib/core/src/irods_client_negotiation.cpp
SwooshyCueb: I left `auto` because the change wasn't necessary, not because it was required.

unit_tests/src/catch2_compat_include/catch2/catch_all.hpp
SwooshyCueb: I don't want to add toggle comments as I anticipate clang-format eventually growing the capability to tolerate single-line empty namespaces, and I think it's silly that it doesn't already.

.clang-tidy
SwooshyCueb: I suppose it's not, unless the default is different in a different version

.clang-tidy
korydraughn: Fine by me.

unit_tests/src/catch2_compat_include/catch2/catch_all.hpp
korydraughn: That's fair.

Resolving.

lib/core/src/irods_client_negotiation.cpp
trel: okay

.clang-tidy
korydraughn: If we're unsure, please remove it and open an issue to investigate it further.

CMakeLists.txt
```diff
@@ -229,12 +246,21 @@ if (IRODS_BUILD_WITH_WERROR)
...
+if (CMAKE_VERSION VERSION_LESS "3.19.0")
+  find_package(fmt "8.1.1" REQUIRED)
+else()
+  find_package(fmt 8.1.1...<10.0.0 REQUIRED)
+endif()
+find_package(spdlog "1.9.2" REQUIRED)
 find_package(LibArchive "3.3.1" REQUIRED)
 
+if (TARGET fmt::fmt-header-only AND NOT TARGET fmt::fmt)
+  add_library(fmt::fmt ALIAS fmt::fmt-header-only)
```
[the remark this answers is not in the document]
SwooshyCueb: Yes, via [`IRODSTargetsWrapper.cmake`](https://github.com/irods/irods/pull/8366/files#diff-eed9ffa7d5693e560a0ba3fe1ee307997b9f520477c4956ef6fd7ffca6e235a1)

SwooshyCueb:
> Seems clang-tidy thinks we need to review `enum error_enum`. Please put that part of the report in an issue for the 5.1.0 milestone so we remember to look at it.

From what I can tell, this is because clang-tidy is confused about our catch2 v2 compatibility shim. It tries to pull it in instead of catch2 v3, despite the buildsystem telling it not to.

.clang-tidy
SwooshyCueb: Removed. I think I added it because I saw it in the release notes and didn't check what the default was.

CMakeLists.txt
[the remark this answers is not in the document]
korydraughn: Nice.

.clang-tidy
korydraughn, line 1: GitHub is reporting that something has changed in this file since I last reviewed it. What's changed?

unit_tests/CMakeLists.txt
korydraughn, line 1: GitHub is reporting that something has changed in this file since I last reviewed it. What's changed?

.clang-tidy
SwooshyCueb: I removed the `WarnOnIntegerToFloatingPointNarrowingConversion` option as per the review comment.

unit_tests/CMakeLists.txt
SwooshyCueb: The docstring for `IRODS_UNIT_TESTS_REPORT_DIR`

alanking (commented):
Checks out. ~Consider addressing clang-format concerns.~ I believe victory was declared with respect to testing, so this seems close

EDIT: Just saw "I've made all the clang-format-suggested changes I intend to.". Up to you.

alanking (approved):
I think this is good to go... Please allow one more approval before #'ing.

SwooshyCueb:
\#'d

emilazy:
@SwooshyCueb As a distro maintainer, I just wanted to say thank you so, *so* much for all the incredible work you’re doing around packaging and updates here and elsewhere!

You may be interested in taking a look at https://github.com/NixOS/nixpkgs/pull/434776 for the remaining few places our third‐party packaging has to stray off the happy path, but your work in 5.0 has taken it from having a very tangled set of overrides to obsolete versions and tricky dependency injections to being a breeze to package, and I deeply appreciate it :)

SwooshyCueb:
@emilazy You have no idea how happy your comment has made me. I had no idea our software was even being packaged for NixOS!

There's still a long road from here to Normal and Boring (you may be interested in watching [my talk from our last conference](https://www.youtube.com/watch?v=3lypO3zWWpc) for more on that) but I think I'm making good progress! I'll take a look at your PR and see if there's anything that I need to add to our list.

PR #8371, opened 2025-04-10
[7962] Remove tests for imeta interactive mode (main)

FifthPotato:
Core tests pass after these removals. Open to any suggestions.


FifthPotato:
> Please update the OP to reference the icommands PR. Do the same for the icommands PR. Always cross reference PRs which depend on each other.
> 
> Take a look at other PRs for this repo if you'd like to see examples.

My original thought was that this PR is technically independent, in that the test removals could happen without breaking testing but the icommands change depends on this one to not break the tests. Should I always link dependencies both ways just in case?

FifthPotato:
Octothorpe'd

Issue #8016, opened 2024-11-12
C++ client-side auth plugins should directly accept passwords


alanking:
- [x] `a_pw` support in main
- [x] erase `a_pw` after using in `pam_password` in main
- [x] cherry-pick changes to 4-3-stable

---

## Feature

Currently, the C/C++ client-side authentication plugins which are used with the iCommands (i.e. via `clientLogin()`) only derive a user's password in one of 2 ways:
 1. Get the obfuscated password from the `.irodsA` file
 2. Get the password from `stdin`

Here are the places where we determine whether we "need" a password or if we have the password already:

native:
https://github.com/irods/irods/blob/13c08dab315ec05f953f76f3e082d9268a2f7283/plugins/auth/src/native.cpp#L101-L133

pam_password:
https://github.com/irods/irods/blob/13c08dab315ec05f953f76f3e082d9268a2f7283/plugins/auth/src/pam_password.cpp#L150-L171

We should support clients directly passing the password to `clientLogin()` via a keyword in the JSON request. This will not require any changes to `clientLogin()` itself - just the client-side plugins.

This feature would be useful for applications which derive the user's password from some source other than `.irodsA` and `stdin`.

Bonus: Update `clientLoginWithPassword()` to also use the new authentication framework as it continues to use legacy authentication plugins. This would entail backward-compatible changes to `clientLoginWithPassword()` such as were done to `clientLogin()` for 4.3.0: https://github.com/irods/irods/commit/c6a91c38399db5f3fd798f5a7e96d2d29ef9602e

alanking:
Part of https://github.com/irods/irods/issues/8257 will be deprecating `clientLoginWithPassword`, which will make this work required, so, linking that here.

PR #8383, opened 2025-04-15
Deprecate legacy auth plugins, absorb some iinit functionality into native auth plugin (main)

alanking:
Addresses #6150 
Addresses #8016
In service of #8257 

Companion PR: https://github.com/irods/irods_client_icommands/pull/561

I didn't add any new tests for the new options mainly because it's just moving existing functionality around (namely, iinit). I can add some if we want, but I didn't feel it was necessary at this time.


plugins/microservices/src/test_user_administration.cpp
```diff
@@ -32,14 +35,15 @@ namespace
...
+        const adm::user_password_property prop{test_user.name};
         IRODS_MSI_NOTHROW(adm::server::modify_user(*_rei->rsComm, test_user, prop));
 
         // Show we can connect to the server as the test user.
         const auto& env = _rei->rsComm->myEnv;
         irods::experimental::client_connection conn{
             irods::experimental::defer_authentication, env.rodsHost, env.rodsPort, {test_user.name, env.rodsZone}};
-        IRODS_MSI_ASSERT(clientLoginWithPassword(static_cast<RcComm*>(conn), prop.value.data()) == 0);
+        const auto ctx = nlohmann::json{{irods::AUTH_PASSWORD_KEY, prop.value.data()}};
+        IRODS_MSI_ASSERT(clientLogin(static_cast<RcComm*>(conn), ctx.dump().c_str()) == 0);
```
korydraughn, line 46: Good stuff.

plugins/authentication/src/native.cpp
```diff
@@ -291,13 +332,71 @@ namespace irods
...
+            // If TTL was provided, get and record the limited password.
+            if (const auto ttl_iter = _req.find(irods::AUTH_TTL_KEY); ttl_iter != _req.end()) {
+                if (const auto& ttl_str = ttl_iter->get_ref<const std::string&>(); !ttl_str.empty()) {
+                    try {
+                        // Specifically check for TTL of 0 here because 0 means "no TTL" for native authentication and
+                        // is a valid value as far as this authentication plugin is concerned. However, the mechanism
+                        // for getting the limited password considers non-positive values invalid. Therefore, we drop
+                        // through to the "else" case for 0 and pass non-zero values to the limited password machinery
+                        // to allow it to perform the value checking.
+                        if (const auto ttl = boost::lexical_cast<int>(ttl_str); 0 != ttl) {
```
korydraughn, line 372: Why use `boost::lexical_cast` instead of `std::stoi`?

scripts/irods/test/test_auth.py
```diff
@@ -547,7 +547,6 @@ def test_iinit_ssl_failure_with_no_default_answers(self):
         self.assertIn(error_string, stderr)
 
         self.assert_basic_iinit_prompts_are_in_stdout(stdout)
-        self.assertIn('Enter your current iRODS password', stdout)
```
korydraughn, line 550: I feel this should be more obvious to me, but why aren't these needed anymore?

korydraughn (commented):
This is good stuff.

plugins/authentication/src/native.cpp
alanking: Mostly because I copied this from elsewhere and tweaked it. Can switch to `std::stoi`.

scripts/irods/test/test_auth.py
alanking: Ah, this was something I meant to discuss before making the PR but it's pretty minor so we can just do it here.

`iinit` has historically prompted for and saved the user's password via `obfSavePw`: https://github.com/irods/irods_client_icommands/blob/aaddbd28143a69830f9bd39823975a6d32ff46ea/src/iinit.cpp#L389-L402

This has been removed because we moved the recording of the authentication file into the plugin. `iinit` connects to the server before calling `clientLogin`. This test intentionally misconfigures a client environment such that it cannot even connect (it's using TLS when TLS is not enabled in the server). So the prompt which used to appear for the password no longer appears.

The bigger change as a result of this is that the irodsA file is not saved by `iinit` in this case. I think this is okay because I don't see any problem with having the user authenticate again once connection with the server is re-established.

Thoughts?

lib/core/include/irods/rcConnect.h
```diff
@@ -219,20 +219,26 @@ int
...
-// clang-format off
+__attribute__((deprecated("Use clientLogin with pam_password scheme, and AUTH_PASSWORD_KEY and TTL_KEY in context.")))
+int clientLoginPam(rcComm_t* conn, char* password, int ttl);
+
+// This function does not actually authenticate. It requests a limited password for the connected client user and
+// then records it as an obfuscated password in the .irodsA file. This is used by iinit exclusively for native
+// authentication with TTL.
+int clientLoginTTL(rcComm_t* conn, int ttl);
+
+// This function only uses by legacy native authentication.
```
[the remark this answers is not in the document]
alanking: Fixed

plugins/authentication/src/native.cpp
alanking: Switched

alanking:
Found a couple of things and addressed review comments. Gonna run tests again

scripts/irods/test/test_auth.py
korydraughn: That sounds fine to me.

plugins/authentication/src/native.cpp
```diff
@@ -378,8 +377,11 @@ namespace irods
                                 {"scheme", scheme_name}, {irods_auth::next_operation, AUTH_CLIENT_START}};
                         }
                     }
-                    catch (const boost::bad_lexical_cast& e) {
-                        THROW(SYS_INVALID_INPUT_PARAM, fmt::format("invalid TTL [{}]", ttl_str));
+                    catch (const std::invalid_argument& e) {
+                        THROW(SYS_INVALID_INPUT_PARAM, fmt::format("Invalid TTL [{}]: {}", ttl_str, e.what()));
+                    }
+                    catch (const std::out_of_range& e) {
+                        THROW(SYS_INVALID_INPUT_PARAM, fmt::format("Invalid TTL [{}]: {}", ttl_str, e.what()));
```
korydraughn, line 384: I take it the lack of a catch block for `std::exception` is intentional as it is handled by an earlier frame in the callstack?

The use of `THROW` near the end of this function hints at that.

alanking: The `std::exception` would eventually be caught by `authenticate_client`, yes. We can handle it here, too, if that's preferred, though.

alanking:
Unit and core tests passed. I'm trying to address some of the clang-tidy issues now

plugins/authentication/src/native.cpp
korydraughn: No, I was just wondering. It's fine as is.

alanking:
I addressed most of the clang-tidy issues. Unit tests passed, running core tests again to make sure

alanking:
Sorry, one more push coming... Forgot to run clang-format one last time

plugins/authentication/src/native.cpp
```diff
@@ -33,30 +35,112 @@
...
+
+        if (const auto err = obfSavePw(0, 0, 0, limited_password.data()); err < 0) {
+            THROW(err, fmt::format("{}: Failed to save limited password to irodsA file. error: [{}]", __func__, err));
+        }
+    } // record_limited_password
+
+    auto get_password_from_client_stdin() -> std::string
+    {
+        termios tty{};
+        std::memset(&tty, 0, sizeof(tty));
```
alanking, line 92: Can remove this now, I think...

korydraughn: If it's POD, then yeah.

alanking: Yep, this is a C struct: https://linux.die.net/man/3/termios

Done

plugins/authentication/src/native.cpp
```diff
@@ -357,7 +362,7 @@ namespace irods
             return resp;
         } // native_auth_client_response
 
-        auto record_authentication_file_op(RcComm& _comm, const nlohmann::json& _req) -> nlohmann::json
+        static auto record_authentication_file_op(RcComm& _comm, const nlohmann::json& _req) -> nlohmann::json
```
korydraughn, line 368: These are static because they don't rely on state associated with a particular instance?

alanking: Yep, apparently. I think that all of these could actually be static, really, but I was just responding to the clang-tidy report.

alanking:
Last clang-tidy-related push, promise :) Unit tests passed, core tests running

alanking:
Core tests passed

alanking:
Squashed

korydraughn (approved):
Pound it.

alanking:
#'d, mergin

PR #8393, opened 2025-04-17
[#8016] Erase AUTH_PASSWORD_KEY from response in pam_password (main)

alanking:
In service of #8016

Only ran `test_pam_password_authentication` and it passes. Can run the whole test suite if desired.

### DIFF
--- a/plugins/authentication/src/pam_password.cpp
+++ b/plugins/authentication/src/pam_password.cpp
@@ -287,8 +287,18 @@ namespace irods
 
             json resp{req};
 
+            // This operation asserts that the key exists in the request above, so we know that it is there.
+            const auto password_iter = resp.find(irods::AUTH_PASSWORD_KEY);
+            const auto password = password_iter->get<std::string>();
+
+            // Erase the password key as soon as it is no longer needed to avoid sending it back in the response.
+            resp.erase(password_iter);
+
             log_auth::trace("getting TTL param");
 
+            // Check TTL parameters before checking the password so that we avoid unnecessary communication with the PAM
+            // service. If we cannot authenticate with the iRODS server because of a bad TTL input, there is no reason
+            // to check the PAM credentials.
             int ttl = 0;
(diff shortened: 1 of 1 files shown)


plugins/authentication/src/pam_password.cpp
```diff
@@ -287,8 +287,18 @@ namespace irods
 
             json resp{req};
 
+            // This operation asserts that the key exists in the request above, so we know that it is there.
+            const auto password_iter = resp.find(irods::AUTH_PASSWORD_KEY);
+            const auto password = password_iter->get<std::string>();
+
+            // Erase the password key as soon as is is no longer needed to avoid sending it back in the response.
```
trel, line 294:
```suggestion
            // Erase the password key as soon as it is no longer needed to avoid sending it back in the response.
```

plugins/authentication/src/pam_password.cpp
```diff
@@ -287,8 +287,18 @@ namespace irods
...
+            resp.erase(password_iter);
+
             log_auth::trace("getting TTL param");
 
+            // Check TTL parameters before checking the password so that we avoid unnecessary communication with the PAM
+            // service. If we cannot authenticate with the iRODS server because of a bad TTL input, there was no reason
```
trel, line 300: not sure... the tense is correct?

```suggestion
            // service. If we cannot authenticate with the iRODS server because of a bad TTL input, there is no reason
```

plugins/authentication/src/pam_password.cpp
alanking: Doe

plugins/authentication/src/pam_password.cpp
alanking: Done

korydraughn (commented):
Squash it.

alanking:
Squashed

korydraughn (approved):
Pound it.

alanking:
#'d, mergin

Issue #8391, opened 2025-04-16
Add `DEPRECATED_AUTHENTICATION_PLUGIN` error code to 4.3


alanking:
- [x] 4-3-stable only

---

### New Feature / Functionality

In the unlikely event that a 4.3 client tries to connect to a 5.0 server using legacy auth plugins, it should probably have a name for the error code that comes back in the rError message.

Error code added here: https://github.com/irods/irods/pull/8383

Probably will be in a PR along with the cherry-pick for https://github.com/irods/irods/issues/8016

Issue #7948, opened 2024-09-04
`iinit` shows "iRODS password" prompt for non-native auth schemes


alanking:
- [x] main (fixed in https://github.com/irods/irods_client_icommands/pull/554)
- [x] 4-3-stable

---

### iRODS Version, OS and Version
4.3.1+

### What did you try to do?
I created an auth plugin which is supposed to function like `pam_permit` (for an experiment). The auth scheme name is new and does not share any officially-supported auth scheme names.

### Expected behavior
I expected `iinit` to not prompt for a password at all because my plugin does not ask for a password and it does not record anything in the `.irodsA` file.

### Observed behavior (including steps to reproduce, if applicable)
```bash
$ iinit
Enter your current iRODS password:
```

This is because `iinit` erroneously returns true in `scheme_uses_iinit_password_prompt` for any unlisted auth schemes:
https://github.com/irods/irods_client_icommands/blob/49562235b1457e1f1db966f02278c59a35d9ff48/src/iinit.cpp#L91-L104

This in turn causes `obfSavePw` to be called...
https://github.com/irods/irods_client_icommands/blob/49562235b1457e1f1db966f02278c59a35d9ff48/src/iinit.cpp#L409-L416

And `obfSavePw` displays this prompt:
https://github.com/irods/irods/blob/bc2a9bf0d8d3587d1aad8bf6a9b3db650e0d6944/lib/core/src/obf.cpp#L305

We should consider requiring auth plugins to do this work (rather than relying on `iinit`) or we should only show the prompt for `native` authentication. I'm not sure why we implemented it this way, but the Kerberos scheme is notably missing here which makes me wonder whether Kerberos used to display this prompt...

alanking:
I believe this issue is resolved in main/pre-5.0 as of this commit: https://github.com/irods/irods_client_icommands/commit/7eeacaf7e588219ba34568a6f035ec65902018a9

PR #8394, opened 2025-04-17
Enable auth plugins to accept passwords in requests, add 5.0 error code (4-3-stable)

alanking:
In service of #7948 
Addresses #8016
Addresses #8391 

Cherry-picked from: #8383
Cherry-picked from: #8393

Companion PR: https://github.com/irods/irods_client_icommands/pull/562

Unit tests passed, auth-based tests are passing. Running the full core test suite now.

### DIFF
--- a/plugins/auth/src/pam_password.cpp
+++ b/plugins/auth/src/pam_password.cpp
@@ -297,8 +297,18 @@ namespace irods
 
             json resp{req};
 
+            // This operation asserts that the key exists in the request above, so we know that it is there.
+            const auto password_iter = resp.find(irods::AUTH_PASSWORD_KEY);
+            const auto password = password_iter->get<std::string>();
+
+            // Erase the password key as soon as it is no longer needed to avoid sending it back in the response.
+            resp.erase(password_iter);
+
             log_auth::trace("getting TTL param");
 
+            // Check TTL parameters before checking the password so that we avoid unnecessary communication with the PAM
+            // service. If we cannot authenticate with the iRODS server because of a bad TTL input, there is no reason
+            // to check the PAM credentials.
             int ttl = 0;
             if (req.contains(irods::AUTH_TTL_KEY)) {
                 if (const auto& ttl_str = req.at(irods::AUTH_TTL_KEY).get_ref<const std::string&>(); !ttl_str.empty()) {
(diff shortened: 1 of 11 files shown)


plugins/auth/src/pam_password.cpp
```diff
@@ -296,8 +297,18 @@ namespace irods
 
             json resp{req};
 
+            // This operation asserts that the key exists in the request above, so we know that it is there.
+            const auto password_iter = resp.find(irods::AUTH_PASSWORD_KEY);
+            const auto password = password_iter->get<std::string>();
+
+            // Erase the password key as soon as is is no longer needed to avoid sending it back in the response.
```
trel, line 304:
```suggestion
            // Erase the password key as soon as it is no longer needed to avoid sending it back in the response.
```

plugins/auth/src/pam_password.cpp
```diff
@@ -296,8 +297,18 @@ namespace irods
...
+            resp.erase(password_iter);
+
             log_auth::trace("getting TTL param");
 
+            // Check TTL parameters before checking the password so that we avoid unnecessary communication with the PAM
+            // service. If we cannot authenticate with the iRODS server because of a bad TTL input, there was no reason
```
trel, line 310:
```suggestion
            // service. If we cannot authenticate with the iRODS server because of a bad TTL input, there is no reason
```

korydraughn (commented):
Awaiting test results.

alanking:
So... `test_native_authentication.test_configurations.test_password_max_time_less_than_password_min_time_makes_ttl_constraints_unsatisfiable` failed when run in the full test suite, but it passed when run by itself and when `test_native_authentication.test_configurations` is run by itself. Gonna need to investigate this.

alanking:
I tried for a very long time to reproduce this issue and was not able to. The only thing in the server log around the failed test was `rcAuthCheck failed` (from the native auth plugin) and the error returned was `CAT_INVALID_AUTHENTICATION`.

The tests have been passing consistently ever since, so I think we can revisit this if it comes up again.

alanking:
Squashed

korydraughn (approved):
Pound it.

alanking:
#'d, mergin

PR #8398, opened 2025-04-21
iRODS 5: Restore eviction logic for expired dns/hostname caches, etc. (main)

korydraughn:
The important changes of this PR are:
- DNS/Hostname caches are now being cleared
  - This step fell through the cracks during the initial redesign of the startup/shutdown process
- Optional config properties in iRODS 4 are now required
- Environment variables defined in server_config.json are now being set
  - Child processes inherit the values

Running tests. Will open issues once the tests are passing.


server/main_server/src/main.cpp
```diff
@@ -280,11 +274,17 @@ auto main(int _argc, char* _argv[]) -> int
         log_server::info("{}: Initializing shared memory for main server process.", __func__);
 
         namespace hnc = irods::experimental::net::hostname_cache;
```
trel, line 276: interested in pulling these out of the experimental namespace?

also still noticed 'log' is in experimental too...

korydraughn: Yes, but I'm afraid we may run out of time to complete that effort since the blast radius is wide. I'm mostly speaking about the logger. The hostname and dns cache are pretty simple to handle.

I'll open issues for that effort so we can circle back to them after everything for iRODS 5 is confirmed to be working.

korydraughn: Created #8399, #8400, #8403, and a few others for various libs.

lib/core/src/rcMisc.cpp
```diff
@@ -4684,7 +4684,8 @@ int get_canonical_name(const char *_hostname, char* _buf, size_t _len)
     snprintf(_buf, _len, "%s", p_addrinfo->ai_canonname);
 
     if (CLIENT_PT != ::ProcessType) {
-        const std::chrono::seconds age{irods::get_dns_cache_eviction_age()};
+        const auto cache_config = irods::get_advanced_setting<nlohmann::json>(irods::KW_CFG_DNS_CACHE);
+        const std::chrono::seconds age{cache_config.at(irods::KW_CFG_EVICTION_AGE_IN_SECONDS).get<int>()};
```
alanking, line 4688: I realize string allocations can throw and we are leaning on the schema validation, but I think we should handle exceptions around this.

lib/core/src/rcMisc.cpp
```diff
@@ -4725,7 +4726,8 @@ int load_in_addr_from_hostname(const char* _hostname, struct in_addr* _out)
     *_out = reinterpret_cast<struct sockaddr_in*>(p_addrinfo->ai_addr)->sin_addr;
 
     if (CLIENT_PT != ::ProcessType) {
-        const std::chrono::seconds age{irods::get_dns_cache_eviction_age()};
+        const auto cache_config = irods::get_advanced_setting<nlohmann::json>(irods::KW_CFG_DNS_CACHE);
+        const std::chrono::seconds age{cache_config.at(irods::KW_CFG_EVICTION_AGE_IN_SECONDS).get<int>()};
```
alanking, line 4730: I realize string allocations can throw and we are leaning on the schema validation, but I think we should handle exceptions around this.

lib/core/src/sockComm.cpp
```diff
@@ -489,7 +489,8 @@ sockOpenForInConn( rsComm_t *rsComm, int *portNum, char **addr, int proto ) {
                     std::strncpy(*addr, alias->data(), alias->size());
                 }
 
-                const std::chrono::seconds age{irods::get_hostname_cache_eviction_age()};
+                const auto cache_config = irods::get_advanced_setting<nlohmann::json>(irods::KW_CFG_HOSTNAME_CACHE);
+                const std::chrono::seconds age{cache_config.at(irods::KW_CFG_EVICTION_AGE_IN_SECONDS).get<int>()};
```
alanking, line 493: I realize string allocations can throw and we are leaning on the schema validation, but I think we should handle `nlohmann::json` exceptions and/or `std::exception` around this.

lib/core/include/irods/irods_server_properties.hpp
```diff
@@ -282,43 +282,6 @@ namespace irods
...
     } // get_advanced_setting
-
-    /// Returns the amount of shared memory that should be allocated for the DNS cache.
-    ///
-    /// \return An integer representing the size in bytes.
-    /// \retval 5000000          If an error occurred or the size was less than or equal to zero.
-    /// \retval Configured-Value Otherwise.
-    ///
-    /// \since 4.2.9
-    auto get_dns_cache_shared_memory_size() noexcept -> int;
```
alanking, line 293: I'm guessing we don't need to deprecate these because these are pretty deep in the server despite being in `lib`?

alanking (commented):
Makes sense overall.

lib/core/src/rcMisc.cpp
korydraughn: Agreed. The original code handled exceptions and this should do the same.

server/main_server/src/main.cpp
```diff
@@ -1461,6 +1464,38 @@ Environment Variables:
...
+
+        static auto start_time = std::chrono::steady_clock::now(); 
+        const auto now = std::chrono::steady_clock::now();
+
+        if (now - start_time < std::chrono::seconds{sleep_time}) {
+            return;
+        }
+
+        start_time = now;
+        irods::experimental::net::dns_cache::erase_expired_entries();
```
korydraughn, line 1480: Consider exceptions.

server/main_server/src/main.cpp
```diff
@@ -1461,6 +1464,38 @@ Environment Variables:
...
+
+        static auto start_time = std::chrono::steady_clock::now(); 
+        const auto now = std::chrono::steady_clock::now();
+
+        if (now - start_time < std::chrono::seconds{sleep_time}) {
+            return;
+        }
+
+        start_time = now;
+        irods::experimental::net::hostname_cache::erase_expired_entries();
```
korydraughn, line 1496: Consider exceptions.

lib/core/src/sockComm.cpp
korydraughn: Agreed. The original code handled exceptions and this should do the same.

lib/core/include/irods/irods_server_properties.hpp
korydraughn: The plan is to mark them deprecated in 4.3.5. An issue for removing these in 5.0.0 would appear as well. I delayed creating the issues to make sure the work could be completed on time.

korydraughn:
All unit tests passed.

All core tests passed except the following:
- `test_delay_queue.Test_Delay_Queue`
- `test_rule_engine_plugin_framework.Test_Plugin_Instance_CppDefault`

Investigating those failures now.

korydraughn:
The test failures appear to be timing related. I've seen the tests pass multiple times now.

I did discover that a cpp-default REP test was leaving an intermediate replica in storage, so I fixed that. I also discovered that many of the delay queue tests do not target a specific REP. I'll open an issue to look into that outside of this PR.

Going to address review comments now.

alanking:
> I did discover that a cpp-default REP test was leaving an intermediate replica in storage, so I fixed that.

We should probably create an issue to address that. It means the L1 descriptor is opened but the servicing agent never closes it. I see that `msiDataObjCreate` is calling `rsDataObjCreate` (which is just a wrapper around `rsDataObjOpen`), so it somehow is getting around the safeguard of the agent handling the close.

korydraughn:
That's true. Will see if I can reproduce that intermediate state.

korydraughn:
After some investigation, I'm certain the failures are due to timing.

The cpp-default REP test uses a `delay_assert` function (similar to `lib.delayAssert`) which only gives the passed callable 5 chances to succeed. The test passes when `ils` lists a replica in the intermediate state or stale state. The test fails whenever the replica isn't shown. With that said, I'm sure things are working (i.e. agents are cleaning up correctly) and the test just needs to be tighten up. This is further supported by the fact that the test session collections are removed without error. If an intermediate replica was present, we'd see error messages during test teardown.

I have not been able to reproduce the test failure I saw with the delay queue tests, but looking at the source for that test suite, it's easy to see that it can be tightened up as well. There are many cases where `irule` is invoked without a specific REP.

The test failures are not caused by the changes in this PR. I'm going to open issues for those test suites and focus on addressing the remaining review comments.

korydraughn:
Created #8414 and #8415 for the test failures.

lib/core/src/rcMisc.cpp
korydraughn: Done.

lib/core/src/rcMisc.cpp
korydraughn: Done.

lib/core/src/sockComm.cpp
korydraughn: Done.

server/main_server/src/main.cpp

server/main_server/src/main.cpp

korydraughn:
Unit tests still pass. No need to rerun the core test suite since the latest changes just add try-catch blocks and tweak one python test.

lib/core/include/irods/irods_server_properties.hpp
korydraughn: Opened #8417 and #8418.

korydraughn:
This is ready for final review.

alanking (commented):
Makes sense to me... Awaiting test results.

I say go ahead and squash to your liking.

korydraughn:
> Makes sense to me... Awaiting test results.

See https://github.com/irods/irods/pull/8398#issuecomment-2822339268.

alanking:
Oh, I misread that. Yeah, go ahead and squash

korydraughn:
Squashed.

alanking (approved):
\# it

korydraughn:
Applied clang-format.

Squashed and pounded.

korydraughn:
Post merge update: Decided to rerun the core tests. All core tests passed.

Issue #8407, opened 2025-04-21
Remove `rcAuthenticate` and `rsAuthenticate`


alanking:
- [x] main

---

`rcAuthenticate` and `rsAuthenticate` are declared but not defined. The definitions for these functions are missing in irods-legacy, so they have been undefined and unused for a long time. Because it is not possible to use this without providing one's own implementation, I don't think there's a need to deprecate it. Let's just remove the API endpoint.

Discovered while working on https://github.com/irods/irods/issues/8257

alanking:
I have also discovered that there is no entry for this API in the table, nor does it have an API number. Let's just get rid of all this.

alanking:
We have decided that this change should only go into 5.0.0 to avoid making client code adjust to the changes. This is now complete.

PR #8410, opened 2025-04-21
[#8257,#8407] Deprecate and replace clientLogin (main)

alanking:
In service of #8257 
Addresses #8407 

Companion PR: https://github.com/irods/irods_client_icommands/pull/564

Unit tests pass. Core tests running now.


lib/core/include/irods/authenticate.h
```diff
@@ -0,0 +1,70 @@
...
+extern "C" {
+#endif
+
+struct RcComm;
+
+// These macros are used extensively throughout the authentication code. Over time, we should move away from using
+// these, but we cannot escape from it for now. Just know that these values are currently very important.
+#define MAX_PASSWORD_LEN 50 // NOLINT(cppcoreguidelines-macro-usage)
+#define CHALLENGE_LEN    64 // NOLINT(cppcoreguidelines-macro-usage)
+#define RESPONSE_LEN     16 // NOLINT(cppcoreguidelines-macro-usage)
```
alanking, line 16: Hm... I should probably make these `NOLINTBEGIN`/`NOLINTEND`. Looks weird next to a macro.

alanking: Done

korydraughn (commented):
Looks good overall.

lib/core/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ set(
   IRODS_LIB_CORE_HEADERS_PUBLIC
   "${CMAKE_CURRENT_SOURCE_DIR}/include/irods/alignPointer.hpp"
   "${CMAKE_CURRENT_SOURCE_DIR}/include/irods/apiHandler.hpp"
+  "${CMAKE_CURRENT_SOURCE_DIR}/include/irods/authenticate.h"
```
korydraughn, line 5: This is in lib/core because `rc_authenticate_client` isn't a real API, correct? It wraps a real API call.

lib/core/src/authenticate.cpp
```diff
@@ -0,0 +1,27 @@
...
+        return 0;
+    }
+
+    RodsEnvironment env{};
+    if (const auto err = getRodsEnv(&env); 0 != err) {
+        return err;
+    }
+
+    const auto ctx = (nullptr != _context) ? nlohmann::json::parse(_context) : nlohmann::json{};
+    return irods::authentication::authenticate_client(*_comm, env, ctx);
```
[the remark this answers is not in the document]
alanking: Yep, this is just how it has always been. I'm willing to add another parameter to optionally pass in a `RodsEnvironment` should the caller already have one at hand. We call `getRodsEnv` inside the auth plugins as well, though, so, it might not buy us much.

`irods::authenticate::authenticate_client` uses the `RodsEnvironment` to derive the authentication scheme:
https://github.com/irods/irods/blob/8665d1a4ae05a25292635d1cb646f9f6088b1c6f/lib/core/include/irods/authentication_plugin_framework.hpp#L153

This could arguably be part of the JSON parameter instead. Just a matter of deciding on an interface.

lib/core/CMakeLists.txt
alanking: `rcAuthenticate` / `rsAuthenticate` apparently was an API at some point, but its definition has been lost to time (since before iRODS 3.3). Because we have removed the function declarations and the include path does not change (it's `irods/authenticate.h` either way), I thought it best to move it into `lib/core`.

lib/core/include/irods/authenticate.h
```diff
@@ -0,0 +1,72 @@
...
+
+struct RcComm;
+
+// These macros are used extensively throughout the authentication code. Over time, we should move away from using
+// these, but we cannot escape from it for now. Just know that these values are currently very important.
+// NOLINTBEGIN(cppcoreguidelines-macro-usage)
+#define MAX_PASSWORD_LEN 50 
+#define CHALLENGE_LEN    64
+#define RESPONSE_LEN     16
+// NOLINTEND(cppcoreguidelines-macro-usage)
```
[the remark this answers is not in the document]
alanking: Done

lib/core/include/irods/authenticate.h
```diff
@@ -0,0 +1,72 @@
...
+/// {
+///     // A password to use for authentication.
+///     "a_pw": <string>,
+///     // Time-To-Live (TTL) value. The server interprets this value as hours.
+///     "a_ttl": <integer>,
+///     // Forcibly show a password prompt even if the password can be derived by other means.
+///     "force_password_prompt": <bool>,
+///     // Record auth file (.irodsA) so the user can be authenticated through that in the future.
+///     "record_auth_file": <bool>
+/// }
```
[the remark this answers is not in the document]
alanking: Created https://github.com/irods/irods/issues/8411 for this and removed the options from here.

lib/core/include/irods/authenticate.h
```diff
@@ -0,0 +1,72 @@
...
+///
+/// \since 5.0.0
+#ifdef __cplusplus
+int rc_authenticate_client(struct RcComm* _comm, const char* _context = nullptr);
+#else
+int rc_authenticate_client(struct RcComm* _comm, const char* _context);
+#endif
+
+#ifdef __cplusplus
+}
```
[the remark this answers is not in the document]
alanking: Done

lib/core/src/authenticate.cpp
```diff
@@ -0,0 +1,27 @@
...
+#include "irods/authentication_plugin_framework.hpp"
+#include "irods/getRodsEnv.h"
+#include "irods/rcConnect.h"
+
+#include <nlohmann/json.hpp>
+
+auto rc_authenticate_client(RcComm* _comm, const char* _context) -> int
+{
+    if (!_comm) {
+        return NULL_VALUE_ERR;
```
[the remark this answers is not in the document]
alanking: Done

lib/core/include/irods/authenticate.h
```diff
@@ -0,0 +1,72 @@
...
+/// /* After handling errors, make API calls as usual. */
+/// \endcode
+/// \endparblock
+///
+/// \since 5.0.0
+#ifdef __cplusplus
+int rc_authenticate_client(struct RcComm* _comm, const char* _context = nullptr);
+#else
+int rc_authenticate_client(struct RcComm* _comm, const char* _context);
+#endif
```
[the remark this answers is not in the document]
korydraughn: Okay. Let's go with the simple solution and force users to pass nullptr to the function.

lib/core/src/authenticate.cpp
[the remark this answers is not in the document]
korydraughn: Passing the auth scheme via the JSON payload or a third parameter makes the most sense to me. Removing the need for a rodsEnv seems like the right approach.

And surprisingly, irods4j does just that. See https://github.com/irods/irods4j/blob/4f5bb6a279449765532f405b8c786aaf4f3cc8d7/src/main/java/org/irods/irods4j/low_level/api/IRODSApi.java#L615-L620.

lib/core/CMakeLists.txt
korydraughn: Makes sense.

I'm starting to think the .h/.cpp file should live under lib/api just because it aligns with the idea of an API/network call. It's also where the other client-side APIs live.

Do you feel it would be surprising to a new reader of the code to find this in lib/core instead of lib/api?

lib/core/src/authenticate.cpp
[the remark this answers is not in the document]
korydraughn: > If no "scheme" is provided, we can derive it from the environment as we were doing before (just passes the buck down one layer).

Will need to hear more about deriving the scheme from the environment. Will add that as a discush item for tomorrow.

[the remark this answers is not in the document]
korydraughn: Will leave this here so I don't forgot my current thinking regarding a derived auth scheme.
- Require the caller to pass a valid auth scheme - i.e. the function does never attempts to derive anything, forcing the caller to be explicit
- Provide an option which the caller can use to instruct the function to invoke `getRodsEnv`

lib/core/CMakeLists.txt
alanking: It would think that would be surprising because many if not most of the headers in `lib/api` declare functions and types which map directly to API calls. `rc_authenticate_client` does not map to an API call, it is more like the interface to the auth plugin framework, not the authentication API (which was built for the authentication framework, which exists in `lib/core`).

However, I see your point that to the average user (or new reader) there is no difference. It's just a library call to them and they probably don't care that the authentication plugin framework is doing most of its work on the client side.

korydraughn: I'm not proposing we change where things live, yet. I'm mostly trying to determine whether the location of the files is important given that a consumer of the library won't necessarily care. All of the headers will live in /usr/include/irods. As for a person working on the server code, it is simply a matter of getting used to the codebase.

lib/core/src/authenticate.cpp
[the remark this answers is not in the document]
korydraughn: After some discussion, we've decided to require that users pass an auth scheme to the plugin. Not providing an auth scheme will result in an error.

lib/core/CMakeLists.txt
alanking: After discussion, we decided to leave this in lib/core. Resolving.

lib/core/src/authenticate.cpp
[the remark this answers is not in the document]
alanking: After discussion, we decided that the caller must provide an auth scheme as part of the JSON input. The auth scheme will not be derived automatically by either `irods::authentication::authenticate_client` nor `rc_authenticate_client`.

alanking:
Given the size of the diff after implementing the proposed changes, I think I'm going to split this into multiple commits, which will require a force-push. I've just pushed what I have now, but plan to re-swizzle commits later. If it helps with reviewing, I will go ahead and do it.

Unit tests passed with latest changes. Running core tests now.

alanking:
I also added some unit tests for the `rc_authenticate_client` function.

korydraughn:
Splitting it up into separate commits sounds helpful. Will wait for your signal that it's ready for review.

alanking:
I think the commits are in a good spot. Need to run tests once more.

korydraughn:
Okay. Will start the review.

lib/core/src/connection_pool.cpp
```diff
@@ -277,7 +280,12 @@ namespace irods
             return;
         }
 
-        if (clientLogin(ctx.conn.get()) != 0) {
+        RodsEnvironment env{};
+        if (const auto err = getRodsEnv(&env); err < 0) {
+            THROW(err, "Failed to get client environment.");
+        }
```
korydraughn, line 286: You can actually replace these lines with `_getRodsEnv` since `getRodsEnv` never fails unless passed a nullptr.

lib/core/src/authenticate.cpp
```diff
@@ -0,0 +1,34 @@
...
+        return irods::authentication::authenticate_client(*_comm, nlohmann::json::parse(_context));
+    }
+    catch (const nlohmann::json::exception& e) {
+        constexpr auto err = JSON_VALIDATION_ERROR;
+        // Do not include the JSON in the error message because it might contain sensitive information.
+        const std::string msg = fmt::format("{}: Exception thrown while parsing JSON input: [{}]", __func__, e.what());
+        allocate_if_necessary_and_add_rError_msg(&_comm->rError, err, msg.c_str());
+        return err;
+    }
+} // rc_authenticate_client
```
korydraughn, line 34: Can L25 throw any other types of exceptions?
Is the catch block for `std::exception` intentionally left out?

L30 can fail to allocate memory, resulting in an exception. If an exception leaks in C world, the application is likely going to crash.

lib/core/include/irods/authenticate.h
```diff
@@ -0,0 +1,61 @@
...
+///
+/// After returning successfully, the client will be authenticated with the connected iRODS server for API requests.
+///
+/// \param[in] _comm The iRODS communication object.
+/// \param[in] _context A string representing a JSON object. \parblock
+///
+/// The JSON object is used in the authentication plugin framework. The inputs will differ based on the authentication
+/// scheme, and not all options are recognized by all authentication schemes. The authentication scheme can be provided
+/// with the "scheme" key and a value of a string containing the name of the scheme to use (e.g. "native"). If no such
+/// key is provided, the authentication scheme will be derived from the local client environment.
```
korydraughn, line 30: > If no such key is provided, the authentication scheme will be derived from the local client environment.

This is referring to the historical behavior, correct?
When you say `key`, you're referring to the `scheme` property, correct?

lib/core/include/irods/authenticate.h
```diff
@@ -0,0 +1,61 @@
...
+///
+/// The JSON object must specify the authentication scheme to use. Minimally, it should look like this:
```
korydraughn, line 32: Seems like this line is in conflict with L29-L30?

Can you explain them?

lib/core/include/irods/authenticate.h
```diff
@@ -0,0 +1,61 @@
...
+/// \code{.js}
+/// {"scheme": "<authentication scheme name>"}
+/// \endcode
+///
+/// If the authentication scheme is not provided in \p _context, an error will be returned.
```
korydraughn, line 37: This seems to be in conflict with L29-L30 too?

unit_tests/src/test_authenticate_client.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+    SECTION("JSON array is not valid")
+    {
+        const auto ctx = nlohmann::json{};
+        CHECK(USER_AUTH_SCHEME_ERR == rc_authenticate_client(&comm, ctx.dump().c_str()));
+    }
+
+    SECTION("JSON array with 'scheme' in it is not valid")
+    {
+        static const auto ctx = nlohmann::json{ia::scheme_name};
```
korydraughn, line 47: L47 results in a JSON array?

I'm assuming `ia::scheme_name` is a string.

unit_tests/src/test_authenticate_client.cpp
```diff
@@ -0,0 +1,95 @@
...
+    // Connect to the server.
+    rErrMsg_t error{};
+    RcComm* comm = rcConnect(env.rodsHost, env.rodsPort, env.rodsUserName, env.rodsZone, NO_RECONN, &error);
+    REQUIRE(comm);
+    const auto disconnect = irods::at_scope_exit{[&comm] { rcDisconnect(comm); }};
+
+    // Show that user cannot perform a query on a protected collection before authenticating.
+    const auto query_str =
+        fmt::format("select COLL_ID where COLL_NAME = '/{}/home/{}'", env.rodsZone, env.rodsUserName);
+    CHECK_THROWS(irods::query{comm, query_str});
```
korydraughn, line 87: May want to double check the macro works as you expect. Catch2 sometimes requires the input to be prepared a certain way. I can never remember the rules for when that's necessary.

See https://github.com/catchorg/Catch2/blob/v2.13.6/docs/assertions.md#exceptions.

The nice thing about L87 is that it's checking that something IS thrown, which means it's much more likely that the test will fail if the assertion is incorrectly specified.

lib/core/src/client_connection.cpp
```diff
@@ -220,4 +213,25 @@ namespace irods::experimental
...
+
+        if (!conn_) {
+            // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+            THROW(SYS_LIBRARY_ERROR, "Invalid connection object");
+        }
+
+        RodsEnvironment env{};
+        if (const auto err = getRodsEnv(&env); err < 0) {
+            THROW(err, "Failed to get client environment");
+        }
```
korydraughn, line 229: This can be rewritten as:
```cpp
RodsEnvironment env;
_getRodsEnv(env);
```

korydraughn (commented):
Looking really good.

lib/core/src/connection_pool.cpp
alanking: True. I'll replace the new uses with this. I was a little hesitant to use it given the leading underscore, but it really isn't much different from `getRodsEnv`.

lib/core/src/authenticate.cpp
alanking: True. I'll remove the rError stuff and just return an error code. Will catch other types of exceptions as well, just in case.

lib/core/include/irods/authenticate.h
alanking: Whoops, I updated `irods::authentication::authenticate_client`, but not this one. Will update.

unit_tests/src/test_authenticate_client.cpp
alanking: I believe so. I can make it more explicit if that seems better.

But, for reference, this uses constructor (5) (https://json.nlohmann.me/api/basic_json/basic_json):
> Creates a JSON value of type array or object from the passed initializer list init. In case type_deduction is true (default), the type of the JSON value to be created is deducted from the initializer list init according to the following rules:
>
>   a. If the list is empty, an empty JSON object value {} is created.
>    b. If the list consists of pairs whose first element is a string, a JSON object value is created where the first elements of the pairs are treated as keys and the second elements are as values.
>    c. In all other cases, an array is created.

The list is not empty and does not consist of pairs, so we fall into case (c).

unit_tests/src/test_authenticate_client.cpp
alanking: Is there something in particular you have in mind here? I just want to ensure that the `irods::query` experiences an error (which, for that library, means that it throws).

unit_tests/src/test_authenticate_client.cpp
alanking: That being said, that means the previous test is not an array but an empty JSON object.

server/core/src/miscServerFunct.cpp
```diff
@@ -128,11 +129,17 @@ svrToSvrConnect( rsComm_t *rsComm, rodsServerHost_t *rodsServerHost ) {
         return status;
     }
 
-    status = clientLogin( rodsServerHost->conn );
+    RodsEnvironment env{};
+    if (const auto err = getRodsEnv(&env); err < 0) {
```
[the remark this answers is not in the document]
alanking: Done

lib/core/src/connection_pool.cpp
alanking: Done

lib/core/src/authenticate.cpp
alanking: Done

lib/core/include/irods/authenticate.h
alanking: Done

lib/core/include/irods/authenticate.h
alanking: Fixed L29-L30. That was left over.

unit_tests/src/test_authenticate_client.cpp
alanking: Now using `nlohmann::json::array` explicitly.

lib/core/src/client_connection.cpp
alanking: Done

alanking:

unit_tests/src/test_authenticate_client.cpp
korydraughn: I'm asking if you've confirmed L87 results in an exception being thrown and caught. Running the test with `-s` will cause it to print the assertions to the terminal so you can see what happened.

This is motivated by experiences with the `*_THROW` assertions not doing what I expected (e.g. not executing). One solution to that problem required wrapping the expression in an immediately invoked lambda.

alanking: Ah, I see. Here's what I'm seeing...
```
-------------------------------------------------------------------------------
rc_authenticate_client: minimum viable inputs
-------------------------------------------------------------------------------
/irods_source/unit_tests/src/test_authenticate_client.cpp:70
...............................................................................

/irods_source/unit_tests/src/test_authenticate_client.cpp:81: PASSED:
  REQUIRE( comm )
with expansion:
  0x0000555fc45dfea0

/irods_source/unit_tests/src/test_authenticate_client.cpp:87: PASSED:
  CHECK_THROWS( irods::query{comm, query_str} )

/irods_source/unit_tests/src/test_authenticate_client.cpp:91: PASSED:
  REQUIRE( 0 == rc_authenticate_client(comm, ctx.dump().c_str()) )
with expansion:
  0 == 0

/irods_source/unit_tests/src/test_authenticate_client.cpp:94: PASSED:
  CHECK( !irods::query{comm, query_str}.empty() )
with expansion:
  true
```

It doesn't have "with expansion" but it appears to have executed and "PASSED"?

korydraughn: Hmm, the lack of an expansion makes me think it didn't actually do what you intended. There are times when the output shows the details of the exception, but I can never remember when or what causes that.

Try the following and see if the output changes.
```cpp
CHECK_THROWS([&comm, &query_str] { irods::query{comm, query_str}; }());
```

alanking: Seems to be the same:
```
-------------------------------------------------------------------------------
rc_authenticate_client: minimum viable inputs
-------------------------------------------------------------------------------
/irods_source/unit_tests/src/test_authenticate_client.cpp:70
...............................................................................

/irods_source/unit_tests/src/test_authenticate_client.cpp:81: PASSED:
  REQUIRE( comm )
with expansion:
  0x000055a2bbf26ea0

/irods_source/unit_tests/src/test_authenticate_client.cpp:88: PASSED:
  CHECK_THROWS( [&comm, &query_str] { irods::query{comm, query_str}; }() )

/irods_source/unit_tests/src/test_authenticate_client.cpp:92: PASSED:
  REQUIRE( 0 == rc_authenticate_client(comm, ctx.dump().c_str()) )
with expansion:
  0 == 0

/irods_source/unit_tests/src/test_authenticate_client.cpp:95: PASSED:
  CHECK( !irods::query{comm, query_str}.empty() )
with expansion:
  true
```
I can of course do something like this if desired:
```c++
try {
    irods::query{comm, query_str};
    CHECK(false);
}
catch (...) {
}
```

unit_tests/src/test_authenticate_client.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+    SECTION("JSON array with expected JSON object in it is not valid")
+    {
+        static const auto ctx = nlohmann::json::array({ia::scheme_name, "native"});
+        CHECK(USER_AUTH_SCHEME_ERR == rc_authenticate_client(&comm, ctx.dump().c_str()));
+    }
+
+    SECTION("Empty JSON object is not valid")
+    {
+        static const auto ctx = nlohmann::json{{}};
```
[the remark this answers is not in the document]
alanking: Would it be acceptable to use this?
```suggestion
        static const auto ctx = nlohmann::json::object{{}};
```

unit_tests/src/test_authenticate_client.cpp
korydraughn: Try calling the original form of the assertion using an authenticated session - i.e. the inputs should cause constructor to succeed resulting in a failing assertion.

Check the output using `-s`. If the assertion fails, I think it's safe to assume your original code was correct and good.

alanking: Here's my diff (put the query just after the authentication)...
```diff
diff --git a/unit_tests/src/test_authenticate_client.cpp b/unit_tests/src/test_authenticate_client.cpp
index 0aae17f84..4892df91b 100644
--- a/unit_tests/src/test_authenticate_client.cpp
+++ b/unit_tests/src/test_authenticate_client.cpp
@@ -90,6 +90,8 @@ TEST_CASE("rc_authenticate_client: minimum viable inputs")
     const auto ctx = nlohmann::json{{ia::scheme_name, env.rodsAuthScheme}};
     REQUIRE(0 == rc_authenticate_client(comm, ctx.dump().c_str()));
 
+    CHECK_THROWS(irods::query{comm, query_str});
+
     // Show that user can now run a query on a protected collection after authenticating.
     CHECK(!irods::query{comm, query_str}.empty());
 }
```
Here's my output:
```
-------------------------------------------------------------------------------
rc_authenticate_client: minimum viable inputs
-------------------------------------------------------------------------------
/irods_source/unit_tests/src/test_authenticate_client.cpp:70
...............................................................................

/irods_source/unit_tests/src/test_authenticate_client.cpp:81: PASSED:
  REQUIRE( comm )
with expansion:
  0x0000555ff6fa2e90

/irods_source/unit_tests/src/test_authenticate_client.cpp:87: PASSED:
  CHECK_THROWS( irods::query{comm, query_str} )

/irods_source/unit_tests/src/test_authenticate_client.cpp:91: PASSED:
  REQUIRE( 0 == rc_authenticate_client(comm, ctx.dump().c_str()) )
with expansion:
  0 == 0

/irods_source/unit_tests/src/test_authenticate_client.cpp:93: FAILED:
  CHECK_THROWS( irods::query{comm, query_str} )
because no exception was thrown where one was expected:

/irods_source/unit_tests/src/test_authenticate_client.cpp:96: PASSED:
  CHECK( !irods::query{comm, query_str}.empty() )
with expansion:
  true
```

korydraughn: Excellent. I think we can trust your original code works as intended.

You can mark this conversation as resolved once things are restored.

alanking: Okay. I think we're good then. Resolving.

korydraughn (commented):
Squash to taste.

korydraughn (approved):
Commit messages look good.

Pound it.

alanking:
#'d, mergin

PR #8425, opened 2025-04-23
[#8407] Remove rcAuthenticate and rsAuthenticate (4-3-stable)

alanking:
Addresses #8407 

Cherry-picked from #8410 

I think confirming that this builds is sufficient, but am willing to run tests.


alanking:
Oh, and in fact, it does not build. :( Switching to draft.

alanking:
The latest changes build, so, marking ready for review. We should probably discuss the implications of this change in a "patch" release (really, it's more like a minor release, but that's just semantics).

korydraughn:
Given these changes would require users to modify their code, let's leave 4.3 as is.

We don't want to force users to change code for 4.3 when iRODS 5 is right around the corner.

alanking:
Sounds good. Closing this and moving the issue back into the 5.0.0 milestone.

Issue #8411, opened 2025-04-21
Document auth plugin JSON request options


alanking:
Auth plugins take JSON "contexts" or "requests" with some options. These may differ from plugin to plugin. Document the options taken by each plugin at the "start" operation which will have some effect.

I believe these are all used by native and pam_password, but not all are used by pam_interactive...
```
/// \code{.js}
/// {
///     // A password to use for authentication.
///     "a_pw": <string>,
///     // Time-To-Live (TTL) value. The server interprets this value as hours.
///     "a_ttl": <integer>,
///     // Forcibly show a password prompt even if the password can be derived by other means.
///     "force_password_prompt": <bool>,
///     // Record auth file (.irodsA) so the user can be authenticated through that in the future.
///     "record_auth_file": <bool>
/// }
/// \endcode
```

PR #8427, opened 2025-04-23
[#8411] Document auth plugin-specific input options (main)

alanking:
Addresses #8411

Not sure if this is the ideal spot, but it felt weird to put it in irods_docs. Few developers will be constructing or interacting with these objects directly, but there aren't many other places to put this information. Open to suggestions.

Here's what the pam_password one looks like:
image

I noticed through this that the auth plugins are in the `irods` namespace but should probably be in the `irods::authentication` namespace. The `authentication_base` class is in `irods::authentication`, so it seems appropriate. I'll look into that.

### DIFF
--- a/plugins/authentication/src/native.cpp
+++ b/plugins/authentication/src/native.cpp
@@ -1,3 +1,5 @@
+/// \file
+
 #include "irods/authentication_plugin_framework.hpp"
 
 #include "irods/authCheck.h"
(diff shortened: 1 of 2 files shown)


alanking:
Created https://github.com/irods/irods/issues/8428 for namespace issue.

Will fix commit message at squash time

korydraughn (approved):
Looks good to me.

I say we roll with the current placement/location you've chosen. We can always change the location of the documentation if it something better is proposed.

alanking:
Sounds good. Fixed commit message and #'d. Mergin

Issue #8378, opened 2025-04-14
Replace SID error code with ZONE_KEY error code


korydraughn:
## Enhancement

See https://github.com/irods/irods/pull/8357#discussion_r2041162326 for details.

Related to #2915.

PR #8429, opened 2025-04-23
[#8378] Replace SID-based error code with ZONE_KEY-based error code (main)

korydraughn:
The following highlights all locations where the `SYS_SIGNED_SID_NOT_MATCHED` is used.

```bash
$ grep -Irn SYS_SIGNED_SID_NOT_MATCHED
server/core/src/irods_server_negotiation.cpp:66:        return ERROR(SYS_SIGNED_SID_NOT_MATCHED, "signed zone_keys do not match");
lib/core/include/irods/rodsErrorTable.h:215:NEW_ERROR(SYS_SIGNED_SID_NOT_MATCHED,                  -147000) // JMC
scripts/irods/test/test_federation.py:1152:                             user.assert_icommand(['ils', '-l', remote_home_collection], 'STDERR', 'SYS_SIGNED_SID_NOT_MATCHED')
```

Given the tiny number of references, there's no need to run the test suite for this PR. We only need to confirm it compiles and that no other error code uses the same integer value.

Now here's something to think about ... Do we want a new error code or do we want a better symbolic name?

I think we can change the symbolic name and compatibility with iRODS 4 would still work because errors are identified by integers. However, doing that would require testing to confirm correctness. I'm currently in the camp of adding a new error code because it's less likely for something to go wrong.

And of course, this new error code will be added to iRODS 4.3.5 and client libraries.


alanking (commented):
Personally, I think the existing error code could just be renamed. It means the same thing on both sides of the conversation regardless of version, and has the same numerical value so it will just show the name from the local client library.

I would think that the only test we would need to confirm is the one being updated in this PR, unless you were referring to something else.

That being said, I don't mind having a new error code.

korydraughn:
> Personally, I think the existing error code could just be renamed. It means the same thing on both sides of the conversation regardless of version, and has the same numerical value so it will just show the name from the local client library.

I mostly agree, but then I started thinking about tools written in other languages (e.g. Go, Java, etc.) and that caused me lose confidence in that approach.

> I would think that the only test we would need to confirm is the one being updated in this PR, unless you were referring to something else.

I was stating there's no need to run tests since the original error code only appeared in exactly three locations throughout the codebase. One in C++ code, the other in the test, and the last one as the definition.

> That being said, I don't mind having a new error code.

Yeah, I think it's the safest of the discussed approaches.

alanking (approved):
New code it is. Ship it

korydraughn:
You made me nervous lol. I'm going to run that one test to be sure everything's still good.

korydraughn:
Added trailing comment to old error code.

The test passed too.
```
 --- IrodsSession: icommand executed by [zonehopper#otherZone] [ils -l /tempZone/home/zonehopper#otherZone] --- 
Assert Command: ils -l /tempZone/home/zonehopper#otherZone
Expecting STDERR: ['ZONE_KEY_SIGNATURE_MISMATCH']
  stdout:                                                                     
    |                   
  stderr:            
    | iRODS filesystem error occurred: [cannot get status: ZONE_KEY_SIGNATURE_MISMATCH]
Output found
```
```
2025-04-23 17:42:15,344 - ==== begin test run results ====
-----
results for [ubuntu-2204-postgres-16_irods-catalog-provider_2]
        passed tests:
                [[  17.2084]s]  [test_federation.Test_ICommands.test_catalog_provider_hosts_other_than_the_first_are_considered__issue_6827]
        skipped tests:
        failed tests:
        return code:[0]
        time elapsed: [    17.21]seconds ([   0]hours [   0.2868]minutes)
-----
All tests passed! :)
time elapsed: [  17.2109]seconds ([   0]hours [ 0.2868]minutes)
```

> Personally, I think the existing error code could just be renamed. It means the same thing on both sides of the conversation regardless of version, and has the same numerical value so it will just show the name from the local client library.

I'm starting to lean back in the direction of just renaming the error code. Success/Failure is communicated through a single integer. Info provided via out parameters and the rError stack are secondary sources. And when it comes to clients written in other languages, the only way they can provide symbolic names is to maintain their own mapping.

I think I'm nearly convinced we can rename the error code.

Are we forgetting anything?

How does the PRC handle error codes?
Does the PRC maintain a table or just report the integer value?

korydraughn:
> How does the PRC handle error codes?
Does the PRC maintain a table or just report the integer value?

The PRC maintains its own mapping, as expected.

https://github.com/irods/python-irodsclient/blob/d83169d2ec9927d45646c3bec97d54d4f31ef92d/irods/exception.py#L228

korydraughn:
The Go library for iRODS also maintains its own mapping.

https://github.com/cyverse/go-irodsclient/blob/d5c6a8edbe7cc8fdbe33c0d5e9606508c252e88d/irods/common/error_code.go#L346

korydraughn:
Same for irods4j.

https://github.com/irods/irods4j/blob/4f5bb6a279449765532f405b8c786aaf4f3cc8d7/src/main/java/org/irods/irods4j/low_level/api/IRODSErrorCodes.java

korydraughn:
I'm going to update this PR to rename the existing error code and rerun that one test to confirm everything's good.

lib/core/include/irods/rodsErrorTable.h
```diff
@@ -212,7 +212,7 @@ NEW_ERROR(SYS_MSSO_OPEN_ERR,                           -135000)
 NEW_ERROR(SYS_MSSO_CLOSE_ERR,                          -136000)
 NEW_ERROR(SYS_RULE_NOT_FOUND,                          -144000) // JMC
 NEW_ERROR(SYS_NOT_IMPLEMENTED,                         -146000) // JMC
-NEW_ERROR(SYS_SIGNED_SID_NOT_MATCHED,                  -147000) // JMC
+NEW_ERROR(SYS_SIGNED_SID_NOT_MATCHED,                  -147000) // EOL since iRODS v5.0.0
```
korydraughn, line 215: - Rename `SYS_SIGNED_SID_NOT_MATCHED` to `ZONE_KEY_SIGNATURE_MISMATCH`
- Remove trailing comment

lib/core/include/irods/rodsErrorTable.h
```diff
@@ -251,6 +251,7 @@ NEW_ERROR(INTERRUPT_DETECTED,                          -182000)
 NEW_ERROR(INVALID_GENQUERY_AGGREGATE_FUNCTION,         -183000)
 NEW_ERROR(INVALID_INPUT_ARGUMENT_NULL_POINTER,         -184000)
 NEW_ERROR(CONNECTION_REFUSED,                          -185000)
+NEW_ERROR(ZONE_KEY_SIGNATURE_MISMATCH,                 -186000)
```
korydraughn, line 254: Remove this.

trel:
having followed along... agreed a rename seems sufficient

lib/core/include/irods/rodsErrorTable.h
korydraughn: Done.

lib/core/include/irods/rodsErrorTable.h
korydraughn: Done.

korydraughn:
Test passed with latest changes.
```
2025-04-23 18:46:11,662 - ==== begin test run results ====
-----
 --- IrodsSession: icommand executed by [zonehopper#otherZone] [ils -l /tempZone/home/zonehopper#otherZone] --- 
Assert Command: ils -l /tempZone/home/zonehopper#otherZone
Expecting STDERR: ['ZONE_KEY_SIGNATURE_MISMATCH']
  stdout:                                                                     
    |                   
  stderr:            
    | iRODS filesystem error occurred: [cannot get status: ZONE_KEY_SIGNATURE_MISMATCH]
Output found
```
```
2025-04-23 18:46:11,662 - ==== begin test run results ====
-----
results for [ubuntu-2204-postgres-16_irods-catalog-provider_2]
        passed tests:
                [[  18.7385]s]  [test_federation.Test_ICommands.test_catalog_provider_hosts_other_than_the_first_are_considered__issue_6827]
        skipped tests:
        failed tests:
        return code:[0]
        time elapsed: [    18.74]seconds ([   0]hours [   0.3123]minutes)
-----
All tests passed! :)
time elapsed: [  18.7414]seconds ([   0]hours [ 0.3124]minutes)
```

korydraughn:
How do we feel about the new symbolic name?

trel:
i think `ZONE_KEY_SIGNATURE_MISMATCH` is fine and good.

also... what else can be done for ... https://github.com/irods/irods/issues/2915

korydraughn:
I know of at least two things that can be addressed. Can look into that in a separate PR.

alanking (commented):
Rename seems good. Squash it

korydraughn:
Squashed.

alanking (approved):
\# it

korydraughn:
Pounded.

Issue #8432, opened 2025-04-24
Failing test: `test_server_rejects_invalid_server_config_options_for_checksum_read_buffer_size_in_bytes__issue_7947`


alanking:
There is an implementation of this test in both `test_irepl` and `test_iput_options`. The test is failing due to new constraints placed on the server configuration value `checksum_read_buffer_size_in_bytes` in this commit: https://github.com/irods/irods/commit/e3af89de2b056dd8ea4405e7fe75d04b89fc3919.

The test should be updated either to expect the configuration reload failures, or it should be removed.

Issue #8428, opened 2025-04-23
Move auth plugin classes to `irods::authentication` namespace


alanking:
- [x] main

---

Currently, these plugins are in the `irods` namespace but the `authentication_base` class from which they inherit is in `irods::authentication` namespace. Let's move these to the same namespace.

Issue #8360, opened 2025-04-08
`pam_password` should return an error when TLS is disabled


alanking:
- [x] irods/irods (main only)
- [x] irods_docs (main only)

---

### Behavioral Change

`pam_password` (and legacy PAM authentication) requires secure communications between client and server. Even if the client and server refuse secure communications (i.e. `CS_NEG_REFUSE`), TLS must still be configured / able to be used in order to use `pam_password`. This is good.

The `pam_password` plugin should not be responsible for enabling/disabling secure communications. It should just return an error if the client and server are not using secure communications.

This is identical to the default behavior of `pam_interactive`. The difference is that `pam_interactive` gives the administrator the option to use insecure communications through an explicit, per-server configuration which is - and this cannot be emphasized enough - **highly discouraged from use in production**: https://github.com/irods/irods_auth_plugin_pam_interactive?tab=readme-ov-file#insecure_mode Adding an `insecure_mode` configuration is the work of #8117.

alanking:
After discussing some more, we have decided we can and should enforce TLS in the server, even if it's "too late". This is a behavioral change, which is okay for 5.0, a major release. In 5.1, we can introduce a step to perform "safe" enforcement of TLS (a la pam_interactive) and that work can be done in #8117.

alanking:
`pam_password` now requires use of TLS and will return an error if it is not enabled. As stated before, we will look into adding a "safe" enforcement of TLS use in the server for 5.1 and include the "insecure mode" feature in #8117.

Closing this one.

Issue #8423, opened 2025-04-23
GitHub Actions: clang-tidy workflow jobs hanging on main


alanking:
- [ ] main

---

Check out the PRs which are open at around the time of writing this (or look at https://github.com/irods/irods/actions/workflows/linter-irods-clang-tidy.yml). Almost all of the jobs are getting cancelled after 360 minutes (6 hours). It's not clear from the output why the jobs are hanging. The problem is not happening on 4-3-stable.

korydraughn:
We have not seen this since the initial creation.

SwooshyCueb:
I still see this in #8366, but I have a workaround.

PR #8434, opened 2025-04-24
Move auth plugins into auth namespace, return error if TLS not enabled for pam_password (main)

alanking:
Addresses #8360 
Addresses #8428 

Unit tests and core tests passed, with an exception noted in #8432.

### DIFF
--- a/plugins/authentication/src/pam_password.cpp
+++ b/plugins/authentication/src/pam_password.cpp
@@ -2,9 +2,6 @@
 
 #include "irods/authentication_plugin_framework.hpp"
 
-#define USE_SSL 1
-#include "irods/sslSockComm.h"
-
 #include "irods/icatHighLevelRoutines.hpp"
 #include "irods/irods_at_scope_exit.hpp"
 #include "irods/irods_auth_constants.hpp"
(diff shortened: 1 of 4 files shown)


plugins/authentication/src/pam_password.cpp
```diff
@@ -2,16 +2,14 @@
...
-#include "irods/sslSockComm.h"
-
 #include "irods/icatHighLevelRoutines.hpp"
 #include "irods/irods_at_scope_exit.hpp"
 #include "irods/irods_auth_constants.hpp"
 #include "irods/irods_client_server_negotiation.hpp"
 #include "irods/irods_logger.hpp"
 #include "irods/irods_pam_auth_object.hpp"
 #include "irods/rcConnect.h"
+#include "irods/sslSockComm.h"
```
alanking, line 12: We may not even need this header anymore, actually

korydraughn (commented):
Changes look good to me.

Is there any documentation in irods/irods_docs which needs to be updated?

plugins/authentication/src/pam_password.cpp
korydraughn: Is anything being used from that header?

alanking: I think the `sslStart` and `sslEnd` calls were the reason for this include. Will confirm that it builds...

alanking:
> Is there any documentation in irods/irods_docs which needs to be updated?

I'll check to see.

Maybe I can address a few of these clang-tidy issues as well. Maybe the hanging issue described in #8423 is healing..?

korydraughn:
Go for it.

alanking:
Yep, wording around TLS requirements with PAM has been adjusted in https://github.com/irods/irods_docs/pull/349.

alanking:
Still waiting on build to complete, but I think we're good.

korydraughn (commented):
The build succeeded. Do the modified tests pass?

What's left for this?

alanking:
Yep, tests are passing. I think I'm happy with it.

korydraughn:
There's still one unresolved comment too.

plugins/authentication/src/pam_password.cpp
alanking: Removed this include and things still compile, so I think we're good. Resolving.

alanking:
Good catch. Resolved convo, and squashed.

korydraughn (approved):
Pound it.

Issue #8439, opened 2025-04-27
Redirect bug from Consumer server in GenQuery2 API


korydraughn:
- [x] main
- [x] 4-3-stable
---
## Bug Report

The following log statement causes the GenQuery2 API to return an error code of -154000 under certain conditions.

https://github.com/irods/irods/blob/c44de02ac648e6000e9c24df0eaa0e7d5fc01f9d/server/api/src/rs_genquery2.cpp#L77

To reproduce, do the following on a Consumer server:
- Set the `api` log category in server_config.json to `trace`
  - Make sure to reload the config for iRODS 5
- Run `iquery "select COLL_NAME"`

I've observed this behavior against tip of main and iRODS 4.3.4.

This is caused by `_input->zone` being null when invoking the logger.

The bug can be worked around by specifying a zone via the `-z` option (e.g. `iquery -z tempZone "select COLL_NAME"`).

PR #8442, opened 2025-04-28
[#8439] GenQuery2: Do not pass nullptr to logger (main)

korydraughn:
Confirmed that the test reproduces the bug and that the C++ change fixes it.

### DIFF
--- a/server/api/src/rs_genquery2.cpp
+++ b/server/api/src/rs_genquery2.cpp
@@ -74,7 +74,7 @@ auto rs_genquery2(RsComm* _comm, Genquery2Input* _input, char** _output) -> int
     }
 
     if (host_info->localFlag != LOCAL_HOST) {
-        log_api::trace("{}: Redirecting request to remote zone [{}].", __func__, _input->zone);
+        log_api::trace("{}: Redirecting request to server [{}].", __func__, host_info->hostName->name);
         return rc_genquery2(host_info->conn, _input, _output);
     }
 
--- a/scripts/irods/test/test_iquery.py
+++ b/scripts/irods/test/test_iquery.py
@@ -3,6 +3,9 @@ import unittest
 
 from . import session
 from .. import lib
+from .. import test
+from ..configuration import IrodsConfig
+from ..controller import IrodsController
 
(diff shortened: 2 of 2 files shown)


alanking (approved):
Looks good to me... # it

korydraughn:
Pounded.

server/api/src/rs_genquery2.cpp
```diff
@@ -74,7 +74,7 @@ auto rs_genquery2(RsComm* _comm, Genquery2Input* _input, char** _output) -> int
     }
 
     if (host_info->localFlag != LOCAL_HOST) {
-        log_api::trace("{}: Redirecting request to remote zone [{}].", __func__, _input->zone);
+        log_api::trace("{}: Redirecting request to server [{}].", __func__, host_info->hostName->name);
```
trel, line 77: why just the server name and not also the zone name?

korydraughn: The zone name is an input argument from the client and is allowed to be null.

korydraughn: I misunderstood your comment. The zone name could be included. I just didn't think to do that since that felt less important than knowing to which server the request was sent.

trel: but when it's NOT null... it IS going to a different zone, so that feels relevant and helpful.     i think that was the initial intent of the line.

korydraughn: Correct. The target zone is still logged.

The original log message was slightly incorrect because it didn't cover redirects within the same zone (i.e. consumer -> provider).

korydraughn: @alanking is referring to these lines. They are the first lines logged whenever the API is invoked (w/ a log level of trace).

https://github.com/irods/irods/blob/a9647277f3010ff5fe9123435bca68181d31b274/server/api/src/rs_genquery2.cpp#L40-L50

trel: gotcha.  resolving.

PR #8444, opened 2025-04-28
Remove python 3 requests as a dependency (main)

korydraughn:
All core tests pass.


cmake/packaging.cmake
```diff
@@ -139,9 +139,9 @@ set(CPACK_DEBIAN_${IRODS_PACKAGE_COMPONENT_SERVER_NAME_UPPERCASE}_PACKAGE_CONTRO
 set(CPACK_RPM_${IRODS_PACKAGE_COMPONENT_SERVER_NAME}_PACKAGE_NAME "irods-server")
 set(CPACK_RPM_${IRODS_PACKAGE_COMPONENT_SERVER_NAME}_PACKAGE_REQUIRES "${IRODS_PACKAGE_DEPENDENCIES_STRING}, irods-runtime = ${CPACK_RPM_PACKAGE_VERSION}-${CPACK_RPM_PACKAGE_RELEASE}, openssl, python3, python3-psutil, libarchive, chkconfig, glibc, libgcc, libstdc++")
 if (IRODS_LINUX_DISTRIBUTION_VERSION_MAJOR EQUAL 7)
```
alanking, line 141: Can't we remove this since we don't support EL7 anymore?

korydraughn: I'm sure we can, but I'm going to leave that to a separate PR so that this one gets in quicker.

alanking: Okay. Let's just leave it, then.

alanking (approved):
\# it

korydraughn:
Added pounds.

Issue #8147, opened 2025-01-30
Server should not signal readiness to service manager until agent factory (and maybe delay server?) are also spun up and ready


SwooshyCueb:
At present, the agent factory and delay server do not tell grandpa that they are finished spinning up, so grandpa usually tells the service manager that startup and reload are finished before the agent factory or delay server are actually ready. Let's introduce a means for the agent factory (and maybe also the delay server) to inform grandpa of their status so that grandpa can be more accurate when communicating the server's status to the service manager.

korydraughn:
The PR merged against this issue makes it so that only the agent factory reports its readiness to the main server process.

The delay server does not need to report its status to the service manager as it is a purely optional service.

If it turns out that communicating the delay server's status to the service manager is important, please open a new issue.

PR #8448, opened 2025-04-28
[#8147] Make agent factory notify main server process of readiness (main)

korydraughn:
This has not been tested yet, but feel free to start reviewing it.

One thing that stands out to me is the fact that this implementation allows the service account user to send SIGUSR2 to the main server process, potentially affecting its ability to accurately detect when the agent factory is ready.

Now, this isn't a huge concern because it requires a user to have access to the service account or root access.

I've left a TODO comment in the code to highlight this situation. We could limit communication of SIGUSR2 to just the agent factory, but I'm not sure that's a good idea. One could argue that allowing the admin to send this signal is actually a good thing.

There's also the question of - when is the agent factory (or config reload) considered to be complete? Is it when the agent factory is able to take a client request? Or maybe it's when the agent factory process has a PID in the process tree?

### DIFF
--- a/server/main_server/src/main.cpp
+++ b/server/main_server/src/main.cpp
@@ -98,9 +98,12 @@ namespace
 
     using log_server = irods::experimental::log::server;
 
-    volatile std::sig_atomic_t g_terminate = 0; // NOLINT(cppcoreguidelines-avoid-non-const-global-variables)
-    volatile std::sig_atomic_t g_terminate_graceful = 0; // NOLINT(cppcoreguidelines-avoid-non-const-global-variables)
-    volatile std::sig_atomic_t g_reload_config = 0; // NOLINT(cppcoreguidelines-avoid-non-const-global-variables)
+    // NOLINTBEGIN(cppcoreguidelines-avoid-non-const-global-variables)
+    volatile std::sig_atomic_t g_terminate = 0;
+    volatile std::sig_atomic_t g_terminate_graceful = 0;
+    volatile std::sig_atomic_t g_reload_config = 0;
+    volatile std::sig_atomic_t g_agent_factory_initialized = 0;
+    // NOLINTEND(cppcoreguidelines-avoid-non-const-global-variables)
 
     // This global variable MUST always hold the max number of child processes forkable
     // by the main server process. It guarantees that the main server process reaps children
@@ -734,6 +737,17 @@ Signals:
             return -1;
         }
 
+        // SIGUSR2 is used by the agent factory to notify the main server process
+        // when it has completed initialization.
+        struct sigaction sa_sigusr2; // NOLINT(cppcoreguidelines-pro-type-member-init)
+        sigemptyset(&sa_sigusr2.sa_mask);
+        sa_sigusr2.sa_flags = 0;
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-type-union-access)
+        sa_sigusr2.sa_handler = [](int) { g_agent_factory_initialized = 1; };
+        if (sigaction(SIGUSR2, &sa_sigusr2, nullptr) == -1) {
+            return -1;
+        }
+
         irods::setup_unrecoverable_signal_handlers();
 
         return 0;
@@ -989,6 +1003,11 @@ Signals:
 
         args.push_back(nullptr);
 
+        // Reset the initialization flag for the agent factory. This flag is how we know when the agent
+        // factory is ready to accept client connections. This flag will be set to 1 when the agent factory
+        // sends the SIGUSR2 signal to the main server process.
+        g_agent_factory_initialized = 0;
+
         g_pid_af = fork();
 
         if (0 == g_pid_af) {
@@ -1009,6 +1028,22 @@ Signals:
         }
 
         log_server::info("{}: Agent Factory PID = [{}].", __func__, g_pid_af);
+
+        // Wait for the agent factory's signal that it has completed initialization. This effectively
+        // means the agent factory has opened its listening socket and is entering its main loop.
+        while (0 == g_agent_factory_initialized) {
+            // If this loop cannot make progress for some reason, allow the admin to stop the server
+            // without needing SIGKILL.
+            if (g_terminate || g_terminate_graceful) {
+                log_server::info(
+                    "{}: Received shutdown instruction. Ending wait loop for agent factory initialization.", __func__);
+                return false;
+            }
+
+            log_server::debug("{}: Waiting for agent factory to complete initialization.", __func__);
+            std::this_thread::sleep_for(std::chrono::milliseconds{250});
+        }
+
         return true;
     } // launch_agent_factory
 
@@ -1130,7 +1165,7 @@ Signals:
         // Wait for the previous agent factory to close its listening socket before launching
         // the new agent factory.
         while (true) {
-            // If this loop cannot make process for some reason, allow the admin to stop the server
+            // If this loop cannot make progress for some reason, allow the admin to stop the server
             // without needing SIGKILL.
             if (g_terminate || g_terminate_graceful) {
                 log_server::info("{}: Received shutdown instruction. Ending reload operation.", __func__);
--- a/server/main_server/src/agent_main.cpp
+++ b/server/main_server/src/agent_main.cpp
@@ -310,6 +310,10 @@ auto main(int _argc, char* _argv[]) -> int
         // This allows rule engine plugins to setup state which doesn't change frequently.
         irods::re_plugin_globals->global_re_mgr.call_setup_operations();
 
+        // Notify the main server process that the agent factory has completed initialization and
+        // is ready for client requests.
+        kill(getppid(), SIGUSR2);
+
         // Enter parent process main loop.
         //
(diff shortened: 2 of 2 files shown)


server/main_server/src/main.cpp
```diff
@@ -734,6 +737,17 @@ Mandatory arguments to long options are mandatory for short options too.
...
+        // SIGUSR2 is used by the agent factory to notify the main server process
+        // when it has completed initialization.
+        struct sigaction sa_sigusr2; // NOLINT(cppcoreguidelines-pro-type-member-init)
+        sigemptyset(&sa_sigusr2.sa_mask);
+        sa_sigusr2.sa_flags = 0;
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-type-union-access)
+        sa_sigusr2.sa_handler = [](int) { g_agent_factory_initialized = 1; }; // TODO Should we constrain this to the agent factory?
+        if (sigaction(SIGUSR2, &sa_sigusr2, nullptr) == -1) {
+            return -1;
+        }
```
korydraughn, line 749: Handle TODO comment.

We can adjust the signal handler to only react to signals from the agent factory. We do that kind of thing in the agent factory code because it makes sense for it to only accept signals from its parent.

It isn't as obvious whether the main server process should only react to this signal from its child.

server/main_server/src/main.cpp
```diff
@@ -1009,6 +1028,21 @@ Mandatory arguments to long options are mandatory for short options too.
...
+            std::this_thread::sleep_for(std::chrono::milliseconds{250});
+            log_server::debug("{}: Waiting for agent factory to complete initialization.", __func__);
+
+            // If this loop cannot make process for some reason, allow the admin to stop the server
+            // without needing SIGKILL.
+            if (g_terminate || g_terminate_graceful) {
+                log_server::info("{}: Received shutdown instruction. Ending wait loop for agent factory initialization.", __func__);
+                return false;
+            }
+        }
```
korydraughn, line 1044: Should this loop terminate after _N_ iterations or _M_ seconds?

We don't do that anywhere else. Perhaps I'm being paranoid and trust the computer will work as intended as long as the design is correct.

server/main_server/src/main.cpp
trel: if the main server can react to sigusr2 from any of its family... we might be able to use that later from a control plane or GUI or something.  but if we don't have a clear use case today, and we can't imagine it being a blocker or headache to limit it to the agent factory, then it seems helpful to constrain it today?

server/main_server/src/main.cpp
```diff
@@ -1009,6 +1028,21 @@ Mandatory arguments to long options are mandatory for short options too.
...
 
         log_server::info("{}: Agent Factory PID = [{}].", __func__, g_pid_af);
+
+        // Wait for the agent factory's signal that it has completed initialization. This effectively
+        // means the agent factory has opened its listening socket and is entering its main loop.
+        while (0 == g_agent_factory_initialized) {
+            std::this_thread::sleep_for(std::chrono::milliseconds{250});
+            log_server::debug("{}: Waiting for agent factory to complete initialization.", __func__);
+
+            // If this loop cannot make process for some reason, allow the admin to stop the server
```
trel, line 1038:
```suggestion
            // If this loop cannot make progress for some reason, allow the admin to stop the server
```

server/main_server/src/main.cpp
trel: computers are different speeds... so feels like M seconds is what an admin will have a better feel for.

we *do* use timeouts for network things... but right, not sure about within the server itself...

as long as the admin can still send `g_terminate || g_terminate_graceful`, maybe we don't need a robust timeout here ... yet.

server/main_server/src/main.cpp
trel: > Is it when the agent factory is able to take a client request? Or maybe it's when the agent factory process has a PID in the process tree?

I think we're most interested in ability to accept requests... the process tree thing would presumably happen 'before' being able to accept requests.. and so therefore would be a 'wider' window... which is less what I think we want than 'ready'.

server/main_server/src/main.cpp
alanking: I somewhat share your concern but I think we can introduce a configuration to time out after a while if this ever becomes an issue.

I'm tempted to say the log message should be bumped up to `info`. If the concern is that we might enter an infinite loop, one might like to know why the server is stuck. There's not really a way to change the log level and see what's going on if it's hanging once the server has started. However, if it takes 1 second from the start of this loop to start the agent factory, you'll get 4 messages in the log at the default level every time the agent factory has to be restarted (which could get annoying).

server/main_server/src/main.cpp
```diff
@@ -1009,6 +1028,21 @@ Mandatory arguments to long options are mandatory for short options too.
         }
 
         log_server::info("{}: Agent Factory PID = [{}].", __func__, g_pid_af);
+
+        // Wait for the agent factory's signal that it has completed initialization. This effectively
+        // means the agent factory has opened its listening socket and is entering its main loop.
+        while (0 == g_agent_factory_initialized) {
+            std::this_thread::sleep_for(std::chrono::milliseconds{250});
+            log_server::debug("{}: Waiting for agent factory to complete initialization.", __func__);
```
alanking, line 1036: Suggesting flipping these since the message about waiting comes after the waiting has occurred :)
```suggestion
            log_server::debug("{}: Waiting for agent factory to complete initialization.", __func__);
            std::this_thread::sleep_for(std::chrono::milliseconds{250});
```

server/main_server/src/main.cpp
```diff
@@ -734,6 +737,17 @@ Mandatory arguments to long options are mandatory for short options too.
...
         }
 
+        // SIGUSR2 is used by the agent factory to notify the main server process
+        // when it has completed initialization.
+        struct sigaction sa_sigusr2; // NOLINT(cppcoreguidelines-pro-type-member-init)
+        sigemptyset(&sa_sigusr2.sa_mask);
+        sa_sigusr2.sa_flags = 0;
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-type-union-access)
+        sa_sigusr2.sa_handler = [](int) { g_agent_factory_initialized = 1; }; // TODO Should we constrain this to the agent factory?
+        if (sigaction(SIGUSR2, &sa_sigusr2, nullptr) == -1) {
```
alanking, line 747:
```suggestion
        if (-1 == sigaction(SIGUSR2, &sa_sigusr2, nullptr)) {
```

server/main_server/src/main.cpp
trel: that does make sense.   not a big threat to leave it open.

server/main_server/src/main.cpp
korydraughn: Given this kind of thing exists in other parts of the redesign, I think we just let it ride until someone reports it as an issue. The server has not shown any behavior that makes me think the loop is an issue.

Will think about the log level bump.

server/main_server/src/main.cpp
korydraughn: I'm going to continue with the existing code since the invocation of `sigaction` is not assignable.

server/main_server/src/main.cpp
korydraughn: Ok. I'm going to leave it open until someone raises an issue.

server/main_server/src/main.cpp
korydraughn: Will do. That suggestion aligns with the rest of the code too.

server/main_server/src/main.cpp
korydraughn: I'm going to leave the log level at debug for now (partly for the reason that @alanking laid out). We can also adjust log levels, etc. My hope is that if an issue does appear at this level, admins can restart the server and reproduce it.

There may also be room for thinking about making the server auto-bump log levels after some criteria is met - e.g. once log message _L_ is executed _N_ times, change the log level of _L_ to a more severe level.

server/main_server/src/main.cpp

server/main_server/src/main.cpp

server/main_server/src/main.cpp
korydraughn: Created #8453.

alanking (commented):
Squash em

korydraughn:
I haven't confirmed this code works, so I'm going to do some testing first.

korydraughn:
Confirmed the changes appear to work.

All unit tests passed.
```
-----
All tests passed! :)
time elapsed: [ 215.9321]seconds ([   0]hours [ 3.5989]minutes)

2025-04-29 16:36:08,737 - version:[4.90.0]
2025-04-29 16:36:08,738 - sha:[e5c19212a64a60b8f328c35af8718670f757bf7b]
```

All core tests passed.
```
-----            
All tests passed! :)
time elapsed: [5562.5941]seconds ([   1]hours [32.7099]minutes)                
                                                                                               
2025-04-29 16:11:28,333 - version:[4.90.0]
2025-04-29 16:11:28,333 - sha:[e5c19212a64a60b8f328c35af8718670f757bf7b]
```

Pounding and merging.

PR #8458, opened 2025-04-30
[8392] Change query in db_set_avu_metadata_op for efficiency (4-3-stable)

JustinKyleJames:
All tests passed.

### DIFF
--- a/plugins/database/src/db_plugin.cpp
+++ b/plugins/database/src/db_plugin.cpp
@@ -8911,7 +8911,9 @@ irods::error db_set_avu_metadata_op(
         std::vector<std::string> bindVars;
         bindVars.push_back( _attribute );
         bindVars.push_back( objIdStr );
-        status = cmlGetMultiRowStringValuesFromSql( "select meta_id from R_OBJT_METAMAP where meta_id in (select meta_id from R_META_MAIN where meta_attr_name=? AND meta_id in (select meta_id from R_OBJT_METAMAP where object_id=?))",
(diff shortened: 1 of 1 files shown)


plugins/database/src/db_plugin.cpp
```diff
@@ -8911,7 +8911,9 @@ irods::error db_set_avu_metadata_op(
         std::vector<std::string> bindVars;
         bindVars.push_back( _attribute );
         bindVars.push_back( objIdStr );
-        status = cmlGetMultiRowStringValuesFromSql( "select meta_id from R_OBJT_METAMAP where meta_id in (select meta_id from R_META_MAIN where meta_attr_name=? AND meta_id in (select meta_id from R_OBJT_METAMAP where object_id=?))",
+        status = cmlGetMultiRowStringValuesFromSql( "SELECT meta_id FROM r_objt_metamap WHERE meta_id IN "
```
trel, line 8914: table names need to remain all caps for multi-database compatibility - we have learned this over time.

JustinKyleJames: I updated the code to make the table names upper-case and the SQL keywords lowercase (including the limit) to match the existing query.

I ran the metadata tests for both postres and mysql with the latest commit.

trel:
code looks good to me.

please explain in the commit message why 'limit 3' is sufficient.

trel:
good - please # it.

korydraughn (approved):

Issue #7229, opened 2023-07-30
Redesign/Refactor/Simplify main server logic


korydraughn:
## Redesign / Refactor
The startup logic for the iRODS server is complicated and difficult to reason about.

Due to a mixture of forking and multi-threading, adding new features to the main server process is tricky. We need to simplify how startup is carried out so that iRODS is in a better position to take on future requirements.

A simpler architecture could be:
- Main iRODS process forks child processes
    - Perhaps this becomes pluggable
- Main iRODS process makes sure all child processes are running, forking new ones if they disappear
- Child processes have one job
    - e.g. Agent factory, Delay server, Control plane, etc
- Agent factory manages port 1247
- Control plane manages port 1248
- Access to global state must be through interfaces or services
    - i.e. Dependency injection
- Initialization of global state must be deterministic
- Everything must be ASan, Ubsan, Tsan friendly

Things to consider:
- How should child processes communicate?
- How should the main server process relay information to children?
- How should child processes relay information to main server process?
- Where does the CRON manager run?
- Who is responsible for the CRON manager?
- How should monitoring be supported?
- What should be monitored?
- What things are considered iRODS 5.0 territory?

To keep things interesting, it must still be backwards compatible with 4.3.

korydraughn:
This is likely an iRODS 6.0 thing, but definitely worth considering at some point.
- https://unixism.net/2019/04/linux-applications-performance-introduction/

That is, we investigate changing the process model to pre-forked or pre-threaded. If we do this, the performance should improve, but we have to make sure there are no memory leaks. That means incorporating Asan and other tools during testing.

korydraughn:
This comment is mostly so others have some idea about what's happening regarding iRODS 5.

### Current Status

- The iRODS 5 server is runnable _(not for production use yet)_
- Startup is faster
- Ownership of components, memory, etc is now clearer
  - e.g. The main server process watches children and handles delay server migration
  - e.g. The agent factory is in charge of all information required by `ips`
  - e.g. The agent factory manages port 1247
- Signal handling for `SIGINT,` `SIGTERM,` and `SIGHUP` are working as intended
  - Signals are sent to child processes appropriately
  - The agent factory and delay server now respond to `SIGHUP,` but only if the main server process sent it
- The main server process and agent factory no longer rely on multiple threads
- The need for a control plane has been removed (so far)
- We've successfully avoided creating dependencies on Python for startup
- The server always performs a graceful shutdown
[...]

PR #8488, opened 2025-05-08
[#7229] Increase rodsLog verbosity (main)

korydraughn:
Required by irods/irods_capability_indexing#172.

### DIFF
--- a/lib/core/src/rodsLog.cpp
+++ b/lib/core/src/rodsLog.cpp
@@ -27,6 +27,7 @@ const static std::map<const int, const std::string> irods_error_map = irods_erro
 #include <unistd.h>
 #include "irods/irods_exception.hpp"
 #include "irods/irods_logger.hpp"
+#include "irods/irods_configuration_keywords.hpp"
 
 #define BIG_STRING_LEN MAX_NAME_LEN+300
 #include <cstdarg>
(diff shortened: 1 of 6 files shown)


alanking (commented):
Seems fine to me. I don't quite understand the changes so I created https://github.com/irods/irods/issues/8489 to update the docs since I think we forgot to do that.

Will review companion PR and come back to this one.

alanking (dismissed):
If it works as expected, # it

alanking:
Ah, got it. Thanks for the explanation. Maybe that will help future mes and yous too :p

korydraughn:
Will handle clang-format after testing.

lib/core/src/rodsLog.cpp
```diff
@@ -69,6 +70,30 @@ std::string create_log_error_prefix() {
...
+        case level::debug:      return LOG_DEBUG7;
+        case level::info:       return LOG_NOTICE;
+        case level::warn:       return LOG_WARNING;
+        case level::error:      return LOG_ERROR;
+        case level::critical:   return LOG_SYS_FATAL;
+    }
+
+    // We should never reach this line because get_level_from_config() always returns
+    // a valid log level (i.e. it catches all exceptions and defaults to level::info).
+    return LOG_SQL;
```
alanking, line 94: Would it make more sense to return `LOG_NOTICE` here, based on the comment?

alanking (commented):
One little suggestion, but otherwise looks good.

lib/core/src/rodsLog.cpp
alanking: Okay. The comment kind of suggested that the default would result in `LOG_NOTICE`, which is why I said that. Perhaps we can add to the comment to explain why `LOG_SQL` is returned instead? What you've written here is good enough explanation, IMO.

korydraughn: Will do.

korydraughn: Done.

alanking (commented):
Looks good. Squash em

korydraughn:
Waiting for core tests to finish/pass. Will squash as soon as the tests report success.

korydraughn:
Made one change to using C++20 `using enum log_ns::level` to reduce some noise. Feature was introduced in Clang 13 so we should be good.

I'll confirm that doesn't break anything by doing some manual testing. Shouldn't need to rerun unit or core tests.

korydraughn:
Investigating two test failures:
- `test_rule_engine_plugin_passthrough`
- `test_rule_engine_plugin_framework.Test_Rule_Engine_Plugin_Framework`

korydraughn:
Fixed the failing tests.

Rerunning core tests to make sure the other tests are still ok.

alanking (commented):
Good. Please squash to taste and apply issue numbers, no #s.

korydraughn:
Squashed.

alanking (approved):
\# em

Issue #6283, opened 2022-03-29
Remove absolute paths from config files and source code


korydraughn:
Having absolute paths in the code base makes it difficult to support certain installation schemes.

See https://github.com/irods/irods/pull/6275#discussion_r835961113 for more details.

korydraughn:
@SwooshyCueb Any thoughts on this?

I tempted to say this is complete, but based on the linked comment, you may be aware of other things that need attention before we close this.

SwooshyCueb:
There's still tons in the test suite and other scripts

SwooshyCueb:
Additionally, as far as I am aware, the schema files still contain absolute paths, which was what this issue was originally opened to address

trel:
most/all of the schemas now have a .in file... which are populated/converted upon installation I believe...

```
    "$id": "file:///@IRODS_CONFIGURATION_SCHEMA_PREFIX@@IRODS_HOME_DIRECTORY@/configuration_schemas/v4/server.json",
```

korydraughn:
Correct. We ship with some defaults, but users are free to change the values during or after setup.

What would be the alternative for this?

SwooshyCueb:
> most/all of the schemas now have a .in file... which are populated/converted upon installation I believe...

Referring back to the linked discussion, we were previously in agreement that having the absolute paths in the schema files, regardless of when they're written down, is a bad idea. In fact, I'll go even further and say that I am not convinced that they should be referring to themselves by filesystem path in the first place.

SwooshyCueb:
Also, to clarify, these values are not set during package installation. They are set by the buildsystem. This is a problem for tarball packages, relocatable RPM packages, etc

korydraughn:
That's reasonable.

Q. How do you imagine these absolute paths be handled?
Q. Is there an established convention for this we can reference?

If you have ideas about how to make this better, please add them here.

korydraughn:
Are you proposing we stop shipping schema files altogether? _(I'm not opposed to shipping fewer things. It's more about making sure we cover the negative space)_

What would the default be for `schema_validation_base_uri` in server_config.json?
Is it disabled on a brand new install?
What happens if internet access isn't allowed due to company policy?

Are these questions answered via the setup script? _(because we've updated it accordingly)_

SwooshyCueb:
> Are you proposing we stop shipping schema files altogether?

[...]

PR #8523, opened 2025-05-19
iRODS 5: Fix upgrade_irods.py's handling of `clients_server_policy` and `tls_client` (main)

korydraughn:
I've confirmed the changes work as intended.


scripts/irods/upgrade_configuration.py
```diff
@@ -302,6 +302,10 @@ def update_base(base, updates):
     new_server_config.pop('server_control_plane_port', None)
     new_server_config.pop('server_control_plane_timeout_milliseconds', None)
 
+    # Add client-server negotiation policy for outbound traffic based on the server account client environment.
+    if "irods_client_server_policy" in service_account_environment:
+        new_server_config["client_server_policy"] = service_account_environment.get("irods_client_server_policy", None)
```
korydraughn, line 307: This can be improved by doing one of the following:
- Remove L306
- Replace `.get` with `[key]`

scripts/irods/upgrade_configuration.py
```diff
@@ -332,9 +336,13 @@ def update_base(base, updates):
     # Add TLS properties for outbound traffic based on the service account client environment.
     add_verify_server_property = False
     if "irods_ssl_ca_certificate_file" in service_account_environment:
+        if "tls_client" not in new_server_config:
+            new_server_config["tls_client"] = {}
         new_server_config["tls_client"]["ca_certificate_file"] = service_account_environment.get("irods_ssl_ca_certificate_file", None)
```
korydraughn, line 341: `.get(key, None)` can be replaced with `[key]` because of the if-statement.

Same for the code that follows.

korydraughn:
With the latest changes, I get the following when the service account does not define `irods_client_server_policy`.

```
irods@0f99e25d38f7:~$ grep -C4 client_server_policy /etc/irods/server_config.json
    "catalog_provider_hosts": [
        "0f99e25d38f7"
    ],
    "catalog_service_role": "provider",
    "client_server_policy": null,
    "connection_pool_refresh_time_in_seconds": 300,
    "controlled_user_connection_list": {
        "control_type": "denylist",
        "users": []
irods@0f99e25d38f7:~$ irodsServer -c
validate_configuration: [
    {
        "valid": false, 
        "evaluationPath": "/properties/client_server_policy/enum", 
        "schemaLocation": "https://irods.org/configuration_schemas/v5/server_config.json#/properties/client_server_policy/enum", 
        "instanceLocation": "/client_server_policy", 
        "error": "'null' is not a valid enum value."
    }
]
```

scripts/irods/upgrade_configuration.py
korydraughn: Done.

scripts/irods/upgrade_configuration.py
korydraughn: Done.

korydraughn:
That comes from the $id property and does not need to be network addressable.

See https://github.com/irods/irods/issues/6283#issuecomment-1631039060.

trel:
yes, i remember that conversation.   just still looks weird to see something as an http endpoint... where there is nothing published.

korydraughn:
I get it. We can host something there, but hard to say whether there's value in doing that.

korydraughn:
I've confirmed upgrades work with and without TLS.

Squashing and pounding so we can start building packages.

korydraughn:
Merging.

PR #8565, opened 2025-06-25
[irods/irods_resource_plugin_s3#2146] Calling fileNotify on close (main)

JustinKyleJames:
All tests passed.  This goes along with irods/irods_resource_plugin_s3#2259 to fix issue irods/irods_resource_plugin_s3#2146. 

This is to fix S3 issue 2146. The issue occurs on large file transfers when the resource is in attached and decoupled modes and a redirect is required.  The following is the rough description assuming the client is connected to server A and the resource is attached to server B.

 1. s3_resolve_resc_hier_operation is run on server A.
 2. The db_reg_data_obj_op operation is called on the catalog provider to create the data object entry. After this there is a record in the database for the replica.
 3. The request is redirected to server B.
 4. On server B, in s3_file_create_operation the physical path is paved over. This allows the plugin to write to the correct S3 key.
 5. If the file is a small file, s3_notify_operation is called on server A. The plugin uses this to pave over the physical path within server A's agent memory.
 6. If the file is a large file, the notification operation is not called on server A. The plugin has no opportunity to pave over the physical path within server A's agent memory.
 7. The file is written to S3 (from server B).

There is no opportunity to update the physical path on server A and therefore the database never gets updated.

This fix is to call fileNotify on the close.  This gives the S3 resource the opportunity to overwrite the physical path in memory which gets represented back in the DB.

Notes:
1. The object ID (reversed) is used as the path in decoupled mode. This is done as it provides a deterministic unique path.
2. I considered a hash of the path in iRODS but this is not guaranteed to be unique as a file can be put and moved and a new file put to this path.
3. A path plus date is not deterministic as the date could change between server A and server B paving over the path, and is also not guaranteed to be unique if a move happens in the same day.
4. The problem with the object ID is that it doesn't get created until step 2, and for large file uploads step 1 is the only opportunity to pave over the path on server A.

This solution seems to check all of the boxes with a simple extra call to fileNotify. This is a NOOP for all other resource types.

### DIFF
--- a/server/api/src/rsDataObjClose.cpp
+++ b/server/api/src/rsDataObjClose.cpp
@@ -53,6 +53,8 @@
 #include "irods/key_value_proxy.hpp"
 #include "irods/replica_access_table.hpp"
 #include "irods/replica_state_table.hpp"
+#include "irods/irods_resource_redirect.hpp"
+#include "irods/irods_logger.hpp"
 
 #define IRODS_REPLICA_ENABLE_SERVER_SIDE_API
 #include "irods/data_object_proxy.hpp"
@@ -69,6 +71,7 @@ namespace
     namespace ill = irods::logical_locking;
     namespace ir = irods::experimental::replica;
     namespace rst = irods::replica_state_table;
+    using log_api = irods::experimental::log::api;
 
     auto apply_static_peps(RsComm& _comm, openedDataObjInp_t& _close_inp, const int _fd, const int _operation_status) -> void
     {
@@ -727,6 +730,23 @@ int rsDataObjClose(rsComm_t* rsComm, openedDataObjInp_t* dataObjCloseInp)
     std::unique_ptr<irods::at_scope_exit<std::function<void()>>> restore_entry;
 
     int ec = 0;
+
+    // Call notify on resource plugin if this is a write. This allows S3 to pave over the physical path
+    // before it is written to the database in case decoupled mode is being used.  See
+    // irods/irods_resource_plugin_s3#2146.
+    if ((l1desc.dataObjInp->openFlags & O_ACCMODE) != O_RDONLY) {
+        irods::file_object_ptr file_obj(new irods::file_object(rsComm, L1desc[fd].dataObjInfo));
+
+        irods::error ret = fileNotify(rsComm, file_obj, irods::WRITE_OPERATION);
+
+        if (!ret.ok()) {
+            const std::string msg = fmt::format("Failed to signal the resource that the data object [{}] was modified.",
+                                                L1desc[fd].dataObjInfo->objPath);
+            ret = PASSMSG(msg, ret);
+            log_api::error(ret.result());
+            return ret.code();
+        }
+    }
 
     // Replica access tokens only apply to write operations.
     if ((l1desc.dataObjInp->openFlags & O_ACCMODE) != O_RDONLY) {


trel:
And the cost of this fileNotify call is a single network round trip from serverB back to serverA for each large file transfer?

server/api/src/rsDataObjClose.cpp
```diff
@@ -728,6 +729,31 @@ int rsDataObjClose(rsComm_t* rsComm, openedDataObjInp_t* dataObjCloseInp)
...
+
+        irods::file_object_ptr file_obj(
+            new irods::file_object(
+                rsComm,
+                L1desc[fd].dataObjInfo));
+
+        irods::error ret = fileNotify(
+            rsComm,
+            file_obj,
+            irods::WRITE_OPERATION);
```
alanking: There is a redirect to remote zones below in the event that this data object is not in this zone. Should we put this call to fileNotify after that? Doesn't matter that much since it would likely be a no-op.

JustinKyleJames:
> And the cost of this fileNotify call is a single network round trip from serverB back to serverA for each large file transfer?

Yes, just one additional fileNotify is called.

server/api/src/rsDataObjClose.cpp
JustinKyleJames: > There is a redirect to remote zones below in the event that this data object is not in this zone. Should we put this call to fileNotify after that? Doesn't matter that much since it would likely be a no-op.

I haven't thought about the remote zone case.  I will have to test it out.

JustinKyleJames:
> Have we thought through any possible surprises from the fact that fileNotify is also being called in DataObjWrite API?
> 
> Do we need to do something similar in replica_close API (when finalizing)?

On your first question, I did some more testing and verified that for small files, the fileNotify in DataObjWrite is the only one that gets called.  For large files, the fileNotify in DataObjClose is the only one that gets called.

I expected to have two fileNotify calls for small files but that seems to not be the case.  In either case, calling that twice is not a problem.

As for the second question, I'll have to do some testing with PRC.

server/api/src/rsDataObjClose.cpp
JustinKyleJames: > I haven't thought about the remote zone case. I will have to test it out.

I tested this scenario in federation and it worked fine.

alanking: Okay. My only other concern about this is when we expect fileNotify to return an error. I'm concerned about the replica access table entry not getting cleaned up just below this. I suppose the L1 descriptor would also not get cleaned up, so perhaps it is actually better this way. Letting it ride.

server/api/src/rsDataObjClose.cpp
```diff
@@ -728,6 +729,31 @@ int rsDataObjClose(rsComm_t* rsComm, openedDataObjInp_t* dataObjCloseInp)
...
+
+        if ( !ret.ok() ) {
+            std::stringstream msg;
+            msg << "Failed to signal the resource that the data object \"";
+            msg << L1desc[fd].dataObjInfo->objPath;
+            msg << "\" was modified.";
```
alanking: I know this came from DataObjWrite, but consider replacing this with `fmt::format`.

alanking (commented):
> On your first question, I did some more testing and verified that for small files, the fileNotify in DataObjWrite is the only one that gets called. For large files, the fileNotify in DataObjClose is the only one that gets called.
>
> I expected to have two fileNotify calls for small files but that seems to not be the case. In either case, calling that twice is not a problem.

This is because DataObjPut does not call DataObjClose on single buffer transfers. This is not great, but explains the behavior.

> As for the second question, I'll have to do some testing with PRC.

Will await results.

server/api/src/rsDataObjClose.cpp
```diff
@@ -728,6 +729,31 @@ int rsDataObjClose(rsComm_t* rsComm, openedDataObjInp_t* dataObjCloseInp)
...
+            ret = PASSMSG( msg.str(), ret );
+            irods::log( ret );
```
korydraughn: Let's use the new logger.

JustinKyleJames: How do I use the new logger with an irods::error() object?  I can't find an example of that.

korydraughn: Like @alanking said, I think you can replace the contents of this entire if-block with a single call to the new logger.
```cpp
log_api::error("{}: Failed to signal the resource that the data object [{}] was modified; error_code=[{}], error_message=[{}]",
               __func__,
               L1desc[fd].dataObjInfo->objPath,
               ret.code(),
               ret.result());
return ret.code();
```
Or ...
```cpp
auto msg = fmt::format("Failed to signal the resource that the data object [{}] was modified."
                       L1desc[fd].dataObjInfo->objPath);
ret = PASSMSG(msg, ret);
log_api::error(ret.result());
return ret.code();
```

korydraughn:
Changes look good. Please run clang-format.

JustinKyleJames:
> Changes look good. Please run clang-format.

done

korydraughn:
Squash it.

JustinKyleJames:
> Squash it.

done

korydraughn:
The lines of your commit message require horizontal scrolling within tig. The content is fine, but consider breaking lines up so that they fit within smaller panes. For example:
```
[irods/irods_resource_plugin_s32146] Calling fileNotify on close

This is to fix S3 issue 2146. The issue occurs on large file transfers when
the resource is in attached and decoupled modes and a redirect is required.
The following is the rough description assuming the client is connected to
server A and the resource is attached to server B.

 1. s3_resolve_resc_hier_operation is run on server A.
 2. The db_reg_data_obj_op operation is called on the catalog provider to
    create the data object entry. After this there is a record in the database
    for the replica.
 3. The request is redirected to server B.
 4. On server B, in s3_file_create_operation the physical path is paved over.
    This allows the plugin to write to the correct S3 key.
 5. If the file is a small file, s3_notify_operation is called on server A. The
    plugin uses this to pave over the physical path within server A's agent memory.
 6. If the file is a large file, the notification operation is not called on server
    A. The plugin has no opportunity to pave over the physical path within server
    A's agent memory.
 7. The file is written to S3 (from server B).

There is no opportunity to update the physical path on server A and therefore
the database never gets updated.

This fix is to call fileNotify on the close. This gives the S3 resource the
opportunity to overwrite the physical path in memory which gets represented back
in the database.

Notes:
1. The object ID (reversed) is used as the path in decoupled mode. This is
   done as it provides a deterministic unique path.
2. I considered a hash of the path in iRODS but this is not guaranteed to be
   unique as a file can be put and moved and a new file put to this path.
3. A path plus date is not deterministic as the date could change between
   server A and server B paving over the path, and is also not guaranteed to be
   unique if a move happens in the same day.
4. The problem with the object ID is that it doesn't get created until step 2,
   and for large file uploads step 1 is the only opportunity to pave over the path
   on server A.

This solution seems to check all of the boxes with a simple extra call to
fileNotify. This is a NOOP for all other resource types.
```

Also, let's replace `DB` in the commit message body with `database`.

JustinKyleJames:
> The lines of your commit message require horizontal scrolling within tig. The content is fine, but consider breaking lines up so that they fit within smaller panes. For example:
> 
> ```
> [irods/irods_resource_plugin_s32146] Calling fileNotify on close
> 
> This is to fix S3 issue 2146. The issue occurs on large file transfers when
> the resource is in attached and decoupled modes and a redirect is required.
> The following is the rough description assuming the client is connected to
> server A and the resource is attached to server B.
> 
>  1. s3_resolve_resc_hier_operation is run on server A.
>  2. The db_reg_data_obj_op operation is called on the catalog provider to
>     create the data object entry. After this there is a record in the database
>     for the replica.
>  3. The request is redirected to server B.
>  4. On server B, in s3_file_create_operation the physical path is paved over.
>     This allows the plugin to write to the correct S3 key.
>  5. If the file is a small file, s3_notify_operation is called on server A. The
>     plugin uses this to pave over the physical path within server A's agent memory.
>  6. If the file is a large file, the notification operation is not called on server
>     A. The plugin has no opportunity to pave over the physical path within server
>     A's agent memory.
>  7. The file is written to S3 (from server B).
> 
> There is no opportunity to update the physical path on server A and therefore
> the database never gets updated.
> 
> This fix is to call fileNotify on the close. This gives the S3 resource the
> opportunity to overwrite the physical path in memory which gets represented back
> in the database.
> 
> Notes:
> 1. The object ID (reversed) is used as the path in decoupled mode. This is
>    done as it provides a deterministic unique path.
> 2. I considered a hash of the path in iRODS but this is not guaranteed to be
>    unique as a file can be put and moved and a new file put to this path.
> 3. A path plus date is not deterministic as the date could change between
>    server A and server B paving over the path, and is also not guaranteed to be
>    unique if a move happens in the same day.
> 4. The problem with the object ID is that it doesn't get created until step 2,
>    and for large file uploads step 1 is the only opportunity to pave over the path
>    on server A.
> 
> This solution seems to check all of the boxes with a simple extra call to
> fileNotify. This is a NOOP for all other resource types.
> ```
> 
> Also, let's replace `DB` in the commit message body with `database`.

I made the changes.

korydraughn (approved):
Pound it.

korydraughn:
Please open an issue in this repo for this specific change and link it to the commit and S3 issue.

The irods/irods issue gives us a milestone and helps with the release notes.

JustinKyleJames:
> Please open an issue in this repo for this specific change and link it to the commit and S3 issue.

I opened the issue, linked it to the original, and added the commit message.  I also pounded it.

korydraughn:
Got it. Merging.

PR #8495, opened 2025-05-08
[6239] fix for mangled collections upon rename and/or move (main)

tsmeele:
Use RDBMS internal function to calculate stringlength as input to substr() operations, rather than relying on a client-calculated literal. This avoids differences in interpretation of string length related to unicode characters that span multiple bytes.

This is a revised pull request, now with a more complete (and tested :) fix. The prior pull request, still open, provided only a fix for a move, not a rename, of a collection.

PR #8567, opened 2025-06-30
Fix for mangled collection names on move, plus tests, plus Oracle (main)

alanking:
Addresses #6239

Supersedes #8495 (@tsmeele still gets credit for the pertinent fix)

Confirmed that (expected) multi-byte character tests fail in the expected manner as shown in #6239 before fix, and tests pass after the fix for both PostgreSQL and MySQL, including in topology from provider and consumer. I have not tested against Oracle, but the documentation states that the functions used should be equivalent.

The not-multi-byte character test (origin of #6239) passes with or without the fix, but was added to ensure we do not go backwards.

Unit tests passed. Running full suite. Will mark ready for review once done, but feel free to review in the meantime.

### DIFF
--- a/plugins/database/src/db_plugin.cpp
+++ b/plugins/database/src/db_plugin.cpp
@@ -10266,16 +10266,16 @@ irods::error db_rename_object_op(
            part, _new_name string, and then (if any for each row) the
            tailing part of the name.
            (In the sql substr function, the index for sql is 1 origin.) */
-        snprintf( collNameSlash, MAX_NAME_LEN, "%s/", collName );
+        snprintf(collNameSlash, MAX_NAME_LEN, "%s/", collName);
         snprintf( slashNewName, MAX_NAME_LEN, "/%s", _new_name );
         if ( isRootDir ) {
             snprintf( slashNewName, MAX_NAME_LEN, "%s", _new_name );
         }
-	/* Instead of preparing the string length here, we have the RDBMS do its own length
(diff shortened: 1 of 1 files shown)


alanking:
Tests passed for PostgreSQL. Running for MySQL now.

plugins/database/src/db_plugin.cpp
```diff
@@ -10700,23 +10710,32 @@ irods::error db_move_object_op(
            part, endCollName string, and then (if any for each row) the
            tailing part of the name.
            (In the sql substr function, the index for sql is 1 origin.) */
-        snprintf( ocLenStr, MAX_NAME_LEN, "%d", ocLen + 1 );
         snprintf( collNameSlash, MAX_NAME_LEN, "%s/", oldCollName );
-        len = strlen( collNameSlash );
-        snprintf( collNameSlashLen, 10, "%d", len );
+
+        /* See remark earlier on rational for having string length calculations
```
trel:
```suggestion
        /* See remark earlier on rationale for having string length calculations
```

alanking:
Core tests passed for PostgreSQL and MySQL. Marking ready for review.

plugins/database/src/db_plugin.cpp

korydraughn (commented):
Squash it.

alanking:
Squashed

korydraughn (approved):
Pound it.

alanking:
#'d, mergin

PR #8568, opened 2025-07-01
Fix for mangled collection names on move, plus tests, plus Oracle (4-3-stable)

alanking:
Addresses https://github.com/irods/irods/issues/6239

Cherry-picked from https://github.com/irods/irods/pull/8567

Confirmed core tests pass for MySQL. Since this is identical to the one on main and we test on PostgreSQL all the time, I think that's good enough.


alanking:
#'d, mergin

Issue #8373, opened 2025-04-11
Possible memory leak in `rsDataObjGet`


alanking:
- [ ] main
- [ ] 4-3-stable

---

## Bug Report

Traced this back to 4.2.9 (of course)

The following code appears to present a number of problems and potential problems:
https://github.com/irods/irods/blob/e9b92d2add9d1412cd4872322acee662327896c8/server/api/src/rsDataObjGet.cpp#L280-L282

First, `dataObjOutBBuf->buf` is allocated, and then immediately gets set to `nullptr` via the `std::memset` on the next line. If `single_buffer_size` is any value above 0, this results in a memleak (and also, I have no idea how this ever worked, or if it does work). To fix this, I believe the line should be:
```
std::memset(dataObjOutBBuf->buf, 0, sizeof(single_buffer_size)); 
```

Doing this creates another problem, though. If `single_buffer_size` is 0, the `std::malloc` behavior is "implementation defined" (https://en.cppreference.com/w/cpp/memory/c/malloc):

> If size is zero, the behavior is implementation defined (null pointer may be returned, or some non-null pointer may be returned that may not be used to access storage, but has to be passed to [std::free](https://en.cppreference.com/w/cpp/memory/c/free)).

If the implementation returns a `nullptr`, we `std::memset` a `nullptr` which is UB in C `memset`. `std::memset` may be different in this regard, but it still seems unsafe. To address this, we should not allow a `single_buffer_size` less than 1.

alanking:
Looking at this again... this is not a leak.

`dataObjOutBBuf` is a `BytesBuf`, the definition for which is found here: https://github.com/irods/irods/blob/86eae2f46342a780563b34f4874e976f4847a879/lib/core/include/irods/rodsDef.h#L179-L182

`single_buffer_size` is an `int`: https://github.com/irods/irods/blob/e9b92d2add9d1412cd4872322acee662327896c8/server/api/src/rsDataObjGet.cpp#L277

`sizeof(single_buffer_size)` will result in the same value as `sizeof(int)`. The first member of `BytesBuf` is `len`, which is an `int`.

Here's the `memset` call we are questioning:
```
 std::memset(dataObjOutBBuf, 0, sizeof(single_buffer_size));
```

All this does is set the value of `len` to 0. However, the intention here was obviously to clear the newly allocated buffer, so we should update it to actually do that.

PR #8570, opened 2025-07-02
[#8373] DataObjGet: Enforce max size for single buffer (main)

alanking:
Addresses #8373

`test_iget` passes. Running other core tests now.

### DIFF
--- a/server/api/src/rsDataObjGet.cpp
+++ b/server/api/src/rsDataObjGet.cpp
@@ -118,7 +118,7 @@ namespace
         BytesBuf* dataObjOutBBuf)
     {
         char *chksumStr = NULL;
-        const auto free_checksum_str = irods::at_scope_exit{[chksumStr] { std::free(chksumStr); }};
+        const auto free_checksum_str = irods::at_scope_exit{[&chksumStr] { std::free(chksumStr); }};
 
         dataObjInfo_t* dataObjInfo = L1desc[_fd].dataObjInfo;
         copyKeyVal( &dataObjInp->condInput, &dataObjInfo->condInput );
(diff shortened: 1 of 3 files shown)


scripts/irods/test/test_iget.py
```diff
@@ -123,3 +121,58 @@ def test_iget_does_not_change_replica_status__issue_5760(self):
...
+            buffer_size_in_mb = 2048
+            server_config["advanced_settings"]["maximum_size_for_single_buffer_in_megabytes"] = buffer_size_in_mb
+
+            with lib.file_backed_up(server_config_path):
+                with open(server_config_path, 'w') as f:
+                    f.write(json.dumps(server_config, sort_keys=True, indent=4, separators=(',', ': ')))
+
+                IrodsController().reload_configuration()
+
+                self.user.assert_icommand(["iget", logical_path, "-"], "STDERR", "CONFIGURATION_ERROR")
```
korydraughn: Let's include the error code in the expected output too.

server/api/src/rsDataObjGet.cpp
```diff
@@ -284,13 +284,32 @@ int rsDataObjGet(
...
-            dataObjOutBBuf->buf = std::malloc(single_buffer_size);
-            std::memset(dataObjOutBBuf, 0, sizeof(single_buffer_size));
-            dataObjOutBBuf->len = single_buffer_size;
+            THROW(CONFIGURATION_ERROR,
+                  fmt::format("{}: [{}] configuration value is invalid. Must be between {} and {}, inclusive.",
+                              __func__,
+                              irods::KW_CFG_MAX_SIZE_FOR_SINGLE_BUFFER,
+                              minimum_value_for_max_single_buffer_size_in_mb,
+                              maximum_value_for_max_single_buffer_size_in_mb));
+        }
```
korydraughn: Seems this could be replaced by adding more constraints on the property within the json schema?

https://github.com/irods/irods/blob/86eae2f46342a780563b34f4874e976f4847a879/schemas/configuration/v5/server_config.json.in#L75-L78

server/api/src/rsDataObjGet.cpp
```diff
@@ -284,13 +284,32 @@ int rsDataObjGet(
...
+
+        const auto buffer_size_in_bytes = buffer_size_in_mb * 1024 * 1024;
+        const auto replica_size = L1desc[fd].dataObjInfo->dataSize;
+        if (replica_size <= buffer_size_in_bytes && UNKNOWN_FILE_SZ != replica_size) {
+            irods::log(LOG_NOTICE, "single buffer put");
```
korydraughn: Please swap out with new logger.

alanking: Whoops, that's not supposed to be there. Leftover from debugging.

server/api/src/rsDataObjGet.cpp
alanking: I thought about doing this, but then I won't be able to test this unless we do not enforce the schema on reloading configuration. Is that something we are interested in trying?

server/api/src/rsDataObjGet.cpp
alanking: Done

server/api/src/rsDataObjGet.cpp
korydraughn: I think we can safely lean on the schema. There are other situations like this in the server. To test those situations, we lean on the config validation check (`irodsServer -c`).

https://github.com/irods/irods/blob/86eae2f46342a780563b34f4874e976f4847a879/scripts/irods/test/test_misc.py#L479-L510

As long as we can guarantee that config validation is performed before the server takes any changes, we're good.

alanking: Okay. I will replace the existing tests and remove the new checks.

scripts/irods/test/test_iget.py
alanking: This test has been removed.

server/api/src/rsDataObjGet.cpp
alanking: The check for the buffer size range has been removed in favor of leaning on the configuration schema. Additionally, configuration validation tests have replaced the former tests.

alanking:
Squashed. New test passed, whole core test suite run is almost done with everything passing so far.

alanking:
Core tests passed

korydraughn:
Squash to taste.

alanking:
Done

korydraughn (approved):
Pound it.

alanking:
#'d, mergin

PR #8571, opened 2025-07-03
[#8373] DataObjGet: Enforce max size for single buffer (4-3-stable)

alanking:
Addresses #8373 

Cherry-picked from https://github.com/irods/irods/pull/8570

The test was not cherry-picked because `test_configuration` does not exist in 4-3-stable and there is no configuration check feature in the server for 4.3. I can add an equivalent test, but I figured this is probably good enough for 4.3.

That said, running core tests now.

### DIFF
--- a/schemas/configuration/v4/server_config.json.in
+++ b/schemas/configuration/v4/server_config.json.in
@@ -31,7 +31,11 @@
                         "cache_clearer_sleep_time_in_seconds": {"type": "integer"}
                     }
                 },
-                "maximum_size_for_single_buffer_in_megabytes": {"type": "integer"},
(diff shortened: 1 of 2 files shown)


korydraughn:
Witnessing a change in the schema for 4.3 had me wondering if the schema version should be bumped. https://docs.irods.org/5.0.1/developers/schema_versions/ also made me wonder.

However, these PRs are addressing a bug and do not result in breaking changes, so I think everything is fine.

Awaiting the test results.

alanking:
Sorry, forgot to report back on this. Core tests passed

korydraughn (approved):
Pound it.

alanking:
#'d, mergin

PR #8573, opened 2025-07-08
[8384] Add additional checksum return checks to rsync (4-3-stable)

FifthPotato:
This will check for `CHECK_VERIFICATION_RESULTS` from `rsDataObjChksum` and then parse the resultant JSON array. If any `error_code` is nonzero, it will still fail and return while logging a message. For the case listed in the ticket, the resultant JSON has a 0 error code-- it's informing the caller that the replica is skipped (due to it being stale). In that case, it will pass and allow the copy to progress. 

Taking suggestions for anything, as always. Of note is that the log message on 289: I don't think it would be wise to wrap `destChksumStr` in `[]` since `[]` is valid JSON, and it would make the log message ambiguous.

### DIFF
--- a/scripts/irods/test/test_irsync.py
+++ b/scripts/irods/test/test_irsync.py
@@ -682,3 +682,55 @@ class Test_iRsync(ResourceBase, unittest.TestCase):
         # Show no data object exists, proving the symlink file was ignored.
         logical_path = f'{self.user0.session_collection}/{symlink_dir_basename}/{symlink_file_basename}'
         self.assertFalse(lib.replica_exists(self.user0, logical_path, 0))
+
+    def test_irsync_succeeds_with_checksum_and_stale_replica__issue_8384(self):
+        # Create a file.
+        test_file_basename = 'irsync_issue_8384.txt'
+        test_file = os.path.join(self.admin.local_session_dir,test_file_basename)
+
+        with open(test_file, 'w') as f:
+            f.write('Random test data.')
+
+        # Set up hierarchy similar to the issue
+        replication_resc = 'repl_resc_8384'
+        child_resc = 'ufs1_8384'
+        other_child_resc = 'ufs2_8384'
+        source_coll = 'thesource_8384'
+        dest_coll = 'thetarget_8384'
+        child_resc_dir = os.path.join(self.admin.local_session_dir, 'ufs1vault_8384')
+        other_child_resc_dir = os.path.join(self.admin.local_session_dir, 'ufs2vault_8384')
+
+        try:
+            self.admin.assert_icommand(['imkdir', f'{self.admin.session_collection}/{source_coll}', f'{self.admin.session_collection}/{dest_coll}'])
+            self.admin.assert_icommand(['iadmin', 'mkresc', replication_resc, 'replication' ], 'STDOUT_SINGLELINE', 'replication')
+            self.admin.assert_icommand(['iadmin', 'mkresc', child_resc, 'unixfilesystem', f'{lib.get_hostname()}:{child_resc_dir}'], 'STDOUT_SINGLELINE', 'unixfilesystem')
+            self.admin.assert_icommand(['iadmin', 'mkresc', other_child_resc, 'unixfilesystem', f'{lib.get_hostname()}:{other_child_resc_dir}'], 'STDOUT_SINGLELINE', 'unixfilesystem')
+            self.admin.assert_icommand(['iadmin', 'addchildtoresc', replication_resc, child_resc])
+            self.admin.assert_icommand(['iadmin', 'addchildtoresc', replication_resc, other_child_resc])
+
+            self.admin.assert_icommand(['iput', test_file, f'{self.admin.session_collection}/{source_coll}'])
+            self.admin.assert_icommand(['iput', '-R', replication_resc, test_file, f'{self.admin.session_collection}/{dest_coll}'])
+            self.admin.assert_icommand(['ichksum', '-n', '0', f'{self.admin.session_collection}/{dest_coll}/{test_file_basename}'], 'STDOUT_SINGLELINE', test_file_basename)
+            self.admin.assert_icommand(['ichksum', '-n', '1', f'{self.admin.session_collection}/{dest_coll}/{test_file_basename}'], 'STDOUT_SINGLELINE', test_file_basename)
+
+            # Set one replica to stale
+            # rsDataObjChksum will return with CHECK_VERIFICATION_RESULTS if any replicas are skipped
+            # Previously, this was treated as a hard error, so irsync terminated early
+            # This should no longer happen
+            self.admin.assert_icommand(['iadmin', 'modrepl', 'logical_path', f'{self.admin.session_collection}/{dest_coll}/{test_file_basename}', 'replica_number', '1', 'DATA_REPL_STATUS', '0'])
+
+            # Should no longer error out due to CHECK_VERIFICATION_RESULTS
+            self.admin.assert_icommand(['irsync', '-K', '-v', '-r', '-R', replication_resc, f'i:{source_coll}', f'i:{dest_coll}' ], 'STDOUT_SINGLELINE', test_file_basename)
+
+            # Copy should have happened and set the replica to good
+            self.assertTrue(lib.get_replica_status(self.admin, test_file_basename, 1) == '1')
+
+        finally:
+            # Clean up
+            self.admin.run_icommand(['irm', '-rf', f'{self.admin.session_collection}/{source_coll}'])
+            self.admin.run_icommand(['irm', '-rf', f'{self.admin.session_collection}/{dest_coll}'])
+            self.admin.run_icommand(['iadmin', 'rmchildfromresc', replication_resc, child_resc])
+            self.admin.run_icommand(['iadmin', 'rmchildfromresc', replication_resc, other_child_resc])
+            self.admin.run_icommand(['iadmin', 'rmresc', child_resc])
+            self.admin.run_icommand(['iadmin', 'rmresc', other_child_resc])
+            self.admin.run_icommand(['iadmin', 'rmresc', replication_resc])
--- a/server/api/src/rsDataObjRsync.cpp
+++ b/server/api/src/rsDataObjRsync.cpp
@@ -11,9 +11,18 @@
 #include "irods/rsDataObjCopy.hpp"
 
 #include "irods/irods_resource_redirect.hpp"
+#include "irods/irods_logger.hpp"
 
 #define IRODS_FILESYSTEM_ENABLE_SERVER_SIDE_API
 #include "irods/filesystem.hpp"
+
+#include <nlohmann/json.hpp>
+
+namespace
+{
+    using json = nlohmann::json;
+    using log_api = irods::experimental::log::api;
+} //namespace
 
 int
 rsDataObjRsync( rsComm_t *rsComm, dataObjInp_t *dataObjInp,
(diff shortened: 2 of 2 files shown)


server/api/src/rsDataObjRsync.cpp
```diff
@@ -261,7 +267,41 @@ rsRsyncDataToData( rsComm_t *rsComm, dataObjInp_t *dataObjInp ) {
         /* use rsDataObjChksum because the path could be in a remote zone */
         status = rsDataObjChksum(rsComm, &dataObjCopyInp.destDataObjInp, &destChksumStr);
 
-        if (status < 0 && status != CAT_NO_ACCESS_PERMISSION && status != CAT_NO_ROWS_FOUND) {
+        if (CHECK_VERIFICATION_RESULTS == status) {
+            try {
+                const auto& verification_result = json::parse(destChksumStr);
```
korydraughn, line 275: Consider checking the `destChksumStr` pointer for null and whether `json::parse` handles null pointers.

server/api/src/rsDataObjRsync.cpp
```diff
@@ -261,7 +267,41 @@ rsRsyncDataToData( rsComm_t *rsComm, dataObjInp_t *dataObjInp ) {
...
+                for (const auto& entry : verification_result) {
+                    if (entry.at("error_code") != 0) {
```
korydraughn: Avoid implicit casts when possible. Use `.get<int>()`.

server/api/src/rsDataObjRsync.cpp
```diff
@@ -261,7 +267,41 @@ rsRsyncDataToData( rsComm_t *rsComm, dataObjInp_t *dataObjInp ) {
...
+                        // "actual" error happened: log it and fail
+                        log_api::error("{}: error in rsDataObjChksum, message = [{}]; error_code = [{}]",
+                                       __func__,
+                                       entry.at("message").get<std::string>(),
```
korydraughn: Can use `.get_ref<const std::string&>()` to avoid a possible memory allocation.

server/api/src/rsDataObjRsync.cpp
```diff
@@ -261,7 +267,41 @@ rsRsyncDataToData( rsComm_t *rsComm, dataObjInp_t *dataObjInp ) {
...
+                                       entry.at("error_code").get<int>());
```
korydraughn: Can avoid another lookup of `error_code` by initializing a variable in the if-stmt.
```cpp
if (const auto ec = entry.at("error_code").get<int>(); ec != 0) {
```

server/api/src/rsDataObjRsync.cpp
```diff
@@ -261,7 +267,41 @@ rsRsyncDataToData( rsComm_t *rsComm, dataObjInp_t *dataObjInp ) {
...
+                        // NOLINTNEXTLINE(cppcoreguidelines-no-malloc, cppcoreguidelines-owning-memory)
+                        free(srcChksumStr);
+                        // NOLINTNEXTLINE(cppcoreguidelines-no-malloc, cppcoreguidelines-owning-memory)
+                        free(destChksumStr);
+                        clearKeyVal(&dataObjCopyInp.srcDataObjInp.condInput);
+                        clearKeyVal(&dataObjCopyInp.destDataObjInp.condInput);
```
korydraughn, line 288: Are there other members of `dataObjCopyInp` which can leak memory?

server/api/src/rsDataObjRsync.cpp
```diff
@@ -11,10 +11,16 @@
...
 #include "irods/irods_resource_redirect.hpp"
+#include "irods/irods_logger.hpp"
 
 #define IRODS_FILESYSTEM_ENABLE_SERVER_SIDE_API
 #include "irods/filesystem.hpp"
 
+#include <nlohmann/json.hpp>
+
+using json = nlohmann::json;
+using log_api = irods::experimental::log::api;
```
korydraughn: Let's put these in an anonymous namespace so they are private within the binary.

server/api/src/rsDataObjRsync.cpp
FifthPotato: Per https://json.nlohmann.me/api/basic_json/parse/#exceptions "Throws [parse_error.101](https://json.nlohmann.me/home/exceptions/#jsonexceptionparse_error101) in case of an unexpected token, or empty input like a null FILE* or char* pointer." That should get caught by the catch statement; should I do it explicitly anyways?

korydraughn: No. I just wanted to make sure we covered that situation.

server/api/src/rsDataObjRsync.cpp
```diff
@@ -261,7 +267,41 @@ rsRsyncDataToData( rsComm_t *rsComm, dataObjInp_t *dataObjInp ) {
...
+                        clearKeyVal(&dataObjCopyInp.srcDataObjInp.condInput);
+                        clearKeyVal(&dataObjCopyInp.destDataObjInp.condInput);
+                        return status;
+                    }
+                }
+            }
+            catch (const json::exception& ex) {
+                log_api::error("{}: json exception from rsDataObjChksum return, destChksumStr = \"{}\"; what = [{}]",
+                               __func__,
+                               destChksumStr,
```
korydraughn: I don't think nullptr is understood by the logger or fmtlib. Let's verify and update this code accordingly.

alanking (commented):
The description in OP wouldn't be a bad thing to include in the commit message body, IMO

server/api/src/rsDataObjRsync.cpp
FifthPotato: I'm not 100% sure-- the cleanup code was copied from the original source from the failing branch, so I assumed they were correct. Should I go hunting for potential leaks?

FifthPotato:
> The description in OP wouldn't be a bad thing to include in the commit message body, IMO

Noted. I'll remember to add that when squash happens.

scripts/irods/test/test_irsync.py
```diff
@@ -682,3 +682,37 @@ def test_irsync_ignore_symlinks_option_is_an_alias_of_the_link_option__issue_753
...
+        source_coll = 'thesource_8384'
+        dest_coll = 'thetarget_8384'
+        child_resc_dir = os.path.join(self.admin.local_session_dir, 'ufs1vault_8384')
+        other_child_resc_dir = os.path.join(self.admin.local_session_dir, 'ufs2vault_8384')
+        self.admin.assert_icommand(['imkdir', f'{self.admin.session_collection}/{source_coll}', f'{self.admin.session_collection}/{dest_coll}'])
+        self.admin.assert_icommand(['iadmin', 'mkresc', replication_resc, 'replication' ], 'STDOUT_SINGLELINE', 'replication')
+        self.admin.assert_icommand(['iadmin', 'mkresc', child_resc, 'unixfilesystem', f'{lib.get_hostname()}:{child_resc_dir}'], 'STDOUT_SINGLELINE', 'unixfilesystem')
+        self.admin.assert_icommand(['iadmin', 'mkresc', other_child_resc, 'unixfilesystem', f'{lib.get_hostname()}:{other_child_resc_dir}'], 'STDOUT_SINGLELINE', 'unixfilesystem')
+        self.admin.assert_icommand(['iadmin', 'addchildtoresc', replication_resc, child_resc])
+        self.admin.assert_icommand(['iadmin', 'addchildtoresc', replication_resc, other_child_resc])
```
alanking: We will need to remove these resources at the conclusion of the test as well to ensure idempotency.

scripts/irods/test/test_irsync.py
```diff
@@ -682,3 +682,37 @@ def test_irsync_ignore_symlinks_option_is_an_alias_of_the_link_option__issue_753
...
+        self.admin.assert_icommand(['iput', test_file, f'{self.admin.session_collection}/{source_coll}'])
+        self.admin.assert_icommand(['iput', '-R', replication_resc, test_file, f'{self.admin.session_collection}/{dest_coll}'])
+        self.admin.assert_icommand(['ichksum', '-n', '0', f'{self.admin.session_collection}/{dest_coll}/{test_file_basename}'], 'STDOUT_SINGLELINE', test_file_basename)
+        self.admin.assert_icommand(['ichksum', '-n', '1', f'{self.admin.session_collection}/{dest_coll}/{test_file_basename}'], 'STDOUT_SINGLELINE', test_file_basename)
+
+        # Set one replica to stale
+        self.admin.assert_icommand(['iadmin', 'modrepl', 'logical_path', f'{self.admin.session_collection}/{dest_coll}/{test_file_basename}', 'replica_number', '1', 'DATA_REPL_STATUS', '0'])
+
+        # Should not error out
+        self.admin.assert_icommand(['irsync', '-K', '-v', '-r', '-R', replication_resc, f'i:{source_coll}', f'i:{dest_coll}' ], 'STDOUT_SINGLELINE', test_file_basename)
```
alanking: What do we expect to have happened as a result of this call? Is the replica supposed to be updated, or not? In either case, we should have some assertions around the expected results.

server/api/src/rsDataObjRsync.cpp
korydraughn: Not right now. Looking at the code around this change, I sense there are many places which can leak if an exception is thrown.

Fixing those potential problems is out of scope for your PR. Let's open an issue to improve memory management within this function. One way to address the problem is to use at_scope_exit to guarantee memory is freed when an exception is thrown.

scripts/irods/test/test_irsync.py
FifthPotato: As it is now, it is meant to be updated. I think it's possible to argue that it should calculate a checksum on the source and compare that to the target before trying (and then stop and refuse to re-copy if the checksums match.) In fact, that is what happens if the source already has a checksum. I discussed some of irsync's weird behavior at length with @trel and @korydraughn last week, and I think we said this was okay as is, but if you have any thoughts on it, I'd like to hear them!

alanking: Okay, I think we should at least confirm that the stale replica has been updated to good status given that is the only thing we can measure here.

scripts/irods/test/test_irsync.py
```diff
@@ -682,3 +682,49 @@ def test_irsync_ignore_symlinks_option_is_an_alias_of_the_link_option__issue_753
...
+
+        # Copy should have happened and set the replica to good
+        self.assertTrue(lib.get_replica_status(self.admin, test_file_basename, 1) == '1')
+
+        # Clean up
```
trel: cleanup is usually in a finally block, i think - so they run even if assertions or code above fails.

scripts/irods/test/test_irsync.py
```diff
@@ -699,32 +699,35 @@ def test_irsync_succeeds_with_checksum_and_stale_replica__issue_8384(self):
...
+            # Should not error out
+            self.admin.assert_icommand(['irsync', '-K', '-v', '-r', '-R', replication_resc, f'i:{source_coll}', f'i:{dest_coll}' ], 'STDOUT_SINGLELINE', test_file_basename)
+
+            # Copy should have happened and set the replica to good
+            self.assertTrue(lib.get_replica_status(self.admin, test_file_basename, 1) == '1')
+
+        finally:
+            # Clean up
+            self.admin.run_icommand(['irm', '-rf', f'{self.admin.session_collection}/{source_coll}'])
+            self.admin.run_icommand(['irm', '-rf', f'{self.admin.session_collection}/{dest_coll}'])
```
trel, line 731: these two may be unnecessary as the session collection is provided by and removed by the testing framework already?   please confirm/test my assertion.

alanking: I believe they need to be removed here in order to remove the resources.

trel: ah, quite right.  resolving!

scripts/irods/test/test_irsync.py
```diff
@@ -682,3 +682,52 @@ def test_irsync_ignore_symlinks_option_is_an_alias_of_the_link_option__issue_753
...
+            self.admin.assert_icommand(['iadmin', 'addchildtoresc', replication_resc, child_resc])
+            self.admin.assert_icommand(['iadmin', 'addchildtoresc', replication_resc, other_child_resc])
+
+            self.admin.assert_icommand(['iput', test_file, f'{self.admin.session_collection}/{source_coll}'])
+            self.admin.assert_icommand(['iput', '-R', replication_resc, test_file, f'{self.admin.session_collection}/{dest_coll}'])
+            self.admin.assert_icommand(['ichksum', '-n', '0', f'{self.admin.session_collection}/{dest_coll}/{test_file_basename}'], 'STDOUT_SINGLELINE', test_file_basename)
+            self.admin.assert_icommand(['ichksum', '-n', '1', f'{self.admin.session_collection}/{dest_coll}/{test_file_basename}'], 'STDOUT_SINGLELINE', test_file_basename)
+
+            # Set one replica to stale
+            self.admin.assert_icommand(['iadmin', 'modrepl', 'logical_path', f'{self.admin.session_collection}/{dest_coll}/{test_file_basename}', 'replica_number', '1', 'DATA_REPL_STATUS', '0'])
```
korydraughn: I feel more wording explaining "why" you're setting one replica to stale is needed. Will be very helpful 6 months from now.

scripts/irods/test/test_irsync.py
```diff
@@ -682,3 +682,52 @@ def test_irsync_ignore_symlinks_option_is_an_alias_of_the_link_option__issue_753
...
+
+            # Should not error out
```
korydraughn: Same thing here. Feels important to provide more details to what this is about.

korydraughn (commented):
Looks good to me. Squash to taste.

FifthPotato:
Squished. Ran clang-format too.

FifthPotato:
Ah, one moment, core tests pass, but getting a single failure on `irods_parallel_transfer_engine` for unit tests. Should I look deeper into that?

korydraughn:
You can if you want, but don't spend too much time on it.

Given it's not in use or a blocker for this PR, you're free to ignore it.

korydraughn:
Don't forget to update the commit message. See https://github.com/irods/irods/pull/8573#pullrequestreview-3016270842.

FifthPotato:
Updated the commit message. I'll ignore that failure for now.

korydraughn:
Agreed.

@FifthPotato Please break up the commit message so it fits into smaller panes.

FifthPotato:
Added some line breaks at or about ~80 chars. Is that okay?

korydraughn:
Yes. There's one line which starts with ` logging`. Please remove the leading space from that line.
```
of it. Now, if any error_code is nonzero, it will still fail and return while
 logging a message (i.e. all 0 error codes indicates everything is fine). For  <<<
the case listed in the ticket, the resultant JSON has a 0 error code: it's
```
```
of it. Now, if any error_code is nonzero, it will still fail and return while
logging a message (i.e. all 0 error codes indicates everything is fine). For   <<<
the case listed in the ticket, the resultant JSON has a 0 error code: it's
```

FifthPotato:
Done.

alanking (approved):
\# it

FifthPotato:
Octothorpe'd

PR #8584, opened 2025-07-18
[8588] python controller errors out early if it detects a terminated process is a zombie (main)

trel:
looking for better warning message / phrasing / instructions for the admin.

something like...

'you need more --init'
or
'go read about init processes'

### DIFF
--- a/scripts/irods/controller.py
+++ b/scripts/irods/controller.py
@@ -223,8 +223,16 @@ class IrodsController(object):
         l = logging.getLogger(__name__)
         for attempt in range(retry_count):
             l.debug('Waiting for iRODS server to shut down. Attempt #%s', attempt)
-            if self.get_server_pid() is None:
+            server_pid = self.get_server_pid()
+            if server_pid is None:
                 return
+
+            # checking for a zombie process
+            proc = psutil.Process(server_pid)
+            if proc.status() == psutil.STATUS_ZOMBIE:
+                l.warning('Terminating the iRODS server referenced by the iRODS PID file resulted in a zombie process. Consider running an init-like process as PID 1.')
+                break
+
             time.sleep(1)
         raise IrodsError('iRODS server failed to shut down.')
 


scripts/irods/controller.py
```diff
@@ -223,8 +223,16 @@ def wait_for_server_to_stop(self, retry_count=100):
...
             l.debug('Waiting for iRODS server to shut down. Attempt #%s', attempt)
-            if self.get_server_pid() is None:
+            server_pid = self.get_server_pid()
+            if server_pid is None:
                 return
+
+            # checking for zombieness
+            proc = psutil.Process(server_pid)
+            if proc.status() == psutil.STATUS_ZOMBIE:
+                l.warning('Server process was terminated, but is now a zombie.  Your environment is not reaping them.')
```
alanking: Sometimes warning messages pose a question to provide help on what might be causing a problem. Maybe something like...

"Terminating server process resulted in a zombie process. Is it running as the init process?"

korydraughn: Here's my attempt:
```
Zombie process referenced by iRODS PID file. Does the environment have
an init-like process running as PID 1?
```
```
Zombie process referenced by iRODS PID file. Your environment is not
reaping zombie processes.
```
```
Zombie process referenced by iRODS PID file. Consider running an init-like
process as PID 1.
```
```
Zombie process referenced by iRODS PID file. Your environment does not
appear to reap zombie processes. Please verify PID 1 supports init-like
functionality.
```

trel: mixing the paint a bit... how does this sound...

```
Terminating server process resulted in a zombie process referenced by the iRODS PID file.  Consider running an init-like process as PID 1.
```

alanking: Any of the suggestions other than mine seem fine to me.

trel: ha.

korydraughn: No preference.

My proposals are mostly about pointing out that the target PID comes from the PID file (in /var/run/irods).

Notice the script doesn't know when the admin launched the server with a custom PID file. I don't think this is an issue since the controller.py logic isn't in the main path for shutdown. The common/expected way to shutdown the server is through the `kill` command or the service manager (if configured for that).

trel: updated the words...

```
Terminating the iRODS server referenced by the iRODS PID file resulted in a zombie process.  Consider running an init-like process as PID 1. 
```

korydraughn: Message looks good to me.

Resolving.

korydraughn (commented):
Squash it.

trel:
squashed, needs an issue number...

trel:
now 8588 exists, commit message updated.

trel:
#'d

korydraughn:
Merging.

Issue #8363, opened 2025-04-08
Intermittent unit test failures for irods_parallel_transfer_engine on Ubuntu24


FifthPotato:
Getting strange unit test failures that crop up and go away seemingly randomly. Testing was being done on ubuntu24.
Relevant log xml follows.

```
    <testcase classname="irods_parallel_transfer_engine.global" name="parallel transfer engine/with restart" time="0.432" status="run">
[...]

PR #8592, opened 2025-07-23
[7774] irsync -l no longer creates target collection (4-3-stable)

FifthPotato:
My first stab at option 3 as mentioned in the original issue.
Seems to work when run in a docker container and with the original test case.
Taking suggestions on naming, style, or anything, as always.

### DIFF
--- a/lib/core/src/miscUtil.cpp
+++ b/lib/core/src/miscUtil.cpp
@@ -2335,16 +2335,8 @@ matchPathname( pathnamePatterns_t *pp, char *name, char *dirname ) {
     return 0;
 }
 
-/* resolveRodsTarget - based on srcPath and destPath, fill in targPath.
- * oprType -
- *      MOVE_OPR - do not create the target coll or dir because rename will
- *        take care of it.
- *      RSYNC_OPR - uses the destPath and the targPath if the src is a
- *        collection
- *      All other oprType will be treated as normal.
- */
-int
-resolveRodsTarget( rcComm_t *conn, rodsPathInp_t *rodsPathInp, int oprType ) {
+static int resolveRodsTargetImpl(rcComm_t* conn, rodsPathInp_t* rodsPathInp, int oprType, bool dryRun)
+{
     rodsPath_t *srcPath, *destPath;
     char srcElement[MAX_NAME_LEN], destElement[MAX_NAME_LEN];
     int status;
@@ -2472,14 +2464,15 @@ resolveRodsTarget( rcComm_t *conn, rodsPathInp_t *rodsPathInp, int oprType ) {
                 if (destPath->objState != EXIST_ST && destPath->objType <= COLL_OBJ_T &&
                     oprType != MOVE_OPR && rodsPathInp->numSrc != 1)
                 {
-                    // mkColl changes the values in destPath, so repopulate the structure
-                    status = mkColl( conn, destPath->outPath );
-                    if ( status < 0 ) {
-                        rodsLog( LOG_ERROR,
-                                 "resolveRodsTarget: Could not create collection %s", destPath->outPath );
-                        return status;
+                    if (!dryRun) {
+                        // mkColl changes the values in destPath, so repopulate the structure
+                        status = mkColl(conn, destPath->outPath);
+                        if (status < 0) {
+                            rodsLog(LOG_ERROR, "resolveRodsTarget: Could not create collection %s", destPath->outPath);
+                            return status;
+                        }
+                        getRodsObjType(conn, destPath);
                     }
-                    getRodsObjType( conn, destPath );
                 }
                 if ( ( destPath->objType == COLL_OBJ_T ||
                        destPath->objType == LOCAL_DIR_T ) &&
@@ -2509,8 +2502,7 @@ resolveRodsTarget( rcComm_t *conn, rodsPathInp_t *rodsPathInp, int oprType ) {
                                     // exists in this block), make sure targPath is made if it is a collection.
                                     if (targPath->objType == COLL_OBJ_T)
                                     {
-                                        if ((status = mkColl( conn, targPath->outPath )) == 0)
-                                        {
+                                        if (!dryRun && (status = mkColl(conn, targPath->outPath)) == 0) {
                                             targPath->objState = EXIST_ST;
                                         }
                                     }
@@ -2520,11 +2512,12 @@ resolveRodsTarget( rcComm_t *conn, rodsPathInp_t *rodsPathInp, int oprType ) {
                                 }
                             }
                             else {
-    #ifdef windows_platform
-                                status = iRODSNt_mkdir( targPath->outPath, 0750 );
-    #else
-                                status = mkdir( targPath->outPath, 0750 );
-    #endif
+                                if (!dryRun) {
+                                    status = mkdir(targPath->outPath, 0750);
+                                }
+                                else {
+                                    status = 0;
+                                }
                                 if ( status < 0 && errno == EEXIST ) {
                                     status = 0;
                                 }
@@ -2545,7 +2538,7 @@ resolveRodsTarget( rcComm_t *conn, rodsPathInp_t *rodsPathInp, int oprType ) {
                 else {
                     /* dest coll does not exist */
                     if ( destPath->objType <= COLL_OBJ_T ) {
-                        if ( oprType != MOVE_OPR ) {
+                        if (!dryRun && oprType != MOVE_OPR) {
                             /* rename does not need to mkColl */
                             status = mkColl( conn, destPath->outPath );
                         }
@@ -2554,13 +2547,14 @@ resolveRodsTarget( rcComm_t *conn, rodsPathInp_t *rodsPathInp, int oprType ) {
                         }
                     }
                     else {
-                        /* use destPath. targPath->outPath not defined.
-                         *  status = mkdir (targPath->outPath, 0750); */
-    #ifdef windows_platform
-                        status = iRODSNt_mkdir( destPath->outPath, 0750 );
-    #else
-                        status = mkdir( destPath->outPath, 0750 );
-    #endif
+                        if (!dryRun) {
+                            /* use destPath. targPath->outPath not defined.
+                             *  status = mkdir (targPath->outPath, 0750); */
+                            status = mkdir(destPath->outPath, 0750);
+                        }
+                        else {
+                            status = 0;
+                        }
                     }
                     if ( status < 0 ) {
                         return status;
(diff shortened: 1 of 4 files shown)


lib/core/src/miscUtil.cpp
```diff
@@ -2335,16 +2335,8 @@ matchPathname( pathnamePatterns_t *pp, char *name, char *dirname ) {
...
- * oprType -
- *      MOVE_OPR - do not create the target coll or dir because rename will
- *        take care of it.
- *      RSYNC_OPR - uses the destPath and the targPath if the src is a
- *        collection
- *      All other oprType will be treated as normal.
- */
 int
-resolveRodsTarget( rcComm_t *conn, rodsPathInp_t *rodsPathInp, int oprType ) {
+resolveRodsTargetImpl( rcComm_t *conn, rodsPathInp_t *rodsPathInp, int oprType, bool dryRun ) {
```
alanking: Can/Should we make this static or move it into an anoymous namespace? Just basing this on the principle of not exposing things until we have to.

Alternatively, could we expose this in the client library on purpose and have the existing `resolveRodsTarget` call it? The `bool` would need to change for the C interface, but then we would have fewer functions floating around, hidden or otherwise.

alanking (commented):
New test looks good

lib/core/src/miscUtil.cpp
```diff
@@ -2595,3 +2590,24 @@ resolveRodsTarget( rcComm_t *conn, rodsPathInp_t *rodsPathInp, int oprType ) {
...
+    return resolveRodsTargetImpl( conn, rodsPathInp, oprType, false );
+}
+
+/* resolveRodsTarget - based on srcPath and destPath, fill in targPath.
+ * As above, but will not create collections.
+ */
+int
+resolveRodsTargetDryRun ( rcComm_t *conn, rodsPathInp_t *rodsPathInp, int oprType ) {
+    return resolveRodsTargetImpl( conn, rodsPathInp, oprType, true );
+}
```
korydraughn: - Move the documentation to the appropriate header files
- Update documentation to follow style seen in newer header files
  - e.g. https://github.com/irods/irods/blob/main/lib/api/include/irods/check_auth_credentials.h

lib/core/src/miscUtil.cpp
korydraughn: Yes please.

Let's make this `static` or wrap it in an anonymous namespace as suggested.

scripts/irods/test/test_irsync.py
```diff
@@ -734,3 +734,20 @@ def test_irsync_succeeds_with_checksum_and_stale_replica__issue_8384(self):
...
+
+        self.user0.assert_icommand(['imkdir', f'{self.user0.session_collection}/{source_coll}'])
+
+        # Make some dummy data objects
+        self.user0.assert_icommand(['itouch', f'{self.user0.session_collection}/{source_coll}/one'])
+        self.user0.assert_icommand(['itouch', f'{self.user0.session_collection}/{source_coll}/two'])
+        self.user0.assert_icommand(['itouch', f'{self.user0.session_collection}/{source_coll}/three'])
+
+        # Perform the irsync
+        self.admin.assert_icommand(['irsync', '-rl', f'i:{self.user0.session_collection}/{source_coll}', f'i:{self.user0.session_collection}/{dest_coll}' ], 'STDOUT', ['one', 'two', 'three'])
```
korydraughn: Does the expected output matter in this test or is it just to keep the assertion from failing?

FifthPotato: A little bit of both? It's more of a just "everything working as expected" assertion since you'd expect all 3 data objects to show up on the `irsync -l` listing. It's not exactly what this test is testing, but I don't think it hurts.

lib/core/src/miscUtil.cpp
FifthPotato: Updated. Is it detailed enough?

FifthPotato:
Also noticed that `irsync -rl` will create local directories, so I updated the code to also prevent that from happening. I think that's what we want?
If so, I'll add a test case for that too.

korydraughn:
Yes, good catch. Please add a test for that if you haven't already.

scripts/irods/test/test_irsync.py
korydraughn: Works for me.

lib/core/src/miscUtil.cpp
```diff
@@ -2520,11 +2515,16 @@ resolveRodsTarget( rcComm_t *conn, rodsPathInp_t *rodsPathInp, int oprType ) {
                                 }
                             }
                             else {
+                                if(!dryRun) {
     #ifdef windows_platform
-                                status = iRODSNt_mkdir( targPath->outPath, 0750 );
+                                    status = iRODSNt_mkdir( targPath->outPath, 0750 );
     #else
```
korydraughn: You can remove blocks which are Windows specific.

I spotted another instance of this in the PR too.

lib/core/src/miscUtil.cpp
```diff
@@ -2595,3 +2601,16 @@ resolveRodsTarget( rcComm_t *conn, rodsPathInp_t *rodsPathInp, int oprType ) {
...
 }
+
+int
+resolveRodsTarget ( rcComm_t *conn, rodsPathInp_t *rodsPathInp, int oprType ) {
+    return resolveRodsTargetImpl( conn, rodsPathInp, oprType, false );
+}
+
+/* resolveRodsTarget - based on srcPath and destPath, fill in targPath.
+ * As above, but will not create collections.
+ */
```
korydraughn: Forgot to remove this?

lib/core/src/miscUtil.cpp
korydraughn: You did what was asked, so it's fine. I'd like to see the documentation fleshed out more, but that's not a requirement of the issue. If you happen to know anything worth adding, please do so. Otherwise, you can mark this as resolved.

scripts/irods/test/test_irsync.py
```diff
@@ -747,7 +749,11 @@ def test_irsync_does_not_create_collections_on_dry_run_issue_7774(self):
         self.user0.assert_icommand(['itouch', f'{self.user0.session_collection}/{source_coll}/three'])
 
         # Perform the irsync
-        self.admin.assert_icommand(['irsync', '-rl', f'i:{self.user0.session_collection}/{source_coll}', f'i:{self.user0.session_collection}/{dest_coll}' ], 'STDOUT', ['one', 'two', 'three'])
+        self.user0.assert_icommand(['irsync', '-rl', f'i:{self.user0.session_collection}/{source_coll}', f'i:{self.user0.session_collection}/{dest_coll}' ], 'STDOUT', ['one', 'two', 'three'])
```
korydraughn: Consider constructing the paths one time and reusing them. It shortens the line length and makes it easier to notice the same paths are being used in multiple assertions.

scripts/irods/test/test_irsync.py
```diff
@@ -747,7 +749,11 @@ def test_irsync_does_not_create_collections_on_dry_run_issue_7774(self):
...
 
         # Assure that the target collection does not exist
         self.user0.assert_icommand(['ils', '-L', dest_coll], 'STDERR_SINGLELINE', 'does not exist')
+
+        # Other case-- ensure that irods-to-local irsync do not create the target dir
+        self.user0.assert_icommand(['irsync', '-rl', f'i:{self.user0.session_collection}/{source_coll}', os.path.join(self.testing_tmp_dir, dest_dir_should_not_exist) ], 'STDOUT', ['one', 'two', 'three'])
+        self.assertFalse(os.path.exists(os.path.join(self.testing_tmp_dir, dest_dir_should_not_exist)))
```
korydraughn: Consider constructing the paths one time and reusing them. It shortens the line length and makes it easier to notice the same paths are being used in multiple assertions.

scripts/irods/test/test_irsync.py
```diff
@@ -738,22 +738,25 @@ def test_irsync_succeeds_with_checksum_and_stale_replica__issue_8384(self):
...
-        self.user0.assert_icommand(['itouch', f'{self.user0.session_collection}/{source_coll}/one'])
-        self.user0.assert_icommand(['itouch', f'{self.user0.session_collection}/{source_coll}/two'])
-        self.user0.assert_icommand(['itouch', f'{self.user0.session_collection}/{source_coll}/three'])
+        self.user0.assert_icommand(['itouch', f'{source_coll_path}/one'])
+        self.user0.assert_icommand(['itouch', f'{source_coll_path}/two'])
+        self.user0.assert_icommand(['itouch', f'{source_coll_path}/three'])
 
         # Perform the irsync
-        self.user0.assert_icommand(['irsync', '-rl', f'i:{self.user0.session_collection}/{source_coll}', f'i:{self.user0.session_collection}/{dest_coll}' ], 'STDOUT', ['one', 'two', 'three'])
+        self.user0.assert_icommand(['irsync', '-rl', f'i:{source_coll_path}', f'i:{dest_coll_path}' ], 'STDOUT', ['one', 'two', 'three'])
```
alanking:
```suggestion
        self.user0.assert_icommand(['irsync', '-rl', f'i:{source_coll_path}', f'i:{dest_coll_path}'], 'STDOUT', ['one', 'two', 'three'])
```

scripts/irods/test/test_irsync.py
```diff
@@ -738,22 +738,25 @@ def test_irsync_succeeds_with_checksum_and_stale_replica__issue_8384(self):
...
 
         # Assure that the target collection does not exist
-        self.user0.assert_icommand(['ils', '-L', dest_coll], 'STDERR_SINGLELINE', 'does not exist')
+        self.user0.assert_icommand(['ils', '-L', dest_coll_path], 'STDERR_SINGLELINE', 'does not exist')
 
         # Other case-- ensure that irods-to-local irsync do not create the target dir
-        self.user0.assert_icommand(['irsync', '-rl', f'i:{self.user0.session_collection}/{source_coll}', os.path.join(self.testing_tmp_dir, dest_dir_should_not_exist) ], 'STDOUT', ['one', 'two', 'three'])
-        self.assertFalse(os.path.exists(os.path.join(self.testing_tmp_dir, dest_dir_should_not_exist)))
+        self.user0.assert_icommand(['irsync', '-rl', f'i:{source_coll_path}', nonexistent_dir_path ], 'STDOUT', ['one', 'two', 'three'])
```
alanking:
```suggestion
        self.user0.assert_icommand(['irsync', '-rl', f'i:{source_coll_path}', nonexistent_dir_path], 'STDOUT', ['one', 'two', 'three'])
```

alanking (commented):
Just nitpicks from me here - seems good. Don't forget to clang-format

korydraughn:
What's the status of this? I see unresolved review comments.

Has this been run through the test suite?

FifthPotato:
> What's the status of this? I see unresolved review comments.
> 
> Has this been run through the test suite?

Running the test suite right now.

FifthPotato:
Passed all core and unit except for `irods_parallel_transfer_engine`, like my last PR. Can we ignore that for this PR as well?

korydraughn:
Yes. You can completely ignore that test.

Run clang-format and squash it.

alanking:
> Passed all core and unit except for `irods_parallel_transfer_engine`, like my last PR.

Linking this issue for another data point... https://github.com/irods/irods/issues/8363

FifthPotato:
Formatted, squished.

FifthPotato:
Octothorpe applied

korydraughn:
Please cherry-pick to the main branch.

Issue #8596, opened 2025-07-24
Support Debian 13


SwooshyCueb:
- [x] `main`
- [x] `5-0-stable`
- [x] other projects

Debian 13 (codenamed "trixie") is currently [planned](https://lists.debian.org/debian-devel-announce/2025/07/msg00003.html) to be released on August 9th. Let's add support once it's out.

Issue #8595, opened 2025-07-24
Bump version numbers for main and 5-0-stable branches


SwooshyCueb:
- [x] `main`
- [x] `5-0-stable`
- [x] other repositories

We have now split `5-0-stable` from `main` in preparation for a 5.0.2 patch release. Since we will no longer be tagging patch releases from `main`, the version number in `main` should be bumped to 5.0.99.
Additionally, due to some of the changes required for EL10 support, it makes sense to go ahead and bump `5-0-stable` to 5.0.2.

Other projects will need to be updated as well.

Issue #8581, opened 2025-07-16
CMake variable for jsoncons library not defined?


korydraughn:
## Bug Report

The development package for iRODS 5.0.1 does not appear to define the cmake variable, `IRODS_EXTERNALS_FULLPATH_JSONCONS`.

Boost and nanodbc have variables defined here.

https://github.com/irods/irods/blob/2d7d0e4d2251dd3a052dfb1f00c410afb7aa834d/cmake/IRODSConfig.cmake.not_yet_installed.in#L19-L24

I'd expect jsoncons to also have an entry considering the top-level cmakelists file mentions boost, nanodbc, and jsoncons.

SwooshyCueb:
jsoncons isn't used by anything else yet so it's not a massive concern, but yes, we should be defining it here

trel:
HTTP API now does!  https://github.com/irods/irods_client_http_api/pull/425

PR #8579, opened 2025-07-15
[#6008,#8099,#8454,#8455,#8581,#8595,#8596,#8611] Update for EL10 and Debian 13 (main)

SwooshyCueb:
Addresses #8455 for main
In service of #6008
In service of #8099
In service of #8454
In service of #8581
In service of #8595
In service of #8596
In service of #8611

Companion PRs:

- Should be merged **before** this PR:
    - irods/externals#280
- Should be merged **with** this PR:
    - #8602
    - irods/irods_client_icommands#580
    - irods/irods_client_icommands#583
- Can be merged at any point once EL10 and Debian 13 externals are present in both the `packages.irods.org` and `core-dev.irods.org` repositories:
    - irods/irods_development_environment#166
    - irods/irods_testing_environment#271
- Should be merged **before** or **with** this PR:
    - irods/irods_capability_indexing#174
    - irods/irods_capability_storage_tiering#351
    - irods/irods_client_http_api#422
        - **NOTE**: adds EL10 GHA workflow that will fail until 5.0.2 is published.
    - irods/irods_client_s3_api#143
    - irods/irods_microservice_plugins_curl#78
    - irods/irods_resource_plugin_s3#2264
    - irods/irods_rule_engine_plugin_audit_amqp#173
    - irods/irods_rule_engine_plugin_logical_quotas#133
    - irods/irods_rule_engine_plugin_logical_quotas#134
    - irods/irods_rule_engine_plugin_metadata_guard#98
    - irods/irods_rule_engine_plugin_python#248
    - irods/irods_python_ci_utilities#23

PR #8602, opened 2025-07-29
[#6008,#8099,#8454,#8455,#8581,#8595,#8596,#8611] Update for EL10 and Debian 13 (5-0-stable)

SwooshyCueb:
Addresses #8455 for 5-0-stable
In service of #6008
In service of #8099
In service of #8454
In service of #8581
In service of #8595
In service of #8596
In service of #8611

Companion to (and should be merged with) irods/irods#8579

### DIFF
--- a/.github/workflows/build-irods-debian.yml
+++ b/.github/workflows/build-irods-debian.yml
@@ -7,7 +7,7 @@ jobs:
     container: ${{ matrix.container }}
     strategy:
       matrix:
-        container: ['debian:12', 'debian:trixie']
+        container: ['debian:12', 'debian:13']
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v2


lib/core/include/irods/irods_query.hpp
```diff
@@ -18,10 +18,31 @@
...
+    constexpr auto format(const irods::query_type& e, format_context& ctx) const
+    {
+        return fmt::formatter<std::underlying_type_t<irods::query_type>>::format(
+            static_cast<std::underlying_type_t<irods::query_type>>(e), ctx);
+    }
+};
+
+namespace irods
+{
+
```
korydraughn: Remove empty line.

korydraughn (commented):
Just one empty line to remove. Once done, squash to taste.

lib/core/include/irods/irods_query.hpp
SwooshyCueb: Done

SwooshyCueb:
Squashed.

pre-squash commit: 2256ed9da550c3c2e141fca0e223c591e01e3aac

lib/core/include/irods/query_builder.hpp
```diff
@@ -17,12 +17,17 @@ namespace irods::experimental
...
     {
     public:
-        auto type(query_type _v) noexcept -> query_builder&
+        auto type(irods::query_type _v) noexcept -> query_builder&
         {
             type_ = _v;
             return *this;
         }
 
+        auto type(query_type _v) noexcept -> query_builder&
```
alanking: For my own understanding... We aren't deprecating `irods::experimental::query_type` here because that would only be allowed to happen in a minor version?

korydraughn: Deprecating anything in the public API requires a bump to the minor version at minimum.

alanking: Got it. And... adding this new function... does not require bumping the minor version?

korydraughn: Hmm, I think it does. Checking the semver wording now.

korydraughn: From https://semver.org/#spec-item-6 and https://semver.org/#spec-item-7:

> 6. Patch version Z (x.y.Z | x > 0) MUST be incremented if only backward compatible bug fixes are introduced. A bug fix is defined as an internal change that fixes incorrect behavior.
> 
> 7. Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backward compatible functionality is introduced to the public API. It MUST be incremented if any public API functionality is marked as deprecated. It MAY be incremented if substantial new functionality or improvements are introduced within the private code. It MAY include patch level changes. Patch version MUST be reset to 0 when minor version is incremented.

This change doesn't fit (6) because there's no bug to fix.

Is the change considered to be _new, backward compatible functionality_? I don't think it introduces new functionality. It is in support of a new platform and doesn't introduce new features to the iRODS server. I think one can classify this as a unique situation because it can fall in the PATCH space or MINOR space.

korydraughn: Ultimately, we may not see a 5.0.2 release because of the introduction of new types and functions in the query lib. Those are additions to the public API which mean the release should be 5.1.0.

Adding this to my notes to think about more.

SwooshyCueb: There is another (IMO much less elegant) approach I could take that would avoid introducing new types and functions (aside from the formatter structs). Every time one of these enums is passed to a function that takes a format string, we would have to cast it to the underlying type first.

SwooshyCueb: Okay, now that I've actually sat down and hammered it out, it's not actually as inelegant as I thought it would be. There's only one place we have to do the cast in this repository, and a few in others.

I'll update PRs shortly. There will be a new one for logical quotas, as the EL10 PR there has already been merged.

SwooshyCueb: I've added a new commit to this PR switching it over to the alternate approach.
The following PRs also have a new commit:
- irods/irods_capability_storage_tiering#351
- irods/irods_client_http_api#422

Additionally, irods/irods_rule_engine_plugin_logical_quotas#134 has been opened.

korydraughn: The latest changes are definitely a push in the right direction for reaching 5.0.2.

---

This PR introduces specializations of `fmt::formatter` for the following:
- `IRODS_STATE_ENUM`
- `ProcState`
- `IRODS_ERROR_ENUM`
- `unsigned char*`

Are the formatters restoring functionality that was available in 5.0.1?

SwooshyCueb: They are restoring functionality that was available in fmtlib 8.

korydraughn: Excellent. That means we can view the formatters as being in service of bug fixes. I think that keeps 5.0.2 as the next 5.x release.

korydraughn:
Squash to taste.

SwooshyCueb:
Squashed

korydraughn:
What's the status of this PR?

SwooshyCueb:
As I said in my comment on #8579 yesterday, it's got two new commits since the last time it was reviewed.

korydraughn (approved):
Please wait for @alanking's approval before pounding.

SwooshyCueb:
\#'d

PR #8608, opened 2025-08-07
[8295] Avoid error in rsDataObjPut when ALL_KW is set (4-3-stable)

FifthPotato:
A bunch of research for a one-line change...

Works when I test it by  hand on the test case in the original issue. Taking suggestions if this change triggers any red flags for anyone.

Still needs a unit test.

### DIFF
--- a/scripts/irods/test/test_irsync.py
+++ b/scripts/irods/test/test_irsync.py
@@ -760,3 +760,48 @@ class Test_iRsync(ResourceBase, unittest.TestCase):
         # Other case-- ensure that irods-to-local irsync do not create the target dir
         self.user0.assert_icommand(['irsync', '-rl', f'i:{source_coll_path}', nonexistent_dir_path], 'STDOUT', ['one', 'two', 'three'])
         self.assertFalse(os.path.exists(nonexistent_dir_path))
+
+    def test_irsync_with_all_keyword_does_not_fail_on_overwrite__issue_8295(self):
+        data_object_name = 'test_dataobj_8295'
+        test_data_object_path = f'{self.user0.session_collection}/{data_object_name}'
+        self.user0.assert_icommand(['itouch', test_data_object_path])
+
+        test_file_path = os.path.join(self.testing_tmp_dir, data_object_name)
+
+        with open(test_file_path, 'w') as f:
+            f.write('potato')
+
+        self.user0.assert_icommand(['irsync', '-a', test_file_path, f'i:{test_data_object_path}'])
+
+    def test_irsync_with_all_keyword__issue_8613(self):
+        data_object_name = 'all_keyword_dataobj'
+        another_resc  = 'resc_all_keyword'
+        data_object_path = f'{self.user0.session_collection}/{data_object_name}'
+
+        test_file_path = os.path.join(self.testing_tmp_dir, data_object_name)
+
+        with open(test_file_path, 'w') as f:
+            f.write('potato')
+
+        self.user0.assert_icommand(['itouch', data_object_path])
+        lib.create_ufs_resource(self.admin, another_resc)
+
+        try:
+            self.user0.assert_icommand(['irepl', '-R', another_resc, data_object_path])
+            self.user0.assert_icommand(['iput', '-f', test_file_path, data_object_path])
+
+            # Replica should be stale after the iput
+            self.assertEqual(lib.get_replica_status(self.user0, data_object_name, 1), '0')
+
+            # Edit the file, or irsync will not do the copy
+            with open(test_file_path, 'w') as f:
+                f.write('mashed potato')
+
+            self.user0.assert_icommand(['irsync', '-a', test_file_path, f'i:{data_object_path}'])
+
+            # Replica should be good after the irsync
+            self.assertEqual(lib.get_replica_status(self.user0, data_object_name, 1), '1')
+
+        finally:
+            self.user0.run_icommand(['irm', '-f', data_object_path])
+            self.admin.run_icommand(['iadmin', 'rmresc', another_resc])
--- a/server/api/src/rsDataObjPut.cpp
+++ b/server/api/src/rsDataObjPut.cpp
@@ -417,9 +417,22 @@ namespace
         if (getValByKey(&_inp.condInput, ALL_KW)) {
             /* update the rest of copies */
             transferStat_t *transStat{};
+
+            const auto* dest_resc_name = getValByKey(&_inp.condInput, DEST_RESC_NAME_KW);
+            const auto stored_resc_name = dest_resc_name ? std::optional<std::string>{dest_resc_name} : std::nullopt;
+
+            // rsDataObjRepl will error if ALL_KW and DEST_RESC_NAME_KW are set simultaneously
+            // It's also set by rsDataObjOpen in some cases, so it needs to be removed here,
+            // not in the client.
+            rmKeyVal(&_inp.condInput, DEST_RESC_NAME_KW);
+
             status = rsDataObjRepl(&_comm, &_inp, &transStat);
             if (transStat) {
                 free(transStat);
+            }
+
+            if (stored_resc_name) {
+                addKeyVal(&_inp.condInput, DEST_RESC_NAME_KW, stored_resc_name.value().c_str());
             }
         }
 


server/api/src/rsDataObjPut.cpp
```diff
@@ -417,6 +417,12 @@ namespace
         if (getValByKey(&_inp.condInput, ALL_KW)) {
             /* update the rest of copies */
             transferStat_t *transStat{};
+
+            // rsDataObjRepl will error if ALL_KW and DEST_RESC_NAME_KW are set simultaneously
+            // It's also set by rsDataObjOpen in some cases, so it needs to be removed here,
+            // not in the client.
+            rmKeyVal(&_inp.condInput, DEST_RESC_NAME_KW);
```
alanking, line 427: We need to restore this after completing the replication to avoid modifying the input structure. (I realize that the input structure is modified in other meaningful/useful ways, but this is technically a change to the API, so we can't do it here.)

server/api/src/rsDataObjPut.cpp
```diff
@@ -418,6 +418,12 @@ namespace
             /* update the rest of copies */
             transferStat_t *transStat{};
 
+            char* stored_resc_name = nullptr;
+
+            if(stored_resc_name = getValByKey(&_inp.condInput, DEST_RESC_NAME_KW)) {
```
korydraughn: Let's initialize the pointer outside of the if-stmt. No need to initialize the pointer to nullptr in this case.

server/api/src/rsDataObjPut.cpp
```diff
@@ -427,6 +433,11 @@ namespace
             if (transStat) {
                 free(transStat);
             }
+
+            if(stored_resc_name) {
+                addKeyVal(&_inp.condInput, DEST_RESC_NAME_KW, stored_resc_name);
+                free(stored_resc_name);
```
korydraughn: If an exception is thrown and `strdup` was called earlier, the call to `free` won't be invoked. I'm looking at `rsDataObjRepl` mostly here.

Consider using something like `std::optional<std::string>`. That will guarantee that memory is free'd appropriately while also giving you a tool that communicates emptiness better than `std::string::empty`.

alanking: Just wanted to note that `rsDataObjRepl` should not throw. That doesn't make this suggestion any less valid, just wanted to mention that.

FifthPotato:
Added a test case that covers the original one seen in the issue. Should I also add a test case to ensure that it indeed overwrites all replicas as expected?

alanking:
> Should I also add a test case to ensure that it indeed overwrites all replicas as expected?

It wouldn't hurt - there are, as far as I'm aware, no tests for `irsync -a` nor `iput -a`. Given that `-a` has to do with updating *at least* one replica, it may also be good to augment the existing test or add a new one such that there are multiple replicas at play.

scripts/irods/test/test_irsync.py
```diff
@@ -760,3 +760,47 @@ def test_irsync_does_not_create_collections_nor_directories_on_dry_run_issue_777
         # Other case-- ensure that irods-to-local irsync do not create the target dir
         self.user0.assert_icommand(['irsync', '-rl', f'i:{source_coll_path}', nonexistent_dir_path], 'STDOUT', ['one', 'two', 'three'])
         self.assertFalse(os.path.exists(nonexistent_dir_path))
+
+    def test_irsync_with_all_keyword_does_not_fail_on_overwrite_issue__issue_8295(self):
```
korydraughn: I think `_overwrite_issue` can just be `_overwrite`?

scripts/irods/test/test_irsync.py
```diff
@@ -760,3 +760,47 @@ def test_irsync_does_not_create_collections_nor_directories_on_dry_run_issue_777
...
+        data_object_name = 'test_dataobj_8295'
+        test_data_object_path = f'{self.user0.session_collection}/{data_object_name}'
+        self.user0.assert_icommand(['itouch', test_data_object_path])
+
+        test_file_path = os.path.join(self.testing_tmp_dir, data_object_name)
+
+        with open(test_file_path, 'w') as f:
+            f.write('potato')
+
+        self.user0.assert_icommand(['irsync', '-a', test_file_path, f'i:{test_data_object_path}'])
```
korydraughn, line 774: What is the expected result of this?

I see that we expect the operation to succeed, but it's not clear what the state of the data object and/or local file is expected to be.

scripts/irods/test/test_irsync.py
```diff
@@ -760,3 +760,47 @@ def test_irsync_does_not_create_collections_nor_directories_on_dry_run_issue_777
...
+
+    def test_irsync_with_all_keyword(self):
```
korydraughn: Needs more `__issue_NNNN`.

scripts/irods/test/test_irsync.py
```diff
@@ -760,3 +760,47 @@ def test_irsync_does_not_create_collections_nor_directories_on_dry_run_issue_777
...
+        another_resc_dir = os.path.join(self.admin.local_session_dir, 'vault_all_keyword')
+        data_object_path = f'{self.admin.session_collection}/{data_object_name}'
+
+        test_file_path = os.path.join(self.testing_tmp_dir, data_object_name)
+
+        with open(test_file_path, 'w') as f:
+            f.write('potato')
+
+        self.admin.assert_icommand(['itouch', data_object_path])
+        self.admin.assert_icommand(['iadmin', 'mkresc', another_resc, 'unixfilesystem', f'{lib.get_hostname()}:{another_resc_dir}'], 'STDOUT_SINGLELINE', 'unixfilesystem')
```
korydraughn: I think we have a convenience function for this - `lib.create_ufs_resource(...)`?

scripts/irods/test/test_irsync.py
```diff
@@ -760,3 +760,47 @@ def test_irsync_does_not_create_collections_nor_directories_on_dry_run_issue_777
...
+
+        try:
+            self.admin.assert_icommand(['irepl', '-R', another_resc, data_object_path])
+            self.admin.assert_icommand(['iput', '-f', test_file_path, data_object_path])
+
+            # Replica should be stale after the iput
+            self.assertTrue(lib.get_replica_status(self.admin, data_object_name, 1) == '0')
```
korydraughn: Prefer `self.assertEqual`.

scripts/irods/test/test_irsync.py
```diff
@@ -760,3 +760,47 @@ def test_irsync_does_not_create_collections_nor_directories_on_dry_run_issue_777
...
+
+            # Edit the file, or irsync will not do the copy
+            with open(test_file_path, 'w') as f:
+                f.write('mashed potato')
+
+            self.admin.assert_icommand(['irsync', '-a', test_file_path, f'i:{data_object_path}'])
+
+            # Replica should be good after the irsync
+            self.assertTrue(lib.get_replica_status(self.admin, data_object_name, 1) == '1')
+        finally:
```
korydraughn: Let's add an empty line above `finally`.

scripts/irods/test/test_irsync.py
```diff
@@ -760,3 +760,47 @@ def test_irsync_does_not_create_collections_nor_directories_on_dry_run_issue_777
...
+            self.assertTrue(lib.get_replica_status(self.admin, data_object_name, 1) == '0')
+
+            # Edit the file, or irsync will not do the copy
+            with open(test_file_path, 'w') as f:
+                f.write('mashed potato')
+
+            self.admin.assert_icommand(['irsync', '-a', test_file_path, f'i:{data_object_path}'])
+
+            # Replica should be good after the irsync
+            self.assertTrue(lib.get_replica_status(self.admin, data_object_name, 1) == '1')
```
korydraughn: Prefer `self.assertEqual`.

scripts/irods/test/test_irsync.py
alanking: If we don't care about the vault path being anything in particular, this could be replaced with that, yes.

scripts/irods/test/test_irsync.py
```diff
@@ -772,3 +772,35 @@ def test_irsync_with_all_keyword_does_not_fail_on_overwrite_issue__issue_8295(se
...
+        finally:
+            self.admin.run_icommand(['iadmin', 'rmresc', another_resc])
```
alanking, line 807: I imagine you would need to remove the data object first (or at least the replica on this resource) before the resource can be removed.

scripts/irods/test/test_irsync.py
```diff
@@ -772,3 +772,35 @@ def test_irsync_with_all_keyword_does_not_fail_on_overwrite_issue__issue_8295(se
...
+        another_resc  = 'resc_all_keyword'
+        another_resc_dir = os.path.join(self.admin.local_session_dir, 'vault_all_keyword')
+        data_object_path = f'{self.admin.session_collection}/{data_object_name}'
+
+        test_file_path = os.path.join(self.testing_tmp_dir, data_object_name)
+
+        with open(test_file_path, 'w') as f:
+            f.write('potato')
+
+        self.admin.assert_icommand(['itouch', data_object_path])
```
alanking: Would be better if we performed the non-admin-y things as a non-rodsadmin, but certainly not required.

korydraughn: I agree. Let's use the non-admin account instead.

scripts/irods/test/test_irsync.py
FifthPotato: Does it specifically pertain to this issue? Per Alan's note above, we never had a test that tests `irsync -a`'s functionality, so I thought this would be for that in general, not specifically related to this issue.
I can add it anyways if you think it makes sense.

korydraughn: You can open an issue which is dedicated to this task and use that.

scripts/irods/test/test_irsync.py
FifthPotato: I think succeeding is the only measurable thing for this case specifically. As noted in the original issue, the copy happens even though the error is thrown. I would expect that the error prevented the `rsDataObjRepl` call from happening properly, but this case doesn't test `irsync -a` for functionality; the other one does. Should I expand this test case in any way, or add a comment?

korydraughn: If the other test covers it, then we're all set .

You can mark this as resolved.

FifthPotato:
Core & unit tests pass. Tested `test_irsync` individually after review changes and still seems to pass.

alanking:
New changes look good. Codacy sees an unused-yet-assigned variable and clang-format is sad. Please take a look at those and then I think this may be good to go. Consider also linking #8613 in the commit message whenever squashing happens.

korydraughn:
Looks like everything has been resolved.

Is there anything else to do for this?

FifthPotato:
> Looks like everything has been resolved.
> 
> Is there anything else to do for this?

I think that's everything. If there are no objections, I'll start squishing.

trel:
agreed - squash em up

FifthPotato:
Squished.

alanking (approved):
\# it

FifthPotato:
Octothorpe'd

alanking:
@FifthPotato - Please cherry-pick to main and 5-0-stable. Thanks

Issue #8083, opened 2024-12-12
Simplify error message if iRODS env file not readable


ll4strw:
- [x] main
- [x] 5-0-stable
- [x] 4-3-stable

---

Is it possible to simplify the error message that is displayed to a user if they mistakenly use an unreadable iRODS environment file to execute icommands? A user could get confused and miss that the problem is `failed to open file [test1]`.
Perhaps also related to #8082, in the sense that there is no need to show the full stack trace? 

```
$ IRODS_ENVIRONMENT_FILE=test1 ils
 ERROR: iRODS Exception:
    file: /irods_source/lib/core/src/irods_environment_properties.cpp
    function: void irods::environment_properties::capture_json(const std::string &)
    line: 113
    code: -1 (Unknown iRODS error)
    message:
        [-]	/irods_source/lib/core/src/irods_configuration_parser.cpp:87:irods::error irods::configuration_parser::load_json_object(const std::string &) :  status [Unknown iRODS error]  errno [Operation not permitted] -- message [failed to open file [test1]]


stack trace:
--------------
 0# irods::stacktrace::dump() const in /lib/libirods_common.so.4.3.2
 1# irods::exception::assemble_full_display_what() const in /lib/libirods_common.so.4.3.2
 2# irods::exception::what() const in /lib/libirods_common.so.4.3.2
 3# irods::log(irods::exception const&) in /lib/libirods_common.so.4.3.2
 4# irods::environment_properties::capture() in /lib/libirods_common.so.4.3.2
 5# irods::environment_properties::environment_properties() in /lib/libirods_common.so.4.3.2
 6# irods::environment_properties::instance() in /lib/libirods_common.so.4.3.2
 7# capture_string_property in /lib/libirods_common.so.4.3.2
 8# getRodsEnvFromFile in /lib/libirods_common.so.4.3.2
 9# getRodsEnv in /lib/libirods_common.so.4.3.2
10# main in ils
11# 0x000070C916A29D90 in /lib/x86_64-linux-gnu/libc.so.6
12# __libc_start_main in /lib/x86_64-linux-gnu/libc.so.6
13# _start in ils


 ERROR: _rcConnect: setRhostInfo error, IRODS_HOST is probably not set correctly status = -302000 USER_RODS_HOST_EMPTY

```

trel:
definitely - yes - we can get this to be more cleanly presented to the user.

alanking:
@MartinFlores751 - Please close if complete. Thanks

PR #8628, opened 2025-08-21
[#8083] Simplify iRODS error message (4-3-stable)

MartinFlores751:
Issue quick link: #8083

This PR simplifies the error produced when using an iRODS environment file that does not exist. Following is an example of the new output:

```console
$ IRODS_ENVIRONMENT_FILE=test1 ils
 ERROR: environment_properties::capture: missing environment file. should be at [test1]
 ERROR: _rcConnect: setRhostInfo error, IRODS_HOST is probably not set correctly status = -302000 USER_RODS_HOST_EMPTY
```

---

I modified `irods::log(const irods::exception& e)` as it was the lowest level this change can happen at, potentially dealing with other related issues as well of client logging. I used `::ProcessType` to use the appropriate logging since `irods::log` is in the common lib.

Core and unit tests both passed.

### DIFF
--- a/lib/core/src/irods_log.cpp
+++ b/lib/core/src/irods_log.cpp
@@ -1,4 +1,5 @@
 #include "irods/irods_log.hpp"
+#include "irods/rcGlobalExtern.h"
 #include "irods/rodsLog.h"
 
 namespace irods {
@@ -19,6 +20,11 @@ namespace irods {
     }
 
     void log(const irods::exception& e) {
-        log(LOG_ERROR, e.what());
+        if (CLIENT_PT == ::ProcessType) {
+            log(LOG_ERROR, e.client_display_what());
+        }
+        else {
+            log(LOG_ERROR, e.what());
+        }
     }
 }


MartinFlores751:
No tests exist for this yet, but I could add one that checks for the `missing environment file` line.

lib/core/src/irods_log.cpp
```diff
@@ -19,6 +20,11 @@ namespace irods {
     }
 
     void log(const irods::exception& e) {
-        log(LOG_ERROR, e.what());
+        if (CLIENT_PT == ::ProcessType) {
+            log(LOG_ERROR, e.client_display_what());
+        }
+        else if (SERVER_PT == ::ProcessType) {
```
alanking: No `else` case makes me a tad itchy here. We should either make this the `else` case, or add one with a log message indicating that the `::ProcessType` is unknown. Thoughts?

korydraughn: Agreed. There are three process types supported by iRODS today. Making this the `else` case would cover all of the process types.

MartinFlores751: Changed to an `else` block.

korydraughn (commented):
Agreed.

MartinFlores751:
Squashed.

alanking (approved):
\# it

MartinFlores751:
Pounded.

PR #8631, opened 2025-08-22
[8590] Overwrite on no good replicas+rsync only trusts good replica checksum (4-3-stable)

FifthPotato:
Leaving in draft for now, since we agreed to stare at this first.

Solves the cases in the issues listed, but incurs another genquery. Taking suggestions, criticisms, etc.
Whatever we eventually end up agreeing on, it'll need a test as well.

### DIFF
--- a/lib/core/src/rsyncUtil.cpp
+++ b/lib/core/src/rsyncUtil.cpp
@@ -4,13 +4,16 @@
 #include "irods/rodsErrorTable.h"
 #include "irods/rodsLog.h"
 #include "irods/miscUtil.h"
+#include "irods/objInfo.h"
 #include "irods/checksum.h"
+#include "irods/filesystem.hpp"
 #include "irods/rcGlobalExtern.h"
 #include "irods/irods_at_scope_exit.hpp"
 #include "irods/irods_log.hpp"
 #include "irods/irods_hasher_factory.hpp"
 #include "irods/irods_path_recursion.hpp"
 #include "irods/irods_exception.hpp"
+#include "irods/irods_query.hpp"
 
 #include <boost/filesystem/operations.hpp>
 #include <boost/filesystem/convenience.hpp>
@@ -561,7 +564,16 @@ rsyncDataToDataUtil( rcComm_t *conn, rodsPath_t *srcPath,
     }
     else if ( strlen( srcPath->chksum ) > 0 && strlen( targPath->chksum ) > 0 ) {
         /* src and trg has a checksum value */
-        if ( strcmp( targPath->chksum, srcPath->chksum ) != 0 ) {
+
+        const static auto good_replica_string = std::to_string(GOOD_REPLICA);
+        const irods::experimental::filesystem::path _path{targPath->outPath};
+        const auto query_str = fmt::format(
+            "select DATA_REPL_STATUS where COLL_NAME = '{}' and DATA_NAME = '{}' and DATA_REPL_STATUS = '1'",
+            _path.parent_path().c_str(),
+            _path.object_name().c_str());
+        irods::query qobj{conn, query_str};
+
+        if (qobj.size() == 0 || strcmp(targPath->chksum, srcPath->chksum) != 0) {
             cpFlag = 1;
         }
     }
--- a/scripts/irods/test/test_irsync.py
+++ b/scripts/irods/test/test_irsync.py
@@ -805,3 +805,48 @@ class Test_iRsync(ResourceBase, unittest.TestCase):
         finally:
             self.user0.run_icommand(['irm', '-f', data_object_path])
             self.admin.run_icommand(['iadmin', 'rmresc', another_resc])
+
+    def test_irsync_overwrites_stale_replicas__issue_8590(self):
+        source_coll = 'thesource_8590'
+        dest_coll = 'thetarget_8590'
+        data_object_name = 'stale_replica_file_8590'
+
+        test_file_path = os.path.join(self.testing_tmp_dir, data_object_name)
+        empty_file_path = os.path.join(self.testing_tmp_dir, 'empty_file_8590')
+
+        source_coll_path = f'{self.user0.session_collection}/{source_coll}'
+        dest_coll_path = f'{self.user0.session_collection}/{dest_coll}'
+
+        dest_data_obj_path = f'{dest_coll_path}/{data_object_name}'
+
+        with open(test_file_path, 'w') as f:
+            f.write('potato')
+
+        with open(empty_file_path, 'w') as f:
+            pass
+
+        self.user0.assert_icommand(['imkdir', source_coll_path])
+        self.user0.assert_icommand(['imkdir', dest_coll_path])
+
+        self.user0.assert_icommand(['iput', test_file_path, f'{source_coll_path}/{data_object_name}'])
+        self.user0.assert_icommand(['iput', empty_file_path, dest_data_obj_path])
+
+        # Stale the replica to check that irsync will not fail with SYS_NO_GOOD_REPLICA when the target is stale
+        self.admin.assert_icommand(['iadmin', 'modrepl', 'logical_path', dest_data_obj_path, 'replica_number', '0', 'DATA_REPL_STATUS', '0'])
+
+        # Should not fail with SYS_NO_GOOD_REPLICA
+        self.user0.assert_icommand(['irsync', '-K', '-v', '-r', f'i:{source_coll_path}', f'i:{dest_coll_path}' ], 'STDOUT_SINGLELINE', data_object_name)
+
+        # Copy should have happened and set the replica to good
+        self.assertEqual(lib.get_replica_status(self.admin, data_object_name, 0), '1')
+
+        # Ensure destination has a checksum
+        self.user0.assert_icommand(['ichksum', dest_data_obj_path], 'STDOUT', [data_object_name, 'sha2:6RwlStWIYKAseI37XBpl1qiEarHcZJYxx9sW/vSvLew='])
+        # Stale the replica to check that irsync will still overwrite even when the checksums match (i.e. do not trust a stale checksum)
+        self.admin.assert_icommand(['iadmin', 'modrepl', 'logical_path', dest_data_obj_path, 'replica_number', '0', 'DATA_REPL_STATUS', '0'])
+
+        # Should cause an overwrite
+        self.user0.assert_icommand(['irsync', '-K', '-v', '-r', f'i:{source_coll_path}', f'i:{dest_coll_path}' ], 'STDOUT_SINGLELINE', data_object_name)
+
+        # Copy should have happened and set the replica to good
+        self.assertEqual(lib.get_replica_status(self.admin, data_object_name, 0), '1')
--- a/scripts/irods/test/test_irsync.py
+++ b/scripts/irods/test/test_irsync.py
@@ -764,12 +764,17 @@ class Test_iRsync(ResourceBase, unittest.TestCase):
     def test_irsync_with_all_keyword_does_not_fail_on_overwrite__issue_8295(self):
         data_object_name = 'test_dataobj_8295'
         test_data_object_path = f'{self.user0.session_collection}/{data_object_name}'
-        self.user0.assert_icommand(['itouch', test_data_object_path])
 
         test_file_path = os.path.join(self.testing_tmp_dir, data_object_name)
+        empty_file_path = os.path.join(self.testing_tmp_dir, 'empty_file_8295')
 
         with open(test_file_path, 'w') as f:
             f.write('potato')
+
+        with open(empty_file_path, 'w') as f:
+            pass
+
+        self.user0.assert_icommand(['iput', empty_file_path, test_data_object_path])
 
         self.user0.assert_icommand(['irsync', '-a', test_file_path, f'i:{test_data_object_path}'])
 
@@ -779,11 +784,15 @@ class Test_iRsync(ResourceBase, unittest.TestCase):
         data_object_path = f'{self.user0.session_collection}/{data_object_name}'
 
         test_file_path = os.path.join(self.testing_tmp_dir, data_object_name)
+        empty_file_path = os.path.join(self.testing_tmp_dir, 'empty_file_8613')
 
         with open(test_file_path, 'w') as f:
             f.write('potato')
 
-        self.user0.assert_icommand(['itouch', data_object_path])
+        with open(empty_file_path, 'w') as f:
+            pass
+
+        self.user0.assert_icommand(['iput', empty_file_path, data_object_path])
         lib.create_ufs_resource(self.admin, another_resc)
 
         try:
--- a/server/api/src/rsDataObjRsync.cpp
+++ b/server/api/src/rsDataObjRsync.cpp
@@ -304,7 +304,9 @@ rsRsyncDataToData( rsComm_t *rsComm, dataObjInp_t *dataObjInp ) {
                 return status;
             }
         }
-        else if (status < 0 && status != CAT_NO_ACCESS_PERMISSION && status != CAT_NO_ROWS_FOUND) {
+        else if (status < 0 && status != CAT_NO_ACCESS_PERMISSION && status != CAT_NO_ROWS_FOUND &&
+                 status != SYS_NO_GOOD_REPLICA)
+        {
             // NOLINTNEXTLINE(cppcoreguidelines-pro-type-vararg, cppcoreguidelines-pro-bounds-array-to-pointer-decay)
             rodsLog(LOG_ERROR,
                     "rsRsyncDataToData: _rsDataObjChksum error for %s, status = %d",


lib/core/src/rsyncUtil.cpp
```diff
@@ -561,7 +563,20 @@ rsyncDataToDataUtil( rcComm_t *conn, rodsPath_t *srcPath,
     }
     else if ( strlen( srcPath->chksum ) > 0 && strlen( targPath->chksum ) > 0 ) {
         /* src and trg has a checksum value */
-        if ( strcmp( targPath->chksum, srcPath->chksum ) != 0 ) {
+
+        const auto path_str = std::string(targPath->outPath);
+        const auto separator_index = path_str.rfind("/");
```
korydraughn: Use the filesystem path class to simplify extraction of path elements and make the code more readable.

The path class provides `.parent_path()` and `.object_name()` for this kind of thing.

See filesystem.cpp for several examples.

lib/core/src/rsyncUtil.cpp
```diff
@@ -561,7 +563,20 @@ rsyncDataToDataUtil( rcComm_t *conn, rodsPath_t *srcPath,
...
+        const auto query_str = fmt::format("select DATA_REPL_STATUS where COLL_NAME = '{}' and DATA_NAME = '{}'",
+                                           path_str.substr(0, separator_index),
+                                           path_str.substr(separator_index + 1));
+        irods::query qobj{conn, query_str};
+
+        rodsLog(LOG_ERROR, path_str.c_str());
+        rodsLog(LOG_ERROR, query_str.c_str());
+        if (std::find_if(qobj.begin(), qobj.end(), [](auto row) { return row[0] == std::to_string(GOOD_REPLICA); }) ==
```
korydraughn: `auto row` is going to result in a copy on every invocation. Use `const auto&`.

korydraughn: Consider pulling `std::to_string(GOOD_REPLICA);` out of the lambda and computing it once. Capture the result and use that instead.

Alternatively, you can define it in the lambda's capture like this:
```cpp
[good = std::to_string(GOOD_REPLICA)](const auto& row) { ...
```

trel:
Please also audit the number of network calls before and after this change.  Additional database calls within the server may not be as noticeable/important.

Going from 1 to 2 will be 100% more, whereas going from 15 to 16 will be a smaller % increase.

lib/core/src/rsyncUtil.cpp
alanking: To take it a step further, it could even be `constexpr` or at least `static`. The value will (hopefully) never change.

FifthPotato:
> Please also audit the number of network calls before and after this change. Additional database calls within the server may not be as noticeable/important.
> 
> Going from 1 to 2 will be 100% more, whereas going from 15 to 16 will be a smaller % increase.

Looks like we're going from 11 (and a half) to 12 (and a half).

My methodology was that I used `strace` on the call before and after and counted each block of write (from the client)/read (from the server) on the socket as a single "network call."

trel:
That's for each file?   That is a lot / more than I expected.

But that also means one more is probably okay for now, as correct is better than fast and wrong.

lib/core/src/rsyncUtil.cpp
```diff
@@ -561,7 +564,18 @@ rsyncDataToDataUtil( rcComm_t *conn, rodsPath_t *srcPath,
     }
     else if ( strlen( srcPath->chksum ) > 0 && strlen( targPath->chksum ) > 0 ) {
         /* src and trg has a checksum value */
-        if ( strcmp( targPath->chksum, srcPath->chksum ) != 0 ) {
+
+        const static auto good_replica_string = std::to_string(GOOD_REPLICA);
+        const irods::experimental::filesystem::path _path{targPath->outPath};
+        const auto query_str = fmt::format("select DATA_REPL_STATUS where COLL_NAME = '{}' and DATA_NAME = '{}'",
+                                           _path.parent_path().string(),
+                                           _path.object_name().string());
```
korydraughn: Use `.c_str()` instead of `.string()` to avoid allocation.

lib/core/src/rsyncUtil.cpp
```diff
@@ -561,7 +564,18 @@ rsyncDataToDataUtil( rcComm_t *conn, rodsPath_t *srcPath,
     }
     else if ( strlen( srcPath->chksum ) > 0 && strlen( targPath->chksum ) > 0 ) {
         /* src and trg has a checksum value */
-        if ( strcmp( targPath->chksum, srcPath->chksum ) != 0 ) {
+
+        const static auto good_replica_string = std::to_string(GOOD_REPLICA);
```
korydraughn, line 568: I think we can just hard code the value into the lambda and avoid this step altogether. Something like this:
```cpp
const auto is_good_replica = [](const auto& row) {
    return row[0] == "1";
};
if (std::none_of(std::begin(qobj), std::end(qobj), is_good_replica) || strcmp(...)) {
```

This gives the lambda a name which captures its intent. The use of `std::none_of` makes the intent of the code a little clearer as well.

You'll need to include `<algorithm>` for `std::none_of` too.

korydraughn: Looking at the code more, why do we not check for good replicas as part of the GenQuery?

FifthPotato:
> That's for each file? That is a lot / more than I expected.
> 
> But that also means one more is probably okay for now, as correct is better than fast and wrong.

Realized I didn't test it per file. Redid the test with 2 files and got 13.5 to 15.5
So it looks like the old rate is 11.5 for 1 file (includes all the startup/teardown stuff) + 2 for each additional file, and the new rate is 12.5 for 1 file + 3 for each additional file.
So it could still be relevant for a lot of files-- 2 to 3 is a 50% increase.

trel:
Good - that's a bit more expected.   Please do some timing tests on a few thousand files in an rsync (both localhost and over a 'real' network)... then we'll make a decision about whether to ship this or keep looking for places to speed up.

FifthPotato:
> Good - that's a bit more expected. Please do some timing tests on a few thousand files in an rsync (both localhost and over a 'real' network)... then we'll make a decision about whether to ship this or keep looking for places to speed up.

Finally got some numbers. I used 3212 files with ~500kb of random data.

Without change, localhost:
6m53.208s
 
With change, localhost:
6m54.409s

Without change, 150ms simulated delay with tc:
89m7.419s

With change, 150ms simulated delay with tc:
105m20.899s

So it looks like it barely matters if there's virtually no delay, but increases the time by 18% for ~3200 files on 150ms.
Do we think this is good enough to keep, or is it time to search for alternative solutions?

trel:
+150ms changed the overall time from ~7m to ~90m?  that seems a lot.

can you run with and without tc...
 - 3000 1kb files?  i expect worse % overhead here.
 - 100 50MB files?  i expect better/less % overhead here.

FifthPotato:
> +150ms changed the overall time from ~7m to ~90m? that seems a lot.
> 
> can you run with and without tc...
> 
>     * 3000 1kb files?  i expect worse % overhead here.
> 
>     * 100 50MB files?  i expect better/less % overhead here.

With or without the new change?

trel:
both, please, so we can see a comparison.

three variables:
- tc - 0ms, 150ms
- file sizes - 10 50MB, 3000 1k, 3212 500k
- additional query - without, with

so 2x2x3...

| case | tc | files | additional query | total time | overhead |
| --- | --- | --- | --- | --- | --- |
| 1 | 0ms | 10 50MB | no | | - |
| 2 | 0ms | 10 50MB | yes | | |
| 3 | 0ms | 3000 1k | no | | - |
| 4 | 0ms | 3000 1k | yes | | |
| 5 | 0ms | 3212 500k | no | 6m53s | - |
| 6 | 0ms | 3212 500k | yes | 6m54s | 0.2% |
| 7 | 150ms | 10 50MB | no | | - |
| 8 | 150ms | 10 50MB | yes | | |
| 9 | 150ms | 3000 1k | no | | - |
| 10 | 150ms | 3000 1k | yes | | |
| 11 | 150ms | 3212 500k | no | 89m7s | - |
| 12 | 150ms | 3212 500k | yes | 105m20s | 18.2% |

FifthPotato:
> both, please, so we can see a comparison.
> 
> three variables:
> 
>     * tc - 0ms, 150ms
> 
>     * file sizes - 10 50MB, 3000 1k, 3212 500k
> 
>     * additional query - without, with
> 
> 
> so 2x2x3...
> case 	tc 	files 	additional query 	total time 	overhead
> 1 	0ms 	10 50MB 	no 		-
> 2 	0ms 	10 50MB 	yes 		
> 3 	0ms 	3000 1k 	no 		-
> 4 	0ms 	3000 1k 	yes 		
> 5 	0ms 	3212 500k 	no 	6m53s 	-
> 6 	0ms 	3212 500k 	yes 	6m54s 	0.2%
> 7 	150ms 	10 50MB 	no 		-
> 8 	150ms 	10 50MB 	yes 		
> 9 	150ms 	3000 1k 	no 		-
> 10 	150ms 	3000 1k 	yes 		
> 11 	150ms 	3212 500k 	no 	89m7s 	-
> 12 	150ms 	3212 500k 	yes 	105m20s 	18.2%

I've got numbers for the remaining cells.

case 1:
real    0m3.911s

case 2:
real    0m3.834s

case 3:
real    6m22.421s

case 4:
real    6m28.283s


case 7:
real    0m23.520s

case 8:
real    0m26.587s

case 9:
real    84m4.605s

case 10:
real    98m4.843s

These tests were done with a testing methodology that's better than the previous ones, so these numbers should be more accurate.

korydraughn:
Please explain the testing methodology used for those results.

FifthPotato:
> Please explain the testing methodology used for those results.

It was basically identical to the previous one, but to ensure the correct codepath was tested, I generated checksums for both source and target, and made sure the target checksums were incorrect, so it'd trigger an overwrite.
Specifically, the setup was all some variant of
```
for i in $(seq 1 3000); do dd if=/dev/random of=testfile bs=1K count=1; iput testfile thesource/testfile$i; ichksum thesource/testfile$i; itouch thetarget/testfile$i; ichksum thetarget/testfile$i; done
```
and then 
```
time irsync -K -v -r i:thesource i:thetarget
```

korydraughn:
You can also render a new table showing the overhead for each case.

FifthPotato:
I've omitted the old cases to avoid confounding factors.

|case|tc|files|new query|total time|overhead|
|---|---|---|---|---|---|
|1|0ms|10 50MB|no|3.911s|-|
|2|0ms|10 50MB|yes|3.834s|0%|
|3|0ms|3000 1K|no|382.421s|-|
|4|0ms|3000 1K|yes|388.283s|1.53%|
|7|150ms|10 50MB|no|23.520s|-|
|8|150ms|10 50MB|yes|26.587s|13.04%|
|9|150ms|3000 1K|no|5045.605s|-|
|10|150ms|3000 1K|yes|5885.843s|16.65%|

trel:
8 is 13%

so mostly, we're looking at a ~15% overhead to be 'correct'.

trel:
and these were single-run measurements, rather than an average over 3-4-5 runs, yes?

FifthPotato:
> 8 is 13%
> 
> so mostly, we're looking at a ~15% overhead to be 'correct'.

Oops, mathed wrong.



> and these were single-run measurements, rather than an average over 3-4-5 runs, yes?

Correct. I don't mind doing extra runs for the shorter ones (if necessary), but the longer ones take a very long time.

korydraughn:
After some discussion, we've agreed that the reported numbers are acceptable.

@FifthPotato Please proceed with next steps for getting this merged.

scripts/irods/test/test_irsync.py
```diff
@@ -805,3 +805,37 @@ def test_irsync_with_all_keyword__issue_8613(self):
...
+        dest_data_obj_path = f'{dest_coll_path}/{data_object_name}'
+
+        with open(test_file_path, 'w') as f:
+            f.write('potato')
+
+        self.user0.assert_icommand(['imkdir', source_coll_path])
+        self.user0.assert_icommand(['imkdir', dest_coll_path])
+
+        self.user0.assert_icommand(['iput', test_file_path, f'{source_coll_path}/{data_object_name}'])
+        self.user0.assert_icommand(['itouch', f'{dest_coll_path}/{data_object_name}'])
```
trel:
```suggestion
        self.user0.assert_icommand(['itouch', dest_data_obj_path])
```

scripts/irods/test/test_irsync.py
```diff
@@ -805,3 +805,37 @@ def test_irsync_with_all_keyword__issue_8613(self):
...
+        # Should not fail with SYS_NO_GOOD_REPLICA
+        self.user0.assert_icommand(['irsync', '-K', '-v', '-r', f'i:{source_coll_path}', f'i:{dest_coll_path}' ], 'STDOUT_SINGLELINE', data_object_name)
+
+        self.user0.assert_icommand(['ichksum', dest_data_obj_path], 'STDOUT_SINGLELINE', data_object_name)
+        self.admin.assert_icommand(['iadmin', 'modrepl', 'logical_path', dest_data_obj_path, 'replica_number', '0', 'DATA_REPL_STATUS', '0'])
+
+        self.user0.assert_icommand(['irsync', '-K', '-v', '-r', f'i:{source_coll_path}', f'i:{dest_coll_path}' ], 'STDOUT_SINGLELINE', data_object_name)
+
+        # Copy should have happened and set the replica to good
+        self.assertTrue(lib.get_replica_status(self.admin, data_object_name, 0) == '1')
```
korydraughn: Prefer `assertEqual` or `assertEquals`, whichever is the true one.

scripts/irods/test/test_irsync.py
```diff
@@ -805,3 +805,37 @@ def test_irsync_with_all_keyword__issue_8613(self):
...
+        self.user0.assert_icommand(['imkdir', source_coll_path])
+        self.user0.assert_icommand(['imkdir', dest_coll_path])
+
+        self.user0.assert_icommand(['iput', test_file_path, f'{source_coll_path}/{data_object_name}'])
+        self.user0.assert_icommand(['itouch', f'{dest_coll_path}/{data_object_name}'])
+
+        self.admin.assert_icommand(['iadmin', 'modrepl', 'logical_path', dest_data_obj_path, 'replica_number', '0', 'DATA_REPL_STATUS', '0'])
+
+        # Should not fail with SYS_NO_GOOD_REPLICA
+        self.user0.assert_icommand(['irsync', '-K', '-v', '-r', f'i:{source_coll_path}', f'i:{dest_coll_path}' ], 'STDOUT_SINGLELINE', data_object_name)
```
trel, line 847: what SHOULD have happened here?   is the target now complete and good?   please assertTrue that after this happens, if so.

scripts/irods/test/test_irsync.py
```diff
@@ -805,3 +805,37 @@ def test_irsync_with_all_keyword__issue_8613(self):
...
+        with open(test_file_path, 'w') as f:
+            f.write('potato')
+
+        self.user0.assert_icommand(['imkdir', source_coll_path])
+        self.user0.assert_icommand(['imkdir', dest_coll_path])
+
+        self.user0.assert_icommand(['iput', test_file_path, f'{source_coll_path}/{data_object_name}'])
+        self.user0.assert_icommand(['itouch', f'{dest_coll_path}/{data_object_name}'])
+
+        self.admin.assert_icommand(['iadmin', 'modrepl', 'logical_path', dest_data_obj_path, 'replica_number', '0', 'DATA_REPL_STATUS', '0'])
```
trel, line 844: add a comment for these setup lines to make sure it's easy for a reader to know the intent

```
# stale the single target replica, so the next steps have work to do
```

trel (commented):
looks good, is there anything else this fix needs to check?

scripts/irods/test/test_irsync.py
```diff
@@ -805,3 +805,37 @@ def test_irsync_with_all_keyword__issue_8613(self):
...
+
+        # Should not fail with SYS_NO_GOOD_REPLICA
+        self.user0.assert_icommand(['irsync', '-K', '-v', '-r', f'i:{source_coll_path}', f'i:{dest_coll_path}' ], 'STDOUT_SINGLELINE', data_object_name)
+
+        self.user0.assert_icommand(['ichksum', dest_data_obj_path], 'STDOUT_SINGLELINE', data_object_name)
+        self.admin.assert_icommand(['iadmin', 'modrepl', 'logical_path', dest_data_obj_path, 'replica_number', '0', 'DATA_REPL_STATUS', '0'])
+
+        self.user0.assert_icommand(['irsync', '-K', '-v', '-r', f'i:{source_coll_path}', f'i:{dest_coll_path}' ], 'STDOUT_SINGLELINE', data_object_name)
```
korydraughn: Consider adding more comments to help the reader understand why things are being done. That speeds up the process of figuring out what the test is doing, removes assumptions, and allows us to determine if an icommand is being used correctly.

scripts/irods/test/test_irsync.py
```diff
@@ -805,3 +805,44 @@ def test_irsync_with_all_keyword__issue_8613(self):
...
+        self.admin.assert_icommand(['iadmin', 'modrepl', 'logical_path', dest_data_obj_path, 'replica_number', '0', 'DATA_REPL_STATUS', '0'])
+
+        # Should not fail with SYS_NO_GOOD_REPLICA
+        self.user0.assert_icommand(['irsync', '-K', '-v', '-r', f'i:{source_coll_path}', f'i:{dest_coll_path}' ], 'STDOUT_SINGLELINE', data_object_name)
+
+        # Copy should have happened and set the replica to good
+        self.assertEqual(lib.get_replica_status(self.admin, data_object_name, 0), '1')
+
+        # Ensure destination has a checksum
+        self.user0.assert_icommand(['ichksum', dest_data_obj_path], 'STDOUT_SINGLELINE', data_object_name)
```
korydraughn: Let's assert the value of the checksum is what we expect. It should be known given we know the contents of the data object.

FifthPotato:
Unit and core tests pass. Squished.

korydraughn:
Please run only your new test in a topology-consumer environment (i.e. `run_topology_tests.py ... --tests <your_new_test>`). There were two test failures discovered for `irsync -a` recently and they only fail when run from a consumer server within a topology environment.

FifthPotato:
Figured out what was breaking the topology tests-- seems like different iCommands have different opinions as to what is the "default" resource when doing `iput`/`itouch`/`irsync`, so I needed to specifically target `demoResc` to make it work properly. Taking more review comments.

alanking:
> Figured out what was breaking the topology tests-- seems like different iCommands have different opinions as to what is the "default" resource when doing `iput`/`itouch`/`irsync`, so I needed to specifically target `demoResc` to make it work properly. Taking more review comments.

That would be... surprising. Is it not the case that the default resource changes depending on the server in our topology tests?

korydraughn:
After some discussion, we've agreed that we need to understand why `-R demoResc` for the irsync tests fixes the problem. It's possible the client's default resource is not defined in the irods_environment.json file or being forwarded to the API call. This would result in the server selecting its default resource. All of that to say, the tests might have revealed a bug in `irsync`.

The work for the `irsync` test failures will be pulled into a separate PR.

_This comment is in regard to commits ab836a2 and 0c5fd34 for issues #8669 and #8668._

FifthPotato:
I've reverted the other two changes for now, and will look into why `-R demoResc` fixes it.

FifthPotato:
Refactored to avoid `itouch` and all the tests should pass now. Open to review comments.

korydraughn (commented):
Latest changes look good.

Did you confirm the two tests pass in a topology-consumer environment?

Also, leave a comment in the PR conversation explaining why itouch was swapped with iput so the full story is captured.

FifthPotato:
> Latest changes look good.
> 
> Did you confirm the two tests pass in a topology-consumer environment?
> 
> Also, leave a comment in the PR conversation explaining why itouch was swapped with iput so the full story is captured.

Yep, they all pass under `run_topology_tests.py`.

I'll leave that comment here: `itouch` has some potentially counterintuitive (but expected?) behavior in how it performs its default resource selection. Namely, it ignores the user's environment choice of default resource, but `iput` and `irsync` obey the environment choice, so `itouch` can end up creating data objects on a different resource, which results in weirdness in topology-consumer environments.

korydraughn:
Good good.

Squash to taste, no pounds.

FifthPotato:
Squished

korydraughn:
Please tell a story in the body of each commit message explaining the why. It doesn't need to be pages of text, just enough to help the reader in cases where they may not have access to github issues.

Consider rewording the top-level commit message for 8590. It's not clear what it means.

FifthPotato:
Pushed some new wording.

korydraughn:
New wording is much better. Please break the long lines so they are easier to read in small panes.

FifthPotato:
Added line breaks so each line is <=80 chars.

korydraughn:
Is github adding the `...` to the top-level comment for 8590? Does that appear when viewing it via git directly?

FifthPotato:
Octothorpe'd, after working on commit messages with Kory.

Issue #7494, opened 2024-02-12
Post-install test fails when using mariadb


SwooshyCueb:
Setting up irods with the mysql plugin pointed at a mariadb server and using the mariadb odbc driver errors out at the very end during teh post-install test:

```
+---------------------------+
| Running Post-Install Test |
+---------------------------+

Calling ['/usr/sbin/irodsTestPutGet'] with options:
{'shell': False, 'stderr': -1, 'stdout': -1}
Command /usr/sbin/irodsTestPutGet returned with code 1.
stderr:
  Error: cannot remove data object ["/tempZone/home/rods/post_install_test.1707763399640759.txt"]
Error encountered running setup_irods:
Traceback (most recent call last):
  File "/var/lib/irods/scripts/setup_irods.py", line 523, in main
    setup_server(irods_config,
  File "/var/lib/irods/scripts/setup_irods.py", line 148, in setup_server
    test_put(irods_config)
  File "/var/lib/irods/scripts/setup_irods.py", line 180, in test_put
    raise IrodsError('Post-install test failed. Please check your configuration.')
irods.exceptions.IrodsError: Post-install test failed. Please check your configuration.
Exiting...
```

However, a quick test with `ils` and `itouch` shows that the server is working, albeit with the `CAT_SUCCESS_BUT_WITH_NO_INFO` behavior described in #7323:

```
root@7913a03a187c:/tmp# sudo -u irods -g irods -H ils
/tempZone/home/rods:
root@7913a03a187c:/tmp# sudo -u irods -g irods -H itouch asdf
Level 0: CAT_SUCCESS_BUT_WITH_NO_INFO: cannot set mtime for [/tempZone/home/rods/asdf]


root@7913a03a187c:/tmp# sudo -u irods -g irods -H ils
/tempZone/home/rods:
  asdf
```

This has been observed with a post-#7485 build of irods in the following environments:
- Ubuntu 22.04 container with MariaDB 10.6 from the Ubuntu repos running in the same container, using the MariaDB ODBC driver from the Ubuntu repos.
- Ubuntu 22.04 container with MariaDB 10.6 from the Ubuntu repos running in the same container, using the MariaDB ODBC driver version 3.1.20 provided by MariaDB.

MartinFlores751:
After a live debugging session with @korydraughn and @trel present, we found that adding a sleep after `create_data_object_if_necessary(...)` fixed this issue.

[...]

Issue #5684, opened 2021-06-02
Data path is truncated in MySQL database


korydraughn:
- [x] master
- [x] 4-2-stable
---
### iRODS Version, OS and Version
4-2-stable

### What did you try to do?
I ran the following test for `test_resource_types.Test_Resource_Unixfilesystem` against an iRODS server backed by MySQL:
https://github.com/irods/irods/blob/1b73314a129c68b975eadd584c26594092240e26/scripts/irods/test/test_resource_types.py#L560

### Expected behavior
I expected the test to pass.

### Observed behavior (including steps to reproduce, if applicable)
The test fails because `ichksum -K` cannot stat the replica. This happens because the `data_path` in `r_data_main` is truncated to 255 bytes.
https://github.com/irods/irods/blob/1b73314a129c68b975eadd584c26594092240e26/scripts/irods/test/test_resource_types.py#L574-L577

PR #8641, opened 2025-08-27
[#7494] Use `option` to set MySQL and MariaDB options (4-3-stable)

MartinFlores751:
Issue quick link: #7494

This PR attempts to fix the issue relating to MariaDB having unexpected behavior by setting `option` to return the number of found rows instead of the number of changed rows. The settings in `option` are set by toggling the relevant bits, e.g., `2` sets bit `1` which is equivalent to enabling `FOUND_ROWS` in MySQL.

`option` is used here since MariaDB does not support some MySQL aliases such as `FOUND_ROWS`. They must be set in `option`.

The relevant MySQL documentation for `option` can be found here: https://dev.mysql.com/doc/connector-odbc/en/connector-odbc-configuration-connection-parameters.html#codbc-dsn-option-flags

The documentation for MariaDB `option` can be found here: https://mariadb.com/docs/connectors/mariadb-connector-odbc/mariadb-connector-odbc-guide#general-connection-parameters

---

Before getting out of draft the following will be completed:
- Test against MySQL with the following `option`:
  - [ ] "10" or both `BIG_PACKETS` and `FOUND_ROWS`
  - [ ] "8" or just `BIG_PACKETS`
  - [ ] "2" or just `FOUND_ROWS`

### DIFF
--- a/scripts/irods/database_connect.py
+++ b/scripts/irods/database_connect.py
@@ -86,8 +86,7 @@ def get_odbc_entry(db_config, catalog_database_type):
             'Server': db_config['db_host'],
             'Port': str(db_config['db_port']),
             'Database': db_config['db_name'],
-            'FOUND_ROWS': '1',
-            'BIG_PACKETS': '1',
+            'option': '10', # Set both FOUND_ROWS (2) and BIG_PACKETS (8)
             'Charset': 'UTF8'
         }
     elif catalog_database_type == 'oracle':


trel:
So we're just 'restoring' example/known/desired behavior... rather than changing anything to fix a bug.

This was already 'solved' for MySQL and MariaDB just doesn't use the same descriptors in the ODBC file.

This is great.

korydraughn:
So everything is working as intended? You're confident the changes are good?

MartinFlores751:
Due to issues in reproducing #5684 and the MariaDB tests possibly needing to be redone (setting `mariadb_use_mysql_odbc_driver` to `False`, see irods/irods_testing_environment#224), the PR will be pulled back into draft and worked on again at a later point.

Issue #8546, opened 2025-05-30
Data objects without any permissions on them aren't visible to GenQuery2


korydraughn:
## Bug Report

Observed in iRODS 4.3.4 while testing irods4j.

Assume the user is a rodsadmin - if there are no ACLs set on a data object, GenQuery2 will not find it.

This is most likely due to GenQuery2 always including permission checks. It assumes that adjusting the permission level is enough to cover all cases.

The code of interest exists in the function containing these lines.

https://github.com/irods/irods/blob/2d7d0e4d2251dd3a052dfb1f00c410afb7aa834d/server/genquery2/src/genquery2_sql.cpp#L408-L411

PR #8642, opened 2025-08-27
[#8546] GenQuery2: Fix ability to find objects with no user permissions (main)

korydraughn:
New tests pass at the bench.

### DIFF
--- a/server/genquery2/src/genquery2_sql.cpp
+++ b/server/genquery2/src/genquery2_sql.cpp
@@ -257,7 +257,7 @@ namespace
 
         std::vector<std::string> processed;
         processed.reserve(_tables.size());
-        processed.emplace_back(std::string{_tables.front()});
+        processed.emplace_back(_tables.front());
         log_gq::debug(fmt::format("processed = [{}]", fmt::join(processed, ", ")));
 
         for (decltype(_tables.size()) i = 0; i < _tables.size() - 1; ++i) {
@@ -322,13 +322,13 @@ namespace
         return sql;
     } // generate_joins_for_metadata_columns
 
-    auto generate_joins_for_permissions(const std::map<std::string, std::string>& _table_aliases) -> std::string
+    auto generate_joins_for_permissions(gq_state& _state, const gq::options& _opts) -> std::string
     {
         // Always include the joins if the query involves columns related to data objects and/or collections.
         // This is required due to how columns in R_OBJT_ACCESS and other tables are handled.
         //
-        // While the additional joins aren't theoretically necessary for a rodsadmin, they are included to
-        // cover cases where the rodsadmin has requested information that only exists in the permission tables.
+        // LEFT JOIN is used instead of INNER JOIN for rodsadmins to allow discovery of data objects and
+        // collections having NO user permissions on them. Using INNER JOIN would hide these objects otherwise.
         //
         //  select d.*
         //  from R_DATA_MAIN d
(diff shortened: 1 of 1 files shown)


server/genquery2/src/genquery2_sql.cpp
```diff
@@ -344,20 +344,39 @@ namespace
...
-                               " inner join R_TOKN_MAIN pct on pcoa.access_type_id = pct.token_id"
-                               " inner join R_USER_GROUP pcug on pcoa.user_id = pcug.group_user_id"
-                               " inner join R_USER_MAIN pcu on pcug.user_id = pcu.user_id",
-                               iter->second);
+            if (const auto iter = _state.table_aliases.find("R_COLL_MAIN"); iter != std::end(_state.table_aliases)) {
+                sql += fmt::format(" inner join R_OBJT_ACCESS pcoa on {}.coll_id = pcoa.object_id"
+                                   " inner join R_TOKN_MAIN pct on pcoa.access_type_id = pct.token_id"
+                                   " inner join R_USER_GROUP pcug on pcoa.user_id = pcug.group_user_id"
+                                   " inner join R_USER_MAIN pcu on pcug.user_id = pcu.user_id",
+                                   iter->second);
```
korydraughn, line 384: Are inner-joins really necessary?
Do left-joins solve the problem for all cases?

server/genquery2/src/genquery2_sql.cpp
```diff
@@ -428,50 +440,35 @@ namespace
...
-                }
-            }
-            else {
+            if (!_opts.admin_mode) {
                 if (d_iter != end && c_iter != end) {
                     sql += fmt::format(" and pdu.user_name = ?"
                                        " and pdu.zone_name = '{zone}'"
                                        " and pdu.user_type_name != 'rodsgroup'"
-                                       " and pdoa.access_type_id >= {perm}"
+                                       " and pdoa.access_type_id >= 1050"
```
trel, line 458: why going to a 'magic' number here?   can we not use a variable for futureproofing / documentation?

korydraughn: Using the integer value for the permission level is about optimization. We're avoiding unnecessary work for the CPU.

I think a comment or documentation showing the various permission levels would cover this.

trel: makes sense.    just want to make sure it's easy for the next reader to understand.

korydraughn: Will add a comment explaining that value.

server/genquery2/src/genquery2_sql.cpp
alanking: Wouldn't it be bad for users to get back results for things which have no permissions on them? I would think that only admins should be allowed to access such items. I could be misunderstanding how this works or would work.

korydraughn: That's what I thought would be the case, but benchtesting this produced the same results as the inner-joins.

More testing is needed to be fully sure. I think left-joins work (i.e. do not return all rows) because the non-admin queries always have a permission check condition attached to them. The table representing the right side of the join will contain null for the columns which are not matched and those nulls will always fail the condition check resulting in them being discarded from the results.

alanking: That makes sense. If we can use similar queries for all paths, that reduces the variables, so that would be good overall.

korydraughn:
Core tests pass.

Going to continue to think about inner/left joins though.

server/genquery2/src/genquery2_sql.cpp
korydraughn: I'm going to move forward with left and inner joins. It's not clear to me what the blast radius is for using left-joins. Limiting the use of left-joins (for this case) to rodsadmins is safer too.

I'll leave a comment highlighting this thought.

korydraughn:
Moving forward with inner-joins for non-admins and left-joins for rodsadmins.

See https://github.com/irods/irods/pull/8642#discussion_r2330173869 for more info.

Addressing review comments and then this will be ready.

korydraughn:
Reran unit tests. They passed.

```
-----
All tests passed! :)
time elapsed: [ 176.2603]seconds ([   0]hours [ 2.9377]minutes)

2025-09-08 12:17:32,998 - version:[5.0.90]
2025-09-08 12:17:32,999 - sha:[74b5683169e518e86cabc552da57ac221019a9bd]
```

Rerunning core tests to make sure everything is good.

server/genquery2/src/genquery2_sql.cpp
korydraughn: Done.

server/genquery2/src/genquery2_sql.cpp
korydraughn: Done.

korydraughn:
All core tests passed.

```
-----
All tests passed! :)
time elapsed: [7447.3064]seconds ([   2]hours [ 4.1218]minutes)

2025-09-08 14:38:55,319 - version:[5.0.90]
2025-09-08 14:38:55,320 - sha:[74b5683169e518e86cabc552da57ac221019a9bd]
```

Waiting for signal to squash and pound.

korydraughn:
Squashed.

korydraughn:
Added pounds.

Issue #8479, opened 2025-05-07
deprecate the control plane and irods-grid


trel:
iRODS introduced the control plane (listening on port 1248) 10 years ago in iRODS 4.1.0 (https://github.com/irods/irods/issues/2298).

It will be removed in iRODS 5.0.0 as part of the redesign of the process model (https://github.com/irods/irods/issues/7229).

- [ ] add deprecation messages to irods_docs
- [ ] add deprecation messages to `irods-grid`

PR #8645, opened 2025-08-28
[#8479] Deprecate irods-grid command (4-3-stable)

korydraughn:
Companion PR at irods/irods_docs#386.

Will confirm the help text is printed correctly soon.

### DIFF
--- a/server/irods_grid/src/irods-grid.cpp
+++ b/server/irods_grid/src/irods-grid.cpp
@@ -25,6 +25,8 @@ template <typename T>
 irods::error usage(T& ostream) {
     ostream << "Usage:  'irods-grid action [ option ] target'" << std::endl;
     ostream << " " << std::endl;
+    ostream << "Note: This command is deprecated and has been removed in iRODS 5." << std::endl;
+    ostream << " " << std::endl;
     ostream << "Sends a message on control channel for possible grid-wide operations" << std::endl;
     ostream << " " << std::endl;
     ostream << "action: ( required ) status, ping, pause, resume, shutdown" << std::endl;


server/irods_grid/src/irods-grid.cpp
```diff
@@ -25,6 +25,8 @@ template <typename T>
 irods::error usage(T& ostream) {
     ostream << "Usage:  'irods-grid action [ option ] target'" << std::endl;
     ostream << " " << std::endl;
+    ostream << "Note: This command is deprecated and will be removed in a later release." << std::endl;
```
alanking: Can it be more specific? Current wording is fine, too.
```suggestion
    ostream << "Note: This command is deprecated and will not be available in iRODS 5." << std::endl;
```

korydraughn: It can. It wasn't clear to me which one felt more appropriate. I think stating iRODS 5 is good since we know it's not in that version.

Will update.

korydraughn:
I've confirmed that the current changes are printed correctly.
```
irods@e5dc9ffecd61:~$ irods-grid -h
Usage:  'irods-grid action [ option ] target'
 
Note: This command is deprecated and will be removed in a later release.
 
Sends a message on control channel for possible grid-wide operations

<snip>
```

Will update the text based on the review comments and this will be ready.

server/irods_grid/src/irods-grid.cpp
korydraughn: Tweaked wording. It is slightly different from the suggested wording.

Let me know what you think.

alanking (commented):
New wording looks good to me. Squash em

korydraughn:
Squashed.

alanking (approved):
\# it

korydraughn:
Pounded.

Issue #6756, opened 2022-12-02
Deprecate and replace all use of `uint`


korydraughn:
- [ ] main
- [ ] 4-3-stable
---
## Refactor
https://github.com/irods/irods/blob/2f602874739f3d0d15551127cddbac181fb88e27/lib/core/include/irods/rodsType.h#L24

korydraughn:
NTS: `grep -Ilrn '\<uint\>' | xargs sed -i 's/\<uint\>/unsigned int/'`

Need to track down where `uint` is declared.

korydraughn:
On my computer, clangd reports the declaration at `/usr/include/x86_64-linux-gnu/sys/types.h`.
```c
146 #ifdef __USE_MISC
147 /* Old compatibility names for C types.  */
148 typedef unsigned long int ulong;
149 typedef unsigned short int ushort;
150 typedef unsigned int uint;
151 #endif
```

PR #8646, opened 2025-08-28
[#6756] Deprecate and replace use of uint (4-3-stable)

korydraughn:
Code compiles, but has not been tested.

This needs to be tested on all platforms we support before merging.

### DIFF
--- a/lib/api/include/irods/getMiscSvrInfo.h
+++ b/lib/api/include/irods/getMiscSvrInfo.h
@@ -12,7 +12,7 @@
 
 typedef struct MiscSvrInfo {
     int serverType;     // RCAT_ENABLED or RCAT_NOT_ENABLED
-    uint serverBootTime;
+    unsigned int serverBootTime;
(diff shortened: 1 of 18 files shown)


alanking (commented):
Looks good to me. Awaiting testing results.

korydraughn:
Limiting this change to the main branch since it deprecates a type and we cannot introduce deprecations into the 5-0-stable branch.

korydraughn:
Will close once I have the PR up for the main branch.

alanking:
> we cannot introduce deprecations into the 5-0-stable branch

But this is 4-3-stable?

korydraughn:
Correct. I'm looking at the weird case of 4-3-stable having a deprecation and 5-0-stable not having it.

We could move forward with this change, but I figured it wouldn't be worth it since we wouldn't be able to remove the deprecated type alias until iRODS 6.

alanking:
Understood

Issue #8577, opened 2025-07-10
unattended setup always prompts for HOST in 5.0.1


kript:
- [x] main - Do not prompt on unattended install
- [x] 5-0-stable - Do not prompt on unattended install
---
### iRODS Version, OS and Version

iRODS 5.0.1
Ubuntu 24.04

### What did you try to do?

Use the unattended installation argument to the setup script

### Expected behavior

No user input required. All information the server needs to build its config files contained in the JSON file

### Observed Behavior (including steps to reproduce, if applicable)

```bash
root@irods-test:~# hostname
irods-test.metadata.school
root@irods-test:~# 
```

In both cases, the script prompts for the hostname, even though it is contained in the `irods_host` entry in the JSON file. Once provided, the script then continues.

#### with an `=`

```bash
root@irods-test:~# python3 /var/lib/irods/scripts/setup_irods.py --json_configuration_file=unattended.json 
iRODS FQDN, hostname, or IP (253 characters max) [irods-test.metadata.school]: 
```

#### without an `=`

```bash
root@irods-test:~# python3 /var/lib/irods/scripts/setup_irods.py --json_configuration_file unattended.json 
iRODS FQDN, hostname, or IP (253 characters max) [irods-test.metadata.school]: 
```

#### unattended.json file

```json
{
    "admin_password": "rods",
    "default_resource_directory": "/var/lib/irods/Vault",
    "default_resource_name": "demoResc",
    "host_system_information": {
        "service_account_user_name": "irods",
        "service_account_group_name": "irods"
    },
    "service_account_environment": {
        "irods_client_server_policy": "CS_NEG_REFUSE",
        "irods_connection_pool_refresh_time_in_seconds": 300,
        "irods_cwd": "/tempZone/home/rods",
        "irods_default_hash_scheme": "SHA256",
        "irods_default_number_of_transfer_threads": 4,
        "irods_default_resource": "demoResc",
        "irods_encryption_algorithm": "AES-256-CBC",
        "irods_encryption_key_size": 32,
        "irods_encryption_num_hash_rounds": 16,
        "irods_encryption_salt_size": 8,
        "irods_home": "/tempZone/home/rods",
        "irods_host": "irods-test.metadata.school",
        "irods_match_hash_policy": "compatible",
        "irods_maximum_size_for_single_buffer_in_megabytes": 32,
        "irods_port": 1247,
        "irods_transfer_buffer_size_for_parallel_transfer_in_megabytes": 4,
[...]

PR #8650, opened 2025-08-29
[#8577] Do not prompt for server host on unattended install (main)

korydraughn:
This code change comes directly from the issue. It skips the prompt when unattended installs are performed. I don't think we need to test this since the change is very obvious.

I'm much more interested in adding support for unattended installation to the testing environment and catching issues during release testing.


alanking:
See also: https://github.com/irods/irods_testing_environment/issues/232

alanking (approved):
\# it

korydraughn:
Pounded.

PR #8654, opened 2025-09-02
[#6756] Deprecate and replace use of uint type alias (main)

korydraughn:
Cherry-picked from #8646.

Code compiles, but has not been tested.

Will test against all OS to be sure things are still working.


alanking (approved):
Looks good. # if tests come back good

Issue #8635, opened 2025-08-26
Remove unused `packaging/generate_checksums.sh`


alanking:
The script `packaging/generate_checksums.sh` appears to be unused. If so, let's get rid of it.

alanking:
`generate_checksums.sh` is not even packaged, so should be safe to remove without breaking backwards compatibility.

trel:
agreed.   it was used to generate output to accompany engineering builds.

trel:
originally for https://github.com/irods/irods/issues/1297

Issue #3403, opened 2016-11-19
davrods+PAM+SSLFIPS equals no-go


bnordgren:
I'm putting this here instead of the davrods project because it looks to me like davrods uses the irods runtime library for authentication. 

DAV clients cannot connect to Apache running in FIPS compliant mode (`SSLFIPS on`). Immediately after successful authentication, davrods attempts to do something with MD5, resulting in a notation in /var/log/httpd/error_log: 

    md5_dgst.c(80): OpenSSL internal error, assertion failed: Digest MD5 forbidden in FIPS mode!
    [Sat Nov 19 14:37:17.855131 2016] [core:notice] [pid 12491] AH00052: child pid 12493 exit signal Aborted (6)

The rodslog contains: 

    Nov 19 15:44:49 pid:13319 NOTICE: Agent process 14019 started for puser=bnordgren and cuser=bnordgren from 127.0.0.1
    Authenticated
    Nov 19 15:44:50 pid:14019 ERROR: [-]    iRODS/server/core/src/rsApiHandler.cpp:520:readAndProcClientMsg :  status [SYS_HEADER_READ_LEN_ERR]  errno [] -- message []
        [-]     iRODS/lib/core/src/sockComm.cpp:199:readMsgHeader :  status [SYS_HEADER_READ_LEN_ERR]  errno [] -- message [failed to call 'read header']
                [-]     libssl.cpp:577:ssl_read_msg_header :  status [SYS_HEADER_READ_LEN_ERR]  errno [] -- message [read 0 expected 4]

    Nov 19 15:44:50 pid:14019 ERROR: [-]    iRODS/server/core/src/rodsAgent.cpp:256:main :  status [SSL_SHUTDOWN_ERROR]  errno [] -- message []
        [-]     iRODS/lib/core/src/sockComm.cpp:162:sockAgentStop :  status [SSL_SHUTDOWN_ERROR]  errno [] -- message [failed to call 'agent stop']
                [-]     libssl.cpp:945:ssl_agent_stop :  status [SSL_SHUTDOWN_ERROR]  errno [] -- message [error completing shutdown of SSL connection]

    Nov 19 15:44:50 pid:13319 NOTICE: Agent process 14019 exited with status 16384

It looks like it is not possible to use iRODS in FIPS compliant mode?

trel:
Also has come up in a recent iRODS-Chat discussion:
   https://groups.google.com/forum/#!topic/irod-chat/BwxZLB8xwr0

[...]

PR #8655, opened 2025-09-02
Replace MD5 for rule hashing and mockarchive (main)

alanking:
In service of #3403 
Addresses #8635

This is the first in a series of pull requests to address the various pieces of #3403. I split them up for ease of reviewing. This one is really small and should have little or no impact on existing deployments. I am going to leave it in draft while I investigate the potential impacts, but my current assessment is that this should not have an impact.

Unit tests passed. Core tests are running now.

### DIFF
--- a/plugins/resources/src/mockarchive.cpp
+++ b/plugins/resources/src/mockarchive.cpp
@@ -41,7 +41,6 @@
 #include <dirent.h>
 #include <openssl/err.h>
 #include <openssl/evp.h>
-#include <openssl/md5.h>
 #if defined(osx_platform)
 #include <sys/malloc.h>
 #else
@@ -173,7 +172,8 @@ irods::error make_hashed_path(
 
     // Initialize the digest context for the derived digest implementation. Must use _ex or _ex2 to avoid
     // automatically resetting the context with EVP_MD_CTX_reset.
-    EVP_MD* message_digest = EVP_MD_fetch(nullptr, "MD5", nullptr);
+    static constexpr const char* algorithm_name = "SHA256";
+    EVP_MD* message_digest = EVP_MD_fetch(nullptr, algorithm_name, nullptr);
     EVP_MD_CTX* context = EVP_MD_CTX_new();
     const auto free_context = irods::at_scope_exit{[&context] { EVP_MD_CTX_free(context); }};
     if (0 == EVP_DigestInit_ex2(context, message_digest, nullptr)) {
@@ -184,11 +184,11 @@ irods::error make_hashed_path(
     }
     EVP_MD_free(message_digest);
 
-    char md5Buf[MAX_NAME_LEN]{};
-    std::strncpy(md5Buf, _path.c_str(), _path.size());
+    char buffer_to_hash[MAX_NAME_LEN]{};
+    std::strncpy(buffer_to_hash, _path.c_str(), _path.size());
 
     // Hash the specified buffer of bytes and store the results in the context.
-    if (0 == EVP_DigestUpdate(context, reinterpret_cast<unsigned char*>(md5Buf), _path.size())) {
+    if (0 == EVP_DigestUpdate(context, reinterpret_cast<unsigned char*>(buffer_to_hash), _path.size())) {
         const auto ssl_error_code = ERR_get_error();
         const auto msg = fmt::format("{}: Failed to calculate digest. error code: [{}]", __func__, ssl_error_code);
         THROW(DIGEST_UPDATE_FAILED, msg);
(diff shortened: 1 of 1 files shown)


server/icat/include/irods/icatHighLevelRoutines.hpp
```diff
@@ -60,8 +60,7 @@ int chlDelColl( rsComm_t *rsComm, collInfo_t *collInfo );
 int chlCheckAuth( rsComm_t *rsComm, const char* scheme, const char *challenge, const char *response,
                   const char *username, int *userPrivLevel, int *clientPrivLevel );
 int chlMakeTempPw( rsComm_t *rsComm, char *pwValueToHash, const char *otherUser );
-int chlMakeLimitedPw( rsComm_t *rsComm, int ttl, char *pwValueToHash );
-int decodePw( rsComm_t *rsComm, const char *in, char *out );
+int chlMakeLimitedPw(rsComm_t* rsComm, int ttl, char* pwValueToHash);
```
trel, line 63: can we just remove the one line here?     and do any reformatting separately?

plugins/resources/src/mockarchive.cpp
```diff
@@ -184,19 +184,20 @@ irods::error make_hashed_path(
...
+    if (0 == EVP_DigestUpdate(context, reinterpret_cast<unsigned char*>(buffer_to_hash), _path.size())) {
         const auto ssl_error_code = ERR_get_error();
         const auto msg = fmt::format("{}: Failed to calculate digest. error code: [{}]", __func__, ssl_error_code);
         THROW(DIGEST_UPDATE_FAILED, msg);
     }
 
     // Finally, retrieve the digest value from the context and place it into a buffer. Use the _ex function here
     // so that the digest context is not automatically cleaned up with EVP_MD_CTX_reset.
-    unsigned char hash[MAX_NAME_LEN];
+    static constexpr auto max_digest_length = EVP_MAX_MD_SIZE;
```
trel, line 199: so this length is now longer than MAX_NAME_LEN?

trel (commented):
surgical.

plugins/resources/src/mockarchive.cpp
korydraughn: Looks like the length doesn't matter much. Just below this, we can see that only the first 16 bytes of the hash are used.

alanking: Hmm... that's a good point. That 16 comes from the length of MD5 hashes, which is 128 bits, or 16 bytes. SHA256 hashes are 256 bits, or 32 bytes. We should probably update that (unless, of course, we don't really care that much about mockarchive's correctness on this front).

The new length is using `EVP_MAX_MD_SIZE` instead of 32 bytes because if we ever decide to move away from SHA256 in the future, we know that `EVP_DigestFinal` outputs *at most* `EVP_MAX_MD_SIZE` bytes, regardless of the hashing algorithm being used.

server/icat/include/irods/icatHighLevelRoutines.hpp
alanking: Sure, I'll try to split these up in a rebase.

plugins/resources/src/mockarchive.cpp
trel: seems good to keep them correct/consistent - for the next reader to not question everything again.

server/icat/include/irods/icatHighLevelRoutines.hpp
alanking: Done

plugins/resources/src/mockarchive.cpp
alanking: Done

alanking:
Core tests passed. Marking ready for review.

korydraughn:
If the mockarchive resource is for testing purposes only, then I agree that it's fine to change.

Do we have that documented somewhere?

alanking:
> If the mockarchive resource is for testing purposes only, then I agree that it's fine to change.
> 
> Do we have that documented somewhere?

The docs (https://docs.irods.org/5.0.1/plugins/composable_resources/#mock-archive) say...

> The mock archive storage resource was created mainly for testing purposes to emulate the behavior of object stores (e.g. WOS). It creates a hash of the file path as the physical name of the Data Object.

It does not specify the nature of the hash and states that it is "mainly for testing purposes". Thoughts?

korydraughn:
Good. We should lean into that more and assert that it should not be used in production environments.

alanking:
> Are the tests still passing with the change to the hash length?


To be honest, I only ran the mockarchive suite after making that change. I will run the whole suite again just in case.

alanking:
Core tests passed.

Should we hold off on the mockarchive part of this change until we confirm that no members are using it at the Technology Working Group?

trel:
fair to wait until Monday.   but this is definitely not being used by anyone.

korydraughn:
Fine with me. Will include it as a topic for discussion.

alanking:
The question was presented at the Technology Working Group this morning and there are no known uses of mockarchive in the deployments of our members. I created https://github.com/irods/irods/issues/8663 to document the resource type as only for testing.

korydraughn:
Good. Is this PR ready or is there more work to be done?

alanking:
I'm happy with it in its current state.

korydraughn:
Squash to taste.

alanking:
I think the commits are where I want them to be. Please let me know if you'd like any changes

alanking:
#'d, mergin

PR #8665, opened 2025-09-08
[#8546] GenQuery2: Fix ability to find objects with no user permissions (5-0-stable)

korydraughn:
Cherry-picked from #8642.


korydraughn:
Added pounds.

Issue #8653, opened 2025-08-29
`imiscsvrinfo` segfaults when talking to a server earlier than 4.3.4


korydraughn:
- [ ] main
- [ ] 5-0-stable
- [ ] 4-3-stable
---
## Bug Report

This is likely caused by the following line.

https://github.com/irods/irods_client_icommands/blob/84d053292251e4be7dd727cb2b2ba42b62af11b6/src/imiscsvrinfo.cpp#L90

The packing instructions changed in 4.3.4 making the icommand incompatible with older versions of iRODS. Notice `certinfo`.

https://github.com/irods/irods/blob/3f6524069c0eac458fc26808b7095a414581a28e/lib/api/include/irods/getMiscSvrInfo.h#L19-L24

This can be fixed by using a custom/compatible packing instruction and `procApiRequest` directly. Whether we do that will depend on the outcome of #8638.

_Originally reported in #8638._

tsmeele:
May i recommend that existing packing instructions are never changed. Instead one could simply add a new packing instruction that extends the prior one.  This method allows clients linked with earlier iRODS client API library to detect that they lack support for a newer data structure being sent, rather than crash because a "known" data structure seemingly mishaves.

korydraughn:
That is a possibility. Now that version numbers communicate more info, we can communicate breaking changes much better. We also have a much better understanding of when a change to a packing instruction can break backward compatibility.

We're currently considering the idea of limiting changes to packing instructions to major releases since that aligns with semver rules. Again, this is something that's in discussion.

korydraughn:
Related to irods/irods_docs#388.

korydraughn:
For the 5-0-stable branch, the icommand detects the situation and reports an error.

The other branches will include fixes which restore the ability to communicate with 4.3.3.

PR #8672, opened 2025-09-15
[#8653] Allow override of packing instructions for API requests (main)

korydraughn:
Companion PR at https://github.com/irods/irods_client_icommands/pull/590.

This PR adds a new version of `procApiRequest` to fix the segfault bug in `imiscsvrinfo`.

This change cannot be part of the 5-0-stable branch because it results in a change to the public API. It can/will be included in the 4.3.5 release though.

There's no need to run the core tests for this since the function is not being used outside of the unit tests.

The most important thing for this work is the documentation and function signature. As long as we get those correct, fixing any bugs within the implementation should be possible.

### DIFF
--- a/lib/core/include/irods/procApiRequest.h
+++ b/lib/core/include/irods/procApiRequest.h
@@ -1,18 +1,172 @@
-#ifndef PROC_API_REQUEST_H__
-#define PROC_API_REQUEST_H__
+#ifndef IRODS_PROC_API_REQUEST_H
+#define IRODS_PROC_API_REQUEST_H
 
+/// \file
+
+#include "irods/packStruct.h"
 #include "irods/rods.h"
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
+/// Executes an iRODS API request.
+///
+/// This function uses the global packing instruction table for serialization and deserialization.
+///
+/// \param[in]  conn        The RcComm used for communication.
+/// \param[in]  apiNumber   The integer which identifies the API to execute. See apiNumberData.h.
+/// \param[in]  inputStruct The data to serialize, typically an input data structure. Can be
+///                         passed \p NULL if no input is necessary.
+/// \param[in]  inputBsBBuf The byte buffer to stream to the server. Can be passed \p NULL if no
+///                         input byte stream is necessary.
+/// \param[out] outStruct   The output data structure to be filled by the server. The output will
+///                         be heap-allocated and assigned to the pointer. Can be passed \p NULL
+///                         if no output is expected.
+/// \param[out] outBsBBuf   The output byte buffer to be filled by the server. Can be passed
+///                         \p NULL if no output stream is expected.
+///
+/// \return An integer indicating success or failure.
+/// \retval >=0 On success.
+/// \retval  <0 On failure.
+///
+/// \b Example
+/// \code{.c}
+/// rcComm_t* conn = // Our iRODS connection.
+///
+/// gridConfigurationInp_t input;
+/// memset(&input, 0, sizeof(gridConfigurationInp_t));
+/// strcpy(input.name_space, "delay_server");
+/// strcpy(input.option_name, "leader");
+///
+/// gridConfigurationOut_t* output = NULL;
+///
+/// int ec = procApiRequest(conn, GET_GRID_CONFIGURATION_VALUE_APN, &input, NULL, (void**) &output, NULL);
+/// if (ec < 0) {
+///     // Handle error (and deallocate memory if needed).
+///     // Make sure to consult the documentation for each API to determine when memory
+///     // needs to be deallocated. Not all APIs require deallocation on failure.
+/// }
+///
+/// // Use the results and deallocate memory.
+///
+/// free(output);
+/// \endcode
 int procApiRequest(rcComm_t *conn,
                    int apiNumber,
                    const void *inputStruct,
                    const bytesBuf_t *inputBsBBuf,
                    void **outStruct,
                    bytesBuf_t *outBsBBuf);
+
+/// Executes an iRODS API request.
+///
+/// Grants the caller control over which packing instructions are used during serialization and
+/// deserialization of the data structures. This form of procApiRequest is useful for maintaining
+/// backward/forward compatibility.
+///
+/// \warning This function is provided for handling changes in packing instructions for an API
+/// only. All other properties of the original API must be honored.
+///
+/// \param[in]  conn        The RcComm used for communication.
+/// \param[in]  apiNumber   The integer which identifies the API to execute. See apiNumberData.h.
+/// \param[in]  packingInstructionTable The table holding various packing instructions. All
+///                         packing instructions needed for serialization of the data structure
+///                         must be defined in the table. If passed \p NULL, the global packing
+///                         instruction table will be used.
+/// \param[in]  inputPackingInstruction The packing instruction to use for serialization. Cannot
+///                         be \p NULL.
+/// \param[in]  inputStruct The data to serialize, typically an input data structure. Can be
+///                         passed \p NULL if no input is necessary.
+/// \param[in]  inputBsBBuf The byte buffer to stream to the server. Can be passed \p NULL if no
+///                         input byte stream is necessary.
+/// \param[in]  outputPackingInstruction The packing instruction to use for deserialization.
+///                         Can be set to \p NULL if no output is expected.
+/// \param[out] outStruct   The output data structure to be filled by the server. The output will
+///                         be heap-allocated and assigned to the pointer. Can be passed \p NULL
+///                         if no output is expected.
+/// \param[out] outBsBBuf   The output byte buffer to be filled by the server. Can be passed
+///                         \p NULL if no output stream is expected.
+///
+/// \return An integer indicating success or failure.
+/// \retval >=0 On success.
+/// \retval  <0 On failure.
+///
+/// \b Example
+/// \code{.c}
+/// // This example produces the same results as the procApiRequest() example.
+///
+/// // Define our custom input structure and packing instruction definitions.
+/// // Mirrors gridConfigurationInp_t.
+/// typedef struct my_input_struct
+/// {
+///     char name_space[2700];
+///     char option_name[2700];
+///     char option_value[2700];
+/// } my_input_struct_t;
+///
+/// const char* my_input_struct_pi_name = "my_input_struct_PI";
+/// const char* my_input_struct_pi_instruction = "str name_space[2700]; str option_name[2700]; str option_value[2700];";
+///
+/// // Define our custom output structure and packing instruction definitions.
+/// // Mirrors gridConfigurationOut_t.
+/// typedef struct my_output_struct
+/// {
+///     char option_value[2700];
+/// } my_output_struct_t;
+///
+/// const char* my_output_struct_pi_name = "my_output_struct_PI";
+/// const char* my_output_struct_pi_instruction = "str option_value[2700];";
+///
+/// // Define our custom packing instruction table for serialization and deserialization.
+/// const packInstruct_t pi_table[] = {
+///     {my_input_struct_pi_name, my_input_struct_pi_instruction, NULL},
+///     {my_output_struct_pi_name, my_output_struct_pi_instruction, NULL},
+///     {PACK_TABLE_END_PI, NULL, NULL}
+/// };
+///
+/// // Use the custom packing instructions to fetch information from the R_GRID_CONFIGURATION
+/// // table.
+///
+/// rcComm_t* conn = // Our iRODS connection.
+///
+/// my_input_struct_t input;
+/// memset(&input, 0, sizeof(my_input_struct_t));
+/// strcpy(input.name_space, "delay_server");
+/// strcpy(input.option_name, "leader");
+///
+/// my_output_struct_t* output = NULL;
+///
+/// int ec = procApiRequest_raw(conn,
+///                             GET_GRID_CONFIGURATION_VALUE_APN,
+///                             pi_table,
+///                             my_input_struct_pi_name,
+///                             &input,
+///                             NULL,
+///                             my_output_struct_pi_name,
+///                             (void**) &output,
+///                             NULL);
+/// if (ec < 0) {
+///     // Handle error (and deallocate memory if needed).
+///     // Make sure to consult the documentation for each API to determine when memory
+///     // needs to be deallocated. Not all APIs require deallocation on failure.
+/// }
+///
+/// // Use the results and deallocate memory.
+///
+/// free(output);
+/// \endcode
+///
+/// \since 4.3.5
+int procApiRequest_raw(rcComm_t* conn,
+                       int apiNumber,
+                       const struct PackingInstruction* packingInstructionTable,
+                       const char* inputPackingInstruction,
+                       const void* inputStruct,
+                       const bytesBuf_t* inputBsBBuf,
+                       const char* outputPackingInstruction,
+                       void** outStruct,
+                       bytesBuf_t* outBsBBuf);
 
 int sendApiRequest(rcComm_t *conn,
                    int apiInx,
@@ -48,7 +202,7 @@ int _cliGetCollOprStat(rcComm_t *conn, collOprStat_t **collOprStat);
 int apiTableLookup(int apiNumber);
 
 #ifdef __cplusplus
-}
+} // extern "C"
 #endif
 
-#endif	// PROC_API_REQUEST_H__
+#endif // IRODS_PROC_API_REQUEST_H
--- a/lib/core/src/procApiRequest.cpp
+++ b/lib/core/src/procApiRequest.cpp
@@ -20,23 +20,299 @@
 
 #include "irods/irods_threads.hpp"
 
-/* procApiRequest - This is the main function used by the client API
- * function to issue API requests and receive output returned from
- * the server.
- * rcComm_t *conn - the client communication handle.
- * int apiNumber - the API number of this call defined in apiNumber.h.
- * void *inputStruct - pointer to the input struct of this API. If there
- *     is no input struct, a NULL should be entered
- * bytesBuf_t *inputBsBBuf - pointer to the input byte stream. If there
- *     is no input byte stream, a NULL should be entered
- * void **outStruct - Pointer to pointer to the output struct. The outStruct
- *     will be allocated by this function and the pointer to this struct is
- *     passed back to the caller through this pointer. If there
- *     is no output struct, a NULL should be entered
- * bytesBuf_t *outBsBBuf - pointer to the output byte stream. If there
- *     is no output byte stream, a NULL should be entered
- *
- */
+namespace
+{
+    auto send_api_request(rcComm_t* conn,
+                          int apiInx,
+                          const PackingInstruction* packingInstructionTable,
+                          const char* packingInstruction,
+                          const void* inputStruct,
+                          const bytesBuf_t* inputBsBBuf) -> int
+    {
+        int status = 0;
+        bytesBuf_t* inputStructBBuf = nullptr;
+        bytesBuf_t* myInputStructBBuf = nullptr;
+
+        cliChkReconnAtSendStart(conn);
+
+        irods::api_entry_table& RcApiTable = irods::get_client_api_table();
+        auto itr = RcApiTable.find(apiInx);
+
+        if (itr == RcApiTable.end()) {
+            rodsLog(LOG_ERROR, "%s: Could not find API entry matching at index [%d].", __func__, apiInx);
+            return SYS_UNMATCHED_API_NUM;
+        }
+
+        if (packingInstruction) {
+            if (inputStruct == nullptr) {
+                cliChkReconnAtSendEnd(conn);
+                return USER_API_INPUT_ERR;
+            }
+            status = pack_struct(inputStruct,
+                                 &inputStructBBuf,
+                                 packingInstruction,
+                                 packingInstructionTable,
+                                 0,
+                                 conn->irodsProt,
+                                 conn->svrVersion->relVersion);
+            if (status < 0) {
+                rodsLog(LOG_ERROR, "%s: Packing instruction error; error_code=[%d]", __func__, status);
+                cliChkReconnAtSendEnd(conn);
+                return status;
+            }
+
+            myInputStructBBuf = inputStructBBuf;
+        }
+
+        if (itr->second->inBsFlag <= 0) {
+            inputBsBBuf = nullptr;
+        }
+
+        irods::network_object_ptr net_obj;
+        irods::error ret = irods::network_factory(conn, net_obj);
+        if (!ret.ok()) {
+            freeBBuf(inputStructBBuf);
+            rodsLog(LOG_ERROR,
+                    "%s: network_factory error: [%s], error_code=[%d]",
+                    __func__,
+                    ret.user_result().c_str(),
+                    ret.code());
+            return static_cast<int>(ret.code());
+        }
+
+        ret = sendRodsMsg(
+            net_obj, RODS_API_REQ_T, myInputStructBBuf, inputBsBBuf, nullptr, itr->second->apiNumber, conn->irodsProt);
+        if (!ret.ok()) {
+            rodsLog(LOG_ERROR, "%s: sendRodsMsg error: [%d].", __func__, ret.code());
+            if (conn->svrVersion != nullptr && conn->svrVersion->reconnPort > 0) {
+                const auto savedStatus = static_cast<int>(ret.code());
+                conn->thread_ctx->lock->lock();
+                const int status1 = cliSwitchConnect(conn);
+                rodsLog(LOG_ERROR,
+                        "%s: cliSwitchConnect error: [%d]; clientState=[%d], agentState=[%d]",
+                        __func__,
+                        status1,
+                        conn->clientState,
+                        conn->agentState);
+                conn->thread_ctx->lock->unlock();
+                if (status1 > 0) {
+                    // Should not be here.
+                    rodsLog(LOG_ERROR, "%s: Switch connection and retry sendRodsMsg.", __func__);
+                    ret = sendRodsMsg(net_obj,
+                                      RODS_API_REQ_T,
+                                      myInputStructBBuf,
+                                      inputBsBBuf,
+                                      nullptr,
+                                      itr->second->apiNumber,
+                                      conn->irodsProt);
+                    if (!ret.ok()) {
+                        rodsLog(LOG_ERROR, "%s: %s; error_code=[%d].", __func__, ret.user_result().c_str(), ret.code());
+                    }
+                    else {
+                        status = savedStatus;
+                    }
+                } // if status1 > 0
+            } // if svrVersion != nullptr ...
+        }
+        else {
+            // be sure to pass along the return code from the
+            // plugin call
+            status = static_cast<int>(ret.code());
+        }
+
+        freeBBuf(inputStructBBuf);
+
+        return status;
+    } // send_api_request
+
+    int procApiReply_raw(rcComm_t* conn,
+                         int apiInx,
+                         const PackingInstruction* packingInstructionTable,
+                         const char* outputPackingInstruction,
+                         void** outStruct,
+                         bytesBuf_t* outBsBBuf,
+                         msgHeader_t* myHeader,
+                         bytesBuf_t* outStructBBuf,
+                         bytesBuf_t* myOutBsBBuf,
+                         bytesBuf_t* errorBBuf)
+    {
+        if (errorBBuf->len > 0) {
+            const int status = unpack_struct(errorBBuf->buf,
+                                             static_cast<void**>(static_cast<void*>(&conn->rError)),
+                                             "RError_PI",
+                                             RodsPackTable,
+                                             conn->irodsProt,
+                                             conn->svrVersion->relVersion);
+            if (status < 0) {
+                rodsLogError(LOG_ERROR, status, "readAndProcApiReply:unpack_struct error. status = %d", status);
+            }
+        }
+
+        const int retVal = myHeader->intInfo;
+
+        /* some sanity check */
+
+        irods::api_entry_table& RcApiTable = irods::get_client_api_table();
+        if (RcApiTable[apiInx]->outPackInstruct != nullptr && outStruct == nullptr) {
+            rodsLog(
+                LOG_ERROR, "readAndProcApiReply: outStruct error for C apiNumber %d", RcApiTable[apiInx]->apiNumber);
+
+            if (retVal < 0) {
+                return retVal;
+            }
+
+            return USER_API_INPUT_ERR;
+        }
+
+        if (RcApiTable[apiInx]->outBsFlag > 0 && outBsBBuf == nullptr) {
+            rodsLog(
+                LOG_ERROR, "readAndProcApiReply: outBsBBuf error for D apiNumber %d", RcApiTable[apiInx]->apiNumber);
+            if (retVal < 0) {
+                return retVal;
+            }
+
+            return USER_API_INPUT_ERR;
+        }
+
+        /* handle outStruct */
+        if (outStructBBuf->len > 0) {
+            if (outStruct != nullptr) {
+                const int status = unpack_struct(outStructBBuf->buf,
+                                                 outStruct,
+                                                 outputPackingInstruction,
+                                                 packingInstructionTable,
+                                                 conn->irodsProt,
+                                                 conn->svrVersion->relVersion);
+                if (status < 0) {
+                    rodsLogError(LOG_ERROR, status, "readAndProcApiReply:unpack_struct error. status = %d", status);
+
+                    if (retVal < 0) {
+                        return retVal;
+                    }
+
+                    return status;
+                }
+            }
+            else {
+                rodsLog(LOG_ERROR,
+                        "readAndProcApiReply: got unneeded outStruct for apiNumber %d",
+                        RcApiTable[apiInx]->apiNumber);
+            }
+        }
+
+        if (myOutBsBBuf != nullptr && myOutBsBBuf->len > 0) {
+            if (outBsBBuf != nullptr) {
+                /* copy to out */
+                *outBsBBuf = *myOutBsBBuf;
+                memset(myOutBsBBuf, 0, sizeof(bytesBuf_t));
+            }
+            else {
+                rodsLog(LOG_ERROR,
+                        "readAndProcApiReply: got unneeded outBsBBuf for apiNumber %d",
+                        RcApiTable[apiInx]->apiNumber);
+            }
+        }
+
+        return retVal;
+    } // procApiReply_raw
+
+    int readAndProcApiReply_raw(rcComm_t* conn,
+                                int apiInx,
+                                const PackingInstruction* packingInstructionTable,
+                                const char* outputPackingInstruction,
+                                void** outStruct,
+                                bytesBuf_t* outBsBBuf)
+    {
+        int status = 0;
+        msgHeader_t myHeader;
+
+        bytesBuf_t outStructBBuf{};
+        bytesBuf_t errorBBuf{};
+
+        cliChkReconnAtReadStart(conn);
+
+        irods::api_entry_table& RcApiTable = irods::get_client_api_table();
+        if (outputPackingInstruction != nullptr && outStruct == nullptr) {
+            rodsLog(LOG_ERROR, "%s: outStruct error for A apiNumber %d", __func__, RcApiTable[apiInx]->apiNumber);
+            cliChkReconnAtReadEnd(conn);
+            return USER_API_INPUT_ERR;
+        }
+
+        if (RcApiTable[apiInx]->outBsFlag > 0 && outBsBBuf == nullptr) {
+            rodsLog(LOG_ERROR, "%s: outBsBBuf error for B apiNumber %d", __func__, RcApiTable[apiInx]->apiNumber);
+            cliChkReconnAtReadEnd(conn);
+            return USER_API_INPUT_ERR;
+        }
+
+        irods::network_object_ptr net_obj;
+        irods::error ret = irods::network_factory(conn, net_obj);
+        if (!ret.ok()) {
+            irods::log(PASS(ret));
+            return static_cast<int>(ret.code());
+        }
+
+        ret = readMsgHeader(net_obj, &myHeader, nullptr);
+        if (!ret.ok()) {
+#ifdef RODS_CLERVER
+            irods::log(PASS(ret));
+#else
+            if (ret.code() == SYS_HEADER_READ_LEN_ERR) {
+                ret = CODE(SYS_INTERNAL_ERR);
+            }
+#endif
+            if (conn->svrVersion != nullptr && conn->svrVersion->reconnPort > 0) {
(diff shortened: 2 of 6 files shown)


lib/core/include/irods/procApiRequest.h
```diff
@@ -1,19 +1,88 @@
...
 
+/// Executes an iRODS API request.
+///
+/// This function uses the global packing instruction table for serialization and deserialization.
+///
+/// \param[in]  conn        The RcComm used for communication.
+/// \param[in]  apiNumber   The integer which identifies the API to execute. See apiNumberData.h.
+/// \param[in]  inputStruct \parblock The data to serialize, typically an input data structure.
+///                         Can be passed \p NULL if no input is necessary. \endparblock
+/// \param[in]  inputBsBBuf The byte buffer to stream to the server.
```
trel:
```suggestion
/// \param[in]  inputBsBBuf \parblock The byte buffer to stream to the server.
```

lib/core/include/irods/procApiRequest.h
```diff
@@ -1,19 +1,88 @@
...
 int procApiRequest(rcComm_t *conn,
                    int apiNumber,
                    const void *inputStruct,
                    const bytesBuf_t *inputBsBBuf,
                    void **outStruct,
                    bytesBuf_t *outBsBBuf);
 
+/// Executes an iRODS API request.
+///
+/// Grants the caller control over which packing instructions are used during serializing and
```
trel:
```suggestion
/// Grants the caller control over which packing instructions are used during serialization and
```

lib/core/include/irods/procApiRequest.h
```diff
@@ -1,19 +1,88 @@
...
+/// deserialization of the data structures. This form of procApiRequest is useful for maintaining
+/// backward compartibility.
```
trel:
```suggestion
/// backward compatibility.
```

lib/core/include/irods/procApiRequest.h
```diff
@@ -1,19 +1,88 @@
...
+/// \param[in]  apiNumber   The integer which identifies the API to execute. See apiNumberData.h.
+/// \param[in]  packingInstructionTable \parblock The table holding various packing instructions.
+///                         All packing instructions needed for serialization of the
+///                         data structure must be defined in the table. If passed \p NULL, the
+///                         global packing instruction table will be used. \endparblock
+/// \param[in]  inputPackingInstruction \parblock The packing instruction to use for serialization.
+///                         Cannot be \p NULL. \endparblock
+/// \param[in]  inputStruct \parblock The data to serialize, typically an input data structure.
+///                         Can be passed \p NULL if no input is necessary. \endparblock
+/// \param[in]  inputBsBBuf The byte buffer to stream to the server.
```
trel:
```suggestion
/// \param[in]  inputBsBBuf \parblock The byte buffer to stream to the server.
```

lib/core/src/procApiRequest.cpp
```diff
@@ -20,23 +20,316 @@
...
+
+        if (itr->second->inBsFlag <= 0) {
+            inputBsBBuf = nullptr;
+        }
+
+        irods::network_object_ptr net_obj;
+        irods::error ret = irods::network_factory(conn, net_obj);
+        if (!ret.ok()) {
+            freeBBuf(inputStructBBuf);
+            rodsLog(LOG_ERROR,
```
trel, line 75: log_api?

lib/core/src/procApiRequest.cpp
```diff
@@ -20,23 +20,316 @@
...
+        bytesBuf_t* inputStructBBuf = nullptr;
+        bytesBuf_t* myInputStructBBuf = nullptr;
+
+        cliChkReconnAtSendStart(conn);
+
+        irods::api_entry_table& RcApiTable = irods::get_client_api_table();
+        auto itr = RcApiTable.find(apiInx);
+
+        if (itr == RcApiTable.end()) {
+            rodsLog(LOG_ERROR, "%s: Could not find API entry matching at index [%d].", __func__, apiInx);
```
trel, line 42: log_api?

lib/core/src/procApiRequest.cpp
```diff
@@ -20,23 +20,316 @@
...
+
+        ret = sendRodsMsg(
+            net_obj, RODS_API_REQ_T, myInputStructBBuf, inputBsBBuf, nullptr, itr->second->apiNumber, conn->irodsProt);
+        if (!ret.ok()) {
+            rodsLog(LOG_ERROR, "%s: sendRodsMsg error: [%d].", __func__, ret.code());
+            if (conn->svrVersion != nullptr && conn->svrVersion->reconnPort > 0) {
+                const auto savedStatus = static_cast<int>(ret.code());
+                conn->thread_ctx->lock->lock();
+                int status1 = cliSwitchConnect(conn);
+                rodsLog(LOG_ERROR,
```
trel, line 91: log_api?

lib/core/src/procApiRequest.cpp
```diff
@@ -20,23 +20,316 @@
...
+                        "%s: cliSwitchConnect error: [%d]; clientState=[%d], agentState=[%d]",
+                        __func__,
+                        status1,
+                        conn->clientState,
+                        conn->agentState);
+                conn->thread_ctx->lock->unlock();
+                if (status1 > 0) {
+                    // Should not be here.
+                    rodsLog(LOG_ERROR, "%s: Switch connection and retry sendRodsMsg.", __func__);
```
trel, line 100: log_api?

lib/core/src/procApiRequest.cpp
```diff
@@ -20,23 +20,316 @@
...
+            }
+            status = pack_struct(inputStruct,
+                                 &inputStructBBuf,
+                                 _packingInstruction,
+                                 _packingInstructionTable,
+                                 0,
+                                 conn->irodsProt,
+                                 conn->svrVersion->relVersion);
+            if (status < 0) {
+                rodsLog(LOG_ERROR, "%s: Packing instruction error; error_code=[%d]", __func__, status);
```
trel, line 59: log_api?

lib/core/src/procApiRequest.cpp
```diff
@@ -20,23 +20,316 @@
...
+                    __func__,
+                    ret.user_result().c_str(),
+                    ret.code());
+            return static_cast<int>(ret.code());
+        }
+
+        ret = sendRodsMsg(
+            net_obj, RODS_API_REQ_T, myInputStructBBuf, inputBsBBuf, nullptr, itr->second->apiNumber, conn->irodsProt);
+        if (!ret.ok()) {
+            rodsLog(LOG_ERROR, "%s: sendRodsMsg error: [%d].", __func__, ret.code());
```
trel, line 86: log_api?

lib/core/src/procApiRequest.cpp
```diff
@@ -20,23 +20,316 @@
...
+                    rodsLog(LOG_ERROR, "%s: Switch connection and retry sendRodsMsg.", __func__);
+                    ret = sendRodsMsg(net_obj,
+                                      RODS_API_REQ_T,
+                                      myInputStructBBuf,
+                                      inputBsBBuf,
+                                      nullptr,
+                                      itr->second->apiNumber,
+                                      conn->irodsProt);
+                    if (!ret.ok()) {
+                        rodsLog(LOG_ERROR, "%s: %s; error_code=[%d].", __func__, ret.user_result().c_str(), ret.code());
```
trel, line 109: log_api?

lib/core/src/procApiRequest.cpp
```diff
@@ -20,23 +20,316 @@
...
+    {
+        if (errorBBuf->len > 0) {
+            int status = unpack_struct(errorBBuf->buf,
+                                       static_cast<void**>(static_cast<void*>(&conn->rError)),
+                                       "RError_PI",
+                                       RodsPackTable,
+                                       conn->irodsProt,
+                                       conn->svrVersion->relVersion);
+            if (status < 0) {
+                rodsLogError(LOG_ERROR, status, "readAndProcApiReply:unpack_struct error. status = %d", status);
```
trel, line 147: log_api?

lib/core/src/procApiRequest.cpp
```diff
@@ -20,23 +20,316 @@
...
+            }
+        }
+
+        int retVal = myHeader->intInfo;
+
+        /* some sanity check */
+
+        irods::api_entry_table& RcApiTable = irods::get_client_api_table();
+        if (RcApiTable[apiInx]->outPackInstruct != nullptr && outStruct == nullptr) {
+            rodsLog(
```
trel, line 157: log_api?

lib/core/src/procApiRequest.cpp
```diff
@@ -20,23 +20,316 @@
...
+
+            if (retVal < 0) {
+                return retVal;
+            }
+
+            return USER_API_INPUT_ERR;
+        }
+
+        if (RcApiTable[apiInx]->outBsFlag > 0 && outBsBBuf == nullptr) {
+            rodsLog(
```
trel, line 168: log_api?

lib/core/src/procApiRequest.cpp
```diff
@@ -20,23 +20,316 @@
...
+        if (outStructBBuf->len > 0) {
+            if (outStruct != nullptr) {
+                int status = unpack_struct(outStructBBuf->buf,
+                                           outStruct,
+                                           outputPackingInstruction,
+                                           packingInstructionTable,
+                                           conn->irodsProt,
+                                           conn->svrVersion->relVersion);
+                if (status < 0) {
+                    rodsLogError(LOG_ERROR, status, "readAndProcApiReply:unpack_struct error. status = %d", status);
```
trel, line 187: log_api?

lib/core/src/procApiRequest.cpp
```diff
@@ -20,23 +20,316 @@
...
+
+                    if (retVal < 0) {
+                        return retVal;
+                    }
+
+                    return status;
+                }
+            }
+            else {
+                rodsLog(LOG_ERROR,
```
trel, line 197: log_api?

lib/core/src/procApiRequest.cpp
```diff
@@ -20,23 +20,316 @@
...
+        }
+
+        if (myOutBsBBuf != nullptr && myOutBsBBuf->len > 0) {
+            if (outBsBBuf != nullptr) {
+                /* copy to out */
+                *outBsBBuf = *myOutBsBBuf;
+                memset(myOutBsBBuf, 0, sizeof(bytesBuf_t));
+            }
+            else {
+                rodsLog(LOG_ERROR,
```
trel, line 210: log_api?

lib/core/src/procApiRequest.cpp
```diff
@@ -20,23 +20,316 @@
...
+
+        memset(&outStructBBuf, 0, sizeof(bytesBuf_t));
+        memset(&errorBBuf, 0, sizeof(bytesBuf_t));
+        /* memset (&myOutBsBBuf, 0, sizeof (bytesBuf_t)); */
+
+        /* some sanity check */
+
+        irods::api_entry_table& RcApiTable = irods::get_client_api_table();
+        if (outputPackingInstruction != nullptr && outStruct == nullptr) {
+            rodsLog(LOG_ERROR, "%s: outStruct error for A apiNumber %d", __func__, RcApiTable[apiInx]->apiNumber);
```
trel, line 236: log_api?

lib/core/src/procApiRequest.cpp
```diff
@@ -20,23 +20,316 @@
...
+        }
+
+        // TODO Do I need to require the outBsFlag be passed in too? \
+    // Then again, RcApiTable[apiInx]->outBsFlag is the apidef_t information. \
+    // Changing the packing instruction DOES NOT affect the output byte stream, so I think this is fine as is. \
+    // What if an API didn't originally have an output byte stream? Is it a different API if it does in the future? This feels like a different API. \
+    // Seems we need to define rules around all of this. Changing packing instructions clearly has downsides/challenges. \
+    // For now, let's move under the assumption that only the packing instructions are modified - i.e. byte stream properties never deviate from the original implementation.
+        if (RcApiTable[apiInx]->outBsFlag > 0 && outBsBBuf == nullptr) {
+            rodsLog(LOG_ERROR, "%s: outBsBBuf error for B apiNumber %d", __func__, RcApiTable[apiInx]->apiNumber);
```
trel, line 242: log_api?

lib/core/src/procApiRequest.cpp
```diff
@@ -20,23 +20,316 @@
...
+#else
+            if (ret.code() == SYS_HEADER_READ_LEN_ERR) {
+                ret = CODE(SYS_INTERNAL_ERR);
+            }
+#endif
+            if (conn->svrVersion != nullptr && conn->svrVersion->reconnPort > 0) {
+                int savedStatus = static_cast<int>(ret.code());
+                /* try again. the socket might have changed */
+                conn->thread_ctx->lock->lock();
+                rodsLog(LOG_DEBUG,
```
trel, line 267: log_api?

lib/core/src/procApiRequest.cpp
```diff
@@ -73,6 +366,44 @@ int procApiRequest(rcComm_t *conn,
...
+        return USER__NULL_INPUT_ERR;
+    }
+
+    freeRError(conn->rError);
+    conn->rError = nullptr;
+
+    const int apiInx = apiTableLookup(apiNumber);
+
+    if (apiInx < 0) {
+        rodsLog(LOG_ERROR, "procApiRequest: apiTableLookup of apiNumber %d failed", apiNumber);
```
trel, line 372: log_api?

lib/core/src/procApiRequest.cpp
```diff
@@ -73,6 +366,44 @@ int procApiRequest(rcComm_t *conn,
...
+        return apiInx;
+    }
+
+    if (const auto ec =
+            send_api_request(conn, apiInx, packingInstructionTable, inputPackingInstruction, inputStruct, inputBsBBuf);
+        ec < 0)
+    {
+        rodsLogError(LOG_DEBUG, ec, "procApiRequest: sendApiRequest failed. status = %d", ec);
```
trel, line 380: log_api?

lib/core/include/irods/procApiRequest.h
```diff
@@ -1,19 +1,88 @@
...
+///                         deserialization. Can be set to \p NULL if no output is expected.
+///                         \endparblock
+/// \param[out] outStruct   \parblock The output data structure to be filled by the server. The
+///                         output will be heap-allocated and assigned to the pointer. Can be
+///                         passed \p NULL if no output is expected. \endparblock
+/// \param[out] outBsBBuf   \parblock The output byte buffer to be filled by the server.
+///                         Can be passed \p NULL if no output stream is expected. \endparblock
+///
+/// \return An integer indicating success or failure.
+/// \retval >=0 On success.
```
trel, line 92: just checking if this is correct?  >=?

lib/core/src/procApiRequest.cpp
trel: or even.. something more critical?  if we shouldn't be here...

lib/core/src/procApiRequest.cpp
```diff
@@ -20,23 +20,316 @@
...
+                                int apiInx,
+                                const PackingInstruction* packingInstructionTable,
+                                const char* outputPackingInstruction,
+                                void** outStruct,
+                                bytesBuf_t* outBsBBuf)
+    {
+        int status = 0;
+        msgHeader_t myHeader;
+
+        /* bytesBuf_t outStructBBuf, errorBBuf, myOutBsBBuf; */
```
trel: ?

lib/core/src/procApiRequest.cpp
```diff
@@ -20,23 +20,316 @@
...
+            cliChkReconnAtReadEnd(conn);
+            return USER_API_INPUT_ERR;
+        }
+
+        // TODO Do I need to require the outBsFlag be passed in too? \
+    // Then again, RcApiTable[apiInx]->outBsFlag is the apidef_t information. \
+    // Changing the packing instruction DOES NOT affect the output byte stream, so I think this is fine as is. \
+    // What if an API didn't originally have an output byte stream? Is it a different API if it does in the future? This feels like a different API. \
+    // Seems we need to define rules around all of this. Changing packing instructions clearly has downsides/challenges. \
+    // For now, let's move under the assumption that only the packing instructions are modified - i.e. byte stream properties never deviate from the original implementation.
```
trel: i think that is an okay assertion for now.

lib/core/src/procApiRequest.cpp
alanking: This is likely happening in a client, so there may be no way to log other than stdout.

lib/core/src/procApiRequest.cpp
```diff
@@ -20,23 +20,316 @@
...
+                rodsLog(LOG_ERROR, "%s: Packing instruction error; error_code=[%d]", __func__, status);
+                cliChkReconnAtSendEnd(conn);
+                return status;
+            }
+
+            myInputStructBBuf = inputStructBBuf;
+        }
+        else {
+            myInputStructBBuf = nullptr;
+        }
```
alanking: `myInputStructBBuf` is initialized to `nullptr` so this `else` is technically not needed. Could be good for clarity, though.

unit_tests/src/test_procApiRequest_raw.cpp
```diff
@@ -0,0 +1,175 @@
...
+                                           pi_name_in,
+                                           &input,
+                                           nullptr,
+                                           "RodsObjStat_PI", // For deserialization.
+                                           static_cast<void**>(static_cast<void*>(&output)),
+                                           nullptr);
+        REQUIRE(ec == SYS_PACK_INSTRUCT_FORMAT_ERR);
+    }
+
+    SECTION("bad output packing instruction")
```
alanking: I noticed we have a truncated input struct test, but no truncated output struct test. Is that even a valid concept?

lib/core/src/procApiRequest.cpp
```diff
@@ -20,23 +20,316 @@
...
+                                      outStruct,
+                                      outBsBBuf,
+                                      &myHeader,
+                                      &outStructBBuf,
+                                      nullptr,
+                                      &errorBBuf);
+        }
+
+        clearBBuf(&outStructBBuf);
+        /* clearBBuf (&myOutBsBBuf); */
```
alanking: Please remove

lib/core/src/procApiRequest.cpp
```diff
@@ -20,23 +20,316 @@
...
+
+        /* bytesBuf_t outStructBBuf, errorBBuf, myOutBsBBuf; */
+        bytesBuf_t outStructBBuf;
+        bytesBuf_t errorBBuf;
+
+        cliChkReconnAtReadStart(conn);
+
+        memset(&outStructBBuf, 0, sizeof(bytesBuf_t));
+        memset(&errorBBuf, 0, sizeof(bytesBuf_t));
+        /* memset (&myOutBsBBuf, 0, sizeof (bytesBuf_t)); */
```
alanking: Please remove

lib/core/src/procApiRequest.cpp
```diff
@@ -20,23 +20,316 @@
...
- *
- */
+namespace
+{
+    auto send_api_request(rcComm_t* conn,
+                          int apiInx,
+                          const PackingInstruction* _packingInstructionTable,
+                          const char* _packingInstruction,
+                          const void* inputStruct,
+                          const bytesBuf_t* inputBsBBuf) -> int
```
alanking, line 30: Inconsistent use of leading underscores in these parameter names, if that matters

alanking (commented):
Looks like the implementation was mostly copied from the existing functions, so I left most of it alone. It would be interesting to see whether you can call the "raw" function from the existing one to consolidate implementations a bit, but that might be risky for the purposes of stability.

korydraughn:
> Looks like the implementation was mostly copied from the existing functions, so I left most of it alone.

Correct. It is a copy with some small modifications.

> It would be interesting to see whether you can call the "raw" function from the existing one to consolidate implementations a bit, but that might be risky for the purposes of stability.

I initially started down that path, but then turned back to avoid breaking everything. It's just safer to work with a copy and leave the existing code as it is.

lib/core/src/procApiRequest.cpp
korydraughn: The new logger is server-side only.

I should be able to switch which logger is used via `::ProcessType` though.

lib/core/src/procApiRequest.cpp
korydraughn: This code is copied from `procApiRequest`. I avoided modifying these calls to make sure it aligns with the original.

We can change it if that's preferred though.

lib/core/src/procApiRequest.cpp
korydraughn: Correct.

lib/core/include/irods/procApiRequest.h
korydraughn: Yes, that's correct.

lib/core/src/procApiRequest.cpp
korydraughn: That comes from copying the original implementation of the function.

lib/core/src/procApiRequest.cpp
korydraughn: Agreed. Do you feel the documentation in the header file captures that well?

lib/core/include/irods/procApiRequest.h
trel: okay

unit_tests/src/test_procApiRequest_raw.cpp
korydraughn: I'm... unsure. It's something I discovered while trying to put together tests and decided it was worth keeping. It's hard to say whether it has value.

lib/core/src/procApiRequest.cpp
trel: i feel the assumption / reasoning / defense of why these are the decisions and this is the approach is only covered here at this point.

lib/core/src/procApiRequest.cpp
korydraughn: Keeping this as is so that it aligns with its sibling implementation.

We can address this in a later PR. I'll open an issue for this.

lib/core/src/procApiRequest.cpp
korydraughn: See https://github.com/irods/irods/pull/8672#discussion_r2369225556.

lib/core/src/procApiRequest.cpp
korydraughn: See https://github.com/irods/irods/pull/8672#discussion_r2369225556.

lib/core/src/procApiRequest.cpp
korydraughn: See https://github.com/irods/irods/pull/8672#discussion_r2369225556.

lib/core/src/procApiRequest.cpp
korydraughn: See https://github.com/irods/irods/pull/8672#discussion_r2369225556.

lib/core/src/procApiRequest.cpp
korydraughn: See https://github.com/irods/irods/pull/8672#discussion_r2369225556.

lib/core/src/procApiRequest.cpp
korydraughn: See https://github.com/irods/irods/pull/8672#discussion_r2369225556.

lib/core/src/procApiRequest.cpp
korydraughn: See https://github.com/irods/irods/pull/8672#discussion_r2369225556.

lib/core/src/procApiRequest.cpp
korydraughn: See https://github.com/irods/irods/pull/8672#discussion_r2369225556.

lib/core/src/procApiRequest.cpp
korydraughn: See https://github.com/irods/irods/pull/8672#discussion_r2369225556.

lib/core/src/procApiRequest.cpp
korydraughn: See https://github.com/irods/irods/pull/8672#discussion_r2369225556.

lib/core/src/procApiRequest.cpp
korydraughn: See https://github.com/irods/irods/pull/8672#discussion_r2369225556.

lib/core/src/procApiRequest.cpp
korydraughn: See https://github.com/irods/irods/pull/8672#discussion_r2369225556.

lib/core/src/procApiRequest.cpp
korydraughn: See https://github.com/irods/irods/pull/8672#discussion_r2369225556.

lib/core/src/procApiRequest.cpp
korydraughn: See https://github.com/irods/irods/pull/8672#discussion_r2369225556.

lib/core/src/procApiRequest.cpp
korydraughn: See https://github.com/irods/irods/pull/8672#discussion_r2369225556.

lib/core/src/procApiRequest.cpp
korydraughn: See https://github.com/irods/irods/pull/8672#discussion_r2369225556.

lib/core/src/procApiRequest.cpp
korydraughn: See https://github.com/irods/irods/pull/8672#discussion_r2369225556.

lib/core/include/irods/procApiRequest.h
korydraughn: Done.

lib/core/include/irods/procApiRequest.h
korydraughn: Done.

lib/core/include/irods/procApiRequest.h
korydraughn: Done.

lib/core/include/irods/procApiRequest.h
korydraughn: Done.

lib/core/src/procApiRequest.cpp
korydraughn: Removed.

lib/core/src/procApiRequest.cpp
korydraughn: Added warning about this to documentation.

lib/core/src/procApiRequest.cpp
korydraughn: Removed.

lib/core/src/procApiRequest.cpp
korydraughn: Done.

lib/core/src/procApiRequest.cpp
korydraughn: Done.

lib/core/src/procApiRequest.cpp
korydraughn: Fixed.

unit_tests/src/test_procApiRequest_raw.cpp
korydraughn: Coming back to this with fresh eyes again, I believe the main takeaway for the truncated structs is that we have the option to stop serialization/deserialization early. That is why the truncated input PI works for the API.

The only reason for using something like this is for optimization purposes IMO.

Will add a test for the output struct.

lib/core/src/procApiRequest.cpp
korydraughn: No need for a new issue. We already have issues open for replacing rodsLog calls.

Resolving.

unit_tests/src/test_procApiRequest_raw.cpp
korydraughn: No test added and confirmed to be passing.

korydraughn:
Running unit tests.

Will squash and pound after they pass.

korydraughn:
I'm seeing an intermittent failure with the "truncated input packing instruction" test.

It's unclear why the test fails sometimes. My guess is the truncated request results in the server seeing garbage.

Given the truncated packing instructions do not have any known use-cases, I'm going to disable those parts of the tests and get this merged. I'll verify the test passes consistently before merging.

unit_tests/src/test_procApiRequest_raw.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+    // The following code is disabled because the test for truncated input packing
+    // instructions can fail unexpectedly, causing testing to take more time than needed.
+    // This behavior has not been observed for the truncated output packing instruction
+    // test.
+    //
+    // The other reason for disabling these tests is due to there being no known use-case
+    // for truncated packing instructions.
+#if 0
+    SECTION("truncated input packing instruction")
```
korydraughn, line 164: State why the tests are still included in the source.

lib/core/include/irods/procApiRequest.h
```diff
@@ -1,19 +1,86 @@
...
+/// \since 4.3.5
+int procApiRequest_raw(rcComm_t* conn,
+                       int apiNumber,
+                       const struct PackingInstruction* packingInstructionTable,
+                       const char* inputPackingInstruction,
+                       const void* inputStruct,
+                       const bytesBuf_t* inputBsBBuf,
+                       const char* outputPackingInstruction,
+                       void** outStruct,
+                       bytesBuf_t* outBsBBuf);
```
korydraughn, line 169: Is `procApiRequest_raw` the best name for this?

korydraughn:
The test passes consistently now.

unit_tests/src/test_procApiRequest_raw.cpp
korydraughn: Done.

unit_tests/src/test_procApiRequest_raw.cpp
```diff
@@ -145,10 +145,13 @@ TEST_CASE("#8653: procApiRequest_raw allows overriding packing instructions")
...
 
-        // The data ID should be empty because the packing instruction isn't configured to
-        // extract and store it in the output data structure. The info is in the response
-        // from the server, it's just ignored by the client.
-        CHECK(std::string_view{output->dataId}.empty());
+        // The remaining member variables likely contain garbage because the data was
+        // never extracted from the server's response. The reason for this is because the
+        // packing instruction does not mention those members. This behavior cannot be
+        // asserted because the output pointer will point to a heap-allocated object. We
+        // cannot know what the initial state of the heap-allocated is.
```
alanking:
```suggestion
        // cannot know what the initial state of the heap-allocated object is.
```

unit_tests/src/test_procApiRequest_raw.cpp
```diff
@@ -207,6 +201,21 @@ TEST_CASE("#8653: procApiRequest_raw allows overriding packing instructions")
...
+        // zero. One of the primary pieces of info we're interested in is the type. That
+        // should be extracted and stored in the output data structure.
+        CHECK(0 == output->objSize);
+        CHECK(COLL_OBJ_T == output->objType);
+
+        // The remaining member variables likely contain garbage because the data was
+        // never extracted from the server's response. The reason for this is because the
+        // packing instruction does not mention those members. This behavior cannot be
+        // asserted because the output pointer will point to a heap-allocated object. We
+        // cannot know what the initial state of the heap-allocated is.
```
alanking:
```suggestion
        // cannot know what the initial state of the heap-allocated object is.
```

unit_tests/src/test_procApiRequest_raw.cpp
```diff
@@ -137,26 +147,20 @@ TEST_CASE("#8653: procApiRequest_raw allows overriding packing instructions")
...
     }
 
-    SECTION("bad input packing instruction")
+    // The following code is disabled because the test for truncated input packing
+    // instructions can fail unexpectedly, causing testing to take more time than needed.
+    // This behavior has not been observed for the truncated output packing instruction
+    // test.
+    //
+    // The other reason for disabling these tests is due to there being no known use-case
+    // for truncated packing instructions.
```
alanking, line 159: Should we document the behavior of truncated input packing instructions as undefined if it is known to cause issues? I see the following in the docstring for `procApiRequest_raw`:
> ```
> /// \warning This function is provided for handling changes in packing instructions for an API
> /// only. All other properties of the original API must be honored. The behavior of this function
> /// is undefined if preconditions are violated.
> ```

Is "preconditions" referring to the truncated input packing instructions, or something else?

alanking (commented):
Looks real close

unit_tests/src/test_procApiRequest_raw.cpp
korydraughn: I don't think we need to document it explicitly. We don't know all of the edge cases around truncated instructions, so it sounds difficult to document the behavior accurately. 

Rewording things such that they define the well-paved path and discourage unknown paths is a much easier thing to do IMO.

The line about preconditions doesn't add anything and is mostly redundant. Will be removing that.

lib/core/include/irods/procApiRequest.h
```diff
@@ -1,19 +1,86 @@
...
+/// \param[out] outStruct   The output data structure to be filled by the server. The output will
+///                         be heap-allocated and assigned to the pointer. Can be passed \p NULL
+///                         if no output is expected.
+/// \param[out] outBsBBuf   The output byte buffer to be filled by the server. Can be passed
+///                         \p NULL if no output stream is expected.
+///
+/// \return An integer indicating success or failure.
+/// \retval >=0 On success.
+/// \retval  <0 On failure.
 int procApiRequest(rcComm_t *conn,
```
korydraughn, line 55: Add example.

lib/core/include/irods/procApiRequest.h
```diff
@@ -1,19 +1,86 @@
...
+///                         if no output is expected.
+/// \param[out] outBsBBuf   The output byte buffer to be filled by the server. Can be passed
+///                         \p NULL if no output stream is expected.
+///
+/// \return An integer indicating success or failure.
+/// \retval >=0 On success.
+/// \retval  <0 On failure.
+///
+/// \since 4.3.5
+int procApiRequest_raw(rcComm_t* conn,
```
korydraughn, line 161: Add example or reference to unit test.

unit_tests/src/test_procApiRequest_raw.cpp
korydraughn: Done.

unit_tests/src/test_procApiRequest_raw.cpp
korydraughn: Done.

unit_tests/src/test_procApiRequest_raw.cpp
alanking: Okay. So the idea is that the packing instructions should always include all of the members required by the API, which is kind of a given?

korydraughn: That's correct. The only focus of the new function is using custom packing instructions. Everything else remains the same.

korydraughn: Removed statement about preconditions.

lib/core/include/irods/procApiRequest.h
korydraughn: Done.

lib/core/include/irods/procApiRequest.h
korydraughn: Done.

lib/core/include/irods/procApiRequest.h
korydraughn: How do we feel about this name?

I'm tempted to say it's likely fine. I don't see us ever changing existing packing instructions again. And looking at the other inputs, I don't see any additional room for customization - i.e. we'd just introduce an entirely new API at that point.

Thoughts?

trel: that sounds reasonable/correct.

alanking: Agreed

korydraughn: Okay. Resolving.

korydraughn:
The changes are now complete. All that's left to do is to squash it.

The new unit test still passes. I verified the changes still resolve the imiscsvrinfo segfault when connected to 4.3.3 too.

Will squash once the signal is given.

lib/core/include/irods/procApiRequest.h
```diff
@@ -29,6 +29,26 @@ extern "C" {
...
+/// gridConfigurationInp_t input;
+/// memset(&input, 0, sizeof(gridConfigurationInp_t));
+/// strcpy(input.name_space, "delay_server");
+/// strcpy(input.option_name, "leader");
+///
+/// gridConfigurationOut_t* output = NULL;
+///
+/// int ec = procApiRequest(conn, GET_GRID_CONFIGURATION_VALUE_APN, &input, NULL, (void**) &output, NULL);
+/// if (ec != 0) {
+///     // Handle error.
```
alanking: Do we need to free the output in the error case, too? Or does it depend on the API?

lib/core/include/irods/procApiRequest.h
```diff
@@ -69,6 +89,67 @@ int procApiRequest(rcComm_t *conn,
...
+///                             GET_GRID_CONFIGURATION_VALUE_APN,
+///                             pi_table,
+///                             my_input_struct_pi_name,
+///                             &input,
+///                             NULL,
+///                             my_output_struct_pi_name,
+///                             (void**) &output,
+///                             NULL);
+/// if (ec != 0) {
+///     // Handle error.
```
alanking: Do we need to free the output in the error case, too? Or does it depend on the API?

lib/core/include/irods/procApiRequest.h
korydraughn: I think it depends on the API.

lib/core/include/irods/procApiRequest.h
korydraughn: I think it depends on the API.

lib/core/include/irods/procApiRequest.h
korydraughn: I'll likely change that `!= 0` to `< 0` on the squash.

alanking: Hmm... okay. Maybe put a reminder in the comment to free the output if necessary here too.

korydraughn: Done.

lib/core/include/irods/procApiRequest.h
korydraughn: Adjusted example.

alanking (commented):
Looks good. Squash em

korydraughn:
Squashed.

alanking (approved):
\# it

korydraughn:
Added pounds.

Issue #8676, opened 2025-09-19
Downgrading iRODS generates stacktrace


korydraughn:
- [x] main
- [x] 5-0-stable
---
## Bug Report

While attempting to **downgrade** my installation of iRODS from 5.0.90 to 5.0.2-rc, I got the following when running `python3 scripts/upgrade_irods.py -v`. This should not happen. The script should print a clean message about downgrades not being supported and exit. No stacktrace.
```
$ python3 scripts/upgrade_irods.py -v
Traceback (most recent call last):
  File "/var/lib/irods/scripts/upgrade_irods.py", line 76, in <module>
    sys.exit(main())
  File "/var/lib/irods/scripts/upgrade_irods.py", line 66, in main
    IrodsController(irods_config).upgrade()
  File "/var/lib/irods/scripts/irods/controller.py", line 78, in upgrade
    upgrade_configuration.raise_exception_if_downgrade_is_detected(self.config)
  File "/var/lib/irods/scripts/irods/upgrade_configuration.py", line 96, in raise_exception_if_downgrade_is_detected
    raise IrodsError(f'Downgrade detected. Downgrading is unsupported, please reinstall iRODS version [{irods_config.version["irods_version"]}] or newer.')
  File "/var/lib/irods/scripts/irods/configuration.py", line 124, in version
    self._version = load_json_config(paths.version_path())
  File "/var/lib/irods/scripts/irods/configuration.py", line 357, in load_json_config
    return lib.open_and_load_json(path)
  File "/var/lib/irods/scripts/irods/lib.py", line 147, in open_and_load_json
    with open(filename, 'rt') as f:
FileNotFoundError: [Errno 2] No such file or directory: '/var/lib/irods/version.json'
```

The stacktrace points to `irods_config.version["irods_version"]` being the problem. `version.json` does not exist following installation of older packages.

https://github.com/irods/irods/blob/989e6c59005039d1b7b1ecf672b57048ff86aca4/scripts/irods/upgrade_configuration.py#L96

PR #8678, opened 2025-09-20
[#8676] Do not stacktrace on detection of server downgrade (5-0-stable)

korydraughn:
When attempting to downgrade from 5.0.90 to 5.0.2, the following is printed.
```
irods:~$ python3 scripts/upgrade_irods.py -v
Downgrade detected. Downgrading is unsupported, please reinstall iRODS version [5.0.90] or newer.
```


trel:
seems clean and minimal.  go for it.

PR #8681, opened 2025-09-22
[#8676] Do not stacktrace on detection of server downgrade (main)

korydraughn:
Cherry-picked from #8678.


trel:
looks identical - go for it.

korydraughn:
Added pound.

PR #8692, opened 2025-10-15
[8691] Make resource quota count the largest overrun (main)

FifthPotato:
Draft so far; do we want to merge fixes piecemeal, or should I try to stuff as much as I can into one PR?

Also taking suggestions on naming, style, approach, as always.

### DIFF
--- a/scripts/irods/test/test_quotas.py
+++ b/scripts/irods/test/test_quotas.py
@@ -97,3 +97,59 @@ class Test_Quotas(resource_suite.ResourceBase, unittest.TestCase):
 
         finally:
             self.admin.assert_icommand(['iadmin', 'rmgroup', 'test_group_3507'])
+    
+    # The following test covers case 10, 11, 12
+    def test_physical_quota_on_single_resource_with_multiple_groups__issue_8691(self):
+        pep_map = {
+            'irods_rule_engine_plugin-irods_rule_language': textwrap.dedent('''
+                acRescQuotaPolicy {
+                    msiSetRescQuotaPolicy("on");
+                }
+            '''),
+            'irods_rule_engine_plugin-python': textwrap.dedent('''
+                def acRescQuotaPolicy(rule_args, callback, rei):
+                    callback.msiSetRescQuotaPolicy('on')
+            ''')
+        }
+
+        try:
+            with temporary_core_file() as core:
+                core.add_rule(pep_map[self.plugin_name])
+                IrodsController().reload_configuration()
+
+                group_name = 'test_group_case_10_11_12'
+                file_name = 'test_file_case_10_11_12'
+                resc_name = 'ufs0_quota_case_10_11_12'
+                lib.make_file(file_name, 4096, contents='arbitrary')
+                lib.create_ufs_resource(self.admin, resc_name)
+
+                self.admin.assert_icommand(['iadmin', 'mkgroup', f'{group_name}_1'])
+                self.admin.assert_icommand(['iadmin', 'mkgroup', f'{group_name}_2'])
+                self.admin.assert_icommand(['iadmin', 'sgq', f'{group_name}_1', resc_name, '5000'])
+                self.admin.assert_icommand(['iadmin', 'sgq', f'{group_name}_2', resc_name, '10000'])
+
+                self.admin.assert_icommand(['iadmin', 'atg', f'{group_name}_1', self.user0.username])
+                self.admin.assert_icommand(['iadmin', 'atg', f'{group_name}_2', self.user0.username])
+
+                self.user0.assert_icommand(['iput', '-R', self.testresc, file_name])
+                self.admin.assert_icommand(['iadmin', 'cu'])
+
+                # Case 10: under quota succeeds
+                with self.subTest(testcase=10, desc="Under quota before and after"):
+                    self.user0.assert_icommand(['icp', '-R', resc_name, file_name, f'{file_name}_case_10'])
+                    self.admin.assert_icommand(['iadmin', 'cu'])
+
+                # Case 11: crossing quota succeeds
+                with self.subTest(testcase=11, desc="Crossing quota"):
+                    self.user0.assert_icommand(['icp', '-R', resc_name, file_name, f'{file_name}_case_11'])
+                    self.admin.assert_icommand(['iadmin', 'cu'])
+
+                # Case 12: should fail with just a single quota violated
+                with self.subTest(testcase=12, desc="Violating quota"):
+                    self.user0.assert_icommand(['icp', '-R', resc_name, file_name, f'{file_name}_case_12'], 'STDERR_SINGLELINE', 'SYS_RESC_QUOTA_EXCEEDED')
+        finally:
+            self.admin.run_icommand(['irm', '-f', f'{file_name}_case_10', f'{file_name}_case_11', f'{file_name}_case_12'])
+            self.admin.run_icommand(['iadmin', 'rmresc', resc_name])
+            self.admin.run_icommand(['iadmin', 'rmgroup', f'{group_name}_1'])
+            self.admin.run_icommand(['iadmin', 'rmgroup', f'{group_name}_2'])
+            IrodsController().reload_configuration()
--- a/server/api/src/rsGetRescQuota.cpp
+++ b/server/api/src/rsGetRescQuota.cpp
@@ -267,6 +267,17 @@ int setRescQuota(
     }
 
     rodsLong_t compare_overrun = resc_quota->quotaOverrun;
+    rescQuota_t* iterator = resc_quota;
+
+    // Traverse the quota result linked list and find the
+    // "most egregious" overrun: i.e., if there are
+    // two groups with quotas on the same resource, find
+    // the one most in violation.
+    while (iterator = iterator->next) {
+        if (compare_overrun < iterator->quotaOverrun) {
+            compare_overrun = iterator->quotaOverrun;
+        }
+    }
     freeAllRescQuota(resc_quota);
     if ( compare_overrun != 0 && // disabled if 0
          compare_overrun + resc_overrun + _data_size > 0 ) {


korydraughn:
Piecemeal is fine. If you identify a design flaw, then that will likely require additional thought.

scripts/irods/test/test_quotas.py
```diff
@@ -97,3 +97,54 @@ def test_filter_out_groups_when_selecting_user__issue_3507(self):
 
         finally:
             self.admin.assert_icommand(['iadmin', 'rmgroup', 'test_group_3507'])
+
+    def test_physical_quota_case_10_11_12(self):
```
korydraughn: Missing issue number suffix.

The cases which the test cover can be captured in a comment at the start of the test rather than the test name.

scripts/irods/test/test_quotas.py
```diff
@@ -97,3 +97,54 @@ def test_filter_out_groups_when_selecting_user__issue_3507(self):
...
+                lib.make_file(file_name, 4096, contents='arbitrary')
+                lib.create_ufs_resource(self.admin, resc_name)
+
+                self.admin.assert_icommand(['iadmin', 'mkgroup', f'{group_name}_1'])
+                self.admin.assert_icommand(['iadmin', 'mkgroup', f'{group_name}_2'])
+                self.admin.assert_icommand(['iadmin', 'sgq', f'{group_name}_1', resc_name, '5000'])
+                self.admin.assert_icommand(['iadmin', 'sgq', f'{group_name}_2', resc_name, '10000'])
+
+                self.admin.assert_icommand(['iadmin', 'atg', f'{group_name}_1', self.admin.username])
+                self.admin.assert_icommand(['iadmin', 'atg', f'{group_name}_2', self.admin.username])
```
korydraughn: Let's add a non-admin user to the group when testing the use-case.

Admin users can sometimes hide bugs.

scripts/irods/test/test_quotas.py
```diff
@@ -97,3 +97,54 @@ def test_filter_out_groups_when_selecting_user__issue_3507(self):
...
+                self.admin.assert_icommand(['icp', '-R', resc_name, file_name, f'{file_name}_case_10'])
+
+                self.admin.assert_icommand(['iadmin', 'cu'])
+
+                # Case 11: crossing quota succeeds
+                self.admin.assert_icommand(['icp', '-R', resc_name, file_name, f'{file_name}_case_11'])
+                self.admin.assert_icommand(['iadmin', 'cu'])
+
+                # Case 12: should fail with just a single quota violated
+                self.admin.assert_icommand(['icp', '-R', resc_name, file_name, f'{file_name}_case_12'], 'STDERR_SINGLELINE', 'SYS_RESC_QUOTA_EXCEEDED')
```
korydraughn: Consider using the `self.subTest()` feature so these stand out on failure.

Not a requirement, just something to consider.

scripts/irods/test/test_quotas.py
```diff
@@ -97,3 +97,54 @@ def test_filter_out_groups_when_selecting_user__issue_3507(self):
...
+                self.admin.assert_icommand(['iadmin', 'mkgroup', f'{group_name}_1'])
+                self.admin.assert_icommand(['iadmin', 'mkgroup', f'{group_name}_2'])
+                self.admin.assert_icommand(['iadmin', 'sgq', f'{group_name}_1', resc_name, '5000'])
+                self.admin.assert_icommand(['iadmin', 'sgq', f'{group_name}_2', resc_name, '10000'])
+
+                self.admin.assert_icommand(['iadmin', 'atg', f'{group_name}_1', self.admin.username])
+                self.admin.assert_icommand(['iadmin', 'atg', f'{group_name}_2', self.admin.username])
+
+                self.admin.assert_icommand(['iput', '-R', self.testresc, file_name])
+
```
trel, line 136: not strictly required - but could be cleaner to also have `iadmin cu` here -  to make sure the totals are up to date.

FifthPotato:
> Looks good so far.
> 
> Does the test capture the bug? Does the fix resolve the bug?

I think so. I wrote the test first, and it failed repeatedly, but this change makes the test pass.

scripts/irods/test/test_quotas.py
FifthPotato: Do we have a stylistic preference on what to pass into the subTest parameter to help distinguish these cases? I could iterate over some descriptor strings?

korydraughn: No preference. Think about what you'd want to see if you never seen the test before. Help the next developer get up to speed quickly.

Again, it's not a requirement. The decision on whether subTests seem useful or the comments being enough is up to you.

scripts/irods/test/test_quotas.py
```diff
@@ -97,3 +97,56 @@ def test_filter_out_groups_when_selecting_user__issue_3507(self):
 
         finally:
             self.admin.assert_icommand(['iadmin', 'rmgroup', 'test_group_3507'])
+    
+    # The following test covers case 10, 11, 12
+    def test_physical_quota__issue_8691(self):
```
korydraughn: This test name doesn't tell us much about the actual test. Think about what can be added to give the reader more context. Otherwise, we may end up with many tests sharing very similar names.

scripts/irods/test/test_quotas.py
```diff
@@ -97,3 +97,58 @@ def test_filter_out_groups_when_selecting_user__issue_3507(self):
...
+                    self.admin.assert_icommand(['iadmin', 'cu'])
+
+                # Case 12: should fail with just a single quota violated
+                with self.subTest(testcase=12, desc="Violating quota"):
+                    self.user0.assert_icommand(['icp', '-R', resc_name, file_name, f'{file_name}_case_12'], 'STDERR_SINGLELINE', 'SYS_RESC_QUOTA_EXCEEDED')
+        finally:
+            self.admin.run_icommand(['irm', '-f', f'{file_name}_case_10', f'{file_name}_case_11', f'{file_name}_case_12'])
+            self.admin.run_icommand(['iadmin', 'rmresc', resc_name])
+            self.admin.run_icommand(['iadmin', 'rmgroup', f'{group_name}_1'])
+            self.admin.run_icommand(['iadmin', 'rmgroup', f'{group_name}_2'])
```
korydraughn, line 154: Missing a reload of the server configuration. Without it, the quota system will remain enabled.

scripts/irods/test/test_quotas.py
```diff
@@ -97,3 +97,58 @@ def test_filter_out_groups_when_selecting_user__issue_3507(self):
 
         finally:
             self.admin.assert_icommand(['iadmin', 'rmgroup', 'test_group_3507'])
+    
+    # The following test covers case 10, 11, 12
```
korydraughn, line 101: Hmm, we'll need to think through where the table will live if the test is going to refer to test cases within it.

FifthPotato: Makes sense. Any ideas where?
Also, do we think the table is in a good enough state to publish (in its 400+ case glory) or are we considering all the icommands to be the same (and thus, I should trim it down a little?)

FifthPotato:
> Do we have any quota tests which exist outside of test_quotas.py?

None that actually seem to test the quota system itself. The most I could find (with a recursive grep on test-y stuff for "quota") is the following:

https://github.com/irods/irods/blob/0b25f949f169c884eb22f8a11c7f1434fd8d0ad8/scripts/irods/test/test_rulebase.py#L905

https://github.com/irods/irods/blob/0b25f949f169c884eb22f8a11c7f1434fd8d0ad8/scripts/irods/test/test_rulebase.py#L928

https://github.com/irods/irods/blob/0b25f949f169c884eb22f8a11c7f1434fd8d0ad8/scripts/irods/test/test_iadmin.py#L112-L118

scripts/irods/test/test_quotas.py
korydraughn: I'm not entirely sure yet.

The table isn't a blocker for this PR though. We can get this merged once complete and circle back to the table conversation and code comments after more decisions have been made.

For now, continue doing what you're doing - that is, implementing good tests and making quotas solid.

korydraughn:
Cool cool.

The icommands repo has several rule files used for testing too. Try grepping that repo for quota terminology.

scripts/irods/test/test_quotas.py
trel: The table will probably go in the main issue for testing all the quotas.   But yes, agreed, let it ride for now.

FifthPotato: Got it. Will resolve this for now, then.

FifthPotato:
> Cool cool.
> 
> The icommands repo has several rule files used for testing too. Try grepping that repo for quota terminology.

Oh, good call. I found these:

https://github.com/irods/irods_client_icommands/blob/main/test/rules/rulemsiQuota.r

https://github.com/irods/irods_client_icommands/blob/main/test/rules/rulemsiSetQuota.r

korydraughn:
I believe those tests are invoked as part of test_all_rules.py.

I think we're in a good spot for running the unit and core tests.

FifthPotato:
> I believe those tests are invoked as part of test_all_rules.py.
> 
> I think we're in a good spot for running the unit and core tests.

Gotcha. Starting those now.

FifthPotato:
Core and unit tests pass.

korydraughn:
Squash to taste.

FifthPotato:
Squished.

korydraughn:
Please update the commit message body to include words explaining what this is about.

FifthPotato:
Added a story to the body.

FifthPotato:
Octothorpe applied

korydraughn:
Please cherry-pick to 4-3-stable, no pounds.

korydraughn:
Merging.

PR #8696, opened 2025-10-23
[4089] Add test to ensure that total and resc quotas together work properly (main)

FifthPotato:
This test passes, so potentially no server changes needed.

### DIFF
--- a/scripts/irods/test/test_quotas.py
+++ b/scripts/irods/test/test_quotas.py
@@ -120,7 +120,7 @@ class Test_Quotas(resource_suite.ResourceBase, unittest.TestCase):
                 group_name = 'test_group_case_10_11_12'
                 file_name = 'test_file_case_10_11_12'
                 resc_name = 'ufs0_quota_case_10_11_12'
-                lib.make_file(file_name, 4096, contents='arbitrary')
+                lib.make_file(os.path.join(self.user0.local_session_dir, file_name), 4096, contents='arbitrary')
                 lib.create_ufs_resource(self.admin, resc_name)
 
                 self.admin.assert_icommand(['iadmin', 'mkgroup', f'{group_name}_1'])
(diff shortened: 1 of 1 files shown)


scripts/irods/test/test_quotas.py
```diff
@@ -153,3 +153,64 @@ def acRescQuotaPolicy(rule_args, callback, rei):
...
+
+                self.admin.assert_icommand(['iadmin', 'atg', f'{group_name}_1', self.user0.username])
+
+                self.user0.assert_icommand(['iput', '-R', self.testresc, file_name])
+                self.admin.assert_icommand(['iadmin', 'cu'])
+
+                # Case 40: under quota succeeds
+                with self.subTest(testcase=40, desc="Under quota before and after"):
+                    self.user0.assert_icommand(['icp', '-R', resc_name, file_name, f'{file_name}_case_40'])
+                    self.admin.assert_icommand(['iadmin', 'cu'])
```
korydraughn: Let's assert the values of each quota too. That will allow us to detect when the numbers are calculated incorrectly.

We should do the same for the test you wrote last week too.

Thoughts?

scripts/irods/test/test_quotas.py
```diff
@@ -153,3 +153,64 @@ def acRescQuotaPolicy(rule_args, callback, rei):
...
+
+                # Case 41: crossing quota succeeds
+                with self.subTest(testcase=41, desc="Crossing quota"):
+                    self.user0.assert_icommand(['icp', '-R', resc_name, file_name, f'{file_name}_case_41'])
```
korydraughn, line 204: I see `icp` is used, which is good/understandable. Have you thought of a plan for other icommands?

I'm not saying you have to cover them right now, just wondering if you've started developing opinions on how to tackle it and whether it's worth covering all of them.

scripts/irods/test/test_quotas.py
```diff
@@ -153,3 +153,64 @@ def acRescQuotaPolicy(rule_args, callback, rei):
...
+
+        try:
+            with temporary_core_file() as core:
+                core.add_rule(pep_map[self.plugin_name])
+                IrodsController().reload_configuration()
+
+                group_name = 'test_group_case_40_41_42'
+                file_name = 'test_file_case_40_41_42'
+                resc_name = 'ufs0_quota_case_40_41_42'
+                lib.make_file(file_name, 4096, contents='arbitrary')
```
alanking: Consider using the session directory to house the file so that it cleans itself up. Otherwise, might need to remove this file in the `finally` block.

scripts/irods/test/test_quotas.py
alanking: More assertions seems good

FifthPotato: To clarify, you're saying I should assert something like a `iadmin lq` after the `iadmin cu` to ensure the numbers are as expected? I could do that.

scripts/irods/test/test_quotas.py
FifthPotato: I have something brewing in my head about potentially trying to iterate over all of them, but it's still nebulous right now. Off the top of my head, I know that `iput` behaves a little differently on the "crossing" case, so I think it might still be valuable to test the other icommands.

scripts/irods/test/test_quotas.py
korydraughn: That's right.

korydraughn: Doing that also helps to confirm the other parts of the quota machinery work as intended.

scripts/irods/test/test_quotas.py
korydraughn: Sounds good to me.

FifthPotato:
Added more assertions. Test now passes after rebasing on tip-of-main.

scripts/irods/test/test_quotas.py
```diff
@@ -176,39 +180,45 @@ def acRescQuotaPolicy(rule_args, callback, rei):
...
                 with self.subTest(testcase=42, desc="Violating resc quota"):
                     self.user0.assert_icommand(['icp', '-R', resc_name, file_name, f'{file_name}_case_42'], 'STDERR_SINGLELINE', 'SYS_RESC_QUOTA_EXCEEDED')
 
-                self.admin.assert_icommand(['iadmin', 'sgq', f'{group_name}_1', resc_name, '10000'])
-                self.admin.assert_icommand(['iadmin', 'sgq', f'{group_name}_1', 'total', '5000'])
-                self.admin.assert_icommand(['iadmin', 'cu'])
+                    self.admin.assert_icommand(['iadmin', 'sgq', f'{group_name}_1', resc_name, '10000'])
+                    self.admin.assert_icommand(['iadmin', 'sgq', f'{group_name}_1', 'total', '10000'])
+                    self.admin.assert_icommand(['iadmin', 'cu'])
+                    self.user0.assert_icommand(['ils', '-rL'], 'STDOUT_SINGLELINE', '')
```
korydraughn: What is this line for? Debugging?

scripts/irods/test/test_quotas.py
```diff
@@ -176,39 +180,45 @@ def acRescQuotaPolicy(rule_args, callback, rei):
...
+                    self.admin.assert_icommand(['iadmin', 'lq'], 'STDOUT', ['quota_over: -1808', 'quota_over: 2288'])
```
korydraughn: What are these lines asserting/checking? Are they specific to the subtest below them?

scripts/irods/test/test_quotas.py
FifthPotato: Oh yeah, my bad, that was when I was trying to figure out what was going on with the quota math. Let me remove those.

scripts/irods/test/test_quotas.py
FifthPotato: Yes, they were meant to make the total quota be in violation instead. Now that you mention it, it makes more sense to group it with the below subtest.

scripts/irods/test/test_quotas.py
```diff
@@ -131,25 +131,94 @@ def acRescQuotaPolicy(rule_args, callback, rei):
...
+
+                self.admin.assert_icommand(['iadmin', 'mkgroup', f'{group_name}_1'])
+                self.admin.assert_icommand(['iadmin', 'sgq', f'{group_name}_1', resc_name, '5000'])
+                self.admin.assert_icommand(['iadmin', 'sgq', f'{group_name}_1', 'total', '15000'])
+
+                self.admin.assert_icommand(['iadmin', 'atg', f'{group_name}_1', self.user0.username])
+
+                self.user0.assert_icommand(['iput', '-R', self.testresc, os.path.join(self.user0.local_session_dir, file_name)])
+                self.admin.assert_icommand(['iadmin', 'cu'])
+                self.admin.assert_icommand(['iadmin', 'lq'], 'STDOUT', ['quota_over: -5000', 'quota_over: -10904'])
```
korydraughn: For all `iadmin lq` assertions, consider including newlines, etc. to delimit substrings. It would help protect against false positives, etc.

For example, the following guarantees that -5000 is matched rather than -50001. Same thing for the beginning of the label (quota_over), etc.
```py
self.admin.assert_icommand(['iadmin', 'lq'], 'STDOUT', ['\nquota_over: -5000\n', '\nquota_over: -10904\n'])
```

FifthPotato:
> One final comment for you to think about.
> 
> I see new commits were added, but there's no mention of whether the tests still pass following them.

Just ran this test individually after the latest commit, and it passed.

korydraughn:
Squash it.

FifthPotato:
Squished.

FifthPotato:
Octothorpe'd.

korydraughn:
@FifthPotato Please cherry-pick to 4-3-stable.

PR #8698, opened 2025-10-29
Adding support for crc64nvme

JustinKyleJames:
1. Added hash option for crc64nvme.
2. Added RESOURCE_OP_READ_CHECKSUM resource plugin operation.
3. Called the above operation before reading file to calculate checksum.  If it succeeds use the checksum returned from it.

I ran the tests (debian 12 / Postgres) and they passed.

### DIFF
--- a/lib/hasher/include/irods/CRC64NVMEStrategy.hpp
+++ b/lib/hasher/include/irods/CRC64NVMEStrategy.hpp
@@ -1,7 +1,10 @@
-#ifndef _CRC64NVME_STRATEGY_HPP_
-#define _CRC64NVME_STRATEGY_HPP_
+#ifndef IRODS_HASHER_CRC64NVME_STRATEGY_HPP
+#define IRODS_HASHER_CRC64NVME_STRATEGY_HPP
 
 #include "irods/HashStrategy.hpp"
+
+#include <boost/any.hpp>
+#include <string>
 
 namespace irods
 {
@@ -20,7 +23,10 @@ namespace irods
         error update(const std::string&, boost::any& context) const override;
         error digest(std::string& messageDigest, boost::any& context) const override;
         bool isChecksum(const std::string&) const override;
+
+      private:
+        static constexpr size_t crc_bits = 64;
     };
 } // namespace irods
 
-#endif // _CRC64NVME_STRATEGY_HPP_
+#endif // IRODS_HASHER_CRC64NVME_STRATEGY_HPP
--- a/lib/hasher/src/CRC64NVMEStrategy.cpp
+++ b/lib/hasher/src/CRC64NVMEStrategy.cpp
@@ -24,12 +24,11 @@ namespace irods
 
     auto CRC64NVMEStrategy::init(boost::any& _context) const -> error
     {
-        const size_t crc_bits = 64;
-        const uint64_t crc_polynomial = 0xAD93D23594C93659;
-        const uint64_t crc_initial_remainder = 0xFFFFFFFFFFFFFFFF;
-        const uint64_t crc_final_xor = 0xFFFFFFFFFFFFFFFF;
-        const bool crc_reflect_input = true;
-        const bool crc_reflect_output = true;
+        constexpr uint64_t crc_polynomial = 0xAD93D23594C93659;
+        constexpr uint64_t crc_initial_remainder = 0xFFFFFFFFFFFFFFFF;
+        constexpr uint64_t crc_final_xor = 0xFFFFFFFFFFFFFFFF;
+        constexpr bool crc_reflect_input = true;
+        constexpr bool crc_reflect_output = true;
 
         // Create the CRC calculator object
         boost::crc_basic<crc_bits>* context = new boost::crc_basic<crc_bits>(
@@ -42,7 +41,6 @@ namespace irods
 
     auto CRC64NVMEStrategy::update(const std::string& data, boost::any& _context) const -> error
     {
-        const size_t crc_bits = 64;
         boost::crc_basic<crc_bits>* context = boost::any_cast<boost::crc_basic<crc_bits>*>(_context);
         context->process_bytes(data.c_str(), data.length());
         return SUCCESS();
@@ -50,14 +48,13 @@ namespace irods
 
     auto CRC64NVMEStrategy::digest(std::string& messageDigest, boost::any& _context) const -> error
     {
-        const size_t crc_bits = 64;
         boost::crc_basic<crc_bits>* context = boost::any_cast<boost::crc_basic<crc_bits>*>(_context);
 
         uint64_t checksum = context->checksum();
 
         // Convert uint64_t to bytes (big-endian for consistency)
         std::array<unsigned char, 8> bytes;
-        for (int i = 0; i < 8; ++i) {
+        for (size_t i = 0; i < bytes.size(); ++i) {
             bytes[7 - i] = static_cast<unsigned char>((checksum >> (i * 8)) & 0xFF);
         }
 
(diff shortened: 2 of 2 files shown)


lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,84 @@
...
+    const std::string CRC64NVME_NAME("crc64nvme");
+
+    auto CRC64NVMEStrategy::init(boost::any& _context) const -> error
+    {
+        const size_t crc_bits = 64;
+        const uint64_t crc_polynomial = 0xAD93D23594C93659;
+        const uint64_t crc_initial_remainder = 0xFFFFFFFFFFFFFFFF;
+        const uint64_t crc_final_xor = 0xFFFFFFFFFFFFFFFF;
+        const bool crc_reflect_input = true;
+        const bool crc_reflect_output = true;
```
alanking: Could these be made `constexpr`?

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+        messageDigest = CRC64NVME_CHKSUM_PREFIX;
+        messageDigest += std::string((char*) out_buffer, out_len);
+
+        return SUCCESS();
+    }
+
+    bool CRC64NVMEStrategy::isChecksum(const std::string& _chksum) const
+    {
+        return boost::starts_with(_chksum, CRC64NVME_CHKSUM_PREFIX);
```
alanking: FWIW, `std::string` has a `starts_with` method now.

lib/hasher/include/irods/CRC64NVMEStrategy.hpp
```diff
@@ -0,0 +1,26 @@
+#ifndef _CRC64NVME_STRATEGY_HPP_
+#define _CRC64NVME_STRATEGY_HPP_
+
+#include "irods/HashStrategy.hpp"
```
alanking, line 4: Let's include these too:
```suggestion
#include "irods/HashStrategy.hpp"

#include <boost/any.hpp>

#include <string>
```

lib/hasher/include/irods/CRC64NVMEStrategy.hpp
```diff
@@ -0,0 +1,26 @@
+#ifndef _CRC64NVME_STRATEGY_HPP_
+#define _CRC64NVME_STRATEGY_HPP_
```
alanking: Maybe this would be better?
```suggestion
#ifndef IRODS_HASHER_CRC64NVME_STRATEGY_HPP
#define IRODS_HASHER_CRC64NVME_STRATEGY_HPP
```

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+        int len = strlen(CRC64NVME_CHKSUM_PREFIX);
+        unsigned long out_len = CHKSUM_LEN - len;
+
+        unsigned char out_buffer[CHKSUM_LEN];
+        base64_encode(bytes.data(), 8, out_buffer, &out_len);
+
+        // The checksum has been extracted from the context.  The context can be deleted.
+        delete context;
+        _context = nullptr;
```
alanking, line 69: Is this supposed to be:
```suggestion
        context = nullptr;
```

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+        messageDigest = CRC64NVME_CHKSUM_PREFIX;
+        messageDigest += std::string((char*) out_buffer, out_len);
```
alanking: Let's use `reinterpret_cast` (or maybe `static_cast`?) here instead of the C-style cast:
```suggestion
        messageDigest += std::string(reinterpret_cast<char*>(out_buffer), out_len);
```

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,84 @@
...
+            crc_polynomial, crc_initial_remainder, crc_final_xor, crc_reflect_input, crc_reflect_output);
+
+        _context = context;
+
+        return SUCCESS();
+    }
+
+    auto CRC64NVMEStrategy::update(const std::string& data, boost::any& _context) const -> error
+    {
+        const size_t crc_bits = 64;
```
alanking: Consider making this a private, static member of the class so that it can be accessed freely. I see that this gets defined in `init` and `digest` as well. Or better yet, can make a type declaration so that that can be used in lieu of `boost::crc_basic<crc_bits>`.

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,84 @@
...
+        std::array<unsigned char, 8> bytes;
+        for (int i = 0; i < 8; ++i) {
+            bytes[7 - i] = static_cast<unsigned char>((checksum >> (i * 8)) & 0xFF);
+        }
+
+        int len = strlen(CRC64NVME_CHKSUM_PREFIX);
+        unsigned long out_len = CHKSUM_LEN - len;
+
+        unsigned char out_buffer[CHKSUM_LEN];
+        base64_encode(bytes.data(), 8, out_buffer, &out_len);
```
alanking: Can use `bytes.size()` instead of 8 on L60 and L68.

If I'm reading this correctly, the 8 on L61 is not referring to the number of bytes but the number of bits in a byte, so that one can stay as an 8.

server/api/src/rsFileChksum.cpp
```diff
@@ -334,6 +335,35 @@ int file_checksum(RsComm* _comm,
...
+    if (ret_err.ok()) {
+        std::string checksum_from_resource;
+        std::string final_scheme_str(final_scheme);
+        ret_err = resc->call<const std::string*, std::string*>(
+            _comm, irods::RESOURCE_OP_READ_CHECKSUM, file_ptr, &final_scheme_str, &checksum_from_resource);
+        if (ret_err.ok()) {
+            strncpy(_calculated_checksum, checksum_from_resource.c_str(), NAME_LEN);
+            return 0;
+        }
+        else {
```
alanking: Technically don't need the `else` here since the `if` block has a `return` statement at its end.

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: done

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: done

lib/hasher/include/irods/CRC64NVMEStrategy.hpp
JustinKyleJames: done

lib/hasher/include/irods/CRC64NVMEStrategy.hpp
JustinKyleJames: done

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: I don't think so.  I am following the same pattern as in lib/hasher/src/SHA256Strategy.cpp.

I think the goal is to null out the context that is in the Hasher class defined in lib/hasherinclude/irods/Hasher.hpp.

If I null out `context` that is just a copy of the pointer.

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: done

server/api/src/rsFileChksum.cpp

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: I made it a static private member.

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: I used reinterpret_cast as static_cast can't force the unsigned char* to be char*.

alanking: Okay - I can never get the different casts straight in my head.

lib/hasher/src/CRC64NVMEStrategy.cpp
alanking: Ah, okay. Just wanted to make sure it wasn't a mistake.

JustinKyleJames:
> Looks good to me. Should we change the branch to main?

If I change it to main I get merge conflicts.  I've opened up a new branch called "crc64nvme_main" but not sure if that requires a new pull request or if we want that.

alanking:
If we can't switch the branch in this PR, opening a new one seems fine. If possible, keeping the diffs between the branches in a separate commit for reviewing purposes would be helpful, but I understand if it's too late for that.

JustinKyleJames:
Closing this pull request after opening #8701.

PR #8700, opened 2025-10-30
[8699] Make total quota update only apply to correct resc_id (main)

FifthPotato:
Seems to work,, though hopefully it doesn't mess up anything else quota-related.
From what I can tell, `mySQL3a` is only intended to update total quotas, so I ensured that it did so. I'm not sure why it didn't previously-- maybe there was some intent?

Either way, since the update only applied when it was a larger overrun, it probably went unnoticed since it didn't actually break anything.

Taking suggestions, as always. I also ran these against the existing quota tests and they seem to pass.

### DIFF
--- a/plugins/database/src/db_plugin.cpp
+++ b/plugins/database/src/db_plugin.cpp
@@ -1371,8 +1371,8 @@ int _modInheritance( int inheritFlag, int recursiveFlag, const char *collIdStr, 
 
 /*
   Set the over_quota values (if any) using the limits and
-  and the current usage; handling the various types: per-user per-resource,
-  per-user total-usage, group per-resource, and group-total.
+  and the current usage; handling the various types: 
+  group per-resource, and group-total.
 
   The over_quota column is positive if over_quota and the negative value
   indicates how much space is left before reaching the quota.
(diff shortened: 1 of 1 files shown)


plugins/database/src/db_plugin.cpp
trel:
```suggestion
and the current usage; handling the various types: group per-resource, and group-total.
```

trel: since we removed the user quotas, this comment resolves to only two scenarios.

trel:
i think update this makes sense.   for 'total', it should update the `resc_id = 0` row only.

it would be good to more thoroughly explain how it DID work for some scenarios a bit more...
it's not yet clear to me what this means...
> since the update only applied when it was a larger overrun

trel:
git blame on that line in the legacy codebase...

https://github.com/irods/irods-legacy/blame/master/iRODS/server/icat/src/icatHighLevelRoutines.c#L8972

I believe this is the commit (and reasoning) that introduced the bug...

https://github.com/irods/irods-legacy/commit/09d7e5d273efa23fea0fdcc9ee24b3d4ddfe9d40

---

Pretty sure I now agree that the fix in this PR is correct.

FifthPotato:
> Looks good to me. Should we make a test for this, or is that being handled in #8696?

The test in #8696 is what caught this in the first place, so it should be sufficient.



> i think update this makes sense. for 'total', it should update the `resc_id = 0` row only.
> 
> it would be good to more thoroughly explain how it DID work for some scenarios a bit more... it's not yet clear to me what this means...
> 
> > since the update only applied when it was a larger overrun

Essentially, this issue only went unnoticed because it only applied the total overrun numbers to the resource when the total overrun was greater (i.e. the overrun was more egregious -> the total quota was smaller than the resource quota). In practice, that means this error only happens when the total quota is more stringent than the resource quota, so it ended up applying first anyways, so the behavior always ended up correct, even if the math wasn't.

trel:
perfect - thanks.

FifthPotato:
If there are no major objections to my implementation, I can start running core tests on this.

trel:
definitely.

FifthPotato:
Core and unit tests pass. Waiting for the go-ahead to octothorpe.

korydraughn:
Given you've touched MySQL code, please run the quota tests against MySQL too.

FifthPotato:
> Given you've touched MySQL code, please run the quota tests against MySQL too.

Quotas tests passed on Ubuntu 22 - MySQL 8.4.

korydraughn:
Squash it too.

FifthPotato:
Squished and octothorpe'd.

alanking (approved):

alanking:
We can just ignore clang-format for this one - it's db_plugin.cpp.

korydraughn:
Right. The db plugin is more sensitive to formatting.

korydraughn:
@FifthPotato Please cherry-pick to 4-3-stable.

PR #8701, opened 2025-10-30
added CRC64/NVME as a valid checksum type (main)

JustinKyleJames:
Companion PR: https://github.com/irods/irods_resource_plugin_s3/pull/2270

This is to replace the pull request for #8698.

I ran the full suite of tests again for these commits.

---

From the original PR.

1. Added hash option for crc64nvme.
2. Added RESOURCE_OP_READ_CHECKSUM resource plugin operation.
3. Called the above operation before reading file to calculate checksum.  If it succeeds use the checksum returned from it.

### DIFF
--- a/scripts/irods/test/resource_suite.py
+++ b/scripts/irods/test/resource_suite.py
@@ -5,6 +5,7 @@ import datetime
 import filecmp
 import getpass
 import hashlib
+import inspect
 import itertools
 import os
 import psutil
@@ -495,6 +496,76 @@ class ResourceSuite(ResourceBase):
         # local cleanup
         if os.path.exists(datafilename):
             os.unlink(datafilename)
+
+    def test_server_side_calculates_crc64nvme_checksum_on_put_operation__issue_8554(self):
+        file1 = f'{inspect.currentframe().f_code.co_name}_f1'
+        file1_size = 2*1024
+
+        config = IrodsConfig()
+
+        try:
+            # create file1
+            lib.make_arbitrary_file(file1, file1_size)
+
+            with lib.file_backed_up(config.server_config_path):
+                config.server_config['default_hash_scheme'] = 'crc64nvme'
+                lib.update_json_file_from_dict(config.server_config_path, config.server_config)
+
+                with lib.file_backed_up(config.client_environment_path):
+                    client_update = {
+                        'irods_default_hash_scheme': 'crc64nvme',
+                    }
+                    session_env_backup = copy.deepcopy(self.admin.environment_file_contents)
+                    self.admin.environment_file_contents.update(client_update)
+
+                    IrodsController().reload_configuration()
+
+                    # calculate the crc64nvme checksum for comparison
+                    checksum_str = 'crc64nvme:'
+                    with open(file1, 'rb') as f:
+                        checksum_str += base64.b64encode(lib.calculate_crc64_nvme(f.read())).decode()
+
+                    self.admin.assert_icommand(['iput', '-k', file1])
+                    self.admin.assert_icommand(['ils', '-L', file1], 'STDOUT_SINGLELINE', checksum_str)
+        finally:
+            self.admin.assert_icommand(['irm', '-f', file1])
+            lib.remove_file_if_exists(file1)
+            IrodsController().reload_configuration()
+
+    def test_client_side_and_server_side_verify_crc64nvme_checksums_on_put_operation__issue_8554(self):
+        file1 = f'{inspect.currentframe().f_code.co_name}_f1'
+        file1_size = 2*1024
+
+        config = IrodsConfig()
+
+        try:
+            # create file1
+            lib.make_arbitrary_file(file1, file1_size)
+
+            with lib.file_backed_up(config.server_config_path):
+                config.server_config['default_hash_scheme'] = 'crc64nvme'
+                lib.update_json_file_from_dict(config.server_config_path, config.server_config)
+
+                with lib.file_backed_up(config.client_environment_path):
+                    client_update = {
+                        'irods_default_hash_scheme': 'crc64nvme',
+                    }
+                    session_env_backup = copy.deepcopy(self.admin.environment_file_contents)
+                    self.admin.environment_file_contents.update(client_update)
+
+                    IrodsController().reload_configuration()
+
+                    # calculate the crc64nvme checksum for comparison
+                    checksum_str = 'crc64nvme:'
+                    with open(file1, 'rb') as f:
+                        checksum_str += base64.b64encode(lib.calculate_crc64_nvme(f.read())).decode()
+
+                    self.admin.assert_icommand(['iput', '-K', file1])
+                    self.admin.assert_icommand(['ils', '-L', file1], 'STDOUT_SINGLELINE', checksum_str)
+        finally:
+            self.admin.assert_icommand(['irm', '-f', file1])
+            lib.remove_file_if_exists(file1)
+            IrodsController().reload_configuration()
 
     def test_local_iput_onto_specific_resource(self):
         # local setup
--- a/scripts/irods/test/test_ichksum.py
+++ b/scripts/irods/test/test_ichksum.py
@@ -12,6 +12,8 @@ import hashlib
 import base64
 import json
 import psutil
+import copy
+import inspect
 
 from . import session
 from . import settings
@@ -571,3 +573,38 @@ C- {5}:
         finally:
             os.unlink(local_file)
             self.admin.assert_icommand(['irm', '-f', object_name])
+
+    def test_ichksum_with_crc64nvme__issue_8554(self):
+        file1 = f'{inspect.currentframe().f_code.co_name}_f1'
+        file1_size = 2*1024
+
+        config = IrodsConfig()
+
+        try:
+            # create file1
+            lib.make_arbitrary_file(file1, file1_size)
+
+            with lib.file_backed_up(config.server_config_path):
+                config.server_config['default_hash_scheme'] = 'crc64nvme'
+                lib.update_json_file_from_dict(config.server_config_path, config.server_config)
+
+                with lib.file_backed_up(config.client_environment_path):
+                    client_update = {
+                        'irods_default_hash_scheme': 'crc64nvme',
+                    }
+                    session_env_backup = copy.deepcopy(self.admin.environment_file_contents)
+                    self.admin.environment_file_contents.update(client_update)
+
+                    IrodsController().reload_configuration()
+
+                    # calculate the crc64nvme checksum for comparison
+                    checksum_str = 'crc64nvme:'
+                    with open(file1, 'rb') as f:
+                        checksum_str += base64.b64encode(lib.calculate_crc64_nvme(f.read())).decode()
+
+                    self.admin.assert_icommand(['iput', file1])
+                    self.admin.assert_icommand(['ichksum', file1], 'STDOUT_SINGLELINE', checksum_str)
+        finally:
+            self.admin.assert_icommand(['irm', '-f', file1])
+            lib.remove_file_if_exists(file1)
+            IrodsController().reload_configuration()
--- a/scripts/irods/lib.py
+++ b/scripts/irods/lib.py
@@ -868,3 +868,28 @@ def set_replica_status(session, logical_path, replica_number, replica_status):
         'replica_number', str(replica_number),
         'DATA_REPL_STATUS', str(replica_status)
     ])
+
+def calculate_crc64_nvme(data: bytes, initial_crc = 0):
+    """
+    Calculates the CRC64-NVMe checksum for the given data.
+
+    Args:
+        data: The bytes object for which to calculate the CRC.
+        initial_crc: The starting CRC value for incremental calculations.
+                     Defaults to 0 for a fresh calculation.
+
+    Returns:
+        The 64-bit unsigned integer representing the CRC64-NVMe checksum.
+    """
+    polynomial = 0x9a6c9329ac4bc9b5
+    crc = initial_crc ^ 0xFFFFFFFFFFFFFFFF  # Initial XOR with all ones
+
+    for byte in data:
+        crc ^= byte
+        for _ in range(8):
+            if crc & 1:
+                crc = (crc >> 1) ^ polynomial
+            else:
+                crc >>= 1
+
+    return (crc ^ 0xFFFFFFFFFFFFFFFF).to_bytes(8) # Final XOR with all ones and return bytes 


lib/hasher/include/irods/CRC64NVMEStrategy.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+#include "irods/HashStrategy.hpp"
+
+#include <boost/any.hpp>
+#include <string>
+
+namespace irods
+{
+    extern const std::string CRC64NVME_NAME;
+    class CRC64NVMEStrategy : public HashStrategy
```
korydraughn, line 16: Let's add an empty line between lines 11 and 12.

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include <openssl/evp.h>
+
+#include <fmt/format.h>
+
+#include <boost/algorithm/string/predicate.hpp>
+#include <boost/crc.hpp>
+
+namespace irods
+{
+
```
korydraughn: Remove empty line.

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,81 @@
...
+    const std::string CRC64NVME_NAME("crc64nvme");
+
+    auto CRC64NVMEStrategy::init(boost::any& _context) const -> error
+    {
+        constexpr uint64_t crc_polynomial = 0xAD93D23594C93659;
+        constexpr uint64_t crc_initial_remainder = 0xFFFFFFFFFFFFFFFF;
+        constexpr uint64_t crc_final_xor = 0xFFFFFFFFFFFFFFFF;
```
korydraughn: Let's add some wording explaining why these values were selected.

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,81 @@
...
+        constexpr bool crc_reflect_input = true;
+        constexpr bool crc_reflect_output = true;
+
+        // Create the CRC calculator object
+        boost::crc_basic<crc_bits>* context = new boost::crc_basic<crc_bits>(
+            crc_polynomial, crc_initial_remainder, crc_final_xor, crc_reflect_input, crc_reflect_output);
```
korydraughn, line 43: Is there a reason why the context is heap allocated?

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    auto CRC64NVMEStrategy::digest(std::string& messageDigest, boost::any& _context) const -> error
+    {
+        boost::crc_basic<crc_bits>* context = boost::any_cast<boost::crc_basic<crc_bits>*>(_context);
+
+        uint64_t checksum = context->checksum();
+
+        // Convert uint64_t to bytes (big-endian for consistency)
+        std::array<unsigned char, 8> bytes;
```
korydraughn, line 87: Missing `<array>` header include.

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,81 @@
...
+        for (size_t i = 0; i < bytes.size(); ++i) {
+            bytes[7 - i] = static_cast<unsigned char>((checksum >> (i * 8)) & 0xFF);
+        }
+
+        int len = strlen(CRC64NVME_CHKSUM_PREFIX);
+        unsigned long out_len = CHKSUM_LEN - len;
+
+        unsigned char out_buffer[CHKSUM_LEN];
```
korydraughn: We use `std::array` earlier in the function and then a C-style array?

Please update this to use `std::array` or the previous line to be a C-style array for consistency.

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,81 @@
...
+        messageDigest += std::string(reinterpret_cast<char*>(out_buffer), out_len);
+
+        return SUCCESS();
+    }
+
+    bool CRC64NVMEStrategy::isChecksum(const std::string& _chksum) const
+    {
+        return _chksum.starts_with(CRC64NVME_CHKSUM_PREFIX);
+    }
+}; //namespace irods
```
korydraughn:
```suggestion
} // namespace irods
```

server/api/src/rsFileChksum.cpp
```diff
@@ -334,6 +335,33 @@ int file_checksum(RsComm* _comm,
     rodsLog(LOG_DEBUG, "file_checksum :: final_scheme [%s]  chkstr_scheme [%s]  hash_policy [%s]",
             final_scheme.data(), chkstr_scheme.c_str(), hash_policy.data());
 
+    irods::file_object_ptr file_ptr{
+        new irods::file_object{_comm, _logical_path, _filename, _resource_hierarchy, -1, 0, O_RDONLY}};
+
+    // ----------
```
korydraughn: Remove line.

server/api/src/rsFileChksum.cpp
```diff
@@ -334,6 +335,33 @@ int file_checksum(RsComm* _comm,
...
+            strncpy(_calculated_checksum, checksum_from_resource.c_str(), NAME_LEN);
+            return 0;
+        }
+        log_api::trace(fmt::format(
+            "{} ::  Negligible error calling RESOURCE_OP_READ_CHECKSUM most likely due to "
+            "checksum read not supported by resource.  Checksum will be calculated by a full file read.  {}",
+            __func__,
+            ret_err.result()));
+    }
+    // ----------
```
korydraughn: Remove line.

server/api/src/rsFileChksum.cpp
```diff
@@ -334,6 +335,33 @@ int file_checksum(RsComm* _comm,
...
+        if (ret_err.ok()) {
+            strncpy(_calculated_checksum, checksum_from_resource.c_str(), NAME_LEN);
+            return 0;
+        }
+        log_api::trace(fmt::format(
+            "{} ::  Negligible error calling RESOURCE_OP_READ_CHECKSUM most likely due to "
+            "checksum read not supported by resource.  Checksum will be calculated by a full file read.  {}",
+            __func__,
+            ret_err.result()));
+    }
```
korydraughn: This needs to go into the fileDriver API.

See the following:
- https://github.com/irods/irods/blob/0dd895c689e53dba3ba11c5d7e83231bac8985fb/server/drivers/include/irods/fileDriver.hpp
- https://github.com/irods/irods/blob/0dd895c689e53dba3ba11c5d7e83231bac8985fb/server/drivers/src/fileDriver.cpp

server/core/include/irods/irods_resource_constants.hpp
```diff
@@ -72,6 +72,7 @@ namespace irods
     extern const std::string RESOURCE_OP_RESOLVE_RESC_HIER;
     extern const std::string RESOURCE_OP_REBALANCE;
     extern const std::string RESOURCE_OP_NOTIFY;
+    extern const std::string RESOURCE_OP_READ_CHECKSUM;
```
korydraughn: Maybe this should be `RESOURCE_OP_READ_CHECKSUM_FROM_STORAGE_DEVICE`?

That makes it more specific and gives us more wiggle room for future checksum operations to be added to the resource plugins.

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include <iostream>
+#include <sstream>
+
+#include <openssl/err.h>
+#include <openssl/evp.h>
+
+#include <fmt/format.h>
+
+#include <boost/algorithm/string/predicate.hpp>
+#include <boost/crc.hpp>
```
korydraughn: Please move the standard C++ headers to the bottom of the list so that the ordering is consistent with other source files.

```cpp
#include "irods/rodsErrorTable.h"
#include "irods/checksum.h"
#include "irods/base64.hpp"

#include <openssl/err.h>
#include <openssl/evp.h>

#include <fmt/format.h>

#include <boost/algorithm/string/predicate.hpp>
#include <boost/crc.hpp>

#include <cstring>
#include <iomanip>
#include <iostream>
#include <sstream>
```

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include "irods/checksum.h"
+#include "irods/base64.hpp"
+
+#include <cstring>
+#include <iomanip>
+#include <iostream>
+#include <sstream>
+
+#include <openssl/err.h>
+#include <openssl/evp.h>
```
korydraughn: Are the OpenSSL headers actually used?

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+#include <fmt/format.h>
+
+#include <boost/algorithm/string/predicate.hpp>
```
korydraughn: This doesn't look like it's needed anymore.

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,81 @@
...
+        uint64_t checksum = context->checksum();
+
+        // Convert uint64_t to bytes (big-endian for consistency)
+        std::array<unsigned char, 8> bytes;
+        for (size_t i = 0; i < bytes.size(); ++i) {
+            bytes[7 - i] = static_cast<unsigned char>((checksum >> (i * 8)) & 0xFF);
+        }
+
+        int len = strlen(CRC64NVME_CHKSUM_PREFIX);
+        unsigned long out_len = CHKSUM_LEN - len;
```
korydraughn: Looks like these lines can be combined since `len` is not used elsewhere.

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,81 @@
+#include "irods/CRC64NVMEStrategy.hpp"
+
+#include "irods/rodsErrorTable.h"
+#include "irods/checksum.h"
+#include "irods/base64.hpp"
+
+#include <cstring>
+#include <iomanip>
+#include <iostream>
+#include <sstream>
```
korydraughn: These headers do not appear to be used.

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,81 @@
+#include "irods/CRC64NVMEStrategy.hpp"
+
+#include "irods/rodsErrorTable.h"
+#include "irods/checksum.h"
+#include "irods/base64.hpp"
```
korydraughn: Please sort for consistency with other source files.

lib/hasher/include/irods/CRC64NVMEStrategy.hpp
```diff
@@ -0,0 +1,32 @@
+#ifndef IRODS_HASHER_CRC64NVME_STRATEGY_HPP
+#define IRODS_HASHER_CRC64NVME_STRATEGY_HPP
+
+#include "irods/HashStrategy.hpp"
+
+#include <boost/any.hpp>
```
korydraughn, line 7: No need to include this header since `any` is part of the public interface of the base class.

lib/hasher/include/irods/CRC64NVMEStrategy.hpp
```diff
@@ -0,0 +1,32 @@
...
+#include <string>
+
+namespace irods
+{
+    extern const std::string CRC64NVME_NAME;
+    class CRC64NVMEStrategy : public HashStrategy
+    {
+      public:
+        CRC64NVMEStrategy(){};
+        virtual ~CRC64NVMEStrategy(){};
```
korydraughn:
```suggestion
        CRC64NVMEStrategy() {}
        virtual ~CRC64NVMEStrategy() {}
```

lib/hasher/include/irods/CRC64NVMEStrategy.hpp
JustinKyleJames: done

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: done

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: This context needs to survive longer than this function call.

This is similar to the others.
https://github.com/irods/irods/blob/b8781c11aef62b61b3ff3e29d17b52bb105fe372/lib/hasher/src/SHA256Strategy.cpp#L30
https://github.com/irods/irods/blob/b8781c11aef62b61b3ff3e29d17b52bb105fe372/lib/hasher/src/SHA256Strategy.cpp#L86

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: done

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: done

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: done

server/api/src/rsFileChksum.cpp

server/api/src/rsFileChksum.cpp

server/api/src/rsFileChksum.cpp
JustinKyleJames: @alanking - Would this break the redirection you were referring to earlier?

lib/hasher/src/CRC64NVMEStrategy.cpp
korydraughn: Sure. If the `boost:crc_basic<crc_bits>` type supports copy/move semantics, then you can avoid the heap allocation. That would also remove the need for anyone to protect against memory leaks.

With that said, code which relies on these hashers might assume data generated by them are heap allocated. If that's true, then we should continue with the existing approach.

I'm fine with your solution overall. Feel free to mark this as resolved.

server/core/include/irods/irods_resource_constants.hpp
JustinKyleJames: Done - also in S3 plugin.

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: done

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: No.  Removed.

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: done

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: removed

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: done

lib/hasher/include/irods/CRC64NVMEStrategy.hpp
JustinKyleJames: removed

lib/hasher/include/irods/CRC64NVMEStrategy.hpp
JustinKyleJames: done

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: removed

server/api/src/rsFileChksum.cpp
alanking: As long as the new fileDriver call is made from this API, the redirection should be handled. Take a look at `rsFileChksum` on L42 in this file and you'll see the redirection logic there. Once it has redirected to the appropriate server, it doesn't *really* matter how the resource operation is invoked other than convention and potentially code reuse, I guess.

server/core/src/irods_resource_constants.cpp
```diff
@@ -42,7 +42,7 @@ namespace irods
     const std::string RESOURCE_OP_RESOLVE_RESC_HIER( "resource_resolve_hierarchy" );
     const std::string RESOURCE_OP_REBALANCE( "resource_rebalance" );
     const std::string RESOURCE_OP_NOTIFY( "resource_notify" );
-    const std::string RESOURCE_OP_READ_CHECKSUM("resource_read_checksum");
+    const std::string RESOURCE_OP_READ_CHECKSUM_FROM_STORAGE_DEVICE("resource_read_checksum");
```
korydraughn: resource_read_checksum
```suggestion
    const std::string RESOURCE_OP_READ_CHECKSUM_FROM_STORAGE_DEVICE("resource_read_checksum_from storage_device");
```

server/api/src/rsFileChksum.cpp
```diff
@@ -349,18 +348,17 @@ int file_checksum(RsComm* _comm,
         std::string checksum_from_resource;
         std::string final_scheme_str(final_scheme);
         ret_err = resc->call<const std::string*, std::string*>(
-            _comm, irods::RESOURCE_OP_READ_CHECKSUM, file_ptr, &final_scheme_str, &checksum_from_resource);
+            _comm, irods::RESOURCE_OP_READ_CHECKSUM_FROM_STORAGE_DEVICE, file_ptr, &final_scheme_str, &checksum_from_resource);
         if (ret_err.ok()) {
             strncpy(_calculated_checksum, checksum_from_resource.c_str(), NAME_LEN);
             return 0;
         }
         log_api::trace(fmt::format(
```
korydraughn: The explicit call to `fmt::format` can be removed.

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,71 @@
...
+        // Convert uint64_t to bytes (big-endian for consistency)
+        std::array<unsigned char, 8> bytes;
+        for (size_t i = 0; i < bytes.size(); ++i) {
+            bytes[7 - i] = static_cast<unsigned char>((checksum >> (i * 8)) & 0xFF);
+        }
+
+        unsigned long out_len = CHKSUM_LEN - strlen(CRC64NVME_CHKSUM_PREFIX);
+
+        std::array<unsigned char, CHKSUM_LEN> out_buffer;
+        base64_encode(bytes.data(), bytes.size(), out_buffer.data(), &out_len);
```
SwooshyCueb: Check the return value of `base64_encode` for errors. (See #8704)

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,71 @@
+#include "irods/CRC64NVMEStrategy.hpp"
+
+#include "irods/base64.hpp"
+#include "irods/checksum.h"
+#include "irods/rodsErrorTable.h"
+
+#include <boost/crc.hpp>
+
+#include <array>
+#include <cstring>
```
SwooshyCueb:
```suggestion
#include <cstdint>
#include <cstring>
```

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,71 @@
...
+        boost::crc_basic<crc_bits>* context = boost::any_cast<boost::crc_basic<crc_bits>*>(_context);
+        context->process_bytes(data.c_str(), data.length());
+        return SUCCESS();
+    }
+
+    auto CRC64NVMEStrategy::digest(std::string& messageDigest, boost::any& _context) const -> error
+    {
+        boost::crc_basic<crc_bits>* context = boost::any_cast<boost::crc_basic<crc_bits>*>(_context);
+
+        uint64_t checksum = context->checksum();
```
SwooshyCueb:
```suggestion
        std::uint64_t checksum = context->checksum();
```

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+namespace irods
+{
+    const std::string CRC64NVME_NAME("crc64nvme");
+
+    auto CRC64NVMEStrategy::init(boost::any& _context) const -> error
+    {
+        constexpr uint64_t crc_polynomial = 0xAD93D23594C93659;
+        constexpr uint64_t crc_initial_remainder = 0xFFFFFFFFFFFFFFFF;
+        constexpr uint64_t crc_final_xor = 0xFFFFFFFFFFFFFFFF;
```
SwooshyCueb:
```suggestion
        constexpr std::uint64_t crc_polynomial = 0xAD93D23594C93659;
        constexpr std::uint64_t crc_initial_remainder = 0xFFFFFFFFFFFFFFFF;
        constexpr std::uint64_t crc_final_xor = 0xFFFFFFFFFFFFFFFF;
```

lib/hasher/include/irods/CRC64NVMEStrategy.hpp
```diff
@@ -0,0 +1,32 @@
+#ifndef IRODS_HASHER_CRC64NVME_STRATEGY_HPP
+#define IRODS_HASHER_CRC64NVME_STRATEGY_HPP
+
+#include "irods/HashStrategy.hpp"
+
+#include <string>
```
SwooshyCueb, line 10:
```suggestion
#include "irods/HashStrategy.hpp"
#include "irods/irods_error.hpp"

#include <boost/any.hpp>

#include <cstdlib>
#include <string>
```

lib/hasher/include/irods/CRC64NVMEStrategy.hpp
```diff
@@ -0,0 +1,32 @@
...
+        {
+            return CRC64NVME_NAME;
+        }
+        error init(boost::any& context) const override;
+        error update(const std::string&, boost::any& context) const override;
+        error digest(std::string& messageDigest, boost::any& context) const override;
+        bool isChecksum(const std::string&) const override;
+
+      private:
+        static constexpr size_t crc_bits = 64;
```
SwooshyCueb:
```suggestion
        static constexpr std::size_t crc_bits = 64;
```

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+    auto CRC64NVMEStrategy::digest(std::string& messageDigest, boost::any& _context) const -> error
+    {
+        boost::crc_basic<crc_bits>* context = boost::any_cast<boost::crc_basic<crc_bits>*>(_context);
+
+        uint64_t checksum = context->checksum();
+
+        // Convert uint64_t to bytes (big-endian for consistency)
+        std::array<unsigned char, 8> bytes;
+        for (size_t i = 0; i < bytes.size(); ++i) {
```
SwooshyCueb:
```suggestion
        for (std::size_t i = 0; i < bytes.size(); ++i) {
```

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,71 @@
...
+            bytes[7 - i] = static_cast<unsigned char>((checksum >> (i * 8)) & 0xFF);
+        }
+
+        unsigned long out_len = CHKSUM_LEN - strlen(CRC64NVME_CHKSUM_PREFIX);
```
SwooshyCueb:
```suggestion
        unsigned long out_len = CHKSUM_LEN - std::strlen(CRC64NVME_CHKSUM_PREFIX);
```

server/api/src/rsFileChksum.cpp
```diff
@@ -334,6 +335,31 @@ int file_checksum(RsComm* _comm,
...
+    irods::resource_ptr resc;
+    irods::error ret_err = file_ptr->resolve(irods::RESOURCE_INTERFACE, ptr);
+    resc = boost::dynamic_pointer_cast<irods::resource>(ptr);
+    if (ret_err.ok()) {
+        std::string checksum_from_resource;
+        std::string final_scheme_str(final_scheme);
+        ret_err = resc->call<const std::string*, std::string*>(
+            _comm, irods::RESOURCE_OP_READ_CHECKSUM_FROM_STORAGE_DEVICE, file_ptr, &final_scheme_str, &checksum_from_resource);
+        if (ret_err.ok()) {
+            strncpy(_calculated_checksum, checksum_from_resource.c_str(), NAME_LEN);
```
SwooshyCueb: Add `<cstring>` include and
```suggestion
            std::strncpy(_calculated_checksum, checksum_from_resource.c_str(), NAME_LEN);
```

lib/hasher/src/CRC64NVMEStrategy.cpp
SwooshyCueb, line 11:
```suggestion
#include "irods/base64.hpp"
#include "irods/checksum.h"
#include "irods/irods_error.hpp"
#include "irods/rodsErrorTable.h"

#include <boost/any.hpp>
#include <boost/crc.hpp>

#include <array>
#include <cstdint>
#include <cstring>
```

server/api/src/rsFileChksum.cpp
JustinKyleJames: > As long as the new fileDriver call is made from this API, the redirection should be handled. Take a look at `rsFileChksum` on L42 in this file and you'll see the redirection logic there. Once it has redirected to the appropriate server, it doesn't _really_ matter how the resource operation is invoked other than convention and potentially code reuse, I guess.

Yes, I didn't get exactly what Kory was asking me to do.

server/core/src/irods_resource_constants.cpp
JustinKyleJames: done

server/api/src/rsFileChksum.cpp

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: done

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: done

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: done

lib/hasher/include/irods/CRC64NVMEStrategy.hpp
JustinKyleJames: done

lib/hasher/include/irods/CRC64NVMEStrategy.hpp
JustinKyleJames: done

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: done

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: done

server/api/src/rsFileChksum.cpp

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: This contradicts https://github.com/irods/irods/pull/8701#discussion_r2479225015.

I added the `cstdint` include.

@korydraughn - thoughts

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: Done - with help of chatgpt.

trel: looks good

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -41,18 +42,22 @@ namespace irods
...
-        unsigned long out_len = CHKSUM_LEN - strlen(CRC64NVME_CHKSUM_PREFIX);
+        unsigned long out_len = CHKSUM_LEN - std::strlen(CRC64NVME_CHKSUM_PREFIX);
 
         std::array<unsigned char, CHKSUM_LEN> out_buffer;
-        base64_encode(bytes.data(), bytes.size(), out_buffer.data(), &out_len);
+        int rc = base64_encode(bytes.data(), bytes.size(), out_buffer.data(), &out_len);
+		if (0 != rc) {
+			const auto msg = fmt::format("{}: Failed to b64 encode hash.", __func__);
+			return ERROR(rc, msg);
+		}
```
alanking: I'm guessing `delete context` is omitted here on purpose?

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,92 @@
...
+        messageDigest += std::string(reinterpret_cast<char*>(out_buffer.data()), out_len);
+
+        return SUCCESS();
+    }
+
+    bool CRC64NVMEStrategy::isChecksum(const std::string& _chksum) const
+    {
+        return _chksum.starts_with(CRC64NVME_CHKSUM_PREFIX);
+    }
+}; // namespace irods
```
korydraughn:
```suggestion
} // namespace irods
```

server/drivers/include/irods/fileDriver.hpp
```diff
@@ -44,4 +44,5 @@ irods::error fileRegistered( rsComm_t* _comm, irods::first_class_object_ptr _obj
 irods::error fileUnregistered( rsComm_t* _comm, irods::first_class_object_ptr _object );
 irods::error fileModified( rsComm_t* _comm, irods::first_class_object_ptr _object );
 irods::error fileNotify( rsComm_t* _comm, irods::first_class_object_ptr _object , const std::string& );
+irods::error fileChecksum( rsComm_t* _comm, irods::first_class_object_ptr _object , const std::string&, std::string&);
```
korydraughn: Why `fileChecksum` instead of `fileReadChecksumFromStorageDevice`?

JustinKyleJames: Fair.  Will change.

JustinKyleJames: done

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: done

alanking (commented):
Had one question

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,92 @@
...
+        for (std::size_t i = 0; i < bytes.size(); ++i) {
+            bytes[7 - i] = static_cast<unsigned char>((checksum >> (i * 8)) & 0xFF);
+        }
+
+        unsigned long out_len = CHKSUM_LEN - std::strlen(CRC64NVME_CHKSUM_PREFIX);
+
+        std::array<unsigned char, CHKSUM_LEN> out_buffer;
+        int rc = base64_encode(bytes.data(), bytes.size(), out_buffer.data(), &out_len);
+		if (0 != rc) {
+			const auto msg = fmt::format("{}: Failed to b64 encode hash.", __func__);
```
korydraughn: b64 -> base64

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: Yes, it is.

Just kidding.  Embarrassingly no.  Done

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: Done

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -49,14 +49,37 @@ namespace irods
...
+        try {
+            context = boost::any_cast<boost::crc_basic<crc_bits>*>(_context);
+        }
+        catch (boost::bad_any_cast& e) {
+            return ERROR(SYS_INVALID_INPUT_PARAM,
+                         "The context sent to CRC64NVMEStrategy::update was not a boost::crc_basic<crc_bits>*");
+        }
+
+        if (nullptr == context) {
+            return ERROR(INVALID_INPUT_ARGUMENT_NULL_POINTER, "A null context was send to CRC64NVMEStrategy::digest.");
```
alanking:
```suggestion
            return ERROR(INVALID_INPUT_ARGUMENT_NULL_POINTER, "A null context was send to CRC64NVMEStrategy::update.");
```

lib/hasher/src/CRC64NVMEStrategy.cpp
alanking, line 97: After further discussion / thought, I'm now wondering whether we ought to free `context` here. The question is... should an error indicate to the caller that the context needs to be freed? I know @SwooshyCueb is making similar changes for the other hashers, so maybe he can weigh in.

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: I literally changed that from digest to update several times.  Must have copied/pasted since the last time.  Done.

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: Looking at `base64_encode` the only way it fails is if the buffer (size identified by `out_len`) was not large enough.

It isn't possible to call this a second time and get a different result since the following would always give the same result:

```
unsigned long out_len = CHKSUM_LEN - std::strlen(CRC64NVME_CHKSUM_PREFIX);
```

So in my eyes is makes sense to clean up the memory.  If the caller happened to try again, they are just going to get a null pointer error returned.

I'll wait to see if @SwooshyCueb has a different opinion on this.

SwooshyCueb: Yeah I think this should be cleaned up before returning. In the hashers that use openssl, it's freed before the call to `base64_encode`.

JustinKyleJames: I am resolving this.

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,117 @@
...
+        try {
+            context = boost::any_cast<boost::crc_basic<crc_bits>*>(_context);
+        }
+        catch (boost::bad_any_cast& e) {
+            return ERROR(SYS_INVALID_INPUT_PARAM,
+                         "The context sent to CRC64NVMEStrategy::update was not a boost::crc_basic<crc_bits>*");
+        }
+
+        if (nullptr == context) {
+            return ERROR(INVALID_INPUT_ARGUMENT_NULL_POINTER, "A null context was send to CRC64NVMEStrategy::update.");
```
korydraughn: send -> sent

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,117 @@
...
+        try {
+            context = boost::any_cast<boost::crc_basic<crc_bits>*>(_context);
+        }
+        catch (boost::bad_any_cast& e) {
+            return ERROR(SYS_INVALID_INPUT_PARAM,
+                         "The context sent to CRC64NVMEStrategy::digest was not a boost::crc_basic<crc_bits>*");
+        }
+
+        if (nullptr == context) {
+            return ERROR(INVALID_INPUT_ARGUMENT_NULL_POINTER, "A null context was send to CRC64NVMEStrategy::digest.");
```
korydraughn: send -> sent

server/api/src/rsFileChksum.cpp
korydraughn: Your comment states the work was done, but `fmt::format` is still in the code?

Perhaps the change wasn't pushed?

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+        unsigned long out_len = CHKSUM_LEN - std::strlen(CRC64NVME_CHKSUM_PREFIX);
+
+        std::array<unsigned char, CHKSUM_LEN> out_buffer;
+        int rc = base64_encode(bytes.data(), bytes.size(), out_buffer.data(), &out_len);
+        if (0 != rc) {
+            delete context;
+            _context = nullptr;
+            const auto msg = fmt::format("{}: Failed to base64 encode hash.", __func__);
+            return ERROR(rc, msg);
```
korydraughn: These lines can be merged together.

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,117 @@
...
+        return SUCCESS();
+    }
+
+    auto CRC64NVMEStrategy::update(const std::string& data, boost::any& _context) const -> error
+    {
+        boost::crc_basic<crc_bits>* context;
+        try {
+            context = boost::any_cast<boost::crc_basic<crc_bits>*>(_context);
+        }
+        catch (boost::bad_any_cast& e) {
```
korydraughn: Needs more `const`.

lib/hasher/src/CRC64NVMEStrategy.cpp
```diff
@@ -0,0 +1,117 @@
...
+        return SUCCESS();
+    }
+
+    auto CRC64NVMEStrategy::digest(std::string& messageDigest, boost::any& _context) const -> error
+    {
+        boost::crc_basic<crc_bits>* context;
+        try {
+            context = boost::any_cast<boost::crc_basic<crc_bits>*>(_context);
+        }
+        catch (boost::bad_any_cast& e) {
```
korydraughn: Missing `const`.

alanking (commented):
Seems good to me. Maybe ready for squashing?

server/api/src/rsFileChksum.cpp
JustinKyleJames: Maybe it wasn't.  I did it now.

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: done

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: done

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: done

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: done

lib/hasher/src/CRC64NVMEStrategy.cpp
JustinKyleJames: done

JustinKyleJames:
I added three test cases for CRC64NVME.  This are similar to the tests that are in the S3 resource plugin.

These are in:
- scripts/irods/test/resource_suite.py
- scripts/irods/test/test_ichksum.py

I tested these with the following but will kick off a full suite:

```
python3 scripts/run_tests.py --run_s=test_resource_types.Test_Resource_Unixfilesystem.test_iput_with_crc64nvme_lowercase_checksum_flag
python3 scripts/run_tests.py --run_s=test_resource_types.Test_Resource_Unixfilesystem.test_iput_with_crc64nvme_uppercase_checksum_flag
python3 scripts/run_tests.py --run_s=test_ichksum.Test_Ichksum.test_ichksum_with_crc64nvme
```
I also added in a `calculate_crc64_nvme()` method into lib.py.  This is also a copy of the one in s3plugin_lib.py.  I wasn't sure if I should remove the latter or keep it so I kept it for now.

Finally, the imports are not sorted in test_ichksum.py so I just added the two new ones at the bottom.

JustinKyleJames:
> > I wasn't sure if I should remove the latter or keep it so I kept it for now.
> 
> To me, this depends on whether the next S3 plugin release needs to be compatible with 5.0 in addition to 5.1. If so, the new function needs to be in both. If it is only going to be compatible with 5.1, then we only need the one in `lib.py`.

It has to have the CRC/NVME core code otherwise these tests will fail anyway. So I am thinking remove from s3 lib and use the one in lib.py.

JustinKyleJames:
I ran a full suite of core tests after the test case updates and it passed.

server/api/src/rsFileChksum.cpp
```diff
@@ -334,6 +336,26 @@ int file_checksum(RsComm* _comm,
...
+    irods::file_object_ptr file_ptr{
+        new irods::file_object{_comm, _logical_path, _filename, _resource_hierarchy, -1, 0, O_RDONLY}};
+
+    // Ask resource server if it can read the checksum directly, any error would go
+    // through the normal file read
+    std::string checksum_from_resource;
+    std::string final_scheme_str(final_scheme);
+    irods::error ret_err = fileChecksumFromStorageDevice(_comm, file_ptr, final_scheme_str, checksum_from_resource);
+    if (ret_err.ok()) {
+        std::strncpy(_calculated_checksum, checksum_from_resource.c_str(), NAME_LEN);
```
korydraughn: `NAME_LEN` -> `CHKSUM_LEN`

Based on @SwooshyCueb's checksum PR, these macros share the same value, so there's no need to rerun tests.

Please double check the values are identical in case I'm getting the macros mixed up.

scripts/irods/test/resource_suite.py
```diff
@@ -496,6 +497,78 @@ def test_local_iput_upper_checksum(self):
...
+        file1_size = 2*1024
+
+        config = IrodsConfig()
+
+        try:
+            # create file1
+            lib.make_arbitrary_file(file1, file1_size)
+
+            with lib.file_backed_up(config.server_config_path):
+
```
korydraughn: Remove empty line to match other uses of `with`.

scripts/irods/test/resource_suite.py
```diff
@@ -496,6 +497,78 @@ def test_local_iput_upper_checksum(self):
         if os.path.exists(datafilename):
             os.unlink(datafilename)
 
+    def test_iput_with_crc64nvme_lowercase_checksum_flag(self):
```
korydraughn: Prefer capturing what the flag is doing instead of identifying the flag.

Here's an alternative.

```suggestion
    def test_server_side_calculates_crc64nvme_checksum_on_put_operation__issue_NNNN(self):
```

scripts/irods/test/resource_suite.py
```diff
@@ -496,6 +497,78 @@ def test_local_iput_upper_checksum(self):
...
+                        checksum_str += base64.b64encode(lib.calculate_crc64_nvme(f.read())).decode()
+
+                    self.admin.assert_icommand(['iput', '-k', file1])
+                    self.admin.assert_icommand(['ils', '-L', file1], 'STDOUT_SINGLELINE', checksum_str)
+        finally:
+            self.admin.assert_icommand(['irm', '-f', file1])
+            lib.remove_file_if_exists(file1)
+            IrodsController().reload_configuration()
+
+    def test_iput_with_crc64nvme_uppercase_checksum_flag(self):
```
korydraughn: Prefer capturing what the flag is doing instead of identifying the flag.

Here's an alternative.

```suggestion
    def test_client_side_and_server_side_verify_crc64nvme_checksums_on_put_operation__issue_NNNN(self):
```

scripts/irods/test/resource_suite.py
```diff
@@ -496,6 +497,78 @@ def test_local_iput_upper_checksum(self):
...
+        file1_size = 2*1024
+
+        config = IrodsConfig()
+
+        try:
+            # create file1
+            lib.make_arbitrary_file(file1, file1_size)
+
+            with lib.file_backed_up(config.server_config_path):
+
```
korydraughn: Remove empty line to match other uses of `with`.

scripts/irods/test/test_ichksum.py
```diff
@@ -571,3 +573,39 @@ def test_ichksum_server_rejects_invalid_server_config_options_for_checksum_read_
         finally:
             os.unlink(local_file)
             self.admin.assert_icommand(['irm', '-f', object_name])
+
+    def test_ichksum_with_crc64nvme(self):
```
korydraughn: Missing issue number: `__issue_NNNN`

scripts/irods/test/test_ichksum.py
```diff
@@ -571,3 +573,39 @@ def test_ichksum_server_rejects_invalid_server_config_options_for_checksum_read_
...
+        file1_size = 2*1024
+
+        config = IrodsConfig()
+
+        try:
+            # create file1
+            lib.make_arbitrary_file(file1, file1_size)
+
+            with lib.file_backed_up(config.server_config_path):
+
```
korydraughn: Remove empty line to match other uses of `with`.

korydraughn:
Have you tested things in a topology, with redirection?

server/api/src/rsFileChksum.cpp

scripts/irods/test/resource_suite.py
JustinKyleJames: I like the empty line there but will remove.

scripts/irods/test/resource_suite.py
JustinKyleJames: done

scripts/irods/test/test_ichksum.py
JustinKyleJames: done

scripts/irods/test/resource_suite.py
JustinKyleJames: done

scripts/irods/test/test_ichksum.py
JustinKyleJames: I thought we only did that for bugs and not enhancements.

Made the change.

scripts/irods/test/resource_suite.py
JustinKyleJames: done

scripts/irods/test/resource_suite.py
JustinKyleJames: done

JustinKyleJames:
> Have you tested things in a topology, with redirection?

I just tested it and it worked.

1. Resource attached to provider, client attached to provider, iput -K with CRC64NVME.
2. Resource attached to provider, client attached to consumer, iput -K with CRC64NVME.
3. Resource attached to consumer, client attached to provider, iput -K with CRC64NVME.
4. Resource attached to consumer, client attached to consumer, iput -K with CRC64NVME.

korydraughn:
Anything else left to do for this?

SwooshyCueb:
If this gets merged before my PR, I'll update my PR to add relevant cases to the unit test I've added.

korydraughn:
Commit message for 1e525cc references incorrect resource op name. Let's include more information about the new resource op in the body of the commit message. That specific change is important to tell a story about.

For 69110ae, can capitalize the beginning of the first word (`added` -> `Added`) to match the other commit messages.

We need a PR which adds the resource PEPs to https://github.com/irods/irods_docs/blob/925e04dcd3f179c961704ff9d7abd6abc3eebb04/dynamic_peps.json too.

JustinKyleJames:
> Commit message for [1e525cc](https://github.com/irods/irods/commit/1e525ccea420d681a107037a22f836c197502e9f) references incorrect resource op name. Let's include more information about the new resource op in the body of the commit message. That specific change is important to tell a story about.
> 
> For [69110ae](https://github.com/irods/irods/commit/69110ae5e5c4f15c43b9893ef0c88c64059a2dbd), can capitalize the beginning of the first word (`added` -> `Added`) to match the other commit messages.
> 
> We need a PR which adds the resource PEPs to https://github.com/irods/irods_docs/blob/925e04dcd3f179c961704ff9d7abd6abc3eebb04/dynamic_peps.json too.

I made the suggested changes. I also limited the new python code to 80 characters wide as you asked for in the other PR. That is a new commit that will need to be squashed.

We need to discuss the PR for the PEP.

korydraughn:
My bad on the line width comment. I was referring to the commit message, not the code. So you can revert the changes to the test or keep them if you like.

JustinKyleJames:
> My bad on the line width comment. I was referring to the commit message, not the code. So you can revert the changes to the test or keep them if you like.

Actually, my bad.  I interpreted the comment as a commit message at first but was only looking at the first line.  I deleted/forced pushed the Python code formatting changes.

korydraughn:
Merging.

Issue #8703, opened 2025-11-04
sha512 hasher seems to be broken


SwooshyCueb:
The sha512 hasher seems to be broken. Setting the `default_hash_scheme` to `SHA512` and then attempting to use `ichksum` either results in `CAT_SQL_ERR` or a garbage value for the checksum. I even got it to give me a demangled symbol name as the hash value once.

PR #8704, opened 2025-11-04
[#8703,#8731,#8734,#8740] Add missing hasher error handling, increase buffer size for sha512 hasher, add hashers unit test (main)

SwooshyCueb:
Addresses #8703
Addresses #8731 
Addresses #8734
Addresses #8740

Discovered this while testing irods/irods_client_globus_connector#127

The sha512 hasher doesn't work, because `CHKSUM_LEN` isn't big enough, causing `base64_encode` to fail. `base64_encode` *does* return an error in this case, but we don't check it in `::digest` calls. What's more, `::digest` also returns an error, but in most cases where it's called *we don't check the return value there either*. On top of that, we don't have tests for the hashers either.

This PR adds the missing error handling, increases the buffer size for the sha512 hasher, and adds a unit test for the hashers.
I've additionally swapped usage of `NAME_LEN` for `CHKSUM_LEN` where appropriate.

### DIFF
--- a/lib/api/include/irods/chkNVPathPerm.h
+++ b/lib/api/include/irods/chkNVPathPerm.h
@@ -1,8 +1,12 @@
 #ifndef CHK_NV_PATH_PERM_H__
 #define CHK_NV_PATH_PERM_H__
 
+// Remove in iRODS 6.0
+#if __has_include("irods/rodsConnect.h")
+#  include "irods/rodsConnect.h"
+#endif
+
 #include "irods/fileOpen.h"
-#include "irods/rodsConnect.h"
 #include "irods/rcConnect.h"
 
 #ifdef __cplusplus
--- a/lib/api/include/irods/execCmd.h
+++ b/lib/api/include/irods/execCmd.h
@@ -1,7 +1,11 @@
 #ifndef EXEC_CMD_H__
 #define EXEC_CMD_H__
 
-#include "irods/rodsConnect.h"
+// Remove in iRODS 6.0
+#if __has_include("irods/rodsConnect.h")
+#  include "irods/rodsConnect.h"
+#endif
+
 #include "irods/procApiRequest.h"
 #include "irods/dataObjInpOut.h"
 


SwooshyCueb:
All tests passing, marking ready for reveiw.

clang-tidy and clang-format aren't happy, but I've already made all the changes in that department that I intend to. This PR makes a lot of tweaks to existing code and I'm not interested in doing a refactor in this PR.

server/api/src/rsBulkDataObjPut.cpp
```diff
@@ -600,7 +600,7 @@ bulkProcAndRegSubfile( rsComm_t *rsComm, const char *_resc_name, const std::stri
         }
         else {
             if ( ( flags & VERIFY_CHKSUM_FLAG ) != 0 && myChksum != NULL ) {
-                char chksumStr[NAME_LEN];
+                char chksumStr[LONG_CHKSUM_LEN];
```
korydraughn: `LONG_CHKSUM_LEN`? Shouldn't this be `CHKSUM_LEN`?

lib/core/include/irods/rodsPackInstruct.h
```diff
@@ -59,7 +59,7 @@
 
 #define PortList_PI "int portNum; int cookie; int sock; int windowSize; str hostAddr[LONG_NAME_LEN];"
 
-#define PortalOprOut_PI "int status; int l1descInx; int numThreads; str chksum[NAME_LEN]; struct PortList_PI;"
+#define PortalOprOut_PI "int status; int l1descInx; int numThreads; str chksum[CHKSUM_LEN]; struct PortList_PI;"
```
korydraughn: Is there any reason to think that changing a packing instruction in this way will break anything?

We might need to hold off of changing packing instructions until we know more.

SwooshyCueb: `NAME_LEN` and `CHKSUM_LEN` are both defined to the same value, and are both in `PackConstantTable`.

server/api/src/rsBulkDataObjPut.cpp
SwooshyCueb: In this particular case, increasing the size of the buffer doesn't result in any incompatibilities, as far as I can tell.

lib/core/include/irods/rodsPackInstruct.h
korydraughn: While I don't think this would break compatibility, we don't have a way to confirm that at this time. Let's hold on changing the packing instructions and data types until iRODS 6.

lib/hasher/src/SHA1Strategy.cpp
```diff
@@ -90,7 +90,11 @@ namespace irods {
         unsigned long out_len = CHKSUM_LEN - len;
 
         unsigned char out_buffer[CHKSUM_LEN];
-        base64_encode( final_buffer, SHA_DIGEST_LENGTH, out_buffer, &out_len );
+        int rc = base64_encode(final_buffer, SHA_DIGEST_LENGTH, out_buffer, &out_len);
+        if (0 != rc) {
+            const auto msg = fmt::format("{}: Failed to b64 encode hash.", __func__);
```
alanking:
```suggestion
            const auto msg = fmt::format("{}: Failed to base64 encode hash.", __func__);
```

lib/hasher/src/SHA256Strategy.cpp
```diff
@@ -90,7 +90,11 @@ namespace irods {
         unsigned long out_len = CHKSUM_LEN - len;
 
         unsigned char out_buffer[CHKSUM_LEN];
-        base64_encode(final_buffer, SHA256_DIGEST_LENGTH, out_buffer, &out_len);
+        int rc = base64_encode(final_buffer, SHA256_DIGEST_LENGTH, out_buffer, &out_len);
+        if (0 != rc) {
+            const auto msg = fmt::format("{}: Failed to b64 encode hash.", __func__);
```
alanking:
```suggestion
            const auto msg = fmt::format("{}: Failed to base64 encode hash.", __func__);
```

lib/hasher/src/SHA512Strategy.cpp
```diff
@@ -90,7 +90,11 @@ namespace irods {
         unsigned long out_len = CHKSUM_LEN - len;
 
         unsigned char out_buffer[CHKSUM_LEN];
-        base64_encode(final_buffer, SHA512_DIGEST_LENGTH, out_buffer, &out_len);
+        int rc = base64_encode(final_buffer, SHA512_DIGEST_LENGTH, out_buffer, &out_len);
+        if (0 != rc) {
+            const auto msg = fmt::format("{}: Failed to b64 encode hash.", __func__);
```
alanking:
```suggestion
            const auto msg = fmt::format("{}: Failed to base64 encode hash.", __func__);
```

plugins/rule_engines/irods_rule_language/src/configuration.cpp
```diff
@@ -437,14 +452,24 @@ int hash_rules_with_copy(const std::vector<std::string>& irods_rule_bases,
...
         try {
-            hasher.update(copy.get_rulebase(irb));
+            ret = hasher.update(copy.get_rulebase(irb));
         }
         catch (std::out_of_range& ex) {
             log_re::warn("hash_rules_with_copy: attempted to access nonexistent rulebase [{}]", irb);
         }
+        if (!ret.ok()) {
+            int status = static_cast<int>(ret.code());
+            log_re::error("{}: failed on hasher update, status = {}", __func__, status);
```
alanking: Might be nice to include `irb` in the log message to indicate which rulebase failed. Up to you

unit_tests/src/test_hashers.cpp
```diff
@@ -0,0 +1,117 @@
...
+#include <fmt/format.h>
+
+#include <string>
+#include <tuple>
+#include <vector>
+
+TEST_CASE("checksum hashers", "[string]")
+{
+    // clang-format off
+    const std::tuple<const std::string, const std::string, const std::string> hash_test_vals = GENERATE(
```
korydraughn, line 21: Interesting. So `GENERATE` is a Catch2 feature.

When running the test with `-s`, do you see all cases in the output?

lib/core/include/irods/rodsPackInstruct.h
alanking: Are we technically *changing* anything here? They are, after all, the same value, as @SwooshyCueb pointed out:
https://github.com/irods/irods/blob/6bc007183bc10ae50f7938da4a2c395e9c6b2ffb/lib/core/include/irods/rodsDef.h#L39-L40

Really asking, by the way. If the packing instruction macro string changes even if the resultant data does not change (and all referenced macros therein already existed in the client library and all past supported server versions vis-a-vis federation), does it actually change anything?

lib/core/include/irods/rodsDef.h
```diff
@@ -38,6 +38,7 @@
 #define TIME_LEN                        32
 #define NAME_LEN                        64
 #define CHKSUM_LEN                      64  /* keep chksum_len the same as the old size */
+#define LONG_CHKSUM_LEN                 128
```
alanking, line 41: Just curious... Why use 128 instead of `CHKSUM_LEN * 2` like it was before it moved in here?

lib/hasher/src/checksum.cpp
```diff
@@ -135,13 +138,21 @@ int chksumLocFile(
     buffer_read.resize( HASH_BUF_SZ );
 
     while ( in_file.read( &buffer_read[0], HASH_BUF_SZ ) ) {
-        hasher.update( buffer_read );
+        ret = hasher.update(buffer_read);
+        if (!ret.ok()) {
+            log_api::error("{}: error on hasher update, result = {}", __func__, ret.result());
```
korydraughn: I'm wondering if these log messages will be a problem in client-side code.

The logger is designed for server-side use only at the moment. This source file is compiled as part of an object file with (PP macro) `IRODS_ENABLE_SYSLOG` defined, yet there is no initialization of the logger performed by any C++ client.

Can you think of a way to trigger one of these code paths in client-side code to see what happens?

lib/core/include/irods/rodsPackInstruct.h
korydraughn: The proposed change is correct and good in theory, but we have no way of confirming the effects due to our limited knowledge of the PackStruct implementation.

korydraughn: Thinking about it more, I suspect that iRODS 6 may end up adding a slightly modified version of the current API to avoid breaks with iRODS 5.

server/api/src/rsBulkDataObjPut.cpp
korydraughn: Agreed.

I just noticed that the comparison of checksums is performed twice. Once at this scope and another time in the call to `verifyChksumLocFile`.

lib/core/include/irods/rodsDef.h
SwooshyCueb: Not sure we want to do that since we're planning on changing `CHKSUM_LEN`. Though at that point we would probably get rid of `LONG_CHKSUM_LEN`

unit_tests/src/test_hashers.cpp
SwooshyCueb: I do, yes.

lib/hasher/src/checksum.cpp
SwooshyCueb: IIRC I asked a while back if we should make the logger work regardless of whether it was running in a local client or in the server, and we decided that outside of the networking plugins, there wasn't really a use case. It is my understanding that in client code, any use of the logger is currently effectively a non-op.

It looks like this function is called in both `putUtil` and `rsyncUtil` so maybe we can test via iput or irsync. I'll look into it. If we do want to switch to a different logging function, which one should I use? I remember one (or more?) is discouraged for use in new code.

lib/core/include/irods/rodsPackInstruct.h
SwooshyCueb: Based on everything I know about packstruct, this code change will not result in any behavioral changes and therefore will not impact compatibility. However, I accept that my understanding of packstruct is quite limited. I've only dug into it enough to be reasonably confident that this change won't break anything. But since we can't verify that, I'm fine with not including this change in this PR. (It looks like I missed one, anyway.) I'll make a note of this in #8721.

server/api/src/rsBulkDataObjPut.cpp
SwooshyCueb: Added another commit removing the redundant check

unit_tests/src/test_hashers.cpp
korydraughn: Nice.

lib/core/include/irods/rodsDef.h
alanking: Okay, was just wondering. Let's just leave this as-is.

lib/hasher/src/checksum.cpp
korydraughn: > It looks like this function is called in both putUtil and rsyncUtil so maybe we can test via iput or irsync.

Yes, please do.

> If we do want to switch to a different logging function, which one should I use? I remember one (or more?) is discouraged for use in new code.

True. `rodsLog` is discouraged, so we're either looking at direct use of `cout`/`cerr` or `fmt::print(stderr, ...)`.

The simplest thing I can think of is to wrap the calls in question with `::ProcessType` conditions. For example:

```cpp
// This is for code which can be invoked on either side of the connection.
if (CLIENT_PT != ::ProcessType) {
    // call the logger
}
```

There's also the option of extending the logger's capabilities to support client-side environments. I believe the client-side implementation is essentially a bunch of empty placeholder implementations. That might have to happen in a later release though.

lib/api/include/irods/fileChksum.h
```diff
@@ -16,13 +16,13 @@ typedef struct FileChksumInp {
...
-    char orig_chksum[NAME_LEN]; // Original incoming checksum.
+    char orig_chksum[CHKSUM_LEN]; // Original incoming checksum.
     rodsLong_t dataSize;
     struct KeyValPair condInput;
 } fileChksumInp_t;
 
-#define fileChksumInp_PI "struct RHostAddr_PI; str fileName[MAX_NAME_LEN]; str rescHier[MAX_NAME_LEN]; str objPath[MAX_NAME_LEN]; int flags; str in_pdmo[MAX_NAME_LEN]; str orig_chksum[NAME_LEN]; double dataSize; struct KeyValPair_PI;"
-#define fileChksumOut_PI "str chksumStr[NAME_LEN];"
+#define fileChksumInp_PI "struct RHostAddr_PI; str fileName[MAX_NAME_LEN]; str rescHier[MAX_NAME_LEN]; str objPath[MAX_NAME_LEN]; int flags; str in_pdmo[MAX_NAME_LEN]; str orig_chksum[CHKSUM_LEN]; double dataSize; struct KeyValPair_PI;"
+#define fileChksumOut_PI "str chksumStr[CHKSUM_LEN];"
```
korydraughn: Revert packing instruction changes.

lib/hasher/src/SHA1Strategy.cpp
```diff
@@ -90,7 +90,11 @@ namespace irods {
         unsigned long out_len = CHKSUM_LEN - len;
 
         unsigned char out_buffer[CHKSUM_LEN];
-        base64_encode( final_buffer, SHA_DIGEST_LENGTH, out_buffer, &out_len );
+        int rc = base64_encode(final_buffer, SHA_DIGEST_LENGTH, out_buffer, &out_len);
+        if (0 != rc) {
+            const auto msg = fmt::format("{}: Failed to base64 encode hash.", __func__);
+            return ERROR(rc, msg);
```
korydraughn:
```suggestion
            auto msg = fmt::format("{}: Failed to base64 encode hash.", __func__);
            return ERROR(rc, std::move(msg));
```

or ...


```suggestion
            return ERROR(rc, fmt::format("{}: Failed to base64 encode hash.", __func__));
```

_This applies to the other hashers too._

SwooshyCueb: Done. Since it was trivial, I went ahead and did the same for the other error messages in the hashers.

lib/api/include/irods/fileChksum.h
```diff
@@ -16,7 +16,7 @@ typedef struct FileChksumInp {
     char objPath[MAX_NAME_LEN];
     int flag;                   // Not used for now.
     char in_pdmo[MAX_NAME_LEN]; // Flag indicating if we are being executed from a pdmo.
-    char orig_chksum[NAME_LEN]; // Original incoming checksum.
+    char orig_chksum[CHKSUM_LEN]; // Original incoming checksum.
```
alanking, line 19: It's a little weird to see `CHKSUM_LEN` here and `NAME_LEN` below in the packing instruction, so it might be good to acknowledge that difference in a comment here, and why it is okay - that is, because `CHKSUM_LEN` and `NAME_LEN` are the same size and will not change.

korydraughn: Agreed.

lib/hasher/src/checksum.cpp
SwooshyCueb: Opened #8743

SwooshyCueb:
Ran tests again, everything still passing

SwooshyCueb:
squashed

korydraughn:
@JustinKyleJames' changes are about to land. They already have the pounds, so let's wait for his to land and then you can add the test cases for the new hasher.

SwooshyCueb:
roger dodger

korydraughn:
@JustinKyleJames' PR has landed.

You're all set to go fast.

SwooshyCueb:
Added crc64nvme to the new unit test, plus a few clang-format suggestions. Should be ready for squash+#.

SwooshyCueb:
\#'d

Issue #8591, opened 2025-07-22
Move icommands back into main repo


SwooshyCueb:
When the decision was made to split the icommands into their own repository, it made sense. However, the reasons for doing so are no longer relevant, and the split repository is an obstacle for Normal and Boring packaging. Let's move the icommands back into the main repo.

SwooshyCueb:
A reddit thread on merging one git repository into a subdirectory of another while preserving commit history: https://www.reddit.com/r/git/comments/sheoo3/merge_git_repo_with_another_repos_subdirectory/

trel:
Original removal of the iCommands was in 831809e95a3 for #2239 in the run-up to 4.2.0 and the use of CMake.

--

A quick read suggests we're looking at a merge commit (the horror!) with the following...

https://git-scm.com/docs/git-merge#Documentation/git-merge.txt---allow-unrelated-histories

Another couple posts suggests the same process...

 - https://jdriven.com/blog/2021/04/how-to-merge-multiple-git-repositories
 - https://gfscott.com/blog/merge-git-repos-and-keep-commit-history/

trel:
where do you propose to attach into the tree?

in a directory called clients?   irods/irods/clients/icommands

just a single client?  irods/irods/client/icommands

somewhere else?

SwooshyCueb:
`clients/icommands` makes the most sense to me

korydraughn:
@SwooshyCueb Please close if complete.

PR #8719, opened 2025-11-04
[#8591] Re-integrate icommands into main repo (separate histories)

SwooshyCueb:
Addresses #8591
Alternate approach: #8725

Commit history of icommands is preserved via a merge commit. This is why there are so many commits in this PR. Additionally, this PR cannot be rebased using standard tooling, so in order to merge, main will likely need to be manually reset to the tip of the PR. I will occasionally manually rebase the PR to keep it up to date with main.

The parents of the merge commit are tip of main for this repo and a child of tip of main for irods_client_icommands. They have unrelated histories, so the new commit graph will have two origin points.


SwooshyCueb:
Rebased. PR is in draft mode while testing, but I don't expect any surprises, so I'd say this is ready for review.
As for which approach I'd take, I'd go with #8725, but I don't feel that strongly about it.

Additionally, I'd like to call attention to this bit of the PR description, as I don't think we've discussed it:
> this PR cannot be rebased using standard tooling, so in order to merge, main will likely need to be manually reset to the tip of the PR.

trel:
agreed, we're going to work with the preferred #8725 first.

SwooshyCueb:
#8725 merged, closing.

PR #8724, opened 2025-11-04
[???] Allow tests to be passed in as regular args to script (main)

MartinFlores751:
This PR attempts to allow for the passing of multiple tests as plain arguments. An example follows:
```console
$ python3 scripts/run_tests.py test_plugin_policy_engine-data_retention test_plugin_policy_engine-verify_checksum test_plugin_event_handler-collection_modified
```

For a clearer example the following would run tests `A`, `B`, and `C`:
```console
$ python3 scripts/run_tests.py A B C
```

---

This update to the script is useful in the case where you wish to run multiple tests directly using `run_tests.py`, but don't wish to set up a bash script to run `for i in "${test_list[@]}"; do python3 scripts/run_tests.py --run_specific-test "${i}"; done`.

I'l briefly mention, the previous bash script is great for setting `ASAN_OPTIONS` to produce better reports (e.g., set `log_path=/tmp/asan-${i}-part-four` to denote the test the log is from and what run you are in).

### DIFF
--- a/scripts/run_tests.py
+++ b/scripts/run_tests.py
@@ -131,7 +131,7 @@ if __name__ == '__main__':
     irods.log.register_file_handler(IrodsConfig().test_log_path)
 
     parser = argparse.ArgumentParser()
-    parser.add_argument('--run_specific_test', metavar='dotted name')
+    parser.add_argument('--run_specific_test', metavar='dotted name', help='deprecated')
     parser.add_argument('--skip_until', action="store")
     parser.add_argument('--run_python_suite', action='store_true')
     parser.add_argument('--run_plugin_tests', action='store_true')


korydraughn:
I thought `--run_specific_test` supported multiple tests. Perhaps I'm getting it mixed with something else.

Does this change affect any of the script's previous functionality?
Does this change affect the testing environment's assumptions about run_tests.py?

MartinFlores751:
> I thought `--run_specific_test` supported multiple tests. Perhaps I'm getting it mixed with something else.

I think you are right? You probably need to do something like `--run_specific_test "a b c d"`. Not sure if you can have multiple `--run_specific_test`s though...

> Does this change affect any of the script's previous functionality?

Previously, any non `--` arguments were tossed to the Python `_` variable. If any script was relying on this behavior, I think it would be to have some "comments" on the script or passed in accidentally.

> Does this change affect the testing environment's assumptions about run_tests.py?

I'm not sure what assumptions the testing environment has, though if you have some pointers on where to look, it would be much appreciated!

korydraughn:
> > I thought `--run_specific_test` supported multiple tests. Perhaps I'm getting it mixed with something else.
> 
> I think you are right? You probably need to do something like `--run_specific_test "a b c d"`. Not sure if you can have multiple `--run_specific_test`s though...

Please verify.

> > Does this change affect the testing environment's assumptions about run_tests.py?
> 
> I'm not sure what assumptions the testing environment has, though if you have some pointers on where to look, it would be much appreciated!

We know the testing environment accepts zero or more tests as an input. I can't point you to one location to figure that out. You're going to have to investigate by running various tests with this change.

MartinFlores751:
> We know the testing environment accepts zero or more tests as an input. I can't point you to one location to figure that out. You're going to have to investigate by running various tests with this change.

Gotcha. I'll just run core, unit, federation, and a random plugin test to check it out then.

alanking:
> We know the testing environment accepts zero or more tests as an input. I can't point you to one location to figure that out. You're going to have to investigate by running various tests with this change.

The test list in the testing environment is maintained by the test "manager", which puts it on a queue from which test "runners" (or executors, if you prefer) pull, *one at a time*, and execute tests. I think that this would not affect the testing environment in its current state.

alanking:
That being said, `--run_specific_tests` only considers the first test provided to it (a single argument). This is what happened when I tried to provide more than one:
```bash
$ python3 scripts/run_tests.py --run_s test_setting_user_password.test_modifying_user_password.test_config_legacy_sets_password_for_native_scheme_but_not_irods_scheme test_setting_user_password.test_modifying_user_password.test_config_sets_password_for_both_native_and_irods_schemes
irods.test.test_setting_user_password.test_modifying_user_password.test_config_legacy_sets_password_for_native_scheme_but_not_irods_scheme ... test_config_legacy_sets_password_for_native_scheme_but_not_irods_scheme (irods.test.test_setting_user_password.test_modifying_user_password.test_config_legacy_sets_password_for_native_scheme_but_not_irods_scheme) ... ok

----------------------------------------------------------------------
Ran 1 test in 15.248s

OK
<__main__.RegisteredTestResult run=1 errors=0 failures=0>
```
I'm guessing that it treats the other one as a positional argument. I don't know what this script does with positional arguments without looking.

And if you provide more than one `--run_specific_test` argument, it looks like `argparse` just takes the last one and ignores any that precede it:
```bash
$ python3 scripts/run_tests.py --run_s test_setting_user_password.test_modifying_user_password.test_config_legacy_sets_password_for_native_scheme_but_not_irods_scheme --run_s test_setting_user_password.test_modifying_user_password.test_config_sets_password_for_both_native_and_irods_schemes
irods.test.test_setting_user_password.test_modifying_user_password.test_config_sets_password_for_both_native_and_irods_schemes ... test_config_sets_password_for_both_native_and_irods_schemes (irods.test.test_setting_user_password.test_modifying_user_password.test_config_sets_password_for_both_native_and_irods_schemes) ... ok

----------------------------------------------------------------------
Ran 1 test in 32.089s

OK
<__main__.RegisteredTestResult run=1 errors=0 failures=0>
```

MartinFlores751:
Most types of tests pass (core, unit, federation, plugin), save for some topology tests in consumer and provider mode. See #8756 and #8773 for more details.

scripts/run_tests.py
```diff
@@ -195,7 +195,7 @@ def clear_irods_shared_memory_files():
             irods.test.settings.HOSTNAME_2 = 'resource2.example.org'
             irods.test.settings.HOSTNAME_3 = 'resource3.example.org'
 
-    test_identifiers = []
+    test_identifiers = [*test_list] if test_list else []
     if options.run_specific_test:
         test_identifiers.append(options.run_specific_test)
```
alanking: So we are allowing for both positional arguments *and* `--run_specific_test`? Should we deprecate the `--run_specific_test` option? Having both seems confusing.

MartinFlores751: We can deprecate and/or remove `--run_specific_test`. I left the `--run_specific_test` option as is because I didn't check if it's a load-bearing `--thing` being used by other scripts.

alanking: The option is still being used in the testing environment, so we would need to update it there if it were to be removed. If it's just deprecated, we can change it later.

korydraughn: What's the next step for this? Have decisions been made?

MartinFlores751: I'll add `deprecated` to the help text of `--run_specific_test`, and if there's a way to mark the flag deprecated in Python too, we can do that as well.

korydraughn: Sounds good.

korydraughn:
GitHub is reporting a conflict too.

scripts/run_tests.py
MartinFlores751: Done.

scripts/run_tests.py
```diff
@@ -131,7 +131,7 @@ def clear_irods_shared_memory_files():
     irods.log.register_file_handler(IrodsConfig().test_log_path)
 
     parser = argparse.ArgumentParser()
-    parser.add_argument('--run_specific_test', metavar='dotted name')
+    parser.add_argument('--run_specific_test', metavar='dotted name', help='deprecated')
```
korydraughn, line 134: What does the help text look like with this change?

Issue #8667, opened 2025-09-10
coordinating resource quotas should have all replicas 'under' it be counted


trel:
> > 2. If you set a quota on a coordinating resource, the "iadmin cu" does not update the usage on that coordinating resource.
> 
> Just one comment to add following discussion about this.
> 
> Q. Should physical quotas be allowed on coordinating resources?
> A. We feel it should. Whether that's a good use of the feature is up to the admin.
> 
> ---
> 
> The question is motivated by a resource hierarchy containing a replication resource and three unixfilesystem resources. If a quota is attached to the replication resource and a user stores a file there, the replication resource will replicate it to its children.
> 
> What should the quota report for the replication resource?
> Should the user's file cost 3x its size due to there being three physical replicas?
> 
> This situation led us to the answer of ... the quota system is about physical storage and should count all replicas. The quota system is a tool and admins are given freedom over where it is applied. 

 _Originally posted by @korydraughn in [#4089](https://github.com/irods/irods/issues/4089#issuecomment-3272404157)_

FifthPotato:
After some discussion with the group, we've discovered the "enforcement" part of this issue is a little more hairy than anticipated. As it stands right now, quotas will only be enforced on resource roots (that is, roots of resource hierarchies or individual resources). Proper enforcement will likely involve adding quota enforcement at each level, which means each resource type will need to account for quotas when a resource hierarchy is being resolved. This will likely manifest as a common function used for all of the resource plugins. We'll be implementing this later.

FifthPotato:
I've written the first half of what is required for this issue in this PR: https://github.com/irods/irods/pull/8745 

This PR adds the math for coordinating resources-- each coordinating resource will now have a usage record generated based on the sum of the usage of all its children. It also adds a basic test case for coordinating resources. Effectively, this is half of the work required to get coordinating resources fully working.

This PR does *not* perform the enforcement of the non-root resources, which is the other half. When we decide to do that, the enforcement code will need to call `setRescQuota` which will update R_QUOTA_MAIN based on these usage records.

PR #8745, opened 2025-11-14
[8667] Modify quota calculations to account for coordinating resources (main)

FifthPotato:
Finally done! Taking suggestions as always.
`test_quotas` passes for every ubuntu24 database we have a testing environment for (mariadb 10.11, 11.4, 11.8; mysql 8, 8.4; postgres 14, 16, 17)

### DIFF
--- a/plugins/database/src/db_plugin.cpp
+++ b/plugins/database/src/db_plugin.cpp
@@ -1469,13 +1469,43 @@ int setOverQuota( rsComm_t *rsComm ) {
     int statementNum = UNINITIALIZED_STATEMENT_NUMBER;
     char myTime[50];
 
+// NOLINTBEGIN(cppcoreguidelines-macro-usage)
+#ifdef ORA_ICAT
+#  define IRODS_INT_TO_STRING(x) " TO_CHAR(" x ") "
+#elif MY_ICAT
+#  define IRODS_INT_TO_STRING(x) " CAST(" x " AS CHAR) "
+#else
+#  define IRODS_INT_TO_STRING(x) " CAST(" x " AS TEXT) "
+#endif
+    // NOLINTEND(cppcoreguidelines-macro-usage)
+
     /* For each defined group limit (if any), get a total usage on that
      * resource for all users in that group: */
     char mySQL1[] = "select sum(quota_usage), UM1.user_id, R_QUOTA_USAGE.resc_id from R_QUOTA_USAGE, R_QUOTA_MAIN, R_USER_MAIN UM1, R_USER_GROUP, R_USER_MAIN UM2 where R_QUOTA_MAIN.user_id = UM1.user_id and UM1.user_type_name = 'rodsgroup' and R_USER_GROUP.group_user_id = UM1.user_id and UM2.user_id = R_USER_GROUP.user_id and R_QUOTA_USAGE.user_id = UM2.user_id and R_QUOTA_MAIN.resc_id = R_QUOTA_USAGE.resc_id group by UM1.user_id, R_QUOTA_USAGE.resc_id";
 
     /* For each defined group limit on total usage (if any), get a
-     * total usage on any resource for all users in that group: */
-    char mySQL2a[] = "select sum(quota_usage), R_QUOTA_MAIN.quota_limit, UM1.user_id from R_QUOTA_USAGE, R_QUOTA_MAIN, R_USER_MAIN UM1, R_USER_GROUP, R_USER_MAIN UM2 where R_QUOTA_MAIN.user_id = UM1.user_id and UM1.user_type_name = 'rodsgroup' and R_USER_GROUP.group_user_id = UM1.user_id and UM2.user_id = R_USER_GROUP.user_id and R_QUOTA_USAGE.user_id = UM2.user_id and R_QUOTA_USAGE.resc_id != %s and R_QUOTA_MAIN.resc_id = %s group by UM1.user_id,  R_QUOTA_MAIN.quota_limit";
+     * total usage on any resource for all users in that group.
+     * The query will sum up total usage for all groups across all resources.
+     * UM1.user_id is a group (filtered by WHERE clause) so the query generates
+     * one row for total usage and links the usage to a quota in R_QUOTA_MAIN
+     * Essentially, this does two things: it gets a total usage and a
+     * defined quota limit. It will use this row later to compute and set
+     * quota_over. Additionally, to avoid multiple counting, this query
+     * will only count totals of leaf (non-coordinating) resources by excluding
+     * any resource that is parent to another (i.e. there exists a resource that
+     * has this resource as a parent)
+     *
+     * This query can be optimized by adding an index on R_RESC_MAIN.resc_parent
+     * to speed up the subquery SELECT. */
+
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays,modernize-avoid-c-arrays)
+    char mySQL2a[] =
+        "select sum(quota_usage), R_QUOTA_MAIN.quota_limit, UM1.user_id from R_QUOTA_USAGE, R_QUOTA_MAIN, R_USER_MAIN "
+        "UM1, R_USER_GROUP, R_USER_MAIN UM2 where R_QUOTA_MAIN.user_id = UM1.user_id and UM1.user_type_name = "
+        "'rodsgroup' and R_USER_GROUP.group_user_id = UM1.user_id and UM2.user_id = R_USER_GROUP.user_id and "
+        "R_QUOTA_USAGE.user_id = UM2.user_id and R_QUOTA_USAGE.resc_id != %s and R_QUOTA_MAIN.resc_id = %s AND NOT "
+        "EXISTS (SELECT resc_id FROM R_RESC_MAIN WHERE resc_parent = " IRODS_INT_TO_STRING(
+            "R_QUOTA_USAGE.resc_id") ") group by UM1.user_id, R_QUOTA_MAIN.quota_limit";
     char mySQL2b[MAX_SQL_SIZE];
 
     char mySQL3a[] = "update R_QUOTA_MAIN set quota_over= %s - ?, modify_ts=? where user_id=? and %s - ? > quota_over and resc_id = %s";
@@ -1527,8 +1557,20 @@ int setOverQuota( rsComm_t *rsComm ) {
     for ( rowsFound = 0;; rowsFound++ ) {
         int status2;
         if ( rowsFound == 0 ) {
-            status = cmlGetFirstRowFromSql( "select sum(quota_usage), R_QUOTA_MAIN.user_id from R_QUOTA_USAGE, R_QUOTA_MAIN where R_QUOTA_MAIN.user_id = R_QUOTA_USAGE.user_id and R_QUOTA_MAIN.resc_id = '0' group by R_QUOTA_MAIN.user_id",
-                                            &statementNum, 0, &icss );
+            /* This query fetches the total usage of a user and matches it
+               to a defined total quota. It's similar to mySQL2a above, but
+               it only works per-user. Since user quotas are officially
+               deprecated, it sees nearly no use, but it also has the updated
+               anti-double counting logic added to it for completeness. */
+            status = cmlGetFirstRowFromSql(
+                "select sum(quota_usage), R_QUOTA_MAIN.user_id from R_QUOTA_USAGE, R_QUOTA_MAIN where "
+                "R_QUOTA_MAIN.user_id = R_QUOTA_USAGE.user_id and R_QUOTA_MAIN.resc_id = '0' AND NOT EXISTS (SELECT "
+                "resc_id FROM R_RESC_MAIN WHERE resc_parent = " IRODS_INT_TO_STRING(
+                    "R_QUOTA_USAGE.resc_id") ") group by R_QUOTA_MAIN.user_id",
+                &statementNum,
+                0,
+                &icss);
+#undef IRODS_INT_TO_STRING
         }
         else {
             status = cmlGetNextRowFromStatement( statementNum, &icss );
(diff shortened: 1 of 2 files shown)


trel:
reading the SQL carefully... but in the meantime...

can you add some more words to the commit message that explains in english what these SQL additions do / limit / find and how?  a paragraph for each of the two main sections you touched would be swell.

plugins/database/src/db_plugin.cpp
```diff
@@ -1383,13 +1383,21 @@ int setOverQuota( rsComm_t *rsComm ) {
...
      * total usage on any resource for all users in that group: */
-    char mySQL2a[] = "select sum(quota_usage), R_QUOTA_MAIN.quota_limit, UM1.user_id from R_QUOTA_USAGE, R_QUOTA_MAIN, R_USER_MAIN UM1, R_USER_GROUP, R_USER_MAIN UM2 where R_QUOTA_MAIN.user_id = UM1.user_id and UM1.user_type_name = 'rodsgroup' and R_USER_GROUP.group_user_id = UM1.user_id and UM2.user_id = R_USER_GROUP.user_id and R_QUOTA_USAGE.user_id = UM2.user_id and R_QUOTA_USAGE.resc_id != %s and R_QUOTA_MAIN.resc_id = %s group by UM1.user_id,  R_QUOTA_MAIN.quota_limit";
+    char mySQL2a[] = "select sum(quota_usage), R_QUOTA_MAIN.quota_limit, UM1.user_id from R_QUOTA_USAGE, R_QUOTA_MAIN, R_USER_MAIN UM1, R_USER_GROUP, R_USER_MAIN UM2 where R_QUOTA_MAIN.user_id = UM1.user_id and UM1.user_type_name = 'rodsgroup' and R_USER_GROUP.group_user_id = UM1.user_id and UM2.user_id = R_USER_GROUP.user_id and R_QUOTA_USAGE.user_id = UM2.user_id and R_QUOTA_USAGE.resc_id != %s and R_QUOTA_MAIN.resc_id = %s AND NOT EXISTS (SELECT resc_id FROM R_RESC_MAIN WHERE resc_parent = " IRODS_INT_TO_STRING("R_QUOTA_USAGE.resc_id") ") group by UM1.user_id, R_QUOTA_MAIN.quota_limit";
```
trel: so this is ignoring coordinating resources (by matching on resc_parent) so that usage isn't double counted...

does this behave with deep/nested trees of three levels or more too?

scripts/irods/test/test_quotas.py
```diff
@@ -228,3 +228,68 @@ def acRescQuotaPolicy(rule_args, callback, rei):
...
+                group_name = 'test_group_case_3_4_5'
+                file_name = 'test_file_case_3_4_5'
+                resc_name = 'ufs0_quota_case_3_4_5'
+                other_resc_name = 'ufs1_quota_case_3_4_5'
+                repl_resc_name = 'repl_quota_case_3_4_5'
+                lib.make_file(os.path.join(self.quota_user.local_session_dir, file_name), 4096, contents='arbitrary')
+                lib.create_ufs_resource(self.admin, resc_name)
+                lib.create_ufs_resource(self.admin, other_resc_name)
+                lib.create_replication_resource(self.admin, repl_resc_name)
+
```
trel, line 312: please add a passthru resource 'above' the replication resource... this will confirm parent-parent-resc relationships are still honored and 'rolled up' correctly.

plugins/database/src/db_plugin.cpp
```diff
@@ -11452,9 +11461,40 @@ irods::error db_calc_usage_and_quota_op(
...
+// Case "as bigint" for postgres
+#    define IRODS_QUOTA_CAST_TYPE_AS " AS BIGINT "
+#endif
+
     status =  cmlExecuteNoAnswerSql(
-                  "insert into R_QUOTA_USAGE (quota_usage, resc_id, user_id, modify_ts) (select sum(R_DATA_MAIN.data_size), R_RESC_MAIN.resc_id, R_USER_MAIN.user_id, ? from R_DATA_MAIN, R_USER_MAIN, R_RESC_MAIN where R_USER_MAIN.user_name = R_DATA_MAIN.data_owner_name and R_USER_MAIN.zone_name = R_DATA_MAIN.data_owner_zone and R_RESC_MAIN.resc_id = R_DATA_MAIN.resc_id group by R_RESC_MAIN.resc_id, user_id)",
+                  "INSERT INTO R_QUOTA_USAGE (quota_usage, resc_id, user_id, modify_ts) "
+                  IRODS_QUOTA_WITH_RECURSIVE
+                  // Begin CTE: recursive table
+                  "resc_usage(quota_usage, resc_id, resc_parent, user_id) AS ("
```
korydraughn: I've never seen this before. Based on the code, this expands to:
```sql
WITH resc_usage(quota_usage, resc_id, resc_parent, user_id) AS (
```

What does `resc_usage(quota_usage, resc_id, resc_parent, user_id)` mean in SQL?

korydraughn (commented):
Good stuff.

Awaiting more words to aid in understanding the SQL, etc.

trel:
these queries are *still* using `data_owner_name`... where they really should be counting quota for anyone who has 'own' permission.  the `data_owner_name` is only the initial uploader/writer.    that can be addressed in a separate PR perhaps... as I'm sure it's not just in the 2-3 queries touched here.

plugins/database/src/db_plugin.cpp
FifthPotato: I don't see why it wouldn't work, but I added the passthru case to deepen the nesting a bit. All assertions still seem to be working.

plugins/database/src/db_plugin.cpp
FifthPotato: A CTE can be essentially thought of as a temporary table that applies to your final query (after the WITH). What this does is assign names to the columns of the temporary table. So, this temporary table is `resc_usage` and its columns are `(quota_usage, resc_id, resc_parent, user_id)`

plugins/database/src/db_plugin.cpp
```diff
@@ -11452,9 +11461,40 @@ irods::error db_calc_usage_and_quota_op(
...
+#    define IRODS_QUOTA_CAST_TYPE_AS " AS SIGNED "
+#else
+#    define IRODS_QUOTA_WITH_RECURSIVE " WITH RECURSIVE "
+// Case "as bigint" for postgres
+#    define IRODS_QUOTA_CAST_TYPE_AS " AS BIGINT "
+#endif
+
     status =  cmlExecuteNoAnswerSql(
-                  "insert into R_QUOTA_USAGE (quota_usage, resc_id, user_id, modify_ts) (select sum(R_DATA_MAIN.data_size), R_RESC_MAIN.resc_id, R_USER_MAIN.user_id, ? from R_DATA_MAIN, R_USER_MAIN, R_RESC_MAIN where R_USER_MAIN.user_name = R_DATA_MAIN.data_owner_name and R_USER_MAIN.zone_name = R_DATA_MAIN.data_owner_zone and R_RESC_MAIN.resc_id = R_DATA_MAIN.resc_id group by R_RESC_MAIN.resc_id, user_id)",
+                  "INSERT INTO R_QUOTA_USAGE (quota_usage, resc_id, user_id, modify_ts) "
```
FifthPotato: This SQL query is the main meat of this PR (and to a degree, of the quota system in general).
In its original iteration, it would sum up the data object sizes and group them by resource and user, effectively getting a total usage per resource, per user, and inserting it into `R_QUOTA_USAGE`.
My changed version is a modified version of it: it is recursive, but its base case is the same as the old query, so it generates a total usage per resource, per user (of all storage resources). One difference from the old query is that it additionally links the parent of the associated resource so it can follow the hierarchy upwards.
Then, it recurses leaf-to-root, and in each recursion, the `resc_id` gets replaced by `resc_parent` so it can walk the hierarchy. Essentially, each leaf resource will generate a row for itself and each of its ancestors with the leaf resource's usage.
Finally, at the end, it will sum up all the generated usage/resource rows one more time (you can't aggregate in the recursive clause of a CTE, so the aggregate has to be done at the end) to get a total usage for each resource in the hierarchy.

plugins/database/src/db_plugin.cpp
```diff
@@ -1441,8 +1449,9 @@ int setOverQuota( rsComm_t *rsComm ) {
     for ( rowsFound = 0;; rowsFound++ ) {
         int status2;
         if ( rowsFound == 0 ) {
-            status = cmlGetFirstRowFromSql( "select sum(quota_usage), R_QUOTA_MAIN.user_id from R_QUOTA_USAGE, R_QUOTA_MAIN where R_QUOTA_MAIN.user_id = R_QUOTA_USAGE.user_id and R_QUOTA_MAIN.resc_id = '0' group by R_QUOTA_MAIN.user_id",
+            status = cmlGetFirstRowFromSql( "select sum(quota_usage), R_QUOTA_MAIN.user_id from R_QUOTA_USAGE, R_QUOTA_MAIN where R_QUOTA_MAIN.user_id = R_QUOTA_USAGE.user_id and R_QUOTA_MAIN.resc_id = '0' AND NOT EXISTS (SELECT resc_id FROM R_RESC_MAIN WHERE resc_parent = " IRODS_INT_TO_STRING("R_QUOTA_USAGE.resc_id") ") group by R_QUOTA_MAIN.user_id",
```
FifthPotato: I'm fairly sure this query is obsolete now because this only calculates totals per-user (but user quotas are no longer a thing). For completeness/correctness, however, I added the double-counting exclusion logic in here as well.

plugins/database/src/db_plugin.cpp
FifthPotato: This query sums up total usage, per group, across all resources. `UM1.user_id` is a group, as shown by the WHERE clause `UM1.user_type_name = 'rodsgroup'`, so it generates a row for total usage and matches it to the relevant defined quota in `R_QUOTA_MAIN`. This essentially does two things in one query: it totals the usage, and it finds the matching quota. Later in this same function, it uses the data in the row to compute "quota_over" and applies it in an `UPDATE`. All that I change here is that I exclude all coordinating resources as part of the count, or else they would double-count against a total quota; I do this by ensuring each counted row in `R_QUOTA_USAGE` is parent to no other resource (i.e. it has no children, so it is a leaf resource).

(Random other thoughts involving this: this query is potentially a bit inefficient since it has to do the SELECT subquery each time on `R_RESC_MAIN`. You could speed it up by  adding an index on R_RESC_MAIN.resc_parent or by including another column in R_QUOTA_USAGE that indicates whether or not the relevant usage row is leaf/nonleaf, but I think changing the schema is...not great.)

plugins/database/src/db_plugin.cpp
FifthPotato: Also, the crazy casting/`CASE` logic is because it looks like `resc_parent` is a VARCHAR, but it's also nullable for some reason, so the `COALESCE` catches nulls, and `''` catches empty string (both of which are valid values for "no parent").

FifthPotato:
Also, re-ran `test_quotas` on all the cases listed in the OP, and they still pass.

trel:
these explanations are great - please add them to the commit messages themselves.

we don't want someone with just the code (and the git history) to have to come back to github and this conversation to find out how these work or what your thinking was.

FifthPotato:
> these explanations are great - please add them to the commit messages themselves.
> 
> we don't want someone with just the code (and the git history) to have to come back to github and this conversation to find out how these work or what your thinking was.

Makes sense. Would we prefer it in the commit message or should I rephrase them as informational and put them as comments on the relevant lines?

trel:
informational is the right tone.  goal is to express intent and reasoning of the developer.

explaining these relatively complex queries is very important.
i think having them in the code as comments is more helpful than in the commit message.

plugins/database/src/db_plugin.cpp
trel: agreed on not touching the schema today.   if we determine (now or later) that there are inefficiencies here, we can create another issue.

plugins/database/src/db_plugin.cpp
trel: adding the double-counting exclusion here is correct.

we still need to calculate per-user quotas at this time.   while we no longer allow per-user quotas to be set... they could still be in a system that was upgraded and not yet zeroed out by the admin.

so please create a 6.0 issue to remove the per-user calculations and point to where they are in the code.  that issue could also require a warning to be logged somewhere if a per-user quota is found in the database.

plugins/database/src/db_plugin.cpp
trel: excellent - yes, please get these into code comments in the most helpful place for the next reader.

plugins/database/src/db_plugin.cpp
trel: excellent - resolving.

plugins/database/src/db_plugin.cpp
```diff
@@ -11477,6 +11495,41 @@ irods::error db_calc_usage_and_quota_op(
 #    define IRODS_QUOTA_CAST_TYPE_AS " AS BIGINT "
 #endif
 
+    /* This query sums up data object sizes and groups them by resource and user.
+       This gets a total usage per resource for each user. This is inserted into
+       R_QUOTA_USAGE. Additionally, after adding the total storage resource
+       usage for each user in the base case, it will recursively add
+       more rows with the same usage for each leaf resource's parent,
```
trel: since it walks up to the parents of coordinating resources as well...

```suggestion
       more rows with the same usage for each resource's parent,
```

plugins/database/src/db_plugin.cpp
trel: done - thanks.  resolving.

plugins/database/src/db_plugin.cpp
```diff
@@ -11452,9 +11479,75 @@ irods::error db_calc_usage_and_quota_op(
...
+       The COALESCE/casting is required because resc_parent is a nullable varchar,
+       so it is needed to catch both cases.
+       Suppose you have a coordinating resource c0 with two children: s0, s1
+
+       Suppose there are two data objects owned by user_id 10001 of size 4096
+       on ids s0, s1.
+       The base case query will generate:
+       quota_usage, resc_id, resc_parent, user_id
+       4096, s0, c0, 10001
+       4096, s1, c0, 10001
```
alanking: Not that it matters much, but I took `c0`, `s0`, and `s1` to be resource *names* whereas `resc_id` and `resc_parent` are going to be resource IDs, which would look like the value shown for `user_id`. That could involve more words to set up values for all of these things, but, just thought I'd point it out. Maybe it could be denoted with, e.g. `c0_id` or something.

FifthPotato: True, I thought about that, but I was thinking that referencing them symbolically was OK. I'll make the modification for clarity, though.

FifthPotato:
Core and unit tests pass.

plugins/database/src/db_plugin.cpp
```diff
@@ -11452,9 +11479,75 @@ irods::error db_calc_usage_and_quota_op(
...
+#    define IRODS_QUOTA_CAST_TYPE_AS " AS BIGINT "
+#endif
+
+    /* This query sums up data object sizes and groups them by resource and user.
+       This gets a total usage per resource for each user. This is inserted into
+       R_QUOTA_USAGE. Additionally, after adding the total storage resource
+       usage for each user in the base case, it will recursively add
+       more rows with the same usage for each resource's parent,
+       effectively generating a usage row for each coordinating resource equal to
+       the sum of its child hierarchy. This repeats leaf-to-root.
```
alanking, line 11800: What happens if we have a hierarchy like this:
```
pt0:passthru
└── repl0:replication               # quota set 
    ├── pt1:passthru
    │   └── repl1:replication       # quota set 
    │       ├── ufs0:unixfilesystem
    │       ├── ufs1:unixfilesystem
    │       └── ufs2:unixfilesystem
    ├── pt2:passthru                # quota set 
    │   └── repl2:replication
    │       ├── ufs3:unixfilesystem # quota set 
    │       ├── ufs4:unixfilesystem # quota set 
    │       └── ufs5:unixfilesystem # quota set 
    └── pt3:passthru
        └── repl3:replication       # quota set 
            ├── ufs6:unixfilesystem # quota set 
            ├── ufs7:unixfilesystem # quota set 
            └── ufs8:unixfilesystem # quota set 
```
I guess the questions which follow are summarily about what happens when there are "holes" in the hierarchy which do not have quotas set. I'm sure we can create a more minimal example than this, and it might be best if we just talk it out offline.

Will `repl2` have a quota because all of its children and its parent resources have quotas, even if a quota is not explicitly set on `repl2` itself?

Will the quota usage on `repl0` reflect all of the quotas on all of its children even though only `pt2` has a quota explicitly set?

FifthPotato: > Will repl2 have a quota because all of its children and its parent resources have quotas, even if a quota is not explicitly set on repl2 itself?

Maybe not in the traditional sense; since you can only write to resource roots (`pt0` in this case) each quota should apply when each coordinating resource does its "extra" action (replicate, random, etc.) but I don't have a test case for this (yet?). Essentially, I don't see quota violation errors bubble up to the root action, but it's possible they're stopped in "extra" actions when they occur. In my big test case table, I marked this as "should succeed, but replication/extra actions may fail (if there are quotas set on the children)". Given the scope of https://github.com/irods/irods/issues/8667, this may be something I tackle in a separate issue or PR.
> Will the quota usage on repl0 reflect all of the quotas on all of its children even though only pt2 has a quota explicitly set?

Usage is calculated without regard to the existence of quotas; in this case, whether a not a quota is set does not influence whether or not it counts toward a usage.

alanking: > whether a not a quota is set does not influence whether or not it counts toward a usage.

Okay, this was the missing piece in my mental model. The usage is calculated for each resource regardless of whether quota limits are set on those resources.

After some discussion, I think I understand what's going on now. Let's make sure quota violations are enforced in the expected manner in each place of a hierarchy by adding tests for quota violation at the root (I think you already have this one), in the middle, and at a leaf.

FifthPotato:
After some more discussion, we've decided we can move forward with this, as the math is correct. However, there is no enforcement for now in the server so this isn't a complete solution to the listed issue.

plugins/database/src/db_plugin.cpp
FifthPotato: Going to punt this for now, see latest comment

FifthPotato:
Also, as of the latest force push, `test_quotas` passes. The two test cases that test for proper non-root enforcement will be added later when proper non-root enforcement is actually added.

plugins/database/src/db_plugin.cpp
FifthPotato: Done. See https://github.com/irods/irods/issues/8774

plugins/database/src/db_plugin.cpp
alanking: See https://github.com/irods/irods/pull/8745#issuecomment-3633868788

scripts/irods/test/test_quotas.py
```diff
@@ -278,3 +278,74 @@ def acRescQuotaPolicy(rule_args, callback, rei):
             self.admin.run_icommand(['iadmin', 'rmresc', other_resc_name])
             self.admin.run_icommand(['iadmin', 'rmgroup', f'{group_name}_1'])
             IrodsController().reload_configuration()
+
+    # The following test covers case 3, 4, 5
+    def test_physical_resource_with_coordinating_resources_issue_8667(self):
```
alanking:
```suggestion
    def test_physical_resource_with_coordinating_resources__issue_8667(self):
```

trel:
The non-root work will happen as part of #8667

alanking (commented):
Please make clang-format happy, whether by addressing it or turning it off at the appropriate places.

Also consider formatting the queries as was done in https://github.com/irods/irods/pull/8762 as well. Up to you

alanking:
> The non-root work will happen as part of #8667

#8667 is the issue linked in this PR. Are we going to keep it open til the non-root work happens, or make a separate issue for that work, or link a different issue for this work?

FifthPotato:
> > The non-root work will happen as part of #8667
> 
> #8667 is the issue linked in this PR. Are we going to keep it open til the non-root work happens, or make a separate issue for that work, or link a different issue for this work?

I think it makes sense to keep the issue open until the non-root work happens. This work doesn't do anything in and of itself (well, aside from the test case) but it's half the solution for the non-root work.

FifthPotato:
> Please make clang-format happy, whether by addressing it or turning it off at the appropriate places.
> 
> Also consider formatting the queries as was done in #8762 as well. Up to you

It is a little bit more difficult due to the macros used for the different DBMSes. I'll see what I can do.

alanking:
> It is a little bit more difficult due to the macros used for the different DBMSes. I'll see what I can do.

That's fine. If it proves messy/weird, just say "Won't fix" in the PR here and we'll call it a day.

trel:
so if this work is for main... and #8667 is 5.1.0...    that signals that we have no intention of cherry-picking this to 4.3?

just saying it out loud...

korydraughn:
@FifthPotato Please explain what this PR addresses now.

FifthPotato:
> @FifthPotato Please explain what this PR addresses now.

This PR adds the math for coordinating resources-- each coordinating resource will now have a usage record generated based on the sum of the usage of all its children. It also adds a basic test case for coordinating resources. Effectively, this is half of the work required to get coordinating resources fully working.

This PR does *not* perform the enforcement of the non-root resources, which is the other half. When we decide to do that, the enforcement code will need to call `setRescQuota` which will update `R_QUOTA_MAIN` based on these usage records.

If I plan on focusing on the multi-owner quota issue now, I suggest this be merged first, as otherwise tip-of-main will have the old queries, and if I modify those directly, merging this in the future will create a nasty merge conflict.

korydraughn:
Good. Let's put that comment in the issue so it's front and center. Link it to this PR so the story is clear.

korydraughn:
> so if this work is for main... and #8667 is 5.1.0... that signals that we have no intention of cherry-picking this to 4.3?
> 
> just saying it out loud...

That sounds correct given we aren't adding support for coordinating resources to 4.3.

@FifthPotato Please correct me if I'm wrong.

FifthPotato:
> Good. Let's put that comment in the issue so it's front and center. Link it to this PR so the story is clear.

Done



> > so if this work is for main... and #8667 is 5.1.0... that signals that we have no intention of cherry-picking this to 4.3?
> > just saying it out loud...
> 
> That sounds correct given we aren't adding support for coordinating resources to 4.3.
> 
> @FifthPotato Please correct me if I'm wrong.

I believe so. I don't see any reason why it wouldn't work for 4.3 but if we're worried about shaking up a stable branch too much, I can see why we're avoiding it.

korydraughn:
Okay. Let's limit this work to iRODS 5.

korydraughn:
What's left for this PR?

FifthPotato:
> What's left for this PR?

Squish and octothorpe, I think. Unless anyone else has any objections?

trel:
for main, i think we're good.  we've got time to find anything missing.

FifthPotato:
Squished

korydraughn (approved):
Pound it.

FifthPotato:
Octothorpe'd

alanking (approved):

PR #8759, opened 2025-12-03
[#8607] Fix zone reports for server-to-server connect errors (4-3-stable)

alanking:
Addresses #8607

New test fails before fix and passes after fix.

Running the rest of the tests now.

Issue #8607, opened 2025-08-06
iadmin moduser <username> type <user_type> failing in some circumstances.


tedgin:
### iRODS Version, OS and Version

iRODS 4.3.1, Ubuntu 18.04

### What did you try to do?

Here are four attempts that I made to use `iadmin moduser` to change the type of a user.  I was only able to change a type from `rodsuser` to `rodsadmin`.  All other attempts failed.

I have a few users with the custom user type `ds-service`. I'm in the process of upgrading to iRODS 4.3.1, which no longer supports custom user types. After upgrading the catalog providers in the dev environment, I tried to change these users to `rodsuser` type.

1. I try to use `iadmin` to change the user's type to `rodsuser`, but it fails with a `SYS_LIBRARY_ERROR`.
   
   ```console
   irods@dev-irods-csp-2:~$ iadmin moduser bisque type rodsuser
   remote addresses: 172.16.12.144 ERROR: rcGeneralAdmin failed with error -167000 SYS_LIBRARY_ERROR
   ```
   
   The following showed up in syslog.

   ```
   Aug  6 16:29:50 dev-irods-csp-2 irodsServer[20264]{"log_category":"api","log_level":"error","log_message":"[_rsGeneralAdmin:899] - [SYS_LIBRARY_ERROR: [json.exception.type_error.304] cannot use at() with array\n\n]","request_api_name":"GENERAL_ADMIN_AN","request_api_number":701,"request_api_version":"d","request_client_user":"rods","request_host":"172.16.12.144","request_proxy_user":"rods","request_release_version":"rods4.3.1","server_host":"dev-irods-csp-2","server_pid":20264,"server_timestamp":"2025-08-06T23:29:50.850Z","server_type":"agent","server_zone":"cyverse.dev"}
   ```

1. Since that didn't work, I tried to change the user's type to `rodsadmin`. It failed with a `CAT_INVALID_USER_TYPE` error.

   ```console
   irods@dev-irods-csp-2:~$ iadmin moduser bisque type rodsadmin
   remote addresses: 172.16.12.144 ERROR: rcGeneralAdmin failed with error -820000 CAT_INVALID_USER_TYPE 
   See 'lt user_type' for a list of valid user types.
   ```

   The following showed up in syslog.

   ```
[...]

PR #8760, opened 2025-12-03
 [#8607] Fix zone reports for server-to-server connect errors (main)

alanking:
Addresses #8607 

Cherry-picked from #8759

Need to run tests


alanking:
Core and unit tests passed

PR #8762, opened 2025-12-04
[8758] Fix quota processing logic (main)

FifthPotato:
Turns out, the quota system has been the victim of a lot of strange assumptions for the many years it's been here.

https://github.com/irods/irods/blob/7902cadf9e4ac9fcb317135d4a46dc81cb5b76a8/plugins/database/src/general_query.cpp#L1748-L1750

For starters, the quota query pair in `general_query.cpp` fetch all quotas from the database, regardless of which one is used. The only difference is that the second query it will only exclude irrelevant resource quotas (but will still fetch all total quotas as well)

It seems that we have been assuming that the first query only fetches total quotas, and the second only fetches resource quotas; this is untrue. The first query will fetch all queries (and it is done whenever a resource name is not provided).

Secondly, the code previously only checked the first row of the output. This was intended to check for the largest overrun (given the previous queries are `ORDER BY quota_over DESC`) but this instead actually checks the lowest overrun. This is because the linked list is constructed "traditionally" with the first element of the output being put at the **tail** of the list.

https://github.com/irods/irods/blob/7902cadf9e4ac9fcb317135d4a46dc81cb5b76a8/server/api/src/rsGetRescQuota.cpp#L160-L171

This is the real reason that https://github.com/irods/irods/issues/4089 point 1 occurred: the existence of another quota would shadow the total quota.

My fix in https://github.com/irods/irods/commit/8e5f20fbe3bbf21b845575e8497c12dcbad82a7e actually resolved this, although I didn't know it at the time, since I didn't know the full nature of these queries. Since the resource query (the second quota query) also fetched total quotas, it would simultaneously enforce the total quota, which masked this assumption completely.

Finally, as a fun note, the `istream` test on line 75 relied on this bug being present: it doesn't specify a resource to `istream` to, so it uses the default one, meaning that it only passed if this bug is here!

Pending a proper test case and a full test suite run. This passes `test_quotas` right now.

As a side note, I could rewrite this to work as it originally intended by reversing the linked list generation (or doing `ORDER BY quota_over ASC`) and then removing both of the linear searches in `rsGetRescQuota.cpp` which would be a little more performant. Maybe that would be preferable?

### DIFF
--- a/plugins/database/src/general_query.cpp
+++ b/plugins/database/src/general_query.cpp
@@ -1734,10 +1734,15 @@ int specialQueryIx( int ix ) {
 
  The caller specifies a user (COL_USER_NAME) and optionally a resource
  (COL_R_RESC_NAME).  Rows are returned with the quotas that apply,
- most severe (most over or closest to going over) first, if any.  For
+ least severe (least over or farthest to going over) first, if any.  For
  global-usage quotas, the returned RESC_ID is 0.  If the user is a
  member of a group with a quota (per-resource or total-usage) a row is
  returned for that quota too.  All in the appropriate order.
+
+ Note: We return rows least-severe first because the linked list of
+ quota structures is generated tail-to-head, so the head of the linked list
+ becomes the most-severe quota. See queRescQuota in rsGetRescQuota.cpp for
+ more details.
  */
 int
 generateSpecialQuery( genQueryInp_t genQueryInp, char *resultingSQL ) {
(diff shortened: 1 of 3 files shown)


alanking (commented):
Please notify when a regression test has been added. Also, please address clang-format and clang-tidy (even if that means adding `NOLINT` comments).

As for the suggestion of an alternative / more performant solution, consider implementing what you had in mind *after* you make a regression test. In this way, you will be able to effectively test both solutions.

FifthPotato:
Ok, did some rebasing and clang stuff. I implemented the "new, better way" and it passes all of `test_quotas` so far, so...reasonable vote that it works as expected?

scripts/irods/test/test_quotas.py
```diff
@@ -72,7 +72,7 @@ def acRescQuotaPolicy(rule_args, callback, rei):
                         lib.make_file(filename_2, 1024, contents='arbitrary')
                         cmd = 'iput -R {0} {1}'.format(self.testresc, filename_2) # should fail
                         self.admin.assert_icommand(cmd.split(), 'STDERR_SINGLELINE', 'SYS_RESC_QUOTA_EXCEEDED')
-                        cmd = 'istream write nopes'
+                        cmd = 'istream -R {0} write nopes'.format(self.testresc)
```
trel, line 75: should we keep the non-specific test as well?

plugins/database/src/general_query.cpp
```diff
@@ -1745,9 +1745,29 @@ generateSpecialQuery( genQueryInp_t genQueryInp, char *resultingSQL ) {
...
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays,modernize-avoid-c-arrays)
+    char quotaQuery1[] =
+        "( select distinct QM.user_id, RM.resc_name, QM.quota_limit, QM.quota_over, QM.resc_id from R_QUOTA_MAIN QM, "
```
trel: how do we feel about formatting these to be readable... while we're in here?

FifthPotato: Do you want me to do a multi-string-literal (like I did for the CTE) and add comments to each part of it?

scripts/irods/test/test_quotas.py
FifthPotato: I think the new test case covers "writing to unaffected resource," so the old one shouldn't be needed, I don't think. It's also a bit obscure as it was. If there's a case I'm missing, I'd be glad to add it, though.

trel: happy for it to be the way it is - just wanted to bring up coverage.   if you're good, i'm good.

plugins/database/src/general_query.cpp
trel: yes, please add a commit that does that work (we can squash before merging if we love it).   i think future-us will thank you.

FifthPotato: Done!

trel:
comments look good.

FifthPotato:
Made it even more readable by putting it through a query formatter.

plugins/database/src/general_query.cpp
```diff
@@ -1734,21 +1734,135 @@ int specialQueryIx( int ix ) {
...
-
+    // Fetches all resource and total quotas for the given user (accounting for the user's group as well)
+    // clang-format off
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays,modernize-avoid-c-arrays)
+    char quotaQuery1[] =
+        // Fetch all group resource quotas for a user
+        // Notice that UM2 is the user and is linked to UG to look up group membership
+          "(SELECT DISTINCT QM.user_id, "
```
trel, line 1759: unindent by two spaces, here and five other SELECTs

plugins/database/src/general_query.cpp
```diff
@@ -1734,21 +1734,135 @@ int specialQueryIx( int ix ) {
 
  The caller specifies a user (COL_USER_NAME) and optionally a resource
  (COL_R_RESC_NAME).  Rows are returned with the quotas that apply,
- most severe (most over or closest to going over) first, if any.  For
+ least severe (least over or farthest to going over) first, if any.  For
  global-usage quotas, the returned RESC_ID is 0.  If the user is a
  member of a group with a quota (per-resource or total-usage) a row is
  returned for that quota too.  All in the appropriate order.
+
+ NB: We return rows least-severe first because the linked list of
```
trel: 'NB' may not be understood by non-native speakers.

```suggestion
 Note: We return rows least-severe first because the linked list of
```

plugins/database/src/general_query.cpp
```diff
@@ -1734,21 +1734,135 @@ int specialQueryIx( int ix ) {
...
+                "R_USER_MAIN UM2, "
+                "R_RESC_MAIN RM "
+           "WHERE QM.resc_id = '0' "
+             "AND RM.resc_name = ? "
+             "AND (QM.user_id = UG.group_user_id "
+                  "AND UM2.user_name = ? "
+                  "AND UM2.zone_name = ? "
+                  "AND UG.user_id = UM2.user_id)) "
+        "UNION "
+        // Fetch all user quotas for a user and a particular resource
```
trel:
```suggestion
        // Fetch all user quotas for a user and a particular resource
        // Eventually can be removed - only group quotas can be set since 5.0
```

plugins/database/src/general_query.cpp
```diff
@@ -1734,21 +1734,135 @@ int specialQueryIx( int ix ) {
...
+                "R_USER_GROUP UG, "
+                "R_USER_MAIN UM2, "
+                "R_RESC_MAIN RM "
+           "WHERE QM.resc_id = '0' "
+             "AND (QM.user_id = UG.group_user_id "
+                  "AND UM2.user_name = ? "
+                  "AND UM2.zone_name = ? "
+                  "AND UG.user_id = UM2.user_id)) "
+        "UNION "
+        // Fetch all user quotas for a particular user
```
trel:
```suggestion
        // Fetch all user quotas for a particular user
        // Eventually can be removed - only group quotas can be set since 5.0
```

plugins/database/src/general_query.cpp
FifthPotato: I think this one is intentional from the formatter-- the selects are in parens, which it treats as being "inside" one level. Only the `UNION`s are top-level, so it doesn't indent those at all.
Do we still want to do this?

trel: fair enough, just caught my eye as they didn't line up.   resolving.

trel:
please to squash with a good/full commit message

server/api/src/rsGetRescQuota.cpp
```diff
@@ -217,10 +217,7 @@ int setRescQuota(
     }
 
     rodsLong_t resc_overrun = 0;
-    ret = irods::get_resource_property<rodsLong_t>(
-                           resc_id,
-                           irods::RESOURCE_QUOTA_OVERRUN,
-                           resc_overrun );
+    ret = irods::get_resource_property<rodsLong_t>(resc_id, irods::RESOURCE_QUOTA_OVERRUN, resc_overrun);
```
korydraughn, line 220: Are we checking the `ret` for errors?

scripts/irods/test/test_quotas.py
```diff
@@ -228,3 +228,50 @@ def acRescQuotaPolicy(rule_args, callback, rei):
...
+                self.admin.assert_icommand(['iadmin', 'atg', f'{group_name}_1', self.quota_user.username])
+
+                self.quota_user.assert_icommand(['iput', '-R', self.testresc, os.path.join(self.quota_user.local_session_dir, file_name)])
+
+                self.quota_user.assert_icommand(['icp', '-R', resc_name, file_name, f'{file_name}_on_{resc_name}'])
+                self.admin.assert_icommand(['iadmin', 'cu'])
+                self.admin.assert_icommand(['iadmin', 'lq'], 'STDOUT', ['\nquota_over: 3096\n'])
+
+                # Should pass and be unaffected by the other quota violation
+                self.quota_user.assert_icommand(['icp', '-R', other_resc_name, file_name, f'{file_name}_on_{other_resc_name}'])
```
korydraughn, line 270: Consider recalculating the quota after this and making sure its not modified.

FifthPotato:
Squished. Going to run full test suite now.

FifthPotato:
Core and unit tests pass.

korydraughn:
The test run included the error check in the cpp file and the new assertions in the python test?

FifthPotato:
> The test run included the error check in the cpp file and the new assertions in the python test?

Yep!
To be specific, it includes everything up to the static cast, because I ran clang-format/tidy after running the tests.

plugins/database/src/general_query.cpp
```diff
@@ -1734,21 +1734,137 @@ int specialQueryIx( int ix ) {
...
+                "R_USER_MAIN UM, "
+                "R_RESC_MAIN RM "
+           "WHERE (QM.resc_id = RM.resc_id "
+                  "OR QM.resc_id = '0') "
+             "AND RM.resc_name = ? "
+             "AND (QM.user_id = UM.user_id "
+                  "AND UM.user_name = ? "
+                  "AND UM.zone_name = ?)) "
+        "ORDER BY quota_over ASC";
+
```
korydraughn: Remove empty line.

plugins/database/src/general_query.cpp
```diff
@@ -1734,21 +1734,137 @@ int specialQueryIx( int ix ) {
...
-
+    // Fetches all resource and total quotas for the given user (accounting for the user's group as well)
+    // clang-format off
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays,modernize-avoid-c-arrays)
+    char quotaQuery1[] =
+        // Fetch all group resource quotas for a user
+        // Notice that UM2 is the user and is linked to UG to look up group membership
```
korydraughn: Let's add some periods to the end of each comment so there's no room for confusion.

_This applies to the other comments for this query._

plugins/database/src/general_query.cpp
```diff
@@ -1734,21 +1734,137 @@ int specialQueryIx( int ix ) {
...
+                           "QM.resc_id "
+           "FROM R_QUOTA_MAIN QM, "
+                "R_RESC_MAIN RM, "
+                "R_USER_GROUP UG, "
+                "R_USER_MAIN UM2 "
+           "WHERE QM.resc_id = RM.resc_id "
+             "AND (QM.user_id = UG.group_user_id "
+                  "AND UM2.user_name = ? "
+                  "AND UM2.zone_name = ? "
+                  "AND UG.user_id = UM2.user_id)) "
```
korydraughn, line 1772: Try feeding the SQL to an AI and see what the SQL would look like using explicit JOINs - e.g. INNER JOINs, LEFT JOINs, etc. whichever is appropriate.

Over time, it's much easier to understand queries when the conditions for JOINs and the WHERE-clause aren't mixed together.

Don't apply the refactored SQL to the PR. Posting what was generated as a response to this review comment is all that's being asked for now.

FifthPotato: This is what the AI gives me.

```
(
    SELECT DISTINCT
        QM.user_id,
        RM.resc_name,
        QM.quota_limit,
        QM.quota_over,
        QM.resc_id
    FROM R_QUOTA_MAIN  QM
    JOIN R_RESC_MAIN   RM ON QM.resc_id = RM.resc_id
    JOIN R_USER_GROUP  UG ON QM.user_id = UG.group_user_id
    JOIN R_USER_MAIN   UM2 ON UG.user_id = UM2.user_id
    WHERE UM2.user_name = ?
      AND UM2.zone_name = ?
)
UNION
(
    SELECT DISTINCT
        QM.user_id,
        RM.resc_name,
        QM.quota_limit,
        QM.quota_over,
        QM.resc_id
    FROM R_QUOTA_MAIN  QM
    JOIN R_USER_GROUP  UG ON QM.user_id = UG.group_user_id
    JOIN R_USER_MAIN   UM2 ON UG.user_id = UM2.user_id
    JOIN R_RESC_MAIN   RM
      ON 1 = 1
    WHERE QM.resc_id = '0'
      AND UM2.user_name = ?
      AND UM2.zone_name = ?
)
UNION
(
    SELECT DISTINCT
        QM.user_id,
        RM.resc_name,
        QM.quota_limit,
        QM.quota_over,
        QM.resc_id
    FROM R_QUOTA_MAIN QM
    JOIN R_USER_MAIN  UM ON QM.user_id = UM.user_id
    JOIN R_RESC_MAIN  RM 
         ON RM.resc_id = QM.resc_id
         OR QM.resc_id = '0'
    WHERE UM.user_name = ?
      AND UM.zone_name = ?
)
ORDER BY quota_over ASC;
```

korydraughn: What are your thoughts on the SQL?

The `ON 1 = 1` stands out to me. I don't see that in your original SQL.

FifthPotato: The AI's reasoning for that line is because the original 2nd block had R_RESC_MAIN joined in without any JOIN conditions on it.
I think it looks reasonably accurate, but I don't know if I'd trust it without rerunning tests. Another potential issue is that I'm not sure how JOIN syntax might differ between DBMSes.

korydraughn: I agree on not blindly trusting what it produced.

Take 10 to 20 minutes to read about the various JOIN types (inner, outer, left, right, full) to get a better understanding of them. That might unlock some new powers for you.

This isn't a blocker for the PR either. Feel free to punt this.

FifthPotato: Will punt for now, but I'll definitely read up on it.

FifthPotato:
Comments are resolved. Do we think we're ready for a final squish?

korydraughn:
If you're happy with the changes and tests are passing, squash it.

FifthPotato:
Squished.

FifthPotato:
Octothorpe'd

alanking:

Issue #8763, opened 2025-12-04
Resource type tests fail on Ubuntu 22.04 with CRC byteorder error


korydraughn:
## Bug Report

While testing #8761 on Ubuntu 22.04 / PostgreSQL 16, all resource type tests failed consistently. The error information is shown below. I think this is independent of the changes within the pull request.

```xml
        <testcase classname="irods.test.test_resource_types.Test_Resource_Deferred" name="test_server_side_calculates_crc64nvme_checksum_on_put_operation__issue_8554" time="12.755" timestamp="2025-12-04T04:38:58" file="scripts/irods/test/test_resource_types.py" line="500">
                <error type="TypeError" message="to_bytes() missing required argument 'byteorder' (pos 2)"><![CDATA[Traceback (most recent call last):
  File "/var/lib/irods/scripts/irods/test/resource_suite.py", line 526, in test_server_side_calculates_crc64nvme_checksum_on_put_operation__issue_8554
    checksum_str += base64.b64encode(lib.calculate_crc64_nvme(f.read())).decode()
  File "/var/lib/irods/scripts/irods/lib.py", line 895, in calculate_crc64_nvme
    return (crc ^ 0xFFFFFFFFFFFFFFFF).to_bytes(8) # Final XOR with all ones and return bytes
TypeError: to_bytes() missing required argument 'byteorder' (pos 2)
]]></error>
```

```xml
        <testcase classname="irods.test.test_resource_types.Test_Resource_Deferred" name="test_client_side_and_server_side_verify_crc64nvme_checksums_on_put_operation__issue_8554" time="14.384" timestamp="2025-12-04T04:26:50" file="scripts/irods/test/test_resource_types.py" line="535">
                <error type="TypeError" message="to_bytes() missing required argument 'byteorder' (pos 2)"><![CDATA[Traceback (most recent call last):
  File "/var/lib/irods/scripts/irods/test/resource_suite.py", line 561, in test_client_side_and_server_side_verify_crc64nvme_checksums_on_put_operation__issue_8554
    checksum_str += base64.b64encode(lib.calculate_crc64_nvme(f.read())).decode()
  File "/var/lib/irods/scripts/irods/lib.py", line 895, in calculate_crc64_nvme
    return (crc ^ 0xFFFFFFFFFFFFFFFF).to_bytes(8) # Final XOR with all ones and return bytes
TypeError: to_bytes() missing required argument 'byteorder' (pos 2)
]]></error>
```

Please investigate.

[...]

PR #8764, opened 2025-12-04
[8763] Defaulting byteorder in to_bytes to big.

JustinKyleJames:
The error described in #8763 was reproduced and the fix for this was verified against Ubuntu 22.04.

Also note that this setting is consistent with the default for `to_bytes` in newer versions of Python.  It is also consistent with the C++ CRC64/NVME code as shown below.

 https://github.com/irods/irods/blob/7902cadf9e4ac9fcb317135d4a46dc81cb5b76a8/lib/hasher/src/CRC64NVMEStrategy.cpp#L86-L90

### DIFF
--- a/scripts/irods/lib.py
+++ b/scripts/irods/lib.py
@@ -892,4 +892,6 @@ def calculate_crc64_nvme(data: bytes, initial_crc = 0):
             else:
                 crc >>= 1
 
-    return (crc ^ 0xFFFFFFFFFFFFFFFF).to_bytes(8) # Final XOR with all ones and return bytes 
+    # Byteorder defaults to 'big' in some versions of Python but not all.  Setting this
+    # to 'big' so it's consistent with CRC64NVMEStrategy.cpp.
+    return (crc ^ 0xFFFFFFFFFFFFFFFF).to_bytes(8, byteorder='big') # Final XOR with all ones and return bytes


scripts/irods/lib.py
```diff
@@ -892,4 +892,6 @@ def calculate_crc64_nvme(data: bytes, initial_crc = 0):
             else:
                 crc >>= 1
 
-    return (crc ^ 0xFFFFFFFFFFFFFFFF).to_bytes(8) # Final XOR with all ones and return bytes 
+    # issue 8763 - Byteorder defaults to 'big' in some versions of Python but not all.  Setting this
+    # to 'big' so that is consistent with how the cpp code sets it.
```
korydraughn: I think we can drop the issue number reference since that information is included in the commit message.
```suggestion
    # Byteorder defaults to 'big' in some versions of Python but not all.  Setting this
    # to 'big' so that it's consistent with how the C++ code sets it.
```

JustinKyleJames: I made the recommended change with a slight tweak:

"consistent with how the C++ code sets it" -> "consistent with the CRC64NVMEStrategy.cpp".

JustinKyleJames:
> Good stuff.
> 
> Spotted `of of` in the body of your commit message.

Fixed the commit message.

scripts/irods/lib.py
korydraughn: Looks good to me.

korydraughn:
Squash it.

JustinKyleJames:
> Squash it.

done

korydraughn (approved):

JustinKyleJames:
> \# it

done

PR #8767, opened 2025-12-05
[#8334] Fix memleaks reported by ASAN (4-3-stable)

alanking:
Addresses #8334

Companion PR: https://github.com/irods/irods_client_icommands/pull/597

Tests which experienced double-free on `iput` sometimes are now passing. Running full test suite now.

The ASAN test script is no longer reporting leaks, except for the one I found in `smsi_delayExec`. I backed out the change which fixed the leak because I do not fully understand how this will affect things that use the `inOutStruct` in the `msParam`s of the `msParamArray` (not) being freed here:
https://github.com/irods/irods/blob/7902cadf9e4ac9fcb317135d4a46dc81cb5b76a8/plugins/rule_engines/irods_rule_language/src/functions.cpp#L1878

It could affect the PREP or other REPs which power major iRODS capabilities. I opened https://github.com/irods/irods/issues/8768 for this one.

### DIFF
--- a/lib/core/src/miscUtil.cpp
+++ b/lib/core/src/miscUtil.cpp
@@ -15,9 +15,12 @@
 
 #include "irods/get_hier_from_leaf_id.h"
 
-#include <fstream>
 #include <boost/filesystem/operations.hpp>
 #include <boost/filesystem/convenience.hpp>
+
+#include <cstdlib>
+#include <cstring>
+#include <fstream>
 
 /* VERIFY_DIV - contributed by g.soudlenkov@auckland.ac.nz */
 #define VERIFY_DIV(_v1_,_v2_) ((_v2_)? (float)(_v1_)/(_v2_):0.0)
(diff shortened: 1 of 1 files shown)


alanking:
Core tests passed for RockyLinux 9 / PostgreSQL 14

lib/core/src/miscUtil.cpp
```diff
@@ -27,6 +27,41 @@ static uint Myumask = INIT_UMASK_VAL;
...
+        clearRodsPath(to);
+
+        // Copy in the non-pointer members.
+        *to = *from;
+
+        // If the rodsObjStat is valid, make a deep copy.
+        if (nullptr != from->rodsObjStat) {
+            // NOLINTNEXTLINE(cppcoreguidelines-owning-memory, cppcoreguidelines-no-malloc)
+            to->rodsObjStat = static_cast<rodsObjStat_t*>(std::malloc(sizeof(rodsObjStat_t)));
+            std::memset(to->rodsObjStat, 0, sizeof(rodsObjStat_t));
```
korydraughn, line 52: Are we including the necessary headers for malloc and memset?

lib/core/src/getUtil.cpp
```diff
@@ -58,6 +58,8 @@ getUtil( rcComm_t **myConn, rodsEnv *myRodsEnv, rodsArguments_t *myRodsArgs,
         }
     }
 
+    dataObjInp_t dataObjOprInp{};
+    const auto clear_cond_input = irods::at_scope_exit{[&dataObjOprInp] { clearKeyVal(&dataObjOprInp.condInput); }};
```
korydraughn, line 63: I think we need to include rcMisc.h for clearKeyVal.

Or maybe that's covered by rodsClient.h in getUtil.h?

lib/core/src/miscUtil.cpp
alanking: Added `cstring` and `cstdlib`

lib/core/src/getUtil.cpp
alanking: I added `rcMisc.h` anyway. It's got include guards.

alanking:
Squashed

alanking:
#'d, mergin

PR #8775, opened 2025-12-09
[#7383] Migrate optparse to argparse (main)

alanking:
Addresses #7383

I'm not really sure how to test this other than just running all the scripts that it touches. I've worked through kinks in `setup_irods.py` and `run_tests.py`. Probably should try `upgrade_irods.py` and the build hook. Not really sure how to test that shared_objects thing.

Anyway, running core tests now and things seem to be going well. I'll try topology and federation tests as well just in case.

Issue #7383, opened 2023-10-15
Python: migrate optparse and getopt usage to argparse


SwooshyCueb:
argparse has been recommended over optparse and getopt for years, and now optparse and getopt are actually deprecated in Python 3.13. Let's migrate.

SwooshyCueb:
I have no objections

korydraughn:
Okay. Moving back to 4.3.2 milestone.

PR #8776, opened 2025-12-10
[#7383] Migrate optparse to argparse (4-3-stable)

alanking:
Addresses #7383

Cherry-picked from https://github.com/irods/irods/pull/8775

I kept the script removals in a separate commit in case we wanted to keep those in 4-3-stable. Will squash if we decide to remove them here as well.


trel:
let's leave them here for 4.3.

korydraughn:
I'm thinking we leave 4.3's scripts as is and let the argparse switch be an iRODS 5 thing.

alanking:
Okay. Should I close this and move #7383 to 5.1.0, then?

korydraughn:
Yeah, let's do that.

Issue #8773, opened 2025-12-09
Provider topology tests fail on tip of main


MartinFlores751:
## Bug Report

The following test suites fail on the tip of main, all with the function `test_server_side_calculates_crc64nvme_checksum_on_put_operation__issue_8554`:
- `test_resource_types.Test_Resource_Compound`
- `test_resource_types.Test_Resource_CompoundWithMockarchive`
- `test_resource_types.Test_Resource_CompoundWithUnivmss`
- `test_resource_types.Test_Resource_Deferred`
- `test_resource_types.Test_Resource_MultiLayered`
- `test_resource_types.Test_Resource_NonBlocking`
- `test_resource_types.Test_Resource_Passthru`
- `test_resource_types.Test_Resource_Random`
- `test_resource_types.Test_Resource_RandomWithinRandom`
- `test_resource_types.Test_Resource_RandomWithinReplication`
- `test_resource_types.Test_Resource_Replication`
- `test_resource_types.Test_Resource_ReplicationToTwoCompound`
- `test_resource_types.Test_Resource_ReplicationToTwoCompoundResourcesWithPreferArchive`
- `test_resource_types.Test_Resource_ReplicationWithinReplication`

### iRODS Version, OS and Version
Tip of main, Ubuntu 24.04 (docker).

### What did you try to do?
Run `run_topology_tests.py` in `provider` mode.

### Expected behavior
The tests pass.

### Observed Behavior (including steps to reproduce, if applicable)

All the tracebacks seem to have the same most recent call.
Here's a snippet of the error:
```
Traceback (most recent call last):
  File "/var/lib/irods/scripts/irods/test/resource_suite.py", line 529, in test_server_side_calculates_crc64nvme_checksum_on_put_operation__issue_8554
    self.admin.assert_icommand(['ils', '-L', file1], 'STDOUT_SINGLELINE', checksum_str)
  File "/var/lib/irods/scripts/irods/test/session.py", line 162, in assert_icommand
    return assert_command(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/lib/irods/scripts/irods/test/command.py", line 76, in assert_command
    return _assert_helper(*args, should_fail=False, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/lib/irods/scripts/irods/test/command.py", line 104, in _assert_helper
    assert result
[...]

PR #8777, opened 2025-12-10
[8756] Skip CRC64/NVME tests while in topology. (main)

JustinKyleJames:
These tests change the default_hash_scheme to "crc64nvme" in server_config.json on the server that is running the test. In some configurations the actual resource written to is not the same as the server where the test is running. This means that the hash scheme won't match what the test is expecting.

Since we have no way to updated the server configuration on all servers, these tests will now be skipped when running in topology.

### DIFF
--- a/scripts/irods/test/resource_suite.py
+++ b/scripts/irods/test/resource_suite.py
@@ -497,6 +497,7 @@ class ResourceSuite(ResourceBase):
         if os.path.exists(datafilename):
             os.unlink(datafilename)
 
+    @unittest.skipIf(test.settings.RUN_IN_TOPOLOGY, "TODO(#6835): Requires support for modifying configuration of remote servers")
     def test_server_side_calculates_crc64nvme_checksum_on_put_operation__issue_8554(self):
         file1 = f'{inspect.currentframe().f_code.co_name}_f1'
         file1_size = 2*1024
@@ -532,6 +533,7 @@ class ResourceSuite(ResourceBase):
             lib.remove_file_if_exists(file1)
             IrodsController().reload_configuration()
 
+    @unittest.skipIf(test.settings.RUN_IN_TOPOLOGY, "TODO(#6835): Requires support for modifying configuration of remote servers")
     def test_client_side_and_server_side_verify_crc64nvme_checksums_on_put_operation__issue_8554(self):
         file1 = f'{inspect.currentframe().f_code.co_name}_f1'
         file1_size = 2*1024
--- a/scripts/irods/test/test_ichksum.py
+++ b/scripts/irods/test/test_ichksum.py
@@ -574,6 +574,7 @@ C- {5}:
             os.unlink(local_file)
             self.admin.assert_icommand(['irm', '-f', object_name])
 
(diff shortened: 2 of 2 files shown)


scripts/irods/test/resource_suite.py
```diff
@@ -497,6 +497,7 @@ def test_local_iput_upper_checksum(self):
         if os.path.exists(datafilename):
             os.unlink(datafilename)
 
+    @unittest.skipIf(test.settings.RUN_IN_TOPOLOGY, "Issue 8756 - skip for topology testing")
```
korydraughn: No need to include the issue number as that will be captured in the commit message.

Instead, you could rewrite this to talk about the future.
```
TODO(#6835): Requires support for modifying configuration of remote servers
```

scripts/irods/test/resource_suite.py
```diff
@@ -532,6 +533,7 @@ def test_server_side_calculates_crc64nvme_checksum_on_put_operation__issue_8554(
             lib.remove_file_if_exists(file1)
             IrodsController().reload_configuration()
 
+    @unittest.skipIf(test.settings.RUN_IN_TOPOLOGY, "Issue 8756 - skip for topology testing")
```
korydraughn: No need to include the issue number as that will be captured in the commit message.

Instead, you could rewrite this to talk about the future.
```
TODO(#6835): Requires support for modifying configuration of remote servers
```

scripts/irods/test/test_ichksum.py
```diff
@@ -574,6 +574,7 @@ def test_ichksum_server_rejects_invalid_server_config_options_for_checksum_read_
             os.unlink(local_file)
             self.admin.assert_icommand(['irm', '-f', object_name])
 
+    @unittest.skipIf(test.settings.RUN_IN_TOPOLOGY, "Issue 8756 - skip for topology testing")
```
korydraughn: No need to include the issue number as that will be captured in the commit message.

Instead, you could rewrite this to talk about the future.
```
TODO(#6835): Requires support for modifying configuration of remote servers
```

korydraughn:
The body of the commit message needs a small tweak.

`updated` -> `update`

JustinKyleJames:
> Do we need to link / reference #8773 as well, or is that a separate issue?

I didn't realize there was a second issue.  Will add.

scripts/irods/test/resource_suite.py
JustinKyleJames: done

scripts/irods/test/test_ichksum.py
JustinKyleJames: done

JustinKyleJames:
> The body of the commit message needs a small tweak.
> 
> `updated` -> `update`

I modified the commit message.

scripts/irods/test/resource_suite.py
JustinKyleJames: done

alanking (commented):
Yep, squash

JustinKyleJames:
> Yep, squash

Done

alanking (approved):
\# it

JustinKyleJames:
> \# it

Done

Issue #6813, opened 2023-01-04
Cannot delete objects with the `delete_object` access level


mstfdkmn:
### iRODS Version, OS and Version
iRODS 4.3.0 centos7

### What did you try to do?
I did try to delete objects on which I have the `delete_object` access level.

### Expected behavior
To be able to move objects to `trash` collection unless `-f` flag is used. If the `-f` flag is used then I expect not to see those objects anywhere in iRODS.

### Observed behavior (including steps to reproduce, if applicable)
I could not remove a collection that I have the `delete_object` access level. I could not remove a data object with the same access level either (without using -f flag).

user2 is granted with the delete_object access level to the test_acl_delete collecion by user1:
```
[user1@dev-u0137480 ~]$ imkdir test_acl_delete
[user1@dev-u0137480 ~]$ ichmod delete_object user2 /u0137480/home/user1/test_acl_delete
[user1@dev-u0137480 ~]$ ils -A /u0137480/home/user1/test_acl_delete
/u0137480/home/user1/test_acl_delete:
        ACL - user1#u0137480:own   user2#u0137480:delete_object
        Inheritance - Disabled
```

user2 now tries to remove the mentioned collection:
```
[user2@dev-u0137480 ~]$ ils -A /u0137480/home/user1/test_acl_delete
/u0137480/home/user1/test_acl_delete:
        ACL - user1#u0137480:own   user2#u0137480:delete_object
        Inheritance - Disabled
[user2@dev-u0137480 ~]$ irm -r /u0137480/home/user1/test_acl_delete
remote addresses: 10.0.2.100 ERROR: rmUtil: rm error for /u0137480/home/user1/test_acl_delete, status = -818000 status = -818000 CAT_NO_ACCESS_PERMISSION
```
logs:
```
irods stdout | {"log_category":"legacy","log_facility":"local0","log_level":"error","log_message":"mvCollToTrash: rcDataObjRename error for /u0137480/trash/home/user2/user1/test_acl_delete, status = -818000","request_api_name":"RM_COLL_AN","request_api_number":679,"request_api_version":"d","request_client_user":"user2","request_host":"10.0.2.100","request_proxy_user":"user2","request_release_version":"rods4.3.0","server_host":"dev-u0137480.irods.t.icts.kuleuven.be","server_pid":24366,"server_timestamp":"2023-01-04T12:39:20.452Z","server_type":"agent"}
```
the trash collection:
```
[...]

PR #8778, opened 2025-12-12
[8750] Quotas count users with ACL permission "own" instead of data_owner_name (main)

FifthPotato:
Leaving in draft for now since I want to bring up some questions before I start writing test cases.
Right now, this generates a usage row based on users with ACL "own" instead of data_owner_name. This creates some weird outcomes that are logical, but potentially not what we want.

1. Ownerless objects
Surrendering "own" on the object makes it not count against a quota at all:
```
irods@aa709bc1b671:~$ iadmin sgq public demoResc 3000
irods@aa709bc1b671:~$ truncate -s 2000 testfile
irods@aa709bc1b671:~$ iput testfile
irods@aa709bc1b671:~$ iadmin cu
irods@aa709bc1b671:~$ iadmin lq
Per resource quotas (if any) for users/groups:
user_name: public
zone_name: tempZone
resc_name: demoResc
quota_limit: 3000
quota_over: -1000
modify_ts: 2025-12-12.20:32:33

Global (total usage) quotas (if any) for users/groups:
No rows found
irods@aa709bc1b671:~$ ichmod delete_object rods testfile
irods@aa709bc1b671:~$ iadmin cu
irods@aa709bc1b671:~$ iadmin lq
Per resource quotas (if any) for users/groups:
user_name: public
zone_name: tempZone
resc_name: demoResc
quota_limit: 3000
quota_over: -3000
modify_ts: 2025-12-12.20:32:33

Global (total usage) quotas (if any) for users/groups:
No rows found
```

2. Multi-owner objects
Objects with multiple owners in the same group will count multiple times against that group's quota:
```
irods@aa709bc1b671:~$ iadmin sgq public demoResc 3000                           
irods@aa709bc1b671:~$ iadmin mkuser alice rodsuser
irods@aa709bc1b671:~$ truncate -s 2000 testfile
irods@aa709bc1b671:~$ iput testfile                                            
irods@aa709bc1b671:~$ iadmin cu
irods@aa709bc1b671:~$ iadmin lq
Per resource quotas (if any) for users/groups:
user_name: public  
zone_name: tempZone
resc_name: demoResc           
quota_limit: 3000                      
quota_over: -1000                                   
modify_ts: 2025-12-12.20:34:38                                                 
                                                                               
Global (total usage) quotas (if any) for users/groups:
No rows found                                                                  
irods@aa709bc1b671:~$ iadmin lg public                                                                                                                         
Members of group public:                                                       
alice#tempZone                                                                 
rods#tempZone                                                                  
irods@aa709bc1b671:~$ ichmod own alice testfile
irods@aa709bc1b671:~$ iadmin cu                     
irods@aa709bc1b671:~$ iadmin lq                                                
Per resource quotas (if any) for users/groups:
user_name: public  
zone_name: tempZone
resc_name: demoResc
quota_limit: 3000             
quota_over: 1000                       
modify_ts: 2025-12-12.20:35:27                      
                                                                               
Global (total usage) quotas (if any) for users/groups:
No rows found
```

Note that this also applies to groups (as the user quota part of the QUOTA_QUERY will catch this case).

Do we have any comments about these cases, or is this all about as expected?

### DIFF
--- a/plugins/database/src/db_plugin.cpp
+++ b/plugins/database/src/db_plugin.cpp
@@ -11826,24 +11826,52 @@ irods::error db_calc_usage_and_quota_op(
        4096, s1_id, 10001, timestamp
        8192, c0_id, 10001, timestamp */
 
+    // clang-format off
     status = cmlExecuteNoAnswerSql(
-        "INSERT INTO R_QUOTA_USAGE (quota_usage, resc_id, user_id, modify_ts) " IRODS_QUOTA_WITH_RECURSIVE
         // Begin CTE: recursive table
-        "resc_usage(quota_usage, resc_id, resc_parent, user_id) AS ("
+        "INSERT INTO R_QUOTA_USAGE (quota_usage, resc_id, user_id, modify_ts)" IRODS_QUOTA_WITH_RECURSIVE "resc_usage(quota_usage, resc_id, resc_parent, user_id) AS "
         // Base case: "old" query that calculates usage on storage resources only
-        "SELECT SUM(R_DATA_MAIN.data_size), R_RESC_MAIN.resc_id, R_RESC_MAIN.resc_parent, R_USER_MAIN.user_id FROM "
-        "R_DATA_MAIN, R_USER_MAIN, R_RESC_MAIN WHERE R_USER_MAIN.user_name = R_DATA_MAIN.data_owner_name AND "
-        "R_USER_MAIN.zone_name = R_DATA_MAIN.data_owner_zone AND R_RESC_MAIN.resc_id = R_DATA_MAIN.resc_id GROUP BY "
-        "R_RESC_MAIN.resc_id, user_id, R_RESC_MAIN.resc_parent UNION ALL "
+        // Usage is calculated by any user with the "own" permission
+          "(SELECT SUM(R_DATA_MAIN.data_size), "
+                  "R_RESC_MAIN.resc_id, "
+                  "R_RESC_MAIN.resc_parent, "
+                  "R_USER_MAIN.user_id "
+           "FROM R_DATA_MAIN, "
+                "R_USER_MAIN, "
+                "R_RESC_MAIN, "
+                "R_OBJT_ACCESS, "
+                "R_TOKN_MAIN "
+           "WHERE R_DATA_MAIN.data_id = R_OBJT_ACCESS.object_id "
+             "AND R_OBJT_ACCESS.user_id = R_USER_MAIN.user_id "
+             "AND R_TOKN_MAIN.token_namespace = 'access_type' "
+             "AND R_OBJT_ACCESS.access_type_id = R_TOKN_MAIN.token_id "
+             "AND R_TOKN_MAIN.token_name = 'own' "
+             "AND R_RESC_MAIN.resc_id = R_DATA_MAIN.resc_id "
+           "GROUP BY R_RESC_MAIN.resc_id, "
+                    "R_USER_MAIN.user_id, "
+                    "R_RESC_MAIN.resc_parent "
         // Recursive case: each parent of a storage resource gets a corresponding usage row for a child usage row
-        "SELECT resc_usage.quota_usage, (CASE WHEN COALESCE(resc_usage.resc_parent, '') = '' THEN NULL ELSE "
-        "CAST(resc_usage.resc_parent " IRODS_QUOTA_CAST_TYPE_AS
-        ") END ), R_RESC_MAIN.resc_parent, resc_usage.user_id FROM resc_usage, R_RESC_MAIN WHERE R_RESC_MAIN.resc_id = "
-        "(CASE WHEN COALESCE(resc_usage.resc_parent, '') = '' THEN NULL ELSE "
-        "CAST(resc_usage.resc_parent " IRODS_QUOTA_CAST_TYPE_AS ") END )) "
+           "UNION ALL SELECT resc_usage.quota_usage, (CASE "
+                                                         "WHEN COALESCE(resc_usage.resc_parent, '') = '' THEN NULL "
+                                                         "ELSE CAST(resc_usage.resc_parent" IRODS_QUOTA_CAST_TYPE_AS ") "
+                                                     "END), R_RESC_MAIN.resc_parent, "
+                                                           "resc_usage.user_id "
+           "FROM resc_usage, "
+                "R_RESC_MAIN "
+           "WHERE R_RESC_MAIN.resc_id = (CASE "
+                                            "WHEN COALESCE(resc_usage.resc_parent, '') = '' THEN NULL "
+                                            "ELSE CAST(resc_usage.resc_parent" IRODS_QUOTA_CAST_TYPE_AS ") "
+                                        "END)) "
         // Sum up usage rows by resource and user id
-        "SELECT SUM(quota_usage), resc_id, user_id, ? FROM resc_usage GROUP BY resc_id, user_id",
+        "SELECT SUM(quota_usage), "
+               "resc_id, "
+               "user_id, "
+               "? "
+        "FROM resc_usage "
+        "GROUP BY resc_id, "
+                 "user_id",
         &icss);
+    // clang-format on
 #undef IRODS_QUOTA_CAST_TYPE_AS_STRING
 #undef IRODS_QUOTA_WITH_RECURSIVE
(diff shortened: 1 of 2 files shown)


trel:
case 1) ownerless objects

I think 'ownerless objects' is behaving as it should / would have to.  Unless we're going to count a different permission level (perhaps... write?) as counting against quotas, the current behavior of not counting them makes sense to me.  I do NOT think any other permission than 'own' makes sense to 'count' at this time.

case 2) multi-owner objects

On first thought... I'm not sure how it could be counted any differently.  So it seems correct.

Then... alternatively to the way you've demonstrated... it could be considered that because alice and rods are both members of public... then the file should only count once against the group quota they are both a part of (and by extension, any groups they are part of).   This suggests that the SQL would have to elide all members of a group and keep totals at the group level... Hmm...  Could be worth looking into, but... harder to explain and document?

It's definitely easier to explain the current behavior... "Every owner of a data object is charged quota".

Interesting for sure.

plugins/database/src/db_plugin.cpp
```diff
@@ -11827,22 +11827,48 @@ irods::error db_calc_usage_and_quota_op(
...
+           "FROM R_DATA_MAIN, "
+                "R_USER_MAIN, "
+                "R_RESC_MAIN, "
+                "R_OBJT_ACCESS, "
+                "R_TOKN_MAIN "
+           "WHERE R_DATA_MAIN.data_id = R_OBJT_ACCESS.object_id "
+             "AND R_OBJT_ACCESS.user_id = R_USER_MAIN.user_id "
+             "AND R_TOKN_MAIN.token_namespace = 'access_type' "
+             "AND R_OBJT_ACCESS.access_type_id = R_TOKN_MAIN.token_id "
+             "AND R_TOKN_MAIN.token_name = 'own' "
```
trel, line 11848: you could maybe(?) skip a join with the token table and just hard code to 1200 rather than dereference 'own'?

what you have is definitely generic and seems correct.   but perhaps slower?   could be useful to test if my assertion above is correct.   can always just mark as a TODO once we determine the queries are correct.

FifthPotato: I think it'll always be slower than a hardcoded value for sure. That being said, though, the ACL values being in a table seems to imply that changing them is allowed. The performance loss is also slightly mitigated by the fact that there are indices set by default on R_TOKN_MAIN, making the lookup pretty snappy. Either way is fine by me, however.

FifthPotato:
> case 1) ownerless objects
> 
> I think 'ownerless objects' is behaving as it should / would have to. Unless we're going to count a different permission level (perhaps... write?) as counting against quotas, the current behavior of not counting them makes sense to me. I do NOT think any other permission than 'own' makes sense to 'count' at this time.

I think this is reasonable. I think the primary issue I personally take with it is that users can surrender permissions as a way of arbitrarily bypassing quota. If the quota is meant to be a way to limit writing to whatever physical resource (cloud, datacenter, bunch-o-disks) it may not perform its job very well if a user can choose to just not contribute to the quota by saying "hey, actually, I don't own this thing!"

> case 2) multi-owner objects
> 
> On first thought... I'm not sure how it could be counted any differently. So it seems correct.
> 
> Then... alternatively to the way you've demonstrated... it could be considered that because alice and rods are both members of public... then the file should only count once against the group quota they are both a part of (and by extension, any groups they are part of). This suggests that the SQL would have to elide all members of a group and keep totals at the group level... Hmm... Could be worth looking into, but... harder to explain and document?
> 
> It's definitely easier to explain the current behavior... "Every owner of a data object is charged quota".
> 
> Interesting for sure.

I think that keeping totals at the group level is a reasonable approach, but it's may be a slightly difficult query to write. Namely, the primary thing that needs to be done is to ensure each data object is only counted once, so as far as I know, this requires at minimum a subquery in order to filter out duplicate data objects before doing a sum. The alternative is doing a group-aware megaquery that does both the usages and the over_quota calculation in one go, but that's even more messy.

FifthPotato:
@trel I saw your response to my e-mail. If we think all of this is OK, I can take this out of draft and begin working on test cases.

plugins/database/src/db_plugin.cpp
trel: let's do the real lookup for now as is - like you say, the table could change (although who knows the other fallout from such a change...).

trel:
surrendering permissions is okay - the admins would have to help them get it back if they wanted it... so not much a real threat i think.

and yes, agreed, let's not do more messy group math for now.   let it ride.

please remove from draft.

alanking (commented):
Agreed on the multi-owner and owner-less objects analysis.

Please quell clang-format. Also consider adding some tests.

FifthPotato:
> Agreed on the multi-owner and owner-less objects analysis.
> 
> Please quell clang-format. Also consider adding some tests.

Done and done. How does it all look now?

scripts/irods/test/test_quotas.py
```diff
@@ -349,3 +349,68 @@ def acRescQuotaPolicy(rule_args, callback, rei):
...
+                self.admin.assert_icommand(['iadmin', 'lq'], 'STDOUT', ['\nquota_over: -1808\n'])
+
+                # Granting the group "own" should count against quota
+                self.quota_user.assert_icommand(['ichmod', 'own', group_name, f'{file_name}_new'])
+                self.admin.assert_icommand(['iadmin', 'cu'])
+                self.admin.assert_icommand(['iadmin', 'lq'], 'STDOUT', ['\nquota_over: 2288\n'])
+                self.quota_user.assert_icommand(['icp', '-R', resc_name, file_name, f'{file_name}_shouldntexist'], 'STDERR_SINGLELINE', 'SYS_RESC_QUOTA_EXCEEDED')
+
+                # Removing owner's "own" permission should no longer count against quota
+                self.quota_user.assert_icommand(['ichmod', 'modify_object', self.quota_user.username, f'{file_name}_new'])
```
alanking, line 406: Removal of the "new" file will (silently) fail in the `finally` block below since we removed the owner's `own` permission here. Make sure the ownership is restored before the removal, or have the `self.admin` user grant itself `own` permission and remove the object.

scripts/irods/test/test_quotas.py
```diff
@@ -349,3 +349,68 @@ def acRescQuotaPolicy(rule_args, callback, rei):
...
+            with temporary_core_file() as core:
+                core.add_rule(pep_map[self.plugin_name])
+                IrodsController().reload_configuration()
+
+                group_name = 'test_group_acl'
+                file_name = 'test_file_acl'
+                resc_name = 'ufs0_quota_acl'
+                user_name = 'test_user_acl'
+
+                other_quota_user = session.mkuser_and_return_session('rodsuser', user_name, 'quotapass', lib.get_hostname())
```
alanking: Why create a session for this user when we don't seem to be using it?

plugins/database/src/db_plugin.cpp
```diff
@@ -11870,6 +11871,7 @@ irods::error db_calc_usage_and_quota_op(
         "GROUP BY resc_id, "
                  "user_id",
         &icss);
+// clang-format on
```
alanking: Increase indentation by one

scripts/irods/test/test_quotas.py
FifthPotato: I could also bump it up to `delete_object` in the test. It'd do the same thing, I think?

scripts/irods/test/test_quotas.py
FifthPotato: I thought I'd need it, but I didn't. Changing!

plugins/database/src/db_plugin.cpp
FifthPotato: Done.

scripts/irods/test/test_quotas.py
alanking: In theory, that is correct. In reality, the expanded permissions model has some issues. I think you are going to run into #6813 with `delete_object`, unfortunately.

FifthPotato: Issue for everything, huh? Alright, I'll do it the proper way.

FifthPotato:
Squished!

FifthPotato:
Octothorpe'd

alanking:
Please cherry-pick to 4-3-stable. Thanks

FifthPotato:
Forgot to mention above-- core and unit tests pass.

PR #8779, opened 2026-01-05
[#7922] Add output assertions to test_msi_replica_truncate (main)

alanking:
Addresses #7922

`test_all_rules.test_msi_replica_truncate` passes

Issue #7922, opened 2024-07-24
`test_all_rules.test_msi_replica_truncate`: Add assertions about output string


alanking:
- [x] main
- [x] 4-3-stable

---

Extensive tests for validating the contents of the output string in the `replica_truncate` API are being included with the resolution of https://github.com/irods/irods/issues/7918. However, there are also tests for `msi_replica_truncate` which are ignoring the output string contents rather explicitly:
https://github.com/irods/irods/blob/11648659c8e94c902d262c408975c6ed623be8ab/scripts/irods/test/test_all_rules.py#L2454-L2458

We should add assertions around the output strings in these tests as well because it is on the microservice to ensure that the rule from which it is being called has access to the output string, as is being done here:
https://github.com/irods/irods/blob/11648659c8e94c902d262c408975c6ed623be8ab/server/re/src/reDataObjOpr.cpp#L3381-L3390

PR #8781, opened 2026-01-06
[#7922] Add output assertions to test_msi_replica_truncate (4-3-stable)

alanking:
Addresses #7922 

Cherry-picked from #8779


alanking:
Confirmed test passed, just in case.

#'d, mergin

Issue #8301, opened 2025-03-18
Deprecate osauth scheme


alanking:
- [ ] 4-3-stable only

---

### Deprecation

Removal is in #8300.

korydraughn:
The current plan is to have the server log a warning on every use of the osauth auth scheme.

This will alert the local admin and trigger them to migrate users from using the deprecated auth scheme.

korydraughn:
Given the challenges of testing osauth, the deprecation of the auth scheme will be documented in the release notes. No changes to the server or clients will take place.

I've closed PR #8783.

korydraughn:
Deprecation of osauth documented in release notes.

See irods/irods_docs#406.

PR #8783, opened 2026-01-06
[#8301] Log warning on use of deprecated "osauth" auth scheme (4-3-stable)

korydraughn:
Putting in draft until I can confirm that the log message appears on every use of the plugin.


korydraughn:
There doesn't appear to be any documentation which demonstrates how to use the osauth scheme.

It's unlikely that anyone running 4.3+ is using osauth. The work to deprecate osauth is starting to look like a single line in the release notes.

korydraughn:
The deprecation of osauth will be reported in the release notes.

As I discovered yesterday, the osauth plugins cannot be triggered via 4.3 clients. You would need an older client to exercise the auth scheme.

Closing.

Issue #8733, opened 2025-11-11
`ips` does not report clients with names containing spaces


korydraughn:
- [x] main
- [x] 4-3-stable
---
## Bug Report

This was observed against iRODS 4.3.4 and 5.x.

In my efforts to update iRODS support in Cyberduck, I updated the client to report a name which is produced by the Cyberduck application. This turned out to be the following.

```
Cyberduck/9.3.0.44005 (Windows 10/10.0.19045.0) (amd64)
```

I soon learned that `ips` was no longer reporting connections from Cyberduck. However, interactions between Cyberduck and iRODS continued working just fine.

Turns out that iRODS doesn't handle more complex client names. This is because it uses whitespace to delimit information.

#### iRODS 5

https://github.com/irods/irods/blob/8fc618789753f38d59db40cc87276e89aecbb6d4/server/main_server/src/agent_main.cpp#L1137-L1149

https://github.com/irods/irods/blob/8fc618789753f38d59db40cc87276e89aecbb6d4/server/api/src/rsProcStat.cpp#L34-L49

#### iRODS 4.3

https://github.com/irods/irods/blob/2b1fa9a6bbb59eadd896c66a43515d935db68016/server/core/src/procLog.cpp#L95-L103

PR #8784, opened 2026-01-06
[#8733] Replace spaces with hyphens for agent info (4-3-stable)

korydraughn:
Putting this in draft until I can determine if there's a way to test this without needing a custom client. I want to use an existing client and override its SP_OPTION value. It's not clear to me that we have something that can do that.

### DIFF
--- a/server/core/src/procLog.cpp
+++ b/server/core/src/procLog.cpp
@@ -9,9 +9,10 @@
 
 #include <boost/lexical_cast.hpp>
 
+#include <algorithm>
+#include <charconv>
 #include <cstdio>
 #include <ctime>
-#include <charconv>
 #include <fstream>
 
 int
@@ -91,15 +92,21 @@ logAgentProc( rsComm_t *rsComm ) {
         return UNIX_FILE_OPEN_ERR - errno;
     }
 
+    const auto replace_spaces = [](char* _s, std::size_t _size) -> char* {
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic)
+        std::replace(_s, _s + _size, ' ', '-');
+        return _s;
+    };
+
     // NOLINTNEXTLINE(cppcoreguidelines-pro-type-vararg)
     std::fprintf(fptr,
                  "%s %s %s %s %s %s %u\n",
-                 rsComm->clientUser.userName,
-                 clientZone,
-                 rsComm->proxyUser.userName,
-                 proxyZone,
-                 progName,
-                 remoteAddr,
+                 replace_spaces(rsComm->clientUser.userName, std::strlen(rsComm->clientUser.userName)),
+                 replace_spaces(clientZone, std::strlen(clientZone)),
+                 replace_spaces(rsComm->proxyUser.userName, std::strlen(rsComm->proxyUser.userName)),
+                 replace_spaces(proxyZone, std::strlen(proxyZone)),
+                 replace_spaces(progName, std::strlen(progName)),
+                 replace_spaces(remoteAddr, std::strlen(remoteAddr)),
                  static_cast<unsigned int>(std::time(nullptr)));
 
     rsComm->procLogFlag = PROC_LOG_DONE;


korydraughn:
Turns out it wasn't too bad having to implement a special client to test the changes.

---

Unrelated to that, it occurred to me that it would be more efficient if the replacement code was executed during agent initialization. However, that only makes sense if admins are constantly running `ips`.

---

The test passes. Taking out of draft now.

test/issue_8733/main.cpp
```diff
@@ -0,0 +1,40 @@
...
+#include <unistd.h>
+
+#include <csignal>
+
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
+volatile std::sig_atomic_t g_stop = 0;
+
+int main()
+{
+    set_ips_display_name("Cyberduck/9.3.0.44005 (Windows 10/10.0.19045.0) (amd64)");
```
alanking: Would it be interesting to make the string a command line argument? We don't need to make it robust or anything - it could just be `argv[1]`...

scripts/irods/test/test_ips.py
```diff
@@ -0,0 +1,45 @@
...
+        # name to the server as 'Cyberduck/9.3.0.44005 (Windows 10/10.0.19045.0) (amd64)'.
+        # Notice the spaces in the name.
+        proc = lib.execute_command_nonblocking(['irods_test_issue_8733'])
+
+        # Give the computer a little time for the process to fully launch.
+        time.sleep(2)
+
+        # Show that server was able to replace spaces in the name reported to the
+        # server by the "irods_test_issue_8733" client.
+        self.user.assert_icommand(['ips'], 'STDOUT', ['  Cyberduck/9.3.0.44005-(Windows-10/10.0.19045.0)-(amd64)  '])
```
alanking: If this assertion fails, we never send the `SIGINT` below. Might want to put a try-finally around this for guaranteed clean exit. It's not as important to clean up here as it is with other tests, though.

scripts/irods/test/test_ips.py
```diff
@@ -0,0 +1,45 @@
...
+    def setUp(self):
+        super(Test_Ips, self).setUp()
+        self.user = self.user_sessions[0]
+
+
+    def tearDown(self):
+        super(Test_Ips, self).tearDown()
+
+
+    @unittest.skipIf(test.settings.RUN_IN_TOPOLOGY, "Skip for Topology Testing")
```
alanking: Don't need to skip this in topology

server/core/src/procLog.cpp
```diff
@@ -91,15 +91,21 @@ logAgentProc( rsComm_t *rsComm ) {
         return UNIX_FILE_OPEN_ERR - errno;
     }
 
+    const auto replace_spaces = [](char* _s, std::size_t _size) -> char* {
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic)
+        std::replace(_s, _s + _size, ' ', '-');
```
alanking, line 97: Do we need the `<algorithm>` header because of this, or nah?

alanking (commented):
Pretty clean. Please appease or silence clang-tidy.

test/issue_8733/main.cpp
korydraughn: Doesn't hurt to do that. Could be useful one day.

scripts/irods/test/test_ips.py
korydraughn: Will update.

test/issue_8733/main.cpp
korydraughn: Done.

scripts/irods/test/test_ips.py
korydraughn: Done.

scripts/irods/test/test_ips.py
korydraughn: Done.

server/core/src/procLog.cpp
korydraughn: Done.

korydraughn:
Fleshed out test a bit more. The client too.

The test passes when run in a single node config and in a topology for provider/consumer.

This is ready for final review.

scripts/irods/test/test_ips.py
```diff
@@ -21,25 +20,25 @@ def tearDown(self):
...
+                ['ips'], 'STDOUT', ['  Cyberduck/9.3.0.44005-(Windows-10/10.0.19045.0)-(amd64)  '])
 
-        # Wait for the client to exit cleanly.
-        proc.send_signal(signal.SIGINT)
-        try:
+            # Instruct the client to exit cleanly.
+            proc.send_signal(signal.SIGINT)
             proc.wait(timeout=10)
+
         except:
```
alanking: This will catch `unittest.AssertionError`s as well, which would silence test failures. In order for this approach to work, we would need to `raise` any `AssertionError`s which are caught. If we only handle the `TimeoutExpired` `Exception` here specifically, then we are back to the first problem in which this won't run when an `AssertionError` is raised.

scripts/irods/test/test_ips.py
```diff
@@ -0,0 +1,51 @@
...
+            if not test.settings.TOPOLOGY_FROM_RESOURCE_SERVER:
+                self.user.assert_icommand(['ips'], 'STDOUT', expected_output)
+
+            # Instruct the client to exit cleanly.
+            proc.send_signal(signal.SIGINT)
+            proc.wait(timeout=10)
+
+        finally:
+            proc.kill();
+            proc.wait();
```
alanking, line 51: Is the intention here to exit quickly (rather than "cleanly") on test failures? Just making sure I understand the intent since I happened to see the original implementation.

alanking (commented):
Looks good. Had one lingering concern about the try-finally thing

scripts/irods/test/test_ips.py
korydraughn: The original implementation which used `except` was incorrect. I actually meant to use `finally`.

With that said, we just want to stop the app, cleanly is preferred, but not required. The OS will clean up the resources. I don't expect we'll ever see issues since the app isn't doing anything but waiting for a signal.

alanking: Okay, I just wanted to make sure that the behavior here was intentional. It looks like the goal is being accomplished, so I will resolve this.

korydraughn:
Squashed.

korydraughn:
Added pounds.

Issue #8746, opened 2025-11-14
Enable session token usage in `pam_password`


alanking:
### New Feature / Functionality

Session tokens are being added in #8697. These could be used in place of time-limited legacy passwords in `R_USER_PASSWORD`. We can introduce a configuration which allows for switching to session tokens for `pam_password` authentication. Note: This might have to also include a configuration on the client side, which could get messy.

PR #8787, opened 2026-01-07
[#8789] Add password_reuse_previous option for pam generated passwords (main)

peterverraedt:
Add the pam_reuse_previous option to be able to disable reuse of previously generated temporary passwords. This is useful when server administrators want to implement proper session timeouts across multiple client machines.

E.g. suppose a user has two devices (work-laptop and hpc-cluster) and authenticates on Monday on hpc-cluster with pam_interactive and a ttl of 7 days. On Wednesday the user authenticates on work-laptop with a ttl of 1 day.

With the default settings (password_extend_lifetime = 1, password_reuse_previous = 1), the session on hpc-cluster will terminate on Thursday instead of next Monday (what the user would expect).

With the already existing option password_extend_lifetime set to 0, the user will get a session on work-laptop that is actually valid until next Monday instead of the expected TTL of 1 day.

With the new option password_reuse_previous = 0, each device will get its own generated native password, and own ttl.

I know you're implemented new session tokens for the upcoming release. Probably pam_interactive and pam_password flows will use the session tokens at some point (see #8746), but as using the session tokens currently requires to use another authentication scheme, adopting this might only be possible after a longer adoption period (all clients need an update). If this is true, we need to patch the current behaviour too to be able to adopt pam_interactive now.

### DIFF
--- a/scripts/irods/test/test_pam_password_authentication.py
+++ b/scripts/irods/test/test_pam_password_authentication.py
@@ -506,6 +506,117 @@ class test_configurations(unittest.TestCase):
             self.auth_session.assert_icommand(
                 ['iinit'], 'STDOUT', 'iRODS password', input=f'{self.auth_session.password}\n')
 
+    def test_password_reuse_previous_set_to_false_has_independent_sessions(self):
+        import time
+
+        min_time_option_name = 'password_min_time'
+        reuse_previous_option_name = 'password_reuse_previous'
+
+        # Stash away the original configuration for later...
+        original_min_time = self.admin.assert_icommand(
+            ['iadmin', 'get_grid_configuration', self.configuration_namespace, min_time_option_name],
+            'STDOUT')[1].strip()
+
+        original_reuse_previous = self.admin.assert_icommand(
+            ['iadmin', 'get_grid_configuration', self.configuration_namespace, reuse_previous_option_name],
+            'STDOUT')[1].strip()
+
+        # Set reuse_previous to 0 so that the a new randomly generated password is retrieved every time.
+        self.admin.assert_icommand(
+            ['iadmin', 'set_grid_configuration', self.configuration_namespace, reuse_previous_option_name, '0'])
+
+        # Make a new session of the existing auth_user. The data is "managed" in the session, so the session
+        # collection shall be shared with the other session. Re-auth first, just in case.
+        self.other_auth_session.assert_icommand(
+            ['iinit'], 'STDOUT', input=f'{self.auth_session.password}\n')
+        self.other_auth_session.assert_icommand(['icd', self.auth_session.session_collection])
+
+        auth_session_env_backup = copy.deepcopy(self.auth_session.environment_file_contents)
+        other_auth_session_env_backup = copy.deepcopy(self.other_auth_session.environment_file_contents)
+        try:
+            client_update = {'irods_authentication_scheme': self.authentication_scheme}
+            self.auth_session.environment_file_contents.update(client_update)
+            self.other_auth_session.environment_file_contents.update(client_update)
+
+            # Set the minimum time to a very short value so that the password expires in a reasonable amount of
+            # time for testing purposes. This value should be higher than 5 seconds to ensure steps in the test
+            # have enough time to complete.
+            ttl = 6
+            self.assertGreater(ttl, 5)
+            option_value = str(ttl)
+            self.admin.assert_icommand(
+                ['iadmin', 'set_grid_configuration', self.configuration_namespace, min_time_option_name, option_value])
+
+            # [0,0] seconds after start
+            # Authenticate with session 1. We assume these calls take less than 0.1 seconds in time.
+            self.auth_session.assert_icommand(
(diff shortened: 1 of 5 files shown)


trel:
seems good and reasonable on first read.

we'll also need to add a default value to the database on install and upgrade.

peterverraedt:
Can I include this change in scripts/irods/database_upgrade.py in schema version 13 or should this come in a next iteration?

The documentation of [iadmin.cpp](https://github.com/irods/irods_client_icommands/blob/main/src/iadmin.cpp) should also be adjusted, but I guess that is just cosmetic.

alanking:
Thanks for this. 

We will need to create an issue and include it in the 5.1.0 milestone. We will also need to include at least one test and ensure that the existing tests pass. 

Also, our semantic versioning directive requires that the default value for the new option reflects the existing behavior so that it can be included in a minor release. Can you confirm that the system behavior is unchanged with the default configuration? It sounds like the default configuration may change behavior for systems with `password_extend_lifetime` set to 0. I might be misunderstanding something.

---

> Can I include this change in scripts/irods/database_upgrade.py in schema version 13 or should this come in a next iteration?

Where you put the upgrade logic is fine. That's the schema version planned for 5.1.0. Should we decide to not include this in 5.1.0, we can adjust the upgrade logic accordingly.
 
> The documentation of [iadmin.cpp](https://github.com/irods/irods_client_icommands/blob/main/src/iadmin.cpp) should also be adjusted, but I guess that is just cosmetic.

Correct, this will be part of resolving the to-be-created issue. It is more than cosmetic in that it reduces the amount of potential email for people like me :)

peterverraedt:
I've created the issue #8789 I'm trying to solve.

> Also, our semantic versioning directive requires that the default value for the new option reflects the existing behavior so that it can be included in a minor release. Can you confirm that the system behavior is unchanged with the default configuration? It sounds like the default configuration may change behavior for systems with password_extend_lifetime set to 0. I might be misunderstanding something.

In case password_reuse_previous is set to true, which should be the default value, nothing changes, and the value password_extend_lifetime dictates as before whether passwords should be extended or not.

peterverraedt:
I've added a test as well, but I'm unsure where to look whether it is executed and/or successful.

alanking:
Here are the instructions on running tests: https://docs.irods.org/5.0.2/developers/testing/#how-to-run-the-tests

I can see in the tests that they probably won't do what we're expecting just based on the variable names. If you don't mind seeing if you can get things working on your own, that would be helpful. Unfortunately, as you know, this is not quite automated yet.

If you don't have the time, the iRODS team can take a look at implementing tests. I think the description(s) of the issue you've provided gives enough context to create several tests in addition to the expected boundary-value tests. Passing tests are a requirement before merging either way, so just let us know.

peterverraedt:
I've run the tests and can confirm my test works fine.

There are already 2 checks that check the behaviour when password_reuse_previous is set to 1, what is the default value: `test_password_extend_lifetime_set_to_false_invalidates_other_authentications_on_expiration`, `test_password_extend_lifetime_set_to_true_extends_other_authentications_past_expiration`. 
The added test checks the behavoiur when password_reuse_previous is set to 0.

If you need more tests added, feel free to suggest some scenarios.

scripts/irods/test/test_pam_password_authentication.py
```diff
@@ -506,6 +506,90 @@ def test_password_extend_lifetime_set_to_false_invalidates_other_authentications
...
+            ttl = 4
+            self.assertGreater(ttl, 3)
+            option_value = str(ttl)
+            self.admin.assert_icommand(
+                ['iadmin', 'set_grid_configuration', self.configuration_namespace, min_time_option_name, option_value])
+
+            # Authenticate with session 1
+            self.auth_session.assert_icommand(
+                ['iinit'], 'STDOUT', 'PAM password', input=f'{self.auth_session.password}\n')
+            self.auth_session.assert_icommand(["ils"], 'STDOUT', self.auth_session.session_collection)
```
alanking: Should we also authenticate with session 2 to ensure that the passwords of both sessions are "valid" for (roughly) the same amount of time? I understand that session 2 was authenticated above, but it's possible that it has been a couple of seconds since then. This concern may be invalid...

alanking (commented):
Read the test again and it looks good. Thinking on it some more and getting a better understanding of the feature, we don't need a test for `password_extend_lifetime` == 0 because that configuration only has an effect when passwords are being reused across sessions.

I had one lingering question/concern, but I think this is basically ready.

scripts/irods/test/test_pam_password_authentication.py
peterverraedt: I've added a check to see whether the pam authentication of session 2 is valid for around the same time as session 1 (both should be valid for 'ttl'). I also updated the comments a bit.

alanking: Looks good. Thanks

peterverraedt:
Squashed commits.

PR #8790, opened 2026-01-07
[#8733] Replace spaces with hyphens for agent info (main)

korydraughn:
Port of #8784.

### DIFF
--- a/server/main_server/src/agent_main.cpp
+++ b/server/main_server/src/agent_main.cpp
@@ -52,6 +52,7 @@
 #include <sys/wait.h>
 #include <unistd.h>
 
+#include <algorithm>
 #include <array>
 #include <charconv>
 #include <chrono>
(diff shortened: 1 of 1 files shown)


alanking:
Implementation appears to have changed slightly from the 4-3-stable version. What's the reason for that? I assume that's why this is in draft?

korydraughn:
The two branches have different implementations for creating the agent info files.

I put it in draft since I haven't run the test yet.

alanking:
> The two branches have different implementations for creating the agent info files.

Why are they different?

korydraughn:
The implementation for creating the agent files changed in the iRODS 5 refactor.

alanking:
Gotcha, thanks.

scripts/irods/test/test_ips.py
```diff
@@ -0,0 +1,51 @@
+import os
```
korydraughn: Remove unused import.

Will make the same change to 4-3-stable once this PR lands.

korydraughn: Done.

korydraughn:
The test passed as expected.

This is ready. Taking out of draft.

server/main_server/src/agent_main.cpp
```diff
@@ -1134,14 +1135,19 @@ namespace
         const auto proc_dir = irods::get_irods_proc_directory();
         const auto pid_file = fmt::format("{}/{}", proc_dir.c_str(), getpid());
 
+        const auto replace_spaces = [](std::string _s) -> std::string {
+            std::replace(std::begin(_s), std::end(_s), ' ', '-');
+            return _s;
+        };
```
alanking, line 1141: Is some level of copy-elision happening here?

For `client_zone`, for instance, we copy it into a new temporary string, then a copy is passed to this, and it returns that string. I guess I'm instinctually wondering whether that argument ought to be a reference?

alanking (approved):
Just had one question. I think it's fine as is so I'm approving it. # it if question does not raise a valid concern

server/main_server/src/agent_main.cpp
korydraughn: It can't be a reference because some of the arguments are `string_view`s. You have to convert string_view to std::string, else you get a build error. I think C++26 fixes that.

The other reason it can't be a reference is due to the incoming RcComm being const. That made some of the members we use read-only.

alanking: Got it. Thanks

korydraughn:
Added pounds.

Issue #8671, opened 2025-09-15
iadmin does not return error code for get/set_grid_configuration


stsnel:
- [x] main
- [x] 4-3-stable

---

### iRODS Version, OS and Version

iRODS 4.3.4 and 5.0.1
Ubuntu 24.04 LTS

### What did you try to do?

Get or set a grid configuration variable using `iadmin` with wrong parameters.

### Expected behavior

If a failure occurs when getting or setting grid configuration variables, `iadmin` returns a non-zero return value.

### Observed Behavior (including steps to reproduce, if applicable)

`iadmin` seems to always have a return value of 0 with these subcommands, irrespective of whether the operation succeeded. For example: 

```
irods@provider:~$ iadmin get_grid_configuration doesnotexist doesnotexist
Failed to get grid configuration for namespace [doesnotexist] and option [doesnotexist] [ec=-808000]
irods@provider:~$ echo $?
0
```

```
irods@provider:~$ iadmin set_grid_configuration doesnotexist doesnotexist something
Failed to set grid configuration for namespace [doesnotexist] and option [doesnotexist] [ec=-808000]
irods@provider:~$ echo $?
0
```

PR #8792, opened 2026-01-08
[#8671] Update iadmin grid config tests for error codes (4-3-stable)

korydraughn:
Depends on irods/irods_client_icommands#598.

The test passes in isolation.

Running core tests to see if any other tests are affected.

### DIFF
--- a/scripts/irods/test/test_iadmin_set_grid_configuration.py
+++ b/scripts/irods/test/test_iadmin_set_grid_configuration.py
@@ -1,5 +1,3 @@
-from __future__ import print_function
-
 import unittest
 
 from . import session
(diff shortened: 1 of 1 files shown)


korydraughn:
Tests which only care about showing help text failed with a segfault. The segfaults were triggered on `iget -h`. The reason why iget segfaults is because one of the variables/structs wasn't cleared before use.

I'll open an issue for this and push the changes up soon.

korydraughn:
Forgot to mention that all other core tests passed. The fix does not require rerunning the core test suite since it only clears a struct on program start.

See the new commit in the icommands PR for the fix.

korydraughn:
The tests that were failing now pass.

korydraughn:
Added pound.

Issue #8413, opened 2025-04-22
`icp -f` doesnt modify the timestamp of the target object


mstfdkmn:
- [x] main
- [x] 4-3-stable

---

### iRODS Version, OS and Version
4.3.4
### What did you try to do?
Attempted to copy a new data object to a collection containing an object with the same name.

### Expected behavior
I expected to see the new object (with new content) would be copied, and its modification time would reflect the copy time.

### Observed Behavior (including steps to reproduce, if applicable)
The new object was copied, and its content is updated, but the modification time remains that of the original object.

```
root@aa959666fe8d:/# echo "12" | istream write test_icp/test.txt
root@aa959666fe8d:/# ils -L test_icp/test.txt
  u0137480          0 default;netapp            3 2025-04-22.12:39 & test.txt
        generic    /netapp/home/u0137480/test_icp/test.txt
root@aa959666fe8d:/# itouch test.txt
root@aa959666fe8d:/# ils -L test.txt
  u0137480          0 default;netapp            0 2025-04-22.12:40 & test.txt
        generic    /netapp/home/u0137480/test.txt
root@aa959666fe8d:/# icp test.txt test_icp
remote addresses: 10.114.48.180 ERROR: cpUtil: cp error for /icts/home/u0137480/test_icp/test.txt, status = -312000 status = -312000 OVERWRITE_WITHOUT_FORCE_FLAG
root@aa959666fe8d:/# icp -f test.txt test_icp
root@aa959666fe8d:/# ils -L test_icp/test.txt
  u0137480          0 default;netapp            0 2025-04-22.12:39 & test.txt
        generic    /netapp/home/u0137480/test_icp/test.txt
```

### iRODS Server Logs (if applicable)

alanking:
I thought that this had something to do with the object names being identical, or copying to a subcollection or something like that. The reason the timestamp does not update is because the source replica size is 0, so this is interpreted as no bytes having been written. If no bytes are written, the modify time does not get updated even if the data object is modified.

I tried this scenario with replicas that have data and everything worked as expected:
```bash
$ ils -Lr
/tempZone/home/rods:
  rods              0 demoResc          241 2026-01-08.19:10 & foo
        generic    /var/lib/irods/Vault/home/rods/foo
  C- /tempZone/home/rods/test_icp
/tempZone/home/rods/test_icp:
  rods              0 demoResc            0 2026-01-08.19:10 & foo
        generic    /var/lib/irods/Vault/home/rods/test_icp/foo
[...]

Issue #7128, opened 2023-06-22
Data object mtime of zero-length files not updating after ingest and istream


avkrishnamurthy:
- [ ] main
- [ ] 4-3-stable
---
### iRODS Version, OS and Version

- 4.3.0
- Ubuntu 20.04

### What did you try to do?

Created files for ingest in landing zone. Ingested files with PUT_SYNC to iRODS collections. Created more files with same name in landing zone. Tried to ingest again from landing zone with PUT_SYNC. 

### Expected behavior

After ingesting the second time with PUT_SYNC, would expect the mtime of the files in the iRODS collection to be updated and be the same as the mtime as the files in the iRODS Vault, but the mtime of files in iRODS collection is not updated.

### Observed behavior (including steps to reproduce, if applicable)

After running the ingest the second time, the files in the iRODS collection do not have updated mtime. The files in the iRODS Vault do have an updated mtime.

If the size of the file(s) is changed, then the file in the iRODS collection updates properly. However, if the size does not change it does not update.

Same thing occurs with istream as well.

```bash
$ istream write goo
$ ils -L goo
  rods              0 demoResc            0 2023-06-22.18:13 & goo 
        generic    /var/lib/irods/Vault/home/rods/goo
$ istream write goo
$ ils -L goo
  rods              0 demoResc            0 2023-06-22.18:13 & goo 
        generic    /var/lib/irods/Vault/home/rods/goo
$ ls -l /var/lib/irods/Vault/home/rods/goo
-rw------- 1 irods irods 0 Jun 22 18:14 /var/lib/irods/Vault/home/rods/goo
```

trel:
So... this lack of updating the mtime is only when the content changes, but with identical filesize and filename?

Your example here shows zero-length files.

Can you try with a file with contents that changes from 'apple' to 'grape', both size 5 (or 6, with an included newline)?

Without a checksum to compare against, we only have filename and filesize, so I'm not sure this is a bug without new bytes being written?  Definitely a corner case with the zero-length files...

avkrishnamurthy:
I just tried it with changing the file contents from "apple" to "grape", and it updated the file in the collection and in the Vault as expected. I also made a new file with the same name and same content with just "apple" and tried that as well, and that updated the file in the collection and Vault also, so it seems like the unexpected behavior only happens with zero-length files, so it is probably not really a bug.

trel:
ah, even better.   okay - so if the file changes, regardless of size, it works as expected.

[...]

PR #8794, opened 2026-01-09
[#8413] Update mtime on empty overwrites for Copy (main)

alanking:
Addresses #8413

### DIFF
--- a/server/api/src/rsDataObjClose.cpp
+++ b/server/api/src/rsDataObjClose.cpp
@@ -361,7 +361,7 @@ namespace
             r.mtime(SET_TIME_TO_NOW_KW);
         }
         else if (OPEN_FOR_WRITE_TYPE == l1desc.openType) {
-            if (l1desc.bytesWritten > 0) {
+            if (bytes_written_in_operation(l1desc)) {
                 // If no errors occurred, bytes being written to an existing replica should
                 // result in staling any sibling replicas. This is because it is not known
                 // whether the new replica is consistent with the siblings.


scripts/irods/test/test_icp.py
```diff
@@ -402,3 +398,50 @@ def test_overwrite_fails_when_force_flag_is_not_specified(self):
...
+        try:
+            # Create the test data.
+            self.user.assert_icommand(["itouch", logical_path_src], "STDOUT")
+            self.user.assert_icommand(["istream", "write", logical_path_dest], "STDOUT", input=content)
+
+            # Get the original modify times for the test data.
+            original_src_mtime = lib.get_replica_mtime(self.user, logical_path_src, 0)
+            original_dest_mtime = lib.get_replica_mtime(self.user, logical_path_dest, 0)
+
+            self.user.assert_icommand(["ils", "-Lr"], "STDOUT") # debugging
```
korydraughn, line 418: So the joining of two flags works. Nice. I can never remember if our testing supports that.

_I'm not suggesting changes here. Just highlighting the flag usage as a single string._

scripts/irods/test/test_icp.py
```diff
@@ -402,3 +398,50 @@ def test_overwrite_fails_when_force_flag_is_not_specified(self):
...
+
+            # Get the new modify times for the test data.
+            new_src_mtime = lib.get_replica_mtime(self.user, logical_path_src, 0)
+            new_dest_mtime = lib.get_replica_mtime(self.user, logical_path_dest, 0)
+
+            # The modify time of the destination object should have been changed to a later time. For this test, we only
+            # want to specify the collection of the destination object, which will cause the destination object to have
+            # the same name as the source object.
+            self.assertEqual(original_src_mtime, new_src_mtime)
+            self.assertLess(original_dest_mtime, new_dest_mtime)
```
korydraughn, line 432: > ... For this test, we only want to specify the collection of the destination object, which will cause the destination object to have the same name as the source object.

It's not clear to me what this means. What am I missing here?

server/api/src/rsDataObjClose.cpp
```diff
@@ -361,7 +361,7 @@ namespace
             r.mtime(SET_TIME_TO_NOW_KW);
         }
         else if (OPEN_FOR_WRITE_TYPE == l1desc.openType) {
-            if (l1desc.bytesWritten > 0) {
+            if (bytes_written_in_operation(l1desc)) {
```
korydraughn, line 364: Does `rs_replica_close` already cover what you're fixing for `rsDataObjClose`?

scripts/irods/test/test_icp.py
alanking: Oh lol That's leftover from an earlier iteration of the test when I thought it had something to do with object names and collection names (see https://github.com/irods/irods/issues/8413#issuecomment-3725303550). I will update that.

server/api/src/rsDataObjClose.cpp
alanking: No. We will need to discuss that one because it would cause some behavioral changes and has more to do with #7128.

scripts/irods/test/test_icp.py
alanking: Done

korydraughn:
Squash it if ready.

alanking:
Squashed

alanking:
#'d, mergin

PR #8795, opened 2026-01-09
[8750] Quotas now only consider ACL "own" permissions for counting (4-3-stable)

FifthPotato:
Partial cherry-pick of https://github.com/irods/irods/pull/8778
Instead of modifying the new query that does the math for coordinating resources, this one instead modifies the query before that (since we aren't adding coordinating resource quotas to 4.3). Test is identical, and `test_quotas` passes in isolation.


korydraughn:
Please run all of the unit/core tests just so there's no surprises during the release testing period.

alanking (commented):
Eyeball test passes here

FifthPotato:
Rebased this and the icommands onto 4-3-stable and rebuilt with debug; core and unit tests pass.

FifthPotato:
Octothorpe'd

PR #8797, opened 2026-01-09
[#8413] Update mtime on empty overwrites for Copy (4-3-stable)

alanking:
Addresses #8413

Cherry-picked from https://github.com/irods/irods/pull/8794

Only ran new test, which passes.


alanking:
#'d, mergin

PR #8805, opened 2026-01-15
[4102] Add test to ensure GenQuery and quota math agree (main)

FifthPotato:
Just a test. The test passes (and so does the rest of `test_quotas`).
Taking suggestions on substance or style, as always

### DIFF
--- a/scripts/irods/test/test_quotas.py
+++ b/scripts/irods/test/test_quotas.py
@@ -354,3 +354,78 @@ class Test_Quotas(resource_suite.ResourceBase, unittest.TestCase):
             self.admin.run_icommand(['iadmin', 'rmuser', user_name])
             self.admin.run_icommand(['iadmin', 'rmgroup', group_name])
             IrodsController().reload_configuration()
+
+
+    def test_quota_calculations_match_genquery_calculations__issue_4102(self):
+        try:
+            with temporary_core_file() as core:
+                core.add_rule(self.pep_to_enable_resc_quota_policy[self.plugin_name])
+                IrodsController().reload_configuration()
+
+                group_name = 'test_quota_genquery'
+                resc_name = 'ufs0_quota_genquery'
+                file_prefix = 'quota_genquery_match'
+                file_count = 10
+                quota_setting = 2048*(file_count+1)+1
+
+                # Capturing groups will capture decimal integers
+                # The only valid integer with a preceding 0 is 0 itself
+                # Otherwise, it must be 1-9 or 1-9 prefixing any number of 0-9
+                quota_re = re.compile(r'quota_limit: (0|-?[1-9][0-9]*)|quota_over: (0|-?[1-9][0-9]*)')
+                genquery_re = re.compile(r'DATA_SIZE = (0|-?[1-9][0-9]*)')
+
+                def file_generator(n, filename_prefix):
+                    seed = 1031
+                    for i in range(n):
+                        seed = (5*seed + 131) % 2048
+                        yield (f'{filename_prefix}_{i}', seed)
+
+                files = list(file_generator(file_count, file_prefix))
+                lib.create_ufs_resource(self.admin, resc_name)
+                for file in files:
+                    lib.make_file(os.path.join(self.quota_user.local_session_dir, file[0]), file[1], contents='arbitrary')
+                    self.quota_user.assert_icommand(['iput', '-R', resc_name, os.path.join(self.quota_user.local_session_dir, file[0])])
+
+                self.admin.assert_icommand(['iadmin', 'mkgroup', group_name])
+                self.admin.assert_icommand(['iadmin', 'sgq', group_name, resc_name, str(quota_setting)])
+                self.admin.assert_icommand(['iadmin', 'atg', group_name, self.quota_user.username])
+
+                self.admin.assert_icommand(['iadmin', 'cu'])
+
+                total_size = sum(map(lambda file: file[1], files))
+
+                # quota_over should be the total size of the files minus the specified quota amount
+                _, quota_out, _ = self.admin.assert_icommand(['iadmin', 'lq'], 'STDOUT', [f'\nquota_over: {total_size - quota_setting}\n'])
+
+                # Sum of all DATA_SIZE from iquest should match total
+                _, genquery_out, _ = self.admin.assert_icommand(['iquest', f'select sum(DATA_SIZE) where COLL_NAME like \'{self.quota_user.session_collection}%\''], 'STDOUT', [f'DATA_SIZE = {total_size}\n'])
+
+                # Linking the outputs of the two commands:
+                # First, parse out each relevant value
+                quota_match = quota_re.findall(quota_out)
+                quota_limit = int(quota_match[0][0])
+                quota_over = int(quota_match[1][1])
+                sum_data_size = int(genquery_re.search(genquery_out).group(1))
+
+                # Next, assert the math is correct
+                self.assertEqual(quota_over, sum_data_size - quota_limit)
+
+                lib.make_file(os.path.join(self.quota_user.local_session_dir, 'quota_genquery_bonus_file'), 2048, contents='arbitrary')
+                self.quota_user.assert_icommand(['iput', '-R', resc_name, os.path.join(self.quota_user.local_session_dir, 'quota_genquery_bonus_file')])
+                self.admin.assert_icommand(['iadmin', 'cu'])
+
+                _, quota_out, _ = self.admin.assert_icommand(['iadmin', 'lq'], 'STDOUT', [f'\nquota_over: {total_size + 2048 - quota_setting}\n'])
+                _, genquery_out, _ = self.admin.assert_icommand(['iquest', f'select sum(DATA_SIZE) where COLL_NAME like \'{self.quota_user.session_collection}%\''], 'STDOUT', [f'DATA_SIZE = {total_size + 2048}\n'])
+
+                # Same as above
+                quota_match = quota_re.findall(quota_out)
+                quota_limit = int(quota_match[0][0])
+                quota_over = int(quota_match[1][1])
+                sum_data_size = int(genquery_re.search(genquery_out).group(1))
+
+                self.assertEqual(quota_over, sum_data_size - quota_limit)
+        finally:
+            self.quota_user.run_icommand(['irm', '-f', 'quota_genquery_bonus_file', *map(lambda file: file[0], files)])
+            self.admin.run_icommand(['iadmin', 'rmresc', resc_name])
+            self.admin.run_icommand(['iadmin', 'rmgroup', group_name])
+            IrodsController().reload_configuration()


scripts/irods/test/test_quotas.py
```diff
@@ -414,3 +415,61 @@ def acRescQuotaPolicy(rule_args, callback, rei):
             self.admin.run_icommand(['iadmin', 'rmuser', user_name])
             self.admin.run_icommand(['iadmin', 'rmgroup', group_name])
             IrodsController().reload_configuration()
+
+
+    def test_quota_calculations_match_genquery_calculations__issue_8750(self):
```
korydraughn: Wrong issue number?

scripts/irods/test/test_quotas.py
```diff
@@ -414,3 +415,61 @@ def acRescQuotaPolicy(rule_args, callback, rei):
...
+                core.add_rule(pep_map[self.plugin_name])
+                IrodsController().reload_configuration()
+
+                group_name = 'test_quota_genquery'
+                resc_name = 'ufs0_quota_genquery'
+                file_prefix = 'quota_genquery_match'
+                file_count = 10
+                quota_setting = 2048*(file_count+1)+1
+
+                files = [ (f'{file_prefix}_{i}', random.randint(1024, 2048)) for i in range(0, file_count) ]
```
alanking: A few things...

1. I don't like random in the integration tests. Even if the values are hard-coded and chosen at random (or generated), at least it will be deterministic. Please tell me why I'm wrong.
2. Why the range 1024 - 2048?
3. These files should be cleaned up in the `finally` block, right? Or `file_prefix` should include `self.quota_user.local_session_dir` so that they are removed when the session gets cleaned up.

korydraughn: Is the 0 in the call to `range` needed? Couldn't it be rewritten as such?


```suggestion
                files = [ (f'{file_prefix}_{i}', random.randint(1024, 2048)) for i in range(file_count) ]
```

scripts/irods/test/test_quotas.py
```diff
@@ -414,3 +415,61 @@ def acRescQuotaPolicy(rule_args, callback, rei):
...
+            'irods_rule_engine_plugin-irods_rule_language': textwrap.dedent('''
+                acRescQuotaPolicy {
+                    msiSetRescQuotaPolicy("on");
+                }
+            '''),
+            'irods_rule_engine_plugin-python': textwrap.dedent('''
+                def acRescQuotaPolicy(rule_args, callback, rei):
+                    callback.msiSetRescQuotaPolicy('on')
+            ''')
+        }
```
alanking: Since this PR just has to do with tests, I'll go ahead and leave this suggestion which I've been thinking about on the past few PRs...

I've seen this same stanza in many tests in this file. Maybe we should consolidate all of them into a helper function or even just a member of the test class that all the tests can access? The name would need to change to something more descriptive (e.g. `pep_to_enable_resc_quota_policy`) but could lead to some minus.

scripts/irods/test/test_quotas.py
```diff
@@ -414,3 +415,61 @@ def acRescQuotaPolicy(rule_args, callback, rei):
...
+                self.admin.assert_icommand(['iquest', f'select sum(DATA_SIZE) where COLL_NAME like \'{self.quota_user.session_collection}%\''], 'STDOUT', [f'DATA_SIZE = {total_size}\n'])
+
+                lib.make_file(os.path.join(self.quota_user.local_session_dir, 'quota_genquery_bonus_file'), 2048, contents='arbitrary')
+                self.quota_user.assert_icommand(['iput', '-R', resc_name, os.path.join(self.quota_user.local_session_dir, 'quota_genquery_bonus_file')])
+                self.admin.assert_icommand(['iadmin', 'cu'])
+
+                self.admin.assert_icommand(['iadmin', 'lq'], 'STDOUT', [f'\nquota_over: {total_size + 2048 - quota_setting}\n'])
+                self.admin.assert_icommand(['iquest', f'select sum(DATA_SIZE) where COLL_NAME like \'{self.quota_user.session_collection}%\''], 'STDOUT', [f'DATA_SIZE = {total_size + 2048}\n'])
+        finally:
+            self.quota_user.run_icommand(['irm', '-f', 'quota_genquery_bonus_file', *map(lambda file: file[0], files)])
```
alanking, line 428: I'll have to keep this trick in my back pocket. Pretty cool

scripts/irods/test/test_quotas.py
FifthPotato: 1. That's fair. I wanted to make the file sizes random to try and show that the "math" part always works, and not just with specific values. I think written correctly it should essentially test the same thing, but maybe it's a little more sensitive if we break the math? I can definitely just set it to hardcoded values if you think that's better, though.
2. Arbitrary. I like my powers of 2, I guess
3. I think I do that with my sneaky unpack at the bottom?

scripts/irods/test/test_quotas.py
FifthPotato: Did this. Seems to work!

scripts/irods/test/test_quotas.py
FifthPotato: Nope, you're right. Just my weird Python muscle memory. Changed!

scripts/irods/test/test_quotas.py
FifthPotato: Oops. Copied from above. Changed.

scripts/irods/test/test_quotas.py
alanking: 1. I suppose that's fair. I just wonder what the difference is between picking 10 values arbitrarily and generating 10 random values. For this test, it's not too different (IMO) so it's innocuous. I just wanted to open up the discussion. Any thoughts on this, @korydraughn?
2. Okay. Maybe leave a comment so that archaeologists will know that there's nothing special about that value range.
3. The sneaky unpack removes the uploaded data objects, but not the local files.

FifthPotato: 1. I think it essentially makes no difference in basically every case. Maybe I was just worried in a weird case where certain math errors could be covered up by the test not being stringent/sensitive enough. But I don't mind it being written either way.
2. Will do!
3. Oh, the file creation on line 378 actually does the join, it's just not in the prefix. Would that be enough?

korydraughn: I think the random values make the test stronger. Like @alanking, I try to avoid random values, but the reasoning behind using them here makes sense to me.

The upside of using random values is that they could help reveal bugs if any are present in the implementation. The downside is reproduction of a bug (found through this test) could be difficult unless you always capture the logs.

With that said, it's not clear to me that the range used could aid in detecting issues. Ultimately, it's hard to say which way to go. I think we pick a direction and keep moving forward. If there's a bug, people will report it (hopefully).

---

This conversation has led me to wonder if we have tests which exercise the bounds of the quota values. What happens if the quota values cannot be represented in the target data types?

scripts/irods/test/test_quotas.py
```diff
@@ -24,16 +25,7 @@ class Test_Quotas(resource_suite.ResourceBase, unittest.TestCase):
...
-
-    def tearDown(self):
-        self.quota_user.__exit__()
-        self.admin.assert_icommand(['iadmin', 'rmuser', 'quotauser'])
-        super(Test_Quotas, self).tearDown()
-
-
-    def test_iquota__3044(self):
-        pep_map = {
+        self.pep_to_enable_resc_quota_policy = {
```
korydraughn, line 27: Let's make this change into a separate commit. Perhaps before your addition of the test.

scripts/irods/test/test_quotas.py
alanking: > Oh, the file creation on line 378 actually does the join, it's just not in the prefix. Would that be enough?

Yes, that looks like what we want here. Let's do that.

---

I defer on the use of the random values here as they do not add (IMO) nor harm anything in this case.

FifthPotato: Since we mostly seem to agree that it doesn't seem to actually add anything, I'll refactor it to avoid the randomness, since non-deterministic tests do seem pretty bad for trying to reproduce issues.
In exchange, I guess I'll pick 10 "ugly" numbers so they don't add up nicely, so it's more likely to ensure the math stays correct.

scripts/irods/test/test_quotas.py
FifthPotato: Did some git-fu, and I think I achieved it!

FifthPotato:
> Looks good.
> 
> I noticed the test does not take the values produced by iquest or iquota and compares them.
> 
> Seems having an assertion for that would tie the two tools together. For example:
> 
>     1. Use iquest to produce a number
> 
>     2. Take iquest result and use it in iquota assertion
> 
> 
> Does that make sense?

Force-pushed with a bunch more changes, including this one. Also reordered the commits to put the pep_map changes first.

scripts/irods/test/test_quotas.py
FifthPotato: The numbers are still ugly, but it's deterministic now. Do we like this better?

scripts/irods/test/test_quotas.py
```diff
@@ -357,32 +356,27 @@ def test_quota_cares_about_acl_own_permission__issue_8750(self):
...
+                core.add_rule(self.pep_to_enable_resc_quota_policy[self.plugin_name])
                 IrodsController().reload_configuration()
 
                 group_name = 'test_quota_genquery'
                 resc_name = 'ufs0_quota_genquery'
                 file_prefix = 'quota_genquery_match'
                 file_count = 10
                 quota_setting = 2048*(file_count+1)+1
+                quota_re = re.compile(r'quota_limit: ([0-9]*)|quota_over: (-?[0-9]*)')
+                genquery_re = re.compile(r'DATA_SIZE = ([0-9]*)')
```
korydraughn: Your regexes will match invalid inputs for the numeric parts.

Please constrain them to only accept correct integers.

scripts/irods/test/test_quotas.py
korydraughn: Nice.

scripts/irods/test/test_quotas.py
alanking: Seems fine. It seems that it's generating the same number for each file, unless I'm reading it wrong. Do we *want* the files to all be the same size? This is not meant to be a leading question.

scripts/irods/test/test_quotas.py
FifthPotato: I think this does what we want now. Do you agree?

scripts/irods/test/test_quotas.py
```diff
@@ -367,8 +367,8 @@ def test_quota_calculations_match_genquery_calculations__issue_4102(self):
                 file_prefix = 'quota_genquery_match'
                 file_count = 10
                 quota_setting = 2048*(file_count+1)+1
-                quota_re = re.compile(r'quota_limit: ([0-9]*)|quota_over: (-?[0-9]*)')
-                genquery_re = re.compile(r'DATA_SIZE = ([0-9]*)')
+                quota_re = re.compile(r'quota_limit: (0|-?[1-9][0-9]*)|quota_over: (0|-?[1-9][0-9]*)')
```
trel, line 374: let's definitely leave a note with the *intent* of the regex.   in case someone later finds a flaw / discrepancy.

scripts/irods/test/test_quotas.py
FifthPotato: Nope, they're all different, promise!
If we're okay with it, then I'll resolve this.

scripts/irods/test/test_quotas.py
korydraughn: It looks correct to me, but please verify it works as intended.

scripts/irods/test/test_quotas.py
alanking: Yep, resolve.

scripts/irods/test/test_quotas.py
FifthPotato: Done.

scripts/irods/test/test_quotas.py
```diff
@@ -367,6 +367,10 @@ def test_quota_calculations_match_genquery_calculations__issue_4102(self):
                 file_prefix = 'quota_genquery_match'
                 file_count = 10
                 quota_setting = 2048*(file_count+1)+1
+
+                # Capturing groups will capture decimal integers
```
alanking: Codacy found trailing whitespace

alanking (commented):
Looks good to me. One last thing from me

scripts/irods/test/test_quotas.py
FifthPotato: Done

korydraughn:
Does this one test suite still pass?

FifthPotato:
> Does this one test suite still pass?

Sure does!

korydraughn:
Squash it.

FifthPotato:
Squished. Left out the pep_map change as a separate commit.

korydraughn:
Yep. You can attach that commit to issue 858. Or, you can open a new issue specifically for it. Up to you.

FifthPotato:
Oops, misclicked the close.
Attached to 858.

korydraughn (approved):
Looks good to me.

Please wait for a second approval before pounding.

alanking (approved):
\# it

FifthPotato:
Octothorpe'd.

alanking:
@FifthPotato - Please cherry-pick / adapt to 4-3-stable. Thanks

Issue #5763, opened 2021-08-05
rxUnregDataObj bypasses logical locking


alanking:
- [ ] master
- [ ] 4-2-stable
---

`rsUnregDataObj` does not check the replica status(es) of the replica or data object unregistered and does not honor logical locking. This is currently being used in `rsDataObjOpen` to unlink newly created replicas which are still in the intermediate status when something goes wrong during the creation of the physical file. This use is fine, but more broadly speaking, `rsUnregDataObj` should honor logical locking just like everything else.

alanking:
The title of this issue is a little bit misleading.

`iunreg` does not use the UnregDataObj API. It uses DataObjTrim and DataObjUnlink.

`ireg` does not use the RegDataObj API. It uses PhyPathReg and RegReplica.

That being said, we should probably test all of these APIs. So, here is a list of the APIs we need to make sure are covered:
 - RegDataObj
 - UnregDataObj
 - PhyPathReg
 - DataObjTrim
 - DataObjUnlink
 - RegReplica

It's a lot easier to write tests involving intermediate replicas by calling the APIs directly rather than using the iCommands, so I'm going to try to avoid that if I can. If all the APIs are made to honor logical locking, then the iCommands which use them will fall in line as well.

alanking:
Looks like we already have tests for unregistering via DataObjUnlink: https://github.com/irods/irods/blob/3ba9fb9b7c734b87ee4f15ee45222882ba794e54/unit_tests/src/test_logical_locking.cpp#L390-L399
We have tests for DataObjTrim, but not with unregister: https://github.com/irods/irods/blob/3ba9fb9b7c734b87ee4f15ee45222882ba794e54/unit_tests/src/test_logical_locking.cpp#L424

Issue #8765, opened 2025-12-05
Logical locking shoud include agent information


peterverraedt:
### New Feature / Functionality

In our irods installation we regularly see HIERARCHY_ERRORS pop up, and they need a lot of manual labour to follow up and repair. The reason is mostly a bad behaving client or a crashing irods agent. Our main concern is to identify easily which data objects are really "locked": some data objects are currently being written to by a client (there is an existing agent holding the write lock), some data objects are left with data_status 2 (because the agent holding the lock crashed for various reasons).

I would suggest the following: 
* if a data object gets a write lock, information about the agent holding the lock will be stored in the catalog. This can be e.g. the irods server name + pid combination.
* if the server detects that an agent crashes (e.g. because of network errors), it checks whether it was holding a write lock, and do some remediation (putting the file to stale, for example, or another state that allows users to reupload the file).

alanking:
Thanks for the suggestion. This is something we had considered when logical locking was first introduced, but backed off the idea since objects "should" never get stuck in this state (if they do, it's a bug, and the bug should be fixed). That being said, the reality of the situation is that they do, and so this would be very helpful even for manual resolution of locked objects, as you've said.

This is also a possible prerequisite task for addressing #5763, so we will look into this. I'm putting it in the 5.1.0 milestone for now since the scope of the work is unknown right now.

korydraughn:
After some discussion, we think that the following could work as a workaround.

Use the resource PEPs (e.g. `pep_resource_open_*`) to write the agent information to the server log.

We'll look into providing an implementation which demonstrates this.

peterverraedt:
There is surely a workaround possible. E.g. we have audit logs, they register when files are opened for writing and contain the pid. We can also regularly query the irods server for currently running agents and cross check the information. We also can query the catalog for objects holding a write lock. By analysing three sources of information (irods logs, irods catalog, current irods processes) we can try to find objects for which the agent crashed "in the past" and do some resolution (marking them as stale, mailing the user, ...).

It would be even better if fewer objects get locked in the first place. E.g. if the network connection is closed prematurely, the agent should know that any opened objects should be released. The proposed method is only one of the possibilities that would make this possible. Any effort that would let the irods server self-heal objects for which it is sure the lock holder died, would be an enhancement. Another possible solution is a lock lease time.

alanking:
> If the network connection is closed prematurely, the agent should know that any opened objects should be released. The proposed method is only one of the possibilities that would make this possible.

[...]

PR #8806, opened 2026-01-20
[#5763] Honor logical locking in various reg/unreg APIs (4-3-stable)

alanking:
Addresses #5763 

I tried to get a bare minimum set of tests written for this, but these APIs are currently not being (directly) tested at all, as far as I can tell.

Unit tests passed. Core tests running.

### DIFF
--- a/server/api/src/rsDataObjOpen.cpp
+++ b/server/api/src/rsDataObjOpen.cpp
@@ -347,7 +347,24 @@ namespace
                 info_json_str = nullptr;
             }};
 
+            // This keyword indicates to the register_physical_path API that logical locking should not be enforced.
+            // Logical locking has already been enforced above, so at this point, we know that logical locking is not
+            // being violated. In order to ensure it is being used in an authorized manner, register_physical_path only
+            // allows use of this keyword for server-to-server connections. If this is not already a server-to-server
+            // connection, the server property is set so that the keyword will be recognized.
+            static_cast<void>(addKeyVal(&l1desc.dataObjInp->condInput, ill::keywords::bypass, ""));
+            bool agent_conn_kw_already_set = irods::server_property_exists(irods::AGENT_CONN_KW);
+            if (!agent_conn_kw_already_set) {
+                static_cast<void>(irods::set_server_property<std::string>(irods::AGENT_CONN_KW, ill::keywords::bypass));
+            }
+
             if (const int ec = rs_register_physical_path(&_comm, l1desc.dataObjInp, &info_json_str); ec < 0 || !info_json_str) {
+                // Remove the logical locking bypass properties as soon as possible.
+                static_cast<void>(rmKeyVal(&l1desc.dataObjInp->condInput, ill::keywords::bypass));
+                if (!agent_conn_kw_already_set) {
+                    static_cast<void>(irods::delete_server_property(irods::AGENT_CONN_KW));
+                }
+
                 irods::log(LOG_ERROR, fmt::format(
                     "[{}:{}] - failed to register physical path "
                     "[error_code=[{}], path=[{}], hierarchy=[{}]",
@@ -369,6 +386,12 @@ namespace
                 freeL1desc(l1_index);
 
                 return ec;
+            }
+
+            // Remove the logical locking bypass properties as soon as possible.
+            static_cast<void>(rmKeyVal(&l1desc.dataObjInp->condInput, ill::keywords::bypass));
+            if (!agent_conn_kw_already_set) {
+                static_cast<void>(irods::delete_server_property(irods::AGENT_CONN_KW));
             }
 
             // Extract the JSON output from the registration API. If the output is not an expected value, a JSON
--- a/server/api/src/rsRegReplica.cpp
+++ b/server/api/src/rsRegReplica.cpp
@@ -4,6 +4,8 @@
 #include "irods/irods_configuration_keywords.hpp"
 #include "irods/irods_file_object.hpp"
 #include "irods/irods_hierarchy_parser.hpp"
+#include "irods/irods_logger.hpp"
+#include "irods/logical_locking.hpp"
 #include "irods/miscServerFunct.hpp"
 #include "irods/objMetaOpr.hpp"
 #include "irods/regReplica.h"
@@ -24,8 +26,11 @@
 
 namespace
 {
-    namespace ir = irods::experimental::replica;
+    using log_api = irods::experimental::log::api;
+
     namespace id = irods::experimental::data_object;
+    namespace ill = irods::logical_locking;
+    namespace ir = irods::experimental::replica;
 
     std::string compute_checksum_for_resc(
         rsComm_t&              _comm,
@@ -140,6 +145,20 @@ namespace
             return SYS_COPY_ALREADY_IN_RESC;
         }
 
+        // Check to see if the object is locked. If so, an error is returned.
+        if (!irods::server_property_exists(irods::AGENT_CONN_KW) ||
+            nullptr == getValByKey(&_inp.condInput, ill::keywords::bypass))
+        {
+            if (const auto ret = ill::try_lock(*dest.get(), ill::lock_type::write); ret < 0) {
+                const auto msg = fmt::format(
+                    "Registering replica not allowed because data object is locked. error code=[{}], path=[{}]",
+                    ret,
+                    source.logical_path());
+                log_api::info("{}: {}", __func__, msg);
+                return ret;
+            }
+        }
+
         if (cond_input.contains(SU_CLIENT_USER_KW)) {
             const int savedClientAuthFlag = _comm.clientUser.authInfo.authFlag;
 
@@ -221,7 +240,22 @@ int rsRegReplica(rsComm_t *rsComm, regReplica_t *regReplicaInp)
             cond_input[IN_REPL_KW] = "";
         }
 
+        // This logical locking bypass is only allowed for server-to-server connections. If this is not a
+        // server-to-server connection, remove the bypass keyword so that logical locking can be properly enforced.
+        bool restore_ill_bypass_keyword = false;
+        if (nullptr != getValByKey(&regReplicaInp->condInput, ill::keywords::bypass)) {
+            if (!irods::server_property_exists(irods::AGENT_CONN_KW)) {
+                restore_ill_bypass_keyword = true;
+                static_cast<void>(rmKeyVal(&regReplicaInp->condInput, ill::keywords::bypass));
+            }
+        }
+
         const auto ec = rcRegReplica(rodsServerHost->conn, regReplicaInp);
+
+        if (restore_ill_bypass_keyword) {
+            static_cast<void>(addKeyVal(&regReplicaInp->condInput, ill::keywords::bypass, ""));
+        }
+
         if (ec < 0) {
             irods::log(LOG_ERROR, fmt::format("[{}:{}] - failed registering replica [{}] ec=[{}]",
                         __func__, __LINE__, regReplicaInp->destDataObjInfo->objPath, ec));
--- a/server/api/src/rsUnregDataObj.cpp
+++ b/server/api/src/rsUnregDataObj.cpp
@@ -1,17 +1,22 @@
-/*** Copyright (c), The Unregents of the University of California            ***
- *** For more information please refer to files in the COPYRIGHT directory ***/
-/* unregDataObj.c
- */
-
-#include "irods/unregDataObj.h"
-#include "irods/icatHighLevelRoutines.hpp"
-#include "irods/fileDriver.hpp"
-#include "irods/miscServerFunct.hpp"
 #include "irods/rsUnregDataObj.hpp"
 
-#include "irods/irods_file_object.hpp"
-#include "irods/irods_stacktrace.hpp"
+#include "irods/dataObjOpr.hpp"
+#include "irods/fileDriver.hpp"
+#include "irods/icatHighLevelRoutines.hpp"
 #include "irods/irods_configuration_keywords.hpp"
+#include "irods/irods_file_object.hpp"
+#include "irods/irods_logger.hpp"
+#include "irods/irods_server_properties.hpp"
+#include "irods/irods_stacktrace.hpp"
+#include "irods/logical_locking.hpp"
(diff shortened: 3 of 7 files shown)


server/api/src/rsRegReplica.cpp
```diff
@@ -140,6 +145,18 @@ namespace
...
         }
 
+        // Check to see if the object is locked. If so, an error is returned.
+        if (nullptr == getValByKey(&_comm.session_props, ill::keywords::bypass)) {
+            if (const auto ret = ill::try_lock(*dest.get(), ill::lock_type::write); ret < 0) {
+                const auto msg = fmt::format(
+                    "Registering replica not allowed because data object is locked. error code=[{}], path=[{}]",
+                    ret,
+                    source.logical_path());
+                log_api::info("{}: {}", __func__, msg);
```
korydraughn, line 157: Use of `info` seems intentional. What's the motivation behind using `info` instead of `error`?

server/api/src/rsUnregDataObj.cpp
```diff
@@ -71,6 +69,32 @@ _rsUnregDataObj( rsComm_t *rsComm, unregDataObj_t *unregDataObjInp ) {
...
+                log_api::error("{}: data object [{}] does not exist. [error code={}]", __func__, inp.objPath, ret);
+                return ret;
+            }
+
+            if (const auto ret = ill::try_lock(*info, ill::lock_type::write); ret < 0) {
+                const auto msg =
+                    fmt::format("Unregister not allowed because data object is locked. error code=[{}], path=[{}]",
+                                ret,
+                                dataObjInfo->objPath);
+                log_api::info("{}: {}", __func__, msg);
```
korydraughn, line 116: Highlighting this is logged as `info`. Awaiting response to previous review comment.

server/core/include/irods/logical_locking.hpp
```diff
@@ -13,6 +13,17 @@ struct RsComm;
...
 {
+    namespace keywords
+    {
+        // This is a special keyword which can be used to explicitly bypass logical locking. The application will
+        // determine whether or not it is honored. This should only be used internally in the iRODS server and is
+        // not meant for client consumption. Furthermore, its usage should be limited to highly specific cases
+        // wherein a replica access token is not available but an object which needs to be modified is locked. It
+        // is recommended that this be used in the session_props of the RsComm so that only internal server
+        // operations can use it.
+        static constexpr const char* bypass = "bypass_logical_locking";
```
korydraughn: I think you need `inline` here so that the compiler defines the string once across all files instead of once per translation unit.

unit_tests/src/test_logical_locking.cpp
```diff
@@ -823,3 +870,332 @@ TEST_CASE("rename", "[write_lock]")
...
+        CHECK(GOOD_REPLICA == ir::replica_status(new_comm, target_object, opened_replica_resc));
+        CHECK(GOOD_REPLICA == ir::replica_status(new_comm, target_object, other_replica_resc));
+    }};
+
+    DataObjInfo info{};
+    std::strncpy(static_cast<char*>(info.objPath), target_object.c_str(), sizeof(info.objPath) - 1);
+    unregDataObj_t unreg_inp{};
+    unreg_inp.dataObjInfo = &info;
+
+    // Populating the rescHier member of the input DataObjInfo is the only way to target things for UnregDataObj.
```
korydraughn: Let's capture that in the header file if we don't already. It can be a separate commit so previous work remains unchanged.

unit_tests/src/test_logical_locking.cpp
```diff
@@ -823,3 +870,332 @@ TEST_CASE("rename", "[write_lock]")
...
+
+    // Replicate the test data object so that it has multiple replicas
+    REQUIRE(unit_test_utils::replicate_data_object(comm, target_object.c_str(), resc_1));
+    REQUIRE(GOOD_REPLICA == ir::replica_status(comm, target_object, other_replica_resc));
+
+    // Create a local file to register.
+    {
+        static constexpr auto contents = std::string_view{"content!"};
+        std::ofstream fstream{filename};
+        fstream << contents;
```
korydraughn: Can be rewritten as ...
```cpp
std::ofstream{filename} << "content!";
```

Not required though. Just caught my eye.

unit_tests/src/test_logical_locking.cpp
```diff
@@ -823,3 +870,332 @@ TEST_CASE("rename", "[write_lock]")
...
+            // Target resource which has no replicas for this data object.
+            addKeyVal(&reg_inp.condInput, DEST_RESC_NAME_KW, resc_2.c_str());
+            REQUIRE(LOCKED_DATA_OBJECT_ACCESS == rcPhyPathReg(&new_comm, &reg_inp));
+            CHECK(!ir::replica_exists(new_comm, target_object.c_str(), resc_2));
+        }
+    }
+
+    SECTION("RegDataObj")
+    {
+        // This API does not target replicas and can only register new data objects. If the data object is locked, that
```
korydraughn, line 1095: Does the header or cpp file say anything like your first statement? If not, let's add that to it.

unit_tests/src/test_logical_locking.cpp
```diff
@@ -823,3 +870,332 @@ TEST_CASE("rename", "[write_lock]")
...
+        // means it exists and the registration is going to fail for that reason, not because it's locked.
+
+        DataObjInfo reg_inp{};
+        std::strncpy(static_cast<char*>(reg_inp.objPath), target_object.c_str(), sizeof(reg_inp.objPath) - 1);
+        std::strncpy(static_cast<char*>(reg_inp.dataType), GENERIC_DT_STR, sizeof(reg_inp.dataType) - 1);
```
korydraughn, line 1100: Does the API actually require `GENERIC_DT_STR`?

unit_tests/src/test_logical_locking.cpp
```diff
@@ -823,3 +870,332 @@ TEST_CASE("rename", "[write_lock]")
...
+        REQUIRE(CATALOG_ALREADY_HAS_ITEM_BY_THAT_NAME == rcRegDataObj(&new_comm, &reg_inp, &reg_out));
+        CHECK(0 == freeAllDataObjInfo(reg_out));
+
+        // Target resource which has no replicas for this data object.
+        std::strncpy(static_cast<char*>(reg_inp.rescHier), resc_2.c_str(), sizeof(reg_inp.rescHier) - 1);
+        REQUIRE(CATALOG_ALREADY_HAS_ITEM_BY_THAT_NAME == rcRegDataObj(&new_comm, &reg_inp, &reg_out));
+        CHECK(!ir::replica_exists(new_comm, target_object.c_str(), resc_2));
+        CHECK(0 == freeAllDataObjInfo(reg_out));
+
+        // REG_REPL_KW is not supported in RegDataObj.
```
korydraughn, line 1125: I take it that the flag is ignored, right?

korydraughn (commented):
Looks good so far.

server/api/src/rsRegReplica.cpp
alanking: Mainly to avoid log spam. This is the level used in the place from which I took the original code, but I'm not partial to it. `error` is fine, too.

server/core/include/irods/logical_locking.hpp
alanking: Will do

unit_tests/src/test_logical_locking.cpp
alanking: Will do

unit_tests/src/test_logical_locking.cpp
alanking: Will do

unit_tests/src/test_logical_locking.cpp
alanking: Will do

unit_tests/src/test_logical_locking.cpp
alanking: Pretty much. I was getting errors about an unrecognized type or something without it.

unit_tests/src/test_logical_locking.cpp
alanking: Correct

alanking:
From core test run, it seems that Phymv is affected by these changes, so I'll need to look into that. Changes forthcoming.

unit_tests/src/test_logical_locking.cpp
alanking: Looking into this a little further, populating `rescHier` doesn't do anything in UnregDataObj. This API is a thin wrapper around `chlUnregDataObj`, which just unregisters all replicas for the specified data object. I'll update the header file anyway and remove this comment.

alanking: Maybe this can just be reduced down to one test...

unit_tests/src/test_logical_locking.cpp
korydraughn: Good to know. Thanks.

server/api/src/rsRegReplica.cpp
korydraughn: Okay.

server/api/src/rsUnregDataObj.cpp
korydraughn: This is fine.

unit_tests/src/test_logical_locking.cpp
alanking: Done

unit_tests/src/test_logical_locking.cpp
alanking: This gave me a compiler error:
```
/irods_source/unit_tests/src/test_logical_locking.cpp:936:36: error: expected ';' at end of declaration
    std::ofstream fstream{filename} << "content!";
                                   ^   
                                   ;   
1 error generated.
```

unit_tests/src/test_logical_locking.cpp
alanking: Done

server/core/include/irods/logical_locking.hpp
alanking: Done

unit_tests/src/test_logical_locking.cpp
korydraughn: You have to remove the variable name, `fstream`.

alanking: Oh :facepalm:

alanking: Done

alanking:
Core tests and unit tests are passing. I would also like to run tests in topology from a resource server to make sure all the flags are getting passed around as they should on redirects.

korydraughn:
Yes, please do.

alanking:
As suspected, the logical locking bypass session property is not propagated to the new server connection when the connection redirects from a catalog consumer to catalog provider. Will need to think of a way to do this without allowing regular clients to use it, or we'll need to come up with another approach (probably something like #8765).

alanking:
Topology-on-consumer tests are still running with latest changes, but all previously observed failures have been resolved. Will report back when those complete. Unit tests passed.

alanking:
Topology-on-consumer tests passed except for the one fixed in #8810. Running core tests once more to be sure, but I don't anticipate any problems.

korydraughn:
Will circle back after testing is successful.

server/api/src/rsDataObjOpen.cpp
```diff
@@ -131,10 +131,19 @@ namespace
                                          p.object_name().c_str(),
                                          kvp.at(RESC_HIER_STR_KW).value());
 
-            for (auto&& row : irods::query{&_conn, gql}) {
-                data_id = std::stoull(row[0]);
-                replica_number = std::stoul(row[1]);
+            auto query = irods::query{&_conn, gql};
+
+            // TODO(#8809) - remove static_cast when empty returns bool
```
korydraughn:
```suggestion
            // TODO(#8809): remove static_cast when empty returns bool
```

server/api/src/rsUnregDataObj.cpp
```diff
@@ -71,6 +90,34 @@ _rsUnregDataObj( rsComm_t *rsComm, unregDataObj_t *unregDataObjInp ) {
...
+
+            // Get data object information. Returns error if object does not exist.
+            DataObjInp inp{};
+            const auto clear_condInput = irods::at_scope_exit{[&inp] { clearKeyVal(&inp.condInput); }};
+            if (nullptr != getValByKey(unregDataObjInp->condInput, ADMIN_KW)) {
+                addKeyVal(&inp.condInput, ADMIN_KW, "");
+            }
+            std::strncpy(static_cast<char*>(inp.objPath), dataObjInfo->objPath, sizeof(inp.objPath) - 1);
+            if (const auto ret = getDataObjInfo(rsComm, &inp, &info, ACCESS_DELETE_OBJECT, 0); ret < 0) {
+                log_api::error("{}: data object [{}] does not exist. [error code={}]", __func__, inp.objPath, ret);
```
korydraughn:
```suggestion
                log_api::error("{}: Data object [{}] does not exist. [error code={}]", __func__, inp.objPath, ret);
```

alanking:
Well, it seems like there are some issues around the commit addressing #8808. Unfortunately, we seem to be leaning on this uninitialized value behavior for some cases and/or it has been masking other issues. I'm going to back out those changes and address them separately.

Running core tests once more.

server/api/src/rsUnregDataObj.cpp
alanking: Done

server/api/src/rsDataObjOpen.cpp
alanking: I reverted this change, so, resolving.

alanking:
The clang-tidy error is captured in #8808. Everything else is a warning.

alanking:
Core tests passed

korydraughn:
Do we anticipate any issues related to topology testing following the revert?

alanking:
I do not, but I can run it again if desired.

korydraughn:
Squash to taste.

alanking:
Squashed. Uninitialized value commit and the revert have been removed

alanking:
#'d, mergin

PR #8810, opened 2026-01-22
[#7128] Fix istream test for topology-on-consumer (4-3-stable)

alanking:
In service of #7128

The test now passes in core, topology-on-consumer, and topology-on-provider.

The reason this fails for topoogy-on-consumer is very surprising to me. `itouch` (with no `-R`) creates the first replica on the local resource (attached to the resource consumer) but `istream` (with no `-R`) creates a new replica on `demoResc`. It may even constitute a bug, but I am not sure about that yet quite yet.

### DIFF
--- a/scripts/irods/test/test_istream.py
+++ b/scripts/irods/test/test_istream.py
@@ -523,10 +523,14 @@ class Test_Istream(session.make_sessions_mixin([('otherrods', 'rods')], [('alice
 
     def test_istream_write_with_various_options_on_empty_replica_with_effects_on_mtime__issue_7128(self):
         logical_path = "/".join([self.user.session_collection, "emptiness"])
+        test_resource = "mtime_resc"
 
         try:
+            # Make a test resource on a remote machine (where applicable).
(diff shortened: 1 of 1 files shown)


korydraughn:
The help text for itouch states `irods_default_resource` in irods_environment.json is ignored.

I think istream honors that property.

Does that change your thoughts on the behavior you witnessed?

korydraughn (commented):
Test changes look fine. Awaiting results.

korydraughn:
Wait, you already ran tests for this, right?

alanking:
> Wait, you already ran tests for this, right?

I just ran the test that I touched in the various modes described in OP, so, sorta.

> The help text for itouch states `irods_default_resource` in irods_environment.json is ignored.
> 
> I think istream honors that property.
> 
> Does that change your thoughts on the behavior you witnessed?

I see. I am not sure that I agree that the behavior is good, but at least it's documented. Good to know for the future.

alanking:
#'d, mergin

Issue #8593, opened 2025-07-23
iRODS 4.3.3 Crash (SIGABRT): Double free or corruption in clearBBuf when handling storage I/O error


messersc:
### iRODS Version, OS and Version

irods v.4.3.3, in a Ubuntu 20.04 docker container with a cephfs Vault mounted on the host VM

### What did you try to do?

Using our omics management system SODAR, we get errors when trying to move a file in irods to another path. This might be caused by high load or slow responses on the underlying storage (cephfs).

#### irods log

```
 {"log_category":"legacy","log_level":"error","log_message":"[rsDataObjClose:803] - [Unknown iRODS error: [update_replica_size_and_throw_on_failure:463] - failed to get size in vault [error_code=[4716061728], path=[/sodarZone/projects/ad/ad163ce0-1db4-4517-b7f0-e90b37f95da9/landing_zones/sodar_user/REDACTED], hierarchy=[base;cephfs-2]]\n\n]","request_api_name":"DATA_OBJ_CLOSE_AN","request_api_number":673,"request_api_version":"d","request_client_user":"sodar_user","request_host":"10.32.105.17","request_proxy_user":"sodar_user","request_release_version":"rods4.3.4","server_host":"sodar.bihealth.org","server_pid":1062467,"server_timestamp":"2025-07-22T14:11:06.365Z","server_type":"agent","server_zone":"sodarZone"}
 {"log_category":"legacy","log_level":"error","log_message":"[rsDataObjClose:803] - [Unknown iRODS error: [update_replica_size_and_throw_on_failure:463] - failed to get size in vault [error_code=[48647108], path=[/sodarZone/projects/ad/ad163ce0-1db4-4517-b7f0-e90b37f95da9/landing_zones/sodar_user/REDACTED], hierarchy=[base;cephfs-2]]\n\n]","request_api_name":"DATA_OBJ_CLOSE_AN","request_api_number":673,"request_api_version":"d","request_client_user":"sodar_user","request_host":"10.32.105.7","request_proxy_user":"sodar_user","request_release_version":"rods4.3.4","server_host":"sodar.bihealth.org","server_pid":1062039,"server_timestamp":"2025-07-22T14:11:06.498Z","server_type":"agent","server_zone":"sodarZone"}
[...]

PR #8811, opened 2026-01-22
[#8593] Network plugins: Set free'd pointers to nullptr (4-3-stable)

alanking:
In service of #8593

Not sure if this fixes the problem because I cannot reproduce it. Still need to run tests, so leaving this in draft for now.


alanking:
Unit and core tests passed. Given that this deals with networking, I'm going to do one run in topology with TLS enabled

alanking:
Topology-on-consumer with TLS enabled passed (except for the failing `istream` test fixed in #8810)

alanking:
#'d, mergin

Issue #8808, opened 2026-01-22
clang-tidy: Possibly uninitialized values used in update_replica_access_table


alanking:
- [x] main

---

```
2026-01-22T15:44:58.6374400Z /home/runner/work/irods/irods/server/api/src/rsDataObjOpen.cpp:149:40: error: 1st function call argument is an uninitialized value [clang-analyzer-core.CallAndMessage,-warnings-as-errors]
2026-01-22T15:44:58.6383198Z                 l1desc.replica_token = rat::create_new_entry(data_id, replica_number, getpid());
```

Relevant code:
https://github.com/irods/irods/blob/3d2abaef56ee8532ea8d902c0e68b5c0ca98fdef/server/api/src/rsDataObjOpen.cpp#L124-L158

It's possible for `data_id` and `replica_number` to be uninitialized here in the case that the query returns no results. These are both `std::uint32_t`s, so initializing to a negative value is technically not a valid solution (although it is incredibly unlikely that a data ID or replica number would be such large values). We should just make sure that the query has results so that the variables have values.

Also note that default initialization won't work either because 0 is a valid value for replica number.

alanking:
Interestingly, this report did not appear in my clang-tidy results on the main branch even though the 4-3-stable PR workflow found this.

PR #8818, opened 2026-01-27
[#8808] DataObjOpen: Rework update to replica access table (main)

alanking:
Addresses #8808

Passing tests:
- [x] unit
- [x] core
- [x] topology-on-consumer
- [x] federation

### DIFF
--- a/server/api/src/rsDataObjOpen.cpp
+++ b/server/api/src/rsDataObjOpen.cpp
@@ -113,48 +113,34 @@ namespace
         update
     };
 
-    auto update_replica_access_table(rsComm_t& _conn,
-                                     update_operation _op,
+    auto update_replica_access_table(update_operation _op,
                                      int _l1desc_index,
-                                     const dataObjInp_t& _input) -> int
+                                     const ir::replica_proxy_t& _replica,
+                                     const char* _replica_token = "") -> int
     {
-        const fs::path p = _input.objPath;
-        const irods::experimental::key_value_proxy kvp{_input.condInput};
-
-        rat::data_id_type data_id;
-        rat::replica_number_type replica_number;
-
-        try {
-            const auto gql = fmt::format("select DATA_ID, DATA_REPL_NUM "
-                                         "where"
-                                         " COLL_NAME = '{}' and"
-                                         " DATA_NAME = '{}' and"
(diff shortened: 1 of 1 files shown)


alanking:
Tests listed above passed

server/api/src/rsDataObjOpen.cpp
```diff
@@ -113,48 +113,34 @@ namespace
...
-
-        try {
-            const auto gql = fmt::format("select DATA_ID, DATA_REPL_NUM "
-                                         "where"
-                                         " COLL_NAME = '{}' and"
-                                         " DATA_NAME = '{}' and"
-                                         " DATA_RESC_HIER = '{}'",
-                                         irods::single_quotes_to_hex(p.parent_path()),
-                                         irods::single_quotes_to_hex(p.object_name()),
-                                         kvp.at(RESC_HIER_STR_KW).value());
```
korydraughn, line 135: From what you can tell, the implementation wasn't leaning on the liveness of GenQuery in any way, correct?

We just used GenQuery because that was the most convenient way to get the data id and replica number.

alanking: It shouldn't be - the information is available in both places from which this function is called.

In the one case, we just registered a new replica in the catalog and that information (including the data ID and replica number) was returned and put into the DataObjInfo of the L1 descriptor. That is demonstrated here:
https://github.com/irods/irods/blob/5f8705fd007c47e362744f6223b9d90e75b8ce17/server/api/src/rsDataObjOpen.cpp#L441-L444

For the other case, we are opening an existing replica for write. In order to have enough information to do that, we had to have already run a query to get all the data object information, which we do here: https://github.com/irods/irods/blob/5f8705fd007c47e362744f6223b9d90e75b8ce17/server/api/src/rsDataObjOpen.cpp#L1013

The information about the replica to be opened is then found here: https://github.com/irods/irods/blob/5f8705fd007c47e362744f6223b9d90e75b8ce17/server/api/src/rsDataObjOpen.cpp#L1095-L1104

So, all the information that query was giving us was already available from prior trips to the database in each case handled by DataObjOpen.

korydraughn: Thanks.

alanking:
#'d, mergin

Issue #8580, opened 2025-07-16
Update library_features.h to include new features for iRODS 5


korydraughn:
## Enhancement

The following items need macros in library_features.h so that other projects can take detect their presence. 
- rc_authenticate_client
- delay rule locking APIs
- access time for genquery
- changes in DISTINCT keyword _(applies to 4.3.5)_

This list is NOT exhaustive.

korydraughn:
#8672 - overriding packing instructions

korydraughn:
#8701 - crc64/nvme checksums (iRODS 5 only)
#8697 - new auth scheme supporting session tokens (iRODS 5 only)

PR #8819, opened 2026-01-27
[#8580] Update feature test macros (4-3-stable)

korydraughn:
Will run the unit tests to confirm correctness.

Of note is the change to the GenQuery2 feature test macro's value.

### DIFF
--- a/lib/core/include/irods/library_features.h
+++ b/lib/core/include/irods/library_features.h
@@ -99,7 +99,17 @@
 
 /// Defined if the development library supports #rc_genquery2.
 ///
+/// - 202503L: User is granted control over the DISTINCT keyword
+/// - 202404L: Initial implementation
+///
 /// \since 4.3.2
-#define IRODS_LIBRARY_FEATURE_GENQUERY2               202404L
+#define IRODS_LIBRARY_FEATURE_GENQUERY2                     202503L
+
+/// Defined if the development library supports #procApiRequest_raw.
+///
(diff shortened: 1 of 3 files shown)


lib/core/include/irods/library_features.h
```diff
@@ -99,7 +99,17 @@
 
 /// Defined if the development library supports #rc_genquery2.
 ///
+/// - 202404L: Initial implementation
+/// - 202503L: User is granted control over the DISTINCT keyword
```
korydraughn: Invert listing.

korydraughn:
Hmm, good question.

My immediate thought is yes. The documentation within the header states that values may change. And that is likely to only happen during major/minor releases.

Will continue to think through that though. Happy to hear detailed thoughts/concerns if you have any.

korydraughn:
I want to add that using feature test macros is a compile-time thing too. As long as users aren't mixing development libraries and runtimes, I think these kinds of changes are safe.

lib/core/include/irods/library_features.h

korydraughn:
Unit tests passed.

```
All tests passed! :)
time elapsed: [ 674.7344]seconds ([   0]hours [11.2456]minutes)

2026-01-27 15:47:14,327 - version:[4.3.4]
2026-01-27 15:47:14,327 - sha:[e82038e7ad70246bc88dbe07673d90f08153a0df]
```

korydraughn:
Added pound.

Issue #8518, opened 2025-05-15
Serialize ticketAdminInp for pep_api_ticket_admin_*


tedgin:
### New Feature / Functionality

- Please describe the feature you are requesting and your use case.

iRODS ticket creation is relatively slow when done in bulk. We have a third-party service that passes out tickets to sets of files that aren't all in the same folder. It can generate over 50 tickets at a time. The service needs to modify the ticket after its created using to separate API calls. Each call takes ~0.3 s, leading to a total of ~0.9 s per ticket.  For fifty tickets, this takes nearly 50 s. The owners of this service aren't please that it takes so long. I would like to use PEPs to automatically set the desired ticket properties when the ticket is created, which would reduce the service to making a single API call per ticket, and hopefully reduce the total ticket creation time to less than 20 s.  

It looks like the PEP I need to use is `pep_api_ticket_admin_post`. Unfortunately, its parameter `ticketAdminInp` isn't serialized, so no information is available about the ticket administration operation.  

It would be great if iRODS serialized the `ticketAdminInp` parameter for the `pep_api_ticket_admin_*` PEPs.

trel:
related to https://github.com/irods/irods/issues/5955

korydraughn:
Reclassifying this as a bug since all data types for RPC API operations should be serializable for the REPF unless explicitly specified as not being serializable.

PR #8837, opened 2026-02-05
[#8518] Serialize TicketAdminInput for policy enforcement (4-3-stable)

korydraughn:
Running unit/core tests.

### DIFF
--- a/server/re/src/irods_re_serialization.cpp
+++ b/server/re/src/irods_re_serialization.cpp
@@ -4,6 +4,7 @@
 #include "irods/irods_plugin_context.hpp"
 #include "irods/msParam.h"
 #include "irods/rodsErrorTable.h"
+#include "irods/ticketAdmin.h"
 
 #include <boost/lexical_cast.hpp>
 #include <boost/range/adaptors.hpp>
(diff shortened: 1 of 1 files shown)


korydraughn:
Unit/Core tests passed.

scripts/irods/test/test_dynamic_peps.py
```diff
@@ -692,3 +692,66 @@ def test_if_MsParamArray_is_exposed__issue_7553(self):
...
+                *v = *input.arg4;
+                msiModAVUMetadata('-C', '{self.admin.session_collection}', 'add', '{avu_prefix}_4', *v, '');
+
+                *v = *input.arg5;
+                msiModAVUMetadata('-C', '{self.admin.session_collection}', 'add', '{avu_prefix}_5', *v, '');
+
+                *v = *input.arg6;
+                if (strlen(*v) == 0) {{
+                    msiModAVUMetadata('-C', '{self.admin.session_collection}', 'add', '{avu_prefix}_6', 'empty', '');
+                }}
```
alanking, line 728: Why do we not add an AVU if `*v` has a non-empty value?

scripts/irods/test/test_dynamic_peps.py
```diff
@@ -692,3 +692,66 @@ def test_if_MsParamArray_is_exposed__issue_7553(self):
...
+                    f'attribute: {avu_prefix}_2\nvalue: {ticket_name}\n',
+                    f'attribute: {avu_prefix}_3\nvalue: {ticket_type}\n',
+                    f'attribute: {avu_prefix}_4\nvalue: {self.admin.session_collection}\n',
+                    f'attribute: {avu_prefix}_5\nvalue: {ticket_name}\n',
+                    f'attribute: {avu_prefix}_6\nvalue: empty\n',
+                    f'attribute: {avu_prefix}_ADMIN_KW\nvalue: 1\n',
+                ])
+
+            finally:
+                if ticket:
```
alanking: Will this ever be true? I don't see `ticket` getting a value anywhere.

scripts/irods/test/test_dynamic_peps.py
korydraughn: At this time, `arg6` isn't set by the TicketAdmin API. This leads to `*v` being an empty string at the policy layer.

If the value is non-empty in the future, this test will catch it and report it to us. It also covers the no value case, which is harder to demonstrate through icommands.

scripts/irods/test/test_dynamic_peps.py
korydraughn: Nice catch. It is a leftover. Will remove.

scripts/irods/test/test_dynamic_peps.py
alanking: Okay. Let's leave a comment about that for if/when that ever happens

korydraughn:
> Looks good overall. Is there any value in creating a unit test for this?
> 
> https://github.com/irods/irods/blob/bb0bf7bbf1883b8e8d3bd4676e8ff70d9257c5f1/unit_tests/src/test_re_serialization.cpp

I don't think so as the python test covers the one and only code path.

scripts/irods/test/test_dynamic_peps.py
korydraughn: Done.

scripts/irods/test/test_dynamic_peps.py
korydraughn: Updated to be a boolean that helps in signaling when removal of the ticket is needed. It's more of a "don't create noise in the log" thing.

korydraughn:
Test still passes.

korydraughn:
Squashed.

alanking (approved):
\# it

korydraughn:
Added pounds.

PR #8838, opened 2026-02-05
[#8518] Serialize TicketAdminInput for policy enforcement (main)

korydraughn:
Ported from #8837.

Test passes in isolation.

Unit tests pass. I don't think running the core tests is necessary since existing tests do not have policy that relies on the TicketAdminInput type, but can do that if desired.


alanking (approved):
I don't think you need to run all the tests again on main. Just the new test is fine

The differences which do exist make sense. # it.

korydraughn:
Will squash and pound once clang-tidy is green.

korydraughn:
Clang-tidy is happy.

Squashing and pounding.

korydraughn:
Added pounds.

Issue #2720, opened 2015-06-01
Tickets will increment both uses and write-file on failed access


tempoz:
If a ticket is already at maximum uses and another access is attempted, it will increment its uses past maximum. This is problematic, as it allows for one fewer use if the maximum uses is later increased.

trel:
not sure this is true anymore...  see recent work for https://github.com/irods/irods/issues/7967

alanking:
Seems to me that we still need a test specific to this issue, or the test added for that issue needs more assertions to make sure it covers the issue described here.

FifthPotato:
If I'm not mistaken, I think line 1142 in here https://github.com/irods/irods/commit/1ab8550efee93f4163304bfabbe3fc371e264833#diff-ebe2f236ca7d17d08e65cf342eacf1aa11e4e3c606046c003769d805970966a0 covers the "uses count" part, in that it shows that the uses count stays at 3 for 3 accesses, which would be consistent with it working as expected (i.e. it doesn't increment past the limit). Should I also add a case in which it gets changed to 5 after 3 uses, then show it can still be used the expected 2 more times?

korydraughn:
> Should I also add a case in which it gets changed to 5 after 3 uses, then show it can still be used the expected 2 more times?

Yes please.

Also, we need to demonstrate the same thing with `iput` - i.e. dup the test and show that `iput` cannot break `uses count`.

korydraughn:
For `iput`, we need to cover the `write-file` count too.

korydraughn:
While working on this issue, I discovered the handling of `write byte count`/`write byte limit` has some issues that aren't as easily fixed as the other properties. Those problems will be addressed in iRODS 5. See #8844 for more information.

This issue will address the remaining ticket options.

PR #8845, opened 2026-02-10
[#2720] Do not increment ticket write-file count after reaching limit (4-3-stable)

korydraughn:
Unit and core tests pass.


korydraughn:
Squashed only.

korydraughn:
Added pounds.

Issue #8843, opened 2026-02-09
`std::memset` called on `RsComm` can result in undefined behavior


korydraughn:
## Bug Report

`RsComm` contains nontrivial types like `std::string`. Calling `std::memset` on such types is not allowed by the C++ standard.

https://github.com/irods/irods/blob/9f629e3f77dd611eae9158077683f654102409f3/server/main_server/src/agent_main.cpp#L667

The passed `RsComm` is cleared before the call to `initServerMain`, so the fix is to simply drop the call to `std::memset`.

PR #8846, opened 2026-02-10
[#2720,#8843] Do not increment ticket write-file count after reaching limit / Do not use memset on RsComm (main)

korydraughn:
Cherry-picked from #8845.

This PR also includes a fix for memset and RsComm.

Unit and core tests pass.


korydraughn:
Added pounds.

PR #8662, opened 2025-09-05
[#7833] Make `update_deprecated_columns.py` work with Python3 (4-3-stable)

MartinFlores751:
Issue quick link: #7833

This PR attempts to address issues in running `update_deprecated_columns.py` on Python 3. This PR also attempts to improve the error reporting on failed connections.

---

Before getting out of draft, test that:
- [ ] The script works on Python 3.
- [ ] The script print improved errors.

Issue #7833, opened 2024-06-24
update_deprecated_columns.py updates needed


JustinKyleJames:
-  [x] 4-3-stable only

---

The update_deprecated_columns.py file no longer works for 4.3/python3 and needs to be updated.  Here are the changes needed:

## Change 1

https://github.com/irods/irods/blob/fc47213018529155ccfdc4365c71ab3dffaf6613/scripts/update_deprecated_database_columns.py#L28

Should be:

```
return int(database_connect.execute_sql_statement(cursor, select_scrubbable_count).fetchone()[0])
```

## Change 2

https://github.com/irods/irods/blob/fc47213018529155ccfdc4365c71ab3dffaf6613/scripts/update_deprecated_database_columns.py#L85

Should be:

```
total_update_count = int(database_connect.execute_sql_statement(cursor, select_update_count).fetchone()[0])
```

## Suggested Change

In addition it would be helpful to print the exception here to catch these types of errors as the printed line can be misleading.

https://github.com/irods/irods/blob/fc47213018529155ccfdc4365c71ab3dffaf6613/scripts/update_deprecated_database_columns.py#L130-L131

alanking:
This script was removed in #7834 and deprecated in https://github.com/irods/irods/issues/7835.

korydraughn:
Nice catch!

trel:
so there's nothing to do here at all, since it's already deprecated?

korydraughn:
Right.

alanking:
#8662 seems to already have done the work, and our deprecation docs (https://docs.irods.org/5.0.1/developers/deprecation) say:

> The feature will continue to be supported for all major versions in which it is deprecated (and previous major versions in which the feature exists).

I think a fix is appropriate since the script doesn't work with any of the platforms 4.3.5 will ship for.

trel:
got it.   that makes sense.

PR #8848, opened 2026-02-11
[#7833] Fix `update_deprecated_columns.py` (4-3-stable)

korydraughn:
Supersedes #8662.

Tested against PostgreSQL, MySQL 8, and MariaDB.


korydraughn:
Added pounds.

Issue #8528, opened 2025-05-21
`msiSetKeyValuePairsToObj` fails with 4.3.4


mstfdkmn:
- [x] main
- [x] 4-3-stable
---
### iRODS Version, OS and Version
4.3.4
### What did you try to do?
We encountered an error with one of our rules after upgrading to version 4.3.4. So I tried tried to set metadata using `msiSetKeyValuePairsToOb` on users and objects.
### Expected behavior
I expected to see that this msi set metadata without any errors.
### Observed Behavior (including steps to reproduce, if applicable)
Metadata cannot be set through that msi and depending on the object type we see different errors in irods log.

The following script can be used to reproduce the issue.

```
root@824cd214c5e2:/# cat test_msi.r
import irods_types

def main(rule_args, callback, rei):

    avu = callback.msiString2KeyValPair(
            "mg.mail-over-quota" + "=" + "true", irods_types.KeyValPair()
        )["arguments"][1]

    #callback.msiSetKeyValuePairsToObj(avu, "/icts/home/rods", "-C")
    callback.msiSetKeyValuePairsToObj(avu, "rods", "-u")
    #callback.msiSetKeyValuePairsToObj(avu, "/icts/home/rods/test.txt", "-d")

INPUT null
OUTPUT ruleExecOut
```

```
root@824cd214c5e2:/# irule -r irods_rule_engine_plugin-python-instance -F test_msi.r
remote addresses: 10.114.48.180 ERROR: rcExecMyRule error.  status = -1828000 RULE_ENGINE_ERROR
```

### iRODS Server Logs (if applicable)

logs when attempted for setting metadata on a user
```
[...]

PR #8851, opened 2026-02-13
[#8528] Add test for PREP and msiSetKeyValuePairsToObj (4-3-stable)

korydraughn:
Will run the test soon. No need to run the full test suite since there's no change to the server.

I'll note that the new test was copied from the one just above it. The only true difference was the use of the bytesbuf type. Other than that, the tests are practically identical.

### DIFF
--- a/scripts/irods/test/test_python_rule_engine_plugin.py
+++ b/scripts/irods/test/test_python_rule_engine_plugin.py
@@ -134,3 +134,49 @@ class Test_Python_Rule_Engine_Plugin(session.make_sessions_mixin([('otherrods', 
             f'attribute: {avu_name}\n',
             f'value: {avu_value}\n'
         ])
+
+    @unittest.skipUnless(plugin_name == 'irods_rule_engine_plugin-python', 'only applicable for python REP')
+    def test_msiSetKeyValPairsToObj_does_not_result_in_rule_engine_error__issue_8528(self):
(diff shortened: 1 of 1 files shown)


scripts/irods/test/test_python_rule_engine_plugin.py
```diff
@@ -134,3 +134,44 @@ def main(rule_args, callback, rei):
...
+            f'value: {avu_value}\n'
+        ])
+        self.admin.assert_icommand(['imeta', 'ls', '-u', self.admin.username], 'STDOUT', [
+            f'attribute: {avu_name}\n',
+            f'value: {avu_value}\n'
+        ])
+        self.admin.assert_icommand(['imeta', 'ls', '-d', data_object], 'STDOUT', [
+            f'attribute: {avu_name}\n',
+            f'value: {avu_value}\n'
+        ])
```
MartinFlores751, line 178: I see that we're testing just about all the items we could attach an AVU to. Think it'd be worth testing the resource case too?

korydraughn: Will do.

korydraughn: Done.

korydraughn:
Had to update test to use an absolute logical path.

Test passed.

PR #8854, opened 2026-02-16
[#8817] Use ccache and ninja in GitHub Actions (main)

korydraughn:

Issue #8817, opened 2026-01-26
Use ccache in GitHub Actions when compiling iRODS code


korydraughn:
- [x] main
- [x] 4-3-stable
---
Reference implementation at  https://github.com/irods/irods_reusable_github_workflows/pull/27

PR #8855, opened 2026-02-16
[#8817] Use ccache and ninja in GitHub Actions (4-3-stable)

korydraughn:
Ported from #8854.


korydraughn:
I suspect the addition of `-j` is negatively affecting the builds. I don't have any evidence to support that thought, but I'll go ahead and switch the build to use ninja to see if that resolves the situation.

korydraughn:
Sticking with ninja.

Squashing.

korydraughn:
Now we wait to see if the cache is used for each build.

korydraughn:
Added pound.

Issue #8857, opened 2026-02-17
Memory leak in Agent executing delay rule


korydraughn:
- [ ] main
- [ ] 4-3-stable
---
## Bug Report

The following information was reported by Address Sanitizer while working on #8575 and #8254.

The function of interest is `rsExecRuleExpression`. The report _(at the bottom of the page)_ points to L38 in rsExecRuleExpression.cpp, but that's because I inserted 3 additional lines to help with debugging. Restoring the code to its original state results in the callstack pointing to L35. L35 matches the original report produced by ASan. I just didn't have the original report available while writing this issue. 

https://github.com/irods/irods/blob/5818a0fb18a3f352b1a5b2eb9f93ab816759cde9/server/api/src/rsExecRuleExpression.cpp#L35

Looking at `r_rule_exec.exe_context` in the database reveals that there are multiple `KeyValPairs` inside of it. The iRODS server does not deallocate memory for those when using `clearMsParam` or `clearMsParamArray`. The reason for this can be observed in the following block.

https://github.com/irods/irods/blob/5818a0fb18a3f352b1a5b2eb9f93ab816759cde9/lib/core/src/msParam.cpp#L573-L587

While not ideal, in the context of delay rule execution, it's mostly fine because the agent will terminate after processing the rule, allowing the OS to reclaim memory resources.

### The Delay Rule

This is the delay rule I used to create the situation.

```py
pep_api_data_obj_put_post(*INSTANCE_NAME, *COMM, *DATAOBJINP, *BUFFER, *PORTAL_OPR_OUT) {
        delay("<INST_NAME>irods_rule_engine_plugin-irods_rule_language-instance</INST_NAME><EF>REPEAT FOR EVER</EF>") {
                writeLine("serverLog", "----- PUT POST FROM DELAY RULE -----");
        }
}
```

### ASan Report _(from modified build of 4-3-stable)_

This is the change I made to the code which resulted in the line reported by ASan being L38 instead of L35. I inserted those lines for debugging purposes.

```diff
diff --git a/server/api/src/rsExecRuleExpression.cpp b/server/api/src/rsExecRuleExpression.cpp
index 4e71e9722..716ac1e86 100644
--- a/server/api/src/rsExecRuleExpression.cpp
+++ b/server/api/src/rsExecRuleExpression.cpp
@@ -9,6 +9,7 @@
 #include "irods/rcMisc.h"
 #include "irods/stringOpr.h"
 
+#include <csignal>
 #include <cstdlib>
[...]

PR #8860, opened 2026-02-19
[#8857] Fix memory leaks stemming from `clearMsParam` (4-3-stable)

korydraughn:
The base commit of this PR has passed unit, core, and PREP tests. I've confirmed that the memory leaks for `rsExecRuleExpression` and `rsRuleExecSubmit` are fixed.

The second commit is temporary. It is there to make the polishing effort more visible - i.e. show that there are no changes in the behavior of the new functions.

Will squash once approved.

### DIFF
--- a/lib/core/include/irods/msParam.h
+++ b/lib/core/include/irods/msParam.h
@@ -130,11 +130,38 @@ extern "C" {
 
 int
 resetMsParam( msParam_t *msParam );
-int
-clearMsParam( msParam_t *msParam, int freeStruct );
-int
-addMsParam( msParamArray_t *msParamArray, const char *label,
-            const char *packInstruct, void *inOutStruct, bytesBuf_t *inpOutBuf );
+
+/// Deallocates memory referenced by members of a MsParam.
+///
+/// This function does not deallocate memory referenced by KeyValPairs.
+///
+/// \param[in] msParam    The pointer to a MsParam to deallocate.
+/// \param[in] freeStruct Indicates whether the memory pointed to by MsParam::inOutStruct should be deallocated.
+///
+/// \returns An integer indicating the status of the operation.
+/// \retval 0        On success.
+/// \retval non-zero On failure.
+int clearMsParam(msParam_t* msParam, int freeStruct);
+
+/// Deallocates memory referenced by members of a MsParam.
+///
+/// Unlike clearMsParam(MsParam*, int), this function deallocates memory referenced by all data members,
+/// including KeyValPairs.
+///
+/// \param[in] msParam The pointer to a MsParam to deallocate.
+///
+/// \returns An integer indicating the status of the operation.
+/// \retval 0        On success.
+/// \retval non-zero On failure.
+///
+/// \since 4.3.5
+int clearMsParamFull(msParam_t* _msParam);
+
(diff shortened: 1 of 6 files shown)


korydraughn:
Will address clang-tidy before this PR is approved.

server/api/src/rsRuleExecSubmit.cpp
```diff
@@ -47,7 +47,7 @@ namespace
         ruleExecInfoAndArg_t* rei_info{};
         irods::at_scope_exit free_rei_info{[&rei_info] {
             if (rei_info) {
-                freeRuleExecInfoStruct(rei_info->rei, (FREE_MS_PARAM | FREE_DOINP));
+                freeRuleExecInfoStructIncludingKeyValPair(rei_info->rei);
```
alanking: Looking at the implementations for `freeRuleExecInfoStruct` and `freeRuleExecInfoStructIncludingKeyValPair`... How is this invocation different?

lib/core/include/irods/msParam.h
```diff
@@ -130,9 +130,33 @@ extern "C" {
...
+///
+/// \returns An integer indicating the status of the operation.
+/// \retval 0        On success.
+/// \retval non-zero On failure.
+int clearMsParam(msParam_t* msParam, int freeStruct);
+
+/// Deallocates memory referenced by members of a MsParam.
+///
+/// This function performs the same operations as clearMsParam(MsParam*, int)
+/// and deallocates any memory owned by KeyValPairs.
```
alanking: Maybe mention that it behaves the same as `clearMsParam(MsParam*, <non-zero-value>)` (that is, that it also clears the `inOutStruct`)? Same for the other docs which make reference to the legacy functions.

alanking (commented):
One minor suggestion. Everything makes sense

server/re/src/reStruct.cpp
```diff
@@ -133,6 +135,56 @@ freeRuleExecInfoInternals( ruleExecInfo_t *rs, int freeSpecialStructFlag ) {
...
+    std::free(_rs);
+} // freeRuleExecInfoStructFull
+
+// NOLINTNEXTLINE(misc-no-recursion)
+void freeRuleExecInfoInternalsFull(ruleExecInfo_t* _rs)
+{
+    if (nullptr != _rs->msParamArray) {
+        clearMsParamArrayFull(_rs->msParamArray);
+        std::free(_rs->msParamArray);
+    }
```
MartinFlores751, line 153: If we're freeing something, we don't need `nullptr` check. The `std::free(...)` and `free(...)` functions do nothing on `NULL`/`nullptr`. Comment applies to all `nullptr` checks in calls to `free(...)`.

lib/core/include/irods/msParam.h
korydraughn: Updated docs.

The reference to the original function was to get doxygen to link to it so readers learned of its existence.

Let me know if the wording feels unclear.

server/re/src/reStruct.cpp
korydraughn: True.

Done.

korydraughn:
Squashed.

korydraughn:
Added pound.

PR #8866, opened 2026-02-23
Fix memory leaks in delay server and rule language (4-3-stable)

korydraughn:
All unit and core tests pass for ec0cc9f. This includes a full test run for the Python REP.

Commit 17351f5 was added after all testing was complete. This commit was tested at the bench. I don't think they require a rerun of the tests.


korydraughn:
Added clang-tidy nolint directives and squashed.

korydraughn:
Added pounds.

Issue #8868, opened 2026-02-23
Delay Server: Remove unnecessary use of `boost::container::pmr::unsynchronized_pool_resource`


korydraughn:
- [ ] main
- [ ] 4-3-stable
---
## Bug Report

The in-memory delay queue uses polymorphic memory resources to limit available memory. This is carried out through the use of `irods::fixed_buffer_resource` wrapped in `boost::unsynchronized_pool_resource`.

While this works, it's possible for the delay server to stop making progress due to the use of the pool resource. Given there's no benefit to keeping the pool resource, we should remove it. Its removal guarantees that the delay server always makes progress.

Issue #8859, opened 2026-02-18
`fixed_buffer_resource` will crash the delay server if the buffer is too small


korydraughn:
- [x] main
- [x] 4-3-stable
---
## Bug Report

While investigating #8575 and #8254, I discovered that attempting to set `maximum_size_of_delay_queue_in_bytes` in server_config.json to a very small value results in the delay server crashing at the following lines.

https://github.com/irods/irods/blob/5818a0fb18a3f352b1a5b2eb9f93ab816759cde9/lib/core/include/irods/fixed_buffer_resource.hpp#L83-L87

The `fixed_buffer_resource` needs to detect this and handle it gracefully by reporting it to the admin.

PR #8869, opened 2026-02-23
[#8859,#8868] Throw exception on small buffer / Remove use of pool memory resource (4-3-stable)

korydraughn:
Unit tests pass.

Will verify the changes work at the bench and then run the core tests.

### DIFF
--- a/server/core/include/irods/irods_delay_queue.hpp
+++ b/server/core/include/irods/irods_delay_queue.hpp
@@ -23,7 +23,7 @@ namespace irods
 {
     class delay_queue // NOLINT(cppcoreguidelines-special-member-functions)
     {
-    public:
+      public:
         explicit delay_queue(std::int64_t _pool_size_in_bytes)
         {
             namespace bpmr = boost::container::pmr;
@@ -53,8 +53,9 @@ namespace irods
         bool contains_rule_id(const std::string& _rule_id)
         {
             std::lock_guard rules_lock{rules_mutex_};
-            return std::any_of(queued_rules_->begin(), queued_rules_->end(),
-                       [&](const auto& q_rule) { return q_rule.data() == _rule_id; });
+            return std::any_of(queued_rules_->begin(), queued_rules_->end(), [&](const auto& q_rule) {
+                return q_rule.data() == _rule_id;
+            });
         }
 
         void enqueue_rule(const std::string& rule_id)
@@ -74,7 +75,7 @@ namespace irods
             }
         }
 
-    private:
+      private:
         std::mutex rules_mutex_;
 #if BOOST_VERSION >= 107200
         std::vector<std::byte> buffer_;
@@ -85,4 +86,3 @@ namespace irods
 } // namespace irods
 
 #endif // IRODS_DELAY_QUEUE_HPP
-


korydraughn:
I've confirmed that the delay server catches and reports the buffer-too-small case.

```
$ tail -f /var/log/irods/irods.log | jq .log_message
"rodsServer Release version rods4.3.4 - API Version d is up"
">>> control plane :: listening on port 1248\n"
"Forking Delay Server (irodsDelayServer) ..."
"Initializing delay server ..."
"Delay server exited with failure: fixed_buffer_resource: Memory alignment error; Buffer may be too small"
"Forking Delay Server (irodsDelayServer) ..."
"Initializing delay server ..."
"Delay server exited with failure: fixed_buffer_resource: Memory alignment error; Buffer may be too small"
"Forking Delay Server (irodsDelayServer) ..."
"Initializing delay server ..."
"Delay server exited with failure: fixed_buffer_resource: Buffer size [8] is too small"
"Forking Delay Server (irodsDelayServer) ..."
"Initializing delay server ..."
"Delay server exited with failure: fixed_buffer_resource: Buffer size [8] is too small"
"Forking Delay Server (irodsDelayServer) ..."
"Initializing delay server ..."
```

Running core tests.

server/core/include/irods/irods_delay_queue.hpp
```diff
@@ -80,16 +71,14 @@ namespace irods
             }
         }
 
-    private:
+      private:
         std::mutex rules_mutex_;
 #if BOOST_VERSION >= 107200
         std::vector<std::byte> buffer_;
 #endif // BOOST_VERSION >= 107200
-        std::unique_ptr<boost::container::pmr::memory_resource> upstream_allocator_;
```
alanking, line 88: Just wondering aloud if there's any value in leaving a note in the code or documentation somewhere to alert future readers to the fact that we have, in fact, thought about and tried using an `unsynchronized_pool_resource` for this and it should be avoided.

lib/core/include/irods/fixed_buffer_resource.hpp
```diff
@@ -77,10 +77,16 @@ namespace irods::experimental::pmr
...
             // Make sure the buffer is aligned for the header type.
             if (!std::align(alignof(header), sizeof(header_storage), buffer_, space_remaining)) {
-                throw std::runtime_error{"fixed_buffer_resource: internal memory alignment error. "};
+                throw std::runtime_error{"fixed_buffer_resource: Memory alignment error; Buffer may be too small"};
             }
 
-            headers_ = new (header_storage) header;
+            if (space_remaining <= sizeof(header)) {
+                throw std::runtime_error{
+                    fmt::format("fixed_buffer_resource: Buffer size [{}] is too small", buffer_size_)};
```
alanking, line 85: Any value in printing the minimum size anywhere? I suppose the developer can just go and look at the `struct` to figure it out.

server/core/include/irods/irods_delay_queue.hpp
korydraughn: Hard to say. Changing the code means someone has to open a PR and go through the code review process. So, we'd catch when someone attempts to introduce a pool resource. There's also the git history of the file.

That said, I'm not against leaving a comment. I'm just unsure if it's necessary.

lib/core/include/irods/fixed_buffer_resource.hpp
korydraughn: Hmm, if the admin wants to limit memory usage, they must set the value to something larger than zero. Setting the value to a very small value (e.g. 5 bytes) doesn't make a lot of sense because there wouldn't be any space left in the buffer to hold rule ids. The admin basically always needs to set the value to hundreds of bytes or more.

So, I don't know if there's a reasonable minimum to report to the admin.

Is there a value that comes to mind for you? Why?

korydraughn:
All core tests passed.

lib/core/include/irods/fixed_buffer_resource.hpp
alanking: Well, the minimum as shown seems to be the size of the `header` struct, plus 1. That, too, is pretty much useless, but just the fact that this check has to be here made me wonder whether we ought to at least put a recommendation or something. I guess the default value would be the recommendation, in some sense. Just trying to combat future email. Resolve if you feel there is nothing to do

server/core/include/irods/irods_delay_queue.hpp
alanking: Up to you. We (and our memory of this change) aren't guaranteed to be here for as long as the project lasts, and I'm guessing whoever comes up with such an idea will not look through the Git history of the file before implementing it. Since we don't leave around design documents, this is the next best thing.

korydraughn: Added comment.

korydraughn:
Squashed.

korydraughn:
2effa2e was missing a word, but I ended up rewriting the commit message so that the change is clear.

Added pounds.

PR #8870, opened 2026-02-24
[#8859,#8868] Throw exception on small buffer / Remove use of pool memory resource (main)

korydraughn:
Cherry-pick of #8869 with a slight tweak to irodsDelayServer.cpp.


korydraughn:
Added pounds.

Issue #8865, opened 2026-02-23
rootless Docker container for Openshift


tolondra:
### New Feature / Functionality

Dear iRODS Team, 

We're having some issues setting up a **rootless container image for OpenShift**.

Our problem basically is the following:

We try to have iRODS running in a container in Openshift, basically we create a Dockerfile with Ubuntu, then add iRODS with apt 
( irods-server irods-database-plugin-postgres irods-runtime irods-icommands ) 

We use an externally provided PostgreSQL from OpenShift, and in the Dockerfile, we set the necessary folders to have owner 1001:1001, which is irods:irods

The container is set to run as 1001:1001, so no root but a defined user. 

When deployed to OpenShift, we map a PVC to /var/lib/irods for persistent storage. 

The problem is that the folder is set to root:irods with 775 permissions when the pvc is mounted, and irods wants to change it, which it can’t. 
 
In principle, this should be writeable, but the installer checks that the folder is owned by irods and then fails. 

See the error message attached: 

```
irods@irods-development:~$ python3 /var/lib/irods/scripts/setup_irods.py --json_configuration_file /tmp/irods/irods_provider.json
Traceback (most recent call last):
  File "/var/lib/irods/scripts/setup_irods.py", line 540, in <module>
    sys.exit(main())
             ^^^^^^
  File "/var/lib/irods/scripts/setup_irods.py", line 526, in main
    setup_server(irods_config,
  File "/var/lib/irods/scripts/setup_irods.py", line 99, in setup_server
    setup_service_account(irods_config, irods_user, irods_group)
 File "/var/lib/irods/scripts/setup_irods.py", line 339, in setup_service_account
    os.lchown(root,
PermissionError: [Errno 1] Operation not permitted: '/var/lib/irods'
```

Have you ever tried to setup a  rootless container image to be used with OpenShift or Kubernetes? 

Thanks in advanbce for your feedback!

Kind regards,

Natalia Bedoya

SwooshyCueb:
At present, the setup script handles some permission stuff, but eventually this responsibility will be handled by the packaging. I'll make sure our next release has an exception handler for this, though. In the meantime, you might have to configure irods manually without using the setup script.

SwooshyCueb:
Quick note before I start working on this @tolondra:

[...]

Issue #7999, opened 2024-11-08
Use `getpwnam` and `getgrnam` one time and reuse values for later calls


korydraughn:
- [ ] main
- [ ] 4-3-stable
---
## Performance

Please investigate reusing the value of `pwd.getpwnam(irods_user).pw_uid` and `grp.getgrnam(irods_group).gr_gid` for all function calls in the following code snippet.

https://github.com/irods/irods/blob/759a039e73b93f0904344d72edbefca8b9df739a/scripts/setup_irods.py#L304-L321

PR #8872, opened 2026-02-24
[#8865] Do not fail setup if chown fails (main)

SwooshyCueb:
Addresses #8865 for main

Not yet tested

### DIFF
--- a/scripts/setup_irods.py
+++ b/scripts/setup_irods.py
@@ -3,6 +3,7 @@
 import argparse
 import os
 import sys
+from pathlib import Path
 
 import irods.setup_options
 
@@ -299,6 +300,13 @@ def get_irods_user_and_group(irods_config):
             input_filter=irods.lib.character_count_filter(minimum=1, field='iRODS Service Account Group'))
     return (irods_user, irods_group)
 
+def try_chown(path, uid, gid, *args, **kwargs):
+    l = logging.getLogger(__name__)
+    try:
+        os.chown(path, uid, gid, *args, **kwargs)
+    except Exception as e:
+        l.warning("Failed to set ownership of %s: %s", path, e)
+
 def setup_service_account(irods_config, irods_user, irods_group):
     l = logging.getLogger(__name__)
     l.info(irods.lib.get_header('Setting up the service account'))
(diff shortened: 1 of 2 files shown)


scripts/setup_irods.py
```diff
@@ -333,31 +341,29 @@ def setup_service_account(irods_config, irods_user, irods_group):
...
+    irods_uid = pwd.getpwnam(irods_user).pw_uid
+    irods_gid = grp.getgrnam(irods_group).gr_gid
+
     l.info('Setting owner of %s to %s:%s',
             irods_config.irods_directory, irods_user, irods_group)
     for (root, _, files) in os.walk(irods_config.irods_directory):
-        os.lchown(root,
-                pwd.getpwnam(irods_user).pw_uid,
-                grp.getgrnam(irods_group).gr_gid)
+        try_chown(root, irods_uid, irods_gid, follow_symlinks=False)
```
trel, line 350: just surfacing the difference between lchown and chown...

korydraughn:
This PR might also cover #7999.

scripts/setup_irods.py
```diff
@@ -299,6 +300,13 @@ def get_irods_user_and_group(irods_config):
             input_filter=irods.lib.character_count_filter(minimum=1, field='iRODS Service Account Group'))
     return (irods_user, irods_group)
 
+def try_chown(path, uid, gid, *args, **kwargs):
+    l = logging.getLogger(__name__)
+    try:
+        os.chown(path, uid, gid, *args, **kwargs)
+    except Exception as e:
+        l.warning("Failed to set ownership of %s: %s", path, e)
```
korydraughn, line 308: This will allow setup to continue and possibly fail at a later point. Unless the admin is checking the output, they could miss the failure.

Seems like this behavior should be an opt-in until we provide examples/documentation on how to put together a rootless container.

scripts/setup_irods.py
SwooshyCueb: Since python 3.3, `lchown` is equivalent to `chown` with `follow_symlinks=False`

trel: perfect.  resolving.

Issue #8874, opened 2026-02-26
Replace usage of `/usr/bin/find` in `Test_ICommands_File_Operations_1.test_irm_rf_nested_coll` with `os.walk` or `Path.walk`.


SwooshyCueb:
- [x] 4-3-stable
- [x] main

-------
`test_irm_rf_nested_coll` tries to call `find` here:
https://github.com/irods/irods/blob/72c2725d00780c6a430520b04914d41cc39949ed/scripts/irods/test/test_icommands_file_operations.py#L209

This would necessitate a package dependency on `findutils` on EL. We shouldn't be calling out to do this anyway, though. Let's replace this with `os.walk` or `Path.walk` instead.

PR #8876, opened 2026-02-26
[#8874] Replace call to find with os.walk in test_irm_rf_nested_coll (main)

SwooshyCueb:
Addresses #8874 for main


SwooshyCueb:
\#'d

Issue #8789, opened 2026-01-07
Add password_reuse_previous option for pam generated passwords


peterverraedt:
This issue proposes to add an option to tweak the behaviour when generating a native password in case pam authentication is used (pam_interactive, pam_password).

Suppose a user has two devices (work-laptop and hpc-cluster) and authenticates on Monday on hpc-cluster with pam_interactive and a ttl of 7 days. On Wednesday the user authenticates on work-laptop with a ttl of 1 day.

With the default settings (password_extend_lifetime = 1), the session on hpc-cluster will terminate on Thursday instead of next Monday (what the user would expect).

With the option password_extend_lifetime set to 0, the user will get a session on work-laptop that is actually valid until next Monday instead of the expected TTL of 1 day.

To provide multiple concurrent sessions on different devices that possibly have different TTLs and certainly have different expiration dates, we propose to add an option "password_reuse_previous" that, if set to false, disables reusing previously generated temporary native passwords altogether. The default value would be true, and this would keep the current behaviour, regardless password_extend_lifetime is set to 0 or 1.

While pam authentication might be coupled with the new session tokens #8746 in a later phase, this change could be useful in production systems right now.

alanking:
@peterverraedt - Thanks for implementing this! We can take care of the documentation before the release, unless you wanted to take a swing at it, of course.

Issue #8769, opened 2025-12-06
`R_USER_CREDENTIALS` not created on upgrade


alanking:
### iRODS Version, OS and Version

5.0.90

Need to add `R_USER_CREDENTIALS` table on upgrade.

See https://github.com/irods/irods/issues/8697#issuecomment-3618952832

Issue #6469, opened 2022-07-11
irsync -r fails to sync two collections with false SYS_COPY_LEN_ERR


tedgin:
### iRODS Version, OS and Version

iRODS 4.2.8, CentOS 7

### What did you try to do?

I attempted to use irsync to copy a collection.  For example, `irsync -r i:/iplant/home/tedgin/analyses i:/iplant/home/tedgin/analyses2`.

### Expected behavior

I expected a new collection to be created with the same contents as the first.

### Observed behavior (including steps to reproduce, if applicable)

irsync fails part way through with a `SYS_COPY_LEN_ERR` error.

```console
tedgin@prod ~? irsync -r i:/iplant/home/tedgin/analyses i:/iplant/home/tedgin/analyses2
remote addresses: 206.207.252.46 ERROR: rsyncDataToDataUtil - Failed to copy the object "/iplant/home/tedgin/analyses/ten-rules-jupyter_analysis1-2020-06-25-18-13-14.1/ten-rules-jupyter/.git/objects/pack/pack-252c1ba0d70e82c5201c3fe9e9f672eae85dcda2.pack" to "/iplant/home/tedgin/analyses2/ten-rules-jupyter_analysis1-2020-06-25-18-13-14.1/ten-rules-jupyter/.git/objects/pack/pack-252c1ba0d70e82c5201c3fe9e9f672eae85dcda2.pack" SYS_COPY_LEN_ERR 
remote addresses: 206.207.252.46 ERROR: rsyncCollToCollUtil: rsyncDataToDataUtil failed for /iplant/home/tedgin/analyses2/ten-rules-jupyter_analysis1-2020-06-25-18-13-14.1/ten-rules-jupyter/.git/objects/pack/pack-252c1ba0d70e82c5201c3fe9e9f672eae85dcda2.pack.stat=-27000 status = -27000 SYS_COPY_LEN_ERR
remote addresses: 206.207.252.46 ERROR: rsyncUtil: rsync error for /iplant/home/tedgin/analyses2 status = -27000 SYS_COPY_LEN_ERR
```

Here's some details about this file that failed.

```console
tedgin@prod ~? ils -L /iplant/home/tedgin/analyses/ten-rules-jupyter_analysis1-2020-06-25-18-13-14.1/ten-rules-jupyter/.git/objects/pack/pack-252c1ba0d70e82c5201c3fe9e9f672eae85dcda2.pack
[...]

PR #8879, opened 2026-03-06
Add new auth table, add irsync tests, add iadmin help text (main)

alanking:
In service of #6469
Addresses #8769
In service of #8789

Upgrades tested:
- [x] Postgres 16
- [x] MySQL 8.4

Passing tests:
- [x] unit tests
- [x] core tests
- [x] topology-on-consumer (no SSL - just the new test)
- [x] topology-on-consumer (SSL - just the new test)

### DIFF
--- a/plugins/database/src/icatSysTables.sql.pp.in
+++ b/plugins/database/src/icatSysTables.sql.pp.in
@@ -154,9 +154,6 @@ create table R_TICKET_ALLOWED_GROUPS ( ticket_id INT64TYPE not null, group_name 
 
 create table R_GRID_CONFIGURATION ( namespace varchar(2700), option_name varchar(2700), option_value varchar(2700));
 
-/* Table for storing password hashes. */
-create table R_USER_CREDENTIALS ( user_id INT64TYPE, hashed_password varchar(2700), salt varchar(32), hashing_algorithm varchar(250), hashing_parameters varchar(2700), create_time varchar(32), modify_time varchar(32), expiration_time varchar(32));
-
 #ifdef mysql
 
 /* The max size for a MySQL InnoDB index prefix is 767 bytes,
--- a/scripts/irods/database_upgrade.py
+++ b/scripts/irods/database_upgrade.py
@@ -252,6 +252,39 @@ def run_update(irods_config, cursor, is_upgrade):
 
         # pam password reuse setting
         database_connect.execute_sql_statement(cursor, "insert into R_GRID_CONFIGURATION values ('authentication', 'password_reuse_previous', '1');")
+
+        # Add R_USER_CREDENTIALS table.
+        if "oracle" == irods_config.catalog_database_type:
+            database_connect.execute_sql_statement(
+                cursor,
+                "create table R_USER_CREDENTIALS"
+                    "("
+                        "user_id "           "integer, "
+                        "hashed_password "   "varchar(2700), "
+                        "salt "              "varchar(32), "
+                        "hashing_algorithm " "varchar(250), "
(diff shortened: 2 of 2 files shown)


scripts/irods/test/resource_suite.py
```diff
@@ -1319,3 +1319,53 @@ def test_iget_data_object_as_user_with_null_access_and_replica_only_in_archive__
...
+    def test_irsync_data_object_with_various_sizes_always_works__issue_6469(self):
+        filename = "test_irsync_data_object_to_new_data_object__issue_6469"
+        try:
+            # Tests a zero-length file, an 8MiB file (streaming / single buffer), and a 128MiB file (parallel transfer).
+            for size in [0, 8 * 1024 ** 2, 128 * 1024 ** 2]:
+                localfile_name = filename + f"_{size}"
+                source_logical_path = "/".join([self.user1.session_collection, localfile_name])
+                dest_logical_path = source_logical_path + ".bak"
+                localfile = os.path.join(self.user1.local_session_dir, localfile_name)
+                contents = str() if size > 0 else "more"
```
korydraughn, line 1333: Is `str()` doing something special?

This isn't a request to change anything. Just trying to determine if this is presenting something new to me.

scripts/irods/test/resource_suite.py
```diff
@@ -1319,3 +1319,53 @@ def test_iget_data_object_as_user_with_null_access_and_replica_only_in_archive__
...
+                    self.assertEqual(source_object_size, dest_object_size)
+
+                    # Rsync the object to a local file (object-to-file sync).
+                    self.user1.assert_icommand(["irsync", f"i:{dest_logical_path}", localfile])
+                    file_size = os.stat(localfile).st_size
+                    self.assertEqual(dest_object_size, file_size)
+
+                    # Rsync the object to a local file (object-to-file create).
+                    self.user1.assert_icommand(["irsync", f"i:{dest_logical_path}", localfile + "_new"])
+                    new_file_size = os.stat(localfile + "_new").st_size
```
korydraughn: Can assign `localfile + "_new"` to a local variable to protect against future modifications.

scripts/irods/test/resource_suite.py
korydraughn, line 1333: How you're managing `contents` is a neat trick. Took me a second to understand it, but I think I got it.

Correct me if I'm wrong, but `contents` is designed to guarantee that the size of the files/data-objects are different at a particular point in each test case, correct?

alanking: `str()` is equivalent to `""`, or, an empty string.

alanking: Right, it's making the data object a non-zero-length object in the zero-length case and a zero-length object in the non-zero-length case.

Admittedly, this excludes some test cases from the table here: https://github.com/irods/irods/issues/6469#issuecomment-3811614685 However, this covers the case outlined by OP in #6469, so, I figured it's better to have the test than not.

scripts/irods/test/resource_suite.py
alanking: Done

scripts/irods/test/resource_suite.py
korydraughn: Okay. Then I was interpreting it correctly.

Thanks.

korydraughn: No objections here.

korydraughn:
Squash to taste if there's nothing left to do.

alanking:
Squashed

alanking:
#'d, mergin

PR #8883, opened 2026-03-13
[#8861] Fix memory leak in Hasher (main)

MartinFlores751:
Issue quick link: #8861.

Hasher now has a destructor that cleans up the context used for the given hash strategy.

### DIFF
--- a/lib/hasher/include/irods/Hasher.hpp
+++ b/lib/hasher/include/irods/Hasher.hpp
@@ -22,7 +22,12 @@ namespace irods {
     class Hasher {
         public:
             Hasher() : _strategy( NULL ) {}
-
+            ~Hasher()
+            {
+                if (nullptr != _strategy) {
+                    _strategy->free_context(_context);
+                }
+            }
             error init( const HashStrategy* );
             error update( const std::string& );
             error digest( std::string& messageDigest );
--- a/lib/hasher/include/irods/ADLER32Strategy.hpp
+++ b/lib/hasher/include/irods/ADLER32Strategy.hpp
@@ -22,6 +22,7 @@ namespace irods {
             error update( const std::string& data, boost::any& context ) const override;
             error digest( std::string& messageDigest, boost::any& context ) const override;
             bool isChecksum( const std::string& ) const override;
+            void free_context(boost::any& _context) const override;
 
             /// Produce a digest in string form based on the provided options.
             ///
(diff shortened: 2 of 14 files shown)


lib/hasher/include/irods/Hasher.hpp
```diff
@@ -15,7 +15,12 @@ namespace irods {
     class Hasher {
         public:
             Hasher() : _strategy( NULL ) {}
-
+            ~Hasher()
```
korydraughn, line 25: Must be marked `virtual`.

korydraughn (commented):
I don't think we can make this change since it likely affects ABI.

This change may need to wait until iRODS 6.

MartinFlores751:
> I don't think we can make this change since it likely affects ABI.
> 
> This change may need to wait until iRODS 6.

I figured. The alternative is to call `update(...)` or `digest(...)` as a form of RAII, since they also do cleanup. I forget the specifics of when they clean up though, and that solution is a bit hacky.

korydraughn:
You might be able to resolve the leak in an ABI friendly manner by using free functions and making them friends of their target hasher. Or, provide a free function overloaded on the type and lean on ADL.

lib/hasher/include/irods/Hasher.hpp
MartinFlores751: After a small discussion with @korydraughn, it was decided that this class does not need `virtual` since it is not inherited from.

alanking:
After a discussion this morning, we've decided that ABI compatibility is only maintained within a given minor release. I think we can move forward with this in 5.1.0 in light of that.

MartinFlores751:
Rebased onto tip of main.

alanking (commented):
Seems good to me. Please confirm that leak is still fixed after rebase. And perhaps run the tests?

korydraughn:
I second what @alanking said.

MartinFlores751:
> Seems good to me. Please confirm that leak is still fixed after rebase. And perhaps run the tests?

Reran the core and unit tests on Debian 13 w/ Postgres 16. All tests passed.

MartinFlores751:
Running the policy composition tests shows the leak is resolved with this PR. The only leaks that show up in testing are from the icommands.

korydraughn (commented):
Squash to taste.

If you haven't already, consider capturing the story behind the change in the commit message body.

MartinFlores751:
Squashed.

korydraughn (approved):
Pound it.

MartinFlores751:
Pounded.

Issue #8424, opened 2025-04-23
Deprecate API endpoints related to legacy authentication


alanking:
- [ ] main

---

The APIs are listed in the docs: https://github.com/irods/irods_docs/blob/main/docs/plugins/pluggable_authentication.md#how-to-know-if-a-client-is-using-legacy-authentication-plugins Just in case:

-    AUTH_REQUEST_AN (703)
-    AUTH_RESPONSE_AN (704)
-    PAM_AUTH_REQUEST_AN (725)
-    AUTH_PLUG_REQ_AN (1201)
-    AUTH_PLUG_RESP_AN (1202)

An extension of the work in https://github.com/irods/irods/issues/8257.

PR #8920, opened 2026-04-09
[#8424] Deprecate legacy authentication API numbers (main)

korydraughn:


alanking (commented):
`rc_authenticate_client` is not an really API endpoint, but rather a C wrapper for the `irods::authentication::authenticate_client` C++ function as an alternative to `clientLogin`.

I think technically the alternative API endpoint for all of these is `AUTHENTICATION_APN` (110000), but this is for use in the client authentication frameworks. Would that be a better suggested alternative?

korydraughn:
Sure. Will update.

korydraughn:
Clang-Tidy still looked at the file.

Will remove that commit before squashing.

alanking (commented):
Looks good. Squash

korydraughn:
Squashed.

alanking (approved):
\# it

korydraughn:
Added pound.

Issue #4770, opened 2020-03-05
Add method for checking remote vaults


alanking:
## Feature

Need a tool for testing which essentially runs `iscan` on a remote physical path. For a given replica in a distributed system, the following scenarios need to be tested:

- If we expect a data object to be created, assert that file is created on disk and registered in the catalog
- If we expect a data object not to be created, assert that the file is NOT created on disk and also not registered in the catalog

There is no currently supported way in our test toolkit to reach out to a remote disk.

PR #8923, opened 2026-04-09
[8450] Added irm and itrim tests

JustinKyleJames:
These test every combination of these actions for all permission levels on a collection and data object.

Notes:

1. At this time this only tests removal of single data objects - no recursive removals.
2. The permissions are only updated on the data object's parent collection.
3. For the 'itrim' test, I modeled it after the 'irm' tests but since there is only one test case, I used the existing 'Test_Irm' class.  All of the setup and tear-down is in the test itself.
4. The tests look for exactly what is being done in the code today.  It makes sense to me.
5. For the itrim, I kept the SIL option (trim fails without error) for future use although that doesn't happen today.  The equivalent cases are DNE for an error reporting the replica does not exist.

### DIFF
--- a/scripts/irods/test/test_irm.py
+++ b/scripts/irods/test/test_irm.py
@@ -109,6 +109,7 @@ class Test_Irm(session.make_sessions_mixin([('otherrods', 'rods')], [('alice', '
 
 @unittest.skipIf(test.settings.RUN_IN_TOPOLOGY, "Checks local resource vault.")
 class test_no_accidental_data_loss_on_unlink__issue_8441(unittest.TestCase):
+
     @classmethod
     def setUpClass(self):
         self.admin = session.mkuser_and_return_session("rodsadmin", "otherrods", "rods", lib.get_hostname())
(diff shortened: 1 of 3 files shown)


scripts/irods/test/test_irm.py
```diff
@@ -174,6 +174,98 @@ def assert_unlink_fails(self, owner, executor, collection_permission, object_per
...
+            'delete_metadata',
+            'create_object',
+            'modify_object',
+            'delete_object',
+            'own',
+        ]
+
+        DEL = 'DEL'  # file deleted
+        SIL = 'SIL'  # file not deleted silently
+        ERR = 'ERR'  # file not deleted with error
```
alanking: Maybe this could be turned into an `enum` instead of strings?

scripts/irods/test/test_irm.py
```diff
@@ -174,6 +174,98 @@ def assert_unlink_fails(self, owner, executor, collection_permission, object_per
         self.assertTrue(lib.replica_exists_on_resource(owner, object_path, self.resource),
                         msg=f"Replica for [{object_path}] was unregistered unexpectedly.")
 
+    def test_irm_force_unlink_all_permission_combinations__issue_8450(self):
```
alanking, line 178: Do we need to skip this test when running in topology as well?

scripts/irods/test/test_irm.py
```diff
@@ -174,6 +174,98 @@ def assert_unlink_fails(self, owner, executor, collection_permission, object_per
...
+                        self.assertEqual(expected[coll_perm][i], outcome,
+                            msg=f"executor={executor.username} coll_perm={coll_perm} obj_perm={obj_perm}: "
+                                f"expected {expected[coll_perm][i]}, got {outcome}")
+                        if outcome == ERR:
+                            self.assertIn("CAT_NO_ACCESS_PERMISSION", err,
+                                msg=f"executor={executor.username} coll_perm={coll_perm} obj_perm={obj_perm}: "
+                                    f"expected CAT_NO_ACCESS_PERMISSION in stderr: [{err}]")
+
+                        # Restore the owner's own permission so subsequent iterations work correctly.
+                        owner.run_icommand(["ichmod", "-M", "own", owner.username, object_path])
```
alanking: Does this need to be in a `finally` block?

scripts/irods/test/test_irm.py
```diff
@@ -174,6 +174,98 @@ def assert_unlink_fails(self, owner, executor, collection_permission, object_per
...
+
+            for coll_perm in all_permissions:
+                for i, obj_perm in enumerate(all_permissions):
+                    with self.subTest(executor=executor.username, coll_perm=coll_perm, obj_perm=obj_perm):
+                        owner.assert_icommand(["ichmod", coll_perm, executor.username, collection_path])
+                        owner.assert_icommand(["ichmod", obj_perm,  executor.username, object_path])
+
+                        out, err, rc = executor.run_icommand(["irm", "-f", object_path])
+
+                        file_gone = not os.path.exists(physical_path)
```
alanking: Would it be better to use `iscan` (or `ifsck`) here?

scripts/irods/test/test_itrim.py
```diff
@@ -374,6 +375,143 @@ def test_itrim_minimum_replicas_to_keep_invalid_values__issue_7502(self):
...
+                'read_metadata',
+                'read_object',
+                'create_metadata',
+                'modify_metadata',
+                'delete_metadata',
+                'create_object',
+                'modify_object',
+                'delete_object',
+                'own',
+            ]
```
alanking: Might be interesting to capture this in `lib.py` or something since this is useful.

scripts/irods/test/test_itrim.py
```diff
@@ -374,6 +375,143 @@ def test_itrim_minimum_replicas_to_keep_invalid_values__issue_7502(self):
...
+
+            DEL = 'DEL'  # replica physically removed
+            SIL = 'SIL'  # replica not removed silently (rc=0, no error)
+            ERR = 'ERR'  # replica not removed, CAT_NO_ACCESS_PERMISSION error
+            DNE = 'DNE'  # replica not removed, object reported as not existing
```
alanking: I think it may help to clarify that the replica was not removed *because* the object was reported as not existing. The first time I read this, I thought that this meant that the replica was not removed and then the object was reported as not existing *after* the attempted trim (if that distinction makes sense).

alanking (commented):
Really like these tables. Just had some improvement suggestions but the approach and implementation are solid

scripts/irods/test/test_irm.py
JustinKyleJames: done

scripts/irods/test/test_irm.py
JustinKyleJames: @alanking - the class itself has the skip if in topology.  Wouldn't that force the test itself to be skipped?

scripts/irods/test/test_irm.py
JustinKyleJames: done

scripts/irods/test/test_irm.py
alanking: Ah, missed that. Yes, that would skip all the test cases in the class. Resolving

scripts/irods/test/test_irm.py
JustinKyleJames: @alanking, those seem to be heavy handed for just checking one file.  What is your thought?

scripts/irods/test/test_itrim.py
JustinKyleJames: Well, I was intending to say something a little different.  The user is informed that the object does not exist.  I changed the wording a bit.

`replica not removed and the reason given is that the object does not exist`

scripts/irods/test/test_irm.py
alanking: Mostly thinking about the fact that we are bound to a single machine in this case. I guess we will be implementing a way out of this limitation in https://github.com/irods/irods/issues/4770. Definitely out of scope for this. Resolving

scripts/irods/test/test_itrim.py
JustinKyleJames: done

alanking (commented):
Looks good. Squash

JustinKyleJames:
> Looks good. Squash

done

alanking (approved):
\# it

JustinKyleJames:
Pound is done

korydraughn:
And the new tests pass?

JustinKyleJames:
> And the new tests pass?

Yes

Issue #6111, opened 2022-01-15
rebalance should continue on data object that has since been moved out-of-tree


trel:
## Enhancement

If a rebalance is started, and a (long) list of (possibly large) data objects is found to be in need of rebalancing (replication)... and due to the unavoidable time-of-check-time-of-use (TOCTOU) window before a particular data object's time has come to be replicated...

It could have been moved out-of-tree, and no longer have a good replica to serve as the source replica for a replication operation.

---

I think this means that line 331...

https://github.com/irods/irods/blob/9c57ce9995501adb3bd439519737168858e08670/plugins/resources/replication/irods_repl_rebalance.cpp#L330-L331

Should be wrapped in a try/catch, handle a thrown `CAT_NO_ROWS_FOUND`, and quietly continue to the next `data_id_to_replicate`.

cc @kript

trel:
This was diagnosed from the following 4.2.7 rodsLog:

```
Jan 11 16:40:44 pid:13643 remote addresses: 127.0.0.1, 172.17.154.47 ERROR: iRODS Exception:
    file: /irods_source/plugins/resources/replication/irods_repl_rebalance.cpp
    function: (anonymous namespace)::ReplicationSourceInfo (anonymous namespace)::get_source_data_object_attributes(rsComm_t *, const rodsLong_t, const std::vector<leaf_bundle_t> &)
    line: 120
    code: -808000 (CAT_NO_ROWS_FOUND)
    message:
        rsGenQuery failed. genquery_inp contents:
maxRows: 256 continueInx: 0 rowOffset: 0
options: 0
selectInp.len: 4
    column: 403 DATA_NAME
    options: 1
    column: 501 COLL_NAME
    options: 1
    column: 421 DATA_MODE
    options: 1
    column: 423 DATA_RESC_ID
    options: 1
sqlCondInp.len: 3
    column: 401 DATA_ID
    condition: = '7259082'
    column: 423 DATA_RESC_ID
[...]

Issue #7984, opened 2024-10-23
allow rebalance to continue for data objects that do not replicate


nief:
hello,

on one of our zone, we have the following simple resource tree:

```
doublecopy:replication
├── disk5:unixfilesystem
└── disk6:unixfilesystem
```
disk5 contains 10 millions files and disk6 has been recently added, hence rebalance of a lot of files has to be done.

So I triggered a rebalance:
```
$ iadmin modresc doublecopy rebalance
remote addresses: 134.158.209.137 ERROR: rcGeneralAdmin failed with error -808000 CAT_NO_ROWS_FOUND
```
I obtained this error rather quickly, see the consumer log containing the following:
```
condition: = '1'\n\n\n\n possible iquest [select DATA_NAME, COLL_NAME, DATA_MODE, DATA_RESC_ID where DATA_ID = '9271259' and DATA_RESC_ID IN ('9271039','9271973') and DATA_REPL_STATUS = '1']\n        failed to rebalance resource: [-]\t/irods_source/server/api/src/rsGeneralAdmin.cpp:1104:int _rsGeneralAdmin(rsComm_t *, generalAdminInp_t *) :  status [CAT_NO_ROWS_FOUND]  
```
The object has indeed a bad repl status and thus cannot be replicated and the request returns `CAT_NO_ROWS_FOUND`.

It makes perfect sense. However the entire process stop just because of this stalled file. 

Is it the expected behavior ? It's because that on millions of files, having some bad files like this may happen for various reasons and the rest of the rebalancing operation could proceed without being interrupted ?

The environment is a mixture of CentOS7 (providers and consumers) and RHEL9 (1 consumer).

thanks,
JY

trel:
We designed rebalance to stop, like this, so there would be no surprises later for the administrator.

We might be able to revisit this logic (aka stale vs good) since many other things in iRODS are better since we initially released `rebalance`... but partial success is very hard to communicate and without stronger guarantees, we wanted to make sure the administrator knew things weren't rebalancing as intended/instructed.

trel:
You an also query ahead of time for data objects with no good replicas and make sure they are handled before the rebalance operation is called.

Potentially, we could even do that in the rebalance call BEFORE starting the rebalance to give a full report to the administrator of things that *would have failed* to replicate.

nief:
Querying in advance and clean up the mess is a good thing in theory but in practice is not that easy. These bad files have to be checked by the owner first to see what happened and to correct or upload them again if needed. 
[...]

PR #8930, opened 2026-04-15
[#6111] Allow rebalance to support partial success (main)

MartinFlores751:
Issue quick link: #6111 

This PR allows for rebalance to progress when encountered with stale replicas that cannot be updated and missing replicas that cannot be populated.

A new error code `REBALANCE_NOT_COMPLETE -1834000` has been created to signal to the user that the rebalance was executed, but manual intervention is required.

---

Before getting this PR out of draft:
- [x] Test on MySQL
- [x] Test on MariaDB
- [x] Remove `rodsLog(...)` calls used for debugging
  - If logging is desired, replace with appropriate function
- [x] Run git-clang-format
- [x] Check on clang-tidy
- [x] Address existing draft PR comments


server/icat/src/icatHighLevelRoutines.cpp
```diff
@@ -4523,6 +4536,68 @@ int chlGetReplListForLeafBundles(
...
+                  _child_idx,
+                  _bundles,
+                  _invocation_timestamp,
+                  _results,
+                  _offset );
+    if (!ret.ok()) {
+        irods::log(PASS(ret));
+    }
+    return ret.code();
+
```
korydraughn, line 4598: Remove empty line.

server/icat/include/irods/icatHighLevelRoutines.hpp
```diff
@@ -229,6 +229,14 @@ int chlGetReplListForLeafBundles(
     const std::string*          _invocation_timestamp,
     dist_child_result_t*        _results );
 
+int chlGetReplListForLeafBundlesOffset(
```
korydraughn, line 232: Add documentation.

plugins/database/src/db_plugin.cpp
```diff
@@ -13647,6 +13645,117 @@ irods::error db_get_repl_list_for_leaf_bundles_op(
...
+
+    // =-=-=-=-=-=-=-
+    // check the context
+    irods::error ret = _ctx.valid();
+    if ( !ret.ok() ) {
+        return PASS( ret );
+    }
+
+    if (_count <= 0) {
+        return ERROR(SYS_INVALID_INPUT_PARAM, boost::format("invalid _count [%d]") % _count);
```
korydraughn, line 13665: Replace `boost::format` with `fmt::format`.

plugins/database/src/db_plugin.cpp
```diff
@@ -13647,6 +13645,117 @@ irods::error db_get_repl_list_for_leaf_bundles_op(
...
+    }
+    if (_bundles->empty()) {
```
korydraughn, line 13667: Missing nullptr check.

plugins/database/src/db_plugin.cpp
```diff
@@ -13647,6 +13645,117 @@ irods::error db_get_repl_list_for_leaf_bundles_op(
...
+        return ERROR(SYS_INVALID_INPUT_PARAM, "no bundles");
+    }
+    if (!_invocation_timestamp) {
+        return ERROR(SYS_INTERNAL_NULL_INPUT_ERR, "invocation timestamp is NULL");
```
korydraughn, line 13671: Use `INVALID_INPUT_ARGUMENT_NULL_POINTER` for null pointer errors.
```suggestion
        return ERROR(INVALID_INPUT_ARGUMENT_NULL_POINTER, "invocation timestamp is NULL");
```

plugins/database/src/db_plugin.cpp
```diff
@@ -13647,6 +13645,117 @@ irods::error db_get_repl_list_for_leaf_bundles_op(
...
+        "select distinct data_id from R_DATA_MAIN "
+        "  where resc_id in (%s) and data_id not in ( "
+        "    select data_id from R_DATA_MAIN "
+        "      where resc_id in (%s) "
+        "  ) and modify_ts <= '%s' "
+        "order by data_id "
+        "limit %d,18446744073709551615") % not_child_array % child_array % _invocation_timestamp->c_str() % _offset).str();
+#else
+    /* Postgres */
+    const std::string query = (boost::format(
```
korydraughn, line 13718: Use fmtlib.

plugins/database/src/db_plugin.cpp
```diff
@@ -13647,6 +13645,117 @@ irods::error db_get_repl_list_for_leaf_bundles_op(
...
+    _results->push_back(atoll(icss.stmtPtr[statement_num]->resultValue[0]));
+
+    for (rodsLong_t i=1; i<_count; ++i) {
+        const int status_cmlGetNextRowFromStatement = cmlGetNextRowFromStatement(statement_num, &icss);
+        if (status_cmlGetNextRowFromStatement == CAT_NO_ROWS_FOUND) {
+            break;
+        }
+        if (status_cmlGetNextRowFromStatement != 0) {
+            cmlFreeStatement(statement_num, &icss);
+            return ERROR(status_cmlGetNextRowFromStatement, boost::format("failed to get row [%d] from query [%s]") % i % query);
```
korydraughn, line 13750: Use fmtlib.

plugins/database/src/db_plugin.cpp
```diff
@@ -13647,6 +13645,117 @@ irods::error db_get_repl_list_for_leaf_bundles_op(
...
+        }
+        _results->push_back(atoll(icss.stmtPtr[statement_num]->resultValue[0]));
+    }
+    cmlFreeStatement(statement_num, &icss);
+    return SUCCESS();
+
```
korydraughn, line 13756: Remove empty line.

plugins/resources/replication/src/irods_repl_rebalance.cpp
```diff
@@ -302,13 +308,14 @@ namespace {
...
-        const size_t                      _bun_idx,
-        const std::vector<leaf_bundle_t>& _bundles,
-        const dist_child_result_t&        _data_ids_to_replicate) {
+    void proc_results_for_rebalance(irods::plugin_context& _ctx,
+                                    const std::string& _parent_resc_name,
+                                    const std::string& _child_resc_name,
+                                    const size_t _bun_idx,
+                                    const std::vector<leaf_bundle_t>& _bundles,
+                                    const dist_child_result_t& _data_ids_to_replicate,
+                                int& _num_repls_to_skip)
```
trel, line 317: indentation

scripts/irods/test/test_iadmin.py
```diff
@@ -509,6 +495,121 @@ def test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380(
             self.admin.assert_icommand("iadmin rmuser %s" % user_name)
             self.admin.assert_icommand("iadmin rmgroup %s" % group_name)
 
+
+    def test_rebalance_under_funny_case__isssue_6111(self):
```
trel, line 499:
```suggestion
    def test_rebalance_under_funny_case__issue_6111(self):
```

scripts/irods/test/test_iadmin.py
```diff
@@ -509,6 +495,121 @@ def test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380(
...
+        output = subprocess.getstatusoutput("hostname")
+        hostname = output[1]
+
+        # Don't use the default rebalance limit
+        # use a smaller limit to...
```
trel, line 504: make this a sentence...

scripts/irods/test/test_iadmin.py
```diff
@@ -509,6 +495,121 @@ def test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380(
...
+
+                # =-=-=-=-=-=-=-
+                # call rebalance function - the thing were actually testing... finally.
+                self.admin.assert_icommand("iadmin modresc pt rebalance", "STDERR", "-1834000 REBALANCE_NOT_COMPLETE", desired_rc=4)
+
+                # =-=-=-=-=-=-=-
+                # visualize our rebalance
+                self.admin.assert_icommand("ils -AL", 'STDOUT_SINGLELINE', "foo")
+
+                # TODO: Check for the absence of '3'
```
trel, line 575: TODO

plugins/resources/replication/src/irods_repl_rebalance.cpp
```diff
@@ -331,7 +338,18 @@ namespace {
...
-            const ReplicationSourceInfo source_info = get_source_data_object_attributes(_ctx.comm(), data_id_to_replicate, _bundles);
+            ReplicationSourceInfo source_info;
+            try {
+                source_info = get_source_data_object_attributes(_ctx.comm(), data_id_to_replicate, _bundles);
+            }
+            catch (const irods::exception& _e) {
+                if (_e.code() != CAT_NO_ROWS_FOUND) {
+                     throw;
+                }
+                rodsLog(LOG_WARNING, "Cannot perform rebalance on id [%lld]. Skipping.", data_id_to_replicate);
```
korydraughn, line 349: Use the new logger. The `resource` category seems like a good fit.

server/icat/include/irods/icatHighLevelRoutines.hpp
trel, line 232: this is titled 'offset'... the database ops above are named 'index'...  is this difference intentional?  if so, please explain.

plugins/resources/replication/src/irods_repl_rebalance.cpp
```diff
@@ -446,7 +470,20 @@ namespace irods {
...
+                ReplicationSourceInfo source_info;
+                try {
+                    source_info =
+                        get_source_data_object_attributes(_ctx.comm(), replica_to_update.data_id, _leaf_bundles);
+                }
+                catch (const irods::exception& _e) {
+                    if (_e.code() != CAT_NO_ROWS_FOUND) {
+                        throw;
+                    }
+                    rodsLog(LOG_WARNING, "Cannot update replica [%lld] for data_id [%lld]. Skipping.", replica_to_update.replica_number, replica_to_update.data_id);
```
korydraughn, line 482: Use the new logger.

plugins/resources/replication/src/irods_repl_rebalance.cpp
```diff
@@ -331,7 +338,18 @@ namespace {
...
+                source_info = get_source_data_object_attributes(_ctx.comm(), data_id_to_replicate, _bundles);
+            }
+            catch (const irods::exception& _e) {
+                if (_e.code() != CAT_NO_ROWS_FOUND) {
+                     throw;
+                }
+                rodsLog(LOG_WARNING, "Cannot perform rebalance on id [%lld]. Skipping.", data_id_to_replicate);
+                _num_repls_to_skip++;
+                continue;
+            }
```
korydraughn, line 352: Should we handle `std::exception`?

plugins/resources/replication/src/irods_repl_rebalance.cpp
```diff
@@ -446,7 +470,20 @@ namespace irods {
...
+                        get_source_data_object_attributes(_ctx.comm(), replica_to_update.data_id, _leaf_bundles);
+                }
+                catch (const irods::exception& _e) {
+                    if (_e.code() != CAT_NO_ROWS_FOUND) {
+                        throw;
+                    }
+                    rodsLog(LOG_WARNING, "Cannot update replica [%lld] for data_id [%lld]. Skipping.", replica_to_update.replica_number, replica_to_update.data_id);
+                    replicas_to_skip++;
+                    continue;
+                }
```
korydraughn, line 485: Should we handle `std::exception`?

plugins/resources/replication/src/irods_repl_rebalance.cpp
```diff
@@ -498,33 +535,41 @@ namespace irods {
...
                 }
-                if (data_ids_needing_new_replicas.empty()) {
-                    break;
+
+                if (repls_to_skip > 0) {
+                    did_skip_some_repls = true;
                 }
 
-                proc_results_for_rebalance(_ctx, _resource_name, child_name, i, _leaf_bundles, data_ids_needing_new_replicas);
+                rodsLog(LOG_WARNING, "Ignored size: [%i].", repls_to_skip);
```
korydraughn, line 567: Use new logger.

scripts/irods/test/test_iadmin.py
korydraughn, line 499: Don't forget to update test name.

scripts/irods/test/test_iadmin.py
korydraughn, line 504: I assume more words will be added later?

scripts/irods/test/test_iadmin.py
```diff
@@ -509,6 +495,121 @@ def test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380(
...
+        self.admin.assert_icommand("iadmin mkresc leaf_c unixfilesystem " + hostname +
+                                   ":/tmp/irods/pydevtest_leaf_c", 'STDOUT_SINGLELINE', "Creating")  # unix
+
+        self.admin.assert_icommand("iadmin addchildtoresc pt repl")
+        self.admin.assert_icommand("iadmin addchildtoresc repl leaf_a")
+        self.admin.assert_icommand("iadmin addchildtoresc repl pt_b")
+        self.admin.assert_icommand("iadmin addchildtoresc repl pt_c1")
+        self.admin.assert_icommand("iadmin addchildtoresc pt_b leaf_b")
+        self.admin.assert_icommand("iadmin addchildtoresc pt_c1 pt_c2")
+        self.admin.assert_icommand("iadmin addchildtoresc pt_c2 leaf_c")
```
korydraughn, line 538: These need to be wrapped in try-finally for clean up purposes.

scripts/irods/test/test_iadmin.py
```diff
@@ -509,6 +495,121 @@ def test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380(
...
+        self.admin.assert_icommand("iadmin mkresc pt_c1 passthru", 'STDOUT_SINGLELINE', "Creating")
+        self.admin.assert_icommand("iadmin mkresc pt_c2 passthru", 'STDOUT_SINGLELINE', "Creating")
+        self.admin.assert_icommand("iadmin mkresc repl replication", 'STDOUT_SINGLELINE', "Creating")
+
+        self.admin.assert_icommand("iadmin mkresc leaf_a unixfilesystem " + hostname +
+                                   ":/tmp/irods/pydevtest_leaf_a", 'STDOUT_SINGLELINE', "Creating")  # unix
+        self.admin.assert_icommand("iadmin mkresc leaf_b unixfilesystem " + hostname +
+                                   ":/tmp/irods/pydevtest_leaf_b", 'STDOUT_SINGLELINE', "Creating")  # unix
+        self.admin.assert_icommand("iadmin mkresc leaf_c unixfilesystem " + hostname +
+                                   ":/tmp/irods/pydevtest_leaf_c", 'STDOUT_SINGLELINE', "Creating")  # unix
```
korydraughn, line 530: Could use `lib.create_ufs_resource()` for these.

Not required, but easier to read.

scripts/irods/test/test_iadmin.py
```diff
@@ -509,6 +495,121 @@ def test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380(
...
+            self.admin.assert_icommand("iadmin rmchildfromresc pt repl")
+
+            self.admin.assert_icommand("iadmin rmresc leaf_c")
+            self.admin.assert_icommand("iadmin rmresc leaf_b")
+            self.admin.assert_icommand("iadmin rmresc leaf_a")
+            self.admin.assert_icommand("iadmin rmresc pt_c2")
+            self.admin.assert_icommand("iadmin rmresc pt_c1")
+            self.admin.assert_icommand("iadmin rmresc pt_b")
+            self.admin.assert_icommand("iadmin rmresc repl")
+            self.admin.assert_icommand("iadmin rmresc pt")
```
korydraughn, line 611: assert_icommand -> run_icommand

scripts/irods/test/test_iadmin.py
```diff
@@ -509,6 +495,121 @@ def test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380(
...
+                    self.admin.assert_icommand(["iadmin", "modrepl", "logical_path", os.path.join(self.admin.session_collection, file), 'replica_number', '1', 'DATA_REPL_STATUS', '0'])
+                    self.admin.assert_icommand(["iadmin", "modrepl", "logical_path", os.path.join(self.admin.session_collection, file), 'replica_number', '2', 'DATA_REPL_STATUS', '0'])
+
+                # =-=-=-=-=-=-=-
+                # visualize our tree
+                self.admin.assert_icommand("ils -AL", 'STDOUT_SINGLELINE', "foo")
+
+                # =-=-=-=-=-=-=-
+                # call rebalance function - the thing were actually testing... finally.
+                self.admin.assert_icommand("iadmin modresc pt rebalance", "STDERR", "-1834000 REBALANCE_NOT_COMPLETE", desired_rc=4)
```
korydraughn, line 569: Prefer testing for nonzero instead of using `desired_rc`.

scripts/irods/test/test_iadmin.py
```diff
@@ -509,6 +495,121 @@ def test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380(
             self.admin.assert_icommand("iadmin rmuser %s" % user_name)
             self.admin.assert_icommand("iadmin rmgroup %s" % group_name)
 
+
+    def test_rebalance_under_funny_case__isssue_6111(self):
+        output = subprocess.getstatusoutput("hostname")
+        hostname = output[1]
```
korydraughn, line 501: `lib.create_ufs_resource()` removes the need for this.

scripts/irods/test/test_iadmin.py
```diff
@@ -509,6 +495,121 @@ def test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380(
...
+                IrodsController().reload_configuration()
+
+                # =-=-=-=-=-=-=-
+                # place data into the resource
+                test_file = "iput_test_file"
+                lib.make_file(test_file, 10)
+                num_children = 10
+                bad_file_indices = (1,3,5)
+                for i in range(num_children):
+                    self.admin.assert_icommand("iput -R pt %s foo%d" % (test_file, i))
```
korydraughn, line 552: Can use `itouch` to create data objects.

korydraughn, line 552: For non-rodsadmin operations, prefer using a non-rodsadmin user session. Helps us catch more bugs.

scripts/irods/test/test_iadmin.py
```diff
@@ -509,6 +495,121 @@ def test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380(
...
+
+                # Trim so we get more replication going
+                for i in range(num_children):
+                    self.admin.assert_icommand("itrim -N2 -n 0 foo%d" % (i), 'STDOUT_SINGLELINE', 'Total size trimmed')
+
+                # Invalidate all replicas of "foo{i}"
+                for file in [f"foo{i}" for i in bad_file_indices]:
+                    self.admin.assert_icommand(["iadmin", "modrepl", "logical_path", os.path.join(self.admin.session_collection, file), 'replica_number', '1', 'DATA_REPL_STATUS', '0'])
```
korydraughn, line 560: Consider constructing the logical path outside of the assertion so it can reused.

scripts/irods/test/test_iadmin.py
```diff
@@ -509,6 +495,121 @@ def test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380(
...
+
+                # =-=-=-=-=-=-=-
+                # visualize our rebalance
+                self.admin.assert_icommand("ils -AL", 'STDOUT_SINGLELINE', "foo")
+
+                # TODO: Check for the absence of '3'
+                # Maybe assert icommand fail?
+                for file in [f"foo{i}" for i in bad_file_indices]:
+                    self.admin.assert_icommand(f"ils -AL {file}", 'STDOUT_SINGLELINE', [" 1 ", " X", f" {file}"])
+                    self.admin.assert_icommand(f"ils -AL {file}", 'STDOUT_SINGLELINE', [" 2 ", " X", f" {file}"])
```
korydraughn, line 579: In regard to L576, there's also the option of capturing the return objects and asserting things on them.

```py
_, out, _ = self.admin.assert_icommand(..., 'STDOUT', [current expectations])
self.assertNotIn('string containing "3"', out)
```

scripts/irods/test/test_iadmin.py
korydraughn, line 499: From what I've read, this test covers the case where data objects have a mixture of good and stale replicas - i.e. the partial success case.

Do we have a test which covers the case where all data objects have no good replica?

korydraughn (commented):
First pass.

Will circle back once this is out of draft.

plugins/database/src/db_plugin.cpp
```diff
@@ -13647,6 +13645,117 @@ irods::error db_get_repl_list_for_leaf_bundles_op(
...
+        for( auto id : (*_bundles)[idx] ) {
+            not_child_stream << id << ",";
+        }
+    } // for idx
+
+    std::string not_child_array = not_child_stream.str();
+    if (not_child_array.empty()) {
+        return SUCCESS();
+    }
+    not_child_array.pop_back(); // trim last ','
```
alanking, line 13702: Could avoid the `pop_back`s and copying with `fmt::join`, I think. Something like this:
```c++
const auto& child_index_bundle = (*_bundles)[_child_index];
const auto child_array = fmt::format(
    "{}", fmt::join(child_index_bundle.start(), child_index_bundle.end(), ","));
```
Might be a little trickier on the part where we are skipping the index, but seems possible.

Also, we need to make sure that `_child_index` does not go past the end of the `_bundles` vector, or use `std::vector::at` instead.

plugins/resources/replication/src/irods_repl_rebalance.cpp
```diff
@@ -498,33 +535,41 @@ namespace irods {
                 THROW(first_error.code(), first_error.result());
             }
         }
+        return static_cast<bool>(replicas_to_skip);
```
alanking, line 538: Would it ever be interesting to actually return this value rather than a `bool`?

scripts/irods/test/test_iadmin.py
alanking: There's also the option of looking at the replica status directly: `lib.get_replica_status`

scripts/irods/test/test_iadmin.py
alanking: Could be done with `lib.replica_exists`

MartinFlores751:
All core tests passed on Ubuntu 24.04 with Postgres 14, MySQL 8.0, and MariaDB 10.11.

plugins/resources/replication/src/irods_repl_rebalance.cpp
MartinFlores751: Done.

scripts/irods/test/test_iadmin.py
MartinFlores751: As much as I like `...funny_case...`, the name will be changed to something more useful. Leaving open until the test name changes.

plugins/database/src/db_plugin.cpp
MartinFlores751: Since this is largely a copy-paste function, `db_get_repl_list_for_leaf_bundles_op` is missing this check too.

Should I add it there too?

plugins/database/src/db_plugin.cpp
MartinFlores751: See previous comment for more details: https://github.com/irods/irods/pull/8930#discussion_r3149669283

Should I also update this in `db_get_repl_list_for_leaf_bundles_op`?

plugins/resources/replication/src/irods_repl_rebalance.cpp
MartinFlores751: Removed.

plugins/resources/replication/src/irods_repl_rebalance.cpp
MartinFlores751: Removed.

plugins/resources/replication/src/irods_repl_rebalance.cpp
MartinFlores751: Removed.

scripts/irods/test/test_iadmin.py
MartinFlores751: Resolving this comment in favor of previous one: https://github.com/irods/irods/pull/8930#discussion_r3149648376

scripts/irods/test/test_iadmin.py
MartinFlores751: Resolving this comment in favor of previous one: https://github.com/irods/irods/pull/8930#discussion_r3087735527

plugins/database/src/db_plugin.cpp
korydraughn: Yes please do as a separate commit. Feel free to open an issue for it.

plugins/database/src/db_plugin.cpp
korydraughn: Not at this time. Open an issue to fix that in 6.0.0.

server/icat/include/irods/icatHighLevelRoutines.hpp
MartinFlores751: Will update.

server/icat/src/icatHighLevelRoutines.cpp
MartinFlores751: Done.

plugins/database/src/db_plugin.cpp
MartinFlores751: Done.

plugins/database/src/db_plugin.cpp
MartinFlores751: Done.

plugins/database/src/db_plugin.cpp
MartinFlores751: Done.

scripts/irods/test/test_iadmin.py
MartinFlores751: Updated test name.

scripts/irods/test/test_iadmin.py
MartinFlores751: Done.

server/icat/include/irods/icatHighLevelRoutines.hpp
MartinFlores751: Done.

scripts/irods/test/test_iadmin.py
MartinFlores751: Done.

scripts/irods/test/test_iadmin.py
MartinFlores751: Done.

scripts/irods/test/test_iadmin.py
MartinFlores751: Done.

plugins/database/src/db_plugin.cpp
MartinFlores751: Done.

plugins/database/src/db_plugin.cpp
MartinFlores751: Done. The commit is separated with a `<BONUS>` tag.

scripts/irods/test/test_iadmin.py
MartinFlores751: Done.

scripts/irods/test/test_iadmin.py
MartinFlores751: Code is unneeded. Removed.

scripts/irods/test/test_iadmin.py
MartinFlores751: Opted for `lib.get_replica_status`. Resolving.

plugins/database/src/db_plugin.cpp
MartinFlores751: Done. Issue also created: #8949.

scripts/irods/test/test_iadmin.py
MartinFlores751: `itouch` couldn't create on non-leaf nodes. Left as is, unless it's more desirable to have three `itouch` calls on the leaf nodes? Can do that if desired.

MartinFlores751: Changed to `user0` in most cases. Resolving.

scripts/irods/test/test_iadmin.py
MartinFlores751: Constructed the path outside. Resolving.

plugins/resources/replication/src/irods_repl_rebalance.cpp
MartinFlores751: Is there anything gained by changing `Incomplete repl` to ` [N] incomplete repl(s)`? Since we do not store the exact ids(?) skipped, I don't think it would be too useful in a client or server error.

Following your thought, is there a benefit to being specific about which function failed? We have separate bools for `irods::update_out_of_date_replicas` and `irods::create_missing_replicas`:
```c++
...
        auto has_bad_repl =
            irods::update_out_of_date_replicas(_ctx, leaf_bundles, batch_size, invocation_timestamp, resource_name);
        auto has_missing_repls =
            irods::create_missing_replicas(_ctx, leaf_bundles, batch_size, invocation_timestamp, resource_name);
        if (has_bad_repl || has_missing_repls) {
            return ERROR(REBALANCE_NOT_COMPLETE, "Missing or stale replica(s) still exist.");
        }
...
```

scripts/irods/test/test_iadmin.py
MartinFlores751: I'll make one. I expect it'll still report `-1834000 REBALANCE_NOT_COMPLETE`, but a test will show if that's right.

plugins/resources/replication/src/irods_repl_rebalance.cpp
MartinFlores751: At a glance, throwing calls in `get_source_data_object_attributes` are caught and rethrown as `irods::exception`. Is there something in particular that catches your eye? Since the control flow here uses exceptions, handling them may change expected behavior.

plugins/resources/replication/src/irods_repl_rebalance.cpp
MartinFlores751: The same function within the `try` block is being discussed here: https://github.com/irods/irods/pull/8930#discussion_r3087754180

Deferring comments to the other thread.

scripts/irods/test/test_iadmin.py
korydraughn: Not required. Just trying to make understanding the various steps in the test easier/faster.

plugins/resources/replication/src/irods_repl_rebalance.cpp
korydraughn: If `get_source_data_object_attributes` re-throws `std::exception` as `irods::exception`, then I think we're okay here.

My comment was motivated by the lack of the `std::exception` catch block, but sounds like it is already covered.

plugins/resources/replication/src/irods_repl_rebalance.cpp
MartinFlores751: In standup discussion, we decided the `Missing or stale replica(s) still exist.` message is enough to prompt the user to look into the failures. If more detail is required or desired, it can be changed in the future.

Resolving.

scripts/irods/test/test_iadmin.py
MartinFlores751: Changed test to use `itouch`. Resolving.

plugins/resources/replication/src/irods_repl_rebalance.cpp
MartinFlores751: Gotcha. Marking comment as resolved.

scripts/irods/test/test_iadmin.py
MartinFlores751: Done.

scripts/irods/test/test_iadmin.py
MartinFlores751: Test created, rebalance still reports `REBALANCE_NOT_COMPLETE`. Resolving.

plugins/database/src/db_plugin.cpp
MartinFlores751: Simplified the first one, gave a good shot at the second. Resolving.

server/icat/include/irods/icatHighLevelRoutines.hpp
MartinFlores751: Done.

MartinFlores751:
Code and tests have changed a bit. I'll have them rerunning in the background.

korydraughn:
All good.

Are there any specific changes which we should pay attention to?

MartinFlores751:
I think the `chlGetReplListForLeafBundlesOffset` function docs and the `db_get_repl_list_for_leaf_bundles_offset_op` function are the most interesting, though that's just what I recall changing most recently.

korydraughn:
Okay. I'll keep that in mind as I look over the code.

server/icat/include/irods/icatHighLevelRoutines.hpp
```diff
@@ -229,6 +229,34 @@ int chlGetReplListForLeafBundles(
     const std::string*          _invocation_timestamp,
     dist_child_result_t*        _results );
 
+/// \brief Wrapper which calls db_get_repl_list_for_leaf_bundles_offset_op
+///
+/// \parblock
+/// Allows for the offsetting of the results returned from the query in
+/// db_get_repl_list_for_leaf_bundles_offset_op.
```
korydraughn: Let's avoid mentioning the actual database plugin operation in the docs since that can change over time.

server/icat/include/irods/icatHighLevelRoutines.hpp
```diff
@@ -229,6 +229,34 @@ int chlGetReplListForLeafBundles(
...
+///
+/// \returns Error code based on whether results are generated successfully.\p
+///
+/// \since 5.0.2
+int chlGetReplListForLeafBundlesOffset(rodsLong_t _count,
+                                       size_t _child_idx,
+                                       const std::vector<leaf_bundle_t>* _bundles,
+                                       const std::string* _invocation_timestamp,
+                                       dist_child_result_t* _results,
+                                       int _offset);
```
korydraughn: You stated that your wording might need some tweaks. Try comparing what you have written to the functions below this one.

server/icat/include/irods/icatHighLevelRoutines.hpp
```diff
@@ -229,6 +229,34 @@ int chlGetReplListForLeafBundles(
...
+/// \endparblock
+///
+/// \param[in] _count Number of results to return at once
+/// \param[in] _child_idx Index to leaf resource of interest
+/// \param[in] _bundles Vector of all leaf resources
+/// \param[in] _invocation_timestamp Timestamp of when rebalance was called
+/// \param[in,out] _results Vector of items needing reblancing
+/// \param[in] _offest Number of results to skip before returning
+///
+/// \returns Error code based on whether results are generated successfully.\p
```
korydraughn: Trailing `\p`?

server/icat/include/irods/icatHighLevelRoutines.hpp
```diff
@@ -229,6 +229,34 @@ int chlGetReplListForLeafBundles(
...
+/// An example use for this function is to process the results in a partial
+/// manner, leaving some results behind that need to be accounted for via
+/// an offset.
+/// \endparblock
+///
+/// \param[in] _count Number of results to return at once
+/// \param[in] _child_idx Index to leaf resource of interest
+/// \param[in] _bundles Vector of all leaf resources
+/// \param[in] _invocation_timestamp Timestamp of when rebalance was called
+/// \param[in,out] _results Vector of items needing reblancing
```
korydraughn:
```suggestion
/// \param[in,out] _results Vector of items needing rebalancing
```

plugins/resources/replication/include/irods/private/irods_repl_rebalance.hpp
```diff
@@ -13,19 +13,17 @@ namespace irods {
...
-        irods::plugin_context& _ctx,
-        const std::vector<leaf_bundle_t>& _leaf_bundles,
-        const int _batch_size,
-        const std::string& _invocation_timestamp,
-        const std::string& resource_name);
+    bool update_out_of_date_replicas(irods::plugin_context& _ctx,
+                                     const std::vector<leaf_bundle_t>& _leaf_bundles,
+                                     const int _batch_size,
+                                     const std::string& _invocation_timestamp,
+                                     const std::string& resource_name);
```
korydraughn:
```suggestion
                                     const std::string& _resource_name);
```

plugins/resources/replication/include/irods/private/irods_repl_rebalance.hpp
```diff
@@ -13,19 +13,17 @@ namespace irods {
...
-        irods::plugin_context& _ctx,
-        const std::vector<leaf_bundle_t>& _leaf_bundles,
-        const int _batch_size,
-        const std::string& _invocation_timestamp,
-        const std::string& resource_name);
+    bool create_missing_replicas(irods::plugin_context& _ctx,
+                                 const std::vector<leaf_bundle_t>& _leaf_bundles,
+                                 const int _batch_size,
+                                 const std::string& _invocation_timestamp,
+                                 const std::string& resource_name);
```
korydraughn:
```suggestion
                                 const std::string& _resource_name);
```

scripts/irods/test/test_iadmin.py
```diff
@@ -509,6 +495,224 @@ def test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380(
             self.admin.assert_icommand("iadmin rmuser %s" % user_name)
             self.admin.assert_icommand("iadmin rmgroup %s" % group_name)
 
+    def test_rebalance_has_partial_success_with_no_good_replicas__issue_6111(self):
+        # Don't use the default rebalance limit.
+        # Use a rebalance smaller limit to simulate large
```
korydraughn: Should this be ...


```suggestion
        # Use a smaller rebalance limit to simulate large
```

scripts/irods/test/test_iadmin.py
```diff
@@ -509,6 +495,224 @@ def test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380(
...
+            lib.create_ufs_resource(self.admin, 'leaf_b')
+            lib.create_ufs_resource(self.admin, 'leaf_c')
+
+            self.admin.assert_icommand("iadmin addchildtoresc pt repl")
+            self.admin.assert_icommand("iadmin addchildtoresc repl leaf_a")
+            self.admin.assert_icommand("iadmin addchildtoresc repl pt_b")
+            self.admin.assert_icommand("iadmin addchildtoresc repl pt_c1")
+            self.admin.assert_icommand("iadmin addchildtoresc pt_b leaf_b")
+            self.admin.assert_icommand("iadmin addchildtoresc pt_c1 pt_c2")
+            self.admin.assert_icommand("iadmin addchildtoresc pt_c2 leaf_c")
```
korydraughn: Please add a unique identifier to the resource names in case the test fails and is unable to clean up.

I tend to include `issue_<number>` in the resource names. The goal is to avoid causing other tests to fail because of name collisions.

scripts/irods/test/test_iadmin.py
```diff
@@ -509,6 +495,224 @@ def test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380(
...
+                core.add_rule(pep_map[self.plugin_name])
+                IrodsController().reload_configuration()
+
+                for file in paths_to_bad_data_objects:
+                    for leaf in ('leaf_a', 'leaf_b', 'leaf_c'):
+                        self.user0.assert_icommand(f"itouch -R {leaf} {file}")
+
+                # Trim so we get more replication going
+                for file in paths_to_bad_data_objects:
+                    self.user0.assert_icommand(f"itrim -N2 -n 0 {file}", 'STDOUT_SINGLELINE', 'Total size trimmed')
```
korydraughn:
```suggestion
                    self.user0.assert_icommand(f"itrim -N2 -n0 {file}", 'STDOUT_SINGLELINE', 'Total size trimmed')
```

scripts/irods/test/test_iadmin.py
```diff
@@ -509,6 +495,224 @@ def test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380(
...
+
+            with temporary_core_file() as core:
+                core.add_rule(pep_map[self.plugin_name])
+                IrodsController().reload_configuration()
+
+                for file in paths_to_bad_data_objects:
+                    for leaf in ('leaf_a', 'leaf_b', 'leaf_c'):
+                        self.user0.assert_icommand(f"itouch -R {leaf} {file}")
+
+                # Trim so we get more replication going
```
korydraughn: Let's update the comments to be more explicit about what's happening. Remove any chance of the reader misunderstanding the story. For example:


```suggestion
                # Trim replica 0 of each data object. This increases the replication
                # workload.
```

scripts/irods/test/test_iadmin.py
```diff
@@ -509,6 +495,224 @@ def test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380(
...
+                for file in paths_to_bad_data_objects:
+                    self.user0.assert_icommand(f"itrim -N2 -n 0 {file}", 'STDOUT_SINGLELINE', 'Total size trimmed')
+
+                # Invalidate all replicas of "foo{i}"
```
korydraughn: Given the function calls within the body, "Invalidate" can only mean "stale". If that's the case, please update the comment to use "Stale" so that there's no room for confusion.

scripts/irods/test/test_iadmin.py
```diff
@@ -509,6 +495,224 @@ def test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380(
...
+                for file in paths_to_bad_data_objects:
+                    lib.set_replica_status(self.admin, file, 1, 0)
+                    lib.set_replica_status(self.admin, file, 2, 0)
+
+                # =-=-=-=-=-=-=-
+                # visualize our tree
+                self.user0.assert_icommand("ils -AL", 'STDOUT_SINGLELINE', "foo")
+
+                # =-=-=-=-=-=-=-
+                # call rebalance function - the thing were actually testing... finally.
```
korydraughn:
```suggestion
                # call rebalance function - the thing we're actually testing... finally.
```

scripts/irods/test/test_iadmin.py
```diff
@@ -509,6 +495,224 @@ def test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380(
...
+                    self.user0.assert_icommand(f"itrim -N2 -n 0 {file}", 'STDOUT_SINGLELINE', 'Total size trimmed')
+
+                # Invalidate all replicas of "foo{i}"
+                for file in paths_to_bad_data_objects:
+                    lib.set_replica_status(self.admin, file, 1, 0)
+                    lib.set_replica_status(self.admin, file, 2, 0)
+
+                # =-=-=-=-=-=-=-
+                # visualize our tree
+                self.user0.assert_icommand("ils -AL", 'STDOUT_SINGLELINE', "foo")
```
korydraughn: This can be rewritten as ...
```suggestion
                self.user0.assert_icommand("ils -AL", 'STDOUT')
```

That is starting to become a convention used in core iRODS tests. It is primarily for forcing output into the XML files so that we can see what the state of the world is.

scripts/irods/test/test_iadmin.py
```diff
@@ -509,6 +495,224 @@ def test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380(
...
+
+                # =-=-=-=-=-=-=-
+                # call rebalance function - the thing were actually testing... finally.
+                rc, _, _ = self.admin.assert_icommand("iadmin modresc pt rebalance", "STDERR", "-1834000 REBALANCE_NOT_COMPLETE")
+                self.assertNotEqual(rc, 0)
+
+                # =-=-=-=-=-=-=-
+                # visualize our rebalance
+                self.user0.assert_icommand("ils -AL", 'STDOUT_SINGLELINE', "foo")
```
korydraughn:
```suggestion
                self.user0.assert_icommand("ils -AL", 'STDOUT')
```

scripts/irods/test/test_iadmin.py
```diff
@@ -509,6 +495,224 @@ def test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380(
...
+
+                # Assert failed replications still exist and no replication performed
+                for file in [os.path.basename(f) for f in paths_to_bad_data_objects]:
+                    self.assertEqual(lib.get_replica_status(self.user0, file, 1), '0')
+                    self.assertEqual(lib.get_replica_status(self.user0, file, 2), '0')
+                    self.assertFalse(lib.replica_exists(self.user0, file, 3))
+
+        finally:
+            # =-=-=-=-=-=-=-
+            # TEARDOWN
```
korydraughn: This can be removed.

scripts/irods/test/test_iadmin.py
```diff
@@ -509,6 +495,224 @@ def test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380(
...
+                # =-=-=-=-=-=-=-
+                # call rebalance function - the thing were actually testing... finally.
+                rc, _, _ = self.admin.assert_icommand("iadmin modresc pt rebalance", "STDERR", "-1834000 REBALANCE_NOT_COMPLETE")
+                self.assertNotEqual(rc, 0)
+
+                # =-=-=-=-=-=-=-
+                # visualize our rebalance
+                self.user0.assert_icommand("ils -AL", 'STDOUT_SINGLELINE', "foo")
+
+                # Assert failed replications still exist and no replication performed
```
korydraughn: Consider using more words so that it's clear what the expectation is.

scripts/irods/test/test_iadmin.py
```diff
@@ -509,6 +495,224 @@ def test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380(
...
+                # Trim so we get more replication going
+                for file in paths_to_bad_data_objects:
+                    self.user0.assert_icommand(f"itrim -N2 -n 0 {file}", 'STDOUT_SINGLELINE', 'Total size trimmed')
+
+                # Invalidate all replicas of "foo{i}"
+                for file in paths_to_bad_data_objects:
+                    lib.set_replica_status(self.admin, file, 1, 0)
+                    lib.set_replica_status(self.admin, file, 2, 0)
+
+                # =-=-=-=-=-=-=-
```
korydraughn: Let's remove all instances of these.

scripts/irods/test/test_iadmin.py
```diff
@@ -509,6 +495,224 @@ def test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380(
...
+            self.admin.run_icommand("iadmin rmresc leaf_a")
+            self.admin.run_icommand("iadmin rmresc pt_c2")
+            self.admin.run_icommand("iadmin rmresc pt_c1")
+            self.admin.run_icommand("iadmin rmresc pt_b")
+            self.admin.run_icommand("iadmin rmresc repl")
+            self.admin.run_icommand("iadmin rmresc pt")
+
+    def test_rebalance_has_partial_success_with_missing_or_stale_replicas__issue_6111(self):
+        # Don't use the default rebalance limit.
+        # Use a rebalance smaller limit to simulate large
```
korydraughn: ???

```suggestion
        # Use a smaller rebalance limit to simulate large
```

scripts/irods/test/test_iadmin.py
```diff
@@ -509,6 +495,224 @@ def test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380(
...
+                    *OUT="replication_rebalance_limit=1";
+                }
+            '''),
+            'irods_rule_engine_plugin-python': textwrap.dedent('''
+                def pep_resource_rebalance_pre(rule_args, callback, rei):
+                    rule_args[0] = 'replication_rebalance_limit=1'
+            ''')
+        }
+
+        # Define number of data objects and data objects will be stale
```
korydraughn: What does the second half of this sentence mean?
```
... and data objects will be stale
```

scripts/irods/test/test_iadmin.py
```diff
@@ -509,6 +495,224 @@ def test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380(
...
+
+        # Define number of data objects
+        num_children = 10
+
+        # Generate full paths to data objects
+        paths_to_bad_data_objects = [os.path.join(self.user0.session_collection, f"foo{i}") for i in range(num_children)]
+
+        try:
+            # =-=-=-=-=-=-=-
+            # STANDUP
```
korydraughn: Can remove this.

plugins/database/src/db_plugin.cpp
```diff
@@ -13647,6 +13648,131 @@ irods::error db_get_repl_list_for_leaf_bundles_op(
...
+        return ERROR(SYS_INVALID_INPUT_PARAM, "no bundles");
+    }
+    if (nullptr == _invocation_timestamp) {
+        return ERROR(INVALID_INPUT_ARGUMENT_NULL_POINTER, "invocation timestamp is NULL");
+    }
+    if (_invocation_timestamp->empty()) {
+        return ERROR(SYS_INVALID_INPUT_PARAM, "invocation timestamp is empty");
+    }
+
+    std::string child_array{fmt::format("{}", fmt::join((*_bundles).at(_child_index), ","))};
```
korydraughn: Curious. Why `(*_bundles)` instead of `_bundles->`?

plugins/database/src/db_plugin.cpp
```diff
@@ -13647,6 +13648,131 @@ irods::error db_get_repl_list_for_leaf_bundles_op(
...
+    if (status_cmlGetFirstRowFromSql == CAT_NO_ROWS_FOUND) {
+        cmlFreeStatement(statement_num, &icss);
+        return SUCCESS();
+    }
+    if (status_cmlGetFirstRowFromSql != 0) {
+        cmlFreeStatement(statement_num, &icss);
+        return ERROR(status_cmlGetFirstRowFromSql, fmt::format("failed to get first row from query [{}]", query));
+    }
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-constant-array-index)
+    _results->push_back(strtoll(icss.stmtPtr[statement_num]->resultValue[0], nullptr, 0));
```
korydraughn: Missing `std::` on `strtoll`.

plugins/database/src/db_plugin.cpp
```diff
@@ -13647,6 +13648,131 @@ irods::error db_get_repl_list_for_leaf_bundles_op(
...
+        if (status_cmlGetNextRowFromStatement == CAT_NO_ROWS_FOUND) {
+            break;
+        }
+        if (status_cmlGetNextRowFromStatement != 0) {
+            cmlFreeStatement(statement_num, &icss);
+            return ERROR(
+                status_cmlGetNextRowFromStatement, fmt::format("failed to get row [{}] from query [{}]", i, query));
+        }
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-constant-array-index)
+        _results->push_back(strtoll(icss.stmtPtr[statement_num]->resultValue[0], nullptr, 0));
```
korydraughn: Missing `std::` on `strtoll`.

plugins/database/src/db_plugin.cpp
korydraughn: What happens if `strtoll` fails?

plugins/database/src/db_plugin.cpp
```diff
@@ -13647,6 +13648,131 @@ irods::error db_get_repl_list_for_leaf_bundles_op(
...
+                    child_array,
+                    _invocation_timestamp->c_str(),
+                    _count);
+#elif MY_ICAT
+    /* MySQL (MariaDB doesn't get 'except' until v10.3)*/
+    const std::string query = fmt::format("select distinct data_id from R_DATA_MAIN "
+                                          "  where resc_id in ({}) and data_id not in ( "
+                                          "    select data_id from R_DATA_MAIN "
+                                          "      where resc_id in ({}) "
+                                          "  ) and modify_ts <= '{}' "
```
korydraughn: Collapse whitespace in string so the SQL is tight.

plugins/database/src/db_plugin.cpp
```diff
@@ -13647,6 +13648,131 @@ irods::error db_get_repl_list_for_leaf_bundles_op(
...
+                                          child_array,
+                                          _invocation_timestamp->c_str(),
+                                          _offset);
+#else
+    /* Postgres */
+    const std::string query = fmt::format("select distinct data_id from R_DATA_MAIN "
+                                          "  where resc_id in ({}) and modify_ts <= '{}' "
+                                          "except "
+                                          "  select data_id from R_DATA_MAIN "
+                                          "    where resc_id in ({}) "
```
korydraughn: Collapse whitespace in string so the SQL is tight.

plugins/database/src/db_plugin.cpp
```diff
@@ -13647,6 +13648,131 @@ irods::error db_get_repl_list_for_leaf_bundles_op(
...
+        return ERROR(SYS_INVALID_INPUT_PARAM, "invocation timestamp is empty");
+    }
+
+    std::string child_array{fmt::format("{}", fmt::join((*_bundles).at(_child_index), ","))};
+    if (child_array.empty()) {
+        return ERROR(SYS_INVALID_INPUT_PARAM, "leaf array is empty");
+    }
+
+    std::string not_child_array;
+    for (size_t idx = 0; idx < _bundles->size(); ++idx) {
```
korydraughn, line 13694: Starting here, add comments to help the reader quickly understand what is happening throughout the function. You don't need to explain everything. Try to focus on the parts that might trip up the reader.

korydraughn (commented):
Getting real close.

MartinFlores751:
Reran core tests. All core tests passed on Ubuntu 24.04 with Postgres 14, MySQL 8.0, and MariaDB 10.11.

korydraughn:
Great. All that's left to do is address the review comments. None of them appear to have any effect on the implementation's behavior, but please re-run your new tests, after making changes, to be sure.

plugins/database/src/db_plugin.cpp
MartinFlores751: TLDR, we should have a `strtoll` helper that does error checking here and wherever else calls this func.

If I recall, the function might not parse the entire string and may report this in the second parameter (which is `nullptr` here). If the value is out of range, it returns the max or min representation and sets `errno`. There may be other errors, but these are what I recall. So, no error handling is done here currently.

plugins/database/src/db_plugin.cpp
MartinFlores751: It's an artifact of using `[_child_index]` before.

Since the line is highlighted, do we prefer exceptions (aka using `.at()`) or returning an `irods::error` with `SYS_INVALID_INPUT_PARAM` if `_child_index >= _bundles->size()` ?

korydraughn: For plugins, always return an error. We want to avoid throwing exceptions across plugin boundaries.

plugins/database/src/db_plugin.cpp
korydraughn: Right. We need some error handling around that call.

Due to time constraints, let's make it work first. We can replace it with a better implementation post release. Perhaps there's a function in stringOpr.h (I think) that does that.

scripts/irods/test/test_iadmin.py
MartinFlores751: Thinking about it, there is no real reason for the trimming if all three replicas are only created via the previous `itouch`. I should just be able to remove `leaf_c` in the `itouch` list and skip the trimming. I'll give it a try and resolve this commit if trimming can be skipped.

scripts/irods/test/test_iadmin.py
MartinFlores751: I'll just trim everything after the hyphen.

scripts/irods/test/test_iadmin.py
MartinFlores751: Related: https://github.com/irods/irods/pull/8930#discussion_r3171116593

Use "stale replicas" to make it clear what's expected.

scripts/irods/test/test_iadmin.py
MartinFlores751: `# Define number of data objects and [the index of the] data objects [that] will be [marked] stale`

Will update.

server/icat/include/irods/icatHighLevelRoutines.hpp
MartinFlores751: Done.

server/icat/include/irods/icatHighLevelRoutines.hpp
MartinFlores751: Removed.

server/icat/include/irods/icatHighLevelRoutines.hpp
MartinFlores751: Done.

scripts/irods/test/test_iadmin.py
MartinFlores751: Swapped.

scripts/irods/test/test_iadmin.py
MartinFlores751: Done.

scripts/irods/test/test_iadmin.py
MartinFlores751: Done.

scripts/irods/test/test_iadmin.py
MartinFlores751: Swapped.

plugins/database/src/db_plugin.cpp
MartinFlores751: Changed to use `[_child_index]` guarded by a check of `_child_index >= _bundles->size()` that returns an `irods::error`.

plugins/database/src/db_plugin.cpp
MartinFlores751: Done.

plugins/database/src/db_plugin.cpp
MartinFlores751: Done.

plugins/resources/replication/include/irods/private/irods_repl_rebalance.hpp
MartinFlores751: Done.

plugins/resources/replication/include/irods/private/irods_repl_rebalance.hpp
MartinFlores751: Done.

scripts/irods/test/test_iadmin.py
MartinFlores751: Done.

scripts/irods/test/test_iadmin.py
MartinFlores751: Turns out `itrim` is still needed, but only one `itouch` is required. Updated the test to reflect this. Also updated the comment to make clear what the purpose of the `itrim` is.

scripts/irods/test/test_iadmin.py
MartinFlores751: Done.

scripts/irods/test/test_iadmin.py
MartinFlores751: Done.

scripts/irods/test/test_iadmin.py
MartinFlores751: Done.

scripts/irods/test/test_iadmin.py
MartinFlores751: Removed.

scripts/irods/test/test_iadmin.py
MartinFlores751: Done.

scripts/irods/test/test_iadmin.py
MartinFlores751: Done.

scripts/irods/test/test_iadmin.py
MartinFlores751: Done.

plugins/database/src/db_plugin.cpp
MartinFlores751: Added a lambda that does error checking.

MartinFlores751: Using error-checked lambda instead.

plugins/database/src/db_plugin.cpp
MartinFlores751: Using error-checked lambda instead.

scripts/irods/test/test_iadmin.py
MartinFlores751: Done.

server/icat/include/irods/icatHighLevelRoutines.hpp
MartinFlores751: This has been updated, but I'm going to give it another pass before resolving.

plugins/database/src/db_plugin.cpp
MartinFlores751: Done.

plugins/database/src/db_plugin.cpp
```diff
@@ -13647,6 +13649,174 @@ irods::error db_get_repl_list_for_leaf_bundles_op(
...
+        cmlFreeStatement(statement_num, &icss);
+        return SUCCESS();
+    }
+    if (status_cmlGetFirstRowFromSql != 0) {
+        cmlFreeStatement(statement_num, &icss);
+        return ERROR(status_cmlGetFirstRowFromSql, fmt::format("failed to get first row from query [{}]", query));
+    }
+
+    // Setting '_base' to zero means the input may be base 10 or
+    // base 8 depending on if the input has a leading zero
```
alanking: Maybe not worth mentioning, but... a prefix of `0x` or `0X` will make it base 16: https://en.cppreference.com/cpp/string/byte/strtol

alanking (commented):
Very nice

server/icat/include/irods/icatHighLevelRoutines.hpp
MartinFlores751: Done.

plugins/database/src/db_plugin.cpp
MartinFlores751: Done.

plugins/database/src/db_plugin.cpp
```diff
@@ -13647,6 +13649,176 @@ irods::error db_get_repl_list_for_leaf_bundles_op(
...
+                    _count);
+#elif MY_ICAT
+    /* MySQL (MariaDB doesn't get 'except' until v10.3)*/
+    const std::string query = fmt::format("select distinct data_id from R_DATA_MAIN "
+                                          "where resc_id in ({}) and data_id not in ( "
+                                          "select data_id from R_DATA_MAIN "
+                                          "where resc_id in ({}) "
+                                          ") and modify_ts <= '{}' "
+                                          "order by data_id "
+                                          "limit {},18446744073709551615",
```
korydraughn: Is the use of `18446744073709551615` intentional? Why is that needed here?

plugins/database/src/db_plugin.cpp
```diff
@@ -13647,6 +13649,176 @@ irods::error db_get_repl_list_for_leaf_bundles_op(
...
+    if (status_cmlGetFirstRowFromSql != 0) {
+        cmlFreeStatement(statement_num, &icss);
+        return ERROR(status_cmlGetFirstRowFromSql, fmt::format("failed to get first row from query [{}]", query));
+    }
+
+    // Setting '_base' to zero means the input may be:
+    // - base 16 if the input starts with "0x" or "0X"
+    // - base 8 if the input has a leading zero
+    // - base 10 if none of the other conditions are met
+    auto checked_strtoll{[](std::string_view _input_string, int _base = 0) -> std::variant<long long, irods::error> {
```
korydraughn: I can't imagine something other than base 10 appearing here. Consider locking it to base 10.

server/icat/include/irods/icatHighLevelRoutines.hpp
```diff
@@ -229,6 +229,34 @@ int chlGetReplListForLeafBundles(
...
+/// \param[in] _count Number of results to return at once
+/// \param[in] _child_idx Index to leaf resource of interest
+/// \param[in] _bundles Vector of all leaf resources
+/// \param[in] _invocation_timestamp Timestamp of when rebalance was called
+/// \param[in,out] _results Vector of items needing rebalancing
+/// \param[in] _offest Number of results to skip before returning
+///
+/// \returns Error code based on whether results are generated successfully.
+///
+/// \since 5.0.2
```
korydraughn:
```suggestion
/// \since 5.1.0
```

scripts/irods/test/test_iadmin.py
```diff
@@ -509,6 +496,210 @@ def test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380(
             self.admin.assert_icommand("iadmin rmuser %s" % user_name)
             self.admin.assert_icommand("iadmin rmgroup %s" % group_name)
 
+    def test_rebalance_has_partial_success_with_no_good_replicas__issue_6111(self):
+        resource_id = '_issue_6111_' + uuid.uuid4().hex
+        # Don't use the default rebalance limit.
```
korydraughn, line 502:
```suggestion
        resource_id = '_issue_6111_' + uuid.uuid4().hex

        # Don't use the default rebalance limit.
```

scripts/irods/test/test_iadmin.py
```diff
@@ -509,6 +496,210 @@ def test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380(
...
+            self.admin.run_icommand(f"iadmin rmresc leaf_a{resource_id}")
+            self.admin.run_icommand(f"iadmin rmresc pt_c2{resource_id}")
+            self.admin.run_icommand(f"iadmin rmresc pt_c1{resource_id}")
+            self.admin.run_icommand(f"iadmin rmresc pt_b{resource_id}")
+            self.admin.run_icommand(f"iadmin rmresc repl{resource_id}")
+            self.admin.run_icommand(f"iadmin rmresc pt{resource_id}")
+
+    def test_rebalance_has_partial_success_with_missing_or_stale_replicas__issue_6111(self):
+        resource_id = '_issue_6111_' + uuid.uuid4().hex
+        # Don't use the default rebalance limit.
```
korydraughn, line 600:
```suggestion
        resource_id = '_issue_6111_' + uuid.uuid4().hex

        # Don't use the default rebalance limit.
```

scripts/irods/test/test_iadmin.py
```diff
@@ -509,6 +496,210 @@ def test_iadmin_rmuser_rmgroup_report_error_on_mismatch_entity_type__issue_7380(
...
+            'irods_rule_engine_plugin-python': textwrap.dedent('''
+                def pep_resource_rebalance_pre(rule_args, callback, rei):
+                    rule_args[0] = 'replication_rebalance_limit=1'
+            ''')
+        }
+
+        # Define number of data objects and the index
+        # of the data objects that will be marked stale
+        num_children = 10
+        bad_file_indices = (1,3,5)
```
korydraughn:
```suggestion
        bad_file_indices = (1, 3, 5)
```

korydraughn (commented):
Looks good overall. Spotted some minor things.

Does this need to go through another round of testing?

plugins/database/src/db_plugin.cpp
MartinFlores751: That is a good question and should probably be documented better.

The limit/offset code is based off existing code found in `generateSQL(...)`: https://github.com/irods/irods/blob/555c3e13bfa93ebc4cb214d419b3aba9493be5bc/plugins/database/src/general_query.cpp#L2144-L2163

Looking at some docs, the magic number seems to be the max number of rows returned, so the code is probably wrong: https://www.geeksforgeeks.org/mysql/mysql-limit-clause/

MartinFlores751: In a mini discussion with @korydraughn, it was decided that the approach will be changed to a combination of `offset` and `limit`. The `limit` will be changed from a magic number to `_count`.

scripts/irods/test/test_iadmin.py
MartinFlores751: Done.

scripts/irods/test/test_iadmin.py
MartinFlores751: Done.

scripts/irods/test/test_iadmin.py
MartinFlores751: Done.

server/icat/include/irods/icatHighLevelRoutines.hpp
MartinFlores751: Done.

plugins/database/src/db_plugin.cpp
MartinFlores751: Set the default to base 10. Left the comments above in case others don't use the default base.

plugins/database/src/db_plugin.cpp
MartinFlores751: Done.

MartinFlores751:
Running the core tests again on Ubuntu 24.04 with Postgres 14, MySQL 8.0, and MariaDB 10.11.

plugins/database/src/db_plugin.cpp
```diff
@@ -13766,7 +13766,7 @@ irods::error db_get_repl_list_for_leaf_bundles_offset_op(irods::plugin_context&
     // - base 16 if the input starts with "0x" or "0X"
     // - base 8 if the input has a leading zero
     // - base 10 if none of the other conditions are met
-    auto checked_strtoll{[](std::string_view _input_string, int _base = 0) -> std::variant<long long, irods::error> {
+    auto checked_strtoll{[](std::string_view _input_string, int _base = 10) -> std::variant<long long, irods::error> {
```
korydraughn: You can hard-code the base and remove the parameter. It's unlikely that anyone will ever change the base at this level.

MartinFlores751: Updated and removed the comment above since it's no longer relevant.

korydraughn:
Awaiting test results.

MartinFlores751:
Some tests in MySQL and MariaDB failed. After looking into the server logs, the issue seems to be the change to use both `limit` and `offset` for the MySQL/MariaDB query suggested in the following comment: https://github.com/irods/irods/pull/8930#discussion_r3210314242 

I'll revert the query to use the previous style of `limit <query_offset>, <querry_limit>` and rerun the tests.

korydraughn:
You can also try connecting to the MySQL/MariaDB directly and running queries involving `LIMIT`, `OFFSET`, and `FETCH FIRST N ROWS ONLY` to see if it's a keyword problem.

MartinFlores751:
I took a look at the MySQL docs: https://dev.mysql.com/doc/refman/8.0/en/select.html

In MySQL, `offset` has to follow `limit` since it's part of the same clause(?). After updating the statement to be `limit {} offset {}`, the failing tests on MySQL and MariaDB pass.

I'm running the core tests again to make sure they all pass with the new changes.

korydraughn:
Please test against postgresql too.

MartinFlores751:
@korydraughn and I discussed some comments left in #7984 and we decided that this PR resolves the issue.

MartinFlores751:
Core and unit tests on Ubuntu 24.04 with Postgres, MySQL, and MariaDB all pass.

korydraughn (commented):
Squash to taste.

MartinFlores751:
Squashed.

korydraughn (approved):
Pound it.

MartinFlores751:
Pounded.

Issue #8928, opened 2026-04-14
univmss resource plugin fails when logical path contains single quotes


korydraughn:
## Bug Report

The univmss resource plugin contains a parser which understands escaped single quotes. The problem is that nothing leading up to the invocation of a univmss script escapes single quotes. This results in errors such as `UNIV_MSS_SYNCTOARCH_ERR` and `UNIV_MSS_UNLINK_ERR`.

To correct this, the univmss resource plugin needs to escape all inputs before passing them to any external scripts.

A better solution would be to remove the need for the parser and adjust functions within the resource plugin to match the expectations. This thought is motivated by the fact that the resource plugin generates strings containing multiple arguments, but then splits them to create an argument vector (i.e. `char*[]`).

_This issue is motivated by the google group conversation at https://groups.google.com/g/irod-chat/c/fOUemtxFjCE/m/vFsaDwRGGAAJ._

PR #8931, opened 2026-04-15
[#8928] univmss: Escape single quotes in paths (main)

korydraughn:
Running core tests.

The implementation is complete. It's mostly polish from point.

### DIFF
--- a/plugins/resources/src/univmss.cpp
+++ b/plugins/resources/src/univmss.cpp
@@ -1,30 +1,34 @@
-#include "irods/msParam.h"
 #include "irods/generalAdmin.h"
-#include "irods/physPath.hpp"
-#include "irods/reIn2p3SysRule.hpp"
-#include "irods/miscServerFunct.hpp"
-#include "irods/rsExecCmd.hpp"
-#include "irods/irods_resource_plugin.hpp"
-#include "irods/irods_file_object.hpp"
-#include "irods/irods_physical_object.hpp"
 #include "irods/irods_collection_object.hpp"
-#include "irods/irods_string_tokenize.hpp"
+#include "irods/irods_file_object.hpp"
 #include "irods/irods_hierarchy_parser.hpp"
+#include "irods/irods_kvp_string_parser.hpp"
+#include "irods/irods_logger.hpp"
+#include "irods/irods_physical_object.hpp"
+#include "irods/irods_re_structs.hpp"
+#include "irods/irods_resource_constants.hpp"
+#include "irods/irods_resource_plugin.hpp"
 #include "irods/irods_resource_redirect.hpp"
 #include "irods/irods_stacktrace.hpp"
-#include "irods/irods_re_structs.hpp"
+#include "irods/irods_string_tokenize.hpp"
+#include "irods/miscServerFunct.hpp"
+#include "irods/msParam.h"
+#include "irods/physPath.hpp"
+#include "irods/reIn2p3SysRule.hpp"
+#include "irods/rsExecCmd.hpp"
 #include "irods/voting.hpp"
 
-#include <boost/lexical_cast.hpp>
-#include <boost/function.hpp>
-#include <boost/any.hpp>
 #include <boost/algorithm/string.hpp>
+#include <boost/any.hpp>
+#include <boost/function.hpp>
+#include <boost/lexical_cast.hpp>
 
 #include <cstring>
 #include <iostream>
 #include <sstream>
-#include <vector>
 #include <string>
+#include <string_view>
+#include <vector>
 
 /// =-=-=-=-=-=-=-
 /// @brief Check the general parameters passed in to most plugin functions
@@ -47,9 +51,53 @@ inline irods::error univ_mss_check_param(
 #define NB_READ_TOUT_SEC        60      /* 60 sec timeout */
 #define NB_WRITE_TOUT_SEC       60      /* 60 sec timeout */
 
-// =-=-=-=-=-=-=-
-/// @brief token to index the script property
-const std::string SCRIPT_PROP( "script" );
+namespace
+{
+    using log_resc = irods::experimental::log::resource;
+
+    // The context string property for identifying the script/binary to execute.
+    const char* const SCRIPT_PROP = "script";
+
+    // The context string property for escaping single quotes in paths.
+    const char* const ESCAPE_SINGLE_QUOTES = "escape_single_quotes";
+
+    // Returns the input string with single quotes escaped when the ESCAPE_SINGLE_QUOTES
+    // property is set to "1". Otherwise, the input string is returned unchanged.
+    auto escape_single_quotes(irods::plugin_context& _ctx, const std::string& _input) -> std::string
+    {
+        std::string resc_name;
+        if (const auto err = _ctx.prop_map().get<std::string>(irods::RESOURCE_NAME, resc_name); !err.ok()) {
+            log_resc::debug("{}: Could not retrieve name of univmss resource.", __func__);
+            resc_name = "<unavailable>";
+        }
+
+        std::string value;
+        if (const auto err = _ctx.prop_map().get<std::string>(ESCAPE_SINGLE_QUOTES, value); err.ok()) {
+            boost::trim(value);
+        }
+        else {
+            // If we're here, then something horrible has happened. The plugin always sets
+            // a default value for this property.
+            log_resc::critical(
+                "{}: Missing context string property [{}] for univmss resource [{}]. Escaping single quotes "
+                "in paths is disabled.",
+                __func__,
+                ESCAPE_SINGLE_QUOTES,
+                resc_name);
+            return _input;
+        }
+
+        if ("1" != value) {
+            log_resc::debug(
+                "{}: Escaping single quotes in paths is disabled for univmss resource [{}].", __func__, resc_name);
+            return _input;
+        }
+
+        log_resc::debug(
+            "{}: Escaping single quotes in paths is enabled for univmss resource [{}].", __func__, resc_name);
+        return boost::replace_all_copy(_input, "'", "\\'");
+    } // escape_single_quotes
+} // anonymous namespace
 
 /// =-=-=-=-=-=-=-
 /// @brief interface for POSIX create
@@ -123,10 +171,12 @@ irods::error univ_mss_file_unlink(
     irods::data_object_ptr fco = boost::dynamic_pointer_cast< irods::data_object >( _ctx.fco() );
     std::string filename = fco->physical_path();
 
+    const auto escaped_filename = escape_single_quotes(_ctx, filename);
+
     execCmd_t execCmdInp;
     memset( &execCmdInp, 0, sizeof( execCmdInp ) );
     snprintf( execCmdInp.cmd, sizeof( execCmdInp.cmd ), "%s", script.c_str() );
-    snprintf( execCmdInp.cmdArgv, sizeof( execCmdInp.cmdArgv ), "rm '%s'", filename.c_str() );
+    snprintf(execCmdInp.cmdArgv, sizeof(execCmdInp.cmdArgv), "rm '%s'", escaped_filename.c_str());
     snprintf( execCmdInp.execAddr, sizeof( execCmdInp.execAddr ), "localhost" );
 
     execCmdOut_t *execCmdOut = NULL;
@@ -186,9 +236,11 @@ irods::error univ_mss_file_stat(
     struct tm mytm;
     time_t myTime;
 
+    const auto escaped_filename = escape_single_quotes(_ctx, filename);
+
     std::memset(&execCmdInp, 0, sizeof(execCmdInp));
     rstrcpy( execCmdInp.cmd, script.c_str(), LONG_NAME_LEN );
-    snprintf( cmdArgv, sizeof( cmdArgv ), "stat '%s' ", filename.c_str() );
+    snprintf(cmdArgv, sizeof(cmdArgv), "stat '%s' ", escaped_filename.c_str());
     rstrcpy( execCmdInp.cmdArgv, cmdArgv, HUGE_NAME_LEN );
     rstrcpy( execCmdInp.execAddr, "localhost", LONG_NAME_LEN );
     status = _rsExecCmd( &execCmdInp, &execCmdOut );
@@ -296,9 +348,11 @@ irods::error univ_mss_file_chmod(
         mode = getDefFileMode();
     }
 
+    const auto escaped_filename = escape_single_quotes(_ctx, filename);
+
     std::memset(&execCmdInp, 0, sizeof(execCmdInp));
     snprintf( execCmdInp.cmd, sizeof( execCmdInp.cmd ), "%s", script.c_str() );
-    snprintf( execCmdInp.cmdArgv, sizeof( execCmdInp.cmdArgv ), "chmod '%s' %o", filename.c_str(), mode );
+    snprintf(execCmdInp.cmdArgv, sizeof(execCmdInp.cmdArgv), "chmod '%s' %o", escaped_filename.c_str(), mode);
     snprintf( execCmdInp.execAddr, sizeof( execCmdInp.execAddr ), "%s", "localhost" );
     execCmdOut_t *execCmdOut = NULL;
     status = _rsExecCmd( &execCmdInp, &execCmdOut );
@@ -346,12 +400,14 @@ irods::error univ_mss_file_mkdir(
     irods::collection_object_ptr fco = boost::dynamic_pointer_cast< irods::collection_object >( _ctx.fco() );
     std::string dirname = fco->physical_path();
 
+    const auto escaped_dirname = escape_single_quotes(_ctx, dirname);
+
     int status = 0;
     execCmd_t execCmdInp;
 
     std::memset(&execCmdInp, 0, sizeof(execCmdInp));
     snprintf( execCmdInp.cmd, sizeof( execCmdInp.cmd ), "%s", script.c_str() );
-    snprintf( execCmdInp.cmdArgv, sizeof( execCmdInp.cmdArgv ), "mkdir '%s'", dirname.c_str() );
+    snprintf(execCmdInp.cmdArgv, sizeof(execCmdInp.cmdArgv), "mkdir '%s'", escaped_dirname.c_str());
     snprintf( execCmdInp.execAddr, sizeof( execCmdInp.execAddr ), "%s", "localhost" );
     execCmdOut_t *execCmdOut = NULL;
     status = _rsExecCmd( &execCmdInp, &execCmdOut );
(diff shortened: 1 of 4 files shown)


plugins/resources/src/univmss.cpp
```diff
@@ -743,40 +799,63 @@ irods::error univ_mss_file_rebalance(
...
+  public:
+    univ_mss_resource(const std::string& _inst_name, const std::string& _context)
+        : irods::resource(_inst_name, _context)
+    {
+        // Default to "0" to avoid breaking existing deployments. This makes it
+        // more noticeable during debugging sessions and removes the need to handle
+        // cases where the property isn't set in the context string.
+        //
+        // TODO(???): Set to "1" in iRODS 6.
+        properties_.set<std::string>(ESCAPE_SINGLE_QUOTES, "0");
```
korydraughn, line 813: Open issue for TODO and handle it later.

plugins/resources/src/univmss.cpp
```diff
@@ -743,40 +799,63 @@ irods::error univ_mss_file_rebalance(
...
+        _flg = false;
+        return SUCCESS();
+    }
+
+    // =-=-=-=-=-=-
+    // override from plugin_base
+    irods::error post_disconnect_maintenance_operation(irods::pdmo_type&)
+    {
+        return ERROR(-1, "nop"); // TODO Replace -1 with concrete error code
+    }
```
korydraughn, line 858: Open issue for TODO and handle it later.

scripts/irods/test/test_resource_types.py
```diff
@@ -2151,6 +2147,111 @@ def test_irepl_with_purgec(self):
...
             os.unlink(filepath)
 
+    def test_univmss_resource_plugin_supports_paths_containing_single_quotes__issue_8928(self):
+        filename = "issue'8928.txt"
+        file_path = f'{self.user0.local_session_dir}/{filename}'
+        local_file_path = Path(file_path)
+        local_file_path.touch()
+
+        try:
+            # TODO(???): Remove this line when the default for "escape_single_quotes" changes in iRODS 6.
```
korydraughn, line 2157: Open issue for TODO.

scripts/irods/test/test_resource_types.py
```diff
@@ -2151,6 +2147,111 @@ def test_irepl_with_purgec(self):
...
+            with self.subTest('iput fails when path contains single quotes and escape_single_quotes not enabled'):
+                self.user0.assert_icommand(['iput', file_path], 'STDERR', ['-550000 UNIV_MSS_SYNCTOARCH_ERR'])
+                self.user0.assert_icommand(['ils', '-l', filename], 'STDOUT', [f' X {filename}\n'])
+                self.user0.assert_icommand(['irm', '-f', filename], 'STDERR', ['-552000 UNIV_MSS_UNLINK_ERR'])
+
+            # TODO(???): Remove this line when the default for "escape_single_quotes" changes in iRODS 6.
```
korydraughn, line 2163: Open issue for TODO.

scripts/irods/test/test_resource_types.py
```diff
@@ -2151,6 +2147,111 @@ def test_irepl_with_purgec(self):
...
+                self.assertTrue(lib.replica_exists(self.user0, f'{self.user0.session_collection}/{new_filename}', 0))
+
+                self.user0.assert_icommand(['imv', new_filename, filename])
+                self.assertTrue(lib.replica_exists(self.user0, f'{self.user0.session_collection}/{filename}', 0))
+                self.assertFalse(lib.replica_exists(self.user0, f'{self.user0.session_collection}/{new_filename}', 0))
+
+        finally:
+            self.user0.run_icommand(['irm', '-f', filename])
+
+            # TODO(???): Remove this line when the default for "escape_single_quotes" changes in iRODS 6.
```
korydraughn, line 2252: Open issue for TODO.

scripts/irods/test/test_resource_types.py
```diff
@@ -2151,6 +2147,111 @@ def test_irepl_with_purgec(self):
...
+
+        try:
+            # TODO(???): Remove this line when the default for "escape_single_quotes" changes in iRODS 6.
+            with self.subTest('iput fails when path contains single quotes and escape_single_quotes not enabled'):
+                self.user0.assert_icommand(['iput', file_path], 'STDERR', ['-550000 UNIV_MSS_SYNCTOARCH_ERR'])
+                self.user0.assert_icommand(['ils', '-l', filename], 'STDOUT', [f' X {filename}\n'])
+                self.user0.assert_icommand(['irm', '-f', filename], 'STDERR', ['-552000 UNIV_MSS_UNLINK_ERR'])
+
+            # TODO(???): Remove this line when the default for "escape_single_quotes" changes in iRODS 6.
+            self.admin.assert_icommand(['iadmin', 'modresc', 'archiveResc', 'context', 'script=univMSSInterface.sh;escape_single_quotes=1;'])
```
korydraughn, line 2164: Use GenQuery to store the original context string, so that it can be restored later.

plugins/resources/src/univmss.cpp
```diff
@@ -47,9 +49,38 @@ inline irods::error univ_mss_check_param(
...
 #define NB_WRITE_TOUT_SEC       60      /* 60 sec timeout */
 
-// =-=-=-=-=-=-=-
-/// @brief token to index the script property
-const std::string SCRIPT_PROP( "script" );
+namespace
+{
+    using log_resc = irods::experimental::log::resource;
+
+    // =-=-=-=-=-=-=-
```
korydraughn, line 56: Remove this.

plugins/resources/src/univmss.cpp
```diff
@@ -47,9 +49,38 @@ inline irods::error univ_mss_check_param(
...
+        std::string value;
+        if (const auto err = _ctx.prop_map().get<std::string>(ESCAPE_SINGLE_QUOTES, value); !err.ok()) {
+            // If we're here, then something horrible has happened. The plugin always sets
+            // a default value for this property.
+            log_resc::error(
+                "{}: Missing context string property [{}] for univmss resource.", __func__, ESCAPE_SINGLE_QUOTES);
+        }
+
+        boost::trim(value);
+
```
korydraughn, line 76: Consider removing the empty line above or below L75.

plugins/resources/src/univmss.cpp
```diff
@@ -743,40 +799,63 @@ irods::error univ_mss_file_rebalance(
...
-            return SUCCESS();
+            return;
         }
 
-        // =-=-=-=-=-=-
-        // override from plugin_base
-        irods::error post_disconnect_maintenance_operation( irods::pdmo_type& ) {
-            return ERROR( -1, "nop" );
+        // Fallback to checking the context string for inappropriate path behavior.
+        // The context string is expected to be the name of script or binary located
```
korydraughn, line 833:
```suggestion
        // The context string is expected to be the name of a script or binary located
```

plugins/resources/src/univmss.cpp
```diff
@@ -743,40 +799,63 @@ irods::error univ_mss_file_rebalance(
...
+        // in msiExecCmd_bin. This aligns with the original behavior of the plugin.
+        if (context_.find("/") != std::string::npos) {
+            log_resc::warn("univmss resource: Expected a single filename which should resides in msiExecCmd_bin. "
```
trel, line 836:
```suggestion
            log_resc::warn("univmss resource: Expected a single filename which should reside in msiExecCmd_bin. "
```

trel (commented):
looks good - just one typo found.

korydraughn:
All unit and core tests pass.

korydraughn:
Will address clang-tidy report.

plugins/resources/src/univmss.cpp
korydraughn: Opened #8934.

plugins/resources/src/univmss.cpp
korydraughn: Opened #8935.

scripts/irods/test/test_resource_types.py
korydraughn: Opened #8935.

scripts/irods/test/test_resource_types.py
korydraughn: Opened #8935.

scripts/irods/test/test_resource_types.py
korydraughn: Opened #8935.

plugins/resources/src/univmss.cpp
korydraughn: Updated TODO comment to include issue number.

plugins/resources/src/univmss.cpp
korydraughn: Removed trailing comment.

scripts/irods/test/test_resource_types.py
korydraughn: Updated TODO comment to include issue number.

scripts/irods/test/test_resource_types.py
korydraughn: Updated TODO comment to include issue number.

scripts/irods/test/test_resource_types.py
korydraughn: Updated TODO comment to include issue number.

scripts/irods/test/test_resource_types.py
korydraughn: Done.

plugins/resources/src/univmss.cpp
korydraughn: Done.

plugins/resources/src/univmss.cpp
korydraughn: Done.

plugins/resources/src/univmss.cpp
korydraughn: Done.

plugins/resources/src/univmss.cpp
korydraughn: Done.

korydraughn:
Added new tests to make sure univmss supports its original style of context string and semicolon delimited context strings.

Those changes do not require a full rerun of the full test suite. I reran the compound test suite to make sure it still passes.

This is now ready for final review. Next step is to update irods/irods_docs to cover the context string enhancement.

plugins/resources/src/univmss.cpp
```diff
@@ -743,40 +797,59 @@ irods::error univ_mss_file_rebalance(
...
 
-        // =-=-=-=-=-=-
-        // override from plugin_base
-        irods::error post_disconnect_maintenance_operation( irods::pdmo_type& ) {
-            return ERROR( -1, "nop" );
+        // Fallback to checking the context string for inappropriate path behavior.
+        // The context string is expected to be the name of a script or binary located
+        // in msiExecCmd_bin. This aligns with the original behavior of the plugin.
+        if (context_.find('/') != std::string::npos) {
+            log_resc::warn("univmss resource: Expected a single filename which should reside in msiExecCmd_bin. "
```
trel: must?

```suggestion
            log_resc::warn("univmss resource: Expected a single filename which must reside in msiExecCmd_bin. "
```

plugins/resources/src/univmss.cpp
```diff
@@ -743,40 +797,59 @@ irods::error univ_mss_file_rebalance(
...
+                           "Context string is set to [{}].",
+                           context_);
```
korydraughn: Apply this same check to the key-value context string logic.

scripts/irods/test/test_resource_types.py
```diff
@@ -2151,6 +2147,183 @@ def test_irepl_with_purgec(self):
...
+
+        # Capture the original context string.
+        _, out, _ = self.admin.assert_icommand(['iquest', '%s', "select RESC_CONTEXT where RESC_NAME = 'archiveResc'"], 'STDOUT')
+        original_resc_context = out.strip()
+
+        try:
+            # TODO(#8935): Remove this line when the default for "escape_single_quotes" changes in iRODS 6.
+            with self.subTest('iput fails when path contains single quotes and escape_single_quotes not enabled'):
+                self.user0.assert_icommand(['iput', file_path], 'STDERR', ['-550000 UNIV_MSS_SYNCTOARCH_ERR'])
+                self.user0.assert_icommand(['ils', '-l', filename], 'STDOUT', [f' X {filename}\n'])
```
alanking: Alternatively, can check replica status directly with `lib.get_replica_status`.

scripts/irods/test/test_resource_types.py
```diff
@@ -2151,6 +2147,183 @@ def test_irepl_with_purgec(self):
...
+                self.user0.assert_icommand(['istream', 'write', filename], input='data')
+                _, out, _ = self.user0.assert_icommand(['ils', '-l', filename], 'STDOUT', [
+                    ' 0 demoResc;cacheResc ',
+                    ' 1 demoResc;archiveResc ',
+                    f' & {filename}\n'
+                ])
+                self.assertNotIn(f' X {filename}\n', out)
+                self.user0.assert_icommand(['irm', '-f', filename])
+
+            # These subtests verify the stage-to-cache functionality behaves as expected.
```
alanking, line 2273: Maybe not worth including, but stage-to-cache also happens on overwrites.

scripts/irods/test/test_resource_types.py
```diff
@@ -2151,6 +2147,183 @@ def test_irepl_with_purgec(self):
...
+                self.assertNotIn(f' X {filename}\n', out)
+                new_filename = filename + "'renamed"
+
+                self.user0.assert_icommand(['imv', filename, new_filename])
+                self.assertFalse(lib.replica_exists(self.user0, f'{self.user0.session_collection}/{filename}', 0))
+                self.assertTrue(lib.replica_exists(self.user0, f'{self.user0.session_collection}/{new_filename}', 0))
+
+                self.user0.assert_icommand(['imv', new_filename, filename])
+                self.assertTrue(lib.replica_exists(self.user0, f'{self.user0.session_collection}/{filename}', 0))
+                self.assertFalse(lib.replica_exists(self.user0, f'{self.user0.session_collection}/{new_filename}', 0))
```
alanking: If this, for whatever reason, fails, that means the cleanup would fail, too, right? Should this get its own try-finally?

plugins/resources/src/univmss.cpp
```diff
@@ -47,9 +49,36 @@ inline irods::error univ_mss_check_param(
...
+    // Returns the input string with single quotes escaped when the ESCAPE_SINGLE_QUOTES
+    // property is set to "1". Otherwise, the input string is returned unchanged.
+    auto escape_single_quotes(irods::plugin_context& _ctx, const std::string& _input) -> std::string
+    {
+        std::string value;
+        if (const auto err = _ctx.prop_map().get<std::string>(ESCAPE_SINGLE_QUOTES, value); !err.ok()) {
+            // If we're here, then something horrible has happened. The plugin always sets
+            // a default value for this property.
+            log_resc::error(
+                "{}: Missing context string property [{}] for univmss resource.", __func__, ESCAPE_SINGLE_QUOTES);
```
alanking: In the off chance that this codepath *is* taken, it looks like the behavior from here is to not escape single quotes (since `"1" != ""`). Should we include that in the error message?

scripts/irods/test/test_resource_types.py
korydraughn: Oh right. I actually tested that at the bench. I'll add a case for that.

scripts/irods/test/test_resource_types.py
korydraughn: I think you're right. I can conditionally run `run_icommand` in the existing finally-block to cover `new_filename`.

plugins/resources/src/univmss.cpp
korydraughn: Good catch. Yes, we should say something about that.

plugins/resources/src/univmss.cpp
korydraughn: Done.

plugins/resources/src/univmss.cpp
korydraughn: Done.

scripts/irods/test/test_resource_types.py
korydraughn: Done.

scripts/irods/test/test_resource_types.py
korydraughn: Done.

scripts/irods/test/test_resource_types.py
korydraughn: Done.

plugins/resources/src/univmss.cpp
korydraughn: Done.

korydraughn:
Adjusted tests to use `lib.get_replica_status_for_resource` where it made sense.

New tests still pass. I don't expect the compound test suite to fail with the latest commits, but rerunning them to be absolutely sure.

Clang-Format is producing weird results for a couple lines, so ignoring those cases.

korydraughn:
Compound test run passed.

All review comments have been addressed.

plugins/resources/src/univmss.cpp
```diff
@@ -64,18 +65,22 @@ namespace
...
-        if (const auto err = _ctx.prop_map().get<std::string>(ESCAPE_SINGLE_QUOTES, value); !err.ok()) {
+        if (const auto err = _ctx.prop_map().get<std::string>(ESCAPE_SINGLE_QUOTES, value); err.ok()) {
+            boost::trim(value);
+        }
+        else {
             // If we're here, then something horrible has happened. The plugin always sets
             // a default value for this property.
-            log_resc::error(
+            log_resc::critical(
                 "{}: Missing context string property [{}] for univmss resource.", __func__, ESCAPE_SINGLE_QUOTES);
```
alanking: This may not be a big deal (given that only the S3 resource is good about this), but we may want to find a way to log the resource name as well. As it stands, we don't know to which instance this is referring. Feel free to skip it or punt since this has not been the case in this resource historically anyway.

plugins/resources/src/univmss.cpp
```diff
@@ -810,6 +815,13 @@ class univ_mss_resource : public irods::resource
         // TODO(#8935): Set to "1" in iRODS 6.
         properties_.set<std::string>(ESCAPE_SINGLE_QUOTES, "0");
 
+        const auto log_warning_if_script_path_is_invalid = [](const std::string_view _sv) {
+            if (_sv.find('/') != std::string_view::npos) {
+                log_resc::warn("univmss resource: Expected a single filename which must reside in msiExecCmd_bin. "
+                               "Script is set to [{}].", _sv);
```
alanking: In the two uses of `log_warning_if_script_path_is_invalid` that I see, one is passing in the name of the script, and one is passing in the whole context string. Maybe we should have this string be expected as a context string and update the message to reflect that? i.e.:
```suggestion
                               "Context string is set to [{}].", _sv);
```

alanking (commented):
Looks good. Noticed a couple of other things but they have to do with log messages, so they're not important

plugins/resources/src/univmss.cpp
korydraughn: I'm interpreting your question in multiple ways. Is the concern that the input string to the lambda will be different depending on the form of the context string?

alanking: Unless I'm misunderstanding something, here's what I'm seeing...

The warning message says: `"Script is set to [{}]."`

Let's say we have a context string like this:
```
escape_single_quotes=0;script=/hello
```

The callsite on R841 passes the script value:
```c++
                if (SCRIPT_PROP == key) {
                    log_warning_if_script_path_is_invalid(value);
                }
```
The warning message will read:
```
Script is set to [/hello].
```
The callsite on R852 passes the whole context string:
```c++
        log_warning_if_script_path_is_invalid(context_);
```

The warning message will read:
```
Script is set to [escape_single_quotes=0;script=/hello].
```

I'm saying we should just have it always print the whole context string so it's easier:
```
Context string is set to [escape_single_quotes=0;script=/hello].
```
Alternatively, we can pick out the script part and print that out, but that will require more code.

korydraughn: I see. The lambda will only ever see the script/filename. Never the semicolon-delimited context string.

The call which passes the `context_` is expected to always be a single filename. It can never be a key-value, semicolon-delimited string due to the earlier check for `=`.

alanking: Ah, I see it now. Thanks. Resolving

plugins/resources/src/univmss.cpp
korydraughn: Agreed.

Done.

alanking (commented):
Seems good to me. Squash if happy

korydraughn:
Given I updated the code a little bit, I'm going to run the compound test suite one more time to be absolutely sure things are still happy.

korydraughn:
`test_resource_types.Test_Resource_CompoundWithUnivmss` passed.

Squashing.

korydraughn:
Squashed.

alanking (approved):
\# it

korydraughn:
Added pounds.

Issue #8729, opened 2025-11-05
Add option to use irods authentication at setup time


alanking:
### New Feature / Functionality

For iRODS 5.1, the default authentication scheme for the account running the iRODS server will continue to be native. We should provide an option to use the new irods authentication scheme instead so that it doesn't need to be changed after setup later.

Issue #7717, opened 2024-04-24
Unattended Install with SSL configuration results in error


0x010C:
- [x] update setup script
- [x] update documentation

---

### iRODS Version, OS and Version
iRODS v4.3.1 on Debian 12

### What did you try to do?
I'm currently automating the deployment of iRODS servers on our infrastructure. For this I use the --json_configuration_file parameter of the setup_irods.py script. This works perfectly until I tried to activate SSL.

```bash
$ cat irods_unattended_installation.json
{
        .....
        "service_account_environment":{
                "irods_client_server_negotiation": "request_server_negotiation",
                "irods_client_server_policy": "CS_NEG_REQUIRE",
                "irods_ssl_certificate_chain_file": "/etc/ssl/irods/fullchain.pem",
                "irods_ssl_certificate_key_file": "/etc/ssl/irods/privkey.pem",
                "irods_ssl_dh_params_file": "/etc/ssl/irods/dhparams.pem",
                ...
        }
}

$ python3 /var/lib/irods/scripts/setup_irods.py --json_configuration_file irods_unattended_installation.json
```

### Expected behavior
I'm expecting to have an up and running iRODS server with SSL enabled.

### Observed behavior (including steps to reproduce, if applicable)
```bash
sudo python3 /var/lib/irods/scripts/setup_irods.py --json_configuration_file irods_unattended_installation.json
Error encountered running setup_irods:
Traceback (most recent call last):
  File "/var/lib/irods/scripts/setup_irods.py", line 523, in main
    setup_server(irods_config,
  File "/var/lib/irods/scripts/setup_irods.py", line 148, in setup_server
    test_put(irods_config)
  File "/var/lib/irods/scripts/setup_irods.py", line 180, in test_put
    raise IrodsError('Post-install test failed. Please check your configuration.')
irods.exceptions.IrodsError: Post-install test failed. Please check your configuration.
```

Looking closer, the script does not update the SSL setting in the /etc/irods/core.re file, although it does for the default resource for example.
```bash
$cat /etc/irods/core.re 
...
acPreConnect(*OUT) { *OUT="CS_NEG_REFUSE"; }
...
acSetRescSchemeForCreate {msiSetDefaultResc("myResc","null"); }
acSetRescSchemeForRepl {msiSetDefaultResc("myResc","null"); }
...
```
[...]

Issue #6785, opened 2022-12-19
Investigate using different authentication schemes in testing


alanking:
## Feature

The python-based testing for iRODS/iCommands uses native authentication for all tests except for the tests found in `test_auth.py`. We should consider adding an option to `run_tests.py` which allows for specifying an authentication scheme. In this way, the entire test suite can be run using, e.g. pam_password.

PR #8939, opened 2026-04-23
Add support for TLS and irods auth in setup_irods.py

alanking:
Addresses #7717
Addresses #8729

There are 2-by-2 things being added here:
 - Choice of auth scheme and TLS (2)
 - Script prompts or unattended install (2)

We don't really have tests for the setup script, so I've just been doing manual tests. We cannot run our test suite with irods authentication (see https://github.com/irods/irods/issues/6785). So that leaves us with TLS using unattended install vs not-unattended install.

In order to run said tests, changes need to be made in the testing environment, the PR for which I will open/link later.

### DIFF
--- a/scripts/irods/database_connect.py
+++ b/scripts/irods/database_connect.py
@@ -432,8 +432,8 @@ def setup_database_values(irods_config, cursor=None, default_resource_directory=
         }
         params = hashing_parameters["parameters"]
         derived_key = hashlib.scrypt(
-            irods_config.admin_password.encode(),
-            salt=salt.encode(),
+            irods_config.admin_password.encode("utf-8"),
+            salt=salt.encode("utf-8"),
             n=params["N"],
             r=params["r"],
             p=params["p"],
@@ -456,7 +456,7 @@ def setup_database_values(irods_config, cursor=None, default_resource_directory=
         irods_config.admin_session_token = str(uuid.uuid4())
         salted_token = salt + irods_config.admin_session_token
         # Hash it and base64-encode it so that it can be represented in the catalog.
-        hashed_session_token = base64.b64encode(hashlib.sha256(salted_token.encode()).digest()).decode("utf-8")
+        hashed_session_token = base64.b64encode(hashlib.sha256(salted_token.encode("utf-8")).digest()).decode("utf-8")
         execute_sql_statement(cursor,
                 "insert into R_USER_SESSION_KEY (user_id, session_key, auth_scheme, session_expiry_ts, create_ts, modify_ts, salt) values (?, ?, ?, ?, ?, ?, ?);",
                 admin_user_id,
--- a/scripts/irods/database_connect.py
+++ b/scripts/irods/database_connect.py
@@ -1,14 +1,19 @@
 from __future__ import print_function
 
+import base64
+import hashlib
 import itertools
 import json
 import logging
 import os
 import pprint
+import random
 import shutil
+import string
 import sys
 import tempfile
 import time
+import uuid
 
 from . import lib
 from . import password_obfuscation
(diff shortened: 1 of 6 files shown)


scripts/setup_irods.py
```diff
@@ -380,6 +399,83 @@ def setup_storage(irods_config):
...
+
+        default_ca_certificate_path = irods_config.server_config["tls_client"].get('ca_certificate_path')
+        client_ca_certificate_path = irods.lib.default_prompt(
+            "Path to CA certificates directory",
+            default=[default_ca_certificate_path] if default_ca_certificate_path else None)
+
+        default_verify_server = irods_config.server_config["tls_client"].get('verify_server') or "cert"
+        verify_server_levels = ["hostname", "cert", "none"]
+        client_verify_server = irods.lib.default_prompt(
+            "Certificate verification level",
```
trel, line 447: The commit message says 'Server verification level'.

I think it should be same as here... 'Certificate verification level'

scripts/setup_irods.py
```diff
@@ -380,6 +399,83 @@ def setup_storage(irods_config):
...
+
+        confirmation_message = (
+            '\n'
+            '-------------------------------------------------------------\n'
+            f'Certificate chain file:         {irods_config.server_config["tls_server"].get("certificate_chain_file")}\n'
+            f'Certificate key file:           {irods_config.server_config["tls_server"].get("certificate_key_file")}\n'
+            f'Diffie-Hellman parameters file: {irods_config.server_config["tls_server"].get("dh_params_file")}\n'
+            f'CA certificate file:            {irods_config.server_config["tls_client"].get("ca_certificate_file", "")}\n'
+            f'CA certificate path:            {irods_config.server_config["tls_client"].get("ca_certificate_path", "")}\n'
+            f'Client verify server level:     {irods_config.server_config["tls_client"].get("verify_server")}\n'
```
trel: similar to here.... 'Certificate verification level:' ?

scripts/setup_irods.py
```diff
@@ -142,6 +153,15 @@ def setup_server(irods_config, json_configuration_file=None, test_mode=False):
         if not os.path.exists(path):
             shutil.copyfile(irods.paths.get_template_filepath(path), path)
 
+    core_re_path = os.path.join(irods_config.core_re_directory, 'core.re')
+
+    # Update core.re to reflect configured TLS. This is done here because unattended installations will not use the TLS
+    # setup function.
+    if "CS_NEG_REQUIRE" == irods_config.server_config.get("client_server_policy", "CS_NEG_REFUSE"):
```
korydraughn: `client_server_policy` is about TLS for server-to-server connections, but it looks like it's being used to decide whether TLS should be involved for pure-client-to-server connections.

What happens if an admin wants server-to-server connections to be secure, but not pure-client-to-server connections?

scripts/setup_irods.py
alanking: Done

scripts/setup_irods.py
alanking: Will update commit message at squashing time.

scripts/irods/database_connect.py
```diff
@@ -412,15 +417,66 @@ def get_next_object_id():
...
+                "N": 16384,
+                "r": 8,
+                "p": 1,
+                "key_length": 64
+            }
+        }
+        params = hashing_parameters["parameters"]
+        derived_key = hashlib.scrypt(
+            irods_config.admin_password.encode(),
+            salt=salt.encode(),
```
korydraughn: Does `.encode()` need `"utf-8"`?

scripts/setup_irods.py
trel: good eye, forgot we had this discussion live on wedneday.   we figured defaulting to TLS out of the box was fine.  an admin who wants what you're suggesting would feel comfortable coming in and touching these after it was setup i think.

korydraughn (commented):
Does managing passwords for users other than the rodsadmin have any dependency on the rodsadmin's credentials, etc?

I'm thinking about how native auth relies on the invoking rodsadmin's information to change passwords, etc.

scripts/setup_irods.py
korydraughn: Got it. So, that part is a documentation exercise.

alanking: I wasn't sure what else I could use to "definitively" know when a server is using TLS or not. I think to support the hypothetical above, one would be using `CS_NEG_DONT_CARE`, which I thought we were trying to phase out. As @trel mentioned, one is always able to go back and modify these things later, too.

So what should we document, and where?

scripts/irods/database_connect.py
alanking: Technically, no. The first argument to `str.encode` defaults to `"utf-8"`: https://docs.python.org/3/library/stdtypes.html#str.encode

However, I guess I did specify `"utf-8"` in the `bytes.decode` calls nearby, which is technically also not needed for the same reason: https://docs.python.org/3/library/stdtypes.html#bytes.decode

Thoughts?

alanking:
> Does managing passwords for users other than the rodsadmin have any dependency on the rodsadmin's credentials, etc?
> 
> I'm thinking about how native auth relies on the invoking rodsadmin's information to change passwords, etc.

Sorry, I don't understand the question. Are you asking what happens if a rodsadmin using irods authentication tries to set a password for another user who is using native authentication?

scripts/irods/database_connect.py
korydraughn: Let's be explicit for those in case Python 3.50 or 4 suddenly decides to change things up on us.

korydraughn:
Let's say native auth is used for a zone only. When changing passwords on behalf of a user, the rodsadmin's .irodsA info is needed. Without that, the rodsadmin cannot change a user's password. See the following.

https://github.com/irods/irods/blob/27b72d946f6d27c20ea316658d9456f4853bda39/unit_tests/src/test_user_administration.cpp#L234-L239

I take it that the new irods auth scheme doesn't have a requirement like that. That is, a rodsadmin using the irods auth scheme can change the password of users freely.

> Are you asking what happens if a rodsadmin using irods authentication tries to set a password for another user who is using native authentication?

That is part of it, yes.

I suspect the rodsadmin needs native auth credentials to change native auth passwords of other users. Hmm, the rodsadmin would need to be using native auth too, perhaps?

scripts/setup_irods.py
korydraughn: The JSON payload for unattended installs can hold options for handling that. It does that for the service account user/group, admin password, etc. We can add one or two options to allow the admin to specify what they want to happen exactly.

https://github.com/irods/irods/blob/27b72d946f6d27c20ea316658d9456f4853bda39/schemas/configuration/v5/unattended_installation.json.in

alanking:
> I take it that the new irods auth scheme doesn't have a requirement like that. That is, a rodsadmin using the irods auth scheme can change the password of users freely.

That's correct as far as irods auth scheme goes.

> I suspect the rodsadmin needs native auth credentials to change native auth passwords of other users. Hmm, the rodsadmin would need to be using native auth too, perhaps?

If changing a user's native password requires the executing rodsadmin to have an .irodsA file and a corresponding password in `R_USER_PASSWORD`, then I suppose that rodsadmin would not be able to do so if using exclusively irods authentication.

I looked around in the tests and found that I only seem to have added tests around this wherein the executing rodsadmin is using native authentication to set other users' passwords (or at least *could* be using native authentication). I don't have any wherein the rodsadmin has only irods auth credentials.

scripts/irods/database_connect.py
alanking: Done

trel:
feels like temporary things could be created (and then destroyed) to satisfy the scenario of modifying a 'native' user with an 'irods-auth' admin?

scripts/setup_irods.py
alanking: After discussion, we have decided that the presence of a `tls_server` stanza is a stronger indicator of the server using TLS, so we will look for that instead. Alternatively, we can use a flag to track what has been configured.

alanking:
After some discussion, we have decided that the irods-only rodsadmin trying to change a native password for another user is possible to work around by creating another rodsadmin which has a native password and having that admin modify the password. We will handle this on a case-by-case basis if users encounter the situation because it is somewhat difficult to get into it and documentation would be noise for the vast majority of readers.

scripts/setup_irods.py
alanking: Just checking for `tls_server` stanza for now. Done

alanking:
After yet more discussion, we've decided that the setup script needs to give administrators the option of selecting a password storage mode so that the use of native authentication can be more easily enabled in a zone set up using irods authentication.

Additionally, we've decided that a session token should only be generated when a user authenticates with irods auth scheme. That means that if the administrator wants to store "both" types of passwords but selects the native auth scheme, the hashed password will be stored in addition to the legacy password, but no session key will be generated until the user tries to authenticate using the irods auth scheme.

I've developed a new prompt for this which sets the password storage mode grid configuration as well as determines whether to create or use an .irodsA file.

While working on this, I added a new prompt to automatically generate and use self-signed certs at setup time if the administrator does not already have one available to use for TLS. This was originally intended to ease testing of the irods auth setup prompts, but we decided it would be useful to include.

---

All of this has introduced significant changes, so I am contemplating force-pushing a newly minted set of commits rather than pushing the diffs. Only @korydraughn and @trel have looked at this so far, so will await the go-ahead or objections from yinz.

alanking:
We've decided to use a separate PR to review the new changes: https://github.com/irods/irods/pull/8950

alanking:
Putting this in draft for now. This will likely be closed.

PR #8942, opened 2026-04-24
Fix boot time tracking / Do not allow server to start if launched by incorrect user (main)

korydraughn:
I've verified that only the user identified in /etc/irods/service_account.config is able to start the server.

I still need to check non-package install and service manager cases. There's also a couple error cases which I need to determine how to handle.

### DIFF
--- a/scripts/irods/test/test_resource_types.py
+++ b/scripts/irods/test/test_resource_types.py
@@ -2235,9 +2235,11 @@ class Test_Resource_CompoundWithUnivmss(ChunkyDevTest, ResourceSuite, unittest.T
 
             # TODO(#8935): Remove this line when the default for "escape_single_quotes" changes in iRODS 6.
             with self.subTest('iput fails when path contains single quotes and escape_single_quotes not enabled'):
-                self.user0.assert_icommand(['iput', file_path], 'STDERR', ['-550000 UNIV_MSS_SYNCTOARCH_ERR'])
+                self.user0.assert_icommand(['iput', '-R', 'demoResc', file_path], 'STDERR', [r'-550\d{3} UNIV_MSS_SYNCTOARCH_ERR'], use_regex=True)
                 self.assertEqual(lib.get_replica_status_for_resource(self.user0, logical_path, 'archiveResc'), '0')
-                self.user0.assert_icommand(['irm', '-f', filename], 'STDERR', ['-552000 UNIV_MSS_UNLINK_ERR'])
+                # We check the first three digits of the error code because the last three digits can change
+                # due to the inclusion of errno.
+                self.user0.assert_icommand(['irm', '-f', filename], 'STDERR', [r'-552\d{3} UNIV_MSS_UNLINK_ERR'], use_regex=True)
 
             # TODO(#8935): Remove this line when the default for "escape_single_quotes" changes in iRODS 6.
             self.admin.assert_icommand(['iadmin', 'modresc', 'archiveResc', 'context', 'script=univMSSInterface.sh;escape_single_quotes=1;'])
@@ -2249,7 +2251,7 @@ class Test_Resource_CompoundWithUnivmss(ChunkyDevTest, ResourceSuite, unittest.T
             with self.subTest('Sync-to-Archive - iput'):
                 # Create a new data object. The bug normally produces an error here. This should succeed
                 # now that single quotes are escaped.
-                self.user0.assert_icommand(['iput', file_path])
+                self.user0.assert_icommand(['iput', '-R', 'demoResc', file_path])
 
                 # Verify that all replicas are good.
                 self.assertEqual(lib.get_replica_status_for_resource(self.user0, logical_path, 'cacheResc'), '1')
@@ -2258,21 +2260,37 @@ class Test_Resource_CompoundWithUnivmss(ChunkyDevTest, ResourceSuite, unittest.T
                 # Remove the data object for the next subtest.
                 self.user0.assert_icommand(['irm', '-f', filename])
 
-            with self.subTest('Sync-to-Archive - itouch'):
-                self.user0.assert_icommand(['itouch', filename])
+            # Skip this itouch case when running in a topology from a catalog consumer.
+            #
+            # itouch does not support targeting the root of a resource hierarchy containing one or
+            # more child resources. It can only be used to target leaf resources.
+            #
+            # This subtest must be run from a catalog provider for the following reasons:
+            # - The default resource of the catalog provider is demoResc (i.e the compound resource)
+            # - Running itouch without specifying a leaf resource results in the connected server using its default resource
+            if not test.settings.TOPOLOGY_FROM_RESOURCE_SERVER:
+                with self.subTest('Sync-to-Archive - itouch + default resource for provider'):
+                    self.user0.assert_icommand(['itouch', filename])
+                    self.assertEqual(lib.get_replica_status_for_resource(self.user0, logical_path, 'cacheResc'), '1')
+                    self.assertEqual(lib.get_replica_status_for_resource(self.user0, logical_path, 'archiveResc'), '1')
+                    self.user0.assert_icommand(['irm', '-f', filename])
+
+            with self.subTest('Sync-to-Archive - itouch + rebalance'):
+                self.user0.assert_icommand(['itouch', '-R', 'cacheResc', filename])
+                self.admin.assert_icommand(['iadmin', 'modresc', 'demoResc', 'rebalance'])
                 self.assertEqual(lib.get_replica_status_for_resource(self.user0, logical_path, 'cacheResc'), '1')
                 self.assertEqual(lib.get_replica_status_for_resource(self.user0, logical_path, 'archiveResc'), '1')
                 self.user0.assert_icommand(['irm', '-f', filename])
 
             with self.subTest('Sync-to-Archive - istream write'):
-                self.user0.assert_icommand(['istream', 'write', filename], input='data')
+                self.user0.assert_icommand(['istream', 'write', '-R', 'demoResc', filename], input='data')
                 self.assertEqual(lib.get_replica_status_for_resource(self.user0, logical_path, 'cacheResc'), '1')
                 self.assertEqual(lib.get_replica_status_for_resource(self.user0, logical_path, 'archiveResc'), '1')
                 self.user0.assert_icommand(['irm', '-f', filename])
 
             # These subtests verify the stage-to-cache functionality behaves as expected.
             with self.subTest('Stage-to-Cache - iget'):
-                self.user0.assert_icommand(['itouch', filename])
+                self.user0.assert_icommand(['iput', '-R', 'demoResc', file_path])
                 self.user0.assert_icommand(['itrim', '-N1', '-n0', filename], 'STDOUT', ['data objects trimmed = 1.'])
                 self.user0.assert_icommand(['iget', filename, '-'], 'STDOUT')
                 self.assertEqual(lib.get_replica_status_for_resource(self.user0, logical_path, 'cacheResc'), '1')
@@ -2288,16 +2306,16 @@ class Test_Resource_CompoundWithUnivmss(ChunkyDevTest, ResourceSuite, unittest.T
                 self.user0.assert_icommand(['itrim', '-N1', '-n2', filename], 'STDOUT', ['data objects trimmed = 1.'])
 
                 # Show that we must use the force flag to overwrite the data object.
-                self.user0.assert_icommand(['iput', file_path], 'STDERR', ['-312000 OVERWRITE_WITHOUT_FORCE_FLAG'])
+                self.user0.assert_icommand(['iput', '-R', 'demoResc', file_path], 'STDERR', ['-312000 OVERWRITE_WITHOUT_FORCE_FLAG'])
 
-                self.user0.assert_icommand(['iput', '-f', file_path], 'STDOUT')
+                self.user0.assert_icommand(['iput', '-f', '-R', 'demoResc', file_path], 'STDOUT')
                 self.assertEqual(lib.get_replica_status_for_resource(self.user0, logical_path, 'cacheResc'), '1')
                 self.assertEqual(lib.get_replica_status_for_resource(self.user0, logical_path, 'archiveResc'), '1')
 
             # This subtest is specifically about renaming data objects.
             with self.subTest('Rename'):
                 self.user0.assert_icommand(['irm', '-f', filename])
-                self.user0.assert_icommand(['itouch', filename])
+                self.user0.assert_icommand(['iput', '-R', 'demoResc', file_path])
                 self.assertEqual(lib.get_replica_status_for_resource(self.user0, logical_path, 'cacheResc'), '1')
                 self.assertEqual(lib.get_replica_status_for_resource(self.user0, logical_path, 'archiveResc'), '1')
 
--- a/server/main_server/src/main.cpp
+++ b/server/main_server/src/main.cpp
@@ -22,6 +22,7 @@
 #include "irods/rcGlobalExtern.h" // For ProcessType
 #include "irods/rcMisc.h"
 #include "irods/rodsClient.h"
+#include "irods/rodsDef.h"
 #include "irods/rodsErrorTable.h"
 #include "irods/rodsLog.h"
 #include "irods/update_replica_access_time.h"
@@ -133,6 +134,7 @@ namespace
     auto print_usage() -> void;
     auto print_version_info() -> void;
 
+    auto set_boot_time_as_environment_variable() -> void;
     auto validate_configuration() -> bool;
     auto daemonize() -> void;
     auto create_pid_file(const std::string& _pid_file) -> int;
@@ -166,7 +168,7 @@ namespace
 
 auto main(int _argc, char* _argv[]) -> int
 {
-    [[maybe_unused]] const auto boot_time = std::chrono::system_clock::now();
+    set_boot_time_as_environment_variable();
 
     bool write_to_stdout = false;
     bool enable_test_mode = false;
@@ -440,6 +442,28 @@ Signals:
         fmt::print(
             "irodsServer v{}.{}.{}-{}\n", IRODS_VERSION_MAJOR, IRODS_VERSION_MINOR, IRODS_VERSION_PATCHLEVEL, commit);
     } // print_version_info
+
+    // This function is meant to be called before the logger is initialized.
+    auto set_boot_time_as_environment_variable() -> void
+    {
+        try {
+            using std::chrono::duration_cast;
+            using std::chrono::seconds;
+            using std::chrono::system_clock;
+
+            const auto now = duration_cast<seconds>(system_clock::now().time_since_epoch());
+            const auto now_string = std::to_string(now.count());
+
+            // Set the boot time as an environment variable so that rsGetMiscSvrInfo can report it
+            // to the client. If there's a failure, notify the administrator and continue the boot process.
+            if (setenv(SERVER_BOOT_TIME, now_string.c_str(), 1) != 0) {
+                fmt::print(stderr, "Warning: Failed to set server boot time as an environment variable.\n");
+            }
+        }
+        catch (const std::exception& e) {
+            fmt::print(stderr, "Error: Caught exception while setting server boot time as an environment variable: {}\n", e.what());
+        }
+    } // set_boot_time_as_environment_variable
 
     auto validate_configuration() -> bool
     {
--- a/server/main_server/src/main.cpp
+++ b/server/main_server/src/main.cpp
@@ -235,7 +235,6 @@ auto main(int _argc, char* _argv[]) -> int
         }
 
         if (create_pid_file(pid_file) != 0) {
-            fmt::print(stderr, "Error: could not create PID file [{}].\n", pid_file);
             return 1;
         }
     }
@@ -661,7 +660,7 @@ Signals:
         // NOLINTNEXTLINE(cppcoreguidelines-pro-type-vararg, hicpp-signed-bitwise)
         const auto fd = open(_pid_file.c_str(), O_RDWR | O_CREAT, S_IRUSR | S_IWUSR);
         if (fd == -1) {
-            fmt::print("Could not open PID file.\n");
+            fmt::print(stderr, "Error: Could not open PID file.\n");
             return 1;
         }
 
@@ -669,7 +668,7 @@ Signals:
         // NOLINTNEXTLINE(cppcoreguidelines-pro-type-vararg)
         const auto flags = fcntl(fd, F_GETFD);
         if (flags == -1) {
-            fmt::print("Could not retrieve open flags for PID file.\n");
+            fmt::print(stderr, "Error: Could not retrieve open flags for PID file.\n");
             return 1;
         }
 
@@ -678,7 +677,7 @@ Signals:
         // Keep in mind that record locks are NOT inherited by forked child processes.
         // NOLINTNEXTLINE(cppcoreguidelines-pro-type-vararg, hicpp-signed-bitwise)
         if (fcntl(fd, F_SETFD, flags | FD_CLOEXEC) == -1) {
-            fmt::print("Could not set FD_CLOEXEC on PID file.\n");
+            fmt::print(stderr, "Error: Could not set FD_CLOEXEC on PID file.\n");
             return 1;
         }
 
@@ -695,21 +694,22 @@ Signals:
         // NOLINTNEXTLINE(cppcoreguidelines-pro-type-vararg)
         if (fcntl(fd, F_SETLK, &input) == -1) {
             if (EAGAIN == errno || EACCES == errno) {
-                fmt::print("Could not acquire write lock for PID file. Another instance "
-                           "could be running already.\n");
+                fmt::print(
+                    stderr,
+                    "Error: Could not acquire write lock for PID file. Another instance could be running already.\n");
                 return 1;
             }
         }
 
         if (ftruncate(fd, 0) == -1) {
-            fmt::print("Could not truncate PID file's contents.\n");
+            fmt::print(stderr, "Error: Could not truncate PID file's contents.\n");
             return 1;
         }
 
         const auto contents = fmt::format("{}\n", getpid());
         // NOLINTNEXTLINE(google-runtime-int)
         if (write(fd, contents.data(), contents.size()) != static_cast<long>(contents.size())) {
-            fmt::print("Could not write PID to PID file.\n");
+            fmt::print(stderr, "Error: Could not write PID to PID file.\n");
             return 1;
         }
 
--- a/server/main_server/src/main.cpp
+++ b/server/main_server/src/main.cpp
@@ -135,6 +135,8 @@ namespace
     auto print_version_info() -> void;
 
     auto set_boot_time_as_environment_variable() -> void;
+    auto terminate_if_launched_with_root_privileges() -> void;
+    auto warn_if_launched_by_non_owner() -> void;
     auto validate_configuration() -> bool;
     auto daemonize() -> void;
     auto create_pid_file(const std::string& _pid_file) -> int;
@@ -207,6 +209,9 @@ auto main(int _argc, char* _argv[]) -> int
             return 0;
         }
 
+        terminate_if_launched_with_root_privileges();
+        warn_if_launched_by_non_owner();
+
         if (vm.count("validate-config") > 0) {
             return validate_configuration() ? 0 : 1;
         }
@@ -251,6 +256,13 @@ auto main(int _argc, char* _argv[]) -> int
         rodsLogSqlReq(0);
 
         init_logger(getpid(), write_to_stdout, enable_test_mode);
+
+        log_server::info("{}: Server launched by real UID=[{}], real GID=[{}], effective UID=[{}], effective GID=[{}].",
+                         __func__,
+                         getuid(),
+                         getgid(),
+                         geteuid(),
+                         getegid());
 
         // Setting up signal handlers here removes the need for reacting to shutdown signals
         // such as SIGINT and SIGTERM during the startup sequence.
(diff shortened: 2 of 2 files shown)


server/main_server/src/main.cpp
```diff
@@ -441,6 +447,96 @@ Mandatory arguments to long options are mandatory for short options too.
...
+
+    auto terminate_if_launched_by_incorrect_system_user() -> void
+    {
+        try {
+            const auto service_acct_file_path = irods::get_irods_config_directory() / "service_account.config";
+            std::ifstream in{service_acct_file_path};
+            if (!in) {
+                // A bad input stream signals that the user may not have the proper permissions to open
+                // the file.
+            }
```
korydraughn: Handle bad stream case.

server/main_server/src/main.cpp
```diff
@@ -441,6 +447,96 @@ Mandatory arguments to long options are mandatory for short options too.
...
+
+            std::string line;
+            while (in && std::getline(in, line)) {
+                if (line.empty()) {
+                    // TODO Do nothing?
+                }
```
korydraughn: Handle TODO.

server/main_server/src/main.cpp
```diff
@@ -441,6 +447,96 @@ Mandatory arguments to long options are mandatory for short options too.
...
+                auto buffer = std::make_unique<char[]>(buffer_size);
+
+                // Lookup the service account user's information. This information comes form the OS.
+                passwd pwd; // NOLINT(cppcoreguidelines-pro-type-member-init)
+                passwd* result{};
+                const auto ec = getpwnam_r(value.c_str(), &pwd, buffer.get(), buffer_size, &result);
+                if (nullptr == result) {
+                    if (ec == 0) {
+                        fmt::print(stderr, "Warning: Service account user [{}] not found in system.\n", value);
```
korydraughn: State that the server is exiting because of this.

Apply to the other failure cases too.

server/main_server/src/main.cpp
```diff
@@ -441,6 +447,96 @@ Mandatory arguments to long options are mandatory for short options too.
             "irodsServer v{}.{}.{}-{}\n", IRODS_VERSION_MAJOR, IRODS_VERSION_MINOR, IRODS_VERSION_PATCHLEVEL, commit);
     } // print_version_info
 
+    auto set_boot_time_as_environment_variable() -> void
+    {
+        try {
+            const auto now = std::to_string(std::time(nullptr));
+
+            // Set the boot time as an environment so that rsGetMiscSvrInfo can report it
```
korydraughn:
```suggestion
            // Set the boot time as an environment variable so that rsGetMiscSvrInfo can report it
```

server/main_server/src/main.cpp
```diff
@@ -441,6 +447,96 @@ Mandatory arguments to long options are mandatory for short options too.
             "irodsServer v{}.{}.{}-{}\n", IRODS_VERSION_MAJOR, IRODS_VERSION_MINOR, IRODS_VERSION_PATCHLEVEL, commit);
     } // print_version_info
 
+    auto set_boot_time_as_environment_variable() -> void
```
korydraughn, line 458: Add a comment to let future devs know that this is designed to be called before the logging system is initialized.

server/main_server/src/main.cpp
```diff
@@ -441,6 +447,96 @@ Mandatory arguments to long options are mandatory for short options too.
...
+        }
+        catch (const std::exception& e) {
+            fmt::print(stderr,
+                       "Warning: An exception was thrown while setting server boot time as an environment variable. "
+                       "Exception: {}\n",
+                       e.what());
+        }
+    } // set_boot_time_as_environment_variable
+
+    auto terminate_if_launched_by_incorrect_system_user() -> void
```
korydraughn: Add a comment to let future devs know that this is designed to be called before the logging system is initialized.

server/main_server/src/main.cpp
```diff
@@ -441,6 +447,96 @@ Mandatory arguments to long options are mandatory for short options too.
...
 
+    auto set_boot_time_as_environment_variable() -> void
+    {
+        try {
+            const auto now = std::to_string(std::time(nullptr));
+
+            // Set the boot time as an environment so that rsGetMiscSvrInfo can report it
+            // to the client. If there's a failure, notify the administrator and continue
+            // the boot process.
+            if (setenv(SERVER_BOOT_TIME, now.c_str(), 1) != 0) {
```
trel: perhaps?

```suggestion
            if (setenv(IRODS_SERVER_BOOT_TIME, now.c_str(), 1) != 0) {
```

server/main_server/src/main.cpp
```diff
@@ -441,6 +447,96 @@ Mandatory arguments to long options are mandatory for short options too.
...
+                // Allocate a buffer large enough to hold the strings of the passwd struct.
+                auto buffer_size = sysconf(_SC_GETPW_R_SIZE_MAX);
+                if (buffer_size == -1) {
+                    // Indeterminate value. Set the buffer to a value large enough to hold the strings
+                    // of the passwd struct.
+                    buffer_size = 16384;
+                }
+                auto buffer = std::make_unique<char[]>(buffer_size);
+
+                // Lookup the service account user's information. This information comes form the OS.
```
trel:
```suggestion
                // Lookup the service account user's information. This information comes from the OS.
```

server/main_server/src/main.cpp
```diff
@@ -441,6 +447,96 @@ Mandatory arguments to long options are mandatory for short options too.
             "irodsServer v{}.{}.{}-{}\n", IRODS_VERSION_MAJOR, IRODS_VERSION_MINOR, IRODS_VERSION_PATCHLEVEL, commit);
     } // print_version_info
 
+    auto set_boot_time_as_environment_variable() -> void
+    {
+        try {
+            const auto now = std::to_string(std::time(nullptr));
```
trel: notice this was a `std::chrono::system_clock::now()` before.

and that there is no code here that uses this yet?

server/main_server/src/main.cpp
```diff
@@ -464,6 +467,76 @@ Mandatory arguments to long options are mandatory for short options too.
...
+                        euid,
+                        pwd.pw_uid,
+                        value);
+                    std::exit(1);
+                }
+            }
+        }
+        catch (const std::exception& e) {
+            fmt::print(stderr,
+                       "Warning: An exception was thrown while setting server boot time as an environment variable. "
```
alanking: I think this message should say something about determining the system user attempting to start the server.

server/main_server/src/main.cpp
```diff
@@ -464,6 +467,76 @@ Mandatory arguments to long options are mandatory for short options too.
...
     } // set_boot_time_as_environment_variable
 
+    auto terminate_if_launched_by_incorrect_system_user() -> void
+    {
+        try {
+            const auto service_acct_file_path = irods::get_irods_config_directory() / "service_account.config";
+            std::ifstream in{service_acct_file_path};
+            if (!in) {
+                // A bad input stream signals that the user may not have the proper permissions to open
+                // the file.
```
alanking: If this happens, we never enter the `while` loop below and it does not terminate the server process. Is that what we want? It seems more likely than not that the wrong user is trying to launch the server if the user cannot open this file due to permissions.

server/main_server/src/main.cpp
```diff
@@ -464,6 +467,76 @@ Mandatory arguments to long options are mandatory for short options too.
...
+                // Allocate a buffer large enough to hold the strings of the passwd struct.
+                auto buffer_size = sysconf(_SC_GETPW_R_SIZE_MAX);
+                if (buffer_size == -1) {
+                    // Indeterminate value. Set the buffer to a value large enough to hold the strings
+                    // of the passwd struct.
+                    buffer_size = 16384;
+                }
+                auto buffer = std::make_unique<char[]>(buffer_size);
+
+                // Lookup the service account user's information. This information comes form the OS.
```
alanking:
```suggestion
                // Lookup the service account user's information. This information comes from the OS.
```

server/main_server/src/main.cpp
```diff
@@ -464,6 +467,76 @@ Mandatory arguments to long options are mandatory for short options too.
...
+                passwd pwd; // NOLINT(cppcoreguidelines-pro-type-member-init)
+                passwd* result{};
+                const auto ec = getpwnam_r(value.c_str(), &pwd, buffer.get(), buffer_size, &result);
+                if (nullptr == result) {
+                    if (ec == 0) {
+                        fmt::print(stderr, "Warning: Service account user [{}] not found in system.\n", value);
+                    }
+                    else {
+                        errno = ec;
+                        perror("getpwnam_r");
```
alanking: Do we know what this looks like when it happens? Maybe we could additionally print some identifiable message here from "us" (e.g. "An error occurred while getting service account user information.") so that we can later easily identify what happened should this occur.

server/main_server/src/main.cpp
```diff
@@ -464,6 +467,76 @@ Mandatory arguments to long options are mandatory for short options too.
...
+            std::ifstream in{service_acct_file_path};
+            if (!in) {
+                // A bad input stream signals that the user may not have the proper permissions to open
+                // the file.
+            }
+
+            std::string line;
+            while (in && std::getline(in, line)) {
+                if (line.empty()) {
+                    // TODO Do nothing?
```
alanking: TODO here. Would this indicate that the file is malformed?

server/main_server/src/main.cpp
korydraughn: I'll switch back to chrono.

> and that there is no code here that uses this yet?

It's not clear to me what you mean. `now` is read by `setenv`, so it is being used?

server/main_server/src/main.cpp
korydraughn: You're proposing we introduce (or rename if possible) a new macro?

server/main_server/src/main.cpp
trel: it wasn't clear what code is using this value anywhere.   are we printing out the time in the log?  or using it to calculate something else?

server/main_server/src/main.cpp
trel: is it a macro?   or just an environment variable?    sorry, didn't know this was already in use anywhere... oh, i'm now seeing rsGetMiscSvrInfo is mentioned.      resolve if you want.

server/main_server/src/main.cpp
korydraughn: The plan is for the server to stop immediately if the file cannot be opened. That part is on the way.

server/main_server/src/main.cpp
korydraughn: I haven't thought very much about this case yet, but I agree with your logic. We control the structure of the file and therefore, an empty line should not be encountered.

server/main_server/src/main.cpp
korydraughn: Based on the man pages, the message will look something like this.

```
getpwnam_r: message containing info describing errno
```

The perror thing comes from the man page demonstrating getpwnam_r, so yes, we can print our own message to help in debugging.

Will update.

server/main_server/src/main.cpp
korydraughn: Yeah, definitely a copy-paste situation.

server/main_server/src/main.cpp
korydraughn: Okay. Resolving.

server/main_server/src/main.cpp
korydraughn: `rsGetMiscSvrInfo` relies on the value to report uptime. Nothing in the main server process relies on it. I don't think anything other than the API endpoint I mentioned uses it.

server/main_server/src/main.cpp

server/main_server/src/main.cpp
korydraughn: Removed.

server/main_server/src/main.cpp

server/main_server/src/main.cpp

server/main_server/src/main.cpp
korydraughn: Done.

server/main_server/src/main.cpp

server/main_server/src/main.cpp

server/main_server/src/main.cpp

server/main_server/src/main.cpp

server/main_server/src/main.cpp

server/main_server/src/main.cpp

server/main_server/src/main.cpp

server/main_server/src/main.cpp
korydraughn: Updated to use chrono lib.

alanking (commented):
Seems to all check out. Anything else to do here?

korydraughn:
I think the implementation is good. All that's left is confirming everything works as intended.

I'm going to try to add some automated tests.

korydraughn:
Here's some examples of what the current set of changes produce.

### 1. Launching the server as root
```
root@b5b2065bf5f7:/# irodsServer -d
Warning: UID [0] of server process does not match UID [999] of service account user [irods]. Exiting.
```

### 2. Launching the server under the wrong user

The testing environment would need to create another user before this can be put into a test. Will update the testing environment to do that.

This covers the root user case, but having a real test involving the root user is much better.

```
root@b5b2065bf5f7:/# su - john 
john@b5b2065bf5f7:~$ irodsServer
Warning: UID [1000] of server process does not match UID [999] of service account user [irods]. Exiting.
```

### 3. Removing read permissions on service_account.config and launching the server under the service account
```
root@b5b2065bf5f7:/# ls -l /etc/irods/service_account.config 
-rw-r--r-- 1 irods irods 64 Apr 27 19:48 /etc/irods/service_account.config
root@b5b2065bf5f7:/# chmod -r /etc/irods/service_account.config 
root@b5b2065bf5f7:/# ls -l /etc/irods/service_account.config 
--w------- 1 irods irods 64 Apr 27 19:48 /etc/irods/service_account.config
root@b5b2065bf5f7:/# su - irods
irods@b5b2065bf5f7:~$ irodsServer -d
Error: Failed to open service account file [/etc/irods/service_account.config]. Cannot determine if user has permission to launch server. Exiting.
```

### 4. Changing the name recorded in service_account.config and launching the server under the original service account
```
irods@b5b2065bf5f7:~$ cat /etc/irods/service_account.config 
IRODS_SERVICE_ACCOUNT_NAME=nope
IRODS_SERVICE_GROUP_NAME=irods
irods@b5b2065bf5f7:~$ irodsServer -d
Warning: Service account user [nope] not found in system. Exiting.
```

I still need to verify the service manager and non-package install cases.

scripts/irods/test/test_misc.py
```diff
@@ -601,6 +602,56 @@ def pep_database_gen_query_access_control_setup_except(rule_args, callback, rei)
...
+            root_user = 'root'
+
+            # Change the username in /etc/irods/service_account.config to root. The value
+            # assigned to IRODS_SERVICE_ACCOUNT_NAME is what the server cares about.
+            # IRODS_SERVICE_ACCOUNT_GROUP_NAME is ignored.
+            with open(paths.service_account_file_path(), 'wt') as f:
+                f.write(f'IRODS_SERVICE_ACCOUNT_NAME={root_user}\n')
+                f.write('IRODS_SERVICE_ACCOUNT_GROUP_NAME=irods\n')
+
+            # Show the server exits immediately due to the user not existing in the system.
```
korydraughn: Fix comment.

scripts/irods/test/test_misc.py
```diff
@@ -601,6 +602,56 @@ def pep_database_gen_query_access_control_setup_except(rule_args, callback, rei)
...
+    def test_server_exits_immediately_when_service_account_config_file_cannot_be_accessed__issue_8587(self):
+        # Remember the current mode of the file. It will be used to restore the permissions.
+        st_mode = os.stat(paths.service_account_file_path()).st_mode
+
+        try:
+            # Remove read permissions on /etc/irods/service_account.config so that the service
+            # account user cannot read the file.
+            os.chmod(paths.service_account_file_path(), st_mode & ~(stat.S_IRUSR | stat.S_IRGRP | stat.S_IROTH))
+
+            # Show the server exits immediately due to the user not existing in the system.
```
korydraughn: Fix comment.

scripts/irods/test/test_misc.py
korydraughn: Done.

scripts/irods/test/test_misc.py
korydraughn: Done.

korydraughn:
I've captured the four test cases shown in https://github.com/irods/irods/pull/8942#issuecomment-4330106690. Test case 1 is simulated because we don't have root access from within the test suite. I think that's ok for now given the fact the C++ code resolves the usernames to IDs before comparing them.

I've verified that service manager support still works.

I've also verified that creating a name mismatch resulted in the iRODS server terminating as intended. The steps taken are below.
- Copy irods.service.pre-253.template to /etc/systemd/system/irods.service
  - My Ubuntu 22.04 system doesn't like `Type=notify-reload` because it's running systemd v249
- Set `User=kory` in irods.service
- Ran `sudo systemctl start irods.service`

Surprisingly, the messages written to stderr appeared in the `tail -f` output of the log file. No different than when config validation fails. I think this is in a good spot.

Unit tests pass. Running core tests now. Will switch to non-package install if core tests pass.

trel:
very nice.

korydraughn:
All core tests passed.

scripts/irods/test/test_misc.py
```diff
@@ -601,6 +602,57 @@ def pep_database_gen_query_access_control_setup_except(rule_args, callback, rei)
...
+
+            # Change the username in /etc/irods/service_account.config to a name that does not
+            # exist in the system/OS. The value assigned to IRODS_SERVICE_ACCOUNT_NAME is what
+            # the server cares about. IRODS_SERVICE_ACCOUNT_GROUP_NAME is ignored.
+            with open(paths.service_account_file_path(), 'wt') as f:
+                f.write(f'IRODS_SERVICE_ACCOUNT_NAME={nonexistent_user}\n')
+                f.write('IRODS_SERVICE_ACCOUNT_GROUP_NAME=irods\n')
+
+            # Show the server exits immediately due to the user not existing in the system.
+            res = subprocess.run(['irodsServer', '-d'], capture_output=True, text=True)
```
alanking: Is there much value in asserting that the server did not, in fact, launch?

alanking (commented):
Just one suggestion which can be ignored

scripts/irods/test/test_misc.py
korydraughn: Can do that. Will take a look.

korydraughn:
Non-package install testing is blocked until #8946 is resolved.

scripts/irods/test/test_misc.py
korydraughn: After re-reading the comment, I don't think it's necessary because we assert the exit code is nonzero and the only way to get that is by the server binary terminating.

Thoughts?

alanking: Seems reasonable. Resolving

korydraughn:
I've confirmed this PR works for non-package installs.

See https://github.com/irods/irods/issues/8946#issuecomment-4340360273 for more details.

alanking (commented):
Squash at your ready

korydraughn:
After some discussion we have decided not to terminate the startup process when the wrong user account is used. Instead, we'll log the UID and GID so that the admin can see it.

The motivation behind this decision is to one day allow a single install to run on the same computer, but with different configuration files. This would be similar to how the iRODS HTTP API works.

alanking:
> After some discussion **we have decided not to terminate the startup process when the wrong user account is used**. Instead, we'll log the UID and GID so that the admin can see it.

This issue - which was opened more recently than the comment post time - seems to contradict this: https://github.com/irods/irods/issues/8960 

> If the server is launched using the root user, the server should report it and exit immediately.

Or is that issue related specifically to the root user?

korydraughn:
#8960 is specifically about the root user.

The original implementation disallowed root, which I feel we should keep since iRODS doesn't need root to run / do its job. Disallowing root is common in many server applications too.

server/main_server/src/main.cpp
```diff
@@ -441,6 +455,72 @@ Mandatory arguments to long options are mandatory for short options too.
...
+        // Guard against setuid-related security vulnerabilities.
+        if (getuid() != geteuid()) {
+            fmt::print(
+                stderr, "Warning: Real UID [{}] and effective UID [{}] must match. Exiting.\n", getuid(), geteuid());
+            std::exit(1);
+        }
+    } // terminate_if_launched_with_root_privileges
+
+    // This function is meant to be called before the logger is initialized.
+    auto warn_if_launched_by_unexpected_user() -> void
```
korydraughn: Might change the name of this to one of the following:
- `warn_if_launched_by_non_owner`
- `warn_if_launched_user_differs_from_configuration_owner`

server/main_server/src/main.cpp
```diff
@@ -441,6 +455,72 @@ Mandatory arguments to long options are mandatory for short options too.
...
+    {
+        const auto home_dir = irods::get_irods_home_directory();
+        struct stat stat_info; // NOLINT(cppcoreguidelines-pro-type-member-init)
+        if (stat(home_dir.c_str(), &stat_info) != 0) {
+            fmt::print(stderr,
+                       "Warning: Stat of service account home directory [{}] failed. Cannot verify if user who "
+                       "launched the server is appropriate.\n",
```
korydraughn: It's not the worst message, but it can be improved.

server/main_server/src/main.cpp
```diff
@@ -441,6 +455,72 @@ Mandatory arguments to long options are mandatory for short options too.
...
+        }
+
+        if (!S_ISDIR(stat_info.st_mode)) {
+            fmt::print(
+                stderr, "Warning: [{}] must be the home directory of the service account user.\n", home_dir.c_str());
+        }
+
+        if (const auto uid = getuid(); uid != stat_info.st_uid) {
+            fmt::print(stderr,
+                       "Warning: Server launched by user with UID [{}]. Expected user with UID [{}].\n",
```
korydraughn:
```suggestion
                       "Warning: Server launched by UID [{}]. Expected UID [{}].\n",
```

server/main_server/src/main.cpp
```diff
@@ -250,6 +256,14 @@ auto main(int _argc, char* _argv[]) -> int
 
         init_logger(getpid(), write_to_stdout, enable_test_mode);
 
+        log_server::info(
+            "{}: Launched by user with real UID=[{}], real GID=[{}], effective UID=[{}], effective GID=[{}].",
```
korydraughn:
```suggestion
            "{}: Launched by real UID=[{}], real GID=[{}], effective UID=[{}], effective GID=[{}].",
```

server/main_server/src/main.cpp
korydraughn: Renamed to `warn_if_launched_by_non_owner`.

server/main_server/src/main.cpp
korydraughn: Adjusted wording.

server/main_server/src/main.cpp

server/main_server/src/main.cpp

SwooshyCueb (approved):

alanking (commented):
Squash when ready

korydraughn:
Squashed, but no pounds.

Running tests to make sure things are still good.

alanking (commented):
Awaiting signal for approval

korydraughn:
Unit tests passed.

Running core tests.

korydraughn:
Core tests passed.

Running topology+consumer tests.

Will verify systemd is still working after that.

korydraughn:
Topology+consumer had two test failures.
- (A) `test_resource_types.Test_Resource_ReplicationToTwoCompoundResourcesWithPreferArchive`
- (B) `test_resource_types.Test_Resource_CompoundWithUnivmss`

They are the same tests which @FifthPotato reported. I suspect they failed in the way too.

The reasons for the failures do not appear to be related to these changes. I re-ran both tests. (A) passed on the second run. (B) did not. (B)'s failures are due to the error code changing. The error code changes because errno is included.

```
Assert Command: irm -f issue'8928.txt
Expecting STDERR: ['-552000 UNIV_MSS_UNLINK_ERR']
  stdout:
    |  
  stderr:
    | remote addresses: 172.18.0.9 ERROR: rmUtil: rm error for /tempZone/home/alice/2026-05-06Z12:15:13--irods-testing-243g3qp_/issue'8928.txt, status = -552115 status = -552115 UNIV_MSS_UNLINK_ERR, Operation now in progress
Output not found
        
FAILED TESTING ASSERTION
```

I'll include a fix for that.

korydraughn:
I've also verified that the changes work with the service manager too.

scripts/irods/test/test_resource_types.py
```diff
@@ -2237,7 +2237,9 @@ def test_univmss_resource_plugin_supports_paths_containing_single_quotes__issue_
             with self.subTest('iput fails when path contains single quotes and escape_single_quotes not enabled'):
                 self.user0.assert_icommand(['iput', file_path], 'STDERR', ['-550000 UNIV_MSS_SYNCTOARCH_ERR'])
                 self.assertEqual(lib.get_replica_status_for_resource(self.user0, logical_path, 'archiveResc'), '0')
-                self.user0.assert_icommand(['irm', '-f', filename], 'STDERR', ['-552000 UNIV_MSS_UNLINK_ERR'])
+                # We check the first three digits of the error code because the last three digits can change
+                # due to the inclusion of errno.
```
alanking, line 2241: Do we know why this is happening? Is it expected/consistent? Should we change the assertion based on where the test is running?

korydraughn: I don't think it's necessary. The errno component isn't important for this particular test. It's more about the base iRODS error code.

korydraughn: The assertion isn't as strong as I'd like it, so I'm going to change it to use a regex.

korydraughn: Updated test.

scripts/irods/test/test_resource_types.py
```diff
@@ -2260,21 +2260,29 @@ def test_univmss_resource_plugin_supports_paths_containing_single_quotes__issue_
...
-                self.assertEqual(lib.get_replica_status_for_resource(self.user0, logical_path, 'archiveResc'), '1')
-                self.user0.assert_icommand(['irm', '-f', filename])
+            # Skip itouch case when running in a topology from a catalog consumer.
+            #
+            # itouch intentionally ignores the default resource property in irods_environment.json.
+            # This test case will fail because the catalog consumer will use its local resource as
+            # the destination resource, resulting in the new replica not being replicated to the
+            # compound resource. Remember, this test suite modifies demoResc, which is the default
+            # for the catalog provider, not the catalog consumers.
+            if test.settings.RUN_IN_TOPOLOGY and not test.settings.TOPOLOGY_FROM_RESOURCE_SERVER:
```
alanking: This will *only* run the test in topology-on-provider (i.e. not in core tests). Is that intentional? I was expecting this, based on the comment above:
```suggestion
            if not test.settings.TOPOLOGY_FROM_RESOURCE_SERVER:
```

alanking (commented):
univmss resource test commit message says:

> last three **digist** of an iRODS error code.

...which I assume should be:

> last three **digits** of an iRODS error code.

Also, I'm guessing that will be linked to #8968?

korydraughn:
> univmss resource test commit message says:
> 
> > last three **digist** of an iRODS error code.
> 
> ...which I assume should be:
> 
> > last three **digits** of an iRODS error code.
> 
> Also, I'm guessing that will be linked to #8968?

That's right.

scripts/irods/test/test_resource_types.py
korydraughn: The intention is for topology+consumer to be skipped.

korydraughn: Wait. I need to think about why that passed for core tests again.

korydraughn: Now I remember. The decision of which resource to create the replica on is left up to the server when no leaf resource is provided to itouch.

The comments hint at that being the case, but I'll work on making it clearer.

korydraughn:
Confirmed the univmss tests pass for core, topology+consumer, and topology+provider.

This is now ready. Will update commit messages on approval.

scripts/irods/test/test_resource_types.py
korydraughn: Done.

alanking: Wait, but... it still looks like it's *only* running this test case for topology-on-provider. How can this test case run for core tests when `test.settings.RUN_IN_TOPOLOGY` is `False` for that case?

korydraughn: Crap. Not sure how I overlooked that.

Will fix.

alanking: No problem - I thought maybe I was just missing something :)

korydraughn: Fixed.

scripts/irods/test/test_resource_types.py
```diff
@@ -2258,21 +2260,37 @@ def test_univmss_resource_plugin_supports_paths_containing_single_quotes__issue_
                 # Remove the data object for the next subtest.
                 self.user0.assert_icommand(['irm', '-f', filename])
 
-            with self.subTest('Sync-to-Archive - itouch'):
-                self.user0.assert_icommand(['itouch', filename])
+            # Skip this itouch case when running in a topology from a catalog consumer.
+            #
+            # itouch does not support targeting the root of resource hierarchy containing one or
```
korydraughn:
```suggestion
            # itouch does not support targeting the root of a resource hierarchy containing one or
```

korydraughn:
Will run the test I modified to make sure it's good.

korydraughn:
Confirmed the update to the `test.settings.TOPOLOGY_FROM_RESOURCE_SERVER` check is producing the expected output. All itouch cases run for the provider (core and topology). Not all itouch cases run for consumers.

Tests:
- `test_resource_types.Test_Resource_CompoundWithUnivmss.test_univmss_resource_plugin_supports_paths_containing_single_quotes__issue_8928`
- `test_resource_types.Test_Resource_CompoundWithUnivmss.test_create_univmss_resource_using_non_key_value_form_of_context_string__issue_8928`
- `test_resource_types.Test_Resource_CompoundWithUnivmss.test_create_univmss_resource_using_key_value_form_of_context_string__issue_8928`

Commit - fbeefe3fb8a418fd32bb12c05ac1b969a2bf7e3f.

scripts/irods/test/test_resource_types.py
korydraughn: Done.

korydraughn:
Awaiting final review and approval.

trel (approved):
seems good to me.

korydraughn:
Added pounds.

Merging.

PR #8950, opened 2026-04-29
[#7717,#8729] Add support for irods auth in setup_irods.py

alanking:
Addresses https://github.com/irods/irods/issues/7717
Addresses https://github.com/irods/irods/issues/8729
Possibly supersedes: https://github.com/irods/irods/pull/8939

### DIFF
--- a/scripts/irods/configuration.py
+++ b/scripts/irods/configuration.py
@@ -199,13 +199,56 @@ class IrodsConfig(object):
     @admin_password.setter
     def admin_password(self, value):
         l = logging.getLogger(__name__)
+        self._admin_password = value
         if not os.path.exists(os.path.dirname(paths.password_file_path())):
             os.makedirs(os.path.dirname(paths.password_file_path()), mode=0o700)
         mtime = int(time.time())
-        with open(paths.password_file_path(), 'wt') as f:
+        with open(paths.password_file_path(), 'wt', opener=lib.read_write_owner_opener) as f:
             l.debug('Writing password file %s', f.name)
             print(encode(value, mtime=mtime), end='', file=f)
         os.utime(paths.password_file_path(), (mtime, mtime))
+
+    def get_admin_password(self, check_legacy_password_file=False):
+        """
+        Get the admin password with the option of just using the cached value.
+
+        Args:
+            check_legacy_password_file: If True, get the password out of the .irodsA file. Else, read the cached value.
+
+        Returns:
+            The admin password.
+        """
+        if check_legacy_password_file:
+            return self.admin_password
+        return self._admin_password
+
+    def set_admin_password(self, value, create_legacy_password_file=False):
+        """
+        Set the admin password with the option of only updating the cached value.
+
+        Args:
+            create_legacy_password_file: If True, write the password to the .irodsA file. Else, only write cached value.
+        """
+        if create_legacy_password_file:
+            self.admin_password = value # use the property setter
+        else:
+            self._admin_password = value # just update cached value
+
+    @property
+    def admin_session_token(self):
+        return self._session_token
+
+    @admin_session_token.setter
+    def admin_session_token(self, value):
+        l = logging.getLogger(__name__)
+        self._session_token = value
+        if not os.path.exists(os.path.dirname(paths.service_account_session_token_file_path())):
+            os.makedirs(os.path.dirname(paths.service_account_session_token_file_path()), mode=0o600)
+        mtime = int(time.time())
+        with open(paths.service_account_session_token_file_path(), 'wt', opener=lib.read_write_owner_opener) as f:
+            l.debug('Writing session token to file: [%s]', f.name)
+            print(value, end='', file=f)
+        os.utime(paths.service_account_session_token_file_path(), (mtime, mtime))
 
     def commit(self, config_dict, path, clear_cache=True, make_backup=False):
         l = logging.getLogger(__name__)
@@ -274,6 +317,10 @@ class IrodsConfig(object):
     @property
     def password_file_path(self):
         return paths.password_file_path()
+
+    @property
+    def session_token_file_path(self):
+        return paths.service_account_session_token_file_path()
 
     @property
     def log_directory(self):
--- a/scripts/irods/database_connect.py
+++ b/scripts/irods/database_connect.py
@@ -1,14 +1,19 @@
 from __future__ import print_function
 
+import base64
+import hashlib
 import itertools
 import json
 import logging
 import os
 import pprint
+import secrets
 import shutil
+import string
 import sys
 import tempfile
 import time
+import uuid
 
 from . import lib
 from . import password_obfuscation
@@ -344,7 +349,7 @@ def create_database_tables(irods_config, cursor, default_resource_directory=None
             except IrodsError as e:
                 raise IrodsError('Database setup failed while running %s:\n%s' % (sql_file, str(e)))
 
-def setup_database_values(irods_config, cursor=None, default_resource_directory=None, default_resource_name=None):
+def setup_database_values(irods_config, cursor=None, default_resource_directory=None, default_resource_name=None, password_storage_mode="legacy"):
     l = logging.getLogger(__name__)
     timestamp = '{0:011d}'.format(int(time.time()))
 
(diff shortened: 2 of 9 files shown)


scripts/irods/configuration.py
```diff
@@ -199,14 +199,46 @@ def admin_password(self):
     @admin_password.setter
     def admin_password(self, value):
         l = logging.getLogger(__name__)
+        self._admin_password = value
         if not os.path.exists(os.path.dirname(paths.password_file_path())):
             os.makedirs(os.path.dirname(paths.password_file_path()), mode=0o700)
         mtime = int(time.time())
-        with open(paths.password_file_path(), 'wt') as f:
+        with open(paths.password_file_path(), 'wt', opener=lib.read_write_owner_opener) as f:
```
korydraughn, line 206: `lib.read_write_owner_opener` opens the file and sets the permissions such that only the owner has access. Group/World permissions are never set.

Is that correct?

scripts/irods/configuration.py
```diff
@@ -199,14 +199,46 @@ def admin_password(self):
...
             l.debug('Writing password file %s', f.name)
             print(encode(value, mtime=mtime), end='', file=f)
         os.utime(paths.password_file_path(), (mtime, mtime))
 
+    def get_admin_password(self, check_legacy_password_file=False):
+        """Get the admin password with the option of just using the cached value."""
```
korydraughn: Consider mentioning what happens if the parameter is set to true.

scripts/irods/configuration.py
```diff
@@ -199,14 +199,46 @@ def admin_password(self):
...
+        l = logging.getLogger(__name__)
+        if check_legacy_password_file:
+            return self.admin_password
+        return self._admin_password
+
+    def set_admin_password(self, value, create_legacy_password_file=False):
+        """Set the admin password with the option of only updating the cached value."""
```
korydraughn: Consider mentioning what happens if the parameter is set to true.

scripts/irods/database_connect.py
```diff
@@ -412,15 +417,74 @@ def get_next_object_id():
...
+        # Derive a key from the provided password.
+        salt = ''.join([secrets.choice(string.ascii_letters + string.digits) for _ in range(16)])
+        algorithm = "scrypt"
+        hashing_parameters = {
+            "N": 16384,
+            "r": 8,
+            "p": 1,
+            "key_length": 64
+        }
+        admin_password = irods_config.get_admin_password(check_legacy_password_file=(password_storage_mode != "hashed"))
```
korydraughn, line 431:
```suggestion
        admin_password = irods_config.get_admin_password(check_legacy_password_file=("hashed" != password_storage_mode))
```

scripts/irods/database_connect.py
```diff
@@ -412,15 +417,74 @@ def get_next_object_id():
...
+                admin_user_id,
+                hashed_session_token,
+                irods_config.server_config["zone_auth_scheme"],
+                '{0:011d}'.format(99999999999),
+                timestamp,
+                timestamp,
+                salt,
+                log_params=False)
+
+    # Write down the password storage mode based on what was requested as setup.
```
korydraughn: "as" -> "at" ?

scripts/irods/database_upgrade.py
```diff
@@ -287,8 +287,17 @@ def run_update(irods_config, cursor, is_upgrade):
...
+        password_storage_mode_query = (
+            "select option_value from R_GRID_CONFIGURATION "
+            "where "
+            "namespace = 'authentication' "
+            "and "
+            "option_name = 'password_storage_mode';"
+        )
+        if len(database_connect.execute_sql_statement(cursor, password_storage_mode_query).fetchall()) == 0:
+            password_storage_mode = "hashed" if "irods" == irods_config.server_config.get("zone_auth_scheme", "native") else "legacy"
+            database_connect.execute_sql_statement(cursor, f"insert into R_GRID_CONFIGURATION values ('authentication', 'password_storage_mode', '{password_storage_mode}');")
```
korydraughn, line 267: The idea here is, upgrading from pre-5.1.0 to 5.1.0 will always result in the password storage mode being set to legacy.

If doing development work and upgrading without bumping the version number or upgrading to 5.1.1+, this code block is always skipped because password storage mode is expected to be defined.

Correct?

If password storage mode is removed (in a 5.1+ deployment) and the admin performs an upgrade, the admin might end up needing to touch the database. The reason being that their deployment could have been operating with a password storage mode of `both`.

scripts/irods/tls.py
```diff
@@ -0,0 +1,68 @@
...
+
+from cryptography import x509
+from cryptography.hazmat.primitives import hashes, serialization
+from cryptography.hazmat.primitives.asymmetric import dh, rsa
+from cryptography.x509.oid import NameOID
+
+from . import lib
+
+
+def generate_tls_certificate_key(directory=None):
```
korydraughn, line 13: Let's document these functions. A line or two is good enough. Just a summary of what to expect.

scripts/setup_irods.py
```diff
@@ -113,35 +122,55 @@ def setup_server(irods_config, json_configuration_file=None, test_mode=False):
             os.makedirs(os.path.dirname(irods_config.client_environment_path), mode=0o700)
         irods_config.commit(json_configuration_dict['service_account_environment'], irods_config.client_environment_path)
         # password
-        irods_config.admin_password = json_configuration_dict['admin_password']
+        _password_storage_mode = json_configuration_dict.get("password_storage_mode", _password_storage_mode)
+        irods_config.set_admin_password(
+            json_configuration_dict['admin_password'],
+            create_legacy_password_file=(_password_storage_mode != "hashed")
+        )
```
korydraughn, line 130: This creates the password file (.irodsA), right?

Is this idempotent?

scripts/setup_irods.py
```diff
@@ -113,35 +122,55 @@ def setup_server(irods_config, json_configuration_file=None, test_mode=False):
...
         if irods_config.is_provider:
-            from irods import database_interface
             l.info(irods.lib.get_header('Configuring the database communications'))
-            database_interface.setup_database_config(irods_config)
+            irods.database_interface.setup_database_config(irods_config)
         # default resource
         default_resource_name, default_resource_directory = setup_storage(irods_config)
+        if prompt_auth:
+            setup_auth(irods_config)
+        if prompt_tls or irods_config.server_config["zone_auth_scheme"] != "native":
```
korydraughn, line 142:
```suggestion
        if prompt_tls or "native" != irods_config.server_config["zone_auth_scheme"]:
```

scripts/setup_irods.py
```diff
@@ -380,6 +408,150 @@ def setup_storage(irods_config):
 
     return (resource_name, resource_vault_path)
 
+
+def setup_auth(irods_config):
+    global _password_storage_mode
+
+    l = logging.getLogger(__name__)
+    l.info(irods.lib.get_header('Setting up auth'))
```
korydraughn: Let's use all of the letters for `auth`.

scripts/setup_irods.py
```diff
@@ -380,6 +408,150 @@ def setup_storage(irods_config):
...
+
+    while True:
+        # Prompt user for which password storage mode to use in this zone.
+        default_password_storage_mode = "both"
+        password_storage_modes = ["legacy", "hashed", "both"]
+        _password_storage_mode = irods.lib.default_prompt(
+            "Password storage mode",
+            default=password_storage_modes,
+            previous=password_storage_modes.index(default_password_storage_mode) + 1,
+            input_filter=irods.lib.set_filter(password_storage_modes, field='Password storage mode'))
```
korydraughn, line 455: How does this relate to the globally initialized value of `_password_storage_mode`?

I noticed that `_password_storage_mode` is initialized to `legacy` at the top of the file.

---

Wait, I think I see it now. The globally initialized value comes into play when the admin does not pass the flags to trigger the prompts, correct?

scripts/setup_irods.py
```diff
@@ -380,6 +408,150 @@ def setup_storage(irods_config):
...
+        else:
+            auth_scheme = "irods" if _password_storage_mode == "hashed" else "native"
+
+        irods_config.server_config['zone_auth_scheme'] = auth_scheme
+
+        confirmation_message = (
+            '\n'
+            '-------------------------------------------------------------\n'
+            f'Password storage mode: {_password_storage_mode}\n'
+            f'Authentication scheme: {irods_config.server_config["zone_auth_scheme"]}\n'
```
korydraughn: I'm wondering if these should have the `iRODS ` prefix like some of the other outputs (e.g. see iRODS port range, administrator, etc.).

Thoughts?

scripts/setup_irods.py
```diff
@@ -380,6 +408,150 @@ def setup_storage(irods_config):
...
+
+    while True:
+        if "tls_server" not in irods_config.server_config:
+            irods_config.server_config["tls_server"] = {}
+
+        if "tls_client" not in irods_config.server_config:
+            irods_config.server_config["tls_client"] = {}
+
+        generate_cert = irods.lib.default_prompt('Generate and use self-signed certificate now?', default=['yes'])
+        if generate_cert.lower() in ('y','yes'):
```
korydraughn:
```suggestion
        if generate_cert.lower() in ('y', 'yes'):
```

scripts/setup_irods.py
```diff
@@ -380,6 +408,150 @@ def setup_storage(irods_config):
...
+            f'Certificate key file:           {irods_config.server_config["tls_server"].get("certificate_key_file")}\n'
+            f'Diffie-Hellman parameters file: {irods_config.server_config["tls_server"].get("dh_params_file")}\n'
+            f'CA certificate file:            {irods_config.server_config["tls_client"].get("ca_certificate_file", "")}\n'
+            f'CA certificate path:            {irods_config.server_config["tls_client"].get("ca_certificate_path", "")}\n'
+            f'Certificate verification level: {irods_config.server_config["tls_client"].get("verify_server")}\n'
+            '-------------------------------------------------------------\n\n'
+            'Please confirm'
+        )
+
+        if irods.lib.default_prompt(confirmation_message, default=['yes']) in ['', 'y', 'Y', 'yes', 'YES']:
```
korydraughn: Does the `.lower()` trick not work here like in the auth prompts?

scripts/setup_irods.py
```diff
@@ -458,21 +630,29 @@ def setup_server_config(irods_config):
...
 
-    irods_config.admin_password = irods.lib.prompt(
+    # Note: This assignment creates an .irodsA file
+    admin_password = irods.lib.prompt(
         'iRODS administrator password',
         input_filter=irods.lib.character_count_filter(minimum=3, maximum=maximum_password_length, field='Admin password'),
         echo=False)
+    irods_config.set_admin_password(
+        admin_password,
+        create_legacy_password_file=(_password_storage_mode != "hashed")
```
korydraughn, line 676:
```suggestion
        create_legacy_password_file=("hashed" != _password_storage_mode)
```

korydraughn (commented):
Looks good to me.

scripts/irods/database_connect.py
alanking: I'm fine with Yoda syntax if everyone else is, but I'm starting to get trained to avoid it in Python: https://docs.astral.sh/ruff/rules/yoda-conditions/ Will address all of these instances if we agree that we prefer the Yoda syntax.

scripts/setup_irods.py
```diff
@@ -380,6 +408,150 @@ def setup_storage(irods_config):
...
+    l.info(irods.lib.get_header('Setting up TLS'))
+
+    while True:
+        if "tls_server" not in irods_config.server_config:
+            irods_config.server_config["tls_server"] = {}
+
+        if "tls_client" not in irods_config.server_config:
+            irods_config.server_config["tls_client"] = {}
+
+        generate_cert = irods.lib.default_prompt('Generate and use self-signed certificate now?', default=['yes'])
```
trel: possible phrasing.   the 'now' feels ... weird?

```suggestion
        generate_cert = irods.lib.default_prompt('Generate and use a self-signed certificate?', default=['yes'])
```

scripts/irods/configuration.py
alanking: Correct

scripts/irods/database_upgrade.py
alanking: > The idea here is, upgrading from pre-5.1.0 to 5.1.0 will always result in the password storage mode being set to legacy.
> 
> If doing development work and upgrading without bumping the version number or upgrading to 5.1.1+, this code block is always skipped because password storage mode is expected to be defined.
> 
> Correct?

Correct. We have to avoid this because the password storage mode is set for fresh installations and since this script runs for new installs as well, we must ensure that the value used at setup doesn't get overwritten here.

> If password storage mode is removed (in a 5.1+ deployment) and the admin performs an upgrade, the admin might end up needing to touch the database. The reason being that their deployment could have been operating with a password storage mode of `both`.

Correct. I would argue that this is an unsupported mode of operation, too.

scripts/setup_irods.py
alanking: Right, the value of `_password_storage_mode` only really changes here.

scripts/setup_irods.py
alanking: I think I just lifted this from somewhere else. Can change it

scripts/setup_irods.py
alanking: I guess it is? It's just using the `admin_password` property that already exists if `create_legacy_password_file` is `True`:
https://github.com/irods/irods/blob/27b72d946f6d27c20ea316658d9456f4853bda39/scripts/irods/configuration.py#L199-L208

scripts/irods/database_connect.py
korydraughn: I didn't know that. That is a perfect reason for not doing the yoda syntax changes in python.

Feel free to ignore them if you like.

scripts/irods/database_upgrade.py
korydraughn: I agree. As long as admins do not touch the database directly (in weird ways), they should never encounter that case.

Resolving.

scripts/setup_irods.py
korydraughn: Yeah, let's do that.

alanking:
After some difficulties with this, I'm going to put it in draft for now. The TLS part is going to be broken out into a separate PR and we may rebase this one with just the auth prompts if it seems good.

alanking:
TLS work to be included in 5.1.0 and has been merged here: https://github.com/irods/irods/pull/8966

This PR represents the `irods` auth piece of the work, and will require a rebase and, of course, some conflict resolution.

alanking:
Rebased and resolved conflicts. Catalog consumers still have a "cold start" problem with irods auth at setup time, which is why this is in draft.

scripts/irods/database_connect.py
korydraughn: Ignore yoda syntax for this.

Resolved.

Issue #8956, opened 2026-05-01
`iqstat` does not check `RcComm` for `nullptr` before use


korydraughn:
## Bug Report

https://github.com/irods/irods/blob/8b80edba937ade8a7ad5f6467844199cebbfd3bf/clients/icommands/src/iqstat.cpp#L175-L178

Issue #8638, opened 2025-08-27
Add message on backwards incompatibility icommands 5.0 to iRODS 4.2.x server


tsmeele:
### New Feature / Functionality
The iRODS 5.0.1 icommands are not compatible with an iRODS 4.2.x server.  As the server reports its version, the icommands could use such information to proactively notify the user of the incompatibility with a more gentle message.  

Examples of current output:

iinit (user has populated irods_environment.json with institute-suggested values (pam)):
```
$ iinit
Connecting as XXX to XXX:1247 ...
Level 0: Error occurred while authenticating user [XXX] [PLUGIN_ERROR_MISSING_SHARED_OBJECT: [-]	/usr/include/irods/irods_load_plugin.hpp:448:error irods::load_plugin(PluginType *&, const std::string &, const std::string &, const std::string &, const Ts &...) [PluginType = irods::authentication::authentication_base, Ts = <char[14]>] :  status [PLUGIN_ERROR_MISSING_SHARED_OBJECT]  errno [] -- message [shared library does not exist [/usr/lib/irods/plugins/authentication/libirods_auth_plugin-pam_client.so]]
$
```
imiscsvrinfo:
```
$ imiscsvrinfo
Segmentation fault (core dumped)
$
```

alanking:
Agreed that the messages should be made better. The segmentation fault in `imiscsvrinfo` is a bug and other iCommands may exhibit similar behavior, so we will treat this as such.

korydraughn:
I've opened #8653 to address the `imiscsvrinfo` segfault.

---

Related to #8605.

If someone attempts to use iRODS 5 icommands against an older version of iRODS, should the icommands stop immediately or print a message and continue? I think I'm in the camp of stopping immediately.

And what about iRODS 5 icommands talking to iRODS 6?

We know how to handle changes in packing instructions (see https://github.com/irods/irods_client_http_api/pull/432), but doing that will result in an increase in complexity.

trel:
clients talking to an older server... most things will work... but that's not very deterministic.

stopping immediately seems abrupt, but definitely clearer.

hmm, not sure...

tsmeele:
The most important is that the user is made _aware_ that unexpected behavior may have arisen from incompatible releases. Note that the average user is likely to have installed the latest software release which, in the past, usually worked.

Therefore it would suffice that a warning message is shown at the start, whenever the client release is newer than the server release.
  
[...]

PR #8957, opened 2026-05-01
[#8638,#8956] Warn user about incompatible icommands / Check for nullptr (main)

korydraughn:
Here's a real example of what users will see with this initial implementation.

```
$ ils
Warning: Major version number mismatch detected between iCommands and server!

   iCommands major version number   : 5
   iRODS server major version number: 4

To avoid compatibility issues, use an iCommands installation which matches the server's major version number.
/tempZone/home/kory:
```

TODOs:
- [x] Confirm all icommands report the warning when connected to an iRODS 4 server
- [x] Confirm environment variable disables warning for one iCommand
- [x] Run unit tests
- [x] Run core tests

### DIFF
--- a/clients/icommands/src/irule.cpp
+++ b/clients/icommands/src/irule.cpp
@@ -180,6 +180,10 @@ main( int argc, char **argv ) {
 
     load_client_api_plugins();
 
+    // Keeps warn_if_connected_to_potentially_incompatible_server from being invoked
+    // more than once.
+    bool print_incompatibility_warning = true;
+
     /* Don't need to parse parameters if just listing available rule_engine_instances */
     if ( argsMap.count( "available" ) ) {
         /* add key val for listing available rule engine instances */
@@ -206,7 +210,7 @@ main( int argc, char **argv ) {
                 exit( 10 );
             }
 
-            /* if the input file name starts with "i:", the get the file from iRODS server */
+            // If the input file name starts with "i:", get the file from the iRODS server.
             if ( !strncmp( fileName, "i:", 2 ) ) {
                 status = getRodsEnv( &myEnv );
 
@@ -221,6 +225,9 @@ main( int argc, char **argv ) {
                 if ( conn == NULL ) {
                     exit( 2 );
                 }
+
+                utils::warn_if_connected_to_potentially_incompatible_server(*conn);
+                print_incompatibility_warning = false;
 
                 status = utils::authenticate_client(conn, myEnv);
                 if ( status != 0 ) {
@@ -450,6 +457,12 @@ main( int argc, char **argv ) {
             rodsLogError( LOG_ERROR, errMsg.status, "rcConnect failure %s",
                           errMsg.msg );
             exit( 2 );
+        }
+
+        if (print_incompatibility_warning) {
+            // NOLINTNEXTLINE(clang-analyzer-deadcode.DeadStores)
+            print_incompatibility_warning = false;
+            utils::warn_if_connected_to_potentially_incompatible_server(*conn);
         }
 
         status = utils::authenticate_client(conn, myEnv);
--- a/clients/icommands/src/utility.hpp
+++ b/clients/icommands/src/utility.hpp
@@ -22,7 +22,7 @@ namespace utils
         // Setting this environment variable is required so that "ips" can display
         // the command name alongside the connection information.
         if (setenv(SP_OPTION, _display_name.data(), /* overwrite */ 1)) {
-            std::cout << "Warning: Could not set environment variable [spOption] for ips.";
+            std::cout << "Warning: Could not set environment variable [spOption] for ips.\n";
         }
     }
 
--- a/clients/icommands/src/utility.hpp
+++ b/clients/icommands/src/utility.hpp
@@ -33,16 +33,19 @@ namespace utils
         return ia::authenticate_client(*_comm, ctx);
     } // authenticate_client
 
-    inline auto option_specified(std::string_view _option, int argc, char** argv) -> bool
+    inline auto option_specified(std::string_view _option, int _argc, char** _argv) -> bool
     {
-        for (int arg = 0; arg < argc; ++arg) {
-            if (!argv[arg]) {
+        for (int arg = 0; arg < _argc; ++arg) {
+            // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic)
+            if (!_argv[arg]) {
                 continue;
             }
 
-            if (_option == argv[arg]) {
+            // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic)
+            if (_option == _argv[arg]) {
                 // parseCmdLineOpt is EVIL and requires this. Please don't ask why.
-                argv[arg] = "-Z";
+                // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic)
+                _argv[arg] = "-Z";
                 return true;
             }
         }
@@ -88,4 +91,3 @@ namespace utils
 } // namespace utils
 
 #endif // IRODS_ICOMMANDS_UTILITY_HPP
-
--- a/clients/icommands/src/utility.hpp
+++ b/clients/icommands/src/utility.hpp
@@ -3,13 +3,17 @@
 
 #include <irods/authentication_plugin_framework.hpp>
 #include <irods/getRodsEnv.h>
+#include <irods/irods_version.h>
 #include <irods/rodsDef.h>
+#include <irods/version.hpp>
+
+#include <fmt/format.h>
+#include <nlohmann/json.hpp>
 
 #include <cstdlib>
+#include <cstring>
 #include <iostream>
 #include <string_view>
-
-#include <nlohmann/json.hpp>
 
 namespace utils
 {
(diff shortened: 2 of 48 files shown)


clients/icommands/src/utility.hpp
```diff
@@ -29,23 +32,40 @@ namespace utils
...
+            fmt::print(stderr,
+                       "Warning: Major version number mismatch detected between iCommands and server!\n\n"
+                       "   iCommands major version number   : {}\n"
+                       "   iRODS server major version number: {}\n\n"
+                       "To avoid compatibility issues, use an iCommands installation which matches the server's major "
+                       "version number.\n",
+                       IRODS_VERSION_MAJOR,
+                       server_version->major);
+        }
+    } // warn_if_connected_to_potentially_incompatible_server
```
korydraughn, line 90: This is the main change of the PR.

korydraughn: Perhaps this should be a `!=` kind of check?

Then again, our policy is to never change packing instructions or signatures of API endpoints. So, this is likely fine.

Hmm.

clients/icommands/src/utility.hpp
```diff
@@ -29,23 +32,40 @@ namespace utils
...
     } // option_specified
+
+    inline auto warn_if_connected_to_potentially_incompatible_server(const RcComm& _comm) -> void
+    {
+        const auto server_version = irods::to_version(_comm.svrVersion->relVersion);
+        if (server_version && server_version->major < IRODS_VERSION_MAJOR) {
+            fmt::print(stderr,
+                       "Warning: Major version number mismatch detected between iCommands and server!\n\n"
+                       "   iCommands major version number   : {}\n"
+                       "   iRODS server major version number: {}\n\n"
```
trel:
```suggestion
                       "   iCommands   : {}\n"
                       "   iRODS Server: {}\n\n"
```

clients/icommands/src/utility.hpp
```diff
@@ -29,23 +32,40 @@ namespace utils
...
+                       "To avoid compatibility issues, use an iCommands installation which matches the server's major "
+                       "version number.\n",
```
trel: maybe this isn't needed?

just the warning itself is enough to make people upgrade or downgrade their iCommands?

korydraughn: That's fair. I struggled a great deal trying to figure out how to phrase that lol. I'm certainly not against removing it.

clients/icommands/src/utility.hpp
korydraughn: Done.

Kept "server" lowercase to match other messages from server.

clients/icommands/src/utility.hpp
korydraughn: Done.

clients/icommands/src/utility.hpp
trel: i think `!=` is stronger.   it will let users know they are behind, too.

can't think of a downside... it's a warning.

korydraughn: Done.

See warning message for case where the server's version is unavailable.

clients/icommands/src/irule.cpp
```diff
@@ -206,7 +210,7 @@ main( int argc, char **argv ) {
                 exit( 10 );
             }
 
-            /* if the input file name starts with "i:", the get the file from iRODS server */
+            /* if the input file name starts with "i:", then get the file from iRODS server */
```
trel: can we change this to match our current comment style?

clients/icommands/src/utility.hpp
```diff
@@ -29,23 +32,49 @@ namespace utils
...
+    //
+    // Prints a warning to stderr if the version of the server cannot be determined.
+    // It is recommended that the user verify they are connecting to a trusted server
+    // before proceeding.
+    inline auto warn_if_connected_to_potentially_incompatible_server(const RcComm& _comm) -> void
+    {
+        const auto server_version = irods::to_version(_comm.svrVersion->relVersion);
+        if (!server_version) {
+            fmt::print(stderr,
+                       "Warning: Could not determine version of server. Make sure you are connected to a trusted iRODS "
```
trel:
```suggestion
                       "Warning: Could not determine server version. Make sure you are connected to a trusted iRODS "
```

clients/icommands/src/utility.hpp
```diff
@@ -29,23 +32,49 @@ namespace utils
...
+                       "server.\n\n");
+        }
+        else if (server_version->major != IRODS_VERSION_MAJOR) {
+            fmt::print(stderr,
+                       "Warning: Major version number mismatch detected between iCommands and server!\n\n"
```
trel: exclamation point feels too much?

```suggestion
                       "Warning: Major version number mismatch detected between iCommands and server.\n\n"
```

clients/icommands/src/utility.hpp
```diff
@@ -29,23 +32,49 @@ namespace utils
...
+                       "   iCommands   : {}\n"
+                       "   iRODS server: {}\n\n",
+                       IRODS_VERSION_MAJOR,
+                       server_version->major);
```
alanking, line 88: Can we provide a way to suppress this message, like an environment variable?

korydraughn: We can. It's not clear to me if we should offer one though.

With the current impl, users can redirect stderr to keep things quiet. I wouldn't say that's a real solution though.

I assume the suppression idea is for people who have scripts which rely on the output of the icommands?

trel: would this currently be output for all icommands, everytime one is used?

alanking: It's moreso for deployments where it is intentional and things are working for their use cases. Or, perhaps a user just wants to install iCommands to do something and gets the iRODS 5 iCommands even though their organization's iRODS server(s) are running iRODS 4.3. We have similar provisions in the PRC to suppress warning messages which seem helpful. It's not a big deal - just a "being nice to the user" thought.

trel: agreed.   people should be allowed to make their tools quiet.

korydraughn: Okay. An environment variable is easy to do. Now for the fun part.

What shall we name it?
Is it generic or targeted?

korydraughn: A targeted name feels appropriate to me. Something like the following _(not ordered)_:
- `IRODS_SILENCE_COMPATIBILITY_WARNINGS`
- `IRODS_IGNORE_ICOMMANDS_SERVER_COMPAT_WARNING`
- `IRODS_IGNORE_ICOMMANDS_SERVER_VERSION_MISMATCH_WARNING`
- `IRODS_SKIP_COMPATIBILITY_CHECK`

Any other suggestions?

trel: feels like a warning, rather than a compatibility check...
and  i like these verbs ... `QUIET`, `SILENCE`, `SUPPRESS`

other clients could honor this too... so it doesn't have to call out icommands i think...

singular or plural... i think i prefer singular...

- `IRODS_QUIET_COMPATIBILITY_WARNING`
- `IRODS_QUIET_COMPATIBILITY_WARNINGS`

trel: mmm, `IGNORE` also works.

korydraughn: I think it should be specific to icommands. Keeps the blast radius narrow and easy to control.

korydraughn: More names ...
- `ICOMMANDS_ENABLE_SERVER_COMPATIBILITY_CHECK`
- `ICOMMANDS_DETECT_SERVER_INCOMPATIBILITY`
- `ICOMMANDS_WARN_ON_SERVER_INCOMPATIBILITY`

trel: oh, flipping it to a positive default...

don't hate `ICOMMANDS_DETECT_SERVER_INCOMPATIBILITY`
or `ICOMMANDS_DETECT_SERVER_MISMATCH`

korydraughn: Might need to include VERSION.
- `ICOMMANDS_DETECT_SERVER_VERSION_INCOMPATIBILITY`
- `ICOMMANDS_DETECT_SERVER_VERSION_MISMATCH`
- `ICOMMANDS_DETECT_VERSION_MISMATCH`
- `ICOMMANDS_DETECT_VERSION_MISMATCH_WITH_SERVER`

trel: favorite...

> `ICOMMANDS_DETECT_SERVER_VERSION_MISMATCH`

korydraughn: Going with that one.

clients/icommands/src/irule.cpp
korydraughn: Done.

clients/icommands/src/utility.hpp
korydraughn: Done.

clients/icommands/src/utility.hpp
korydraughn: Done.

clients/icommands/src/utility.hpp
korydraughn: Done.

trel:
looks good - okay to squash from my perspective.

alanking (commented):
Looks good. Squash

korydraughn:
I'll run the tests and verify everything works following the squash.

korydraughn:
Rebased on tip of main.

All unit tests passed.
All core tests passed except `test_logical_quotas.py`.

@FifthPotato I ran the core tests against Ubuntu 22.04 + MariaDB 11.4. Did you see the logical quotas test suite pass against MariaDB? Are you able to reproduce this?

```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuite name="irods.test.test_logical_quotas.Test_Logical_Quotas-20260507025411" tests="16" file="irods/test/test_logical_quotas.py" time="134.510" timestamp="2026-05-07T02:56:25" failures="11" errors="0" skipped="0">
```

Most failures report the following.

```
Assert Command: iadmin modrepl logical_path /tempZone/home/lquotauser/2026-05-07Z02:55:23--irods-testing-xwk40utn/test_logical_quota_replica_4 replica_number 0 DATA_REPL_STATUS 1
Expecting EMPTY: ['']
  stdout:
    | Level 0: db_mod_data_obj_meta_op - Failed to find file in database by its logical path.
    | Level 0: db_mod_data_obj_meta_op - Failed to find file in database by its logical path.
  stderr:
    | rcModDataObjMeta failed when modifying replica: [CAT_UNKNOWN_FILE (-817000)]
Unexpected output on stderr

FAILED TESTING ASSERTION
```

Will open an issue for that. I don't think the changes in this PR are the cause of those failures.

korydraughn:
Confirmed all icommands report the warning when connected to 4.3.5.

Confirmed `ils` does not show the warning when the environment variable `ICOMMANDS_DETECT_SERVER_VERSION_MISMATCH` is set to `0`.

Testing complete. This is now ready.

korydraughn:
I've opened #8970 to address the logical quotas tests. @FifthPotato is working on it.

trel (approved):
looks good.   i think it's ready to squash the one commit.

korydraughn:
Squashed that one commit.

korydraughn:
Adding pounds and merging.

korydraughn:
Added pounds.

Merging.

Issue #8545, opened 2025-05-29
need /etc/tmpfiles.d/irods.conf file to ensure /var/run/irods exists with ownership


trel:
Discovered after two installs where `/var/run` is a tmpfs... and so `/var/run/irods` does not persist across a reboot...

We need to create `/etc/tmpfiles.d/irods.conf` with the following content...

```
# Create the irods location to hold PID files
#
# Type Path Mode User Group Age Argument
d     /var/run/irods   0755 irods irods
```

SwooshyCueb:
This must be generated by setup, as the user and group for the service account do not exist prior, and also may not even be `irods:irods`

korydraughn:
Some systems will report messages like the following due to the use of the legacy path prefix `/var/run`.

```
/etc/tmpfiles.d/irods.conf:1: Line references path below legacy directory /var/run/, updating /var/run/irods → /run/irods; please update the tmpfiles.d/ drop-in file accordingly
```

For such cases, the path in `/etc/tmpfiles.d/irods.conf` should be changed to `/run/irods.conf`.

PR #8962, opened 2026-05-04
[#8545] Update post-setup notes for PID file directory management (main)

korydraughn:
Companion PR: irods/irods_docs#423

Verified the changes result in a file being put into /etc/tmpfiles.d if the directory exists. Tested on:
- Ubuntu 22.04 - no /etc/tmpfiles.d, so irods.conf is not created
- Ubuntu 24.04 - creates file
- Rocky Linux 9 - creates file

### DIFF
--- a/scripts/setup_irods.py
+++ b/scripts/setup_irods.py
@@ -4,6 +4,7 @@ import argparse
 import os
 import pathlib
 import sys
+import textwrap
 
 import irods.setup_options, irods.tls
 
@@ -187,21 +188,49 @@ def setup_server(irods_config, json_configuration_file=None, test_mode=False, op
     if not args.skip_post_install_test:
         test_put(irods_config)
 
+    l.info(irods.lib.get_header('Stopping iRODS...'))
+    controller.stop()
+
+    l.info(irods.lib.get_header('iRODS is configured and ready to be started'))
+
     # extract the "irods_version" property from the version.json.dist file.
     # this guarantees that setup always uses the correct version information.
     with open('.'.join([irods_config.version_path, 'dist'])) as f:
         irods_version_string = json.load(f)['irods_version']
 
-    l.info(irods.lib.get_header('Log Configuration Notes'))
-    l.info(('iRODS uses syslog for logging. If your OS has a running syslog service, messages\n'
-            'will appear in the system log file (normally located at /var/log/syslog).\n\n'
-            'See the following for more information about configuring an iRODS-specific log file:\n\n'
-            f'  https://docs.irods.org/{irods_version_string}/system_overview/troubleshooting/'))
+    l.info(irods.lib.get_header('Post Setup Notes'))
+    l.info(textwrap.dedent(
+        f'''\
+        Logging
+        -------
+        iRODS uses syslog for logging. If your OS has a running syslog service, messages
+        will appear in the system log file (normally located at /var/log/syslog).
 
-    l.info(irods.lib.get_header('Stopping iRODS...'))
-    controller.stop()
+        See the following for information about configuring an iRODS-specific log file:
 
-    l.info(irods.lib.get_header('iRODS is configured and ready to be started'))
+            https://docs.irods.org/{irods_version_string}/system_overview/server_log/
+
+        Service Manager
+        ---------------
+        iRODS supports integration with service managers such as systemd. This
+        functionality is not enabled by default.
+
+        See the following for information about managing iRODS via a service manager:
+
+            https://docs.irods.org/{irods_version_string}/getting_started/running/#managing-irods-via-the-service-manager
+
+        PID File Management
+        -------------------
+        On an OS where /run (or /var/run) is mounted as tmpfs, the iRODS PID file
+        directory may not persist across reboots. This can prevent iRODS from starting
+        successfully on installations using an install prefix of "/".
+
+        See the following for information about PID file management:
+
+            https://docs.irods.org/{irods_version_string}/getting_started/running/#server-pid-file
+        '''
+    ))
+
 
 def replace_in_file(filepath, original, replacement, flags=0):
     with open(filepath, "r+") as f:
--- a/scripts/setup_irods.py
+++ b/scripts/setup_irods.py
@@ -658,13 +658,11 @@ def main():
     optional_prompts = []
 
     irods.log.register_file_handler(irods_config.setup_log_path)
+    llevel = logging.INFO
     if None != args.verbose and args.verbose > 0:
-        llevel = logging.NOTSET
         if args.verbose == 1:
-            llevel = logging.INFO
-        elif args.verbose == 2:
             llevel = logging.DEBUG
-        irods.log.register_tty_handler(sys.stdout, llevel, logging.WARNING)
+    irods.log.register_tty_handler(sys.stdout, llevel, logging.WARNING)
 
     if args.server_log_level != None:
         irods_config.injected_environment['spLogLevel'] = str(args.server_log_level)


scripts/setup_irods.py
```diff
@@ -364,6 +365,36 @@ def setup_service_account(irods_config, irods_user, irods_group):
     #owner of top-level directory changed, clear the cache
     irods_config.clear_cache()
 
+def configure_tmpfiles_for_pid_file_directory(irods_user, irods_group):
+    l = logging.getLogger(__name__)
+    l.info(irods.lib.get_header('Adding configuration to tmpfiles.d for PID file directory'))
+
+    tmpfiles_dir = '/etc/tmpfiles.d'
+    if not os.path.exists(tmpfiles_dir):
```
alanking: If you care to look into it... It seems that `pathlib` is the new, hip way of dealing with paths: https://docs.astral.sh/ruff/rules/os-path-join/ Feel free to ignore

alanking (commented):
Seems fine to me. Might be worth eyeballs from @SwooshyCueb and/or @trel

scripts/setup_irods.py
```diff
@@ -364,6 +365,36 @@ def setup_service_account(irods_config, irods_user, irods_group):
...
+    run_dir = '/run'
+    if not os.path.exists(run_dir):
+        run_dir = '/var/run'
+        if not os.path.exists(run_dir):
+            return
+
+    with open(tmpfiles_conf, 'wt') as f:
+        f.write('# Create the iRODS location to hold PID files.\n')
+        f.write('#\n')
+        f.write('# Type Path Mode User Group Age Argument\n')
```
trel: Argument?

```suggestion
        f.write('# Type Path Mode User Group Age\n')
```

korydraughn:
Ran the catch2 unit tests for ub22, ub24, and rocky9. They passed.

Build: c68db0654a3b32644012d81f1913452bbaaa5c6a

scripts/setup_irods.py
korydraughn: Updated to use pathlib.

scripts/setup_irods.py
korydraughn: Removed.

korydraughn:
Going to re-run the catch2 tests for those three OS and then manually verify that setup is doing the right thing.

korydraughn:
Unit tests passed for the mentioned OS.

Checking installations now.

scripts/setup_irods.py
```diff
@@ -364,6 +366,38 @@ def setup_service_account(irods_config, irods_user, irods_group):
     #owner of top-level directory changed, clear the cache
     irods_config.clear_cache()
 
+def configure_tmpfiles_for_pid_file_directory(irods_user, irods_group):
+    l = logging.getLogger(__name__)
+    l.info(irods.lib.get_header('Adding configuration to tmpfiles.d for PID file directory'))
+
+    tmpfiles_dir = Path('/etc/tmpfiles.d')
```
korydraughn: This does not consider non-package installs. Same for `/run` and `/var/run`.

Need to think through how these directories should be handled for non-package install.

korydraughn:
Confirmed `/etc/tmpfiles.d/irods.conf` is installed into the correct location for package installs.

SwooshyCueb:
I'd rather see this in our packaging than in a setup script. Moreover, I'm not sure we need a tmpfiles config for `/run/irods` if we're already putting it in the systemd unit file via `RuntimeDirectory`

korydraughn:
> I'd rather see this in our packaging than in a setup script.

I agree. Until we move away from setup allowing admins to create service account users, setup has to handle this.

> Moreover, I'm not sure we need a tmpfiles config for `/run/irods` if we're already putting it in the systemd unit file via `RuntimeDirectory`

That assumes everyone is running iRODS via systemd, which we have no way of confirming.

That said, until I confirm the non-package install case is covered, this PR is effectively blocked.

scripts/setup_irods.py
```diff
@@ -364,6 +366,38 @@ def setup_service_account(irods_config, irods_user, irods_group):
...
+    if not tmpfiles_dir.exists():
+        l.info(f'[{tmpfiles_dir}] does not exist. Skipping.')
+        return
```
korydraughn: This isn't necessary for non-package installs since the server creates the PID file under `<install_prefix>/var/run`. The function should return if performing a non-package install.

Need to figure out how to detect a non-package install during setup. These things come to mind:
- Check the install prefix used by paths.py
- Expose an option on setup_irods.py which informs setup when a non-package install is being configured

There might be other/better ways to do it.

scripts/setup_irods.py
korydraughn: See https://github.com/irods/irods/pull/8962/changes#r3204280323.

scripts/setup_irods.py
```diff
@@ -364,6 +366,38 @@ def setup_service_account(irods_config, irods_user, irods_group):
...
+
+    # Do not proceed if both /run and /var/run are missing.
+    run_dir = Path('/run')
+    if not run_dir.is_dir():
+        run_dir = Path('/var/run')
+        # is_dir() follows symlinks by default. This is important because
+        # /var/run tends to be a symlink to /run on modern systems.
+        if not run_dir.is_dir():
+            return
```
korydraughn: Use `irods.paths.runstate_directory()`.

scripts/setup_irods.py
korydraughn: Checking the install prefix is unlikely to be the answer because that only changes if the user specifies something other than the default. If they do not specify a prefix, the files will land in the same locations as a packaged install.

We could query the system, but that means dealing with platform differences.

I'm starting to think a flag on setup_irods.py is the answer.

scripts/setup_irods.py
korydraughn: Done.

scripts/setup_irods.py
korydraughn: Checking SYSCONFDIR is enough to handle this. No need for a new flag. See the code comment for more info.

Done.

korydraughn:
Confirmed that setup handles tmpfiles configuration correctly for the following:
- Ubuntu 22.04 - prompt-based install, unattended install
- Ubuntu 24.04 - prompt-based install, unattended install
- Rocky Linux 9 - prompt-based install, unattended install

The handling of tmpfiles is skipped on systems which do not have an /etc/tmpfiles.d.

All unit and core tests passed. Tested against Ub24.

MartinFlores751 (approved):
Changes look good.

korydraughn:
I'm starting to think this should be handled by documentation.

We'd provide a template file with placeholders and include wording in the post-setup message about systemd and tmpfiles.d.

That aligns with how we handled syslog and logrotate.

korydraughn:
If we accept this PR and a sysadmin wants to use systemd or some other service manager, they might need to remove the tmpfiles.d config to avoid management issues with the PID file directory.

Documenting that is equally acceptable since iRODS doesn't install service manager files.

korydraughn:
Another alternative is for the sysadmin to tell iRODS to write its PID file to a location other than /run/irods. For example:

```sh
# Assume we're in the service account home directory.
irods $ mkdir pid_dir
irods $ irodsServer -p $(pwd)/pid_dir/irods-server.pid -d
```

No tmpfiles.d config file needed.

korydraughn:
After some discussion with @trel, we've agreed to handle this as documentation and maybe provide a template file for the tmpfiles.d case. That means updating the post-setup message and docs at docs.irods.org.

No changes to the setup script will be made.

korydraughn:
Removed the logic which generates a template file. There's no point in producing a template file since it is trivial to write yourself. Instead, a link to an example config file at docs.irods.org is referenced instead _(section in docs not updated yet)_. This aligns with how we handle syslog and logrotate.

Here's what the post-setup notes look like now.

```
+------------------+
| Post Setup Notes |
+------------------+

Logging
-------
iRODS uses syslog for logging. If your OS has a running syslog service, messages
will appear in the system log file (normally located at /var/log/syslog).

See the following for information about configuring an iRODS-specific log file:

    https://docs.irods.org/5.0.90/system_overview/server_log/

Service Manager
---------------
iRODS supports integration with service managers such as systemd. This
functionality is not enabled by default.

See the following for information about managing iRODS via a service manager:

    https://docs.irods.org/5.0.90/getting_started/running/#managing-irods-via-the-service-manager

PID File Management
-------------------
On OS where /run (or /var/run) is mounted as tmpfs, the iRODS PID file
directory may not persist across reboots. This can prevent iRODS from starting
successfully on installations using an install prefix of "/".

See the following for information about PID file management:

    https://docs.irods.org/5.0.90/getting_started/running/#server-pid-file

```

scripts/setup_irods.py
```diff
@@ -189,21 +188,47 @@ def setup_server(irods_config, json_configuration_file=None, test_mode=False, op
...
+iRODS supports integration with service managers such as systemd. This
+functionality is not enabled by default.
+
+See the following for information about managing iRODS via a service manager:
+
+    https://docs.irods.org/{irods_version_string}/getting_started/running/#managing-irods-via-the-service-manager
+
+PID File Management
+-------------------
+On OS where /run (or /var/run) is mounted as tmpfs, the iRODS PID file
```
trel: 'On OS where' ... feels clunky.   "On an OS where" ?

korydraughn: Will update.

korydraughn: Done.

korydraughn:
Apparently, the post-setup messages are not being shown to the admin during setup. They are written to the log file for setup. I doubt admins are looking at that file unless something failed.

In order to see the messages during setup, you have to pass `-v` to the setup script.

That means a lot of admins are missing important information post-setup.

I've verified this is true for 4.3.5 too, which implies all of 4.3 and possibly 4.2 share this behavior.

@trel Is that surprising to you?

trel:
Hmm, yes, that is surprising to me.   I'm pretty sure I see printouts when we've done training...

korydraughn:
I've confirmed the setup script printed the messages until the release of 4.2.9.

The change in behavior appears to stem from https://github.com/irods/irods/commit/ec41d9e650b9065e21795db2762fd8c63a9be379.

Setting the log level to `logging.INFO` results in the messages appearing. Confirmed by modifying the setup script for 4.2.9.

trel:
ah, very interesting...  5y?!

korydraughn:
Ran unit tests against ub22, ub24, and rocky9. All tests passed.

Confirmed INFO-level log messages now appear without needing to pass `-v` to setup_irods.py. Also confirmed DEBUG-level messages are enabled when `-v` is passed.

Squashed. Checking commit messages one last time before adding pounds and merging.

Issue #8190, opened 2025-02-11
Investigate use of `irule` in `test_delay_queue_with_long_job(...)`


MartinFlores751:
In `test_delay_queue.py:202`, `irule` is used without specifying the rule engine instance.

This should be updated to produce appropriate errors if issues occur in `irule`.

Check for other uses of `irule` in the function, and possibly the file.

korydraughn:
The line referenced in the issue is:

https://github.com/irods/irods/blob/aa83f741bacd8476583fb3d7a76aa06de92a26ec/scripts/irods/test/test_delay_queue.py#L202

korydraughn:
Confirmed there are many other cases throughout test_delay_queue.py which do not specify a REP instance.

PR #8983, opened 2026-05-15
[#8190] Pass REP instance to irule in delay queue tests (main)

korydraughn:
test_delay_queue.py passes with these changes.

I've looked through each test to make sure their original intent is maintained.

### DIFF
--- a/scripts/irods/test/test_delay_queue.py
+++ b/scripts/irods/test/test_delay_queue.py
@@ -1,14 +1,8 @@
-from __future__ import print_function
-import sys
-if sys.version_info < (2, 7):
-    import unittest2 as unittest
-else:
-    import unittest
-
 import json
 import os
 import textwrap
 import time
+import unittest
 
 from . import resource_suite
 from . import session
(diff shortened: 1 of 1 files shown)


scripts/irods/test/test_delay_queue.py
FifthPotato, line 16: Instead of declaring a variable with the plugin instance every time, could you use the value of `IrodsConfig().default_rule_engine_instance` everywhere, similar to how this was declared/used? It could avoid the re-declaration of a local `rep_instance` each time.

FifthPotato (commented):
Looks pretty good! Just one nitpick and not much else.

scripts/irods/test/test_delay_queue.py
korydraughn: Will update.

korydraughn: Done.

korydraughn:
Forgot that @FifthPotato isn't available at this time.

Confirmed the test_delay_queue tests pass with these changes.

Squashing and merging.

Issue #8980, opened 2026-05-14
Server does not enforce absolute path for PID file


korydraughn:
If the server is launched using the following command, it exits immediately with a code of 0.

```
mkdir pid_dir
irodsServer -p pid_dir/irods.pid -d
```

The reason for this behavior is because `-d` (daemonize) is processed before the PID file path. The path needs to be validated before daemonizing the server. For this specific case, we only need to confirm the path starts with a forward slash.

https://github.com/irods/irods/blob/a997385ae36e1ff6eb6bc30d203e39105a382355/server/main_server/src/main.cpp#L223-L239

This is only an issue when `-d` is used. Relative paths work just fine without `-d`. However, we want to make sure the server honors the help text, so absolute paths need to be enforced.

PR #8984, opened 2026-05-15
[#8980] Reject relative paths to PID file on server startup (main)

korydraughn:
~Working on failed assertion. Putting in draft until that is fixed.~

New tests pass in isolation and when run with other test suites. No need to run the full test suite for this since the code changes apply to startup logic and help text.

This is now ready.

### DIFF
--- a/clients/icommands/src/istream.cpp
+++ b/clients/icommands/src/istream.cpp
@@ -118,45 +118,56 @@ int main(int argc, char* argv[])
 
 auto usage() -> void
 {
-    std::cout << "Usage: istream read [-R RESC_NAME] [-o INTEGER] [-c INTEGER] LOGICAL_PATH\n"
-                 "Usage: istream read [-n REPLICA_NUMBER] [-o INTEGER] [-c INTEGER] LOGICAL_PATH\n"
-                 "Usage: istream write [-R RESC_NAME] [-k] [-o INTEGER] [-c INTEGER] [--no-trunc] [-a] LOGICAL_PATH\n"
-                 "Usage: istream write [-n REPLICA_NUMBER] [-k] [-o INTEGER] [-c INTEGER] [--no-trunc] [-a] LOGICAL_PATH\n"
-                 "\n"
-                 "Streams bytes to/from iRODS via stdin/stdout.\n"
-                 "Reads bytes from the target data object and prints them to stdout.\n"
-                 "Writes bytes from stdin to the target data object.\n"
-                 "\n"
-                 "Writing to a non-existent data object will create it by default.\n"
-                 "Writing to an existing data object always truncates it.  Truncation can be\n"
-                 "disabled by passing --no-trunc.  However, using --no-trunc disables creation\n"
-                 "of data objects.\n"
-                 "\n"
-                 "If the client-side of the connection is interrupted before the data object\n"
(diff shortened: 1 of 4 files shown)


server/main_server/src/main.cpp
```diff
@@ -220,6 +220,15 @@ auto main(int _argc, char* _argv[]) -> int
             return 1;
         }
 
+        auto pid_file = irods::get_irods_runstate_directory() / "irods/irods-server.pid";
+        if (const auto iter = vm.find("pid-file"); std::end(vm) != iter) {
+            pid_file = iter->second.as<std::string>();
+            if (!pid_file.is_absolute()) {
```
korydraughn: Use `pid_file.is_relative()`. The intent becomes clearer and it results in the exact same behavior.

clients/icommands/src/istream.cpp
```diff
@@ -118,45 +118,54 @@ int main(int argc, char* argv[])
...
-                 "-c, --count     The number of bytes to read/write.  Value must be positive.\n"
-                 "                Defaults to all bytes.\n"
-                 "-a, --append    Appends bytes read from stdin to the data object.  Implies\n"
-                 "                --no-trunc.\n"
-                 "    --no-trunc  Does not truncate the data object.  Disables creation of data\n"
-                 "                objects.  Ignored by write operations when --append is set.\n"
-                 "-k, --checksum  Compute checksum.\n"
-                 "-h, --help      Prints this message\n";
+    std::cout
+        << "Usage: istream read [-R RESOURCE_NAME] [-o INTEGER] [-c INTEGER] LOGICAL_PATH\n"
```
korydraughn: Use a raw-string literal for this. Just makes editing the help text easier.

server/main_server/src/main.cpp

clients/icommands/src/istream.cpp
korydraughn: Done.

Added a summary line to istream, just like itouch.

scripts/irods/test/test_misc.py
```diff
@@ -601,6 +604,112 @@ def pep_database_gen_query_access_control_setup_except(rule_args, callback, rei)
...
+            # Stop the server so that the test can launch it with a different PID file.
+            # This avoids address-in-use issues, etc.
+            controller.stop()
+
+            # Test launching the server in the foreground with a custom PID file.
+            #
+            # "start_new_session" is used to make sure the server is launched under a different
+            # process group/session. Not doing this results in signals like SIGUSR1 (used by the server)
+            # being sent to this test, causing it to be interrupted and exiting.
+            pid_file_path = '/tmp/irods_server_foreground_issue_8980.pid'
```
korydraughn: This works, but I'm going to switch it to the service account or /etc/irods directory to avoid /tmp.

korydraughn: Done.

korydraughn:
Confirmed the new tests still pass with the recent changes. Tested only the new tests against ub22, ub24, and rocky9.

Even though running the full core test suite isn't necessary, I did it anyway over the weekend to be sure and all of the tests passed. The commit was  a938fa4b8459586538e0978b187a5973a24130db.

Squashing, pounding and merging.

trel:
nice!

Issue #8749, opened 2025-11-14
Release activities for 5.1.0


korydraughn:
- [ ] Bump server version
- [ ] Bump icommands version
- [ ] Release notes

korydraughn:
Consider documenting changes to database PEP in regard to no-scramble password option.

See conversation starting at https://github.com/irods/irods/pull/8732#discussion_r2515283236 for details.

---

UPDATE: Providing a picture of the linked conversation since clicking the link doesn't automatically jump to the conversation.

<img width="741" height="867" alt="Image" src="https://github.com/user-attachments/assets/735e8464-36a8-455b-a977-c1ac302c022f" />

korydraughn:
Consider bumping the GenQuery2 feature test macro to accommodate for the latest changes.
- Exposed create_ts and modify_ts for metadata attachment
- Fixed group permission expansion bug
  - _This is the primary motivator because it changes behavior in a significant way._
- Return error info via rError stack

korydraughn:
Other things to consider for the feature test macros.
- Built-in implementation of Logical Quotas
- Hashers and digest output format
- Rebalance improvements _(probably not since this isn't an API thing)_

PR #8996, opened 2026-05-29
[#8749] Bump version to 5.1.0 (main)

korydraughn:


test/c_api_test/test.c
```diff
@@ -15,7 +15,7 @@
 #include "irods/dataObjClose.h"
 #include "irods/checksum.h"
```
trel, line 16: are there any other .h files needed here we need to include/enforce?

korydraughn: My guess is that there's several files which need to be included.

korydraughn:
Taking this out of draft since we need to bump the version.

test/c_api_test/test.c
korydraughn: Any additional thoughts regarding this? If not, I'm going to resolve this conversation.

korydraughn:
Rebased this on top of main.

test/c_api_test/test.c
trel: not from me.

korydraughn:
Will pound and merge after github actions reports success.

korydraughn:
GitHub Actions succeeded.

Adding pound and merging.

Issue #8999, opened 2026-06-04
`rsModDataObjMeta` allows garbage to be assigned to create_ts


korydraughn:
While investigating https://github.com/irods/irods_client_http_api/issues/474, I discovered that  `rsModDataObjMeta` does not apply the same logic to the create time as it does for modify time, expiry time, and access time. All of those columns have the same traits as far as the database is concerned.

All of the timestamp columns are converted to integers and formatted before updating the catalog. Here's an example demonstrating the difference.

```sh
$ ils
/tempZone/home/rods:
$ itouch foo
$ ils -l
/tempZone/home/rods:
  rods              0 demoResc            0 2026-06-04.12:06 & foo
$ iquest "select DATA_CREATE_TIME where DATA_NAME = 'foo'"
DATA_CREATE_TIME = 01780589179
------------------------------------------------------------
$ iadmin modrepl logical_path /tempZone/home/rods/foo replica_number 0 DATA_CREATE_TIME thinger
$ iquest "select DATA_CREATE_TIME where DATA_NAME = 'foo'"
DATA_CREATE_TIME = thinger
------------------------------------------------------------
$ iadmin modrepl logical_path /tempZone/home/rods/foo replica_number 0 DATA_MODIFY_TIME thinger
$ iquest "select DATA_CREATE_TIME, DATA_MODIFY_TIME where DATA_NAME = 'foo'"
DATA_CREATE_TIME = thinger
DATA_MODIFY_TIME = 00000000000
------------------------------------------------------------
```

The reason that happens is because there is no code for the create time like there is for the other time columns.

https://github.com/irods/irods/blob/677792c11d9819d941d4dabc5e81fcddcf2d7147/plugins/database/src/db_plugin.cpp#L2354-L2391

Adding a code block for `DATA_CREATE_KW` should be all that's needed to resolve the issue.

trel:
wild.   yes, please make it good.

PR #9000, opened 2026-06-04
[#8999] rsModDataObjMeta: Format string as timestamp for DATA_CREATE_KW (main)

korydraughn:
Confirmed the new test passes at the bench.

Will run the unit tests and core tests next.

Will open an issue about the 11 byte limit.

### DIFF
--- a/unit_tests/src/test_rc_mod_data_obj_meta.cpp
+++ b/unit_tests/src/test_rc_mod_data_obj_meta.cpp
@@ -1,12 +1,15 @@
 #include <catch2/catch_all.hpp>
 
+#include "irods_error_enum_matcher.hpp"
+#include "unit_test_utils.hpp"
+
 #include "irods/client_connection.hpp"
 #include "irods/dataObjInpOut.h"
 #include "irods/dstream.hpp"
 #include "irods/filesystem.hpp"
 #include "irods/irods_at_scope_exit.hpp"
-#include "irods_error_enum_matcher.hpp"
 #include "irods/irods_exception.hpp"
+#include "irods/irods_query.hpp"
 #include "irods/objInfo.h"
 #include "irods/rcMisc.h"
 #include "irods/replica.hpp"
@@ -14,7 +17,6 @@
 #include "irods/rodsDef.h"
 #include "irods/rodsErrorTable.h"
 #include "irods/transport/default_transport.hpp"
-#include "unit_test_utils.hpp"
 
 #include <fmt/format.h>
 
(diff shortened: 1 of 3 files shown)


unit_tests/src/test_rc_mod_data_obj_meta.cpp
```diff
@@ -148,3 +150,55 @@ TEST_CASE("using ADMIN_KW with rcModDataObjMeta")
         }
     }
 } // rc_data_obj_repl
+
+TEST_CASE("rsModDataObjMeta formats string as timestamp for DATA_CREATE_KW")
```
alanking: Worth mentioning the issue number here?

alanking (commented):
Checks out. Any interest in addressing clang-tidy in db_plugin, or just let it ride since it aligns with the existing code blocks?

unit_tests/src/test_rc_mod_data_obj_meta.cpp
korydraughn: I originally had it there, but removed it. I flip flop on that all the time.

I can add it back if it seems helpful. Either way is fine with me.

korydraughn:
> Checks out. Any interest in addressing clang-tidy in db_plugin, or just let it ride since it aligns with the existing code blocks?

Ah, that must've appeared after I fixed the other clang-tidy report. Will review the latest report.

korydraughn:
> Will open an issue about the 11 byte limit.

Opened #9001.

korydraughn:
Catch2 unit tests pass.

Hmm, clang-tidy is not reporting a problem in db_plugin.cpp.

Running core tests.

alanking:
> Hmm, clang-tidy is not reporting a problem in db_plugin.cpp.

Oh, I think this is because the file has gotten too big for clang-tidy to analyze it. lol Here's what it says:
> Terminated by timeout: clang-tidy -line-filter=[{"name":"plugins/database/src/db_plugin.cpp","lines":[[2393,2405]]}] -quiet -p=build/compile_commands.json plugins/database/src/db_plugin.cpp

 I imagine it would have issues with the naked `char` array, C-style cast, etc. Up to you if you actually want to address them since the code block does match the existing ones (not that that's a great reason to keep it the way that it is).

korydraughn:
Will consider looking into that.

korydraughn:
All core tests pass except one.

It is a modrepl test which needs its assertion updated. This was expected.

korydraughn:
See latest commit for the test change/fix.

Test passes now.

unit_tests/src/test_rc_mod_data_obj_meta.cpp
korydraughn: Will include the issue number since I've included in the dstream PR.

korydraughn: Done.

alanking (commented):
Looks good. Squash

korydraughn:
Squashed.

alanking (approved):
\# it

korydraughn:
Added pounds.

Issue #9002, opened 2026-06-05
Narrowing conversion error possible in default_transport.hpp


korydraughn:
The following lines involve an implicit cast from `std::streamsize` (i.e. `long int`)  to `int`. This can result in precision lost.

https://github.com/irods/irods/blob/677792c11d9819d941d4dabc5e81fcddcf2d7147/lib/core/include/irods/transport/default_transport.hpp#L202

https://github.com/irods/irods/blob/677792c11d9819d941d4dabc5e81fcddcf2d7147/lib/core/include/irods/transport/default_transport.hpp#L215

korydraughn:
I noticed I used `_LOST` instead of `_LOSS` in the new error error code.

After some discussion with @alanking, we've agreed to rename the error code to `NARROWING_CONVERSION_ERROR`.

trel:
> NARROWING_CONVERSION_ERROR 

i like that.

Issue #8459, opened 2025-04-30
Investigate how to expose iRODS error codes through dstream


korydraughn:
- [ ] main
- [ ] 4-3-stable
---
## Enhancement

The dstream library is modeled after the C++ fstream library. Neither of these expose error information received from the OS.

Please investigate how we can expose iRODS error codes to users through the dstream library.

```cpp
dstream in; // Assume this points to a valid data object.

std::string s;
in >> s; // Assume this fails. How do we expose the error code(s)?
```

korydraughn:
Related to #8627.

PR #9003, opened 2026-06-05
[#8459,#9002] dstream: Track error codes and check for precision lost (main)

korydraughn:
Derived from https://github.com/irods/irods/compare/main...korydraughn:irods:8459.m.original.complete.

Will run the tests. I don't expect there to be any issues.


alanking (commented):
Tests?

korydraughn:
On the way. I meant to write those before opening this.

korydraughn:
Added tests.

korydraughn:
Clang-tidy wants more const. Will take care of that and run the tests.

korydraughn:
Also need to add a feature test macro so that the HTTP API can detect when `last_error` is available.

unit_tests/src/test_dstream.cpp
```diff
@@ -305,6 +310,96 @@ TEST_CASE("dstream", "[iostreams]")
...
+
+    SECTION("passing an invalid offset to seek")
+    {
+        const auto data_object = sandbox / "data_object.txt";
+
+        io::odstream out{tp, data_object};
+        CHECK(out);
+        CHECK_FALSE(out.seekp(-1, std::ios::beg));
+        auto errc = irods::experimental::make_error_code(out.last_error());
+        CHECK(irods::experimental::get_irods_error_code(errc) == UNIX_FILE_LSEEK_ERR);
```
alanking, line 348: Why do we make an error code and get an error code here, but not in the precision error test below?

alanking (commented):
Just had one comprehension question

unit_tests/src/test_dstream.cpp
korydraughn: The non-precision test cases were failing because errno was being included in the codes (sometimes).

We use the highlighted logic to extract the pure iRODS error code.

The precision test cases, don't have that problem because the transport always returns the pure error code - i.e. no syscalls or errno involved.

alanking: Got it, thanks. Resolving.

alanking (commented):
Squash

korydraughn:
> Derived from [main...korydraughn:irods:8459.m.original.complete](https://github.com/irods/irods/compare/main...korydraughn:irods:8459.m.original.complete).
> 
> Will run the tests. I don't expect there to be any issues.

No need to run the core tests for this. The code in this PR was pulled from the referenced branch `8459.m.original.complete`. The only difference is that I had to provide a base implementation for `last_error()` since the pure virtual function would require users to modify their code. Nothing outside of the Catch2 tests use the new code. 

I've verified that the full dstream Catch2 test suite passes too.

Squashing.

korydraughn:
Squashed.

The other changes in the branch which this PR comes from will have to wait until iRODS 6 unless we want to introduce macros like `IRODS_NODISCARD` which then changes what the code does depending on the version of the dev library. I'm still considering this. Happy to talk about it for anyone interested.

alanking (approved):
> The other changes in the branch which this PR comes from will have to wait until iRODS 6 unless we want to introduce macros like `IRODS_NODISCARD` which then changes what the code does depending on the version of the dev library. I'm still considering this. Happy to talk about it for anyone interested.

Right, we can add those later. Is there an issue for it?

\# when ready

korydraughn:
> Right, we can add those later. Is there an issue for it?

Not yet. Will do that soon so it's not lost.

korydraughn:
Going to change the summary commit message for the first commit from:

```
[8459] dstream: Track iRODS error codes
```

to:

```
[8459] dstream: Track and make iRODS error codes available
```

korydraughn:
I'm thinking about renaming the feature test macro to the following.

```
IRODS_LIBRARY_FEATURE_DSTERAM
```

The reason being that there are a few features which it will grow later.
- Support for the ADMIN_KW
- Support for more keyword options (i.e. condInput)

Then we just bump the integer whenever they appear. No new macro needed.

Thoughts?

_I'm just now noticing that the macro has DSTERAM instead of DSTREAM. Will fix that._

alanking:
`IRODS_LIBRARY_FEATURE_DSTREAM` seems reasonable. I guess it doesn't matter when it was first implemented because we will change the value whenever a new feature is added to it.

korydraughn:
True. I'm wondering if it matters that dstream already exists. Is it confusing to not have the macro be more explicit in this case?

korydraughn:
All of the other macros are for completely new things.

alanking:
I think it would be better to have the more general feature test macros, personally. It would align more with the C++ ones on which (I assume) these are based: https://en.cppreference.com/cpp/feature_test

korydraughn:
Yep.

Okay. Going with that.

korydraughn:
Running unit tests to make sure the library features test passes.

alanking (commented):
Squash on passing tests

korydraughn:
Tests pass.

Squashed.

korydraughn:
Adding pounds.

korydraughn:
Added pounds.

alanking (approved):

PR #9010, opened 2026-06-11
[#9017] add test for issue python rule plugin issue 258

d-w-moore:


scripts/irods/test/test_prep_genquery_iterator.py
```diff
@@ -1109,3 +1183,4 @@ def pep_api_touch_pre(rule_args, callback, rei):
 
         finally:
             IrodsController().reload_configuration()
+#
```
trel: remove?

alanking:
Does this supersede https://github.com/irods/irods_rule_engine_plugin_python/pull/260?

trel:
> Does this supersede [irods/irods_rule_engine_plugin_python#260](https://github.com/irods/irods_rule_engine_plugin_python/pull/260)?

i believe it does.

scripts/irods/test/test_prep_genquery_iterator.py
```diff
@@ -172,6 +172,80 @@ def main(args,callback,rei):
...
+                    )
+                    expected_result_rows = len(results_list := list(query))
+                    expected_total_rows = query.total_rows()
+                    callback.writeLine('stdout',repr([locals()[key] for key in {output_table_columns}]))
+
+                INPUT null
+                OUTPUT ruleExecOut
+                ''').format(**ns))
+
+            with self.subTest(msg='Test failed for parameter combination:',  **parameters_for_test):
```
alanking: What does this `msg` mean? Are we expecting a failure below?

scripts/irods/test/test_prep_genquery_iterator.py
```diff
@@ -172,6 +172,80 @@ def main(args,callback,rei):
...
+                output, err, rc = self.admin.run_icommand("irule -r irods_rule_engine_plugin-python-instance -F rule.r")
+                self.assertTrue(rc == 0, "icommand status ret = {r} output = '{o}' err='{e}'".format(r=rc,o=output,e=err))
```
alanking: Could be `assertEqual`

scripts/irods/test/test_prep_genquery_iterator.py
```diff
@@ -172,6 +172,80 @@ def main(args,callback,rei):
...
+        )
+
+        # Test collection creation
+        for n in range(10):
+            self.admin.assert_icommand(
+            f"""imkdir -p {self.admin.session_collection}/issue_{n}_258"""
+            f"""          {self.admin.session_collection}/Issue_{n}_258""")
+
+        for test_vector in table_of_test_vectors:
+            output_table_columns = [col_name for col_name in table_columns if not col_name.endswith('_')]
```
alanking: It looks like we might be able to move this outside the loop? Maybe I'm missing something.

scripts/irods/test/test_prep_genquery_iterator.py
```diff
@@ -172,6 +172,80 @@ def main(args,callback,rei):
...
+                    query = Query(
+                        callback,
+                        ["COLL_NAME"],
+                        """COLL_NAME like '%/issue%' and COLL_PARENT_NAME = '{self.admin.session_collection}'"""
+                        , case_sensitive={case_sensitive_}
+                        , offset={offset_}
+                        , limit={limit_}
+                    )
+                    expected_result_rows = len(results_list := list(query))
+                    expected_total_rows = query.total_rows()
```
alanking: I see that we have to name these variables as such since we want to print only the `output_table_columns`, but it is a little confusing that these are in fact the *actual* values, not the *expected* values. To avoid confusion, can we change how we are doing this print, or maybe leave an explanatory comment above?

alanking (commented):
Shaping up

scripts/irods/test/test_prep_genquery_iterator.py
d-w-moore: My understanding is the `msg` prints out only when there's a failure.  I could be wrong.

d-w-moore:
> > Does this supersede [irods/irods_rule_engine_plugin_python#260](https://github.com/irods/irods_rule_engine_plugin_python/pull/260)?
> 
> i believe it does.

yes

scripts/irods/test/test_prep_genquery_iterator.py
d-w-moore: I think I'm not quite understanding how I would change the print. Maybe it would be better if I just left out the `expected_` in each variable name?

scripts/irods/test/test_prep_genquery_iterator.py
alanking: Oh, I just read the docs and it looks like you're right. I could've sworn I've seen these messages on success too. Anyway, good to know. Resolving

scripts/irods/test/test_prep_genquery_iterator.py
alanking: It would certainly make it a bit more readable, IMO. The result would look something like...
```python
result_rows = len(results_list := list(query))
total_rows = query.total_rows()
callback.writeLine('stdout',repr(result_rows))
callback.writeLine('stdout',repr(total_rows))
```
This, of course, kind of makes for multiple sources of truth, but it certainly seems less confusing to me. Up to you whether/how you want to do it.

scripts/irods/test/test_prep_genquery_iterator.py
d-w-moore: done

scripts/irods/test/test_prep_genquery_iterator.py
d-w-moore: done

scripts/irods/test/test_prep_genquery_iterator.py
d-w-moore: done

scripts/irods/test/test_prep_genquery_iterator.py
d-w-moore: @alanking ok see if what I've done there makes sense?

alanking: Yes, I think it is less confusing now. Thanks. Resolving.

alanking (commented):
Okay, so what's left to do here? Test passes? Runs along with the other ones when the PREP test hook is executed?

d-w-moore:
Passes test suite

scripts/irods/test/test_prep_genquery_iterator.py
```diff
@@ -172,6 +172,91 @@ def main(args,callback,rei):
         self.assertTrue(rc == 0, "icommand status ret = {r} output = '{o}' err='{e}'".format(r=rc,o=output,e=err))
         self.assertEqual(output.strip(), "[1,3]")
 
+    @unittest.skipIf(plugin_name == 'irods_rule_engine_plugin-irods_rule_language', 'only applicable for python REP')
+    def test_query_params__issue_258__irods__irods_rule_engine_plugin_python(self):
```
korydraughn: Rename test to follow existing conventions.

d-w-moore: Since the issue number is not local, is there a guide to proper conventions?
Do we want something like this?:

`test_query_params_are_cumulative__irods_rule_engine_plugin_python__issue_258`

korydraughn: Open an issue in this repo (irods/irods) and reference the PREP issue in it.

Then use the irods/irods issue number in the test name. You can add a comment to the test which references the PREP issue number.

alanking: https://github.com/irods/irods/issues/9017 has been created and it is linked to https://github.com/irods/irods_rule_engine_plugin_python/issues/258. Just waiting on the test name to be updated now.

d-w-moore: test passed under the irods testing environment.

korydraughn: Good.

Update the test name to reference the irods/irods issue number instead of the PREP issue number and then we can resolve this conversation.

d-w-moore: Ok if we like the test name now, I can squash down to something like

```
[_9017] add test for total_rows fix (PREP issue 258)
```

korydraughn: You can reference the PREP issue number in the summary line or the body of the commit message. For example:

```
[9017,irods/irods_rule_engine_plugin_python 258] add test for total_rods fix
```

korydraughn (commented):
Squash it.

trel:
appears to be squashed.

korydraughn:
Please drop the first instance of the word `issue` from the commit message.

```
[_9017] add test for python rule plugin issue 258
```

korydraughn (approved):
Please resolve the one review comment and pound it.

scripts/irods/test/test_prep_genquery_iterator.py
d-w-moore: done

korydraughn:
Merging.

PR #9020, opened 2026-07-13
[9012] ilsresc returns nonzero on nonexistent resc (main)

FifthPotato:
Taking comments.

### DIFF
--- a/scripts/irods/test/test_ilsresc.py
+++ b/scripts/irods/test/test_ilsresc.py
@@ -36,7 +36,7 @@ class Test_Ilsresc(session.make_sessions_mixin(admins, users), unittest.TestCase
 
             # Show that iquest does not print an error when the total number of resources
             # is a multiple of the GenQuery page size.
-            _, err, ec = self.admin.run_icommand(['ilsresc', '-l'])
+            _, _, ec = self.admin.run_icommand(['ilsresc', '-l'])
             self.assertEqual(ec, 0)
             self.assertNotIn('rcGenQuery failed with error -808000 CAT_NO_ROWS_FOUND', err)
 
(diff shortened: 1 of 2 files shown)


trel:
i don't suppose we can make the test assertion any stronger...  since we don't know the value of `status`... even though it will *usually* be `CAT_NO_ROWS_FOUND`...

scripts/irods/test/test_ilsresc.py
```diff
@@ -48,3 +48,6 @@ def test_ilsresc_does_not_print_CAT_NO_ROWS_FOUND_at_page_boundary__issue_5155(s
     def test_ilsresc_reports_an_error_on_unknown_zone__issue_6022(self):
         self.admin.assert_icommand(['ilsresc', '-z', 'unknown_zone_issue_6022'], 'STDERR', ['-26000 SYS_INVALID_ZONE_NAME'])
 
+    def test_ilsresc_returns_nonzero_on_nonexistent_resource__issue_9012(self):
+        _, err, ec = self.admin.run_icommand(['ilsresc', '-l', 'nonexistent_resc'])
```
alanking, line 52: Could make `err` a placeholder `_` as well, since it's not being used. Else, we can make an assertion as to its value.

alanking (commented):
I think the only way to make the assertion stronger is to show that the return value is 1 and not non-zero. However, the only requirement of the issue is that the value be non-zero, so I think it's fine. Just my two cents

trel:
okay, yes, agreed.  non-zero it is.

korydraughn:
@FifthPotato Have you run this against the core test suite?

FifthPotato:
> @FifthPotato Have you run this against the core test suite?

Not yet. Will do so now.

korydraughn:
See @alanking's review comments too.

FifthPotato:
All core tests pass except for `test_delay_queue.Test_Delay_Queue`, and after checking, it shouldn't be related? Nothing related to `ilsresc` in the failing block.

alanking (commented):
Looks good. The test you mentioned is very timing dependent, so I'm pretty sure that's not related to your change. Just run it once more in isolation to be sure and I think we'll be good. I think this is good to squash

FifthPotato:
Squished.

alanking (approved):
\# it

FifthPotato:
Octothorpe'd.

PR #9022, opened 2026-07-16
[#9027] istream: Report dstream error codes on failure (main)

korydraughn:
test_istream.py tests pass.

Running core tests suite now.

Putting in draft until I see the tests pass.

### DIFF
--- a/clients/icommands/src/istream.cpp
+++ b/clients/icommands/src/istream.cpp
@@ -39,7 +39,8 @@ auto check_input_arguments(const po::variables_map& vm) -> std::tuple<bool, int>
 
 auto canonical(const std::string& path, rodsEnv& env) -> std::optional<std::string>;
 
-auto stream_bytes(std::istream& in, std::ostream& out, int_type count) -> int;
+template <typename InputStream, typename OutputStream>
+auto stream_bytes(InputStream& in, OutputStream& out, int_type count) -> int;
 
 auto read_data_object(rodsEnv& env,
                       const po::variables_map& vm,
@@ -216,7 +217,8 @@ auto canonical(const std::string& path, rodsEnv& env) -> std::optional<std::stri
     return p;
 }
 
-auto stream_bytes(std::istream& in, std::ostream& out, int_type count) -> int
+template <typename InputStream, typename OutputStream>
+auto stream_bytes(InputStream& in, OutputStream& out, int_type count) -> int
 {
     if (count < 0) {
         std::cerr << "Error: Invalid byte count.\n";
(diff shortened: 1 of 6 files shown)


clients/icommands/src/istream.cpp
```diff
@@ -51,7 +51,7 @@ auto write_data_object(rodsEnv& env, const po::variables_map& vm, io::client::de
 
 int main(int argc, char* argv[])
 {
-    utils::set_ips_display_name(boost::filesystem::path{argv[0]}.filename().c_str());
+    utils::set_ips_display_name("istream");
```
trel, line 54: clang-tidy didn't like this as it was?

korydraughn: Clang-tidy doesn't like the `argv[0]`. It identifies it as pointer arithmetic.

korydraughn:
All core tests pass.

clients/icommands/src/istream.cpp
korydraughn: I hard coded the string since the binary's name is unlikely to ever change.

trel: right - fair enough.

korydraughn:
All unit tests pass.

Will open issues for commits next.

clients/icommands/src/istream.cpp
```diff
@@ -232,12 +234,23 @@ auto stream_bytes(std::istream& in, std::ostream& out, int_type count) -> int
         }
 
         if (!in.eof()) {
-            std::cerr << "Error: Failed to read all bytes.\n";
+            if constexpr (std::is_same_v<InputStream, io::idstream>) {
+                std::cerr << "Error: Failed to read all bytes [error code=" << in.last_error() << "].\n";
```
alanking, line 238: The dstream error reporting is all in the client library, correct? There's no situation in which this error code would come back as 0 or something due to its not being set in the past?

I am thinking about the compatibility between 5.1.x `istream` and a 5.0.x server.

alanking (commented):
Just had one question. Otherwise seems good

clients/icommands/src/istream.cpp
korydraughn: > The dstream error reporting is all in the client library, correct?

Yes.

> There's no situation in which this error code would come back as 0 or something due to its not being set in the past?
> 
> I am thinking about the compatibility between 5.1.x `istream` and a 5.0.x server.

For this specific case, I do not believe we'd ever see 0 reported from `last_error()`.

alanking: Okay, just making sure.

alanking:
Please rebase and squash

korydraughn:
Rebased and squashed.

korydraughn:
I'm going to update test_logical_quotas.py's use of istream so that they include the logical quotas error code.

alanking (approved):
\# it

korydraughn:
test_logical_quotas.py changes passed.

Adding pounds.

korydraughn:
Added pounds.

PR #9023, opened 2026-07-16
[#9026] Report error when Touch API fails to create data object (main)

korydraughn:

No new tests for this PR since it's unclear how to trigger the error branch.

These changes can be viewed as a bug fix because they allow the touch API to detect and report when the creation of a data object fails. I'll open an issue for that if we don't have one already. I'll update the commit message to reflect that as well.

Putting in draft for now.

### DIFF
--- a/.github/workflows/build-irods-debian.yml
+++ b/.github/workflows/build-irods-debian.yml
@@ -14,7 +14,7 @@ jobs:
 
     steps:
       - name: Checkout Repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@v7
 
       - name: Install Prerequisites
         run:  |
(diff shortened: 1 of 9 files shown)


korydraughn:
All unit tests pass.

Running core tests.

korydraughn:
All core tests pass.

Will open issues for commits next.

korydraughn:
Added issue numbers to commits and squashed.

alanking (commented):
Didn't spot anything that looked wrong. I'm assuming that we do *not* consider changing the output like this as a breaking change?

korydraughn:
That crossed my mind, but I don't consider error message strings to be in the category of a breaking change. They are purely for the human IMO - this should be documented if we all agree.

alanking (approved):
Seems reasonable. # it

korydraughn:
Added pounds.

Issue #7074, opened 2023-04-27
Agent crashes when rule input is mishandled


alanking:
### iRODS Version, OS and Version
4.2.12ish (probably affects all versions)

### What did you try to do?
I had the following rule:
```python
# rule.r
check_access {
    *access = 0;
    msiCheckAccess("*logical_path", "read object", *access);
}
INPUT *logical_path="/tempZone/home/alan/foo"
OUTPUT ruleExecOut
```
And I executed the rule file via `irule` as user `alan`:
```bash
$ irule -r irods_rule_engine_plugin-irods_rule_language-instance -F rule.r /tempZone/home/rods/foo
```

### Expected behavior
Regardless of whether the rule is correct or the invocation of `irule` is correct, I expect agents not to crash.

### Observed behavior (including steps to reproduce, if applicable)
The agent crashes with the following stacktrace:
```
Caught signal [11]. Dumping stacktrace and exiting

Dumping stack trace
<0>     Offset:         Address: 0x7f5b7e2f1143 /usr/lib/libirods_server.so.4.2.11(+0xb52143) [0x7f5b7e2f1143]
<1>     Offset:         Address: 0x7f5b7b472980 /lib/x86_64-linux-gnu/libpthread.so.0(+0x12980) [0x7f5b7b472980]
<2>     Offset: 0x137   Address: 0x7f5b787f1437 nextRuleGenStringExpression(pointer*, ParserContext*, token*)
```
And the output from `irule`:
```
remote addresses: 127.0.0.1 ERROR: rcExecMyRule error.  status = -154000 SYS_INTERNAL_ERR
```
Please investigate.

alanking:
Just in case I don't finish in time, here's the function mentioned in the stacktrace (the name is generated via macros):
https://github.com/irods/irods/blob/3af66900a0189690deaece96d61f26a52b6c3f7b/plugins/rule_engines/irods_rule_language/src/parser.cpp#L1273-L1338

korydraughn:
How did we end up in NREP parser code? Does the crash happen in there?

trel:
so ... we could easily rr this?   and see where it breaks?

alanking:
> How did we end up in NREP parser code? Does the crash happen in there?

I suppose so, based on the stacktrace.

> so ... we could easily rr this? and see where it breaks?

The plan is to sniff things out with address sanitizer first, yes.

alanking:
[...]

PR #9029, opened 2026-07-17
[#7074] Prevent crash on invalid rule input (main)

MartinFlores751:
Issue quick link: #7074

This PR prevents the iRODS agent from crashing by checking the return code from a helper function before assigning a token type.

### DIFF
--- a/plugins/rule_engines/irods_rule_language/src/parser.cpp
+++ b/plugins/rule_engines/irods_rule_language/src/parser.cpp
@@ -255,8 +255,16 @@ Token* nextTokenRuleGen( Pointer* e, ParserContext *context, int rulegen, int pa
                 }
             }
             else if ( pathLiteral && ch == '/' ) { /* path */
-                nextStringBase( e, token->text,MAX_TOKEN_TEXT_LEN, "),; \t\r\n", 0, '\\', 0, token->vars ); /* path can be used in a foreach loop or assignment, or as a function argument */
-                token->type = TK_PATH;
+                // Handle no string found case
+                // path can be used in a foreach loop or assignment, or as a function argument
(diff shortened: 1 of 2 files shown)


korydraughn (commented):
Code looks good.

Have you run the unit and core tests?

MartinFlores751:
Ran the core and unit tests. There was only one test failure in `test_ilsresc`, in the test `test_ilsresc_does_not_print_CAT_NO_ROWS_FOUND_at_page_boundary__issue_5155`, which was already fixed in a separate commit in main.

All other tests passed.

korydraughn:
Squash it.

korydraughn:
Also, add more context to the body of the commit message. Feels important to capture what you learned.

MartinFlores751:
Squashed and commit message added.

korydraughn:
Do we know why the agent crashes without that check?

MartinFlores751:
The agent crashes because the `vars` member of the variable `token` was not set, and the assumption internally is that a valid path token will have a `vars` sentinel value set. In this case the `vars` member is supposed to be set in `nextStringBase(...)`. However, since the end of the string is reached before one of the character delimiters is found, `vars` is unset and a `-1` is returned from the function. Previously, the return code of `nextStringBase(...)` was unchecked, so the agent eventually got hold of the poorly formed token.

korydraughn:
Please include some version of that wording in the commit message body.

MartinFlores751:
Updated the commit message.

korydraughn:
Excellent.

Pound it.

MartinFlores751:
Pounded.

korydraughn (approved):

korydraughn:
Merging.

Issue #8416, opened 2025-04-22
test_delay_queue_with_long_job can fail sometimes


korydraughn:
- [ ] main
- [ ] 4-3-stable
---
## Testing

I observed this failure while testing #8398. I had to run test_delay_queue.py a few times to trigger it, which indicates it may be timing related.

```xml
        <testcase classname="irods.test.test_delay_queue.Test_Delay_Queue" name="test_delay_queue_with_long_job" time="28.381" timestamp="2025-04-22T19:22:42" file="scripts/irods/test/test_delay_queue.py" line="123">
                <!--Test to ensure that a long-running delayed job does not block other jobs from executing.

        Schedules a long-running job, a block of short-running jobs to be run as soon as possible, and a block
        of short-running jobs to be run at a later time (but before the long-running job completes).

        The delay server is limited to 2 concurrent threads of execution. One of the threads will be occupied
        with the long-running job while the other will execute the short-running jobs. This demonstrates the
        delay server's ability to execute multiple jobs per query as well as to pick up new jobs even as a
        long-running job continues to execute.
        -->
                <failure type="AssertionError" message="5 != 4"><![CDATA[Traceback (most recent call last):
  File "/var/lib/irods/scripts/irods/test/test_delay_queue.py", line 232, in test_delay_queue_with_long_job
    self.assertEqual(delay_job_batch_size, self.count_strings_in_queue('later:'))
AssertionError: 5 != 4
]]></failure>
```

Here's a link to the line which fails.

https://github.com/irods/irods/blob/8d5a54324a2c710fc41d2cb5fb2e469eac609d68/scripts/irods/test/test_delay_queue.py#L232

Please investigate.

FifthPotato:
Data point: also ran into this while testing for https://github.com/irods/irods/pull/9020
Likely unrelated, but notably, this one happens consistently.

PR #9031, opened 2026-07-21
[8975] Guard clauses for argument overflow in icmds (main)

FifthPotato:
Individual tests pass, but full suite hasn't been run yet.
Not entirely sure about the approach here, so taking comments. I did consider some other approaches such as just printing usage when overflowing, but I was worried that wasn't clear enough what the issue was in case a user organically ran into this.

### DIFF
--- a/clients/icommands/src/iadmin.cpp
+++ b/clients/icommands/src/iadmin.cpp
@@ -1140,6 +1140,10 @@ getInput( char *cmdToken[], int maxTokens ) {
     nTokens = 0;
     tokenFlag = 0;
     for ( i = 0; i < lenstr; i++ ) {
+        if (nTokens >= maxTokens) {
+            fprintf(stderr, "Too many command line arguments.\n");
+            return 0;
+        }
         if ( ttybuf[i] == '\n' ) {
             ttybuf[i] = '\0';
             cmdToken[nTokens++] = cpTokenStart;
@@ -1837,6 +1841,12 @@ main( int argc, char **argv ) {
     int i;
     const int maxCmdTokens = 20;
     char *cmdToken[maxCmdTokens];
+
+    if (argc - argOffset > maxCmdTokens) {
+        fprintf(stderr, "Too many command line arguments.\n");
+        return 1;
+    }
+
     for ( i = 0; i < maxCmdTokens; i++ ) {
         cmdToken[i] = "";
     }
--- a/scripts/irods/test/test_iadmin.py
+++ b/scripts/irods/test/test_iadmin.py
@@ -2242,6 +2242,11 @@ class Test_Iadmin(resource_suite.ResourceBase, unittest.TestCase):
         finally:
             self.admin.run_icommand(['iadmin', 'rmzone', good_zone_name])
 
+    def test_iadmin_prints_error_message_on_argument_overflow__issue_8975(self):
+        # 20 arguments is the limit
+        ec, _, _ = self.admin.assert_icommand(['iadmin', *['potato'] * 21], 'STDERR', 'Too many command line arguments.')
+        self.assertNotEqual(ec, 0)
+
 class Test_Iadmin_Resources(resource_suite.ResourceBase, unittest.TestCase):
 
     def setUp(self):
--- a/clients/icommands/src/igroupadmin.cpp
+++ b/clients/icommands/src/igroupadmin.cpp
@@ -144,6 +144,10 @@ getInput( char *cmdToken[], int maxTokens ) {
     nTokens = 0;
     tokenFlag = 0;
     for ( i = 0; i < lenstr; i++ ) {
+        if (nTokens >= maxTokens) {
+            fprintf(stderr, "Too many command line arguments.\n");
+            return;
+        }
         if ( ttybuf[i] == '\n' ) {
             ttybuf[i] = '\0';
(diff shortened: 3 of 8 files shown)


korydraughn:
Remove pound from commit message.

clients/icommands/src/iadmin.cpp
```diff
@@ -1140,6 +1140,10 @@ getInput( char *cmdToken[], int maxTokens ) {
     nTokens = 0;
     tokenFlag = 0;
     for ( i = 0; i < lenstr; i++ ) {
+        if (nTokens >= maxTokens) {
+            fprintf(stderr, "Argument count exceeds maximum.\n");
+            return 0;
```
trel, line 1145: return 1?

i see the tests are happy... so... honest question... where does this get returned that it is okay that an error occurred?

clients/icommands/src/iadmin.cpp
```diff
@@ -1837,6 +1841,12 @@ main( int argc, char **argv ) {
     int i;
     const int maxCmdTokens = 20;
     char *cmdToken[maxCmdTokens];
+
+    if (argc - argOffset > maxCmdTokens) {
+        fprintf(stderr, "Argument count exceeds maximum.\n");
```
korydraughn: How do you feel about the following?

```
Too many command line arguments.
```

korydraughn (commented):
I think the approach you've taken is what we want. Printing the help text is likely too noisy.

Definitely run the unit and core tests to make sure things are happy.

clients/icommands/src/iadmin.cpp
FifthPotato: I can change it to that.

clients/icommands/src/iadmin.cpp
FifthPotato: This is interactive mode!

FifthPotato: Specifically, the return value of this function just isn't cared about at all. Not sure why it returns `int` here...
In the other files, the corresponding function returns `void`.

https://github.com/FifthPotato/irods/blob/f8bfba060ea02eb49ddbd28eca14616c27de942c/clients/icommands/src/iadmin.cpp#L1997

trel: yep, good.  thanks.   resolving.

scripts/irods/test/test_iadmin.py
```diff
@@ -2244,7 +2244,7 @@ def test_modzone_does_not_accept_invalid_zone_names_for_remote_zones__issue_7722
 
     def test_iadmin_prints_error_message_on_argument_overflow__issue_8975(self):
         # 20 arguments is the limit
-        ec, _, _ = self.admin.assert_icommand(['iadmin', *['potato'] * 21], 'STDERR', 'Argument count exceeds maximum.')
+        ec, _, _ = self.admin.assert_icommand(['iadmin', *['potato'] * 21], 'STDERR', 'Too many command line arguments.')
```
korydraughn, line 2247: How do these tests behave if you mix in options/flags e.g. `isysmeta ls -l <more_args>`?

Notice the `-l` in the example.

FifthPotato: From checking by hand, `-l` gets treated as a flag and ignored for the count, but everything else (`ls`, `<more_args>`) counts towards the 20 limit.
Are we okay with that?

korydraughn: Does that mean the presence of multiple flags do not impact the bug you're fixing?

FifthPotato: I believe so. Passing `-l` and `-l -v` to `isysmeta ls` did not change the argument count (nor did it seem to trigger ASAN). Only `ls` seemed to be counted as an argument.

korydraughn:
Have you run this through the unit and core tests?

FifthPotato:
> Have you run this through the unit and core tests?

Running those now.

FifthPotato:
All tests pass except for `test_delay_queue.Test_Delay_Queue`, but I'm reasonably sure that's unrelated. Likely another occurrence of https://github.com/irods/irods/issues/8416 ?

korydraughn:
I doubt the test failures are related too, but check the test output to be sure.

FifthPotato:
> I doubt the test failures are related too, but check the test output to be sure.

```
======================================================================                   FAIL [27.591s]: test_delay_queue_with_long_job (irods.test.test_delay_queue.Test_Delay_Qu
eue)                                                                                     Test to ensure that a long-running delayed job does not block other jobs from executing.
----------------------------------------------------------------------                                         
2026-07-23 18:45:15,824 - Traceback (most recent call last):
  File "/var/lib/irods/scripts/irods/test/test_delay_queue.py", line 229, in test_delay_q
ueue_with_long_job                                                                       
    self.assertFalse(lib.metadata_attr_with_value_exists(
AssertionError: True is not false
```
Looks like it's in the same test, but a different failure?

korydraughn:
I don't see any instances of the icommands you modified in that test. I think we're good to go.

Squash to taste.

FifthPotato:
Squished.

korydraughn (approved):
Pound it.

FifthPotato:
Octothorpe'd.

Issue #8933, opened 2026-04-17
`ils` should not core dump on incorrect file or directory permissions


MartinFlores751:
### iRODS Version, OS and Version
Tip of main (6e8d56370) debug build, container using Ubuntu 24.04

### What did you try to do?
`ils` after copying the `.irods` dir of the `irods` user to another user (`dev`).

### Expected behavior
`ils` is angry because of bad file permissions and doesn't core dump.

### Observed Behavior (including steps to reproduce, if applicable)

```console
...
dev@19c85e3a97fb:/workspaces/irods$ sudo ls -dla ~/.irods
drwx------ 2 root root 4096 Apr 17 18:00 /home/dev/.irods
dev@19c85e3a97fb:/workspaces/irods$ sudo ls -la ~/.irods
total 20
drwx------ 2 root root 4096 Apr 17 18:00 .
drwxr-x--- 1 dev  dev  4096 Apr 17 18:00 ..
-rw-r--r-- 1 root root   12 Apr 17 18:00 .irodsA
-rw------- 1 root root  875 Apr 17 18:00 irods_environment.json
dev@19c85e3a97fb:/workspaces/irods$ ils
terminate called after throwing an instance of 'boost::filesystem::filesystem_error'
  what():  boost::filesystem::status: Permission denied [system:13]: "/home/dev/.irods/irods_environment.json"
Aborted (core dumped)
dev@19c85e3a97fb:/workspaces/irods$ echo $?
134
```

Issue seems to be the exception thrown from `boost::filesystem::exists(...)`:
https://github.com/irods/irods/blob/6e8d5637059a085a1e561178f18170ffdac8be72/lib/core/src/irods_environment_properties.cpp#L89-L92

MartinFlores751:
Related #4008

korydraughn:
@MartinFlores751 Update the milestone and close if complete.

PR #9033, opened 2026-07-24
[#8933] Prevent `ils` core dump on directory permission issues (main)

MartinFlores751:
Issue quick link: #8933

This PR fixes a core dump by updating the `boost::filesystem::exists(...)` function call to use a non-throwing version.

### DIFF
--- a/lib/core/src/irods_environment_properties.cpp
+++ b/lib/core/src/irods_environment_properties.cpp
@@ -13,7 +13,7 @@
 #include <boost/algorithm/string.hpp>
 #include <boost/lexical_cast.hpp>
 #include <boost/filesystem.hpp>
-
+#include <boost/system/detail/error_code.hpp>
 
 namespace irods {
     error get_json_environment_file(
@@ -86,8 +86,12 @@ namespace irods {
             return;
         }
 
-        if (!boost::filesystem::exists(json_file)) {
-            rodsLog(LOG_ERROR, "environment_properties::capture: missing environment file. should be at [%s]", json_file.c_str());
+        boost::system::error_code ec{};
+        if (!boost::filesystem::exists(json_file, ec)) {
+            rodsLog(LOG_ERROR,
+                    "environment_properties::capture: %s. Expected file at [%s].",
+                    ec.message().c_str(),
+                    json_file.c_str());
             return;
         }
 
--- a/scripts/irods/test/test_ils.py
+++ b/scripts/irods/test/test_ils.py
@@ -1,5 +1,7 @@
 import copy
 import os
+import pathlib
+import stat
 import unittest
 
 from . import session
@@ -32,6 +34,18 @@ class Test_Ils(resource_suite.ResourceBase, unittest.TestCase):
         self.admin.assert_icommand(['ils', '-l', rods_filename_1, rods_filename_2], 'STDOUT_MULTILINE', filename_1, filename_2)
         os.system('rm -f ' + filename_1)
         os.system('rm -f ' + filename_2)
+
+    def test_ils_does_not_core_dump_on_bad_file_permissions__issue_8933(self):
+        # Save original permissions
+        parent_dir = str(pathlib.Path(self.user0._environment_file_path).parent)
+
+        dir_stats = os.stat(parent_dir)
+        try:
+            os.chmod(parent_dir, 0o007)
+            self.user0.assert_icommand(['ils'], 'STDERR_SINGLELINE', 'environment_properties::capture: Permission denied.')
+        finally:
+            # Restore file permissions
+            os.chmod(parent_dir, stat.S_IMODE(dir_stats.st_mode))
 
     def test_genquery_fallback_for_collection_permissions__issue_4636(self):
         # Remove the specific query. This will cause the fallback to be used.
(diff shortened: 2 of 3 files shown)


MartinFlores751:
Ran the core and unit tests. Only one test fails, `test_iinit_does_not_crash_from_long_home_directory__issue_5411`, from the updated error string not matching the expected one.

lib/core/src/irods_environment_properties.cpp
```diff
@@ -86,7 +87,11 @@ namespace irods {
             return;
         }
 
-        if (!boost::filesystem::exists(json_file)) {
+        boost::system::error_code ec{};
+        if (!boost::filesystem::exists(json_file, ec)) {
+            if (ec) {
+                rodsLog(LOG_ERROR, "Boost filesystem error: [%s]", ec.message().c_str());
```
MartinFlores751: An example of the current error message is as follows:
```
ERROR: Boost filesystem error: [No such file or directory]
ERROR: environment_properties::capture: missing environment file. should be at [/tmp/tmp2_g88422/crazy/long/temp/home/directory/like/this/is/longer/than/the/solos/in/chameleon/.irods/irods_environment.json]
...
```

Did we want the error message to be reported differently?

korydraughn: Let's drop Boost from the error message. Let's drop the square brackets too.

Is the Boost error about the file path in the environment_properties message?

lib/core/src/irods_environment_properties.cpp
```diff
@@ -1,4 +1,5 @@
 #include "irods/irods_environment_properties.hpp"
+#include "boost/system/detail/error_code.hpp"
```
korydraughn: Group with other boost header includes.

lib/core/src/irods_environment_properties.cpp
korydraughn: Hmm, should we combine the two error messages?

MartinFlores751: > Hmm, should we combine the two error messages?

I think that'd be fine. Something like: `environment_properties::capture: [boost message]. Expected file at [file path].`

We can drop the brackets around the boost message here too.

lib/core/src/irods_environment_properties.cpp
MartinFlores751: Done.

lib/core/src/irods_environment_properties.cpp
korydraughn: That seems good to me. Let's see what it looks like for real.

MartinFlores751: Updated the error message. It now looks like: `ERROR: environment_properties::capture: No such file or directory. Expected file at [/tmp/tmpiwilghk0/crazy/long/temp/home/directory/like/this/is/longer/than/the/solos/in/chameleon/.irods/irods_environment.json].`

korydraughn: I like it.

korydraughn:
Squash it.

korydraughn:
Does the test need to be updated?

MartinFlores751:
> Does the test need to be updated?

Yep! Updated the test.

scripts/irods/test/test_ils.py
```diff
@@ -33,6 +35,18 @@ def test_ils_of_multiple_data_objects__issue_3520(self):
...
         os.system('rm -f ' + filename_2)
 
+    def test_ils_does_not_core_dump_on_bad_file_permissions__issue_8933(self):
+        # Save original permissions
+        parent_dir = str(pathlib.Path(self.user0._environment_file_path).parent)
+
+        dir_stats = os.stat(parent_dir)
+        try:
+            os.chmod(parent_dir, 0o007)
+            self.user0.assert_icommand(['ils'], 'STDERR_SINGLELINE', '[Permission denied]')
```
korydraughn: Does this assertion still pass?

MartinFlores751: It did not. Updated the test and now it passes.

scripts/irods/test/test_ils.py
```diff
@@ -33,6 +35,18 @@ def test_ils_of_multiple_data_objects__issue_3520(self):
...
         os.system('rm -f ' + filename_2)
 
+    def test_ils_does_not_core_dump_on_bad_file_permissions__issue_8933(self):
+        # Save original permissions
+        parent_dir = str(pathlib.Path(self.user0._environment_file_path).parent)
+
+        dir_stats = os.stat(parent_dir)
+        try:
+            os.chmod(parent_dir, 0o007)
+            self.user0.assert_icommand(['ils'], 'STDERR_SINGLELINE', 'Permission denied')
```
korydraughn: To strengthen the assertion, please include more context around `Permission denied`. For example:
```
'environment_properties::capture: Permission denied.'
```

korydraughn (commented):
Please run that one test again after making changes. If it passes, squash it.

scripts/irods/test/test_ils.py
MartinFlores751: Updated the assertion string.

MartinFlores751:
Squashed.

korydraughn (approved):
Pound it.

MartinFlores751:
Pounded.

PR #9035, opened 2026-07-29
[9034] Fix bug in logical quotas calculation (main)

FifthPotato:
Had worked on this for so long, but didn't realize this was an issue until now...oops.
`test_logical_quotas` passes on all our tested DBMSes.
Furthermore, the test has been demonstrated to capture the issue: it fails without the `cpp` changes, and passes with.

Since this system is so new and basically doesn't interact with anything, this is a pretty good indicator that these changes are OK. I can, however, run the full suite just to make sure nothing was overlooked.

Taking comments.

### DIFF
--- a/scripts/irods/test/test_logical_quotas.py
+++ b/scripts/irods/test/test_logical_quotas.py
@@ -4207,3 +4207,86 @@ class Test_Logical_Quotas(
             )
 
             self.admin.assert_icommand(["iadmin", "rmzone", zone_name])
+
+    def test_logical_quotas_do_not_count_similarly_prefixed_collections__issue_9034(self):
+        file_name = "test_logical_quota_similar_name"
+        file_path = os.path.join(self.quota_user.local_session_dir, file_name)
+        subcoll_path = f"{self.quota_user.session_collection}/subcoll"
+        uncounted_subcoll_path = f"{self.quota_user.session_collection}/subcollsimilar"
+        file_size = 4096
+        max_bytes = 10000
+        lib.make_file(
+            file_path,
+            file_size,
+            contents="arbitrary",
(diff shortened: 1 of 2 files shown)


scripts/irods/test/test_logical_quotas.py
```diff
@@ -4207,3 +4207,86 @@ def test_logical_quotas_set_on_remote_zones(self):
...
+                        "<unset>",
+                        str(-max_bytes),
+                        "<unenforced>",
+                    )
+                )
+                in out
+            )
+
+        finally:
+            self.admin.assert_icommand(
```
trel: don't assert in the finally

trel (commented):
please update commit message to say what you fixed rather than just 'bug'

korydraughn (commented):
No need to rerun the whole suite for this.

Will approve once review comments are resolved.

scripts/irods/test/test_logical_quotas.py
FifthPotato: Changed.

FifthPotato:
> please update commit message to say what you fixed rather than just 'bug'

Added some words. How does it look?

korydraughn:
Commit message looks good. I see the test has been updated to use `run_icommand` too.

Do all logical quotas tests pass?

FifthPotato:
> Commit message looks good. I see the test has been updated to use `run_icommand` too.
> 
> Do all logical quotas tests pass?

Just re-ran them again. They pass for all DBMSes we test against.

korydraughn (approved):
Excellent.

Pound it.

FifthPotato:
Octothorpe'd.

Issue #8997, opened 2026-06-03
Consider logging stacktrace dumps from agent factory


SwooshyCueb:
At present, stacktrace dumps are logged by the main server process: https://github.com/irods/irods/blob/677792c11d9819d941d4dabc5e81fcddcf2d7147/server/main_server/src/main.cpp#L382
Implementation: https://github.com/irods/irods/blob/677792c11d9819d941d4dabc5e81fcddcf2d7147/server/main_server/src/main.cpp#L1486-L1582
This results in low-quality stacktrace dump logging when the stacktrace dump is from an agent, particularly when a plugin or plugin dependency is part of the trace, as these libraries are not loaded at the time that the stacktrace dump file is loaded.

Now that we load plugins in the long-running agent factory in 5.0, we can improve the quality of our stacktrace dump logging by logging stacktrace dumps from agents in the agent factory rather than in the main server process.

I've been playing around with a quick-n-dirty version of this in my [`st-in-af.m`](https://github.com/irods/irods/compare/main...SwooshyCueb:irods:st-in-af.m) branch. Other things might pop up in this branch as I work on other issues, so as of this writing, the relevant commit is da218dc86f06a0af6f28701e6975e3c291ea844f

PR #9045, opened 2026-08-03
[#8997] Move processing of stacktrace files to agent factory (main)

korydraughn:
Running catch2 unit tests and core tests.

### DIFF
--- a/scripts/irods/test/test_stacktrace.py
+++ b/scripts/irods/test/test_stacktrace.py
@@ -1,9 +1,5 @@
 import sys
-
-if sys.version_info < (2, 7):
-    import unittest2 as unittest
-else:
-    import unittest
+import unittest
 
 from . import session
 from .. import test
@@ -35,8 +31,7 @@ class Test_Stacktrace(session.make_sessions_mixin([('otherrods', 'rods')], []), 
                 log_offset = lib.get_file_size_by_path(config.server_log_path)
                 rep_name = 'irods_rule_engine_plugin-irods_rule_language-instance'
                 self.admin.assert_icommand_fail(['irule', '-r', rep_name, 'msiSegFault()', 'null', 'ruleExecOut'])
-                for msg in [' 0# 0x', '"stacktrace_agent_pid":']:
-                    lib.delayAssert(lambda: lib.log_message_occurrences_greater_than_count(msg=msg, count=0, start_index=log_offset))
+                lib.delayAssert(lambda: lib.log_message_occurrences_greater_than_count(msg='"stacktrace_agent_pid":', count=0, start_index=log_offset))
 
         finally:
             IrodsController(config).reload_configuration()
--- a/server/main_server/src/agent_main.cpp
+++ b/server/main_server/src/agent_main.cpp
@@ -332,6 +332,8 @@ auto main(int _argc, char* _argv[]) -> int
                 break;
             }
 
+            log_stacktrace_files();
+
             if (g_terminate_graceful) {
                 log_af::info("{}: Received graceful shutdown instruction. Exiting agent factory main loop.", __func__);
                 break;
@@ -351,8 +353,6 @@ auto main(int _argc, char* _argv[]) -> int
                     log_af::info("{}: Received shutdown instruction. Exiting agent factory select() loop.", __func__);
                     break;
                 }
-
-                log_stacktrace_files();
 
                 if (g_terminate_graceful) {
                     log_af::info(
--- a/server/main_server/src/agent_main.cpp
+++ b/server/main_server/src/agent_main.cpp
@@ -326,6 +326,8 @@ auto main(int _argc, char* _argv[]) -> int
         fd_set sockMask;
         FD_ZERO(&sockMask); // NOLINT(readability-isolate-declaration)
 
+        struct timeval time_out; // NOLINT(cppcoreguidelines-pro-type-member-init)
+
         while (true) {
             if (g_terminate) {
                 log_af::info("{}: Received shutdown instruction. Exiting agent factory main loop.", __func__);
@@ -342,13 +344,14 @@ auto main(int _argc, char* _argv[]) -> int
             // NOLINTNEXTLINE(hicpp-signed-bitwise, cppcoreguidelines-pro-bounds-constant-array-index)
             FD_SET(svrComm.sock, &sockMask);
 
-            int numSock = 0;
-            struct timeval time_out; // NOLINT(cppcoreguidelines-pro-type-member-init)
+            // Always set the timeval structure (if used) before invoking select(). This is required because
+            // select() modifies the structure on each invocation.
+            constexpr auto time_out_microseconds = 500 * 1000;
             time_out.tv_sec = 0;
-            time_out.tv_usec = 500 * 1000;
-            const auto original_time_out = time_out;
+            time_out.tv_usec = time_out_microseconds;
 
-            while ((numSock = select(svrComm.sock + 1, &sockMask, nullptr, nullptr, &time_out)) == -1) {
+            int socket_ready_count = 0;
+            while ((socket_ready_count = select(svrComm.sock + 1, &sockMask, nullptr, nullptr, &time_out)) == -1) {
                 if (g_terminate) {
                     log_af::info("{}: Received shutdown instruction. Exiting agent factory select() loop.", __func__);
                     break;
@@ -360,9 +363,9 @@ auto main(int _argc, char* _argv[]) -> int
                     break;
                 }
 
-                // "select" modifies the timeval structure, so reset it.
+                // select() modifies the timeval structure, so reset it.
                 time_out.tv_sec = 0;
-                time_out.tv_usec = 500 * 1000;
+                time_out.tv_usec = time_out_microseconds;
 
                 if (EINTR == errno) {
                     log_af::trace("{}: select() interrupted", __func__);
@@ -385,9 +388,7 @@ auto main(int _argc, char* _argv[]) -> int
                 break;
             }
 
-            if (0 == numSock) {
-                // "select" modifies the timeval structure, so reset it.
-                time_out = original_time_out;
+            if (0 == socket_ready_count) {
                 continue;
             }
 
--- a/server/main_server/src/agent_main.cpp
+++ b/server/main_server/src/agent_main.cpp
@@ -1012,7 +1012,7 @@ namespace
                 continue;
             }
 
-            auto slash_pos = p.rfind("/");
+            auto slash_pos = p.rfind('/');
 
             if (slash_pos == std::string::npos) {
                 log_af::trace("{}: Skipping [{}]. No forward slash separator found.", __func__, p);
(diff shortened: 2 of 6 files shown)


korydraughn:
Unit tests pass.

```
-----
All tests passed! :)
time elapsed: [ 185.0037]seconds ([   0]hours [ 3.0834]minutes)

2026-08-03 15:47:14,549 - version:[5.0.90]
2026-08-03 15:47:14,549 - sha:[b300c3fb633391c29aab991815734ecf19e2ec70]
2026-08-03 15:47:14,549 - message:[]
```

korydraughn:
Running core tests.

scripts/irods/test/test_stacktrace.py
```diff
@@ -35,8 +31,7 @@ def test_stacktraces_appear_in_log__issue_4382(self):
                 log_offset = lib.get_file_size_by_path(config.server_log_path)
                 rep_name = 'irods_rule_engine_plugin-irods_rule_language-instance'
                 self.admin.assert_icommand_fail(['irule', '-r', rep_name, 'msiSegFault()', 'null', 'ruleExecOut'])
-                for msg in [' 0# 0x', '"stacktrace_agent_pid":']:
-                    lib.delayAssert(lambda: lib.log_message_occurrences_greater_than_count(msg=msg, count=0, start_index=log_offset))
+                lib.delayAssert(lambda: lib.log_message_occurrences_greater_than_count(msg='"stacktrace_agent_pid":', count=0, start_index=log_offset))
```
MartinFlores751, line 34: Why was the `0# 0x` check removed from the stack trace messages?

server/main_server/src/agent_main.cpp
```diff
@@ -338,10 +345,11 @@ auto main(int _argc, char* _argv[]) -> int
             FD_SET(svrComm.sock, &sockMask);
 
             int numSock = 0;
-            struct timeval time_out; // NOLINT(cppcoreguidelines-pro-type-member-init)
+
+            // "select" modifies the timeval structure, so reset it.
```
MartinFlores751: I don't think the `"select"` comment is relevant to this particular portion of the code... It might be better to have the comment explain `time_out_in_microseconds`. Some examples could be to describe how we determined `500 * 1000` is a good value or to state that `time_out_in_microseconds` is also used to reset the struct later on.

server/main_server/src/agent_main.cpp
```diff
@@ -323,12 +326,16 @@ auto main(int _argc, char* _argv[]) -> int
         fd_set sockMask;
         FD_ZERO(&sockMask); // NOLINT(readability-isolate-declaration)
 
+        struct timeval time_out; // NOLINT(cppcoreguidelines-pro-type-member-init)
```
MartinFlores751, line 329: Is it worth considering braced initialization so members are zeroed prior to first use?

korydraughn:
All core tests pass.

Will address clang-tidy report next.

server/main_server/src/agent_main.cpp
```diff
@@ -980,6 +986,105 @@ namespace
...
+                          pid);
+
+            try {
+                // Convert the epoch value to ISO8601 format.
+                log_af::trace("{}: Converting epoch seconds to UTC timestamp.", __func__);
+                using boost::chrono::system_clock;
+                using boost::chrono::time_fmt;
+                const auto tp = system_clock::from_time_t(std::stoll(epoch_seconds));
+                std::ostringstream utc_ss;
+                utc_ss << time_fmt(boost::chrono::timezone::utc, "%FT%T") << tp;
```
MartinFlores751, line 1054: Do we prefer boost chrono to std chrono? If std chrono is used, I think we can take advantage of fmt lib's time formatting options.

server/main_server/src/agent_main.cpp
```diff
@@ -980,6 +986,105 @@ namespace
...
+        const auto sleep_time =
+            irods::get_advanced_setting<int>(irods::KW_CFG_STACKTRACE_FILE_PROCESSOR_SLEEP_TIME_IN_SECONDS);
+        static auto start_time = std::chrono::steady_clock::now();
+        const auto now = std::chrono::steady_clock::now();
+
+        if (now - start_time < std::chrono::seconds{sleep_time}) {
+            return;
+        }
+
+        start_time = now;
```
MartinFlores751: Minor detail. `start_time` can be renamed to `last_log_time` or something more specific.

server/main_server/src/agent_main.cpp
```diff
@@ -980,6 +986,105 @@ namespace
...
     } // handle_client_request
 
+    auto log_stacktrace_files() -> void
+    {
+        const auto sleep_time =
+            irods::get_advanced_setting<int>(irods::KW_CFG_STACKTRACE_FILE_PROCESSOR_SLEEP_TIME_IN_SECONDS);
+        static auto start_time = std::chrono::steady_clock::now();
+        const auto now = std::chrono::steady_clock::now();
+
+        if (now - start_time < std::chrono::seconds{sleep_time}) {
```
MartinFlores751: Optional. Can make `std::chrono::seconds{sleep_time}` a `constexpr` and give it a name.

server/main_server/src/agent_main.cpp
```diff
@@ -980,6 +986,105 @@ namespace
...
+
+            ++slash_pos;
+            const auto first_dot_pos = p.find(".", slash_pos);
+
+            if (first_dot_pos == std::string::npos) {
+                log_af::trace("{}: Skipping [{}]. No dot separator found.", __func__, p);
+                continue;
+            }
+
+            const auto last_dot_pos = p.rfind(".");
```
MartinFlores751: Is there a reason why we use `.rfind(...)` instead of using `.find(".", first_dot_pos + 1)` or an equivalent? I think `.find(...)` would follow the specified `<epoch_seconds>.<epoch_milliseconds>.<agent_pid>` format more exactly, where an `.rfind(...)` would allow some wiggle room.

MartinFlores751: Though to be fair, both allow for "wiggle room", it's just a difference in where that room is provided.

MartinFlores751 (commented):
Looks pretty good. Most comments are just about style. There are a few questions about the code too.

scripts/irods/test/test_stacktrace.py
korydraughn: It didn't add any value. `stacktrace_agent_pid` is a clear indicator for what's being tested.

server/main_server/src/agent_main.cpp
korydraughn: I'd prefer to stick to the man page example for this. If the example used `memset` to clear the struct, that would be signal to use braced initialization.

server/main_server/src/agent_main.cpp
korydraughn: > I don't think the `"select"` comment is relevant to this particular portion of the code.

What makes it feel irrelevant? Is it because it's above the `constexpr` variable or do you feel more words are needed?

scripts/irods/test/test_stacktrace.py
MartinFlores751: And every stack trace (`0# 0x`) caught and reported by the agent will have an associated `stacktrace_agent_pid`, right? If that's the case, this can be resolved.

server/main_server/src/agent_main.cpp
MartinFlores751: Further down in the code is the exact same comment with an actual `select(...)` call nearby that is then being "reset". Unless I'm missing something, the code highlighted in this comment is more of an initialization than a "reset after a select call".

From another perspective, when I first saw this comment, it was not obvious to me how there would be a "select" that would require a "reset" at this specific point.

server/main_server/src/agent_main.cpp
MartinFlores751: Sticking to examples is fine. Does the example in the man page use memset?

server/main_server/src/agent_main.cpp
korydraughn: This logic was introduced back in 4.3.0 (4+ years old). I don't think we had fmtlib at the time this was written.

I think adjusting it is out of scope for this PR though. Improving the performance can happen in a separate PR.

server/main_server/src/agent_main.cpp
korydraughn: I don't remember. I think it leans heavily on the filename following the naming convention.

True on the wiggle room comment.

server/main_server/src/agent_main.cpp
korydraughn: That results in a compiler error due to the use of `constexpr`.

scripts/irods/test/test_stacktrace.py
korydraughn: That's correct.

server/main_server/src/agent_main.cpp
korydraughn: It doesn't.

server/main_server/src/agent_main.cpp
korydraughn: It is both an initialization and reset. It might be hard to spot due to it being viewed as a diff.

The outer loop (where this comment is) is encountered when `select` returns a value greater than or equal to 0. Look for `if (0 == numSock) {` on L391.

MartinFlores751: Gotcha, I see it now. I think the dual-purpose code doing both init/reset with only a comment stating "reset" threw me off.

Maybe the comment could be something like `Set the timeout for "select"`, to avoid the reset/init wording? Think that would help avoid any misunderstandings. Thoughts?

server/main_server/src/agent_main.cpp
MartinFlores751: Gotcha. This comment can be resolved.

server/main_server/src/agent_main.cpp
MartinFlores751: Sounds good. Might want to create an issue for this so it's not forgotten.

server/main_server/src/agent_main.cpp
korydraughn: Will update.

server/main_server/src/agent_main.cpp
MartinFlores751: Doh. If you want to give the `std::chrono::seconds{sleep_time}` a name you can assign it to a `const` then (or something else that compiles). Otherwise, this issue can be resolved.

server/main_server/src/agent_main.cpp
MartinFlores751: Not a blocker. You can create an issue for this to look further into it later.

korydraughn: Will open an issue for this after it is merged.

server/main_server/src/agent_main.cpp
korydraughn: I'm going to leave this as is since it's a pretty clear what's happening.

server/main_server/src/agent_main.cpp
korydraughn: Done.

server/main_server/src/agent_main.cpp
korydraughn: Done.

server/main_server/src/agent_main.cpp
korydraughn: Will open an issue for this after it is merged.

korydraughn:
Reran unit and core tests.

All tests passed.

trel:
squash em

korydraughn:
Squashed.

trel:
10 commits is correct?

korydraughn:
Yes.

trel:
pound em if you're ready

korydraughn:
Added pounds.

korydraughn:
Merging.

trel (approved):
ship it

server/main_server/src/agent_main.cpp
korydraughn: Opened #9046.

server/main_server/src/agent_main.cpp
korydraughn: Opened #9047.

PR #9048, opened 2026-08-06
[9032] Replace collection UPDATE with LIKE clause

FifthPotato:
Passes unit and core tests across all DBMSes we test against (on el9).
I'm still kind of suspicious of this result, so this definitely warrants additional scrutiny.

When testing against a 10-deep nested subcollection with about 100,000 collections total, `imv`-ing the root collection takes about half the time.

### DIFF
--- a/plugins/database/src/db_plugin.cpp
+++ b/plugins/database/src/db_plugin.cpp
@@ -10760,7 +10760,9 @@ irods::error db_rename_object_op(
         cllBindVars[cllBindVarCount++] = parentCollName;
         cllBindVars[cllBindVarCount++] = slashNewName;
         cllBindVars[cllBindVarCount++] = collName;
-        cllBindVars[cllBindVarCount++] = collNameSlash;
+        cllBindVars[cllBindVarCount++] = parentCollName;
+        cllBindVars[cllBindVarCount++] = slashNewName;
+        cllBindVars[cllBindVarCount++] = collName;
         cllBindVars[cllBindVarCount++] = collNameSlash;
         cllBindVars[cllBindVarCount++] = collName;
         if ( logSQL != 0 ) {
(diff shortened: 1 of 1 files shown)


korydraughn:
Do you have the specific numbers from your testing for the 100k collections?

korydraughn:
Have you considered providing the schema for `R_COLL_MAIN`, the old SQL, and the new SQL to an AI to see what it finds?

It might spot something we haven't considered.

FifthPotato:
> Do you have the specific numbers from your testing for the 100k collections?

Old:
```
[irods@71ce61736ce6 ~]$ time imv dir_0 dir_00

real    0m9.319s
user    0m0.052s
sys     0m0.021s
[irods@71ce61736ce6 ~]$ time imv dir_00 dir_0

real    0m9.100s
user    0m0.053s
sys     0m0.017s
[irods@71ce61736ce6 ~]$ time imv dir_0 dir_00

real    0m10.531s
user    0m0.051s
sys     0m0.019s

[irods@71ce61736ce6 ~]$ time imv dir_00 dir_0

real    0m8.993s
user    0m0.056s
sys     0m0.017s
```

New:

```
[irods@1c61ac139e26 ~]$ time imv dir_0 dir_00

real    0m4.841s
user    0m0.053s
sys     0m0.019s
[irods@1c61ac139e26 ~]$ time imv dir_00 dir_0

real    0m5.036s
user    0m0.054s
sys     0m0.020s
[irods@1c61ac139e26 ~]$ time imv dir_0 dir_00

real    0m5.149s
user    0m0.056s
sys     0m0.016s
[irods@1c61ac139e26 ~]$ time imv dir_00 dir_0

real    0m5.027s
user    0m0.059s
sys     0m0.017s
```

plugins/database/src/db_plugin.cpp
```diff
@@ -10783,33 +10789,6 @@ irods::error db_rename_object_op(
...
 
-        /* like above, but for the parent_coll_name's */
-        cllBindVars[cllBindVarCount++] = parentCollName;
-        cllBindVars[cllBindVarCount++] = slashNewName;
-        cllBindVars[cllBindVarCount++] = collName;
-        cllBindVars[cllBindVarCount++] = collNameSlash;
-        cllBindVars[cllBindVarCount++] = collNameSlash;
-        cllBindVars[cllBindVarCount++] = collName;
-        if ( logSQL != 0 ) {
-            log_sql::debug("chlRenameObject SQL 10");
```
trel, line 10794: does anything look for this debug output?  this was the way some tests passed back in the day...

plugins/database/src/db_plugin.cpp
```diff
@@ -10760,21 +10760,27 @@ irods::error db_rename_object_op(
...
             log_sql::debug("chlRenameObject SQL 9");
         }
 #ifdef ORA_ICAT
         status = cmlExecuteNoAnswerSql(
-            "update R_COLL_MAIN set coll_name = SUBSTR(coll_name, 1, LENGTH(?)) || ? || SUBSTR(coll_name, LENGTH(?) + "
-            "1) where SUBSTR(parent_coll_name, 1, LENGTH(?)) = ? or parent_coll_name = ?",
+            "update R_COLL_MAIN set coll_name = SUBSTR(coll_name, 1, LENGTH(?)) || ? || SUBSTR(coll_name, "
+            "LENGTH(?) + 1), parent_coll_name = SUBSTR(parent_coll_name, 1, "
+            "LENGTH(?)) || ? || SUBSTR(parent_coll_name, LENGTH(?) + 1) "
+            "where parent_coll_name LIKE (? || '%') or parent_coll_name = ?",
```
trel, line 10776: some sleuthing suggests this `LIKE (? || '%')` construct still will not use an index on `parent_coll_name`, even if it exists.


> How to make it SARGable (practically)
> Bind the parameter as the full pattern including the % so the predicate shape is “clean”:
> 
> - `parent_coll_name LIKE ?` with `?` bound to `prefix%`
>
> In other words, do the concatenation in your application / caller, not inside SQL:
> 
> - application sets parameter = `'abc%'`
> - SQL uses `parent_coll_name LIKE ?`
> 



and then separately... i'm not convinced this is enough indexing... as is...

```
Indexes:
    "idx_coll_main2" UNIQUE, btree (parent_coll_name, coll_name)
    "idx_coll_main3" UNIQUE, btree (coll_name)
    "idx_coll_main1" btree (coll_id)
```
i think we might need another one JUST on `parent_coll_name`?

DrakoRod: Hey guys, reviewing the code, I've a couple of questions. 

I guess that the objective of using the next code `LIKE (? || '%')` in the query builder is to concatenate the `%` operator with the parameter received. The question is: If so, could it be used directly in the query builder without having to concatenate it into the query?

For example, just like that:

`"where parent_coll_name LIKE '?%' or parent_coll_name = ?",`

I don't know how the query builder works, but the problem is that when we use this type of syntax `LIKE (? || '%')`, the query planner recognize different that the simple `LIKE '?%'` and instead using the existing index, try to find a index that process a specific filter: `(? || '%')` because of that don't use the above indexes.

PR #9051, opened 2026-08-14
[9050] Fix overcounting bug with LIKE wildcards in coll names (main)

FifthPotato:
Seems to work as intended.

Ran the new test in a before/after style against one DBMS. As expected, it failed before and succeeded after.

Ran `test_logical_quotas` for all DBMSes we test against, and all pass.

Taking comments.

### DIFF
--- a/plugins/database/src/db_plugin.cpp
+++ b/plugins/database/src/db_plugin.cpp
@@ -16644,7 +16644,9 @@ irods::error db_calc_logical_usage_and_quota_op(irods::plugin_context& _ctx, [[m
                    "R_COLL_MAIN RCM1, "
                    "R_COLL_MAIN RCM2 "
               "WHERE R_DATA_MAIN.coll_id = RCM2.coll_id "
-                "AND (RCM2.coll_name LIKE CONCAT(RCM1.coll_name, ?, '%') "
+                // Escape all LIKE wildcards from the coll name by replacing them out.
+                // MySQL only: backslashes are escape characters by default, so they need to be doubled.
+                "AND (RCM2.coll_name LIKE CONCAT(REPLACE(REPLACE(REPLACE(RCM1.coll_name, '\\\\', '\\\\\\\\'), '%', '\\\\%'), '_', '\\\\_'), ?, '%') ESCAPE '\\\\' "
                      "OR RCM2.coll_name = RCM1.coll_name)) ranked_objs "
            // Only count one data object that is top-ranked.
            "WHERE ranked_objs.replica_rank = 1 "
@@ -16686,7 +16688,7 @@ irods::error db_calc_logical_usage_and_quota_op(irods::plugin_context& _ctx, [[m
                    "R_COLL_MAIN RCM2 "
               "WHERE R_DATA_MAIN.coll_id = RCM2.coll_id "
                 "AND RCM1.coll_id = R_LOGICAL_QUOTA_MAIN.coll_id "
-                "AND (RCM2.coll_name LIKE (RCM1.coll_name || ? || '%') "
+                "AND (RCM2.coll_name LIKE (REPLACE(REPLACE(REPLACE(RCM1.coll_name, '\\', '\\\\'), '%', '\\%'), '_', '\\_') || ? || '%') ESCAPE '\\' "
                      "OR RCM2.coll_name = RCM1.coll_name)) AS ranked_objs "
            "WHERE ranked_objs.replica_rank = 1)",
 #endif
--- a/scripts/irods/test/test_logical_quotas.py
+++ b/scripts/irods/test/test_logical_quotas.py
@@ -4290,3 +4290,171 @@ class Test_Logical_Quotas(
                     "0",
                 ]
             )
+
+
+    def test_logical_quotas_do_not_overcount_collections_with_like_wildcards__issue_9050(self):
+        file_name = "test_logical_quota_like_wildcards"
+        file_path = os.path.join(self.quota_user.local_session_dir, file_name)
+
+        # The failure that this test guards against is as follows:
+        # With a quota set on .../%potato, logical quotas will count data objects
+        # whose collection path matches .../%potato/%
+        # This means a data object like .../badpotato/obj will not be counted, but
+        # .../badpotato/inner/obj will be, since obj's enclosing collection name
+        # matches the pattern. (.../badpotato/inner matches .../%potato/%)
+        subcoll_path = f"{self.quota_user.session_collection}/%potato"
+        other_subcoll_path = f"{self.quota_user.session_collection}/_tomato"
+        uncounted_subcoll_path = f"{self.quota_user.session_collection}/badpotato"
+        other_uncounted_subcoll_path = f"{self.quota_user.session_collection}/btomato"
+
+        deep_subcoll = f"inner"
+
+        file_size = 4096
+        max_bytes = 10000
+        max_objects = 10000
+        lib.make_file(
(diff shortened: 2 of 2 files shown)


plugins/database/src/db_plugin.cpp
```diff
@@ -16644,7 +16644,7 @@ irods::error db_calc_logical_usage_and_quota_op(irods::plugin_context& _ctx, [[m
                    "R_COLL_MAIN RCM1, "
                    "R_COLL_MAIN RCM2 "
               "WHERE R_DATA_MAIN.coll_id = RCM2.coll_id "
-                "AND (RCM2.coll_name LIKE CONCAT(RCM1.coll_name, ?, '%') "
+                "AND (RCM2.coll_name LIKE CONCAT(REPLACE(REPLACE(REPLACE(RCM1.coll_name, '\\\\', '\\\\\\\\'), '%', '\\\\%'), '_', '\\\\_'), ?, '%') ESCAPE '\\\\' "
```
korydraughn, line 16649: Add a comment to help the next person understand the what and the why.

plugins/database/src/db_plugin.cpp
```diff
@@ -16686,7 +16686,7 @@ irods::error db_calc_logical_usage_and_quota_op(irods::plugin_context& _ctx, [[m
                    "R_COLL_MAIN RCM2 "
               "WHERE R_DATA_MAIN.coll_id = RCM2.coll_id "
                 "AND RCM1.coll_id = R_LOGICAL_QUOTA_MAIN.coll_id "
-                "AND (RCM2.coll_name LIKE (RCM1.coll_name || ? || '%') "
+                "AND (RCM2.coll_name LIKE (REPLACE(REPLACE(REPLACE(RCM1.coll_name, '\\', '\\\\'), '%', '\\%'), '_', '\\_') || ? || '%') ESCAPE '\\' "
```
korydraughn, line 16691: Add a comment to help the next person understand the what and the why.

scripts/irods/test/test_logical_quotas.py
```diff
@@ -4290,3 +4290,170 @@ def test_logical_quotas_do_not_count_similarly_prefixed_collections__issue_9034(
...
+    def test_logical_quotas_do_not_overcount_collections_with_like_wildcards__issue_9050(self):
+        file_name = "test_logical_quota_like_wildcards"
+        file_path = os.path.join(self.quota_user.local_session_dir, file_name)
+
+        # The failure being tested for is as follows:
+        # With a quota set on .../%potato, logical quotas will count data objects
+        # whose collection path matches .../%potato/%
+        # This means a data object like .../badpotato/obj will not be counted, but
+        # .../badpotato/inner/obj will be, since obj's enclosing collection name
+        # matches the pattern. (.../badpotato/inner matches .../%potato/%)
```
korydraughn: This only explains the failure case, correct?

scripts/irods/test/test_logical_quotas.py
```diff
@@ -4290,3 +4290,170 @@ def test_logical_quotas_do_not_count_similarly_prefixed_collections__issue_9034(
...
+
+            self.admin.assert_icommand(
+                [
+                    "iadmin",
+                    "set_logical_quota",
+                    subcoll_path,
+                    "bytes",
+                    str(max_bytes),
+                ]
+            )
```
korydraughn, line 4337: I see byte count is tested. Do we need an equivalent test for number of data objects?

scripts/irods/test/test_logical_quotas.py
FifthPotato: It's meant to highlight what the test is trying to guard against.

scripts/irods/test/test_logical_quotas.py
FifthPotato: I could just change the quota to set both bytes and data objects and change the `llq` asserts to look for object count (being unchanged) as well. I think that'd cover the extremely unlikely edge case. Thoughts?

plugins/database/src/db_plugin.cpp
FifthPotato: Added some words.

plugins/database/src/db_plugin.cpp
FifthPotato: Didn't add words here, but the comments for the MySQL query apply here as well.

scripts/irods/test/test_logical_quotas.py
korydraughn: I felt that's what it was communicating. Consider rephrasing L4299 so that is communicated clearly.

scripts/irods/test/test_logical_quotas.py
korydraughn: That seems fine. The CLI is hitting the same code paths regardless, correct?

The main goal is to cover both cases.

FifthPotato: > That seems fine. The CLI is hitting the same code paths regardless, correct?
> 
> The main goal is to cover both cases.

In essence, yes. Since this code is in the recalculation step, only bugs relating to recalculation could pop up. So there is a tiny chance that it calculates object count wrong and that would be something you'd need to guard against.

I'll make the change.

korydraughn: Can you say more about this? It hasn't clicked for me how that can happen.

> So there is a tiny chance that it calculates object count wrong and that would be something you'd need to guard against.

How do we guard against it? Or are you saying that is something the admin has to make a decision about? Are there any tips we need to put in documentation (docs.irods.org) to help admins?

scripts/irods/test/test_logical_quotas.py
FifthPotato: Done.

scripts/irods/test/test_logical_quotas.py
FifthPotato: If my changes to the recalculation query somehow messed up the math for object calculation (I'm pretty sure it didn't) then a test that checks the math for both byte and object calculation would catch that (as opposed to a test that only checks the byte calculations).
That being said, since I'm only modifying the WHERE clause here, that should never happen. However, more paranoid tests usually aren't a bad thing.

I personally wouldn't be able to think of how this would mess up object calculation, if that's what you're worried about.

FifthPotato:
After most recent changes, re-ran `test_logical_quotas` on all tested DBMSes. Still passes.

scripts/irods/test/test_logical_quotas.py
korydraughn: > If my changes to the recalculation query somehow messed up the math for object calculation (I'm pretty sure it didn't) then a test that checks the math for both byte and object calculation would catch that (as opposed to a test that only checks the byte calculations).

Good. That's what we want the test to do.

> That being said, since I'm only modifying the WHERE clause here, that should never happen. However, more paranoid tests usually aren't a bad thing.

If you're confident the test covers things, then there's no need for additional tests.

> I personally wouldn't be able to think of how this would mess up object calculation, if that's what you're worried about.

Sounds like you've thought it through.

Mark this conversation as resolved if you feel there's nothing else to add.

FifthPotato:
Is there anything left to review, or am I free to squash?

korydraughn (commented):
Squash it.

FifthPotato:
Squished.

korydraughn (approved):
Pound it.

FifthPotato:
Octothorpe'd.

Issue #9059, opened 2026-08-26
`get_new_connection` can ignore requested client username in delay server


korydraughn:
Affects all versions of iRODS 4.3 and later.

The following code is used by the delay server to process delay rules. If the function is handed a nonempty `_client_user` and the `delay_rule_executors` list is empty, the function ignores the client username and falls back to connecting as the local rodsadmin.

This doesn't have any negative impact on the processing of delay rules because user identity is defined by information in the REI context. However, this bug does make it more difficult to write policy that depends on the name of the connected user. Fixing this also allows the log file to present the correct username.

https://github.com/irods/irods/blob/b669dda15abc4a1b21412d258fccc61331d6a46b/server/delay_server/src/irodsDelayServer.cpp#L187-L214

## Workaround

Add the hostname/FQDN of one or more iRODS servers to the `delay_rule_executors` list in server_config.json.

PR #9060, opened 2026-08-26
[#9059] Delay Server: Honor client username when creating connection (main)

korydraughn:
Unit tests and core tests pass.


scripts/irods/test/test_all_rules.py
```diff
@@ -1943,6 +1943,50 @@ def test_irule_cannot_modify_the_random_scheme_via_acSetVaultPathPolicy__issue_8
...
+                    delay_rule = f'delay("<INST_NAME>{rep_name}</INST_NAME>") {{ writeLine("serverLog", "issue #9059"); }}'
+                    self.user0.assert_icommand(['irule', '-r', rep_name, delay_rule, 'null', 'ruleExecOut'])
+
+                    # Show the user's name and zone are attached as metadata to their session
+                    # collection. The metadata will only exist if the PEP defined in core.re fires.
+                    def metadata_attr_with_value_exists_on_collection(attr, value):
+                        print(f'Looking for attr:[{attr}] value:[{value}] on collection:[{self.user0.session_collection}]')
+                        out, _, _ = self.user0.run_icommand(
+                            ['iquest', '%s', f"select COLL_NAME where META_COLL_ATTR_NAME = '{attr}' and META_COLL_ATTR_VALUE = '{value}'"])
+                        print(out)
```
trel, line 1980: print statement?

trel:
logic looks good to me

scripts/irods/test/test_all_rules.py
```diff
@@ -1943,6 +1943,50 @@ def test_irule_cannot_modify_the_random_scheme_via_acSetVaultPathPolicy__issue_8
...
+        server_config_filename = paths.server_config_path()
+        with open(server_config_filename) as f:
+            svr_cfg = json.load(f)
+        svr_cfg['advanced_settings']['delay_server_sleep_time_in_seconds'] = 2
+
+        try:
+            # Update server_config.json.
+            new_server_config = json.dumps(svr_cfg, sort_keys=True, indent=4, separators=(',', ': '))
+            with lib.file_backed_up(server_config_filename):
+                with open(server_config_filename, 'w') as f:
```
FifthPotato: Do we like the double open here (paired with the one on 1949), or do we prefer a `open -> read -> seek -> write -> truncate` deal instead? Either way is probably fine, but it's something to consider.

FifthPotato (commented):
Logic looks good! Just one nitpick python comment to look at that you're free to disregard.

Also, a small point of confusion on the commit message: `Before this commit, the delay server would always connect as the local rodsadmin.` This wording seems slightly misleading because it relies on the end of the previous sentence,  `...when the executors list is empty` to make it true. On first read, I had thought that it meant it would always *always* connect as the local rodsadmin.
Maybe a slight change of wording on the last sentence to make it clearer, like `Before this commit, the executors list being empty would cause the delay server to always connect as the local rodsadmin.` or similar?

trel:
agree with the always always comment and fix above.

scripts/irods/test/test_all_rules.py
korydraughn: That function is modeled after ones provided by lib.py.

The print statements allow us to see what is happening for situations involving `run_icommand`. `run_icommand` doesn't write its output to stdout/stderr like `assert_icommand`.

trel: okay, if intentional, no worries.   just was unexpected.   resolving.

scripts/irods/test/test_all_rules.py
korydraughn: Updated.

korydraughn:
Test is still passing.

korydraughn:
Updated the commit message for b14df71.

FifthPotato (approved):
Commit message and python change look good to me!

korydraughn:
Merging.

PR #9064, opened 2026-08-30
5423.m ibun double free (main)

trel:
The double free mentioned in the original issue (4-2-stable) is no longer reproducible - it was checked on 4.3.5 and 5.0.2.

However, inspecting the code resulted in detecting that bundling a mounted collection into a replication resource was still a problem.  Could make a new issue if we think it correct.

This passes relevant tests - has not passed a full test run.

### DIFF
--- a/scripts/irods/test/test_ibun.py
+++ b/scripts/irods/test/test_ibun.py
@@ -237,6 +237,67 @@ class Test_ibun(resource_suite.ResourceBase, unittest.TestCase):
             if os.path.exists(local_file):
                 os.unlink(local_file)
 
+    def test_ibun_reports_resource_mismatch_for_mounted_collection__issue_5423(self):
+        replication = 'issue_5423_repl'
+        resource_0 = 'issue_5423_ufs0'
+        resource_1 = 'issue_5423_ufs1'
+        mount_resource = 'issue_5423_mount_resc'
+        collection_path = os.path.join(self.admin.session_collection, 'issue_5423_src')
+        mounted_collection = os.path.join(collection_path, 'mnt')
+        archive_path = collection_path + '.zip'
+        mountpoint_directory = os.path.join(self.admin.local_session_dir, 'issue_5423_phymount')
+        rule_file = os.path.join(self.admin.local_session_dir, 'issue_5423_register_mount.r')
+
+        rule_text = '''\
+register_issue_5423_mount {{
+    msiCollCreate(*irodsColl, 0, *status);
+    msiPhyPathReg(*irodsColl, *irodsResc, *phyDir, "mountPoint", *status);
+}}
+
+INPUT *irodsColl="{mounted_collection}", *irodsResc="{mount_resource}", *phyDir="{mountpoint_directory}"
+OUTPUT ruleExecOut
+'''.format(**locals())
+
+        try:
+            lib.create_replication_resource(self.admin, replication)
+            lib.create_ufs_resource(self.admin, resource_0, test.settings.HOSTNAME_2)
+            lib.create_ufs_resource(self.admin, resource_1, test.settings.HOSTNAME_3)
+            lib.create_ufs_resource(self.admin, mount_resource, test.settings.HOSTNAME_1)
+            lib.add_child_resource(self.admin, replication, resource_0)
+            lib.add_child_resource(self.admin, replication, resource_1)
+
+            os.mkdir(mountpoint_directory)
+            with open(os.path.join(mountpoint_directory, 'foo'), 'w') as f:
+                f.write('payload')
+
+            with open(rule_file, 'w') as f:
+                f.write(rule_text)
+
+            self.admin.assert_icommand(['imkdir', collection_path])
+            self.admin.assert_icommand(['irule', '-F', rule_file])
+            self.admin.assert_icommand(['ils', '-AL', mounted_collection], 'STDOUT_SINGLELINE', 'mountP')
+
+            out, err, ec = self.admin.run_icommand(
+                ['ibun', '-c', '-Dzip', '-R', replication, archive_path, collection_path])
+            self.assertNotEqual(0, ec)
+            self.assertEqual('', out)
+            self.assertIn('SYS_COPY_NOT_EXIST_IN_RESC', err)
+            self.assertNotIn('SYS_INTERNAL_ERR', err)
+
+        finally:
+            self.admin.run_icommand(['imcoll', '-U', mounted_collection])
+            self.admin.run_icommand(['irm', '-f', archive_path])
+            self.admin.run_icommand(['irm', '-r', '-f', collection_path])
+            self.admin.run_icommand(['iadmin', 'rmchildfromresc', replication, resource_0])
+            self.admin.run_icommand(['iadmin', 'rmchildfromresc', replication, resource_1])
+            self.admin.run_icommand(['iadmin', 'rmresc', replication])
+            self.admin.run_icommand(['iadmin', 'rmresc', resource_0])
+            self.admin.run_icommand(['iadmin', 'rmresc', resource_1])
+            self.admin.run_icommand(['iadmin', 'rmresc', mount_resource])
+            shutil.rmtree(mountpoint_directory, ignore_errors=True)
+            if os.path.exists(rule_file):
+                os.unlink(rule_file)
+
     def test_ibun(self):
         test_file = "ibun_test_file"
         lib.make_file(test_file, 1000)
--- a/scripts/irods/test/test_ibun.py
+++ b/scripts/irods/test/test_ibun.py
@@ -183,6 +183,60 @@ class Test_ibun(resource_suite.ResourceBase, unittest.TestCase):
         do_test_ibun_atop_existing_archive_file_in_replication_hierarchy(self, 1038425, 1044480)
         do_test_ibun_atop_existing_archive_file_in_replication_hierarchy(self, 1040425, 1054720)
 
+    def test_ibun_zip_to_replication_resource_with_archive_name_matching_source_collection__issue_5423(self):
+        replication = 'issue_5423_repl'
+        resource_0 = 'issue_5423_ufs0'
+        resource_1 = 'issue_5423_ufs1'
+        collection_name = 'new'
+        filename = 'foo'
+        local_file = os.path.join(self.user0.local_session_dir, filename)
+        collection_path = os.path.join(self.user0.session_collection, collection_name)
+        put_path = os.path.join(collection_path, filename)
+        archive_path = os.path.join(self.user0.session_collection, collection_name + '.zip')
+
+        def assert_archive_replicas_are_good():
+            query = "select DATA_RESC_NAME, DATA_REPL_STATUS where COLL_NAME = '{}' and DATA_NAME = '{}'".format(
+                os.path.dirname(archive_path), os.path.basename(archive_path))
+            out, err, ec = self.admin.run_icommand(['iquest', '%s...%s', query])
+
+            self.assertEqual(0, ec)
+            self.assertEqual(0, len(err))
+
+            rows = [line.split('...') for line in out.splitlines()]
+            self.assertEqual(2, len(rows), out)
+            self.assertEqual(set([resource_0, resource_1]), set([row[0] for row in rows]))
+
+            for _, repl_status in rows:
+                self.assertEqual(1, int(repl_status))
+
+        try:
+            lib.create_replication_resource(self.admin, replication)
+            lib.create_ufs_resource(self.admin, resource_0, test.settings.HOSTNAME_2)
+            lib.create_ufs_resource(self.admin, resource_1, test.settings.HOSTNAME_3)
+            lib.add_child_resource(self.admin, replication, resource_0)
+            lib.add_child_resource(self.admin, replication, resource_1)
+
+            lib.make_file(local_file, 1024, 'random')
+            self.user0.assert_icommand(['imkdir', collection_path])
+            self.user0.assert_icommand(['iput', local_file, put_path])
+
+            self.user0.assert_icommand(['ibun', '-c', '-Dzip', '-R', replication, archive_path, collection_path])
+            assert_archive_replicas_are_good()
+
+            self.user0.assert_icommand(['ibun', '-c', '-Dzip', '-f', '-R', replication, archive_path, collection_path])
+            assert_archive_replicas_are_good()
+
+        finally:
+            self.user0.run_icommand(['irm', '-f', archive_path])
+            self.user0.run_icommand(['irm', '-r', '-f', collection_path])
+            lib.remove_child_resource(self.admin, replication, resource_0)
+            lib.remove_child_resource(self.admin, replication, resource_1)
+            lib.remove_resource(self.admin, replication)
+            lib.remove_resource(self.admin, resource_0)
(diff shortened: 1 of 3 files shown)


korydraughn:
> However, inspecting the code resulted in detecting that bundling a mounted collection into a replication resource was still a problem.

What does this mean?

> Could make a new issue if we think it correct.

A new issue sounds good. Especially if you've identified a bug.

trel:
have created https://github.com/irods/irods/issues/9067

korydraughn:
Does this pass the unit and core tests?

trel:
yes, core and unit tests pass.

server/api/src/rsStructFileBundle.cpp
```diff
@@ -117,6 +118,57 @@ namespace
...
+        // =-=-=-=-=-=-
+        // Check the source collection before creating the archive object. This
+        // prevents the archive path from shadowing the source collection during
+        // recursive collection reads.
+        chkObjPermAndStat_t chkObjPermAndStatInp;
+        memset(&chkObjPermAndStatInp, 0, sizeof(chkObjPermAndStatInp));
+        rstrcpy(chkObjPermAndStatInp.objPath, structFileBundleInp->collection, MAX_NAME_LEN);
+        chkObjPermAndStatInp.flags = CHK_COLL_FOR_BUNDLE_OPR;
+        addKeyVal(&chkObjPermAndStatInp.condInput, RESC_NAME_KW, resource.c_str());
+        addKeyVal(&chkObjPermAndStatInp.condInput, RESC_HIER_STR_KW, resc_hier.c_str());
```
korydraughn, line 160: `RESC_ID_KW` was set in the original code. Is it still needed?

server/api/src/rsStructFileBundle.cpp
```diff
@@ -117,6 +118,57 @@ namespace
...
+        status = rsChkObjPermAndStat(rsComm, &chkObjPermAndStatInp);
+        clearKeyVal(&chkObjPermAndStatInp.condInput);
```
korydraughn, line 162: The call to clear the `KeyValPair` instance will leak memory if `rsChkObjPermAndStat` throws an exception.

You can use `irods::at_scope_exit` to guard against that.

```suggestion
        chkObjPermAndStat_t chkObjPermAndStatInp{};
        irods::at_scope_exit clear_condinput{[&chkObjPermAndStatInp] { clearKeyVal(&chkObjPermAndStatInp.condInput); }};
        rstrcpy(chkObjPermAndStatInp.objPath, structFileBundleInp->collection, MAX_NAME_LEN);
        chkObjPermAndStatInp.flags = CHK_COLL_FOR_BUNDLE_OPR;
        addKeyVal(&chkObjPermAndStatInp.condInput, RESC_NAME_KW, resource.c_str());
        addKeyVal(&chkObjPermAndStatInp.condInput, RESC_HIER_STR_KW, resc_hier.c_str());
        status = rsChkObjPermAndStat(rsComm, &chkObjPermAndStatInp);
```

server/api/src/rsStructFileBundle.cpp
```diff
@@ -117,6 +118,57 @@ namespace
...
+        {
+            return OVERWRITE_WITHOUT_FORCE_FLAG;
+        }
+
+        // =-=-=-=-=-=-
+        // get the resc hier string
+        std::string resc_hier;
+        char* resc_hier_ptr = getValByKey(&structFileBundleInp->condInput, RESC_HIER_STR_KW);
+        if (!resc_hier_ptr) {
+            rodsLog(LOG_NOTICE, "%s :: RESC_HIER_STR_KW is NULL", __FUNCTION__);
```
korydraughn, line 133: Use new logger. Prefer `__func__` over `__FUNCTION__`.

```suggestion
            log_api::info("{}: RESC_HIER_STR_KW is NULL", __func__);
```

For that suggestion to work, you need to add the following in the anonymous namespace block on L41.

```cpp
```

server/api/src/rsStructFileBundle.cpp
```diff
@@ -117,6 +118,57 @@ namespace
...
+            return SYS_INVALID_RESC_INPUT;
+        }
+
+        resc_hier = resc_hier_ptr;
+
+        irods::hierarchy_parser parser;
+        if (const auto err = parser.set_string(resc_hier); !err.ok()) {
+            irods::log(PASSMSG("failed to parse resource hierarchy", err));
```
korydraughn, line 141: Use the new logger.

I don't know that we need to use `PASSMSG` for this. The log instruction could just be this.

```suggestion
            log_api::error("{}: Failed to parse resource hierarchy string [error code=[{}], hierarchy string=[{}]].", __func__, err.code(), resc_hier);
```

server/api/src/rsStructFileBundle.cpp
```diff
@@ -117,6 +118,57 @@ namespace
...
+            return err.code();
+        }
+
+        std::string resource;
+        if (const auto err = parser.last_resc(resource); !err.ok()) {
+            irods::log(PASSMSG("failed to get leaf resource from hierarchy", err));
```
korydraughn, line 147: Use the new logger.

Same thing here with the use of `PASSMSG`.

server/api/src/rsStructFileBundle.cpp
```diff
@@ -117,6 +118,57 @@ namespace
...
+        addKeyVal(&chkObjPermAndStatInp.condInput, RESC_NAME_KW, resource.c_str());
+        addKeyVal(&chkObjPermAndStatInp.condInput, RESC_HIER_STR_KW, resc_hier.c_str());
+        status = rsChkObjPermAndStat(rsComm, &chkObjPermAndStatInp);
+        clearKeyVal(&chkObjPermAndStatInp.condInput);
+
+        if (status < 0) {
+            rodsLog(LOG_ERROR,
+                    "rsStructFileBundle: rsChkObjPermAndStat of %s error. stat = %d",
+                    chkObjPermAndStatInp.objPath,
+                    status);
```
korydraughn, line 168: Use new logger.

```suggestion
            log_api::error(
                    "{}: rsChkObjPermAndStat failed [error code=[{}], logical path=[{}]].",
                    __func__,
                    status,
                    chkObjPermAndStatInp.objPath);
```

server/api/src/rsChkObjPermAndStat.cpp
```diff
@@ -293,38 +293,41 @@ chkCollForBundleOpr( rsComm_t *rsComm,
...
+    if (collEnt->dataName != nullptr) {
         free( collEnt->dataName );
     }
-    if ( collEnt->resource != NULL ) {
+    if (collEnt->resource != nullptr) {
         free( collEnt->resource );
     }
+    if (collEnt->resc_hier != nullptr) {
+        free(collEnt->resc_hier);
+    }
```
korydraughn, line 330: No need to check for nullptr. `free` is a no-op if the pointer is null.

server/api/src/rsChkObjPermAndStat.cpp
```diff
@@ -293,38 +293,41 @@ chkCollForBundleOpr( rsComm_t *rsComm,
...
         collEnt->dataName = strdup( collEnt->dataName );
     }
-    if ( collEnt->resource != NULL ) {
+    if (collEnt->resource != nullptr) {
         collEnt->resource = strdup( collEnt->resource );
     }
-    if ( collEnt->resc_hier != NULL ) {
+    if (collEnt->resc_hier != nullptr) {
         collEnt->resc_hier = strdup( collEnt->resc_hier );
     }
```
korydraughn, line 310: This code is a little strange. I have to assume this function `saveCollEntForChkColl` is used for a very specific situation since it assumes the pointers aren't pointing to some form of allocated memory?

Do you have any insight into the use of this function and assumptions surrounding it?

server/api/src/rsChkObjPermAndStat.cpp
```diff
@@ -147,7 +147,7 @@ chkCollForBundleOpr( rsComm_t *rsComm,
                              "chkCollForBundleOpr: specColl resc %s does not match %s",
                              collEnt->specColl.resource, resource );
                     rsCloseCollection( rsComm, &handleInx );
-                    freeCollEntForChkColl( collEnt );
+                    free(collEnt);
```
korydraughn, line 150: I take it these calls were changed to `free` for the following reason?

https://github.com/irods/irods/blob/81df38807bc406a239718f70b31d6f58321970b5/lib/filesystem/src/collection_iterator.cpp#L85-L86

Is this one of the places where the double free would happen?

scripts/irods/test/test_ibun.py
```diff
@@ -183,6 +183,121 @@ def do_test_ibun_atop_existing_archive_file_in_replication_hierarchy(self, files
         do_test_ibun_atop_existing_archive_file_in_replication_hierarchy(self, 1038425, 1044480)
         do_test_ibun_atop_existing_archive_file_in_replication_hierarchy(self, 1040425, 1054720)
 
+    def test_ibun_zip_to_replication_resource_with_archive_name_matching_source_collection__issue_5423(self):
+        replication = 'issue_5423_repl'
```
korydraughn, line 187:
```suggestion
        replication_resource = 'issue_5423_repl'
```

scripts/irods/test/test_ibun.py
```diff
@@ -183,6 +183,121 @@ def do_test_ibun_atop_existing_archive_file_in_replication_hierarchy(self, files
...
+        finally:
+            self.user0.run_icommand(['irm', '-f', archive_path])
+            self.user0.run_icommand(['irm', '-r', '-f', collection_path])
+            lib.remove_child_resource(self.admin, replication, resource_0)
+            lib.remove_child_resource(self.admin, replication, resource_1)
+            lib.remove_resource(self.admin, replication)
+            lib.remove_resource(self.admin, resource_0)
+            lib.remove_resource(self.admin, resource_1)
+            if os.path.exists(local_file):
+                os.unlink(local_file)
```
korydraughn, line 238: I think this is unnecessary since the file is in `self.user0.local_session_dir`.

scripts/irods/test/test_ibun.py
```diff
@@ -183,6 +183,121 @@ def do_test_ibun_atop_existing_archive_file_in_replication_hierarchy(self, files
...
+
+            lib.make_file(local_file, 1024, 'random')
+            self.user0.assert_icommand(['imkdir', collection_path])
+            self.user0.assert_icommand(['iput', local_file, put_path])
+
+            self.user0.assert_icommand(['ibun', '-c', '-Dzip', '-R', replication, archive_path, collection_path])
+            assert_archive_replicas_are_good()
+
+            self.user0.assert_icommand(['ibun', '-c', '-Dzip', '-f', '-R', replication, archive_path, collection_path])
+            assert_archive_replicas_are_good()
```
korydraughn, line 227: Some story telling around these lines would be helpful. Six months from now, we might not understand the full picture from just looking at the current form of the test.

scripts/irods/test/test_ibun.py
```diff
@@ -183,6 +183,121 @@ def do_test_ibun_atop_existing_archive_file_in_replication_hierarchy(self, files
...
+            lib.remove_child_resource(self.admin, replication, resource_0)
+            lib.remove_child_resource(self.admin, replication, resource_1)
+            lib.remove_resource(self.admin, replication)
+            lib.remove_resource(self.admin, resource_0)
+            lib.remove_resource(self.admin, resource_1)
+            if os.path.exists(local_file):
+                os.unlink(local_file)
+
+    def test_ibun_reports_resource_mismatch_for_mounted_collection__issue_5423(self):
+        replication = 'issue_5423_repl'
```
korydraughn, line 241:
```suggestion
        replication_resource = 'issue_5423_repl'
```

scripts/irods/test/test_ibun.py
```diff
@@ -183,6 +183,121 @@ def do_test_ibun_atop_existing_archive_file_in_replication_hierarchy(self, files
...
+        mountpoint_directory = os.path.join(self.admin.local_session_dir, 'issue_5423_phymount')
+        rule_file = os.path.join(self.admin.local_session_dir, 'issue_5423_register_mount.r')
+
+        rule_text = '''\
+register_issue_5423_mount {{
+    msiCollCreate(*irodsColl, 0, *status);
+    msiPhyPathReg(*irodsColl, *irodsResc, *phyDir, "mountPoint", *status);
+}}
+
+INPUT *irodsColl="{mounted_collection}", *irodsResc="{mount_resource}", *phyDir="{mountpoint_directory}"
```
korydraughn, line 257: These variables can be baked directly into the rule body.

scripts/irods/test/test_ibun.py
```diff
@@ -183,6 +183,121 @@ def do_test_ibun_atop_existing_archive_file_in_replication_hierarchy(self, files
...
+
+            self.assertEqual(0, ec)
+            self.assertEqual(0, len(err))
+
+            rows = [line.split('...') for line in out.splitlines()]
+            self.assertEqual(2, len(rows), out)
+            self.assertEqual(set([resource_0, resource_1]), set([row[0] for row in rows]))
+
+            for _, repl_status in rows:
+                self.assertEqual(1, int(repl_status))
```
korydraughn, line 210: Seems this can be replaced by a few calls to the following function?

https://github.com/irods/irods/blob/81df38807bc406a239718f70b31d6f58321970b5/scripts/irods/lib.py#L648

scripts/irods/test/test_ibun.py
```diff
@@ -183,6 +183,121 @@ def do_test_ibun_atop_existing_archive_file_in_replication_hierarchy(self, files
...
+
+            self.admin.assert_icommand(['imkdir', collection_path])
+            self.admin.assert_icommand(['irule', '-F', rule_file])
+            self.admin.assert_icommand(['ils', '-AL', mounted_collection], 'STDOUT_SINGLELINE', 'mountP')
+
+            out, err, ec = self.admin.run_icommand(
+                ['ibun', '-c', '-Dzip', '-R', replication, archive_path, collection_path])
+            self.assertNotEqual(0, ec)
+            self.assertEqual('', out)
+            self.assertIn('SYS_COPY_NOT_EXIST_IN_RESC', err)
```
korydraughn, line 284: Let's include the error code with the symbolic name to guard against future substring issues.

scripts/irods/test/test_ibun.py
```diff
@@ -183,6 +183,121 @@ def do_test_ibun_atop_existing_archive_file_in_replication_hierarchy(self, files
...
+            self.admin.run_icommand(['irm', '-r', '-f', collection_path])
+            self.admin.run_icommand(['iadmin', 'rmchildfromresc', replication, resource_0])
+            self.admin.run_icommand(['iadmin', 'rmchildfromresc', replication, resource_1])
+            self.admin.run_icommand(['iadmin', 'rmresc', replication])
+            self.admin.run_icommand(['iadmin', 'rmresc', resource_0])
+            self.admin.run_icommand(['iadmin', 'rmresc', resource_1])
+            self.admin.run_icommand(['iadmin', 'rmresc', mount_resource])
+            shutil.rmtree(mountpoint_directory, ignore_errors=True)
+            if os.path.exists(rule_file):
+                os.unlink(rule_file)
```
korydraughn, line 299: This is probably unnecessary since the file is created in the local_session_dir.

Issue #9065, opened 2026-08-31
`std::string_view::back` invoked on empty string


korydraughn:
The following lines of code trigger undefined behavior if `_logical_path` is an empty string.

https://github.com/irods/irods/blob/69874094328aba0667d0b56156a6483b4d3668c6/lib/core/src/rodsPath.cpp#L513

https://github.com/irods/irods/blob/69874094328aba0667d0b56156a6483b4d3668c6/lib/core/src/rodsPath.cpp#L529

See https://cppreference.com/cpp/string/basic_string_view/back.

This was observed while testing against Ubuntu 26.04. I did not see this on Ubuntu 22.04.

The following test would trigger `SIGABRT`.

https://github.com/irods/irods/blob/69874094328aba0667d0b56156a6483b4d3668c6/unit_tests/src/test_logical_paths_and_special_characters.cpp#L221-L226

_Also, use a nonempty string for the `SECTION` title._

PR #9066, opened 2026-08-31
[#9065] Check for empty string to avoid `SIGABRT` (main)

korydraughn:
Unit tests pass.

Running core tests.


unit_tests/src/test_logical_paths_and_special_characters.cpp
```diff
@@ -218,7 +218,7 @@ TEST_CASE("test getLastPathElement")
         REQUIRE(last_path_element == ""s);
     }
 
-    SECTION("")
+    SECTION("empty path")
```
korydraughn: Consider renaming to `empty string`.

korydraughn:
Core tests pass.

unit_tests/src/test_logical_paths_and_special_characters.cpp
korydraughn: Changed to `empty path string`.

korydraughn:
This is now ready for final review.

trel (approved):

PR #9068, opened 2026-09-02
[9055] Expose modify_ts for logical quotas (main)

FifthPotato:
Ran `test_logical_quotas` against all of our tested DBMSes and it passes.

Taking comments as always.

### DIFF
--- a/clients/icommands/src/iadmin.cpp
+++ b/clients/icommands/src/iadmin.cpp
@@ -31,6 +31,7 @@
 #include <termios.h>
 #include <unistd.h>
 
+#include <boost/date_time.hpp>
 #include <boost/lexical_cast.hpp>
 #include <fmt/compile.h>
 #include <fmt/format.h>
@@ -1019,12 +1020,29 @@ auto list_logical_quotas(const char* _coll_name = nullptr) -> int
         const std::string max_objects_string = std::to_string(quotaEntry.max_objects);
         const std::string bytes_over_string = std::to_string(quotaEntry.over_bytes);
         const std::string objects_over_string = std::to_string(quotaEntry.over_objects);
-        fmt::print("Collection name: {}\nMaximum bytes: {}\nMaximum objects: {}\nBytes over: {}\nObjects over: {}\n",
+
+        std::string modify_ts_string = "?";
+        try {
+            const auto pt = boost::posix_time::from_time_t(std::strtoll(quotaEntry.modify_time, nullptr, 10));
+
+            std::ostringstream ss;
+            ss.imbue(std::locale(ss.getloc(), new boost::posix_time::time_facet{"%Y-%m-%d.%H:%M:%S"}));
+            ss << pt;
+
+            modify_ts_string = ss.str();
+        }
+        catch (const std::exception& e) {
+            fmt::print(stderr, "Timestamp conversion error: {}", e.what());
+        }
+
+        fmt::print("Collection name: {}\nMaximum bytes: {}\nMaximum objects: {}\nBytes over: {}\nObjects over: "
+                   "{}\nLast modified: {}\n",
                    quotaEntry.coll_name,
                    (byte_limit_enforced ? max_byte_string.c_str() : "<unset>"),
                    (object_limit_enforced ? max_objects_string.c_str() : "<unset>"),
                    (byte_limit_enforced ? bytes_over_string.c_str() : "<unenforced>"),
-                   (object_limit_enforced ? objects_over_string.c_str() : "<unenforced>"));
+                   (object_limit_enforced ? objects_over_string.c_str() : "<unenforced>"),
+                   modify_ts_string);
         if (i != logicalQuotaList->len - 1) {
             fmt::print("\n");
         }
--- a/lib/api/include/irods/get_logical_quota.h
+++ b/lib/api/include/irods/get_logical_quota.h
@@ -68,9 +68,19 @@ typedef struct LogicalQuota
     ///
     /// \since 5.1.0
     rodsLong_t over_objects;
+
+    /// The last modification time of the quota, as seconds from the epoch.
+    /// This is set on recalculation (even if no values were changed), creation,
+    /// and update of a logical quota.
+    ///
+    /// \since 5.1.0
+    char modify_time[TIME_LEN];
+
 } logicalQuota_t;
 
-#define logicalQuota_PI "str *collName; double maxBytes; double maxObjects; double overBytes; double overObjects;"
+#define logicalQuota_PI                                                                           \
+    "str *collName; double maxBytes; double maxObjects; double overBytes; double overObjects; str " \
+    "modify_time[TIME_LEN];"
 
 /// The type that represents a list of logical quota entries.
 ///
--- a/scripts/irods/test/test_logical_quotas.py
+++ b/scripts/irods/test/test_logical_quotas.py
@@ -1,7 +1,10 @@
-import os
-import sys
-import unittest
 import contextlib
+import datetime
+import os
+import re
+import sys
+import time
+import unittest
 
 from . import session
 from .. import database_connect
@@ -4454,6 +4457,89 @@ class Test_Logical_Quotas(
                     "iadmin",
                     "set_logical_quota",
                     other_subcoll_path,
+                    "0",
+                    "0",
+                ]
+            )
+
+
+    def test_logical_quota_timestamps__issue_9055(self):
+        subcoll_path = f"{self.quota_user.session_collection}/potato"
+        max_bytes = 10000
+        max_objects = 50
+
+        timestamp_regex = re.compile(r'Last modified: (.*)')
+        try:
+            self.quota_user.assert_icommand(["imkdir", subcoll_path])
+            self.admin.assert_icommand(
+                [
+                    "iadmin",
+                    "set_logical_quota",
+                    subcoll_path,
+                    str(max_bytes),
+                    str(max_objects),
+                ]
+            )
+            _, out, _ = self.admin.assert_icommand(
+                ["iadmin", "list_logical_quotas"], "STDOUT_SINGLELINE", subcoll_path
+            )
+
+            self.assertTrue(
+                (
+                    self.llq_output_template
+                    % (
+                        subcoll_path,
+                        str(max_bytes),
+                        str(max_objects),
+                        str(-max_bytes),
+                        str(-max_objects),
+                    )
+                )
+                in out
+            )
+
+            first_ts = timestamp_regex.search(out)
+
+            # Assert that we found something in "Last modified"
+            self.assertNotEqual(first_ts.group(1), None)
+
+            # Wait two seconds to induce a modify_time change
+            time.sleep(2)
+
+            self.admin.assert_icommand(["iadmin", "calculate_logical_usage"])
+
+            _, out, _ = self.admin.assert_icommand(
+                ["iadmin", "list_logical_quotas"], "STDOUT_SINGLELINE", subcoll_path
+            )
+
+            # All except timestamp should be unchanged, since nothing was added
+            self.assertTrue(
+                (
+                    self.llq_output_template
+                    % (
+                        subcoll_path,
+                        str(max_bytes),
+                        str(max_objects),
+                        str(-max_bytes),
+                        str(-max_objects),
+                    )
+                )
+                in out
+            )
+
+            second_ts = timestamp_regex.search(out)
+
+            self.assertNotEqual(second_ts.group(1), None)
+
+            # Assert that the first timestamp is before the second
+            self.assertTrue(datetime.datetime.strptime(first_ts.group(1), '%Y-%m-%d.%H:%M:%S') < datetime.datetime.strptime(second_ts.group(1), '%Y-%m-%d.%H:%M:%S'))
+
+        finally:
+            self.admin.run_icommand(
+                [
+                    "iadmin",
+                    "set_logical_quota",
+                    subcoll_path,
                     "0",
                     "0",
                 ]
--- a/server/core/include/irods/logical_quota_utilities.hpp
+++ b/server/core/include/irods/logical_quota_utilities.hpp
@@ -5,13 +5,25 @@
 
 #include <sys/types.h>
 
+#include <cstdint>
 #include <functional>
 #include <type_traits>
+#include <string>
+#include <tuple>
+#include <vector>
 
 struct RsComm;
 
 namespace irods::logical_quotas
 {
+    /// Type that represents a vector of quotas.
+    ///
+    /// Used internally to pass around lists of quotas.
+    ///
+    /// \since 5.1.0
+    using quota_vector =
+        std::vector<std::tuple<std::string, std::int64_t, std::int64_t, std::int64_t, std::int64_t, std::string>>;
+
     /// Enum that represents a quota violation mode.
     ///
     /// Supports bitwise operations to represent different combinations of quota types being violated.
--- a/server/icat/include/irods/icatHighLevelRoutines.hpp
+++ b/server/icat/include/irods/icatHighLevelRoutines.hpp
@@ -3,6 +3,7 @@
 
 /// \file
 
+#include "irods/logical_quota_utilities.hpp"
 #include "irods/objInfo.h"
 #include "irods/ruleExecSubmit.h"
 #include "irods/rcConnect.h"
@@ -676,9 +677,7 @@ auto chl_set_logical_quota(RsComm* _comm, const char* _coll_name, const char* _b
 /// \retval <0 On failure.
 ///
 /// \since 5.1.0
-auto chl_check_logical_quota(
-    RsComm* _comm,
-    const char* _coll_name,
-    std::vector<std::tuple<std::string, std::int64_t, std::int64_t, std::int64_t, std::int64_t>>* _quota_values) -> int;
+auto chl_check_logical_quota(RsComm* _comm, const char* _coll_name, irods::logical_quotas::quota_vector* _quota_values)
+    -> int;
 
 #endif // IRODS_ICAT_HIGHLEVEL_ROUTINES_HPP
--- a/server/re/src/irods_re_serialization.cpp
+++ b/server/re/src/irods_re_serialization.cpp
@@ -2,6 +2,7 @@
 
 #include "irods/irods_logger.hpp"
 #include "irods/irods_plugin_context.hpp"
+#include "irods/logical_quota_utilities.hpp"
 #include "irods/msParam.h"
 #include "irods/rodsErrorTable.h"
 #include "irods/ticketAdmin.h"
@@ -1441,12 +1442,11 @@ namespace irods::re_serialization
         return SUCCESS();
     } // serialize_ticketAdminInp_ptr
 
-    static error serialize_ptr_to_vector_of_tuples_of_std_string_and_four_int64s(boost::any _p,
+    static error serialize_ptr_to_vector_of_tuples_of_std_string_and_five_int64s(boost::any _p,
                                                                                  serialized_parameter_t& _out)
     {
         try {
-            using tupletype = std::tuple<std::string, std::int64_t, std::int64_t, std::int64_t, std::int64_t>;
-            const std::vector<tupletype>* inp = boost::any_cast<std::vector<tupletype>*>(_p);
+            const irods::logical_quotas::quota_vector* inp = boost::any_cast<irods::logical_quotas::quota_vector*>(_p);
             if (inp) {
                 _out["size"] = std::to_string(inp->size());
 
@@ -1455,12 +1455,13 @@ namespace irods::re_serialization
                 _out["_size"] = std::to_string(inp->size());
 
                 for (size_t i = 0; i < inp->size(); i++) {
-                    tupletype tup = inp->at(i);
+                    irods::logical_quotas::quota_vector::value_type tup = inp->at(i);
                     _out[std::to_string(i) + ":std_string"] = std::get<0>(tup);
                     _out[std::to_string(i) + ":std_int64_1"] = std::get<1>(tup);
                     _out[std::to_string(i) + ":std_int64_2"] = std::get<2>(tup);
                     _out[std::to_string(i) + ":std_int64_3"] = std::get<3>(tup);
                     _out[std::to_string(i) + ":std_int64_4"] = std::get<4>(tup);
+                    _out[std::to_string(i) + ":std_int64_5"] = std::get<5>(tup);
                 }
             }
             else {
@@ -1470,14 +1471,14 @@ namespace irods::re_serialization
         catch (const boost::bad_any_cast& e) {
             return ERROR(INVALID_ANY_CAST,
                          fmt::format("{}: failed to cast pointer to [std::vector<std::tuple<std::string, std::int64_t, "
-                                     "std::int64_t, std::int64_t>>*]: {}",
+                                     "std::int64_t, std::int64_t, std::int64_t>>*]: {}",
                                      __func__,
                                      e.what()));
         }
         catch (const std::exception& e) {
             return ERROR(SYS_LIBRARY_ERROR,
                          fmt::format("{}: failed to serialize [std::vector<std::tuple<std::string, std::int64_t, "
-                                     "std::int64_t, std::int64_t>>*]: {}",
+                                     "std::int64_t, std::int64_t, std::int64_t>>*]: {}",
                                      __func__,
                                      e.what()));
         }
@@ -1546,9 +1547,8 @@ namespace irods::re_serialization
             {std::type_index(typeid(structFileExtAndRegInp_t*)), serialize_structFileExtAndRegInp_ptr},
             {std::type_index(typeid(msParamArray_t**)), serialize_msParamArray_ptr_ptr},
             {std::type_index(typeid(TicketAdminInput*)), serialize_ticketAdminInp_ptr},
-            {std::type_index(
-                 typeid(std::vector<std::tuple<std::string, std::int64_t, std::int64_t, std::int64_t, std::int64_t>>*)),
-             serialize_ptr_to_vector_of_tuples_of_std_string_and_four_int64s}};
+            {std::type_index(typeid(irods::logical_quotas::quota_vector*)),
+             serialize_ptr_to_vector_of_tuples_of_std_string_and_five_int64s}};
         return the_map;
     } // get_serialization_map
 
--- a/server/re/src/irods_re_serialization.cpp
+++ b/server/re/src/irods_re_serialization.cpp
@@ -1442,7 +1442,7 @@ namespace irods::re_serialization
         return SUCCESS();
     } // serialize_ticketAdminInp_ptr
 
-    static error serialize_ptr_to_vector_of_tuples_of_std_string_and_five_int64s(boost::any _p,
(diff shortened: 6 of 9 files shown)


server/icat/include/irods/icatHighLevelRoutines.hpp
```diff
@@ -679,6 +679,7 @@ auto chl_set_logical_quota(RsComm* _comm, const char* _coll_name, const char* _b
 auto chl_check_logical_quota(
     RsComm* _comm,
     const char* _coll_name,
-    std::vector<std::tuple<std::string, std::int64_t, std::int64_t, std::int64_t, std::int64_t>>* _quota_values) -> int;
+    std::vector<std::tuple<std::string, std::int64_t, std::int64_t, std::int64_t, std::int64_t, std::int64_t>>*
```
korydraughn: Consider introducing a type alias and using that everywhere. That will help with maintaining all of the code which have this declaration.

lib/api/include/irods/get_logical_quota.h
```diff
@@ -68,9 +68,18 @@ typedef struct LogicalQuota
     ///
     /// \since 5.1.0
     rodsLong_t over_objects;
+
+    /// The last modification time of the quota, as seconds from the epoch.
+    /// This is set on recalculation (even if no values were changed), creation,
+    /// and update of a logical quota.
+    ///
+    /// \since 5.1.0
+    rodsLong_t modify_ts;
```
korydraughn: I think many if not all iRODS APIs return the raw timestamp strings as is.

I'm wondering if we should do the same thing for this. Hmm.

clients/icommands/src/iadmin.cpp
```diff
@@ -1019,12 +1020,29 @@ auto list_logical_quotas(const char* _coll_name = nullptr) -> int
...
+            const auto pt = boost::posix_time::from_time_t(quotaEntry.modify_ts);
+
+            std::ostringstream ss;
+            ss.imbue(std::locale(ss.getloc(), new boost::posix_time::time_facet{"%Y-%m-%d.%H:%M:%S"}));
+            ss << pt;
+
+            modify_ts_string = ss.str();
+        }
+        catch (const std::exception& e) {
+            std::cerr << "Timestamp conversion error: " << e.what() << '\n';
```
korydraughn: Prefer `fmt::print(stderr, ...)`

scripts/irods/test/test_logical_quotas.py
```diff
@@ -4458,3 +4461,86 @@ def test_logical_quotas_do_not_overcount_collections_with_like_wildcards__issue_
...
+
+            first_ts = timestamp_regex.search(out)
+
+            # Assert that we found something in "Last modified"
+            self.assertNotEqual(first_ts.group(1), None)
+
+            # Wait two seconds to induce a modify_ts change
+            time.sleep(2)
+
+            self.admin.assert_icommand(["iadmin", "calculate_logical_usage"])
```
korydraughn, line 4509: The idea is that the mtime is only updated when `calculate_logical_usage` is invoked, correct?

lib/api/include/irods/get_logical_quota.h
korydraughn: All members of this struct share the same name as their database column?

scripts/irods/test/test_logical_quotas.py
FifthPotato: It can also be updated if one updates an existing quota (i.e. with `iadmin set_logical_quota`).

lib/api/include/irods/get_logical_quota.h
FifthPotato: Are there any examples? From what I understand, it's generally frowned upon to send stringified integers over the network since they're generally bigger than their integer equivalents.

scripts/irods/test/test_logical_quotas.py
korydraughn: Ah, got it.

So, a person should view that timestamp as the time at which the entry in the catalog was last modified -- i.e. the what and how behind the modification doesn't matter.

lib/api/include/irods/get_logical_quota.h
FifthPotato: Correct. Well, with the exception of `coll_name` since that's linked to another table. But the others all match.

korydraughn: > Are there any examples?

`rcObjStat`, `rcModColl`, and GenQuery are a few examples.

https://github.com/irods/irods/blob/81df38807bc406a239718f70b31d6f58321970b5/lib/api/include/irods/objStat.h#L16-L17

Try grepping for `TIME_LEN`. That might identify others.

> From what I understand, it's generally frowned upon to send stringified integers over the network since they're generally bigger than their integer equivalents.

I agree, but I think iRODS uses strings to stay flexible. My guess is that it's also due to legacy reasons.

server/icat/include/irods/icatHighLevelRoutines.hpp
FifthPotato: Is there a preferred header for doing so? A quick recursive grep tells me that the following files use this definition:
```
server/re/src/irods_re_serialization.cpp
server/api/src/rs_get_logical_quota.cpp
server/icat/src/icatHighLevelRoutines.cpp
server/icat/include/irods/icatHighLevelRoutines.hpp
plugins/database/src/db_plugin.cpp
```
However, on first glance, I don't see a common header amongst these.

lib/api/include/irods/get_logical_quota.h
korydraughn: Makes sense. I was wondering if there was a common name we use for time-based members.

I'm mostly staring at the `_ts` suffix. What if we expanded `_ts` to `_time`?

`modify_time` is closer to the name in `rodsObjStat` (see https://github.com/irods/irods/pull/9068#discussion_r3916296824) and makes it clear to the reader that the member is related to time. New readers might not pick up on the `_ts` being short for timestamp _(even I'm assuming that's what it stands for)_.

scripts/irods/test/test_logical_quotas.py
FifthPotato: That's the gist of it!

server/icat/include/irods/icatHighLevelRoutines.hpp
korydraughn: Could put it in logical_quotas_utilities.hpp?

scripts/irods/test/test_logical_quotas.py
korydraughn: Good. Resolving.

lib/api/include/irods/get_logical_quota.h
FifthPotato: Hmm, I see. As long as we don't mind the minor efficiency hit, I can change it, though it will require a little refactoring. 
It does seem to work as-is, so do we think it's worth it to change it over?

korydraughn: Given the mtime is stored in the table as a string, let's just return it as a string.

Yes, it's not as efficient, but it's consistent with the rest of the design and familiar to users. We can always change it later, in a major release.

FifthPotato: Done.

FifthPotato: Done.

server/icat/include/irods/icatHighLevelRoutines.hpp
FifthPotato: Done.

server/core/include/irods/logical_quota_utilities.hpp
```diff
@@ -5,13 +5,26 @@
...
 #include <type_traits>
+#include <string>
+#include <tuple>
+#include <vector>
 
 struct RsComm;
 
 namespace irods::logical_quotas
 {
+
```
korydraughn: Remove empty line.

clients/icommands/src/iadmin.cpp
```diff
@@ -1019,12 +1020,29 @@ auto list_logical_quotas(const char* _coll_name = nullptr) -> int
         const std::string max_objects_string = std::to_string(quotaEntry.max_objects);
         const std::string bytes_over_string = std::to_string(quotaEntry.over_bytes);
         const std::string objects_over_string = std::to_string(quotaEntry.over_objects);
-        fmt::print("Collection name: {}\nMaximum bytes: {}\nMaximum objects: {}\nBytes over: {}\nObjects over: {}\n",
+
+        std::string modify_ts_string = "?";
+        try {
+            const auto pt = boost::posix_time::from_time_t(std::strtoll(quotaEntry.modify_time, NULL, 10));
```
korydraughn:
```suggestion
            const auto pt = boost::posix_time::from_time_t(std::strtoll(quotaEntry.modify_time, nullptr, 10));
```

korydraughn (commented):
Looks good to me.

Please run the logical quotas tests. Squash to taste if passing.

FifthPotato:
Implemented suggested changes. Ran `test_logical_quotas` on all DBMSes again, and all pass. Squished.

server/re/src/irods_re_serialization.cpp
```diff
@@ -1441,12 +1442,11 @@ namespace irods::re_serialization
         return SUCCESS();
     } // serialize_ticketAdminInp_ptr
 
-    static error serialize_ptr_to_vector_of_tuples_of_std_string_and_four_int64s(boost::any _p,
+    static error serialize_ptr_to_vector_of_tuples_of_std_string_and_five_int64s(boost::any _p,
```
korydraughn: Do we have any tests that verify these values are available to the NREP and PREP?

korydraughn: Is this name still correct?

server/re/src/irods_re_serialization.cpp
```diff
@@ -1470,14 +1471,14 @@ namespace irods::re_serialization
         catch (const boost::bad_any_cast& e) {
             return ERROR(INVALID_ANY_CAST,
                          fmt::format("{}: failed to cast pointer to [std::vector<std::tuple<std::string, std::int64_t, "
-                                     "std::int64_t, std::int64_t>>*]: {}",
+                                     "std::int64_t, std::int64_t, std::int64_t>>*]: {}",
```
korydraughn: Should this be `tuple<1 string, 4 ints, 1 string>`?

server/re/src/irods_re_serialization.cpp
FifthPotato: No, needs an update. Good catch.

FifthPotato: No. I've hand-tested it with the NREP to check that the keys show up, but it seems difficult to get the NREP to work with 64-bit integer values.
Should I figure out a test for this?

server/re/src/irods_re_serialization.cpp
FifthPotato: Yup.

server/re/src/irods_re_serialization.cpp
korydraughn: Yes. We need to make sure those values are available to the NREP and PREP.

I think you're going to have to add support for that type to the PREP.

FifthPotato:
Added tests that test the serialization parameter for NREP and PREP.
NREP tested and passing on all DBMSes. PREP tested and passing on one.

scripts/irods/test/test_logical_quotas.py
```diff
@@ -4458,3 +4464,192 @@ def test_logical_quotas_do_not_overcount_collections_with_like_wildcards__issue_
...
+                [
+                    "iadmin",
+                    "set_logical_quota",
+                    subcoll_path,
+                    "0",
+                    "0",
+                ]
+            )
+
+    def test_logical_quota_re_serialization_parameter(self):
```
korydraughn: Missing issue number.

scripts/irods/test/test_logical_quotas.py
```diff
@@ -4458,3 +4464,192 @@ def test_logical_quotas_do_not_overcount_collections_with_like_wildcards__issue_
...
+                        msiGetValByKey(*quotas, str(*i) ++ ":std_int64_2", *D);
+                        msiGetValByKey(*quotas, str(*i) ++ ":std_int64_3", *E);
+                        msiGetValByKey(*quotas, str(*i) ++ ":std_int64_4", *F);
+                        msiModAVUMetadata('-C', '{self.quota_user.session_collection}', 'set', 'TEST_MARKER_re_serialization_path', *A, '');
+                        msiModAVUMetadata('-C', '{self.quota_user.session_collection}', 'set', 'TEST_MARKER_re_serialization_modify_time', *B, '');
+                        msiModAVUMetadata('-C', '{self.quota_user.session_collection}', 'set', 'TEST_MARKER_re_serialization_max_bytes', *C, '');
+                        msiModAVUMetadata('-C', '{self.quota_user.session_collection}', 'set', 'TEST_MARKER_re_serialization_max_objects', *D, '');
+                        msiModAVUMetadata('-C', '{self.quota_user.session_collection}', 'set', 'TEST_MARKER_re_serialization_over_bytes', *E, '');
+                        msiModAVUMetadata('-C', '{self.quota_user.session_collection}', 'set', 'TEST_MARKER_re_serialization_over_objects', *F, '');
```
korydraughn: Can pull the attribute names out and store them in variables so that they can be used here and in the assertions that follow.

korydraughn: Could replace `TEST_MARKER` with the issue number in the attribute names? Would make the attribute names more unique.

scripts/irods/test/test_logical_quotas.py
```diff
@@ -4458,3 +4464,192 @@ def test_logical_quotas_do_not_overcount_collections_with_like_wildcards__issue_
...
+                )
+
+                self.assertTrue(lib.metadata_attr_with_value_exists(self.quota_user, 'TEST_MARKER_re_serialization_path', self.quota_user.session_collection))
+                self.assertTrue(lib.metadata_attr_with_value_exists(self.quota_user, 'TEST_MARKER_re_serialization_max_bytes', str(max_bytes)))
+                self.assertTrue(lib.metadata_attr_with_value_exists(self.quota_user, 'TEST_MARKER_re_serialization_max_objects', str(max_objects)))
+                self.assertTrue(lib.metadata_attr_with_value_exists(self.quota_user, 'TEST_MARKER_re_serialization_over_bytes', str(-max_bytes)))
+                self.assertTrue(lib.metadata_attr_with_value_exists(self.quota_user, 'TEST_MARKER_re_serialization_over_objects', str(-max_objects)))
+
+                out, _, _ = self.quota_user.run_icommand(['iquest', '%s',
+                    "select META_DATA_ATTR_VALUE where META_DATA_ATTR_NAME = 'TEST_MARKER_re_serialization_modify_time'"])
```
korydraughn: You can pass `STDOUT` to `assert_icommand` to get the output to appear in the XML file. You also get error checking.
```suggestion
                _, out, _ = self.quota_user.assert_icommand(['iquest', '%s',
                    "select META_DATA_ATTR_VALUE where META_DATA_ATTR_NAME = 'TEST_MARKER_re_serialization_modify_time'"], 'STDOUT')
```

server/re/src/irods_re_serialization.cpp
```diff
@@ -1455,12 +1455,13 @@ namespace irods::re_serialization
...
-                    _out[std::to_string(i) + ":std_int64_2"] = std::get<2>(tup);
-                    _out[std::to_string(i) + ":std_int64_3"] = std::get<3>(tup);
-                    _out[std::to_string(i) + ":std_int64_4"] = std::get<4>(tup);
+                    irods::logical_quotas::quota_vector::value_type tup = inp->at(i);
+                    _out[std::to_string(i) + ":std_string_1"] = std::get<0>(tup);
+                    _out[std::to_string(i) + ":std_int64_1"] = std::to_string(std::get<1>(tup));
+                    _out[std::to_string(i) + ":std_int64_2"] = std::to_string(std::get<2>(tup));
+                    _out[std::to_string(i) + ":std_int64_3"] = std::to_string(std::get<3>(tup));
+                    _out[std::to_string(i) + ":std_int64_4"] = std::to_string(std::get<4>(tup));
+                    _out[std::to_string(i) + ":std_string_2"] = std::get<5>(tup);
```
korydraughn: Consider dropping the `std_` and replacing `int64` with `int`. Those pieces of information feel like they tie the data to C++ a little too tightly.

FifthPotato:
`test_logical_quotas` still passes.

scripts/irods/test/test_logical_quotas.py
```diff
@@ -4458,3 +4464,196 @@ def test_logical_quotas_do_not_overcount_collections_with_like_wildcards__issue_
...
+                            str(max_bytes),
+                            str(max_objects),
+                            str(-max_bytes),
+                            str(-max_objects),
+                        )
+                    )
+                    in out
+                )
+
+                self.assertTrue(lib.metadata_attr_with_value_exists(self.quota_user, full_parameter_names['path'], self.quota_user.session_collection))
```
korydraughn: `lib.metadata_attr_with_value_exists` doesn't support taking a path to a collection.

You're going to need to add a version that does to lib.py.

scripts/irods/test/test_logical_quotas.py
```diff
@@ -4458,3 +4464,196 @@ def test_logical_quotas_do_not_overcount_collections_with_like_wildcards__issue_
...
+                self.assertTrue(lib.metadata_attr_with_value_exists(self.quota_user, full_parameter_names['max_bytes'], str(max_bytes)))
+                self.assertTrue(lib.metadata_attr_with_value_exists(self.quota_user, full_parameter_names['max_objects'], str(max_objects)))
+                self.assertTrue(lib.metadata_attr_with_value_exists(self.quota_user, full_parameter_names['over_bytes'], str(-max_bytes)))
+                self.assertTrue(lib.metadata_attr_with_value_exists(self.quota_user, full_parameter_names['over_objects'], str(-max_objects)))
```
korydraughn: These functions only test for existence of metadata. They do not confirm that the metadata is attached to the collection.

scripts/irods/test/test_logical_quotas.py
```diff
@@ -4458,3 +4464,196 @@ def test_logical_quotas_do_not_overcount_collections_with_like_wildcards__issue_
...
+
+                _, out, _ = self.quota_user.assert_icommand(['iquest', '%s',
+                    f"select META_DATA_ATTR_VALUE where META_DATA_ATTR_NAME = '{full_parameter_names['modify_time']}'"], 'STDOUT')
```
korydraughn: - Needs to use `META_COLL_` instead of `META_DATA_`
- Needs `COLL_NAME = '<path>'` condition

scripts/irods/test/test_logical_quotas.py
korydraughn: I have an implementation of this, but I haven't opened a PR for it yet.

You can do the same. We can race.

FifthPotato: Are we happy with this impl?

scripts/irods/lib.py
```diff
@@ -576,6 +576,20 @@ def metadata_attr_with_value_exists(session, attr, value):
     print(out)
     return value in out
 
+def metadata_attr_with_value_exists_on_named_coll_or_obj(session, attr, value, path, coll_or_obj='collection'):
+    print('looking for attr:[{0}] value:[{1}] on {3} [{2}]'.format(attr, value, path, coll_or_obj))
+
+    type_functions = {
+                    'data_object': lambda a, p: session.run_icommand(['iquest', '%s',
+            'select META_DATA_ATTR_VALUE where META_DATA_ATTR_NAME = \'{}\' and COLL_NAME = \'{}\' and DATA_NAME = \'{}\''.format(a, *p.rsplit('/', 1))]),
```
trel: needs another parameter to the format() call?

scripts/irods/lib.py
```diff
@@ -576,6 +576,20 @@ def metadata_attr_with_value_exists(session, attr, value):
     print(out)
     return value in out
 
+def metadata_attr_with_value_exists_on_named_coll_or_obj(session, attr, value, path, coll_or_obj='collection'):
```
korydraughn: Let's drop the 5th parameter and add dedicated functions for each target. That gives each function one responsibility and removed the chance of mismatch args for `path` and `coll_or_obj`.

It also allows us to use a more explicit function name.

scripts/irods/lib.py
FifthPotato: Nope, python magic!
(from REPL)
```
INS|>>> a = 'potato'
INS|>>> p = '/tempZone/home/rods/tomato'
INS|>>> 'select META_DATA_ATTR_VALUE where META_DATA_ATTR_NAME = \'{}\' and COLL_NAME = \'{}\' and DATA_NAME = \'{}\''.format(a, *p.rsplit('/', 1))
"select META_DATA_ATTR_VALUE where META_DATA_ATTR_NAME = 'potato' and COLL_NAME = '/tempZone/home/rods' and DATA_NAME = 'tomato'"
```

Too evil?

scripts/irods/test/test_logical_quotas.py
korydraughn: Looked at the new impl. See https://github.com/irods/irods/pull/9068#discussion_r4039354523.

scripts/irods/lib.py
```diff
@@ -576,20 +576,25 @@ def metadata_attr_with_value_exists(session, attr, value):
...
-def metadata_attr_with_value_exists_on_named_coll_or_obj(session, attr, value, path, coll_or_obj='collection'):
-    print('looking for attr:[{0}] value:[{1}] on {3} [{2}]'.format(attr, value, path, coll_or_obj))
-
-    type_functions = {
-                    'data_object': lambda a, p: session.run_icommand(['iquest', '%s',
-            'select META_DATA_ATTR_VALUE where META_DATA_ATTR_NAME = \'{}\' and COLL_NAME = \'{}\' and DATA_NAME = \'{}\''.format(a, *p.rsplit('/', 1))]),
-                    'collection': lambda a, p: session.run_icommand(['iquest', '%s',
-            'select META_COLL_ATTR_VALUE where META_COLL_ATTR_NAME = \'{}\' and COLL_NAME = \'{}\''.format(a, p)])
-    }
+def metadata_attr_with_value_exists_on_named_obj(session, attr, value, path):
```
korydraughn: `on_named_obj` -> `on_data_object`

scripts/irods/lib.py
```diff
@@ -576,20 +576,25 @@ def metadata_attr_with_value_exists(session, attr, value):
...
+    print('looking for attr:[{0}] value:[{1}] on data object [{2}]'.format(attr, value, path))
+
+    out, _, _ = session.run_icommand(['iquest', '%s',
+            'select META_DATA_ATTR_VALUE where META_DATA_ATTR_NAME = \'{}\' and COLL_NAME = \'{}\' and DATA_NAME = \'{}\''.format(attr, *path.rsplit('/', 1))]),
+
+    print(out)
+    return value in out
+
+def metadata_attr_with_value_exists_on_named_coll(session, attr, value, path):
```
korydraughn: `on_named_coll` -> `on_collection`

scripts/irods/test/test_logical_quotas.py
```diff
@@ -4634,11 +4634,11 @@ def pep_database_check_logical_quota_post(rule_args, callback, rei):
...
-                self.assertTrue(lib.metadata_attr_with_value_exists_on_named_coll_or_obj(self.quota_user, full_parameter_names['max_bytes'], str(max_bytes), self.quota_user.session_collection))
-                self.assertTrue(lib.metadata_attr_with_value_exists_on_named_coll_or_obj(self.quota_user, full_parameter_names['max_objects'], str(max_objects), self.quota_user.session_collection))
-                self.assertTrue(lib.metadata_attr_with_value_exists_on_named_coll_or_obj(self.quota_user, full_parameter_names['over_bytes'], str(-max_bytes), self.quota_user.session_collection))
-                self.assertTrue(lib.metadata_attr_with_value_exists_on_named_coll_or_obj(self.quota_user, full_parameter_names['over_objects'], str(-max_objects), self.quota_user.session_collection))
+                self.assertTrue(lib.metadata_attr_with_value_exists_on_named_coll(self.quota_user, full_parameter_names['path'], self.quota_user.session_collection, self.quota_user.session_collection))
```
korydraughn: session_collection is passed twice?

scripts/irods/lib.py
korydraughn: Seems fine to me as long as it doesn't allow the test to pass with incorrect arguments.

FifthPotato: This one isn't used in this test, but the genquery looks correct to me.

scripts/irods/test/test_logical_quotas.py
FifthPotato: Yup, intentional. The quota is set on the session collection, so the value of the AVU is also the session collection.
So it's an AVU valued with the session collection set on the session collection.

korydraughn: Ah, I missed that.

Resolving.

scripts/irods/lib.py
korydraughn: It looks correct to me as well.

Add a comment to those functions stating the `path` is expected to be an absolute logical path. I think that's the only part that's missing.

FifthPotato: Done.

scripts/irods/lib.py
```diff
@@ -576,6 +576,29 @@ def metadata_attr_with_value_exists(session, attr, value):
     print(out)
     return value in out
 
+# Checks data object at "path" for metadata with attribute "attr" and value "value"
```
korydraughn:
```suggestion
# Checks data object at "path" for metadata with attribute "attr" and value "value".
```

scripts/irods/lib.py
```diff
@@ -576,6 +576,29 @@ def metadata_attr_with_value_exists(session, attr, value):
...
+def metadata_attr_with_value_exists_on_data_object(session, attr, value, path):
+    print('looking for attr:[{0}] value:[{1}] on data object [{2}]'.format(attr, value, path))
+
+    out, _, _ = session.run_icommand(['iquest', '%s',
+            'select META_DATA_ATTR_VALUE where META_DATA_ATTR_NAME = \'{}\' and COLL_NAME = \'{}\' and DATA_NAME = \'{}\''.format(attr, *path.rsplit('/', 1))]),
+
+    print(out)
+    return value in out
+
+# Checks collection at "path" for metadata with attribute "attr" and value "value"
```
korydraughn:
```suggestion
# Checks collection at "path" for metadata with attribute "attr" and value "value".
```

korydraughn (commented):
Once those last two review comments are resolved, run the new test to confirm it still passes.

Let us know the results. Expecting this to land today.

FifthPotato:
`test_logical_quotas` passes for both NREP and PREP. Also did clang-format.

korydraughn:
Squash to taste.

FifthPotato:
Squished.

korydraughn (approved):
Pound it.

FifthPotato:
Octothorpe'd.

korydraughn:
Merging.

Issue #8993, opened 2026-05-27
ObjStat API can return info for intermediate replicas


alanking:
https://github.com/irods/irods/blob/6dc24af5f43eb0576add6a6c36f300a4bc9c15fd/server/api/src/rsObjStat.cpp#L277-L315

If I am reading the above code snippet correctly, the ObjStat API is looking for the first non-stale replica to act as the "representative" replica for a data object. The comparison on L278 suggests that not only are good replicas considered, but intermediate and locked replicas as well.

I'm certain that the original intent here was to find good replicas, but the introduction of additional replica statuses in 4.2.9 means this code may no longer work as intended. Let's update this to only look at good replicas.

alanking:
This isn't a big deal because, if logical locking is working as intended, you will never have a situation where one replica is good and another replica is intermediate or locked. However, it is more correct to specifically check for good replica statuses here.

korydraughn:
@MartinFlores751 Close if complete.

korydraughn:
@MartinFlores751 I don't see your Catch2 test listed in https://github.com/irods/irods/blob/16f2f8d4cf8f12d89616c657ba2b8a92b42e875b/unit_tests/unit_tests_list.json.

Please confirm your new ObjStat API tests run via the testing environment.

PR #9071, opened 2026-09-08
[#8993] Only use good replicas for stat when possible (main)

MartinFlores751:
Issue quick link: #8993 

This PR introduces a change to make ObjStat prefer using information from good replicas where possible. A majority of the changes introduced are from a new unit test asserting the behavior of `rcObjStat(...)`.

### DIFF
--- a/unit_tests/src/test_data_obj_stat_api.cpp
+++ b/unit_tests/src/test_data_obj_stat_api.cpp
@@ -248,7 +248,6 @@ TEST_CASE_METHOD(TestFixture, "Stat on data object with mixed stale and good rep
 
     // We expect the good replica size
     REQUIRE(res->objSize == 0);
-    REQUIRE(res->objSize != bad_size);
 }
 
 TEST_CASE_METHOD(TestFixture, "Stat on data object with only stale replicas")
@@ -264,7 +263,6 @@ TEST_CASE_METHOD(TestFixture, "Stat on data object with only stale replicas")
 
     // We expect the first replica to give the stat when both replicas are stale
     REQUIRE(res->objSize == bad_size_one);
-    REQUIRE(res->objSize != bad_size_two);
 }
 
 TEST_CASE_METHOD(TestFixture, "Stat on data object with invalid status")
@@ -278,7 +276,6 @@ TEST_CASE_METHOD(TestFixture, "Stat on data object with invalid status")
 
     // We expect to have the size of the good replica
     REQUIRE(res->objSize == 0);
-    REQUIRE(res->objSize != bad_size);
 }
 
 TEST_CASE_METHOD(TestFixture, "Stat on data object with only invalid status")
@@ -298,5 +295,4 @@ TEST_CASE_METHOD(TestFixture, "Stat on data object with only invalid status")
 
     // We expect the first replica to give the stat when both replicas are stale
     REQUIRE(res->objSize == bad_size_one);
-    REQUIRE(res->objSize != bad_size_two);
 }
--- a/unit_tests/src/test_data_obj_stat_api.cpp
+++ b/unit_tests/src/test_data_obj_stat_api.cpp
@@ -66,12 +66,12 @@ auto set_replica_status(RcComm& _comm,
     });
 
     // Specify the data object we want to mess with
-    DataObjInfo info_two{};
-    std::strncpy(static_cast<char*>(info_two.objPath), _path.c_str(), MAX_NAME_LEN - 1);
-    info_two.replNum = replica;
+    DataObjInfo info{};
+    std::strncpy(static_cast<char*>(info.objPath), _path.c_str(), MAX_NAME_LEN - 1);
+    info.replNum = replica;
 
     // Create the required input
-    ModDataObjMetaInp inp_two{&info_two, kvp.get()};
+    ModDataObjMetaInp inp_two{&info, kvp.get()};
     return rcModDataObjMeta(&_comm, &inp_two);
 }
 
--- a/unit_tests/src/test_data_obj_stat_api.cpp
+++ b/unit_tests/src/test_data_obj_stat_api.cpp
@@ -303,7 +303,7 @@ TEST_CASE_METHOD(TestFixture, "Stat on data object with only invalid status")
     REQUIRE(set_replica_status(
                 comm, test_data_object, 1, bad_status_two, {{DATA_SIZE_KW, std::to_string(bad_size_two)}}) >= 0);
 
-    REQUIRE(irods::experimental::replica::replica_status(comm, test_data_object, 0) == bad_size_one);
+    REQUIRE(irods::experimental::replica::replica_status(comm, test_data_object, 0) == bad_status_one);
     REQUIRE(irods::experimental::replica::replica_status(comm, test_data_object, 1) == bad_status_two);
 
     auto res{stat(conn, test_data_object)};
--- a/unit_tests/src/test_data_obj_stat_api.cpp
+++ b/unit_tests/src/test_data_obj_stat_api.cpp
@@ -1,5 +1,8 @@
 #include <catch2/catch_all.hpp>
 
+#include <boost/uuid.hpp>
+
+#include "boost/uuid/random_generator.hpp"
 #include "irods/client_connection.hpp"
 #include "irods/connection_pool.hpp"
 #include "irods/data_object_modify_info.h"
@@ -11,22 +14,27 @@
 #include "irods/irods_pack_table.hpp"
 #include "irods/rodsClient.h"
 #include "irods/rodsErrorTable.h"
+#include "irods/touch.h"
 #include "irods/transport/default_transport.hpp"
 
-TEST_CASE("data_object_modify_info")
-{
-    namespace fs = irods::experimental::filesystem;
-    namespace io = irods::experimental::io;
+#include "irods/resource_administration.hpp"
 
-    auto& api_table = irods::get_client_api_table();
-    auto& pack_table = irods::get_pack_table();
-    init_api_table(api_table, pack_table);
+#include <array>
+#include <climits>
+#include <filesystem>
 
-    auto conn_pool = irods::make_connection_pool();
-    auto conn = conn_pool->get_connection();
+namespace fs = irods::experimental::filesystem;
+namespace io = irods::experimental::io;
+namespace adm = irods::experimental::administration;
+
+TEST_CASE("data_obj_stat_api")
+{    
+    load_client_api_plugins();
 
     rodsEnv env;
     _getRodsEnv(env);
+
+    irods::experimental::client_connection conn;
 
     const auto sandbox = fs::path{env.rodsHome} / "irods_unit_tests_sandbox";
     const auto path = sandbox / "dstream_data_object.txt";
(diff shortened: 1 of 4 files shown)


unit_tests/src/test_data_obj_stat_api.cpp
```diff
@@ -0,0 +1,303 @@
+#include <catch2/catch_all.hpp>
+
+#include "irods/client_connection.hpp"
+#include "irods/connection_pool.hpp"
```
korydraughn: This header file does not appear to be used.

unit_tests/src/test_data_obj_stat_api.cpp
```diff
@@ -0,0 +1,303 @@
...
+#include "irods/rodsClient.h"
+#include "irods/rodsDef.h"
+#include "irods/rodsErrorTable.h"
+#include "irods/rodsKeyWdDef.h"
+#include "irods/rodsType.h"
+#include "irods/touch.h"
+#include "irods/transport/default_transport.hpp"
+
+#include <boost/asio/ip/host_name.hpp>
+#include "boost/uuid/random_generator.hpp"
```
korydraughn:
```suggestion
#include <boost/uuid/random_generator.hpp>
```

unit_tests/src/test_data_obj_stat_api.cpp
```diff
@@ -0,0 +1,303 @@
...
+    dataObjInp_t input{};
+    std::strncpy(static_cast<char*>(input.objPath), _path.c_str(), std::strlen(_path.c_str()));
+
+    rodsObjStat* output{};
+
+    REQUIRE(rcObjStat(&_comm, &input, &output) >= 0);
+    return {output, freeRodsObjStat};
+}
+
+auto set_replica_status(RcComm& _comm,
```
korydraughn, line 56: Let's use a word other than "status". That word has a very specific meaning and can confuse the reader.

unit_tests/src/test_data_obj_stat_api.cpp
```diff
@@ -0,0 +1,303 @@
...
+    auto [kvp, lm] = irods::experimental::make_key_value_proxy();
+    kvp[REPL_STATUS_KW] = std::to_string(status);
+    kvp[ADMIN_KW] = "";
+
+    std::for_each(std::cbegin(_additional_kvp_args), std::cend(_additional_kvp_args), [&kvp](auto& _thing) {
+        kvp[_thing.first] = _thing.second;
+    });
+
+    // Specify the data object we want to mess with
+    DataObjInfo info_two{};
```
korydraughn: What does the `_two` mean?

unit_tests/src/test_data_obj_stat_api.cpp
```diff
@@ -0,0 +1,303 @@
...
+
+    irods::experimental::client_connection conn;
+
+    const auto sandbox = fs::path{static_cast<char*>(env.rodsHome)} / "irods_unit_tests_sandbox";
+    const auto path = sandbox / "dstream_data_object.txt";
+
+    fs::client::create_collection(conn, sandbox);
+
+    irods::at_scope_exit cleanup{
+        [&] { fs::client::remove_all(conn.operator RcComm&(), sandbox, fs::remove_options::no_trash); }};
```
korydraughn: Please use `static_cast`. That is easier for new readers to understand and more common.

unit_tests/src/test_data_obj_stat_api.cpp
```diff
@@ -0,0 +1,303 @@
...
+    DataObjInfo info_two{};
+    std::strncpy(static_cast<char*>(info_two.objPath), _path.c_str(), MAX_NAME_LEN - 1);
+    info_two.replNum = replica;
+
+    // Create the required input
+    ModDataObjMetaInp inp_two{&info_two, kvp.get()};
+    return rcModDataObjMeta(&_comm, &inp_two);
+}
+
+TEST_CASE("Stat on single data object")
```
korydraughn, line 80: The description of the test needs to cover its goal. For cases where the the description is very long, resort to code comments for capturing the goal.

Here's my attempt.

```
rcObjStat of data object with single good replica returns info about single good replica
```

unit_tests/src/test_data_obj_stat_api.cpp
```diff
@@ -0,0 +1,303 @@
...
+
+    // Create a data object in iRODS.
+    // This is used in all future sections.
+    {
+        io::client::default_transport transport{conn};
+        io::odstream out{transport, path};
+    }
+
+    auto res{stat(conn, path)};
+    REQUIRE(res->objSize == 0);
```
korydraughn, line 104: Let's assert the status of the replica is good and the replica number.

korydraughn, line 104: Assert `res` is valid before deref.

unit_tests/src/test_data_obj_stat_api.cpp
```diff
@@ -0,0 +1,303 @@
...
+}
+
+struct TestFixture
```
korydraughn, line 107:
```suggestion
struct test_fixture
```

Consider including the issue number in the name. Or maybe in the string that describes the test case.

korydraughn, line 107: Please document what this produces. No need for doxygen style comments for this.

See the following for an example.

https://github.com/irods/irods/blob/81df38807bc406a239718f70b31d6f58321970b5/scripts/irods/test/test_federation.py#L1804-L1822

unit_tests/src/test_data_obj_stat_api.cpp
```diff
@@ -0,0 +1,303 @@
...
+        // Cleanup the resources
+        adm::client::remove_resource(conn, res_regis_pt.resource_name);
+        adm::client::remove_resource(conn, res_regis_repl.resource_name);
+        adm::client::remove_resource(conn, res_regis_b.resource_name);
+        adm::client::remove_resource(conn, res_regis_a.resource_name);
+
+        // Cleanup the temp directories
+        std::filesystem::remove_all(res_b_path);
+        std::filesystem::remove_all(res_a_path);
+    }
```
korydraughn: Probably need to wrap all of this in a try-catch block since destructors are not allowed to throw exceptions.

unit_tests/src/test_data_obj_stat_api.cpp
```diff
@@ -0,0 +1,303 @@
...
+                       .resource_type = adm::resource_type::unixfilesystem,
+                       .host_name = hostname,
+                       .vault_path = res_b_path.string()};
+        REQUIRE_NOTHROW(adm::client::add_resource(conn, res_regis_b));
+        res_regis_repl = {
+            .resource_name = fmt::format("repl-{}", test_uui), .resource_type = adm::resource_type::replication};
+        REQUIRE_NOTHROW(adm::client::add_resource(conn, res_regis_repl));
+        res_regis_pt = {
+            .resource_name = fmt::format("pt-{}", test_uui), .resource_type = adm::resource_type::passthrough};
+        REQUIRE_NOTHROW(adm::client::add_resource(conn, res_regis_pt));
```
korydraughn: I think you can drop the passthru resource.

unit_tests/src/test_data_obj_stat_api.cpp
```diff
@@ -0,0 +1,303 @@
...
+
+    REQUIRE(rcObjStat(&_comm, &input, &output) >= 0);
+    return {output, freeRodsObjStat};
+}
+
+auto set_replica_status(RcComm& _comm,
+                        const fs::path& _path,
+                        int replica,
+                        int status,
+                        const std::unordered_map<std::string, std::string>& _additional_kvp_args) -> int
```
korydraughn, line 60: What's the motivation behind handling `status` independently of `_additional_kvp_args`?

unit_tests/src/test_data_obj_stat_api.cpp
```diff
@@ -0,0 +1,303 @@
...
+{
+    constexpr rodsLong_t bad_size{10};
+    auto& comm{static_cast<RcComm&>(conn)};
+    REQUIRE(set_replica_status(comm, test_file, 0, STALE_REPLICA, {{DATA_SIZE_KW, std::to_string(bad_size)}}) >= 0);
+
+    auto res{stat(conn, test_file)};
+
+    // We expect the good replica size
+    REQUIRE(res->objSize == 0);
+    REQUIRE(res->objSize != bad_size);
```
korydraughn: Is L252 needed?

unit_tests/src/test_data_obj_stat_api.cpp
```diff
@@ -0,0 +1,303 @@
...
+    REQUIRE(set_replica_status(comm, test_file, 0, STALE_REPLICA, {{DATA_SIZE_KW, std::to_string(bad_size_one)}}) >= 0);
+
+    constexpr rodsLong_t bad_size_two{20};
+    REQUIRE(set_replica_status(comm, test_file, 1, STALE_REPLICA, {{DATA_SIZE_KW, std::to_string(bad_size_two)}}) >= 0);
+
+    auto res{stat(conn, test_file)};
+
+    // We expect the first replica to give the stat when both replicas are stale
+    REQUIRE(res->objSize == bad_size_one);
+    REQUIRE(res->objSize != bad_size_two);
```
korydraughn: Is L268 needed?

unit_tests/src/test_data_obj_stat_api.cpp
```diff
@@ -0,0 +1,303 @@
...
+    constexpr rodsLong_t bad_size{10};
+    constexpr auto bad_status{42};
+    auto& comm{static_cast<RcComm&>(conn)};
+    REQUIRE(set_replica_status(comm, test_file, 0, bad_status, {{DATA_SIZE_KW, std::to_string(bad_size)}}) >= 0);
+
+    auto res{stat(conn, test_file)};
+
+    // We expect to have the size of the good replica
+    REQUIRE(res->objSize == 0);
+    REQUIRE(res->objSize != bad_size);
```
korydraughn: Is L282 needed?

unit_tests/src/test_data_obj_stat_api.cpp
```diff
@@ -0,0 +1,303 @@
...
+    constexpr rodsLong_t bad_size_two{20};
+    constexpr auto bad_status_two{56709};
+    REQUIRE(set_replica_status(comm, test_file, 1, bad_status_two, {{DATA_SIZE_KW, std::to_string(bad_size_two)}}) >=
+            0);
+
+    auto res{stat(conn, test_file)};
+
+    // We expect the first replica to give the stat when both replicas are stale
+    REQUIRE(res->objSize == bad_size_one);
+    REQUIRE(res->objSize != bad_size_two);
```
korydraughn: Is L302 needed?

korydraughn (commented):
Looking good so far.

Please apply the comments regarding test descriptions and pointer dereferences to all test cases.

unit_tests/src/test_data_obj_stat_api.cpp
MartinFlores751: What could "status" mean here in the context of replicas? It's mirroring the function of the Python test suite `set_replica_status(...)` save for the last new argument.

unit_tests/src/test_data_obj_stat_api.cpp
MartinFlores751: Is there a case where `REQUIRE(rcObjStat(&_comm, &input, &output) >= 0);` is true but `output` is set to `nullptr`? If not, then the check is already done in the `stat(...)` call.

unit_tests/src/test_data_obj_stat_api.cpp
MartinFlores751: To match the `set_replica_status(...)` Python function closely. Also to have `_additional_kvp_args` only be for the `kvp`. Selecting the replica seems to be done with `DataObjInfo`, not the `kvp`.

unit_tests/src/test_data_obj_stat_api.cpp
MartinFlores751: Removed.

unit_tests/src/test_data_obj_stat_api.cpp
MartinFlores751: Done.

unit_tests/src/test_data_obj_stat_api.cpp
MartinFlores751: Copy-paste artifact. Removed.

unit_tests/src/test_data_obj_stat_api.cpp
MartinFlores751: Removed `.operator RcComm&()`. Used implicit conversion.

unit_tests/src/test_data_obj_stat_api.cpp
MartinFlores751: Removed.

unit_tests/src/test_data_obj_stat_api.cpp
MartinFlores751: Removed.

unit_tests/src/test_data_obj_stat_api.cpp
MartinFlores751: Removed.

unit_tests/src/test_data_obj_stat_api.cpp
MartinFlores751: Removed.

unit_tests/src/test_data_obj_stat_api.cpp
MartinFlores751: Removed the passthru resource.

unit_tests/src/test_data_obj_stat_api.cpp
MartinFlores751: Added try-catch block.

unit_tests/src/test_data_obj_stat_api.cpp
korydraughn: I forgot to remove that comment. It wasn't until later that I noticed it does set the replica status.

Resolving.

unit_tests/src/test_data_obj_stat_api.cpp
korydraughn: > Is there a case where `REQUIRE(rcObjStat(&_comm, &input, &output) >= 0);` is true but `output` is set to `nullptr`?

I don't think so, but if there is, the assertion would make it very clear to us.

> If not, then the check is already done in the `stat(...)` call.

You sure? I didn't see a check for nullptr in `stat`. Do you have a line number?

unit_tests/src/test_data_obj_stat_api.cpp
korydraughn: Got it.

You've confirmed that using `.replNum` targets the correct replica?

unit_tests/src/test_data_obj_stat_api.cpp
MartinFlores751: > I don't think so, but if there is, the assertion would make it very clear to us.

Fair enough, will add an extra assertion.

> You sure? I didn't see a check for nullptr in `stat`. Do you have a line number?

I didn't use enough words. I meant if `rcObjStat(...)` only returns `nullptr` with a negative status code, then the null check is effectively performed by the `REQUIRE(... >= 0)` before returning the `unique_ptr`.

unit_tests/src/test_data_obj_stat_api.cpp
MartinFlores751: > Got it.
> 
> You've confirmed that using `.replNum` targets the correct replica?

I've based the `.replNum` off the `iadmin` code. 

There's more code above the highlighted portion, but this should be what's relevant: https://github.com/irods/irods/blob/81df38807bc406a239718f70b31d6f58321970b5/clients/icommands/src/iadmin.cpp#L421-L447

Still, I'll add some additional checks to ensure the replicas are being handled correctly. I believe this comment is relevant: https://github.com/irods/irods/pull/9071#discussion_r3969156110

unit_tests/src/test_data_obj_stat_api.cpp
MartinFlores751: Done.

MartinFlores751: All replicas are asserted before the `stat(...)` call now. Resolving.

unit_tests/src/test_data_obj_stat_api.cpp
MartinFlores751: Linked comment resolved. Resolving.

unit_tests/src/test_data_obj_stat_api.cpp
MartinFlores751: Updated the test names. Resolving.

unit_tests/src/test_data_obj_stat_api.cpp
MartinFlores751: Done.

MartinFlores751: Done.

unit_tests/src/test_data_obj_stat_api.cpp
```diff
@@ -104,7 +104,21 @@ TEST_CASE("Stat on single data object")
...
 
-struct TestFixture
+// This structure provides the setup and teardown for tests that
+// request its use.
+//
+// The structure provides:
+//  - a connection that may be used throughout the test
+//  - a UUID to help prevent naming collisions
+//  - a test specific collection (sandbox)
+//  - a simple resource heirarchy, replication with two unix resources beneath it
```
korydraughn: heirarchy -> hierarchy

korydraughn (commented):
Awaiting results of unit and core tests.

unit_tests/src/test_data_obj_stat_api.cpp
MartinFlores751: Fixed.

MartinFlores751:
Rebased onto tip of main. Running core and unit tests.

MartinFlores751:
Ran the core and unit tests. All tests passed.

korydraughn:
Nice. Squash to taste.

MartinFlores751:
Squashed.

korydraughn (approved):
Pound it.

The commit message body talks about tests so I think you can drop the part about tests on the summary line.

MartinFlores751:
Updated the commit message. Pounded.

PR #9078, opened 2026-09-15
Plugin improvements - logging, regexes, UB (main)

korydraughn:
Related to irods/irods_rule_engine_plugin_python#268.

The change to `RuleExistsHelper` is an ABI break, but given that ABI stability isn't a guarantee and the change results in a safer implementation, I think it's acceptable.

Will run tests soon.


korydraughn:
Clang-tidy is complaining about unused function parameters and the use of `static`. Will address those later.

All unit tests pass.

Running core tests.

korydraughn:
All core tests passed.

trel (approved):
nothing jumps out to me.   tests pass.  i say get it in.

